PR #889, opened 2021-07-22
Make ScriptSession a LivenessScope instead of LivenessNode

nbauernfeind:


grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionState.java
```diff
@@ -472,7 +473,7 @@ private synchronized void setDependencies(final List<ExportObject<?>> parents) {
 
             this.parents = parents;
             dependentCount = parents.size();
-            parents.forEach(this::manage);
+            parents.stream().filter(Objects::nonNull).forEach(this::manage);
```
rcaudy, line 476: How we would have null parents ?

rcaudy (approved):

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionState.java
nbauernfeind: As discussed offline; for optional dependencies if doing so is convenient. An alternative impl for the ConsoleService's bindTableToVariable looks like this:

```java
    @Override
    public void bindTableToVariable(BindTableToVariableRequest request, StreamObserver<BindTableToVariableResponse> responseObserver) {
            final SessionState.ExportObject<Table> exportedTable = ticketRouter.resolve(session, request.getTableId());
            final SessionState.ExportObject<ScriptSession> exportedConsole = !request.hasConsoleId() ? null
                    : ticketRouter.resolve(session, request.getConsoleId());

            session.nonExport()
                    .requiresSerialQueue()
                    .require(exportedTable, exportedConsole)
                    .onError(responseObserver::onError)
                    .submit(() -> {
                        ScriptSession scriptSession = exportedConsole != null ? exportedConsole.get() : globalSessionProvider.getGlobalSession();
                        Table table = exportedTable.get();
                        scriptSession.setVariable(request.getVariableName(), table);
                        scriptSession.manage(table);
                        responseObserver.onNext(BindTableToVariableResponse.getDefaultInstance());
                        responseObserver.onCompleted();
                    });
        });
    }
```

PR #891, opened 2021-07-22
fixes #880 listFlights is now compatible with proper FlightClient

### LINKED ISSUE #880: listFlights client IndexOutOfBoundsException
Calling listFlights is causing the following on my client:

```
java.lang.IndexOutOfBoundsException
        at java.nio.Buffer.checkIndex(Buffer.java:551)
        at java.nio.HeapByteBuffer.getInt(HeapByteBuffer.java:379)
        at com.google.flatbuffers.Table.__reset(Table.java:301)
        at org.apache.arrow.flatbuf.Message.__init(Message.java:31)
        at org.apache.arrow.flatbuf.Message.__assign(Message.java:32)
        at org.apache.arrow.flatbuf.Message.getRootAsMessage(Message.java:30)
        at org.apache.arrow.flatbuf.Message.getRootAsMessage(Message.java:29)
        at org.apache.arrow.vector.ipc.message.MessageSerializer.readMessage(MessageSerializer.java:706)
        at org.apache.arrow.vector.ipc.message.MessageSerializer.deserializeSchema(MessageSerializer.java:213)
        at org.apache.arrow.flight.FlightInfo.<init>(FlightInfo.java:98)
        at org.apache.arrow.flight.FlightClient.lambda$listFlights$0(FlightClient.java:130)
        at org.apache.arrow.flight.grpc.StatusUtils$1.next(StatusUtils.java:223)
        at io.deephaven.client.examples.ListFlights.call(ListFlights.java:50)
        at io.deephaven.client.examples.ListFlights.call(ListFlights.java:23)
        at io.deephaven.client.examples.ListFlights.main(ListFlights.java:61)
```

This is happening after I bind a variable via the web ui:

```groovy
x = emptyTable(1).view("X=42L")
```
succeeds.

And then the listFlights breaks.

Note: listFlights against an empty server works.

nbauernfeind:


grpc-api/src/main/java/io/deephaven/grpc_api/arrow/FlightServiceGrpcImpl.java
```diff
@@ -460,13 +460,12 @@ private void sealAndExport() {
...
+                resultTable.sealTable(() -> resultExportBuilder.submit(() -> {
                     // transfer ownership to submit's liveness scope, drop our extra reference
                     resultTable.manageWithCurrentScope();
                     resultTable.dropReference();
                     return resultTable;
-                });
+                }));
 
                 observer.onCompleted();
                 onRequestDone();
```
devinrsmith, line 471: I'm not familiar w/ the full code-paths; but typically, I see the onCompleted() calls being done inside the submit runnable. Is it correct to call onCompleted() outside of it here? (Also, I assume an onError is attached to the resultExportBuilder).

nbauernfeind: This specific example is a little confusing because there are several different things going on at once.

I'm going to start with the simple cases you're talking about:
```
session.newExport(id)
    .require(dependency)
    .submit(() -> {
        observer.onNext(dependency.get().toProtobufType());
        observer.onComplete();
    });
```

You can't call onComplete until after having sent all data that you will send to that observer. Most of the time the data you want to send isn't available until after it has been resolved via the session's dependency mechanisms.

In the above example, that observer is supposed to ping a message every time it processes a Schema or RecordBatch from the user -- that way they know approximately how much is uploaded and whether or not it matches their expectations. Once we've processed the last RecordBatch we do not need that observer anymore.

However, I see your valid point that an error in resultExportBuilder won't get propagated to the user through that observer. They will find out about the failure through other mechanisms (dependency propagation, the export listener stream). But you're right, we should complete inside of the submit callback -- and the user will know when it's available if we do it that way.

devinrsmith (approved):

grpc-api/src/main/java/io/deephaven/grpc_api/arrow/FlightServiceGrpcImpl.java
```diff
@@ -460,15 +462,14 @@ private void sealAndExport() {
                 }
 
                 // no more changes allowed; this is officially static content
-                resultTable.sealTable();
-                resultExportBuilder.submit(() -> {
+                resultTable.sealTable(() -> resultExportBuilder.submit(() -> {
                     // transfer ownership to submit's liveness scope, drop our extra reference
                     resultTable.manageWithCurrentScope();
                     resultTable.dropReference();
+                    observer.onCompleted();
```
niloc132, line 469: wrap this up so we can still export if network dies at this point? the table still exists and is sealed and exported, and the user will hopefully be back shortly

grpc-api-client/src/main/java/io/deephaven/grpc_api_client/table/BarrageSourcedTable.java
```diff
@@ -436,6 +441,10 @@ private synchronized void realRefresh() {
         }
 
         if (sealed) {
+            if (onSealRunnable != null) {
+                onSealRunnable.run();
```
niloc132, line 445: probably stupid question, but what if we seal but never call this, is that possible? seal racing with release for example?

and what if the runnable throws? (for example, the export func runs synchronously, and onCompleted throws)

grpc-api/src/main/java/io/deephaven/grpc_api/arrow/FlightServiceGrpcImpl.java
```diff
@@ -238,11 +238,11 @@ public void doPutCustom(final InputStream request, final StreamObserver<Flight.P
             }
             final BarragePutMetadata app_metadata = mi.app_metadata;
             if (app_metadata == null) {
-                throw GrpcUtil.statusRuntimeException(Code.INVALID_ARGUMENT, "no app_metadata provided");
+                throw GrpcUtil.statusRuntimeException(Code.INVALID_ARGUMENT, "No app_metadata provided");
             }
             final ByteBuffer ticketBuf = app_metadata.rpcTicketAsByteBuffer();
             if (ticketBuf == null) {
-                throw GrpcUtil.statusRuntimeException(Code.INVALID_ARGUMENT, "no rpc ticket provided");
+                throw GrpcUtil.statusRuntimeException(Code.INVALID_ARGUMENT, "No rpc ticket provided");
```
niloc132, line 245: what's the goal in this? it isn't possible for rpcTicketAsByteBuffer() to return null

niloc132 (approved):

grpc-api/src/main/java/io/deephaven/grpc_api/table/TableServiceGrpcImpl.java
```diff
@@ -290,10 +290,8 @@ public void exportedTableUpdates(final ExportedTableUpdatesRequest request, fina
     }
 
     public static ExportedTableCreationResponse buildTableCreationResponse(final TableReference tableRef, final Table table) {
-        final String[] columnNames = table.getDefinition().getColumnNamesArray();
-        final ColumnSource<?>[] columnSources = table.getColumnSources().toArray(ColumnSource.ZERO_LENGTH_COLUMN_SOURCE_ARRAY);
         final FlatBufferBuilder builder = new FlatBufferBuilder();
-        builder.finish(BarrageSchemaUtil.makeSchemaPayload(builder, table.getDefinition(), table.getAttributes()));
+        builder.finish(BarrageSchemaUtil.makeSchemaPayload(builder, table));
```
niloc132, line 294: let's fix this too, so clients only have one way to decode schemas.

JS API change, in ClientTableState.java, roughly line 983 - this is the current code:
```
    public void applyTableCreationResponse(ExportedTableCreationResponse def) {
        // by definition, the ticket is now exported and connected
        handle.setState(TableTicket.State.EXPORTED);
        handle.setConnected(true);

        Schema schema = Schema.getRootAsSchema(new io.deephaven.javascript.proto.dhinternal.flatbuffers.ByteBuffer(def.getSchemaHeader_asU8()));
        ```
        
This should instead read a Message, confirm there is a Schema inside, read the Schema, and then pass that along. Here's how we read BarrageRecordBatches in that way, from WorkerConnection.java, around line 1030:

                    Message headerMessage = Message.getRootAsMessage(new io.deephaven.javascript.proto.dhinternal.flatbuffers.ByteBuffer(data.getDataHeader_asU8()));
                    if (body.limit() == 0 && headerMessage.headerType() != MessageHeader.BarrageRecordBatch) {
                        // a subscription stream presently ignores schemas and other message types
                        return;
                    }
                    BarrageRecordBatch header = headerMessage.header(new BarrageRecordBatch());

nbauernfeind: good idea: https://github.com/deephaven/deephaven-core/pull/926

PR #893, opened 2021-07-23
Add deephaven types submodule for convenience when using DynamicTable…

jjbrosnan:
Without this addition, users must repeatedly use jpy.get_type() when using DynamicTableWriter.  The deephaven.types submodule "aliases" jpy.get_type to something that is nicer


hythloda (approved):
I love it!  Not sure where to put this in the whole package though.

jmao-denver:
would be nice to see a before/after code example in the description

Integrations/python/deephaven/types.py
```diff
@@ -0,0 +1,43 @@
...
+"""
+Functions to return types from jpy for convenience
+"""
+
+import jpy
+
+# Java primitives
+
+def Byte():
+    return jpy.get_type("byte")
```
jmao-denver, line 14: not super important, but we can avoid repeated call to the function by creating a global constant to represent a Java type.

jjbrosnan:
Now has global variables instead of function calls.

Without deephaven.types:

```
from deephaven import jpy
import random, string, threading, time
from deephaven import types as dht

# Fetch the object
DynamicTableWriter = jpy.get_type("io.deephaven.db.v2.utils.DynamicTableWriter")

# Create the object
tableWriter = DynamicTableWriter(["A", "B"], [jpy.get_type("int"), jpy.get_type("java.lang.String")])

# Set the name of the table that the DynamicTableWriter will write to
t = tableWriter.getTable()

# Define the function to make a live, ticking table
def thread_func():
    for i in range(10):
        a = random.randint(1, 100)
        b = random.choice(string.ascii_letters)
        # The logRow method adds a row to the table
        tableWriter.logRow(a, b)
        time.sleep(1)

# Create the thread and write to the live table
thread = threading.Thread(target = thread_func)
thread.start()
```

With deephaven.types:

```
from deephaven import jpy
import random, string, threading, time
from deephaven import types as dht

# Fetch the object
DynamicTableWriter = dht.DynamicTableWriter

# Create the object
tableWriter = DynamicTableWriter(["A", "B"], [dht.Int, dht.String])

# Set the name of the table that the DynamicTableWriter will write to
t = tableWriter.getTable()

# Define the function to make a live, ticking table
def thread_func():
    for i in range(10):
        a = random.randint(1, 100)
        b = random.choice(string.ascii_letters)
        # The logRow method adds a row to the table
        tableWriter.logRow(a, b)
        time.sleep(1)

# Create the thread and write to the live table
thread = threading.Thread(target = thread_func)
thread.start()
```

jmao-denver:
It seems a bit strange to put DynamicTableWriter in the types module. I believe it fits better in TableTools.

jmao-denver (approved):

jmao-denver (dismissed):

Integrations/python/deephaven/types.py
```diff
@@ -0,0 +1,25 @@
...
+
+"""
+Functions to return types from jpy for convenience
+"""
+
+import jpy
+
+# Java primitives
+
+Bool = jpy.get_type("boolean")
```
chipkent, line 13: I am concerned that we might not be able to do it this simple way.  This would need to be confirmed with Devin.  

If you look in https://github.com/deephaven/deephaven-core/blob/deephaven-types/Integrations/python/deephaven/__init__.py you see: `# NB: this must be defined BEFORE importing .jvm_init or .start_jvm (circular import)`.  I'm concerned that (1) the simple case may not work because of JVM initialization problems or (2) the simple case does work, so we need to clean up the code.

See also cases like this https://github.com/deephaven/deephaven-core/blob/deephaven-types/Integrations/python/deephaven/TableTools/__init__.py, where `_defineSymbols` is used to defer the initialization.

Either way, you need to discuss with @devinrsmith and fire up the code to make sure that things work as expected.

chipkent (changes requested):

jmao-denver (approved):

jjbrosnan:
In Integrations/python/deephaven/__init__.py, `if jpy.has_jvm():` is needed.  When the checks are run, they are not run where a JVM is present, so they fail despite having no issues in deephaven.  Either way, this also works with no issues (that I've been able to find yet).

Integrations/python/deephaven/__init__.py
```diff
@@ -136,6 +135,8 @@ def initialize():
 Figure = None  # variable for initialization
 PlottingConvenience = None  # variable for initialization
 
+if jpy.has_jvm():
+    DynamicTableWriter = jpy.get_type("io.deephaven.db.v2.utils.DynamicTableWriter")
```
chipkent, line 139: My concern is that this is not done in the same way as all of the other imported types.  Why is this being done differently than `PythonFunction`, `PythonListenerAdapter`, etc.?

Integrations/python/deephaven/types.py
```diff
@@ -0,0 +1,29 @@
...
+
+"""
+Functions to return types from jpy for convenience
+"""
+
+import jpy
+
+# Raise a SystemError if jpy is not aware of the JVM yet
+if not jpy.has_jvm():
+    raise SystemError("No java functionality can be used until the JVM has been initialized through the jpy module")
```
chipkent, line 13: Confirming that if you start a clean console and run the following that everything works.
```
from deephaven.types import *
```

chipkent (changes requested):

Integrations/python/deephaven/types.py
jjbrosnan: Confirmed.  Just restarted everything and it works for all of the types.

Integrations/python/deephaven/__init__.py
jjbrosnan: This isn't done the same as the other methods in the module because it's different than those listed.  DynamicTableWriter is only meant to be a convenience alias for a jpy.get_type() call to fetch the DynamicTableWriter object from io.deephaven.db.v2.utils.  It shouldn't have to be any more complex than this.

chipkent: What you are saying does not make sense.

1) The assertion that `DynamicTableWriter` is different is just not true.  Look at `PythonFunction` and `PythonListenerAdapter`.  Those are just imported types.
2) `PickedResult` and `Config` are not like what you did.  All of these methods use `__initializer__`, as can be seen here:
```
def initialize():
    __initializer__(jpy.get_type("io.deephaven.configuration.Configuration"), Config)
    __initializer__(jpy.get_type("io.deephaven.db.util.PickledResult"), PickledResult)
    __initializer__(jpy.get_type("io.deephaven.integrations.python.PythonListenerAdapter"), PythonListenerAdapter)
    __initializer__(jpy.get_type("io.deephaven.integrations.python.PythonFunction"), PythonFunction)
```

This is not the same as:
```
if jpy.has_jvm():
    DynamicTableWriter = jpy.get_type("io.deephaven.db.v2.utils.DynamicTableWriter")
```

jjbrosnan:
The code in this PR is now consistent with how other submodules and methods are written in deephaven.  I have tested it and found no bugs thus far.

jmao-denver (approved):

Integrations/python/deephaven/__init__.py
```diff
@@ -23,6 +23,8 @@
 
 * TableTools imported as ttools *(`tt` is frequently used for time table)*
 
+* Types as dht
```
chipkent, line 26:
```suggestion
* Types imported as dht
```

Integrations/python/deephaven/Types/__init__.py
```diff
@@ -0,0 +1,74 @@
+# 
+# Copyright (c) 2016 - 2021 Deephaven Data Labs and Patent Pending
+# 
+
+"""
+A submodule to return types from jpy for convenience
```
chipkent, line 6: I think you can make a more clear description.

Integrations/python/deephaven/Types/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+Bool = None
+Byte = None
+Char = None
+Double = None
+Float = None
+Int = None
+Long = None
+Short = None
+String = None
+
```
chipkent, line 24: It seems like date time should be here.

Integrations/python/deephaven/Types/__init__.py
```diff
@@ -0,0 +1,74 @@
+#
```
chipkent, line 1: I have some angst about calling this `Types` because the name is not pythonic.  We have used `TableTools` and such because those are autogenerated classes that map to Java objects.  You could argue that using caps was a bad idea.
@jmao-denver what is our religion?  Are we going with caps and just being different?

Integrations/python/deephaven/TableManipulation/__init__.py
```diff
@@ -8,13 +8,15 @@
...
 
-__all__ = ['ColumnRenderersBuilder', 'DistinctFormatter', 'DownsampledWhereFilter', 'LayoutHintBuilder',
-           'SmartKey', 'SortPair', 'TotalsTableBuilder', 'WindowCheck']
+__all__ = ['ColumnRenderersBuilder', 'DBDateTime', 'DistinctFormatter', 'DownsampledWhereFilter', 'DynamicTableWriter', 
+           'LayoutHintBuilder', 'SmartKey', 'SortPair', 'TotalsTableBuilder', 'WindowCheck']
 
 # None until the first successful _defineSymbols() call
 ColumnRenderersBuilder = None   #: Class to build and parse the directive for Table.COLUMN_RENDERERS_ATTRIBUTE (io.deephaven.db.v2.ColumnRenderersBuilder).
 DistinctFormatter = None        #: Class to create distinct and unique coloration for each unique input value (io.deephaven.db.util.DBColorUtil$DistinctFormatter).
+DBDateTime = None               #: Class to create a DBDateTime instance in Python (io.deephaven.db.tables.utils.DBDateTime).
```
chipkent, line 17: This doesn't seem to belong here.  This module is documented to be `Functionality to display and modify tables.`  I don't see how a date time does that.  It makes far more sense in `Types`.

chipkent (changes requested):

chipkent (changes requested):

jcferretti:
We just merged https://github.com/deephaven/deephaven-core/pull/1088
This borrowed some parts of your idea.
Previously discussed here: https://illumon.slack.com/archives/G01K6UHJ00K/p1629394210482000.

jjbrosnan:
This branch is outdated and the content has been added via Cristian's that got merged yesterday and PR 1117 that got merged today.

PR #894, opened 2021-07-23
take an initial snapshot in QueryTable#snapshot if requested or if left table is static

### LINKED ISSUE #824: Snapshot does not work OOTB with static tables
**Description**

The .snapshot() method does not freeze tables in time, as described in the enterprise docs [here](https://deephaven.io/enterprise/docs/onboarding/become-an-expert/use-table-snapshots/#creating-a-static-snapshot).

**Steps to reproduce**

Execute the following commands:

`from deephaven import TableTools as ttools`
`data = ttools.timeTable("00:00:10").update("X=i", "Y=new int[]{0,1,2}").tail(5).ungroup()`
`static_data = ttools.emptyTable(1).snapshot(data)`

**Expected results**

I expected `static_data` to contain exactly what `data` contained at the time of creating `static_data`, in a static table. You will get this desired result if you instead use

`static_data = ttools.emptyTable(1).snapshot(data, True)`

where this second optional parameter is not described anywhere in enterprise docs.

**Actual results**

`static_data` will be an empty table with the correct schema. No data from `data` gets copied.

**Versions**
 - Deephaven: 0.1.0
 - OS: masOS Big Sur 11.4
 - Browser: Safari 14.1.1
 - Docker: 3.3.3

nbauernfeind:
Fixes #824

### DIFF
--- a/DB/src/main/java/io/deephaven/db/v2/QueryTable.java
+++ b/DB/src/main/java/io/deephaven/db/v2/QueryTable.java
@@ -1860,7 +1860,7 @@ public class QueryTable extends BaseTable {
         final SnapshotInternalListener listener = new SnapshotInternalListener(this, lazySnapshot, tableToSnapshot,
                 result, resultLeftColumns, resultRightColumns, resultIndex);
 
-        if (doInitialSnapshot) {
+        if (!isRefreshing() || doInitialSnapshot) {
             if (!isRefreshing() && tableToSnapshot.isLive() && !lazySnapshot) {
                 // if we are making a static copy of the table, we must ensure that it does not change out from under us
                 ConstructSnapshot.callDataSnapshotFunction("snapshotInternal",


DB/src/main/java/io/deephaven/db/v2/QueryTable.java
```diff
@@ -1860,7 +1860,7 @@ private Table snapshotInternal(Table tableToSnapshot, boolean doInitialSnapshot,
         final SnapshotInternalListener listener = new SnapshotInternalListener(this, lazySnapshot, tableToSnapshot,
                 result, resultLeftColumns, resultRightColumns, resultIndex);
 
-        if (doInitialSnapshot) {
+        if (!isRefreshing() || doInitialSnapshot) {
```
rcaudy, line 1863: I hate the idea of making `snapshot` have different semantics for static tables. I'm not convinced we should merge this.

nbauernfeind: Ahh, yes, I see. Then we just need to fix the documentation. Closing.

PR #897, opened 2021-07-24
make default behavior of QueryTable#snapshot take an initial snapshot

### LINKED ISSUE #824: Snapshot does not work OOTB with static tables

nbauernfeind:
Fixes #824.

### DIFF
--- a/DB/src/main/java/io/deephaven/db/tables/Table.java
+++ b/DB/src/main/java/io/deephaven/db/tables/Table.java
@@ -2287,7 +2287,7 @@ public interface Table extends LongSizedDataStructure, LivenessNode, TableOperat
     Table snapshot(Table rightTable, boolean doInitialSnapshot, String... stampColumns);
 
     default Table snapshot(Table rightTable, String... stampColumns) {
-        return snapshot(rightTable, false, stampColumns);
+        return snapshot(rightTable, true, stampColumns);
     }
 
     Table snapshotIncremental(Table rightTable, boolean doInitialSnapshot, String... stampColumns);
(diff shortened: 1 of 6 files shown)


DB/src/main/java/io/deephaven/db/tables/Table.java
```diff
@@ -2287,7 +2287,7 @@ default Table sort(Collection<SortColumn> columnsToSortBy) {
     Table snapshot(Table rightTable, boolean doInitialSnapshot, String... stampColumns);
 
     default Table snapshot(Table rightTable, String... stampColumns) {
```
rcaudy, line 2289: Can you clarify the JavaDoc on `TableOperations` to specify the default behavior? Also, please make sure @devinrsmith and the dev-rel folks know about this change, for QST and documentation respectively.

rcaudy (commented):
See comment regarding docs and communication.

rcaudy (approved):
Make sure the docs team hear's about this, and understands that it is a changer relative to DHE.

Issue #293, opened 2021-04-09
EPIC: Kafka support


mofojed:
- I want to be able to easily connect to a single or multiple Kafka topic(s).  I want to support a happy path…. where I can provide 4-5 arguments like: schema-service url, topic-url, credentials, additional filtering information related to the topic, maybe a starting sequence number.  I will expect Deephaven to handle and fully relieve me of worrying about routing tables, in-worker ingestion services, and schema creation.  

_Revising the remainder of the description: the MVP version of community's Kafka support likely will *not* have any support to ease publication._
- I want to be able to publish streaming data to Kafka topics with a simple writeKafka() capability that writes to a Kafka-interface I control (outside of DH).  I want to be able to provide minimal configuration: Credentials, an interface-url, zookeeper broker endpoint.

rcaudy:
I don't think this was closed intentionally.

rcaudy:
Closing. We've met our "MVP" Kafka goals. We're planning a September 20201 epic for Kafka publishing.

PR #917, opened 2021-07-27
WIP: stream table interfaces for discussion

rcaudy:
Just putting this up as a discussion aid.

### DIFF
--- a/DB/src/main/java/io/deephaven/stream/StreamConsumer.java
+++ b/DB/src/main/java/io/deephaven/stream/StreamConsumer.java
@@ -1,0 +1,28 @@
+package io.deephaven.stream;
+
+import io.deephaven.db.v2.sources.chunk.Attributes.Values;
+import io.deephaven.db.v2.sources.chunk.WritableChunk;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Chunk-oriented consumer for streams of data.
+ */
+@FunctionalInterface
+public interface StreamConsumer {
+
+    /**
+     * <p>Accept a batch of rows splayed across per-column {@link WritableChunk chunks} of {@link Values values}.
+     *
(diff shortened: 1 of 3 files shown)


DB/src/main/java/io/deephaven/stream/StreamConsumer.java
```diff
@@ -0,0 +1,28 @@
...
+     * needed.
+     *
+     * <p>Implementations will generally have a mechanism for determining the expected number and type of input
+     * chunks, but this is not dictated at the interface level.
+     *
+     * @param data Per-column {@link WritableChunk chunks} of {@link Values values}.
+     *             Must all have the same {@link WritableChunk#size() size}.
+     */
+    @SuppressWarnings("unchecked") // There's no actual possibility of heap-pollution, here.
+    void accept(@NotNull WritableChunk<Values>... data);
```
cpwright, line 27: No need to be writable.

rcaudy: If you want to return it to the pool, they do, unless you want an `instanceof` check for each one. You opposed making `Chunk` implement `PoolableChunk` last time it came up. Also, since they are most certainly writable at the producer, I see no reason not to make it clear that the consumer can do whatever it wants with the chunks it receives.

rcaudy: Related, it might be nice to somehow say "these are *not* resettable", since that would interfere with our ability to return them properly.

rcaudy: Actually, I take that part (about resettable) back. If we get a resettable chunk, we own the backing storage up to its capacity, and we only need to return the view to the pool. I think that is semantically clean enough for me.

rcaudy:
Link to #293

cpwright (approved):

PR #921, opened 2021-07-27
Chunk column source.

cpwright:


DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/CharChunkColumnSource.java
```diff
@@ -0,0 +1,133 @@
...
+    @Override
+    public void addChunk(final Chunk<? extends Attributes.Values> chunk) {
+        addChunk(chunk.asCharChunk());
+    }
+
+    @Override
+    public void clear() {
+        totalSize = 0;
+        data.clear();
+        offsets.clear();
```
rcaudy, line 154: You want to use `offsets.reset()`, I think. `clear()` throws away the backing array.

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/CharChunkColumnSource.java
```diff
@@ -0,0 +1,133 @@
...
+import io.deephaven.db.v2.utils.OrderedKeys;
+import io.deephaven.util.QueryConstants;
+import org.apache.commons.lang3.mutable.MutableInt;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.ArrayList;
+
+public class CharChunkColumnSource extends AbstractColumnSource<Character> implements ImmutableColumnSourceGetDefaults.ForChar, ChunkColumnSource<Character> {
+    private final ArrayList<CharChunk<? extends Attributes.Values>> data = new ArrayList<>();
+    private final TLongArrayList offsets = new TLongArrayList();
```
rcaudy, line 24: We should be clearer about what we're storing. `startOffsets`? `firstOffsets`?

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/CharChunkColumnSource.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private int getChunkIndex(final long start) {
+        return getChunkIndex(start, 0);
+    }
+
+    private int getChunkIndex(final long start, final int startIndex) {
+        int index = offsets.binarySearch(start, startIndex, offsets.size());
+        if (index < 0) {
+            index = -index - 2;
```
rcaudy, line 134: This may be mostly about aesthetics, and the more I think about it the more I think you should read it and then do exactly what you want if you don't find it persuasive.

If you store last offset instead of first, you can make this a little prettier:
1. You can use `~1` for the negative index conversion.
2. You can throw an exception if the corrected result is `>= offsets.size()`.

Of course, the trade-off is you'd need to change your calculation around L82:
```
 final int offsetWithinChunk = (int)(s - (chunkIndex == 0 ? 0 :  offsets.get(chunkIndex - 1)));
```
I'm pretty sure this would be a better pattern than relying on the check at L89 to terminate a request for more keys than present, or an AIOOBE from `binarySearch`.

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/CharChunkColumnSource.java
```diff
@@ -0,0 +1,133 @@
...
+                    return charChunk;
+                }
+                return ((ChunkGetContext)context).resettableCharChunk.resetFromChunk(charChunk, offset, length);
+            }
+        }
+        return getChunkByFilling(context, orderedKeys);
+    }
+
+    @Override
+    public void fillChunk(@NotNull FillContext context, @NotNull WritableChunk<? super Attributes.Values> destination, @NotNull OrderedKeys orderedKeys) {
```
rcaudy, line 83: I'm kind of suspicious that you should just implement `com.illumon.iris.db.v2.sources.chunk.page.PageStore` and introduce a new kind of `ChunkPage` that has-a chunk instead of the current impl that is-a chunk.

That said, I think this code is perfectly correct barring other comments.

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/CharChunkColumnSource.java
```diff
@@ -0,0 +1,133 @@
...
+        final MutableInt startChunkIndex = new MutableInt(0);
```
rcaudy, line 84: `searchStartChunkIndex`?

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/CharChunkColumnSource.java
```diff
@@ -0,0 +1,133 @@
...
+                final long rangeLength = e - s + 1;
+                final int chunkRemaining = chunkSize - offsetWithinChunk;
+                final int length = rangeLength > chunkRemaining ? chunkRemaining : (int) rangeLength;
+                Assert.gtZero(length, "length");
+                destination.copyFromChunk(charChunk, offsetWithinChunk, destinationOffset.intValue(), length);
+                destinationOffset.add(length);
+                startChunkIndex.setValue(chunkIndex + 1);
+                s += length;
+            }
+        });
```
rcaudy, line 102: I think you are missing a `destination.setSize(orderedKeys.intSize)`.

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/CharChunkColumnSource.java
```diff
@@ -12,6 +12,13 @@
 
 import java.util.ArrayList;
 
+/**
+ *  A column source backed by CharChunks.
```
rcaudy, line 16: `CharChunks` -> `{@link CharChunk char chunks}`
`getChunk` -> `{@link #getChunk}`
"chunk, or a" -? "chunk or a"

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/CharChunkColumnSource.java
```diff
@@ -101,19 +108,35 @@ public void fillPrevChunk(@NotNull FillContext context, @NotNull WritableChunk<?
...
     private int getChunkIndex(final long start) {
         return getChunkIndex(start, 0);
     }
 
-    private int getChunkIndex(final long start, final int startIndex) {
-        int index = offsets.binarySearch(start, startIndex, offsets.size());
+    /**
+     * Given an index within this column's address space; return the chunk that contains the index.
+     *
+     * @param start the data index to find the corresponding chunk for
```
rcaudy, line 125: Can you run the auto-formatter on this?

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/CharChunkColumnSource.java
```diff
@@ -101,19 +108,35 @@ public void fillPrevChunk(@NotNull FillContext context, @NotNull WritableChunk<?
...
+    private int getChunkIndex(final long start, final int startChunk) {
+        int index = offsets.binarySearch(start, startChunk, offsets.size());
         if (index < 0) {
             index = -index - 2;
         }
         return index;
     }
 
-    public void addChunk(final CharChunk<? extends Attributes.Values> chunk) {
+    private void addChunk(final CharChunk<? extends Attributes.Values> chunk) {
```
rcaudy, line 139: Why make this private? Wouldn't we rather code have the option to use the typed version?

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/ChunkColumnSource.java
```diff
@@ -29,6 +45,18 @@
...
+     * Append a chunk of data to this column source.
+     *
+     * @param chunk the chunk of data to add
+     */
     void addChunk(Chunk<? extends Attributes.Values> chunk);
+
+    /**
+     *  Reset the column source to be ready for reuse.
+     *
+     *  Cear will discard the currently held chunks.  This should not be  called if a table will continue to reference
```
rcaudy, line 58: extra space in "be  called"

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/ChunkColumnSource.java
```diff
@@ -29,6 +45,18 @@
...
+     *  the column source; as it violates the immutability contract.
+     */
     void clear();
```
rcaudy, line 61: The clear implementations should also return chunks to the pool. See my previous claims about `PoolableChunk`; what should we do about it?

rcaudy:
Link to #293

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/CharChunkColumnSource.java
cpwright: It wasn't used outside, so by making it private I didn't need to duplicate the Javadoc.  I could consider making it public again if it would make sense from the producer.

rcaudy: We can punt until we have a use-case.

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/ChunkColumnSource.java
cpwright: WritableChunk is not pretty, but the path of least resistance.

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/CharChunkColumnSource.java
cpwright: I kind of hate the zero check.

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/ByteChunkColumnSource.java
```diff
@@ -34,62 +35,63 @@ protected ByteChunkColumnSource() {
...
         }
         return getChunkByFilling(context, orderedKeys);
     }
 
     @Override
-    public void fillChunk(@NotNull FillContext context, @NotNull WritableChunk<? super Attributes.Values> destination, @NotNull OrderedKeys orderedKeys) {
-        final MutableInt startChunkIndex = new MutableInt(0);
+    public void fillChunk(@NotNull final FillContext context, @NotNull final WritableChunk<? super Attributes.Values> destination, @NotNull final OrderedKeys orderedKeys) {
+        final MutableInt searchStartChunkIndex = new MutableInt(0);
         final MutableInt destinationOffset = new MutableInt(0);
```
rcaudy, line 90: You know, we could just use `destination.size()` for this...

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/ChunkColumnSource.java
```diff
@@ -44,6 +45,37 @@
...
 
+    /**
+     * Create a new ChunkColumnSource for the given chunk type and data type.
+     *
+     * @param sharedOffsetForData an array list representing the shared offsets for data across several ChunkColumnSources
+     * @param chunkType           the type of chunk
+     * @param dataType            the datatype for the newly created column source
+     * @return an empty ChunkColumnSource
+     */
+    static ChunkColumnSource<?> make(TLongArrayList sharedOffsetForData, ChunkType chunkType, Class<?> dataType) {
```
rcaudy, line 56: Maybe put the optional list at the end?

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/ChunkColumnSource.java
```diff
@@ -44,6 +45,37 @@
...
+        switch (chunkType) {
+            case Char:
+                return new CharChunkColumnSource(sharedOffsetForData);
+            case Byte:
```
rcaudy, line 60: Do we need a "BooleanAsByte" source?

DB/src/main/java/io/deephaven/db/v2/sources/ByteAsBooleanColumnSource.java
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
+ */
+
+package io.deephaven.db.v2.sources;
+
+import io.deephaven.db.util.BooleanUtils;
```
rcaudy, line 7: Prefer `io.deephaven.util.BooleanUtils`.

DB/src/main/java/io/deephaven/db/v2/sources/ByteAsBooleanColumnSource.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public ByteAsBooleanColumnSource(ColumnSource<Byte> alternateColumnSource) {
+        super(Boolean.class);
+        this.alternateColumnSource = alternateColumnSource;
+    }
+
+    @Override
+    public Boolean get(long index) {
+        final byte byteValue = alternateColumnSource.getByte(index);
+        if (byteValue == QueryConstants.NULL_BYTE) {
```
rcaudy, line 30: This might merit a comment. Took me a minute to think through the possibilities, but I agree that this seems necessary given that `ReadOnlyIndex.NULL_KEY` should map to `NULL_BYTE` .

DB/src/main/java/io/deephaven/db/v2/sources/ByteAsBooleanColumnSource.java
```diff
@@ -0,0 +1,110 @@
...
+
+    private final ColumnSource<Byte> alternateColumnSource;
+
+    public ByteAsBooleanColumnSource(ColumnSource<Byte> alternateColumnSource) {
+        super(Boolean.class);
+        this.alternateColumnSource = alternateColumnSource;
+    }
+
+    @Override
+    public Boolean get(long index) {
```
rcaudy, line 28: `final`?

DB/src/main/java/io/deephaven/db/v2/sources/ByteAsBooleanColumnSource.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    public <ALTERNATE_DATA_TYPE> ColumnSource<ALTERNATE_DATA_TYPE> doReinterpret(@NotNull final Class<ALTERNATE_DATA_TYPE> alternateDataType) throws IllegalArgumentException {
+        //noinspection unchecked
+        return (ColumnSource<ALTERNATE_DATA_TYPE>) alternateColumnSource;
+    }
+
+    private class ToBooleanFillContext implements FillContext {
+        final FillContext alternateFillContext;
```
rcaudy, line 62: I think it makes more sense to wrap a `GetContext` than a `FillContext`+`WritableByteChunk` pair, and then use `getChunk` rather than `fillChunk`. That way you exploit the underlying source's possibly high-quality `getChunk` impl.

Maybe I can persuade you to fix `io.deephaven.db.v2.sources.BooleanAsByteColumnSource` at the same time, in the same way.

DB/src/main/java/io/deephaven/db/v2/sources/LongAsDateTimeColumnSource.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public LongAsDateTimeColumnSource(ColumnSource<Long> alternateColumnSource) {
+        super(DBDateTime.class);
+        this.alternateColumnSource = alternateColumnSource;
+    }
+
+    @Override
+    public DBDateTime get(long index) {
+        final long longValue = alternateColumnSource.getLong(index);
+        if (longValue == QueryConstants.NULL_LONG) {
```
rcaudy, line 30: You can just use:
`io.deephaven.db.tables.utils.DBTimeUtils#nanosToTime`

Same applies to the `convertToDBDateTime` implementation.

DB/src/main/java/io/deephaven/db/v2/sources/LongAsDateTimeColumnSource.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    public <ALTERNATE_DATA_TYPE> ColumnSource<ALTERNATE_DATA_TYPE> doReinterpret(@NotNull final Class<ALTERNATE_DATA_TYPE> alternateDataType) throws IllegalArgumentException {
+        //noinspection unchecked
+        return (ColumnSource<ALTERNATE_DATA_TYPE>) alternateColumnSource;
+    }
+
+    private class ToDateTimeFillContext implements FillContext {
+        final FillContext alternateFillContext;
```
rcaudy, line 62: Same commentary as on the other class; better to fill from get.

rcaudy: And same as on the other class, we should fix `io.deephaven.db.v2.sources.DatetimeAsLongColumnSource` too.

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/CharChunkColumnSource.java
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * Given an index within this column's address space; return the chunk that contains the index.
+     *
+     * @param start      the data index to find the corresponding chunk for
+     * @param startChunk the first chunk that may possibly contain start
+     * @return the chunk index within data and offsets
+     */
+
+    private int getChunkIndex(final long start, final int startChunk) {
+        int index = firstOffsetForData.binarySearch(start, startChunk, firstOffsetForData.size());
```
rcaudy, line 138: Should we insist that all chunks except the last one have the same size, so we don't have to do a binary search? I guess it would make this tool less general.

DB/src/main/java/io/deephaven/db/v2/sources/ByteAsBooleanColumnSource.java
cpwright: We could also take the opportunity to fix the definition of NULL_BOOLEAN_AS_BYTE to be identical to NULL_BYTE since we do not have any legacy to care about.

DB/src/main/java/io/deephaven/db/v2/sources/chunkcolumnsource/CharChunkColumnSource.java
cpwright: No, but I added an escape for when you march through in order.

DB/src/main/java/io/deephaven/db/v2/sources/ByteAsBooleanColumnSource.java
rcaudy: I think that can be done. I need to think a little bit about it on the Parquet side. This merits a ticket of its own.

rcaudy: https://github.com/deephaven/deephaven-core/issues/927

rcaudy (approved):

PR #922, opened 2021-07-27
Add dockerized redpanda; do an initial very simple spike end-to-end

jcferretti:
Refactored docker-compose.yml common parts to docker-compose-common.yml; created two new docker-compose files in their
own subdirectories (docker-compose likes project definitions matching directory names): redpanda, redpanda-standalone.

redpanda-standalone allows to run redpanda on its own.
redpanda runs redpanda together with deephaven core.

The diff also includes all the basics for a simple end-to-end spike where a message published in redpanda shows up
in a DH table.


rcaudy:
Link to #293

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
```diff
@@ -0,0 +1,309 @@
...
+     */
+    private boolean pollOnce(final Duration timeout) {
+        final ConsumerRecords<?, ?> records;
+        try {
+            records = consumer.poll(timeout);
+        } catch (WakeupException we) {
+            return false;
+        }
+        for (final ConsumerRecord<?, ?> record : records) {
+            final int partition = record.partition();
```
cpwright, line 293: We need to multiplex to the right partition destination.

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/GenericRecordConsumerRecordToTableWriterAdapter.java
```diff
@@ -0,0 +1,406 @@
...
+    private final Map<String, Function<GenericRecord, ?>> columnToValueFunction;
+    private final boolean allowNullKeys;
+    private final boolean allowNullValues;
+
+    private GenericRecordConsumerRecordToTableWriterAdapter(
+            final TableWriter<?> writer,
+            final String kafkaPartitionColumn,
+            final String offsetColumnName,
+            final String timestampColumnName,
+            String rawKeyColumnName,
```
cpwright, line 45: This is effectively subsuming simple for the key column; but doing generic records for the value column.

redpanda-standalone/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+version: '3.7'
```
niloc132, line 1: other files are 3.8, can we make everything use the same version?

niloc132 (approved):

redpanda-standalone/docker-compose.yml
jcferretti: We discussed, and checked and moved everything to 3.4

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
```diff
@@ -0,0 +1,298 @@
...
+            final Consumer<ConsumerRecord<?, ?>> consumer = partitionToConsumer.apply(partition);
+            try {
+                consumer.accept(record);
+            } catch (Exception ex) {
+                ++messagesWithErr;
+                log.error().append(logPrefix).append("Exception while processing Kafka message:").append(ex);
```
rcaudy, line 287: If we hit any errors, maybe we should deliver a failure to the table downstream. Not necessarily worth plumbing at this stage.

rcaudy (commented):
This is a very partial review, just wanted to comment on a thing I noticed.

Generators/src/main/java/io/deephaven/python/PythonStaticGenerator.java
```diff
@@ -19,7 +19,7 @@
     private static final List<String> customTableTools = Arrays.asList("col", "byteCol", "shortCol", "intCol", "longCol",
             "floatCol", "doubleCol", "charCol", "newTable", "colSource", "objColSource");
     private static final List<String> customParquetTools = Arrays.asList("deleteTable", "readTable", "writeParquetTables", "writeTable", "writeTables");
-    private static final List<String> customTableInputHandler = Arrays.asList("c", "cKey", "getRestrictedColumnNames");
```
cpwright, line 22: meant to be removed?

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
```diff
@@ -0,0 +1,298 @@
...
+        final Set<TopicPartition> assignments = consumer.assignment();
+        log.info().append(logPrefix).append("Partition Assignments: ").append(assignments.toString()).endl();
+
+        if (assignments.size() != openPartitions.size()) {
+            throw new UncheckedDeephavenException(logPrefix +  "Partition assignments do not match request: assignments=" + assignments + ", request=" + openPartitions);
+        }
+
+        for (final TopicPartition topicPartition : assignments) {
+            final long seekOffset = partitionToInitialSeekOffset.applyAsLong(topicPartition.partition());
+            if (seekOffset == -1) {
```
cpwright, line 207: For Community, a very normal pattern might be not to seek at all; and let offsets be managed by the broker/zookeeper.

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
```diff
@@ -0,0 +1,298 @@
...
+        final DecimalFormat rateFormat = new DecimalFormat("#.0000");
+        while (openPartitions.size() == expectedPartitionCount) {
+            final long beforePoll = System.nanoTime();
+            final long nextReport = lastReportNanos + reportIntervalNanos;
+            final long remainingNanos = beforePoll > nextReport ? 0 : (nextReport - beforePoll);
+            boolean noMore = pollOnce(Duration.ofNanos(remainingNanos));
+            if (noMore) {
+                break;
+            }
+            final long afterPoll = System.nanoTime();
```
cpwright, line 250: possibly have the printout on the final noMore result?

grpc-api/grpc-api.gradle
```diff
@@ -51,7 +51,7 @@ dependencies {
     testCompile TestTools.projectDependency(project, 'DB'),
                 TestTools.projectDependency(project, 'Util')
 
-    runtimeOnly project(':Numerics'), project(':Integrations')
+    runtimeOnly project(':Numerics'), project(':Integrations'), project(':KafkaIngester')
```
cpwright, line 54: Why does our API depend on Kafka?

jcferretti: grpc-api is mis-named.  This is essentially the worker code; Numerics and Integrations and Plot are here so that the worker can implement them.

KafkaIngester/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(KafkaTools .class);
+
+    public static Table simpleConsumeToTable(
+            @NotNull final Properties kafkaConsumerProperties,
+            @NotNull final String topic) {
+        final DynamicTableWriter tableWriter = new DynamicTableWriter(SIMPLE_TABLE_DEFINITION);
+        final ConsumerRecordToTableWriterAdapter adapter = SimpleConsumerRecordToTableWriterAdapter.make(
+                tableWriter, KAFKA_PARTITION_COLUMN_NAME, OFFSET_COLUMN_NAME, TIMESTAMP_COLUMN_NAME, KEY_COLUMN_NAME, VALUE_COLUMN_NAME);
+        if (!kafkaConsumerProperties.containsKey(KEY_DESERIALIZER_PROPERTY)) {
```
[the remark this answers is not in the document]
jcferretti: This is intended as a first step that needs to be iterated to our actual API.  In my imagination, KafkaTools remains, this method changes, likely many more methods are added.  While we do that, we have something that keeps working and if it breaks we know.

Or rather, we would have that if I had added any tests!  But in making it easy for us to actually spin up redpanda and send data, I hope to make it common for us while we change this stuff to test with the web IDE etc.

KafkaIngester/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -0,0 +1,77 @@
...
+    public static final String TIMESTAMP_COLUMN_NAME = "timestamp";
+    public static final String KEY_COLUMN_NAME = "key";
+    public static final String VALUE_COLUMN_NAME = "value";
+    public static final String KEY_DESERIALIZER_PROPERTY = "key.deserializer";
+    public static final String VALUE_DESERIALIZER_PROPERTY = "value.deserializer";
+    public static final String DEFAULT_DESERIALIZER = "org.apache.kafka.common.serialization.StringDeserializer";
+
+    private static final Logger log = LoggerFactory.getLogger(KafkaTools .class);
+
+    public static Table simpleConsumeToTable(
```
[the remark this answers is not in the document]
jcferretti: This is intended for our python/groovy integration.  I called this KafkaTools to follow the pattern of TableTools and ParquetTools.  This file is the basis for all the other changes in this PR that wire things up to be callable from python.

The implementation happens to be here right now because it needed to be somewhere, but likely as this stuff grows it should move somewhere (eg, ParqutTools calls to pretty big files for ParquetTableWriter and such in different, parquet specific modules.  Since we want this stuff to be based on other stuff and not TableWriter, this implementation is scaffolding for integration right now, it is fully expected to change, etc.

Generators/src/main/java/io/deephaven/python/PythonStaticGenerator.java
jcferretti: Yes, unused.

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
jcferretti: That is a good point, and from an API perspective it kinda kills the int function... unless we want to mess with -2 for... and -3 for...

cpwright: We could also not be jerks and use constants with -2 and -3. :)

KafkaIngester/src/main/java/io/deephaven/kafka/KafkaTools.java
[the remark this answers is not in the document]
cpwright: I meant "Consume -> Consumer"; but it doesn't take a consumer.  It just reads a bit off.  KafkaTools is fine.

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/GenericRecordConsumerRecordToTableWriterAdapter.java
jcferretti: Lost, I don't understand your comment.  Feel free to steer me in the direction of the actual change/fix here.
Keep in mind tho that these classes will go away, we won't do TableWriter based, correct?

cpwright: It's more of a generic comment that we want to split the ConsumerRecordToStreamPublisher adapters by "generic stuff", "key stuff" and "value stuff".  Pointing out that primitives keys are important enough to have them built into the generic record adapter.

cpwright (approved):

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
jcferretti: https://github.com/deephaven/deephaven-core/issues/934

PR #926, opened 2021-07-28
Standardize on Arrow's Flight Schema representation

nbauernfeind:


grpc-api/src/main/java/io/deephaven/grpc_api/session/TicketRouter.java
```diff
@@ -190,12 +192,17 @@ public void visitFlightInfo(final @Nullable SessionState session, final Consumer
                 .build();
     }
 
-    private static ByteString schemaBytesFromTable(final Table table) {
+    public static ByteString schemaBytesFromTable(final Table table) {
+        return schemaBytesFromTable(table.getDefinition(), table.getAttributes());
+    }
+
+    public static ByteString schemaBytesFromTable(final TableDefinition table,
```
niloc132, line 199: it seems weird for this to be in _this_ class, since it basically leans on makeSchemaPayload in BarrageSchemaUtil and adds a bit of padding...? maybe BarrageSchemaUtil for this and its new overload?

niloc132 (approved):

PR #930, opened 2021-07-29
StreamToTableAdapter implementation and simple unit test.

cpwright:


DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -29,11 +54,83 @@ public StreamToTableAdapter(@NotNull final TableDefinition tableDefinition,
...
+        table.setRefreshing(true);
+        table.setAttribute(Table.STREAM_TABLE_ATTRIBUTE, Boolean.TRUE);
+    }
+
+    @NotNull
+    private ChunkColumnSource<?> [] makeChunkSources(Class<?>[] columnTypesArray, String[] columnNamesArray) {
+        final ChunkColumnSource<?> [] sources = new ChunkColumnSource[columnNamesArray.length];
+        final TLongArrayList offsets = new TLongArrayList();
+        for (int ii = 0; ii < columnTypesArray.length; ++ii) {
+            sources[ii]  = ChunkColumnSource.make(chunkTypes[ii], columnTypesArray[ii], offsets);
```
rcaudy, line 82: This highlights that we've forgotten to plumb component type for these sources. You should take a definition instead of parallel arrays and do that. Also, you have no reason to take the names array. Lastly, could be a static method.

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -29,11 +54,83 @@ public StreamToTableAdapter(@NotNull final TableDefinition tableDefinition,
...
+        return sources;
     }
 
-    public Table table() {
+    @NotNull
+    private SwitchColumnSource [] makeSwitchSources(ChunkColumnSource<?> [] wrapped, Map<String, ColumnSource<?>> visibleSourcesMap) {
+        final SwitchColumnSource [] switchSources = new SwitchColumnSource[wrapped.length];
+        final Map<String, SwitchColumnSource<?>> result = new LinkedHashMap<>();
+        for (int ii = 0; ii < wrapped.length; ++ii) {
+            final SwitchColumnSource<?> switchSource = new SwitchColumnSource<>(wrapped[ii]);
```
rcaudy, line 102: Change the constructor for `SwitchColumnSource` to use the super constructor that takes component type, please. This method can be static, like the others.

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -29,11 +54,83 @@ public StreamToTableAdapter(@NotNull final TableDefinition tableDefinition,
...
+        final ChunkColumnSource<?> [] sources = new ChunkColumnSource[columnNamesArray.length];
+        final TLongArrayList offsets = new TLongArrayList();
+        for (int ii = 0; ii < columnTypesArray.length; ++ii) {
+            sources[ii]  = ChunkColumnSource.make(chunkTypes[ii], columnTypesArray[ii], offsets);
+        }
+        return sources;
+    }
+
+    @NotNull
+    private NullValueColumnSource<?> [] makeNullColumnSources(ChunkColumnSource<?> [] chunkSources) {
```
rcaudy, line 88: Could be static and take the definition. Maybe better?

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -29,11 +54,83 @@ public StreamToTableAdapter(@NotNull final TableDefinition tableDefinition,
...
+        columnNamesArray = tableDefinition.getColumnNamesArray();
+
+        chunkTypes = tableDefinitionToChunkTypes(columnTypesArray);
+        currentChunkSources = makeChunkSources(columnTypesArray, columnNamesArray);
+        nullColumnSources = makeNullColumnSources(currentChunkSources);
+
+        final LinkedHashMap<String, ColumnSource<?>> visibleSources = new LinkedHashMap<>();
+        switchSources = makeSwitchSources(currentChunkSources, visibleSources);
+
+        bufferChunkSources = makeChunkSources(columnTypesArray, columnNamesArray);
```
rcaudy, line 68: Why do we need this in the constructor? Seems inconsistent with the rest of the code.

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -21,6 +36,16 @@
 
     private final QueryTable table;
 
+    private final Index index;
```
rcaudy, line 39: Can you humor me and rearrange these so that the final ones are before and separate from the mutable ones, and in some logical order?

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -21,6 +36,16 @@
...
+    private ChunkColumnSource<?> [] bufferChunkSources;
+    private ChunkColumnSource<?> [] currentChunkSources;
+    private ChunkColumnSource<?> [] prevChunkSources;
+    private NullValueColumnSource<?> [] nullColumnSources;
+    private final SwitchColumnSource<?> [] switchSources;
+    private final ChunkType [] chunkTypes;
+    private final Class<?>[] columnTypesArray;
```
jcferretti, line 46: Space or no space, pick one? :-)

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -46,11 +143,65 @@ public void close() {
...
+        final long newSize;
+
+        synchronized (this) {
+            newSize = bufferChunkSources == null ? 0 : bufferChunkSources[0].getSize();
+
+            if (oldSize == 0 && newSize == 0) {
+                return;
+            }
+
+            ChunkColumnSource<?> [] oldPrev = prevChunkSources;
```
jcferretti, line 157: maybe this can be final?

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -46,11 +143,65 @@ public void close() {
...
+                for (int ii = 0; ii < switchSources.length; ++ii) {
+                    //noinspection unchecked
+                    switchSources[ii].setNewCurrent((ColumnSource) nullColumnSources[ii]);
+                }
+            } else {
+                for (int ii = 0; ii < switchSources.length; ++ii) {
+                    switchSources[ii].setNewCurrent((ColumnSource) bufferChunkSources[ii]);
+                }
+            }
+            // current becomes prev, buffer becomes current, buffer is set to cleared prev for the next pass
```
jcferretti, line 176: I find these almost exact textual descriptions of what the code right after does (which already use meaningful variable names to the point of literally sounding like reading the code) risky from a "code evolution" perspective while adding little value.  As in, you change the code below later, you are also forced to rewrite the comment.

My vote would be for a much more general "set the stage for next pass" comment (which is totally useful to separate the sections of what the code does, so I recognize the value of that in your comment).  But this is an opinion for your consideration, not a request for a change, I don't mind if you leave it.

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -29,11 +54,83 @@ public StreamToTableAdapter(@NotNull final TableDefinition tableDefinition,
...
+
+        columnTypesArray = tableDefinition.getColumnTypesArray();
+        columnNamesArray = tableDefinition.getColumnNamesArray();
+
+        chunkTypes = tableDefinitionToChunkTypes(columnTypesArray);
+        currentChunkSources = makeChunkSources(columnTypesArray, columnNamesArray);
+        nullColumnSources = makeNullColumnSources(currentChunkSources);
+
+        final LinkedHashMap<String, ColumnSource<?>> visibleSources = new LinkedHashMap<>();
+        switchSources = makeSwitchSources(currentChunkSources, visibleSources);
```
rcaudy, line 66: Doesn't it make more sense to initialize the switch sources from the null sources? We should then skip allocating the current sources, like the buffer ones.

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -46,11 +143,65 @@ public void close() {
...
     @SafeVarargs
     @Override
     public final void accept(@NotNull final WritableChunk<Attributes.Values>... data) {
         // Accumulate data into buffered column sources
+        synchronized (this) {
+            if (bufferChunkSources == null) {
+                bufferChunkSources = makeChunkSources(columnTypesArray, columnNamesArray);
+            }
+            if (data.length != bufferChunkSources.length) {
+                throw new IllegalStateException();
```
jcferretti, line 200: What state is illegal? (exception text)

Also consider that IllegalState percolating up makes the worker bail/exit; UncheckedDeephavenException is what you can use when you want the exception to get to the toplevel, notify the user, abort whatever the current thing is, but not kill the worker etc.

Maybe this code is running in the KafkaIngester thread tho?   In that case maybe a TODO with the new issue linked for the error handling stuff we discussed.

In case is useful, for your convenience, that issue's link: https://github.com/deephaven/deephaven-core/issues/934

DB/src/test/java/io/deephaven/stream/TestStreamToTableAdapter.java
```diff
@@ -0,0 +1,192 @@
...
+        LiveTableMonitor.DEFAULT.runWithinUnitTestCycle(adapter::refresh);
+        TstUtils.assertTableEquals(empty, result);
+        TestCase.assertEquals(0, listener.getCount());
+
+        final WritableChunk<Attributes.Values> [] chunks = new WritableChunk[3];
+        final WritableObjectChunk<Object, Attributes.Values> woc = WritableObjectChunk.makeWritableChunk(3);
+        chunks[0] = woc;
+        woc.set(0, "Collins");
+        woc.set(1, "Armstrong");
+        woc.set(2, "Aldrin");
```
jcferretti, line 138: No Gagarin? ;-)

jcferretti (approved):

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -46,11 +143,65 @@ public void close() {
     public void refresh() {
         streamPublisher.flush();
         // Switch columns, update index, deliver notification
+
+        final long oldSize = index.size();
+        final long newSize;
+
+        synchronized (this) {
```
rcaudy, line 150: We only need to hold the lock while swapping the buffered sources. Meaning we should clear out the prevColumnSources `(if oldSize != 0)`, then grab our lock, then check buffered size, then do the swap under the lock, then release the lock and do the rest of our maintenance.

rcaudy (changes requested):

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -75,31 +68,43 @@ public StreamToTableAdapter(@NotNull final TableDefinition tableDefinition,
...
-        return sources;
     }
 
     @NotNull
-    private SwitchColumnSource [] makeSwitchSources(ChunkColumnSource<?> [] wrapped, Map<String, ColumnSource<?>> visibleSourcesMap) {
-        final SwitchColumnSource [] switchSources = new SwitchColumnSource[wrapped.length];
-        final Map<String, SwitchColumnSource<?>> result = new LinkedHashMap<>();
+    private static SwitchColumnSource<?> [] makeSwitchSources(TableDefinition definition, NullValueColumnSource<?> [] wrapped, Map<String, ColumnSource<?>> visibleSourcesMap) {
+        final SwitchColumnSource<?> [] switchSources = new SwitchColumnSource[wrapped.length];
+        final Class<?> [] columnTypesArray = definition.getColumnTypesArray();
```
rcaudy, line 103: Do we get anything for making these arrays?

DB/src/main/java/io/deephaven/db/v2/sources/SwitchColumnSource.java
```diff
@@ -6,22 +6,33 @@
...
         this.currentSource = currentSource;
         this.updateCommitter = new UpdateCommitter<>(this, SwitchColumnSource::clearPrevious);
+        this.onPreviousCommitted = onPreviousCommitted;
     }
 
     private void clearPrevious() {
+        final ColumnSource<T> captured = prevSource;
         prevCycle = -1;
         prevSource = null;
+        onPreviousCommitted.accept(captured);
```
rcaudy, line 35: This will NPE if someone give you a null `onPreviousCommitted`.

rcaudy (approved):

Issue #697, opened 2021-06-02
Column statistics


pete-petey:
@niloc132 Do note: Robb made these faster on Enterprise side.  I assume that is irrelevant... just noting it though.  (Perhaps worthy of a chat with him.)

mattrunyon:
To make searching this ticket easier, here is the web-UI error hovering on a table column name.
Error: java.lang.UnsupportedOperationException: getColumnStatistics

mofojed:
@niloc132 is Column Statistic support possible now/is it blocked on anything else?

Issue #37, opened 2021-01-29
Implement PartitionedTable, delete TableMap


niloc132:
Applies to aggregation, treetables, rollups.

mofojed:
@niloc132 Is this expected for August? Wondering if UI should disable some of the aggregation features that don't work.

niloc132:
No, it is not expected to work by then.

rcaudy:
I'm going to hijack this issue under a new name, since it documents the design that was pursued. We will not actually need a gRPC API for SmartKeys.
Part of #1762

PR #931, opened 2021-07-29
Remove @JsMethod from JsTable and JsTreeTable features that do not currently work.

mofojed:
- Related to #37 and #697, there is some table functionality that does not work yet.
- Comment out the methods that do not currently work (shouldn't have an API that doesn't work).
- UI checks for the presence of the method for whether to show the option or not. When they are added back, they should appear in the UI again. See https://github.com/deephaven/web-client-ui/issues/136

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
@@ -559,7 +559,8 @@ public class JsTable extends HasEventHandling implements HasTableBinding, HasLif
         }
         return Promise.resolve(new JsTable(this));
     }
-
+    // TODO: #37: Need SmartKey support for this functionality
+    // @JsMethod
     public Promise<JsTotalsTable> getTotalsTable(Object config) {
         // fetch the handle and wrap it in a new jstable. listen for changes
         // on the parent table, and re-fetch each time.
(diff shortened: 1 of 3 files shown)


web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
```diff
@@ -572,15 +559,29 @@ public void internalSubscribe(JsArray<Column> columns, TableSubscription sub) {
         }
         return Promise.resolve(new JsTable(this));
     }
-
-    @JsMethod
-    public Promise<JsTotalsTable> getTotalsTable(@JsOptional Object config) {
+    // TODO: #37: Need SmartKey support for this functionality
+    // @JsMethod
+    public Promise<JsTotalsTable> getTotalsTable(Object config) {
```
niloc132, line 564: that config is still optional, right?

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -865,8 +866,9 @@ public boolean isIncludeConstituents() {
         return sourceTable.then(t -> Promise.resolve(t.getTotalsTableConfig()));
     }
 
-    @JsMethod
-    public Promise<JsTotalsTable> getTotalsTable(@JsOptional Object config) {
+    // TODO: #37: Need SmartKey support for this functionality
+    // @JsMethod
+    public Promise<JsTotalsTable> getTotalsTable(Object config) {
```
niloc132, line 871: still optional here, below

niloc132 (commented):
Arguably you should also just kill the getTreeTable call, anything else that can return a TreeTable from the idesession?

niloc132 (approved):

Issue #868, opened 2021-07-19
Parquet: Support identifying hierarchical partitioning and schema from Spark/Dask metadata files


rcaudy:
We can and should be able to filter partitions according to the values they contain when thrift metadata files are present.
Part of #294, maybe.

rcaudy:
I'm moving the requirements for this issue a bit. We've got good support for partitioning columns driven by by metadata files, but we're lacking support for predicate pushdown to the row group level or below. See https://github.com/deephaven/deephaven-core/issues/968 for future work.

PR #954, opened 2021-08-02
Support multiple row groups in Parquet, clean up regioned column sources, and fix dictionary writing/symbol tables

### LINKED ISSUE #857: We only read row group zero in a parquet file.  If a file has more rowgroups we will be missing rows.
https://github.com/deephaven/deephaven-core/blob/65bfdbfe7a199bc85c2123ddfe93d22e8979345b/DB/src/main/java/io/deephaven/db/v2/locations/local/ReadOnlyParquetTableLocation.java#L65

### LINKED ISSUE #866: Clean up regioned column sources
Part of #294 
We have a fairly brittle top layer that doesn't allow for "differently shaped" data at the lower level very well. Contexts attempt to be good enough for all region types, etc.
We can probably do a very clean refactoring here with relatively minimal effort.

rcaudy:
Closes #857 
Closes #866


DB/src/main/java/io/deephaven/db/v2/sources/regioned/RegionedColumnSourceManager.java
```diff
@@ -241,69 +248,97 @@ public ImmutableTableLocationKey getKey(@NotNull final EmptyTableLocationEntry e
...
-                        + ", larger than maximum supported location size " + RegionedColumnSource.REGION_CAPACITY_IN_ELEMENTS);
-            }
+            final ReadOnlyIndex updateIndex = location.getIndex();
+            try {
+                if (updateIndex == null) {
+                    // This should be impossible - the subscription buffer transforms a transition to null into a pending exception
+                    throw new TableDataException("Location " + location + " is no longer available, data has been removed");
+                }
+                if (!indexAtLastUpdate.subsetOf(updateIndex)) { // Bad change
+                    throw new IllegalStateException("Index keys removed at location " + location + ": " + indexAtLastUpdate.minus(updateIndex));
```
jcferretti, line 306: Is an IllegalStateException better than an Assert here?
The behavior on the grpc_api main for IllegalStateExceptions, if I remember correctly, is not great.
Also, maybe we need tickets to ensure that, on asserts failed, we get a heap dump, and on IllegalStateException,
we don't leave the worker in a foobar state.

DB/src/main/java/io/deephaven/db/v2/sources/regioned/RegionedColumnSourceWithDictionary.java
```diff
@@ -0,0 +1,299 @@
...
+
+        @Override
+        public ColumnRegionLong<DictionaryKeys> getRegion(final int regionIndex) {
+            final ColumnRegionObject<DATA_TYPE, Values> sourceRegion = RegionedColumnSourceWithDictionary.this.getRegion(regionIndex);
+            if (sourceRegion instanceof ColumnRegion.Null) {
+                return nullRegion;
+            }
+            ColumnRegionLong<DictionaryKeys>[] localWrappers;
+            ColumnRegionLong<DictionaryKeys> wrapper;
+            if ((localWrappers = wrapperRegions).length > regionIndex && (wrapper = localWrappers[regionIndex]) != null) {
```
jcferretti, line 108: Why do the assignments outside of the synchronized block?

classpaths.gradle
```diff
@@ -109,17 +109,18 @@ dependencies {
...
-           'com.jcraft:jsch:0.1.51',
-           "org.eclipse.jgit:org.eclipse.jgit:$Classpaths.JGIT_VERSION",
-           "org.eclipse.jgit:org.eclipse.jgit.ssh.jsch:$Classpaths.JGIT_VERSION",
-           'org.scala-lang:scala-library-all:2.12.2',
-           'org.scala-lang:scala-compiler:2.12.2',
-           'org.yaml:snakeyaml:1.22',
-           'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.9.8',
-           'com.github.javaparser:javaparser-core:2.0.0',
-           'com.tdunning:t-digest:3.2'
+            "org.codehaus.groovy:groovy:$GROOVY_VERSION",
```
jcferretti: Was this change intentional?

rcaudy: I intentionally corrected bad whitespace on the left.
The only material change is a new dependency:
```
            'com.fasterxml.jackson.datatype:jackson-datatype-jdk8:2.9.8',
```

DB/src/main/java/io/deephaven/db/v2/sources/regioned/RegionedColumnSourceWithDictionary.java
rcaudy: Not sure I follow the question. I don't want to read the `volatile wrapperRegions` reference twice, nor do I want to do the array lookup twice. It would *probably be fine* in this case, but it's something I avoid out of habit and hygiene.
I'm reasonably convinced I have the right thread-safety decisions here; by making `localWrappers` volatile I avoid any risk that someone might see an incompletely copied version of the array, and wrapper sources are immutable.

DB/src/main/java/io/deephaven/db/v2/sources/regioned/RegionedColumnSourceManager.java
rcaudy: This could arguably happen with a bad publisher, it doesn't have to be an engine bug. I tend to think for possibly-user-facing errors we should use exceptions with clean messages, rather than assert. I'm sensitive to your heap dump point, though, and so I'm planning to change to:
```
                    Assert.statementNeverExecuted("Index keys removed at location " + location + ": " + indexAtLastUpdate.minus(updateIndex));
```
The comment regarding `grpc_api` main is not something I intend to worry about at this layer; if we encounter an error we throw an exception. The calling code needs to be wrapped in a try-catch block in order to surface appropriate errors to the user. We shouldn't crash just because an engine op failed, unless we setup failure listeners to make that happen deliberately.

DB/src/main/java/io/deephaven/db/v2/parquet/ParquetTableWriter.java
```diff
@@ -396,81 +415,95 @@ public Integer get() {
...
-                Object nullValue = getNullValue(columnType);
+                final boolean supportNulls = supportNulls(columnType);
+                final Object bufferToWrite = transferObject.getBuffer();
+                final Object nullValue = getNullValue(columnType);
                 try (final OrderedKeys.Iterator lengthIndexIt = lengthIndex != null ? lengthIndex.getOrderedKeysIterator() : null;
                      final ChunkSource.GetContext lengthSourceContext = lengthSource != null ? lengthSource.makeGetContext(targetSize) : null;
                      final OrderedKeys.Iterator it = index.getOrderedKeysIterator()) {
-                    IntBuffer repeatCount = lengthSource != null ? IntBuffer.allocate(targetSize) : null;
+                    final IntBuffer repeatCount = lengthSource != null ? IntBuffer.allocate(targetSize) : null;
                     for (int step = 0; step < stepsCount; step++) {
```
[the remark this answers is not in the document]
rcaudy: I just marked the field `final`, I didn't sign up to make everything perfect in this file on this pass. That said, I've changed it to `++step`.

DB/src/main/java/io/deephaven/db/v2/locations/parquet/topage/ToPageWithDictionary.java
```diff
@@ -44,59 +46,61 @@ public final ChunkType getChunkType() {
...
     @NotNull
-    public final StringSetImpl.ReversibleLookup<T> getReversibleLookup() {
+    public final StringSetImpl.ReversibleLookup<DATA_TYPE> getReversibleLookup() {
         return dictionary;
     }
 
     @Override
     @NotNull
-    public final ToPage<Attributes.DictionaryKeys, int[]> getDictionaryKeysToPage() {
-        return new ToPage<Attributes.DictionaryKeys, int[]>() {
```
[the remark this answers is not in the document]
rcaudy: Long story, but totally worth explaining. It was a bit of a trade-off.

The short answer is: because that's what `convertResult` returns now.

Most of the code that uses `getDictionaryKeysToPage` is just looking to get the `int[]` returned by `getResult` (as an `Object`) and convert it to the intended upward-facing data type before wrapping that in a chunk and caching it. The exception is the symbol table code up in `RegionedColumnSourceWithDictionary`. We expose `long` ids that live in the same region address space they came from, but with their sub-region offsets within the dictionary's space.

For example, in a nested structure, maybe I'm in outer region 20, inner region 5 (out of 10, so 4 bits of address space), and the dictionary offset is 1911, so the resulting symbol id is:
```
20 << 43 | 5 << 39 | 1911
178670639515511
```
So, to implement this, I need to get `1911` and its friends from the lower level, and then OR on the higher levels' region offsets. The easiest way is to fill a chunk of longs from the bottom (dictionary) region, and than manipulate it in-place.

Note that (`SoftReference`-based) caching built into our Parquet regions is on the materialized results (so after `convertResult`, not `getResult`). I decided it made more sense to convert to longs and cache that, so that my `fillChunk` looks like an array copy, than to cache the ints and have to do a row-by-row cast when retrieving values. This also avoided adding yet-another wrapper page.

DB/src/main/java/io/deephaven/db/v2/locations/parquet/local/ParquetColumnLocation.java
```diff
@@ -70,157 +105,250 @@ public String getImplementationName() {
...
+        // pageStores array
+        return columnChunkReaders != null || pageStores != null;
+    }
 
-        return pageStore != null;
+    private ParquetTableLocation tl() {
+        return (ParquetTableLocation) getTableLocation();
     }
 
+    private static final ColumnDefinition<Long> FIRST_KEY_COL_DEF = ColumnDefinition.ofLong("__firstKey__");
```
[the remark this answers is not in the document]
rcaudy: I hate GitHub. I wrote you a long reply, and it got wiped out by a backpage.

Shorter version: those aren't the column names we write to file. These are:
```
    public static final String BEGIN_POS = "dh_begin_pos";
    public static final String END_POS = "dh_end_pos";
    public static final String GROUPING_KEY = "dh_key";
```

That said, the original author of this file likely had a healthy fear of clashing with user column names. It's just good hygiene, like brushing your teeth. Those two names are just used for now to construct ephemeral `ColumnDefinition` objects used in `ToPage` construction. That said, I would leave them alone. Down the long and winding roadmap a bit, we'll be exposing grouping info as tables and deleting the key->`Index` map fandango you're used to, at which point we will definitely want reserved or at least non-user-like names for these columns.

DB/src/main/java/io/deephaven/db/v2/locations/local/ParquetMetadataFileLayout.java
```diff
@@ -0,0 +1,235 @@
...
+                    throw new TableDataException("Unexpected number of path elements in " + filePathString + " for partitions " + partitions.keySet());
+                }
+                final boolean useHiveStyle = filePath.getName(0).toString().contains("=");
+                for (int pi = 0; pi < numPartitions; ++pi) {
+                    final String pathElement = filePath.getName(pi).toString();
+                    final ColumnDefinition columnDefinition;
+                    final String partitionKey;
+                    final String partitionValueRaw;
+                    if (useHiveStyle) {
+                        final String[] pathComponents = pathElement.split("=");
```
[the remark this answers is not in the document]
rcaudy: Debatable, but not an unreasonable point. If someone wanted to have "=" in their value (but not their column name) I might as well accept it. Will change it.

DB/src/main/java/io/deephaven/db/v2/locations/local/FileTableLocationKey.java
```diff
@@ -22,21 +22,24 @@
...
     protected final File file;
+    private final int order;
 
     private int cachedHashCode;
 
     /**
      * Construct a new FileTableLocationKey for the supplied {@code file} and {@code partitions}.
      *
      * @param file        The file (or directory) that backs the keyed location. Will be adjusted to an absolute path.
+     * @param order      Explicit ordering index, taking precedence over other fields
```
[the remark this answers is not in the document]
rcaudy:
```
     * @param order      Explicit ordering value for this location key. {@link Comparable#compareTo(Object)} will sort
     *                   FileTableLocationKeys with a lower {@code order} before other keys. Comparing this ordering
     *                   value takes precedence over other fields.
 ```

Also, on the `compareTo`:
```
    /**
     * Precedence-wise this implementation compares {@code order}, then applies a {@link PartitionsComparator} to
     * {@code partitions}, then compares {@code file}.
     * @inheritDoc
     */
 ```

DB/src/main/java/io/deephaven/db/tables/utils/ParquetTools.java
```diff
@@ -438,9 +451,31 @@ public static Table readMultiFileTable(
             allColumns.add(ColumnDefinition.fromGenericType(partitionKey, getUnboxedTypeIfBoxed(partitionValue.getClass()), ColumnDefinition.COLUMNTYPE_PARTITIONING, null));
         }
         allColumns.addAll(schemaInfo.getFirst());
-        return readMultiFileTable(recordingLocationKeyFinder, schemaInfo.getSecond(), new TableDefinition(allColumns));
+        return readPartitionedTable(recordingLocationKeyFinder, schemaInfo.getSecond(), new TableDefinition(allColumns));
+    }
+
+    /**
+     * Reads in a table using metadata files found in the supplied directory.
```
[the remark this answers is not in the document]
rcaudy: Methods now say:
```
     * @see SingleParquetFileLayout
     * @see ParquetMetadataFileLayout
     * @see KeyValuePartitionLayout
     * @see FlatParquetLayout
```
`ParquetMetadataFileLayout` now has:
```
/**
 * <p>{@link TableLocationKeyFinder Location finder} that will examine a parquet metadata file to discover locations.
 *
 * <p>Note that we expect to find the following files:
 * <ul>
 *     <li>{@code _metadata} - A file containing Parquet metadata for all {@link RowGroup row groups} in all
 *     {@code .parquet} files for the entire data set, including schema information non-partitioning columns and
 *     key-value metadata</li>
 *     <li>{@code _common_metadata} <i>(optional)</i> - A file containing Parquet metadata with schema information
 *     that applies to the entire data set, including partitioning columns that are inferred from file paths rather than
 *     explicitly written in {@link org.apache.parquet.format.ColumnChunk column chunks} within {@code .parquet}
 *     files</li>
 * </ul>
 */
```

DB/src/main/java/io/deephaven/db/v2/locations/parquet/topage/ToPageWithDictionary.java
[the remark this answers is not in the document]
jcferretti: Rewriting the int[] to long[] is sad, but I guess such is life.

DB/src/main/java/io/deephaven/db/v2/sources/regioned/RegionedColumnSourceWithDictionary.java
jcferretti: This is the expression in the if and also in the later if inside the synchronized:
`(localWrappers = wrapperRegions).length > regionIndex && (wrapper = localWrappers[regionIndex]) != null)`

Outside the synchronized block, if that expression evaluates to true, we bail.  So the assignments to localWrappers and wrapper are useless at that point.  So it is confusing for me why we are doing them.

rcaudy: Discussed on the phone.

rcaudy:
Closes #868 too

jcferretti (approved):

PR #955, opened 2021-08-02
Fix #950 SortedRanges.SearchIterator hasNext() still returns true after exhaustion via advance

### LINKED ISSUE #950: SortedRanges.SearchIterator hasNext() still returns true after exhaustion via advance
My code advances a `ReadOnlyIndex.SearchIterator` past the end (using `advance`, which correctly returns `false`). Afterwards, `hasNext()` returns `true`.

I believe this is the correct fix to apply:
```
diff --git a/DB/src/main/java/io/deephaven/db/v2/utils/sortedranges/SortedRanges.java b/DB/src/main/java/io/deephaven/db/v2/utils/sortedranges/SortedRanges.java
index 9c187f8c..99dbfb99 100644
--- a/DB/src/main/java/io/deephaven/db/v2/utils/sortedranges/SortedRanges.java
+++ b/DB/src/main/java/io/deephaven/db/v2/utils/sortedranges/SortedRanges.java
@@ -557,6 +557,9 @@ public abstract class SortedRanges extends RefCountedCow<SortedRanges> implement
                 return true;
             }
             if (sar == null || nextRangeIdx == sar.count) {
+                currRangeStart = currRangeEnd = -1;
+                closeImpl();
                 return false;
             }
             int p = sar.unpackedBinarySearch(v, nextRangeIdx);
```

We should also check the other SearchIterator implementations, and maybe write a unit test.

jcferretti:


DB/src/main/java/io/deephaven/db/v2/utils/sortedranges/SortedRanges.java
```diff
@@ -556,7 +556,13 @@ public final boolean advanceImpl(final long v) {
                 }
                 return true;
             }
-            if (sar == null || nextRangeIdx == sar.count) {
+            if (sar == null) {
```
rcaudy, line 559: Is it really worth having two `if`s since `closeImpl` checks for `null`?

DB/src/main/java/io/deephaven/db/v2/utils/sortedranges/SortedRanges.java
```diff
@@ -875,10 +881,15 @@ public boolean advance(final long v) {
                 rangeCurr = Math.min(v, rangeCurr);
                 return true;
             }
-            if (nextRangeIdx < 0 || sar == null) {
+            if (sar == null) {
```
rcaudy, line 884: Same question as above. Also, did this iterator have "the" bug, or just a resource leak?

rcaudy (dismissed):

DB/src/main/java/io/deephaven/db/v2/utils/sortedranges/SortedRanges.java
jcferretti: Not the bug.

rcaudy (approved):

Issue #412, opened 2021-04-19
Decouple barrage's metadata from Flight's RecordBatch


nbauernfeind:
The conclusive feedback from arrow's dev list is to integrate with Arrow Flight via `app_metadata` data fields.

The criteria for completing this work is:
- flatbuffer defined for a delta (including how many record batches in total for adds followed by mods); must also include the indexes for mod columns
- split adds and mods into separate record batches (client must accumulate) -- does not need to generate multiple adds/mods (that is #179)
- some form of polymorphic DoExchange
- implement DoExchange for subscriptions
- flatbuffer for the actual subscription
- replace new session logic with arrow flight handshakes (deferred as #997)

nbauernfeind:
This work was completed in #1020.

Follow up tasks are:
- #179 
- #997

PR #956, opened 2021-08-03
Clean up proto packages/namespaces, remove flight ticket

### LINKED ISSUE #937: Clean up proto files
- [ ] Add python codegen
- [ ] Capitalize .proto files that we provide
- [ ] Put files in top-level of the source dir so that languages like python are happier - might need to consider adding a package to it again, will discuss with @jmao-denver

niloc132:
Will make it easier to auto-generate python code for grpc calls, and start first steps of updating to barrage v0.2.0.

Also adds some TODOs for followup as part of #412. JS Flight implementation is temporarily removed, will also return in 412.

Fixes #937


grpc-api/src/main/java/io/deephaven/grpc_api/barrage/BarrageClientSubscription.java
```diff
@@ -38,39 +37,33 @@
     private volatile boolean connected = false;
 
     private final String logName;
-    private final Flight.Ticket handle;
```
niloc132, line 41: This was apparently unused, so I just removed it rather than change the type.

proto/proto-backplane-grpc/Dockerfile
```diff
@@ -2,23 +2,33 @@ FROM deephaven/protoc:local-build
 
 COPY src/main/proto /includes
 
-COPY dependencies /dependencies
+#COPY dependencies /dependencies
```
niloc132, line 5: remove this

proto/proto-backplane-grpc/build.gradle
```diff
@@ -21,9 +21,9 @@ TaskProvider<Task> generateProtobuf = Docker.registerDockerTask(project, 'genera
       exclude 'build'
       exclude 'build.gradle'
     }
-    from(zipTree(configurations.download.singleFile)) {
-      include("flight.proto").into("dependencies/arrow/flight/protocol")
-    }
+//    from(zipTree(configurations.download.singleFile)) {
```
niloc132, line 24: remove this

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/ticket.proto
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
+ */
+
+syntax = "proto3";
+
+package io.deephaven.proto.backplane.grpc;
+
+option java_multiple_files = true;
+option optimize_for = SPEED;
+
```
niloc132, line 11: add file comments that this is copied from arrow

proto/proto-backplane-grpc/Dockerfile
niloc132: changed my mind, leaving this to re-add it with #412

proto/proto-backplane-grpc/build.gradle
niloc132: changed my mind, leaving this to re-add it with #412

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionState.java
```diff
@@ -597,7 +619,7 @@ public T get() {
         /**
          * @return the export id or NON_EXPORT_ID if it does not have one
```
nbauernfeind, line 620: javadoc and method name do not match up; should we rename to `getTicket`?

grpc-api/src/main/java/io/deephaven/grpc_api/session/TicketRouter.java
```diff
@@ -70,6 +85,7 @@ public TicketRouter(final Set<TicketResolver> resolvers) {
      * @param <T> the expected return type of the ticket; this is not validated
      * @return an export object; see {@link SessionState} for lifecycle propagation details
      */
+    //TODO #412 use this or remove it?
```
nbauernfeind, line 88: we can drop this TODO, too (and the ones below; or again, I will look for and clear these)

proto/proto-backplane-grpc-util/src/main/java/io/deephaven/grpc_api/util/ExportTicketHelper.java
```diff
@@ -116,6 +143,17 @@ public static int descriptorToExportId(final Flight.FlightDescriptor descriptor)
...
+        return exportIdToDescriptor(ticketToExportId(ticket));
+    }
+
+    /**
+     * Convenience method to convert from a Flight.Ticket to a Flight.FlightDescriptor.
+     *
+     * @param ticket the ticket to convert
+     * @return a flight descriptor that represents the ticket
+     */
+    // TODO #412 use this or remove it
```
nbauernfeind, line 156: Same on these two TODOs.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/console.proto
```diff
@@ -46,11 +46,11 @@ message GetConsoleTypesResponse {
 }
 
 message StartConsoleRequest {
-    arrow.flight.protocol.Ticket result_id = 1;
+    io.deephaven.proto.backplane.grpc.Ticket result_id = 1;
```
nbauernfeind, line 49: should we drop the prefixes instead? they should be in the same namespace now

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionState.java
niloc132: hmm, not even my fault!

I should remove the "or NON_EXPORT_ID..." right?

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/console.proto
niloc132: Nope, console is in .script.grpc, since it is a different package it needs qualified imports (same as ETCR references in this file)

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionState.java
nbauernfeind: I think I would state:
```
@return the ticket for this export; note if this is a non-export the received ticket will not resolve to anything and is considered an invalid ticket
```

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/console.proto
nbauernfeind: I wonder if it make sense to be in another package anymore. Especially if it gets the package convention from the file structure in python. Either way, we can table this for now. Just food for thought.

nbauernfeind (approved):

Issue #874, opened 2021-07-21
DH Python Client: doPut-based interface to push data to DH tables


jmao-denver:
Symmetrical to table snapshotting which is based on Flight's doGet,  the ability to populate a DH table with Flight's doPut call is essential to a minimal viable Python client.

jmao-denver:
closed by #1094

Issue #873, opened 2021-07-21
Flesh out the query/batch operations in the Python client API


jmao-denver:
Repurpose the ticket to complete all the query/batch operations in the Python client API.
 
QST is a totally different direction and should be considered for future feature addition.

Original description: In the spirit of having as much as commonalities as possible with other planned DH clients and taking advantage of the good work already done in the Java client, we want to have a QST-based Query interface that is extensible and easy to use.

jmao-denver:
closed by #1094

Issue #776, opened 2021-06-23
Write a prototype for a Python API that can deliver data to a Python client from Deephaven


pete-petey:
Pete's comments about this ticket -- these are not a mandate:

- The goal is to focus on this one aspect of the Python API -- accessing Deephaven data from a Python client.  (Yes, arguably, at some point it should be part of a larger, more multi-functional, "proper" Python client, but for this exercise let's ignore that.

- The word "prototype" for this ticket is appropriate.  We want to build something that simply convinces ourselves that the approach taken is the right one.

-- The data accessible to the Python client must be deephaven tables in static form, snapshots, and updates.  

-- Per Jianfeng, it is likely wise to first explore whether Python-GRPC API and PyArrow (or Py-Arrow-FLight + Barrage) can directly provide what the user wants (without involving ourselves in wrapping C++).

@niloc132  @devinrsmith  @nbauernfeind  will all be helpful I am sure.  (Ryan too, but not having him watch it.)

jmao-denver:
closed by #1094

PR #959, opened 2021-08-03
feature-776, implemented the Python client with limited functionality

jmao-denver:
missing some of the table ops, but otherwise is usable, especially with the support of doGet and doPut. No ticking table support yet.

partially #776, #873, #874


jmao-denver:
No embedded Python Doc either, will add after #961 is finished.

margaretkennedy (approved):

pyclient/deephaven/session.py
```diff
@@ -0,0 +1,279 @@
...
+from deephaven.table_service import TableService
+
+
+class Session:
+    def __init__(self, host='localhost', port=10000, user="", password=""):
+        self._r_lock = threading.RLock()
+        self._last_ticket = 0
+        self._ticket_bitarray = BitArray(1024)
+        self.host = os.environ.get("DHCE_HOST", host)
+        self.port = os.environ.get("DHCE_PORT", port)
```
devinrsmith, line 22: Typically, we want an explicitly passed in value to take precedence over an environment variable, right?

I think we want the precedence:

1. Manually value
2. Env value
3. Default value

pyclient/deephaven/batch_assembler.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def visit(self, op):
+        if isinstance(op, NoneOp):
+            self._current_table_ref = table_pb2.TableReference(ticket=op.table.ticket)
+        elif isinstance(op, UpdateOp):
+            self.build_update(op)
+        elif isinstance(op, TailOp):
+            self.build_tail()
+        elif isinstance(op, HeadOp):
+            self.build_head()
```
devinrsmith, line 31: In java, the visit method is implemented by each concrete type to make sure that no "isinstance" checks need to be done. I'm not sure exactly what it would look like here, but maybe something like:

```python
def build(self, op):
  op.walk(BuildVisitor())
```

pyclient/deephaven/batch_assembler.py
```diff
@@ -0,0 +1,53 @@
...
+    def _set_last_op(self, op):
+        self._last_op = op
+
+    def visit(self, op):
+        if isinstance(op, NoneOp):
+            self._current_table_ref = table_pb2.TableReference(ticket=op.table.ticket)
+        elif isinstance(op, UpdateOp):
+            self.build_update(op)
+        elif isinstance(op, TailOp):
+            self.build_tail()
```
devinrsmith, line 29: I'm assuming this hasn't been tested out, since `op` looks like it needs to be passed through, but isn't?

pyclient/deephaven/qst.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class TableOp(ABC):
+    def __init__(self, parent=None):
+        self.parent = parent
+        if parent:
+            self.parent.child = self
+        self.child = None
+
+    def accept(self, visitor):
```
devinrsmith, line 12: In java, this is an abstract method - each concrete impl is response for implementing. Given that overloading is a bit different though, I might except each concrete impl calls a method based on its own type:

```python
class UpdateOp(TableOp):
    def accept(self, visitor):
      visitor.visitUpdate(self)
      
class ViewOp(TableOp):
    def accept(self, visitor):
      visitor.visitView(self)
...
```

pyclient/setup.py
```diff
@@ -0,0 +1,20 @@
+from setuptools import setup
+
+setup(
+    name='deephaven',
```
devinrsmith, line 4: I wonder if we should have a distinct name for this package, so we don't confuse it for the package installed on the server?

devinrsmith (commented):
Very fast first pass. I'd like to dig in more next week.

pyclient/setup.py
chipkent: I agree that there is a problem.  With the same name, we have confusion between client and server, and we have problems with generating pydocs in a reasonable way.

Maybe `deephaven_client`?

pyclient/tests/test_query.py
```diff
@@ -0,0 +1,25 @@
+import time
+
+from tests.testbase import BaseTestCase
+
+
+class QueryTestCase(BaseTestCase):
```
chipkent, line 6: This is only testing linear DAGs, so I can't see what the plan is for non-linear cases.

pyclient/tests/test_query.py
```diff
@@ -0,0 +1,25 @@
...
+        self.assertEqual(10, rlt.size)
+        time.sleep(1)
+        rlt2 = query.exec()
+        self.assertEqual(rlt.size, rlt2.size)
+
+    def test_tail_update_ticking(self):
+        tbl = self.session.time_table(period=100000)
+        query = self.session.query(tbl).update(column_specs=["Col1=i + 1"]) \
+            .tail(5).update(column_specs=["Col2=i*i"])
+        rlt = query.exec()
```
chipkent, line 21: Looking at this example, I have two questions:
1) How do you handle multiple inputs to the query
2) How do you handle multiple outputs to the query
In both cases, it seems like what is here may have problems.

pyclient/tests/testbase.py
```diff
@@ -0,0 +1,32 @@
+import unittest
+import warnings, os
+
+from deephaven import Session
+from tests.gen_test_data import make_random_csv
+
+
+class BaseTestCase(unittest.TestCase):
+    csv_file = 'test_csv_file'
```
chipkent, line 9: you don't want a `.csv` suffix?

chipkent (changes requested):

jmao-denver:
because of the lack of sufficient documentation, the PR is very hard to review.  I am closing it and resubmit a new one #1094 which contains a more complete implementation and Sphinx generated documentation.

PR #965, opened 2021-08-03
dramatically reduce log spam for unauthenticated requests

nbauernfeind:
The jsapi should stop making requests that require authentication once it receives an unauthenticated response. We will tackle that in another review, and maybe not all jsapi end points all at once (I'm thinking the auto complete ones are a nice target).

### DIFF
--- a/grpc-api/src/main/java/io/deephaven/grpc_api/util/GrpcUtil.java
+++ b/grpc-api/src/main/java/io/deephaven/grpc_api/util/GrpcUtil.java
@@ -22,8 +22,7 @@ public class GrpcUtil {
             lambda.run();
         } catch (final StatusRuntimeException err) {
             if (err.getStatus().equals(Status.UNAUTHENTICATED)) {
-                // most of the time we do not want to log this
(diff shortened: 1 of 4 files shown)


grpc-api/src/main/java/io/deephaven/grpc_api/util/GrpcUtil.java
```diff
@@ -22,7 +21,12 @@ public static void rpcWrapper(final Logger log, final StreamObserver<?> response
         try (final SafeCloseable ignored = LivenessScopeStack.open()) {
             lambda.run();
         } catch (final StatusRuntimeException err) {
-            log.error().append(err).endl();
+            if (err.getStatus().equals(Status.UNAUTHENTICATED)) {
+                // most of the time we do not want to log this
+                log.debug().append("ignoring unauthenticated request").endl();
```
rcaudy, line 26: Should we log (to debug) what we're ignoring?
"Ignoring unauthenticated request"

rcaudy (dismissed):

PR #977, opened 2021-08-06
Kafka: Stream firstBy and lastBy chunked operators, address other engine features for stream-tables

### LINKED ISSUE #978: Kafka: Build and integrate stream-table first and last by chunked operators
Linked to #293

### LINKED ISSUE #979: Kafka: Integrate stream tables into `Table` and `QueryTable`
Linked to #293 
Requirements:
1. Introduce an `isStream()` interface method, analogous to `isAddOnly()`.
2. Update stream table JavaDocs to note that the contract is that data never exists for more than one cycle. Even if no new data is added, all data will still be removed on the next cycle. This is important for the correctness of stream-preserving filters with fancy filtering.
2. Introduce a `dropStream()` interface method that does a `copy()` and eliminates the `STREAM_TABLE_ATTRIBUTE` from the result. Necessary to allow aggs to "work like normal" on just the data available each cycle, rather than over the full history.
3. Attribute passes through most operations, including view/select/updateView/update/dropColumns, sort (theoretically), where (including in, not in), all joins, flatten, coalesce, wouldMatch, reverse, ByExternal, preview
4. `treeTable()` should refuse to work on streams (it just doesn't make sense to do)
5. `rollup()` should refuse to work on streams if including constituents
6. `by()` should refuse to work on streams. We could make it work as a stream-preserving op (converting downstream mods to removes+adds), but then it would have totally other semantics from other stream aggs, and this seems bad. Users must `dropStream()` to allow `by` on just the current-cycle contents, same as other operations.
7. The aggregation helper should ignore upstream removes and assert no mods or shifts when its input is a stream table.
8.  `ComboAggregateFactory#makeAggregationContextFactory` and `QueryTable#byNoMemo` should treat stream tables the same as add-only tables. `FirstOrLastByAggregationFactory` should use the new operators.
9. Write a helper for "record full history in-memory via `snapshotHistory()`".

rcaudy:
Closes #978 
Closes #979
Part of #293

### DIFF
--- a/DB/src/main/java/io/deephaven/db/v2/by/StreamFirstByChunkedOperator.java
+++ b/DB/src/main/java/io/deephaven/db/v2/by/StreamFirstByChunkedOperator.java
@@ -5,7 +5,8 @@ import io.deephaven.db.tables.Table;
 import io.deephaven.db.tables.select.MatchPair;
 import io.deephaven.db.v2.QueryTable;
 import io.deephaven.db.v2.ShiftAwareListener;
-import io.deephaven.db.v2.sources.*;
+import io.deephaven.db.v2.sources.WritableChunkSink;
+import io.deephaven.db.v2.sources.WritableSource;
 import io.deephaven.db.v2.sources.chunk.Attributes.*;
 import io.deephaven.db.v2.sources.chunk.*;
 import io.deephaven.db.v2.utils.Index;
@@ -15,54 +16,12 @@ import io.deephaven.db.v2.utils.ShiftedOrderedKeys;
 import io.deephaven.util.SafeCloseableList;
 import org.jetbrains.annotations.NotNull;
 
-import java.lang.ref.SoftReference;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.Map;
-
 /**
  * A firstBy aggregation operator for stream tables.
  *
  * @see Table#STREAM_TABLE_ATTRIBUTE
  */
-public class StreamFirstByChunkedOperator implements IterativeChunkedAggregationOperator {
-
-    private static final int COPY_CHUNK_SIZE = ArrayBackedColumnSource.BLOCK_SIZE;
-
-    /**
-     * The number of result columns. This is the size of {@link #resultColumns} and the length of {@link #inputColumns}
-     * and {@link #outputColumns}.
-     */
-    private final int numResultColumns;
-
-    /**
-     * Result columns, parallel to {@link #inputColumns} and {@link #outputColumns}.
-     */
-    private final Map<String, ArrayBackedColumnSource<?>> resultColumns;
-
-    /**
-     * <p>Input columns, parallel to {@link #outputColumns} and {@link #resultColumns}.
-     * <p>These are the source columns from the upstream table, reinterpreted to primitives where applicable.
-     */
-    private final ColumnSource<?>[] inputColumns;
-
-    /**
-     * <p>Output columns, parallel to {@link #inputColumns} and {@link #resultColumns}.
-     * <p>These are the result columns, reinterpreted to primitives where applicable.
-     */
-    private final WritableSource<?>[] outputColumns;
-
-    /**
-     * Cached pointer to the most-recently allocated {@link #redirections}.
-     */
-    private SoftReference<LongArraySource> cachedRedirections;
-
-    /**
-     * Map from destination slot to first key. Only used during a step to keep track of the appropriate rows to copy into
-     * the output columns.
-     */
-    private LongArraySource redirections;
+public class StreamFirstByChunkedOperator extends StreamFirstOrLastByChunkedOperator {
 
     /**
      * <p>The next destination slot that we expect to be used.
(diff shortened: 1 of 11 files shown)


DB/src/main/java/io/deephaven/db/v2/by/StreamFirstByChunkedOperator.java
```diff
@@ -0,0 +1,237 @@
...
+        inputColumns = new ColumnSource[numResultColumns];
+        outputColumns = new WritableSource[numResultColumns];
+        final Map<String, ArrayBackedColumnSource<?>> resultColumnsMutable = new LinkedHashMap<>(numResultColumns);
+        for (int ci = 0; ci < numResultColumns; ++ci) {
+            final MatchPair resultPair = resultPairs[ci];
+            final ColumnSource<?> streamSource = streamTable.getColumnSource(resultPair.left());
+            final ArrayBackedColumnSource<?> resultSource = ArrayBackedColumnSource.getMemoryColumnSource(0, streamSource.getType(), streamSource.getComponentType());
+            resultColumnsMutable.put(resultPair.left(), resultSource);
+            inputColumns[ci] = ReinterpretUtilities.maybeConvertToPrimitive(streamSource);
+            outputColumns[ci] = (WritableSource<?>) ReinterpretUtilities.maybeConvertToPrimitive(resultSource);
```
cpwright, line 54: This feels backwards; that we would want to have the outputColumn be an array backed column source of the primitive not of the type from the streamSource; and then you would wrap it in the opposite.

DB/src/main/java/io/deephaven/db/v2/by/StreamLastByChunkedOperator.java
```diff
@@ -0,0 +1,254 @@
...
+        outputColumns = new WritableSource[numResultColumns];
+        permuteKernels = new PermuteKernel[numResultColumns];
+        final Map<String, ArrayBackedColumnSource<?>> resultColumnsMutable = new LinkedHashMap<>(numResultColumns);
+        for (int ci = 0; ci < numResultColumns; ++ci) {
+            final MatchPair resultPair = resultPairs[ci];
+            final ColumnSource<?> streamSource = streamTable.getColumnSource(resultPair.left());
+            final ArrayBackedColumnSource<?> resultSource = ArrayBackedColumnSource.getMemoryColumnSource(0, streamSource.getType(), streamSource.getComponentType());
+            resultColumnsMutable.put(resultPair.left(), resultSource);
+            inputColumns[ci] = ReinterpretUtilities.maybeConvertToPrimitive(streamSource);
+            outputColumns[ci] = (WritableSource<?>) ReinterpretUtilities.maybeConvertToPrimitive(resultSource);
```
cpwright, line 60: Again my backwards assertion.

DB/src/main/java/io/deephaven/db/v2/by/StreamFirstByChunkedOperator.java
```diff
@@ -0,0 +1,237 @@
...
+            Assert.eq(inputColumns[ci].getChunkType(), "inputColumns[ci].getChunkType()", outputColumns[ci].getChunkType(), "outputColumns[ci].getChunkType()");
+        }
+        resultColumns = Collections.unmodifiableMap(resultColumnsMutable);
+        cachedRedirections = new SoftReference<>(redirections = new LongArraySource());
+    }
+
+    @Override
+    public void ensureCapacity(final long tableSize) {
+        redirections.ensureCapacity(tableSize - previousTableSize);
```
cpwright, line 63: This is subtle enough that you should have a comment.  Maybe a comment on the overall strategy of knowing that you can only have new things when they are bigger than previousTableSize.

DB/src/main/java/io/deephaven/db/v2/by/StreamFirstByChunkedOperator.java
```diff
@@ -0,0 +1,237 @@
...
+                         final Chunk<? extends Values> values, // Unused
+                         @NotNull final LongChunk<? extends KeyIndices> inputIndices,
+                         @NotNull final IntChunk<KeyIndices> destinations,
+                         @NotNull final IntChunk<ChunkPositions> startPositions,
+                         final IntChunk<ChunkLengths> length, // Unused
+                         @NotNull final WritableBooleanChunk<Values> stateModified) {
+        for (int ii = 0; ii < startPositions.size(); ++ii) {
+            final int startPosition = startPositions.get(ii);
+            final long destination = destinations.get(startPosition);
+            redirections.set(destination - previousTableSize, inputIndices.get(startPosition));
```
cpwright, line 104: How are you checking that this isn't just an additon to a previous state and should be ignored?

DB/src/main/java/io/deephaven/db/v2/by/StreamFirstByChunkedOperator.java
rcaudy: I'm pretty sure you're wrong about this; we (you?) did a good job designing the array-backed column sources for this use case:
`DateTimeArraySource.ReinterpretedAsLong` is just as efficient for filling with longs as `LongArraySource`, and `DateTimeArraySource` is more efficient for getting `DBDateTimes` out of than `LongAsDateTimeColumnSource` wrapped around `LongArraySource`.
BooleanArraySource.ReinterpretedAsByte` is similarly good for filling with bytes, and I think `BooleanArraySource` is way better for reading `Booleans` from than `ByteAsBooleanColumnSource` wrapped around `ByteArraySource`.

DB/src/main/java/io/deephaven/db/v2/by/StreamLastByChunkedOperator.java
rcaudy: See note in https://github.com/deephaven/deephaven-core/pull/977#discussion_r684845459

DB/src/main/java/io/deephaven/db/v2/by/StreamFirstByChunkedOperator.java
rcaudy: Sufficiently clear?
```
    /**
     * <p>The next destination slot that we expect to be used.
     * <p>Any destination less than this one can safely be ignored while processing adds since the first row can never
     * change once a destination has been created given that we ignore removes.
     */
    private long nextDestination;

    /**
     * <p>The first destination that we used on the current step (if we used any). At the very beginning of a step, this
     * is equivalent to {@link #nextDestination} and also the result table's size.
     * <p>We use this as an offset shift for {@code redirections}, so that {@code redirections} only needs to hold first
     * source keys for newly-added destinations, rather than the entire space.
     * <p>At the end of a step, this is updated to prepare for the next step.
     */
    private long firstDestinationThisStep;
```

DB/src/main/java/io/deephaven/db/v2/by/StreamFirstByChunkedOperator.java
rcaudy: Oops, I wasn't. Fixed in next commit.

rcaudy:
This still needs unit tests and testing, but it may be complete otherwise.

DB/src/main/java/io/deephaven/db/v2/by/StreamFirstOrLastByChunkedOperator.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @Override
+    public final Map<String, ? extends ColumnSource<?>> getResultColumns() {
+        return resultColumns;
+    }
+
+    @Override
+    public final void startTrackingPrevValues() {
+        Arrays.stream(outputColumns).forEach(ColumnSource::startTrackingPrevValues);
```
cpwright, line 79: For firstBy I think this is unnecessary, because the columns will never change after you write a value down.

cpwright (commented):
I would approve if there were unit tests.

DB/src/main/java/io/deephaven/db/v2/by/StreamFirstOrLastByChunkedOperator.java
rcaudy: To clarify for posterity: this makes sense because it's undefined to read previous values for rows that weren't present on the previous cycle.

cpwright (approved):

PR #989, opened 2021-08-06
remove QueyScope from Deephaven Python package, fixes #846

### LINKED ISSUE #846: Unable to add Python objects to the QueryScope
Python objects can not be added to the QueryScope manually.  The same objects can be added automatically at the top level without problems.  I suspect this is a problem with assigning Python objects to `java.lang.Object` as illustrated below.

```
from deephaven import QueryScope

class MyClass:
    def __init__(self):
        pass

    def do_it(self):
        return 3


def in_func(t):
    mc2 = MyClass()
    #This fails with an error
    QueryScope.addParam("mc2", mc2)
    return t.update("X = mc2.do_it()")

t = emptyTable(3)

# this works fine
mc1 = MyClass()
t1 = t.update("X = mc1.do_it()")

# This results in an error
t2 = in_func(t)

```

image

I suspect the problem is an inability to assign a Python object to a `java.lang.Object`.  This can be seen here:

```

import jpy

class MyClass:
    def __init__(self):
        pass

    def do_it(self):
        return 3

mc1 = MyClass()

a = jpy.array("java.lang.Object", 3)
a[0] = mc1
```
image

jmao-denver:
The QueryScope wrapper in Deephaven Python package is very confusing to use and doesn't do what it is expected to do. We have decided to simply use Python's own globals() to achieve the goal of making Python objects accessible by the DH Query engine. By removing the wrapper from the Deephaven package eliminates the chance of leaving in place incorrect old sample codes which would have continued to work otherwise.


Generators/Generators.gradle
```diff
@@ -215,8 +215,8 @@ def pythonStaticMethodsArgs = ['io.deephaven.util.calendar.Calendars',
                                'io.deephaven.db.tables.utils.DBTimeUtils',
                                     '/Generators/src/main/java/io/deephaven/pythonPreambles/DBTimeUtilsPreamble.txt',
                                     '/Integrations/python/deephaven/DBTimeUtils/__init__.py',
-                               'io.deephaven.db.tables.select.QueryScope',
-                                    '/Integrations/python/deephaven/QueryScope/__init__.py',
+//                               'io.deephaven.db.tables.select.QueryScope',
+//                                    '/Integrations/python/deephaven/QueryScope/__init__.py',
```
chipkent, line 219: Just delete it if we aren't using it.

Generators/src/main/java/io/deephaven/python/PythonStaticGenerator.java
```diff
@@ -23,7 +23,7 @@
     // which methods should just be skipped
     private static final List<String> skipGeneration = Arrays.asList(
             "io.deephaven.db.tables.utils.TableTools,display",
-            "io.deephaven.db.tables.select.QueryScope,addObjectFields",
+//            "io.deephaven.db.tables.select.QueryScope,addObjectFields",
```
chipkent, line 26: Just delete it if we aren't using it.

Integrations/python/deephaven/__init__.py
```diff
@@ -81,13 +81,17 @@
 from .conversion_utils import convertToJavaArray, convertToJavaList, convertToJavaArrayList, convertToJavaHashSet, \
     convertToJavaHashMap, getJavaClassObject
 
+
+class QueryScope(object):
+    pass
+
+
```
chipkent, line 88: Seems weird.  Can't we get rid of this?

Generators/Generators.gradle
chipkent: Confirm that there aren't any header files for this case that need deleting.  I don't think there are, but just making sure.

chipkent (changes requested):

Generators/Generators.gradle
jmao-denver: deleted and global search all didn't find any leftover references to QueryScope.

Integrations/python/deephaven/__init__.py
jmao-denver: deleted.

Generators/src/main/java/io/deephaven/python/PythonStaticGenerator.java
jmao-denver: done.

chipkent (approved):

PR #998, opened 2021-08-08
Plumb KafkaIngester into StreamToTableAdapter so that we use chunks for ingestion instead of tablewriters.

cpwright:


KafkaIngester/src/main/java/io/deephaven/kafka/ingest/SimpleConsumerRecordToTableWriterAdapter.java
```diff
@@ -73,50 +74,178 @@ private SimpleConsumerRecordToTableWriterAdapter(
...
-            final String kafkaPartitionColumnName,
-            final String offsetColumnName,
-            final String timestampColumnName,
-            final String keyColumnName,
-            @NotNull final String valueColumnName
-    ) {
-        return (TableWriter<?> tw) -> new SimpleConsumerRecordToTableWriterAdapter(
-                tw, kafkaPartitionColumnName, offsetColumnName, timestampColumnName, keyColumnName, valueColumnName);
-    }
+//    public static Function<TableWriter<?>, ConsumerRecordToTableWriterAdapter> makeFactory(
```
jcferretti, line 77: Assuming you will blast this commented out code?

KafkaIngester/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -299,37 +302,44 @@ public static Table consumeToTable(
...
+        streamPublisher.setChunkFactory(() -> streamToTableAdapter.makeChunksForDefinition(CHUNK_SIZE), streamToTableAdapter::chunkTypeForIndex);
+
+        final MutableInt colIdx = new MutableInt();
+        final ToIntFunction<ColumnDefinition<?>> orNull = (final ColumnDefinition<?> colDef) -> {
             if (colDef == null) {
-                return null;
+                return -1;
             }
-            return colDef.getName();
+            return colIdx.getAndAdd(1);
```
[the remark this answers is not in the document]
cpwright: I'm trying to understand your assertion here; given that we might be skipping some of the columns, so we aren't passing it though directly.  We are essentially saying "does this exist" and incrementing the counter when it exists.  So you might be asking about columns 0...4; but you would end up with 0, -1, 1, -1, 2 if you had a value, no key, and only cared about partition and timestamp; but not offset.

DB/src/main/java/io/deephaven/db/v2/StreamTableToAppendOnlyTable.java
```diff
@@ -0,0 +1,82 @@
...
+public class StreamTableToAppendOnlyTable {
+    /**
+     * Convert a Stream Table to an in-memory append only table.
+     *
+     * @param streamTable the input stream table
+     * @return an append-only in-memory table representing all data encountered in the stream
+     */
+    public static Table streamToAppendOnlyTable(Table streamTable) {
+        return QueryPerformanceRecorder.withNugget("streamToAppendOnlyTable", () -> {
+            Object streamAttribute = streamTable.getAttribute(Table.STREAM_TABLE_ATTRIBUTE);
```
rcaudy, line 28: Can change to `BaseTable.isStream()` now that it exists.

DB/src/main/java/io/deephaven/db/v2/StreamTableToAppendOnlyTable.java
```diff
@@ -0,0 +1,82 @@
...
+                columns.put(nameColumnSourceEntry.getKey(), newColumn);
+                sourceColumns[colIdx] = existingColumn;
+                destColumns[colIdx++] = newColumn;
+            }
+
+            Index index = Index.FACTORY.getEmptyIndex();
+            final QueryTable result = new QueryTable(index, columns);
+            result.setRefreshing(true);
+            result.setAttribute(Table.ADD_ONLY_TABLE_ATTRIBUTE, true);
+            result.setFlat();
```
rcaudy, line 50: We could mark the stream tables from `StreamToTableAdapter` flat...

DB/src/main/java/io/deephaven/db/v2/StreamTableToAppendOnlyTable.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    public static Table streamToAppendOnlyTable(Table streamTable) {
+        return QueryPerformanceRecorder.withNugget("streamToAppendOnlyTable", () -> {
+            Object streamAttribute = streamTable.getAttribute(Table.STREAM_TABLE_ATTRIBUTE);
+            if (!(streamAttribute instanceof Boolean) || !(Boolean) streamAttribute) {
+                throw new IllegalArgumentException("Input is not a stream table!");
+            }
+
+            final Map<String, ArrayBackedColumnSource> columns = new LinkedHashMap<>();
+            final int columnCount = streamTable.getColumnSourceMap().size();
```
rcaudy, line 34: `streamTable.getDefinition().getColumns().length` is a little better. Or hold the map so you don't do things that might have side effects or allocations twice.

DB/src/main/java/io/deephaven/db/v2/StreamTableToAppendOnlyTable.java
```diff
@@ -0,0 +1,82 @@
...
+            final ColumnSource [] sourceColumns = new ColumnSource[columnCount];
+            final ArrayBackedColumnSource [] destColumns = new ArrayBackedColumnSource[columnCount];
+            int colIdx = 0;
+            for (Map.Entry<String, ? extends ColumnSource> nameColumnSourceEntry : streamTable.getColumnSourceMap().entrySet()) {
+                final ColumnSource<?> existingColumn = nameColumnSourceEntry.getValue();
+                final ArrayBackedColumnSource<?> newColumn = ArrayBackedColumnSource.getMemoryColumnSource(0, existingColumn.getType(), existingColumn.getComponentType());
+                columns.put(nameColumnSourceEntry.getKey(), newColumn);
+                sourceColumns[colIdx] = existingColumn;
```
rcaudy, line 42: I think you should be reinterpreting these.

DB/src/main/java/io/deephaven/db/v2/StreamTableToAppendOnlyTable.java
```diff
@@ -0,0 +1,82 @@
...
+                destColumns[colIdx++] = newColumn;
+            }
+
+            Index index = Index.FACTORY.getEmptyIndex();
```
rcaudy, line 46: I was surprised that there was no initial copy here (and associated call to `ConstructSnapshot` or `checkInitiateTableOperations()`). It feels like it would be marginally more correct to capture that state, although irrelevant in the likely-common case of making one of these under the LTM lock soon after constructing a `StreamToTableAdapter`.

DB/src/main/java/io/deephaven/db/v2/StreamTableToAppendOnlyTable.java
```diff
@@ -0,0 +1,82 @@
...
+ * Note, this table will grow without bound as new stream values are encountered.
+ */
+public class StreamTableToAppendOnlyTable {
+    /**
+     * Convert a Stream Table to an in-memory append only table.
+     *
+     * @param streamTable the input stream table
+     * @return an append-only in-memory table representing all data encountered in the stream
+     */
+    public static Table streamToAppendOnlyTable(Table streamTable) {
```
rcaudy, line 26: We should give this a better name/and or path to use it.
Maybe make a `StreamTables` interface that offers this and other helpers? Or maybe put an instance method on `StreamToTableAdapter`?

DB/src/main/java/io/deephaven/db/v2/StreamTableToAppendOnlyTable.java
```diff
@@ -0,0 +1,82 @@
...
+                        throw new IllegalArgumentException("Stream tables should not modify or shift!");
+                    }
+                    final long newRows = upstream.added.size();
+                    if (newRows == 0) {
+                        return;
+                    }
+                    final long currentSize = index.size();
+                    columns.values().forEach(c -> c.ensureCapacity(currentSize + newRows));
+
+                    final Index newRange = Index.FACTORY.getIndexByRange(currentSize, currentSize + newRows - 1);
```
rcaudy, line 65: I think you can use `CURRENT_FACTORY` here, and for modified and removed as well.

DB/src/main/java/io/deephaven/db/v2/StreamTableToAppendOnlyTable.java
```diff
@@ -0,0 +1,82 @@
...
+
+                    ChunkUtils.copyData(sourceColumns, upstream.added, destColumns, newRange, false);
+                    index.insertRange(currentSize, currentSize + newRows - 1);
+
+                    final ShiftAwareListener.Update downstream = new ShiftAwareListener.Update();
+                    downstream.shifted = IndexShiftData.EMPTY;
```
rcaudy, line 71: What kind of crazy person thinks of these fields in this order?

rcaudy (commented):
This is a review of the query engine side of things only. Note a mix of cosmetic comments and legitimate concerns.

KafkaIngester/KafkaIngester.gradle
```diff
@@ -42,3 +42,36 @@ dependencies {
...
+def createDeephavenTestExecTask = {
+    taskName, mainClass -> tasks.create(taskName, JavaExec, { JavaExec task ->
+        new File("$rootDir/tmp/workspace").mkdirs()
+        new File("$rootDir/tmp/logs").mkdirs()
+
+        task.workingDir "$rootDir"
+        task.classpath = sourceSets.test.runtimeClasspath
+        task.main = mainClass
+
+        task.jvmArgs '-DConfiguration.rootFile=dh-tests.prop',
```
rcaudy, line 55: This whole block is indented weird.

It also has some boilerplate that should maybe come out:
```
                    '-Dlog4j.configuration=log4j.teamcity.xml',
                    '-Ddisable.pdsport.rotate=true',
                    '-Djava.awt.headless=true'
```

Then again, I see that you copied this from elsewhere.

KafkaIngester/KafkaIngester.gradle
```diff
@@ -42,3 +42,36 @@ dependencies {
...
+def createMainExecTask = {
+    taskName, mainClass -> tasks.create(taskName, JavaExec, { JavaExec task ->
+        task.workingDir "$rootDir"
+        task.classpath = sourceSets.main.runtimeClasspath
+        task.main = mainClass
+        return
+    })
+}
+
+createDeephavenTestExecTask('replicateKafka', 'io.deephaven.kafka.ingest.ReplicateKafka')
```
rcaudy, line 77: Why is this a `TestExecTask` instead of a `MainExecTask`?

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/BooleanFieldCopier.java
```diff
@@ -0,0 +1,29 @@
...
+        this.fieldName = fieldName;
+    }
+
+    @Override
+    public void copyField(ObjectChunk<Object, Attributes.Values> inputChunk, WritableChunk<Attributes.Values> publisherChunk) {
+        final WritableBooleanChunk<Attributes.Values> output = publisherChunk.asWritableBooleanChunk();
+        for (int ii = 0; ii < inputChunk.size(); ++ii) {
+            final GenericRecord genericRecord =  (GenericRecord)inputChunk.get(ii);
+            final Boolean value = (Boolean)genericRecord.get(fieldName);
+            output.set(ii, value == null ? QueryConstants.NULL_BOOLEAN : value);
```
rcaudy, line 26: This is buggy and unused. I would delete it and not generate it,

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/CharFieldCopier.java
```diff
@@ -0,0 +1,26 @@
...
+        this.fieldName = fieldName;
+    }
+
+    @Override
+    public void copyField(ObjectChunk<Object, Attributes.Values> inputChunk, WritableChunk<Attributes.Values> publisherChunk) {
+        final WritableCharChunk<Attributes.Values> output = publisherChunk.asWritableCharChunk();
+        for (int ii = 0; ii < inputChunk.size(); ++ii) {
+            final GenericRecord genericRecord =  (GenericRecord)inputChunk.get(ii);
+            final Character value = (Character)genericRecord.get(fieldName);
+            output.set(ii, value == null ? QueryConstants.NULL_CHAR : value);
```
rcaudy, line 23: I was suspicious that you were doing a 0-based fill but weren't setting size. The only use I see is this, from `io.deephaven.kafka.ingest.GenericRecordChunkAdapter#handleChunk`:
```
                for (int cc = 0; cc < chunkOffsets.length; ++cc) {
                    final WritableChunk<Attributes.Values> publisherChunk = publisherChunks[chunkOffsets[cc]];
                    final int existingSize = publisherChunk.size();
                    publisherChunk.setSize(existingSize + inputChunk.size());
                    fieldCopiers[cc].copyField(inputChunk, publisherChunk);
                }
```
So the calling pattern looks like an append, but you're using `set` with a 0-based index instead of `add` or similar. Probably a bug...
If you change the field copiers, don't forget to do the non-replicated ones (String and Object).

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/FieldCopier.java
```diff
@@ -0,0 +1,9 @@
+package io.deephaven.kafka.ingest;
+
+import io.deephaven.db.v2.sources.chunk.Attributes;
+import io.deephaven.db.v2.sources.chunk.ObjectChunk;
+import io.deephaven.db.v2.sources.chunk.WritableChunk;
+
+public interface FieldCopier {
+    void copyField(ObjectChunk<Object, Attributes.Values> inputChunk, WritableChunk<Attributes.Values> publisherChunk);
```
rcaudy, line 8: Needs JavaDoc.

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -84,6 +96,21 @@ public StreamToTableAdapter(@NotNull final TableDefinition tableDefinition,
         }
     }
 
+    @NotNull
+    private static WritableChunk<?> makeChunk(io.deephaven.db.tables.ColumnDefinition<?> cd, int size) {
```
rcaudy, line 100: Any particular reason for the fully-qualified class name?

KafkaIngester/src/main/java/io/deephaven/kafka/StreamPublisherImpl.java
```diff
@@ -0,0 +1,49 @@
...
+import io.deephaven.stream.StreamPublisher;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.function.IntFunction;
+import java.util.function.Supplier;
+
+public class StreamPublisherImpl implements StreamPublisher {
+    private StreamConsumer streamConsumer;
+    private WritableChunk[] chunks;
+    private java.util.function.Supplier<WritableChunk[]> chunkFactory;
```
rcaudy, line 15: Just `Supplier` will do...

KafkaIngester/src/main/java/io/deephaven/kafka/StreamPublisherImpl.java
```diff
@@ -0,0 +1,49 @@
...
+    @Override
+    public void register(@NotNull StreamConsumer consumer) {
+        if (streamConsumer != null) {
+            throw new IllegalStateException("Can not register multiple StreamConsumers.");
+        }
+        streamConsumer = consumer;
+    }
+
+    @Override
+    public void flush() {
```
rcaudy, line 43: This almost certainly needs to be `synchronized`, as does whatever fills the chunks.

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/KeyOrValueProcessor.java
```diff
@@ -0,0 +1,17 @@
+package io.deephaven.kafka.ingest;
+
+import io.deephaven.db.v2.sources.chunk.Attributes;
+import io.deephaven.db.v2.sources.chunk.ObjectChunk;
+import io.deephaven.db.v2.sources.chunk.WritableChunk;
+
+public interface KeyOrValueProcessor {
+    /**
+     * After consuming a set of generic records for a batch that are not raw objects, we pass the keys or values to
+     * an appropriate handler.  The handler must know it's data types and offsets within the publisher chunks, and
```
rcaudy, line 10: it's -> its

rcaudy (commented):
Partial review of the KafkaIngestor side of things.

DB/src/main/java/io/deephaven/db/v2/utils/ChunkUtils.java
```diff
@@ -574,6 +576,68 @@ public static void copyData(ChunkSource.WithPrev<? extends Attributes.Values> sr
...
+                        throw new IllegalArgumentException("Source must not equal destination!");
+                    }
+                }
+                destinations[ss].ensureCapacity(destAllKeys.lastKey() + 1);
+                sourceContexts[ss] = sources[ss].makeGetContext(minSize);
+                destContexts[ss] = destinations[ss].makeFillFromContext(minSize);
+            }
+
+            while (srcIter.hasMore()) {
+                assert(destIter.hasMore());
```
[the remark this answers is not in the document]
cpwright: No, just copy and paste.

DB/src/main/java/io/deephaven/db/v2/StreamTableToAppendOnlyTable.java
cpwright: I am pretty sure that you run into issues when you notify with current indices instead of full up indices.

DB/src/main/java/io/deephaven/db/v2/StreamTableToAppendOnlyTable.java
cpwright: The kind that puts the empty crap that is needed for initialization first, and then the one thing you actually care about.

DB/src/main/java/io/deephaven/db/v2/StreamTableToAppendOnlyTable.java
rcaudy: `StreamTools.toAppendOnlyTable`.

rcaudy: `StreamTableTools.streamToAppendOnlyTable` because Python? Eww

rcaudy (commented):
Incremental review on changes to already reviewed files.

KafkaIngester/src/main/java/io/deephaven/kafka/StreamPublisherImpl.java
```diff
@@ -12,7 +12,7 @@
 public class StreamPublisherImpl implements StreamPublisher {
     private StreamConsumer streamConsumer;
```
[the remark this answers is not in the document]
cpwright: You can only set them once at initialization.

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/PojoConsumerRecordToStreamPublisherAdapter.java
```diff
@@ -603,7 +603,7 @@ public void consumeRecords(List<? extends ConsumerRecord<?, ?>> records) throws
         }
     }
 
-    public void consumeRecord(ConsumerRecord<?, ?> record) throws IOException {
+    private void consumeRecord(ConsumerRecord<?, ?> record) throws IOException {
```
rcaudy, line 606: Do this in JSON, too?

KafkaIngester/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -252,74 +252,96 @@ public static Table consumeToTable(
...
 
         if (!kafkaConsumerProperties.containsKey(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG)) {
             if (keySchema != null) {
                 kafkaConsumerProperties.setProperty(
                         ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, KafkaAvroDeserializer.class.getName());
+                columnDefinitions.addAll(Arrays.asList(keyColumns));
             } else {
                 kafkaConsumerProperties.setProperty(
                         ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, STRING_DESERIALIZER);
+//                columnDefinitions.add(ColumnDefinition.ofString("Key"));
```
rcaudy, line 272: What's with the commented-out definitions?

rcaudy (commented):
All done. Mostly just the TODOs/comments to discuss.

rcaudy (approved):

PR #1029, opened 2021-08-11
Support a delegating getChunk implementation on SwitchColumnSource, and associated cleanups

### LINKED ISSUE #1024: Query Engine: SwitchColumnSource should implement Get by Delegation instead of Filling
As a developer, I want SwitchColumnSource to delegate getChunk to the inner source so that we do not need to make copies of our data unnecessarily.

rcaudy:
Closed #1024

### DIFF
--- a/DB/src/main/java/io/deephaven/db/v2/sources/SwitchColumnSource.java
+++ b/DB/src/main/java/io/deephaven/db/v2/sources/SwitchColumnSource.java
@@ -1,22 +1,24 @@
 package io.deephaven.db.v2.sources;
 
 import io.deephaven.base.verify.Assert;
-import io.deephaven.db.v2.sources.chunk.Attributes;
-import io.deephaven.db.v2.sources.chunk.SharedContext;
-import io.deephaven.db.v2.sources.chunk.WritableChunk;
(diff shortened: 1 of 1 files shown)


DB/src/main/java/io/deephaven/db/v2/sources/SwitchColumnSource.java
```diff
@@ -1,213 +1,275 @@
...
+                    ? prevContext = makeContext(prevSource)
+                    : prevContext;
         }
 
         @Override
         public void close() {
-            currentContext.close();
-            if (prevContext != null) {
-                prevContext.close();
+            //noinspection EmptyTryBlock
```
cpwright, line 88: This seems like a very round about way of closing things and null checking.

cpwright (approved):

DB/src/main/java/io/deephaven/db/v2/sources/SwitchColumnSource.java
rcaudy: As discussed, it's just better in every way except clarity to the uninitiated.

PR #1030, opened 2021-08-11
TicketRouter: throw StatusRuntimeException instead of NPE for missing ticket

nbauernfeind:


grpc-api/src/main/java/io/deephaven/grpc_api/session/TicketRouter.java
```diff
@@ -74,7 +74,11 @@ public TicketRouter(final Set<TicketResolver> resolvers) {
     public <T> SessionState.ExportObject<T> resolve(
             @Nullable final SessionState session,
             final Ticket ticket) {
-        return resolve(session, ticket.getTicket().asReadOnlyByteBuffer());
+        final ByteBuffer bb = ticket.getTicket().asReadOnlyByteBuffer();
```
rcaudy, line 77: Can you clarify what can be null here? `Ticket`? `Ticket.getTicket()`? Just the byte buffer?

rcaudy: And is this different from the call to `resolve` on a `io.deephaven.proto.backplane.grpc.Ticket`?

nbauernfeind: No this is the same call. The ticket is a null byte buffer if the client did not send it. I found a jsapi call that isn't sending one and it just NPEs instead of giving a more useful error.

rcaudy: I think you probably are wrong about which call in the chain is causing the NPE.

nbauernfeind:
discarding; it was indeed different -- Colin and I merged fixes for as many of these as we found in the now merged app_metadata changeset

PR #1034, opened 2021-08-11
Downgrade Alpine to avoid a DNS issue on older mac docker installs

niloc132:
Apparently in older linux kernels or docker installs (the alpine bug report isn't sure which), DNS sometimes doesn't work, and we can't build this image since github.com fails to resolve in the second line of this `RUN`.

In Alpine 3.12, unzip doesn't create the directory automatically, so we need to make it ahead of time.

Alpine 3.12 EOLs May 1, 2022, we should update by then, decide that we aren't concerned with older kernels (or docker installs), or pick a different image for this proxy.

https://github.com/alpinelinux/docker-alpine/issues/155

cc @hythloda

### DIFF
--- a/grpc-proxy/Dockerfile
+++ b/grpc-proxy/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.io/library/alpine:3.13
+FROM docker.io/library/alpine:3.12
 
 COPY contents/ .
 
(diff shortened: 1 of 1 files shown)


devinrsmith (changes requested):
Can you create a ticket and add a comment linking to it in the Dockerfile?

niloc132:
Issue created and referenced from Dockerfile

devinrsmith (approved):

PR #1042, opened 2021-08-11
JSAPI: wire up selectDistinct; fixes the select values box in advanced filters

nbauernfeind:
This gets this widget working again:
Screen Shot 2021-08-11 at 4 22 17 PM

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
@@ -537,15 +537,14 @@ public class JsTable extends HasEventHandling implements HasTableBinding, HasLif
 
     @JsMethod
     public Promise<JsTable> selectDistinct(Column[] columns) {
+        final ClientTableState state = state();
         // We are going to forget all configuration for the current state
         // by just creating a new, fresh state.  This should be an optional flatten()/copy() step instead.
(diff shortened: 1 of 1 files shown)


web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
```diff
@@ -538,12 +537,17 @@ public void internalSubscribe(JsArray<Column> columns, TableSubscription sub) {
...
         // We are going to forget all configuration for the current state
         // by just creating a new, fresh state.  This should be an optional flatten()/copy() step instead.
         String[] columnNames = Arrays.stream(columns).map(Column::getName).toArray(String[]::new);
-        final ClientTableState distinct = workerConnection.newState((c, cts, metadata) -> {
-                    workerConnection.tableServiceClient().selectDistinct(new SelectDistinctRequest(), metadata, c::apply);
+
+
+        final ClientTableState distinct = workerConnection.newState((c, state, metadata) -> {
+                    SelectDistinctRequest request = new SelectDistinctRequest();
+                    request.setSourceId(state().getHandle().makeTableReference());
```
niloc132, line 547: please move state() back to the top of the method - we want to record the current state at the moment that this was invoked, rather than risk this coming back later with a different state (such as if the user called applySort on the table right after selectDistinct - not 100% that this would cause it, but we do want to be sure to get the right instance.)

niloc132 (approved):

PR #1043, opened 2021-08-11
Json support on top of stream adapters, plus refactorings towards reuse facilitating that.

jcferretti:


DB/src/main/java/io/deephaven/db/tables/utils/TableTools.java
```diff
@@ -1644,4 +1645,48 @@ public static String nullTypeAsString(final Class<?> dataType) {
...
+        }
+        return dataType;
+    }
+
+    public static ColumnDefinition<?> colDef(
+            final String columnName, final String dataTypeStr) {
+        return ColumnDefinition.fromGenericType(columnName, typeFromName(dataTypeStr));
+    }
+
+    public static ColumnDefinition<?> colDef(
```
cpwright, line 1688: This feels somehow duplicative with io.deephaven.db.tables.ColumnDefinition#fromGenericType(java.lang.String, java.lang.Class<T>); or like we are not doing it in the same coherent way as Devin's stuff.

KafkaIngester/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -339,6 +342,88 @@ public static Table consumeToTable(
...
+            @NotNull final String topic,
+            @NotNull final IntPredicate partitionFilter,
+            @NotNull final IntToLongFunction partitionToInitialOffset,
+            final ColumnDefinition<?>[] valueColumns,
+            final Map<String, String> columnNameToJsonField) {
+        final int nCols = 3 + valueColumns.length;
+        final ColumnDefinition<?>[] commonColumns = new ColumnDefinition[nCols];
+        int c = 0;
+        c += getCommonCols(commonColumns, 0, consumerProperties, partitionFilter == ALL_PARTITIONS);
+        final int[] commonColsIndices = new int[3];
```
cpwright, line 365: We are repeating this thing over and over; in the next iteration we should refactor this to be common.

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/JsonNodeObjectFieldCopier.java
```diff
@@ -0,0 +1,24 @@
...
+    public JsonNodeObjectFieldCopier(String fieldName) {
+        this.fieldName = fieldName;
+    }
+
+    @Override
+    public void copyField(ObjectChunk<Object, Attributes.Values> inputChunk, WritableChunk<Attributes.Values> publisherChunk, int sourceOffset, int destOffset, int length) {
+        final WritableObjectChunk<Object, Attributes.Values> output = publisherChunk.asWritableObjectChunk();
+        for (int ii = 0; ii < length; ++ii) {
+            final JsonNode node = (JsonNode) inputChunk.get(ii + sourceOffset);
+            output.set(ii + destOffset, JsonNodeUtil.getValue(node, fieldName, true, true));
```
cpwright, line 21: What kind of objects do we expect here?

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/ReplicateKafka.java
```diff
@@ -6,6 +6,6 @@
 
 public class ReplicateKafka {
     public static void main(String [] args) throws IOException {
-        ReplicatePrimitiveCode.charToAllButBoolean(CharFieldCopier.class, ReplicatePrimitiveCode.MAIN_SRC);
+        ReplicatePrimitiveCode.charToAllButBoolean(GenericRecordCharFieldCopier.class, ReplicatePrimitiveCode.MAIN_SRC);
```
cpwright, line 9: Why isn't your JSON version here?

cpwright (dismissed):

KafkaIngester/src/main/java/io/deephaven/kafka/KafkaTools.java
jcferretti: That is the idea, my next step is to refactor all the 3 calls into one, with arguments that are explicit avro versus json versus simple "specs".

DB/src/main/java/io/deephaven/db/tables/utils/TableTools.java
jcferretti: Is Devin's stuff usable right now?  Last time I asked (which was Friday) the answer was no.
Strictly speaking, for this diff we only need a way to access in a not-too-convoluted way ColumnDefinition "factories" from the python implementation of consumeToTable; yesterday I was actually thinking that it would be better to do an argument like
```
[ ('Symbol', 'string'), ('Price', 'double') ]
```
instead of
```
[ colDef('Symbol', 'string'), colDef('Price', 'double') ]
```
ie, expect python tuples in the argument, and apply colDef or something like it under the hood, instead of expect the user to know column definition objects/call colDef.

I'd like to do that in the next pass tho, if we agree that's a good way.

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/ReplicateKafka.java
jcferretti: Because is wrong... :-)  Fixing...

KafkaIngester/src/main/java/io/deephaven/kafka/ingest/JsonNodeObjectFieldCopier.java
jcferretti: We discussed over slack, we are deleting.

cpwright (approved):

PR #1058, opened 2021-08-12
add compose files for standalone docker containers

mikeblas:


containers/groovy-examples/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+    image: ghcr.io/deephaven/grpc-api:${VERSION:-latest}
+    expose:
+      - '8080'
+    volumes:
+      - ./docker/core/data:/data
+      - api-cache:/cache
+    environment:
+      - JAVA_TOOL_OPTIONS=-Xmx4g -Dio.deephaven.console.type=groovy
+
+
```
chipkent, line 14:
```suggestion
```
Inconsistent spacing

containers/groovy-examples/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
+version: "3.4"
+
+services:
+  grpc-api:
+    image: ghcr.io/deephaven/grpc-api:${VERSION:-latest}
+    expose:
+      - '8080'
+    volumes:
+      - ./docker/core/data:/data
```
chipkent, line 9: The `./docker/core/data` path seems funny for this use case.  Kind of Rube Goldberg.  Should we be using `./data` instead?

containers/groovy-examples/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+      - web
+      - grpc-proxy
+      - grpc-api
+    ports:
+      - "${PORT:-10000}:10000"
+
+  examples:
+    image: ghcr.io/deephaven/examples
+    volumes:
+      - ./docker/core/data:/data
```
chipkent, line 44: The `./docker/core/data` path seems funny for this use case.  Kind of Rube Goldberg.  Should we be using `./data` instead?

containers/groovy/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
+version: "3.4"
+
+services:
+  grpc-api:
+    image: ghcr.io/deephaven/grpc-api:${VERSION:-latest}
+    expose:
+      - '8080'
+    volumes:
+      - ./docker/core/data:/data
```
chipkent, line 9: The `./docker/core/data` path seems funny for this use case.  Kind of Rube Goldberg.  Should we be using `./data` instead?

containers/groovy/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+      - api-cache:/cache
+    environment:
+      - JAVA_TOOL_OPTIONS=-Xmx4g -Dio.deephaven.console.type=groovy
+
+
```
chipkent, line 14:
```suggestion
```

containers/python-examples/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
+version: "3.4"
+
+services:
+  grpc-api:
+    image: ghcr.io/deephaven/grpc-api:${VERSION:-latest}
+    expose:
+      - '8080'
+    volumes:
+      - ./docker/core/data:/data
```
chipkent, line 9: The `./docker/core/data` path seems funny for this use case.  Kind of Rube Goldberg.  Should we be using `./data` instead?

containers/python-examples/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+      - web
+      - grpc-proxy
+      - grpc-api
+    ports:
+      - "${PORT:-10000}:10000"
+
+  examples:
+    image: ghcr.io/deephaven/examples
+    volumes:
+      - ./docker/core/data:/data
```
chipkent, line 43: The `./docker/core/data` path seems funny for this use case.  Kind of Rube Goldberg.  Should we be using `./data` instead?

containers/python/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
+version: "3.4"
+
+services:
+  grpc-api:
+    image: ghcr.io/deephaven/grpc-api:${VERSION:-latest}
+    expose:
+      - '8080'
+    volumes:
+      - ./docker/core/data:/data
```
chipkent, line 9: The `./docker/core/data` path seems funny for this use case.  Kind of Rube Goldberg.  Should we be using `./data` instead?

containers/python/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+      - api-cache:/cache
+    environment:
+      - JAVA_TOOL_OPTIONS=-Xmx4g -Dio.deephaven.console.type=python
+
+
```
chipkent, line 14:
```suggestion
```

containers/python-examples/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+services:
+  grpc-api:
+    image: ghcr.io/deephaven/grpc-api:${VERSION:-latest}
+    expose:
+      - '8080'
+    volumes:
+      - ./docker/core/data:/data
+      - api-cache:/cache
+    environment:
+      - IO_DEEPHAVEN_CONSOLE_TYPE=python
```
chipkent, line 12: This is not present in `python`.  It raises questions about what is different in the files and why.  Have you done diffs of the files to look for differences?

containers/groovy-examples/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
+version: "3.4"
```
chipkent, line 1: @devinrsmith I see a bunch of repetition and some inconsistency here.  Do we want to keep things in single files, as they are, or use something like https://stackoverflow.com/questions/33408118/inheritance-or-nesting-with-docker-compose or https://serversforhackers.com/dockerized-app/compose-separated?  My concerns are DRY and inconsistencies.  I can see that avoiding these problems may increase the problems with users deploying, so the changes might not be worth it.

containers/groovy-examples/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+
```
chipkent, line 2: @mikeblas What is your plan for the README?  A second PR?

chipkent (changes requested):

containers/groovy-examples/docker-compose.yml
mikeblas: Since we want the files to be downloadable, do we have a choice? Inheritance would mean having the user download multiple files, not just one.

containers/groovy-examples/docker-compose.yml
mikeblas: A second PR, yes. I have something written, but it isn't ready until we settle on the naming (including path) for these files. And I need to get the authoring tools installed and working before I can write it.

containers/python-examples/docker-compose.yml
mikeblas: Fixed!

containers/groovy-examples/docker-compose.yml
mikeblas: see https://github.com/deephaven/deephaven.io/pull/491

containers/python-examples/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+  envoy:
+    image: ghcr.io/deephaven/envoy:${VERSION:-latest}
+    depends_on:
+      - web
+      - grpc-proxy
+      - grpc-api
+    ports:
+      - "${PORT:-10000}:10000"
+
+  examples:
```
chipkent, line 40: Does running examples without any arguments cause the data to be downloaded / updated?

containers/groovy-examples/docker-compose.yml
chipkent: I"m not against single files if we can keep them consistent.

containers/groovy-examples/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+  envoy:
+    image: ghcr.io/deephaven/envoy:${VERSION:-latest}
+    depends_on:
+      - web
+      - grpc-proxy
+      - grpc-api
+    ports:
+      - "${PORT:-10000}:10000"
+
+  examples:
```
chipkent, line 40: Does running examples without any arguments cause the data to be downloaded / updated?

chipkent (changes requested):

containers/python-examples/docker-compose.yml
mikeblas: No. As with the standalone verison, it prints the help message that gives the commands and their arguments.

containers/groovy-examples/docker-compose.yml
mikeblas: No. It will print the usage message for the download script, as it did before. This hasn't changed.

chipkent (dismissed):

chipkent (dismissed):

containers/groovy-examples/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+    image: ghcr.io/deephaven/examples
+    volumes:
+      - ./docker/core/data:/data
+
+volumes:
+    data:
```
devinrsmith, line 46: We shouldn't be creating a proper docker volume in these 4 cases since we are doing it as a bind mount.

devinrsmith (changes requested):

containers/groovy-examples/docker-compose.yml
mikeblas: Fixed!

docker/README.md
```diff
@@ -3,8 +3,3 @@
 This `docker/` directory serves the local docker development process by being a convenient place
 for [bind mounts](https://docs.docker.com/storage/bind-mounts/). Developers may structure data
 however they see fit in any subdirectories.
-
-By convention though, the main [docker-compose.yml](../docker-compose.yml) file will use the
-`docker/core/data/` directory as the bind mount for the `/data` volume and the
-`docker/core/cache/` directory as the bind mount for the `/cache` volume with respect to the
-`grpc-api` service.
```
chipkent, line 10: Do we need to indicate where `/data` is bind mounted?  We deleted the comments related to this.

chipkent (changes requested):

docker/README.md
mikeblas: We could, but it seems awkward to describe the `./data` mount in the apparently unrelated `./docker/README.md` file. If we do write something, where should it live?

mikeblas: Maybe this is what we want?

```
By convention though, the main[docker-compose.yml](../docker-compose.yml) file will use the `./data` directory as the bind mount for the `/data` volume and the `./docker/core/cache/` directory as the bind mount for the `/cache` volume with respect to the grpc-api service.
```

devinrsmith (approved):

hythloda (approved):
Love it! We need a redpanda version.  But also don't want to hold this up to get that.

docker/README.md
mikeblas: Per chat, I've used the above text.

PR #1059, opened 2021-08-12
Propagate errors from Kafka to tables.

cpwright:
I tested this using the following code snippet:
```
# New System.err configured
from deephaven.KafkaTools import *
t = consumeToTable({'bootstrap.servers' : 'redpanda:29092'}, 'orders', value_json=[ ('Symbol', 'string'), ('Side', 'string'), ('Price', 'double'), ('Qty', 'int') ], table_type="append")
PrintListenerType = jpy.get_type("io.deephaven.db.util.PrintListener")
pl = PrintListenerType("Streaming Append Table", t)
```

And then sent the following valid JSON:
```
python3 kafka-produce.py orders 'str:{ "Symbol" : "MSFT", "Side" : "BUY", "Price" : "278.86", "Qty"= "501" }'
```
Which adds a row to our table and prints it from the listener.

And then the following invalid JSON:
```
python3 kafka-produce.py orders 'str:{ "Symbol" : "MSFT", "Side" : "BUY", "Price" : "278.86", "Qty"= "501" }'
```
Which does not add a row to our table, and the listener receives and prints the failure.

The UI does not surface failed tables properly; which is why I tested this way.


DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -222,6 +231,13 @@ public void refresh() {
     }
 
     private void doRefresh() {
+        synchronized (this) {
+            // if we have an enqueued failure we want to process it first, before we allow the streamPublisher to flush itself
+            if (enqueuedFailure != null) {
+                throw new UncheckedDeephavenException(MultiException.maybeWrapInMultiException("Multiple errors encountered while ingesting stream", enqueuedFailure.toArray(new Exception[0])));
```
jcferretti, line 237: This is a pretty long line.  And github wrapping is nil for review.

jcferretti (approved):

jcferretti (approved):

PR #1064, opened 2021-08-13
New KafkaTools.consumeToTable general API.

jcferretti:
Preliminary draft, to elicit comments about the Spec objects.

The intended usage is:

```
Table t = KafkaTools.consumeToTable(
    kafkaConsumerProperties,
    "myTopic",
    KafkaTools.ALL_PARTITIONS,
    KafkaTools.ALL_PARTITIONS_DONT_SEEK,
    // up to this point it has looked the same, what follow is different
    KafkaTools.simpleSpec("Date"), // spec for key, simple one value; in this case, type from properties.
    KafkaTools.avroSpec(avroSchema, Function.identity())  // spec for value.
);
```


KafkaIngester/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -195,20 +203,274 @@ else if (unionType0 == Schema.Type.NULL) {
...
+    }
+
+    public static void avroSchemaToColumnDefinitions(
+            final List<ColumnDefinition> columns,
+            final Schema schema
+    ) {
+        avroSchemaToColumnDefinitions(columns, schema, DIRECT_MAPPING);
+    }
+
+    public enum KeyOrValue {
```
jcferretti, line 227: Interesting new API part starts here.

KafkaIngester/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -604,14 +852,36 @@ private static int getCommonCols(
...
+        final String typeProperty;
+        final String deserializerProperty;
+        final String nameProperty;
+        final String nameDefault;
+        switch(keyOrValue) {
+            case KEY:
+                typeProperty = KEY_COLUMN_TYPE_PROPERTY;
+                deserializerProperty = ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG;
+                nameProperty = KEY_COLUMN_NAME_PROPERTY;
+                nameDefault =KEY_COLUMN_NAME_DEFAULT;
```
cpwright, line 869: spacing after equal

KafkaIngester/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -195,20 +203,274 @@ else if (unionType0 == Schema.Type.NULL) {
...
+
+        final KeyOrValueProcessor keyProcessor = getProcessor(keySpec, tableDefinition, streamToTableAdapter, keyIngestData);
+        final KeyOrValueProcessor valueProcessor = getProcessor(valueSpec, tableDefinition, streamToTableAdapter, valueIngestData);
+
+        final ConsumerRecordToStreamPublisherAdapter adapter = KafkaStreamPublisher.make(streamPublisher,
+                commonColumnIndices[0],
+                commonColumnIndices[1],
+                commonColumnIndices[2],
+                keyProcessor, valueProcessor,
+                Function.identity(),
```
cpwright, line 360: should the transformation be part of the KeyIngestData, because doesn't JSON need to do the parsing here?

cpwright (commented):
I need to leave a comment so that I am up-to-date and can review from after this merge resolution.

Integrations/python/deephaven/KafkaTools/__init__.py
```diff
@@ -80,72 +89,163 @@ def _custom_avroSchemaToColumnDefinitions(*args):
...
+
+
+@_passThrough
+def avro(*args, **kwargs):
+    if len(args) < 1 and len(args) > 2:
+        raise Exception("one or two positional arguments expected, instead got " + len(args))
+    if _isStr(args[0]):
+        have_actual_schema = False
+        schema_name = args[0]
+        if len(args) < 2 or not _isStr(args[1]):
```
cpwright, line 164: That is interesting; I would have assumed there was a default.

Integrations/python/deephaven/KafkaTools/__init__.py
```diff
@@ -80,72 +89,163 @@ def _custom_avroSchemaToColumnDefinitions(*args):
...
+        maybe_dict_arg = 2
+    else:
+        have_actual_schema = True
+        schema = args[0]
+        maybe_dict_arg = 1
+    mapping = kwargs.pop('mapping', None)
+    mapping_only = kwargs.pop('mapping_only', None)
+    if mapping is not None and mapping_only is not None:
+        raise Exception(
+            "only one keyword argument betwee 'mapping' and " +
```
cpwright, line 178: betwee -> between

Integrations/python/deephaven/KafkaTools/__init__.py
```diff
@@ -198,6 +314,31 @@ def getAvroSchema(schemaServerUrl, resourceName, version):
...
+def ignoreSpec():
+    """
+    :return: io.deephaven.kafka.KafkaTools.KeyOrValueSpec
+    """
+    
+    return _java_type_.ignoreSpec()
+
+
+@_passThrough
+def jsonSpec(*args):
```
cpwright, line 327: When do we expect these to be used instead of the friendly ones above?

Integrations/src/main/java/io/deephaven/integrations/python/PythonTools.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /**
+     * Convert a map to a function, mapping any nonexistent keys in the map to themselves.
+     *
+     * @param map           the map to convert to function; the caller gives up ownership to the returned object
+     * @return              the resulting function
+     */
+    @SuppressWarnings("unused")
+    public static Function<String, String> functionStrStrFromMap(final Map<String, String> map
```
cpwright, line 32: I think the overload is not quite clear ; the identity behavior is not exactly normal; so I would add Identity or something to the function name to make it less ambiguous.

cpwright (commented):
If I am reading this right, we still always call record.value() from the KafkaIngester; even if we have an IGNORE spec.  I think we should fix that and eliminate the need for a value chunk in that case.

Integrations/python/deephaven/KafkaTools/__init__.py
```diff
@@ -80,72 +89,163 @@ def _custom_avroSchemaToColumnDefinitions(*args):
...
-        streaming_table = _java_type_.consumeToTable(r[0], r[1], r[2], r[3], key_avro_schema, mapping, value_avro_schema, mapping)
-    elif value_json is not None:
-        if not isinstance(value_json, list):
-            raise Exception('value_json keyword argument should be a list.')
-        col_def_list = _tuplesListToColDefsList(value_json)
-        streaming_table = _java_type_.consumeJsonToTable(r[0], r[1], r[2], r[3], col_def_list)
-    else:
-        raise Exception('Unknown keyword arguments: ' + str(kwargs))
     table_type = kwargs.pop('table_type', None)
     if table_type is None or table_type == 'append':
```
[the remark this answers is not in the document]
jcferretti: Done.

Integrations/python/deephaven/KafkaTools/__init__.py
jcferretti: Since is a string I am kinda hesitant to guess "1", I've seen examples of things like '0.1', '1.1', '1', is kinda all over the place I think

Integrations/python/deephaven/KafkaTools/__init__.py
jcferretti: They shouldn't, I am missing listing them for ignoring on PythonStaticGenerator, thanks.

Integrations/src/main/java/io/deephaven/integrations/python/PythonTools.java
jcferretti: Done.

cpwright (dismissed):

cpwright (approved):

PR #1067, opened 2021-08-13
Fix bugs and missing files preventing Java2NumpyCopyTests from compiling.

alexpeters1208:
Java2NumpyTest now compiles, runs, and passes. Had to add a few images to the right test directory, make some small changes to .build in integrations, and modify the test file slightly.


Integrations/build.gradle
```diff
@@ -35,9 +35,7 @@ sourceSets {
         resources.srcDirs = ['src/main/resources']
     }
     test {
-        java.srcDirs = []
-        // TODO: test doesn't compile
-        //java.srcDirs = ['src/test/java']
+        java.srcDirs = ['src/test/java']
         resources.srcDirs = ['test_python', 'src/test/resources']
```
chipkent, line 39: Have you confirmed that the tests run without the one extra line from Colin?

Integrations/src/main/java/io/deephaven/integrations/numpy/Java2NumpyCopy.java
```diff
@@ -230,23 +230,23 @@ public static void copySlice(final Table t, final long rowStart, final long[] da
...
+     * @param rowStart first row of data to copy
+     * @param data     array to copy data into
+     * @param nRow     number of rows to copy; also the number of rows in <code>data</code>.
+     * @param nCol     number of table columns; also the number of columns in <code>data</code>.
+     */
+    public static void copySlice(final Table t, final long rowStart, final boolean[] data, final int nRow, final int nCol) {
+        assertCopySliceArgs(t, rowStart, data == null ? -1 : data.length, nRow, nCol);
+        assert data != null;
+        copySlice(t, rowStart, nRow, nCol, boolean.class, "(boolean)", (cs, k, idx) -> data[idx] = cs.getBoolean(k));
+    }
```
chipkent, line 247: This comment seems to indicate that these constructors were intentionally removed so that byte[] is used.  Does the boolean cast throw an error?  Are you certain this code is executed?

Integrations/src/main/java/io/deephaven/integrations/numpy/Java2NumpyCopy.java
```diff
@@ -497,23 +497,21 @@ public static void copyRand(final Table t, final long[] data, final int nRow, fi
...
+     * @param data array to copy data into
+     * @param nRow number of rows to copy; also the number of rows in <code>data</code>.
+     * @param nCol number of table columns; also the number of columns in <code>data</code>.
+     * @param rows indices of rows to copy.  Null causes rows to be randomly generated.
+     */
+    public static void copyRand(final Table t, final boolean[] data, final int nRow, final int nCol, final long[] rows) {
+        assertCopyRandArgs(t, data == null ? -1 : data.length, nRow, nCol, rows);
+        assert data != null;
+        copyRand(t, nRow, nCol, rows, boolean.class, "(boolean)", (cs, k, idx) -> data[idx] = cs.getBoolean(k));
+    }
```
chipkent, line 514: This comment seems to indicate that these constructors were intentionally removed so that byte[] is used.  Does the boolean cast throw an error?  Are you certain this code is executed?

chipkent: Looking at the code, I think there is a test.  Possibly the concern is `Java2NumpyCopy.java:382`.  It looks like we automatically cast from type A to type B.  Maybe there is something funny there.  It isn't obvious to me what the problem is.  Just make certain that the unit tests are getting run.  You can also run them with code coverage in IJ to make sure that the code is getting executed.

Integrations/src/main/java/io/deephaven/integrations/numpy/Java2NumpyCopy.java
chipkent: Looking at the code, I think there is a test.  Possibly the concern is `Java2NumpyCopy.java:382`.  It looks like we automatically cast from type A to type B.  Maybe there is something funny there.  It isn't obvious to me what the problem is.  Just make certain that the unit tests are getting run.  You can also run them with code coverage in IJ to make sure that the code is getting executed.

chipkent (changes requested):

Integrations/src/main/java/io/deephaven/integrations/numpy/Java2NumpyCopy.java
alexpeters1208: Tested with coverage and all of these methods are covered.

Integrations/build.gradle
alexpeters1208: One extra line from Colin?

chipkent (approved):

Integrations/build.gradle
niloc132: restoring the java.srcDirs is necessary to tell the java plugin to compile files in that dir - i am not at all sure we need test_python in resources.srcDirs though.

The later line (to be added in an upcoming commit?) that changes "test" to "check" is a convenience to be able to easily just run java tests, instead of re-running python tests as part of java tests. With it change to "check", the python tests will still run as part of the build, you can confirm this by checking the "check" ci task output.

niloc132: In other words, the tests should run either way with the other change, this just makes it a little nicer to just run java tests for debugging and such.

Integrations/src/test/java/io/deephaven/integrations/numpy/Java2NumpyCopyTest.java
```diff
@@ -7,9 +7,9 @@
 
 public class Java2NumpyCopyTest extends BaseArrayTestCase {
 
-    final private String file = Configuration.getInstance().getDevRootPath() + "/Integrations/src/test/resources/io/deephaven/integrations/numpy/dh.jpg";
-    final private String q1 = Configuration.getInstance().getDevRootPath() + "/Integrations/src/test/resources/io/deephaven/integrations/numpy/quadrant1.jpg";
-    final private String q2 = Configuration.getInstance().getDevRootPath() + "/Integrations/src/test/resources/io/deephaven/integrations/numpy/quadrant2.png";
+    final private String file = Configuration.getInstance().getDevRootPath() + "/Integrations/src/test/java/io/deephaven/integrations/numpy/dh.jpg";
```
niloc132, line 10: consider just io/deephaven/integration/... for each instead, and wherever you consume these paths, use the classloader to load them instead

chipkent: I took a look at this with Alex.  As written, the `DbFile` and `DbFile` impl take `java.io.File` as input, and not `InputStream` which would come from the classloader.  It is not worth a rework to do this change right now.

niloc132 (approved):

Integrations/src/test/java/io/deephaven/integrations/numpy/Java2NumpyCopyTest.java
alexpeters1208: Since this fix is not going in this time, this is ready to close.

PR #1079, opened 2021-08-17
TableMap support in KafkaTools

### LINKED ISSUE #1038: Kafka: KafkaIngester to TableMaps for different partitions
As a user, I want to split the Kafka partitions to individual tables within a TableMap rather than a merged table so that I can enable more parallelism.

Part of #293

rcaudy:
Closes #1038


redpanda-apicurio/examples/python/kafka-produce-avro.py
```diff
@@ -41,7 +41,7 @@
 # The last command above should create a table with columns: [ KafkaPartition, KafkaOffset, KafkaTimestamp, Symbol, Price ]
 #
 # Run this script on the host (not on a docker image) to generate one row:
-# $ python3 ./kafka-produce-avro.py share_price avro/share_price.json str:Symbol=MSFT str:Side=BUY double:Price=274.82 int:Qty=200
+# $ python3 ./kafka-produce-avro.py share_price 0 avro/share_price.json str:Symbol=MSFT str:Side=BUY double:Price=274.82 int:Qty=200
```
cpwright, line 44: I wonder if we should start using argparse here to avoid making these examples too cumbersome.

KafkaIngester/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -303,7 +313,75 @@ private Json(
...
+    /**
+     * Map "Python-friendly" table type name to a {@link TableType}.
+     *
+     * @param typeName The friendly name
+     * @return The mapped {@link TableType}
+     */
+    public static TableType friendlyNameToTableType(@NotNull final String typeName) {
+        // @formatter:off
+        switch (typeName) {
+            case "streaming"    : return TableType.Stream;
```
cpwright, line 371: Should we change these to "stream" to match the TableType and also "StreamTable" as opposed to "StreamingTable".

cpwright (approved):

KafkaIngester/src/main/java/io/deephaven/kafka/KafkaTools.java
rcaudy: I think yes, if @jcferretti agrees.

redpanda-apicurio/examples/python/kafka-produce-avro.py
rcaudy: I don't want to re-write the examples in this PR. Happy if @jcferretti  wants to do so in a subsequent one.

jcferretti: Not accepting "thrown under the bus offer" on this one.  Either this change is at a good enough usability threshold or not.  Whoever breaks the usability good enough threshold fixes it.  In the same PR or a subsequent one as a result of a negotiation with their reviewers.  I propose Charles makes a call on whether this is breaking the usability threshold or not.

KafkaIngester/src/main/java/io/deephaven/kafka/KafkaTools.java
jcferretti: You native and educated english speakers make a call here.  To me, "streaming table" sounds like a sentence with more obvious meaning than "stream table".  "stream table" sounds weird to me, like two nouns in the same sentence leaving me in a "What?" state.

rcaudy: He didn't agree, but I did it anyway.

cpwright (approved):

PR #1081, opened 2021-08-17
Move away from using **kwargs in python Kafka API.

jcferretti:
Thanks to Jianfeng for pointing out.

### DIFF
--- a/Generators/src/main/java/io/deephaven/pythonPreambles/KafkaToolsPreamble.txt
+++ b/Generators/src/main/java/io/deephaven/pythonPreambles/KafkaToolsPreamble.txt
@@ -161,7 +161,6 @@ def consumeToTable(*args, **kwargs):
 
 @_passThrough
 def avro(schema, **kwargs):
-    schema_version = kwargs.pop('schema_version', "latest")  # default may never be used if schema is not of str type.
     mapping = kwargs.pop('mapping', None)
     mapping_only = kwargs.pop('mapping_only', None)
     if mapping is not None and mapping_only is not None:
@@ -190,9 +189,17 @@ def avro(schema, **kwargs):
         have_actual_schema = False
     elif instanceof(schema, _avro_schema_jtype_):
         have_actual_schema = True
+        if kwargs.has_key('schema_version'):
+            raise Exception("keyword argument 'schema_version' is only expected if schema is of str type.")
     else:
         raise Exception("first positional argument for schema expected to be of " +
                         "type str or avro schema type, instead got " + str(schema))
+
+    schema_version = kwargs.pop('schema_version', 'latest')  # default may never be used if schema is not of str type.
+
+    if len(kwargs) > 0:
+        raise Exception("excess keyword arguments not understood given: " + str(kwargs))
+
     if have_mapping:
         if have_actual_schema:
             return _java_type_.avroSpec(schema, mapping)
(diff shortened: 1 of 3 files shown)


Generators/src/main/java/io/deephaven/pythonPreambles/KafkaToolsPreamble.txt
```diff
@@ -73,56 +74,58 @@ def _passThrough(wrapped, instance, args, kwargs):
...
-        fieldNamesArray = jpy.array('java.lang.String', dict.keys())
-        columnNamesArray = jpy.array('java.lang.String', dict.values())
-        mapping = _java_type_.fieldNameMappingFromParallelArrays(fieldNamesArray, columnNamesArray)
-        return _java_type_.avroSchemaToColumnDefinitions(schema, mapping)
-    raise Exception('too many arguments: ' + len(args))
+def _custom_avroSchemaToColumnDefinitions(schema, mapping:dict = None):
+    if mapping is None:
+        return _java_type_.avroSchemaToColumnDefinitions(schema)
+
+    if not isinstance(mapping, dict):
```
cpwright, line 81: Still necessary with that mapping:dict specifier?

Generators/src/main/java/io/deephaven/pythonPreambles/KafkaToolsPreamble.txt
```diff
@@ -131,68 +134,69 @@ def consumeToTable(*args, **kwargs):
...
     elif not isinstance(offsets, jpy.JType):
         raise Exception(
-            "type " + type(offsets).__name__ +
-            "  of keyword argument 'offsets' not recognized; only str or dict allowed")
+            "value " + str(offsets) + " of type " + type(offsets).__name__ +
+            "  not recognized for argument 'offsets'; only str, dict or predefined constants allowed")
 
-    key = kwargs.pop('key', FROM_PROPERTIES)
-    value = kwargs.pop('value', FROM_PROPERTIES)
+    if key is None:
```
cpwright, line 140: It seems like you should use FROM_PROPERTIES in the method signature default.

Generators/src/main/java/io/deephaven/pythonPreambles/KafkaToolsPreamble.txt
```diff
@@ -206,43 +210,37 @@ def avro(schema, **kwargs):
...
 
 @_passThrough
-def simple(*args):
-    if len(args) < 1 or len(args) > 2:
-        raise Exception("one or two arguments expected, instead got " + len(args))
-    colName = args[0]
-    if not _isStr(colName):
-        raise Exception("column_name argument needs to be of str type, instead got " + colName)
+def simple(column_name:str, data_type_str:str = 'string'):
+    if not _isStr(column_name):
```
cpwright, line 233: still needed?

Generators/src/main/java/io/deephaven/pythonPreambles/KafkaToolsPreamble.txt
jcferretti: We should, but is sad and we can't.  FROM_PROPERTIES is an object that is not initialized by the time the def definition is processed.   If we say `def(..., some_arg=SOME_CONSTANT, ...)` for a `SOME_CONSTANT` that we initialize from a Java object static via jpy, when python processed the `def` it finds the initial value of `None` for the constant and default arguments are never evaluated again, only once at the time the def is processed, so it doesn't work.   See also https://illumon.slack.com/archives/C028TLCCSPR/p1629244827144200

Generators/src/main/java/io/deephaven/pythonPreambles/KafkaToolsPreamble.txt
jcferretti: Yes, there is no enforcement from python, only warnings from linters/IDEs.

PR #1082, opened 2021-08-18
Application mode for long running workers

### LINKED ISSUE #103: gRPC API for fetching various objects by name from a scope
With only one scope type at this time, we can only fetch from a console session, but in the future we will probably want to be able to support different kinds of scopes. Focusing on console support also means that we need to make the fetch operations specific to consoles and their languages.

### LINKED ISSUE #853: Provide a command line argument to a script to seed the application
As a developer, I want to be able to specify a script on the command line to seed the session so that I can share my work easily with others.

Related to #695

### LINKED ISSUE #854: Provide list of objects in the existing session when a client connects
As a client UI, I want a list of objects that are available on the server so that I can open tables/plots that may have been created before I connected.

Relates to #695

### LINKED ISSUE #923: Add TableService getExportedTableCreationResponse gRPC
Users that want to recover from a lost ExportedTableCreationMessage, want to acquire a duplicate ETCR, or get information from any other kind of ticket (without exporting it), should be able to do so.

This method belongs on (and easily integrates with) the table API.

### LINKED ISSUE #952: Support canceling tickets instead of just releasing them
Presently the only option a grpc client has is to release a ticket while pending. The export will still finish its work in that case, and then any downstream work will begin before this item is released, so that those downstream items should be able to finish.

A cancel operation instead would optionally stop the currently finishing work, and also prevent downstream work from beginning. If canceled after the work is completed, the ticket will simply be released.

The logic for this is implemented, but it is not expose via gRPC at this time.

nbauernfeind:
New Flags:
```
  -Ddeephaven.console.disable   # prevents any user from accessing any repl session
  -Ddeephaven.application.dir=/path/to/app.d  # instruct the worker where to find configured applications
```

Due to the restriction that we have no more than one deephaven script session, python and groovy applications cannot be mixed at this time. Note that multiple python (or groovy) scripts share the same query scope which bleed into each other, and into the query scope available to a REPL user. The `deephaven.console.disabled` flag is handy here, but the user is also welcome to encapsulate any state they wish does not leak to the REPL. The variety of options here yield a ton of flexibility.

Supportive Functionality:
- Fixes #103: ExportFromTicket enables a ticketRouter.resolve(source_ticket) and export into a session managed result_ticket (this is the modern version of fetch)
- Fixes #923: support get ETCR from a tickerRouter.resolve(source_ticket)
- Fixes #854: provide list of fields in query scope if not in console disabled mode, provide list of application fields if Application.dir flag supplied
- Fixes #853: provide command line flags (and other configuration) to support application mode
- The listFields() call subscribes to future updates, too.
- Added ExportNotification.State.RUNNING to distinguish between queued and executing.

Bug Fixes:
- Fixes #952: release proactively cancels; there is otherwise a race between steps 2 and 3 depending on network latency and out-of-ordering of messages. 1) export e0, 2) export e1 depends on e0, 3) release e0. Instead, the client would need to wait until it sees that all depending exports are PENDING before releasing e0. With a proactive cancel, the client still has to keep track of the same state, but cannot release until all depending exports are EXPORTED. I've introduced a ReleaseRequest in case we want to add-back the "smart-release" functionality --- but better the client gets bit by the proactive cancel and quickly realizes they need to track all of the export state changes.
- wrap new console sessions in a delegating script session that upgrades modified query scope variables to be "new query scope" variables so that the UI may better distinguish multi-window scenarios where a query scope field is overwritten

Sample Groovy Application
---------------------------
`/data/app.d/01-groovy.app`
```properties
type=script
scriptType=groovy
enabled=true
id=io.deephaven.db.appmode.ApplicationConfigs.app01
name=My Groovy Application
file_0=01-groovy.groovy
file_1=01-groovy-qs.groovy
```
`/data/app.d/01-groovy.groovy`
Description: use ApplicationContext and lambdas to encapsulate your application state
```groovy
import io.deephaven.appmode.ApplicationContext
import io.deephaven.db.tables.utils.TableTools

def start = { ApplicationState app ->
    size = 42
    app.setField("hello", TableTools.emptyTable(size))
    app.setField("world", TableTools.timeTable("00:00:01"))
}

ApplicationContext.initialize(start)
```

`/data/app.d/01-groovy.groovy`
Description: you can leak content via the query scope
```groovy
import io.deephaven.db.tables.utils.TableTools

// Use QueryScope! Careful; this leaks into the REPL state!
size = 42
hello = TableTools.emptyTable(size)
world = TableTools.timeTable("00:00:01")
```

Sample Static Class Application
--------------------------------
`/data/app.d/00-static.app`
```properties
type=static
class=io.deephaven.db.appmode.ApplicationConfigs$App00
```
Class Impl:
```java
    public static class App00 implements Application.Factory {

        @Override
        public final Application create() {
            Field<Table> hello = StandardField.of("hello", TableTools.emptyTable(42).view("I=i"), "A table with one column 'I' and 42 rows, 0-41.");
            Field<Table> world = StandardField.of("world", TableTools.timeTable("00:00:01"));
            return Application.builder()
                    .id(App00.class.getName())
                    .name("My Class Application")
                    .fields(Fields.of(hello, world))
                    .build();
        }
    }
```

Note that the DynamicApplication can add/remove fields during the runtime of the application, and is just as easy an API to use.


grpc-api/src/test/app.d/01-groovy-b.groovy
```diff
@@ -0,0 +1,4 @@
+import io.deephaven.appmode.ApplicationContext
+
+app = ApplicationContext.get()
```
devinrsmith, line 3: We should be careful in examples. `app` escapes to the global query scope. Likely, want to do `def app = ...`

grpc-api/src/test/java/io/deephaven/db/appmode/ApplicationConfigTest.java
```diff
@@ -0,0 +1,51 @@
+package io.deephaven.db.appmode;
+
+import org.junit.Ignore;
+import org.junit.Test;
+
+import java.io.IOException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Ignore("cannot predict working directory; tests can't predictably find files")
```
devinrsmith, line 10: We should try to fix this.

application-mode/src/main/java/io/deephaven/db/appmode/ApplicationState.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public synchronized void removeFields(Iterable<String> names) {
+        for (String name : names) {
+            removeField(name);
+        }
+    }
+
+    public Ticket ticketForField(Field<?> field) {
+        return null;
```
devinrsmith, line 100: Todo?

DB/src/main/java/io/deephaven/db/util/DelegatedScriptSession.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Supplier;
+
+/**
+ * The delegated script session delegates all calls to another script session. When evaluating a script
+ * it massages the Changes to the QueryScope so that any modifications that are being seen for the first
+ * time by the api-client come across as new entries and not as modified entries.
+ */
+public class DelegatedScriptSession implements ScriptSession {
```
devinrsmith, line 23: In what context is this necessary? I think I missed it elsewhere in the review...

devinrsmith (changes requested):

web/client-api/src/main/java/io/deephaven/ide/shared/IdeSession.java
```diff
@@ -66,17 +68,17 @@ public IdeSession(
 
     // TODO: #37: Need SmartKey support for this functionality
     @JsIgnore
-    public Promise<JsTreeTable> getTreeTable(String name) {
-        return connection.getTreeTable(name, result);
+    public Promise<JsTreeTable> getTreeTable(JsVariableDefinition name) {
```
mofojed, line 71: Why did all these methods (`getTable`, `getTreeTable`, `getFigure`) all change to require a variable definition now? That's changing the existing API, and doesn't really give us anything of value over just passing the `name` (or `id`) of the variable.

web/client-api/src/main/java/io/deephaven/web/client/api/console/JsVariableDefinition.java
```diff
@@ -1,58 +1,83 @@
...
 
     @JsProperty
+    public String getType() {
+        return type;
+    }
+
+    @JsProperty @Deprecated
     public String getName() {
-        return name;
+        return title;
```
mofojed, line 66: I think this should return the `id`, as that's the unique name/equivalent

mofojed: Actually I think I'm misunderstanding something. The `name` seems to map to `title` instead of `id`.
This changes the API considerably, and IMO makes it more difficult for the client to use. The client doesn't care about these arbitrary IDs that seem to be completely random, and stores variables based on the name in that session (of which there can only be one in that scope). Within that scope (the IdeSession), it wants to get an object with that name, it should just be able to call `getTable(name)` and get that object, it shouldn't require a full definition.

nbauernfeind: We decided to deprecate name, because it is no longer a unique identifier.

An "old variable" gets mapped like this:
id = name
title = name
description = ""
type = type

Title is just a different word for `name` and represents a human readable name for the table, but is not unique and cannot identify the variable on the server.

nbauernfeind: > client doesn't care about these arbitrary IDs that seem to be completely random
Well, now it needs to, because it is not random it is what the server uses to identify the variable. It looks like nonsense because it is binary and is b64 encoded here.

mofojed: When the client runs code like:
```
session.runCode(`t = emptyTable(100).update("A=i")`)
```
After that completes, it's expected there's a table with the name `t` within that scope. Are you saying that is invalid? I don't understand why the client would care about the arbitrary IDs when the session/name combo should also be unique.

How are the id's generated? Will they be consistent if you restart the server?

mofojed: I guess I'm saying given the IdeSession, the variable `name` should already be unique... so why force the client to require knowledge of the id when you should be able to derive that?

And what's the point of having the `.getTable`, `.getTreeTable`, etc. methods if they still take an object just like `.getObject`? It's assumed client already knows they are tables (hence calling that method), there's no checks right now that the type actually matches the method. The session, type, and variable name are all known already, so the id should be deriveable

nbauernfeind: The id's are persistent and consistent. For application fields they are structured like this: "a/$app_id/f/$field_name". For Query scope they are structured like "s/$field_name". These ticket patterns are not meant for the client to understand or construct on their own -- just use it to help the server identify which of these you want if they are named the same. DHE would shadow the variables exposed from the "application-script" equivalent -- which, I understand, is why remote REPL is more-or-less disabled for PQs. This solves the shadowing problem by introducing a "hybrid" mode where you can still connect to the console while in application mode.

There is a restriction that a REPL should not be able to destroy application state -- down the road when we are not sharing script sessions, this enables all connecting clients to get their own query scope namespace while still having access to the application mode exported fields.

Ideally I think the client would just keep around the JsVariableDefinition and use it to look things up (getObject, for example, already takes a JsVariableDefinition).

> And what's the point of having the .getTable, .getTreeTable, etc. methods if they still take an object just like .getObject? It's assumed client already knows they are tables (hence calling that method), there's no checks right now that the type actually matches the method. The session, type, and variable name are all known already, so the id should be deriveable

@niloc132 recommended that the client knows the type of the variable, that it should try to load it as that type -- but that is hidden from the jsapi user directly. I'm not sure I see a problem calling `getObject()` as it delegates to getTable/getTreeTable/etc. The delegate methods are there to help the jsapi catch race conditions for when a field's type changes before the UI knew about it. (i.e. load a layout that knows this tab is a Figure, but it turned out to be a Table because the application developer changed his mind but didn't change your custom layout)

mofojed: Okay, so in terms of an IdeSession - can an IdeSession have more than one $app_id then? Or is it a Query scope?
I don't understand what you mean about catching race conditions - how does it help catch a race condition? The `.getTable` function only really reads the `id` from the definition to actually retrieve the object, it doesn't check the type or title or anything. That's all just printed out as debug info.

nbauernfeind: `getTable` calls an API method that verifies the retrieved variable is a Table and not any other kind of type. The race condition is between the API client's knowledge of the field (e.g. as a client I call `getTable` which expects to find a Table or else error) and between the server state changing (e.g. two chrome tabs are open, while one is fetching a query scope variable named `t` the other changes the variable -- one request will race the other and will cause getTable to fail in one ordering but not the other).

An IdeSession can have multiple app_id's yes. This is one thing Colin was discussing with you w.r.t. the QueryConfig pattern in DHE. An Application is a bag of fields. With this layer of abstraction, it becomes easier to think about how one might put together a dashboard that combines data from different workers.

mofojed: Okay. I think I'll be okay because I'm already calling just `getObject`... but to re-iterate - I don't see any value in exposing `.getTable`/`.getFigure`. etc and require the entire `VariableDefinition` to be passed in; or if you do, you should assert immediately that the type matches the method.

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -516,86 +529,130 @@ public void onError(Throwable throwable) {
...
-    public Promise<JsTable> getPandas(String name, Ticket script) {
+    public Promise<JsTable> getPandas(JsVariableDefinition varDef) {
         return whenServerReady("get a pandas table").then(serve -> {
-            JsLog.debug("innerGetPandasTable", name, " started");
+            JsLog.debug("innerGetPandasTable", varDef.getTitle(), " started");
             return newState(info,
                     (c, cts, metadata) -> {
-                        JsLog.debug("performing fetch for ", name, " / ", cts, " (" + LazyString.of(cts::getHandle), ",", script, ")");
+                        JsLog.debug("performing fetch for ", varDef.getTitle(), " / ", cts, " (" + LazyString.of(cts::getHandle), ")");
 //                        if (script != null) {
```
niloc132, line 557: script param is removed, kill this commented out code

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -516,86 +529,130 @@ public void onError(Throwable throwable) {
...
+                        } else {
+                            created[created.length] = fd;
+                        }
+                    }
+                }
+
+                notifyFieldsChangeListeners(new JsVariableChanges(created, updated, removed));
+            });
+            fieldsChangeUpdateStream.onEnd(status -> {
+                //TODO handle reconnect
```
niloc132, line 634: call checkStatus in here

web/client-api/src/main/java/io/deephaven/web/client/api/console/JsVariableDefinition.java
```diff
@@ -1,58 +1,83 @@
...
-    private String name;
-    private String type;
-
-    @JsIgnore
     public static JsVariableDefinition from(Object definitionObject) {
         if (definitionObject instanceof JsVariableDefinition) {
             return (JsVariableDefinition)definitionObject;
         } else {
-            return new JsVariableDefinition(Js.cast(definitionObject));
+            return new JsVariableDefinition(Js.<JsPropertyMap<Object>>cast(definitionObject));
```
niloc132, line 18: use Js.asPropertyMap(definitionObject)

web/client-api/src/main/java/io/deephaven/web/client/api/console/JsVariableChanges.java
```diff
@@ -17,18 +14,37 @@
                                 OTHERWIDGET = "OtherWidget",
                                 PANDAS = "Pandas";
 
-    private JsVariableDefinition[] created;
-    private JsVariableDefinition[] updated;
-    private JsVariableDefinition[] removed;
-
-    private static JsVariableDefinition[] convertDefinitions(VariableDefinition[] definitions) {
-        return Arrays.stream(definitions).map(def -> new JsVariableDefinition(def.getName(), def.getType())).toArray(JsVariableDefinition[]::new);
+    public static String getVariableTypeFromFieldCase(int fieldCase) {
```
niloc132, line 17: maybe make an array to hold these, and make the indexes line up with what we expect here? probably easier than if/else/etc

if we have to update, that'll be easier to change too

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/application.proto
```diff
@@ -0,0 +1,103 @@
...
+  Ticket ticket = 1;
+  string field_name = 2;
+  FieldType field_type = 3;
+  string field_description = 4;
+  string application_name = 5; // display-friendly identification
+  string application_id = 6; // computer-friendly identification
+
+  message FieldType {
+    oneof field {
+      RemovedField removed = 1;
```
niloc132, line 55: consider making removed = 1000

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/application.proto
```diff
@@ -0,0 +1,103 @@
...
+ * Represents a batch of fields.
+ */
+message FieldsChangeUpdate {
+  repeated FieldInfo fields = 1;
+}
+
+/*
+ * A lightweight object describing the exposed field.
+ */
+message FieldInfo {
```
niloc132, line 45: @mofojed any chance you want an icon field here? easy to add later, but to differentiate different plots, different "custom" things.

niloc132: Also, why isn't this a table? wouldnt it make future filtering easier, sorting, eventual access control...

grpc-api/src/test/app.d/01-groovy.groovy
```diff
@@ -0,0 +1,7 @@
+import io.deephaven.appmode.ApplicationContext
+import io.deephaven.db.tables.utils.TableTools
```
niloc132, line 2: this should be unnecessary, right? or are we not doing "free" imports any more?

probably should write the kind of code we expect to have users write.

--

also, would it not make sense to have the AppContext import and app = AppContext.get() binding already done for you for these scripts? since any script that does anything interesting will need to do it anyway..

grpc-api/src/main/java/io/deephaven/appmode/ApplicationContext.java
```diff
@@ -0,0 +1,29 @@
+package io.deephaven.appmode;
+
+import io.deephaven.db.appmode.ApplicationState;
+
+/**
+ * This application context can be used to get access to the application state from
+ * within script applications.
```
niloc132, line 7: comment about how this is actually part of the app mode API, but lives in grpc-api, and why it doesn't behave as a dagger singleton, can't be an singleton instance

grpc-api/src/main/java/io/deephaven/appmode/ApplicationFactory.java
```diff
@@ -0,0 +1,93 @@
+package io.deephaven.appmode;
+
+import com.google.rpc.Code;
+import io.deephaven.db.appmode.DynamicApplication;
```
niloc132, line 4: packages are weird - why does appmode, in the appmode project, use the io.deephaven.db.appmode package, while not depending on db, but we have io.deephaven.appmode in the grpc-api project?

can we move this to grpc-api, and remove db from packages in appmode (that are explicitly not part of db)?

grpc-api/src/main/java/io/deephaven/grpc_api/app_mode/ApplicationTicketResolver.java
```diff
@@ -0,0 +1,71 @@
...
+
+    }
+
+    public void visitAllApplications(Consumer<ApplicationState> consumer) {
+
+    }
+
+    @Override
+    public <T> SessionState.ExportObject<T> resolve(@Nullable SessionState session, ByteBuffer ticket) {
+        throw new UnsupportedOperationException("not yet implemented");
```
niloc132, line 39: TODOs and tickets?

grpc-api/src/main/java/io/deephaven/grpc_api/app_mode/ApplicationTicketResolver.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public String getLogNameFor(ByteBuffer ticket) {
+        throw new UnsupportedOperationException("not yet implemented");
+    }
+
+    @Override
+    public void forAllFlightInfo(@Nullable SessionState session, Consumer<Flight.FlightInfo> visitor) {
+        // TODO NOCOMMIT (NATE): implement
```
niloc132, line 69: remove/ticket/etc

grpc-api/src/main/java/io/deephaven/grpc_api/console/ConsoleServiceGrpcImpl.java
```diff
@@ -360,23 +329,16 @@ public void closeDocument(CloseDocumentRequest request, StreamObserver<CloseDocu
     public void fetchFigure(FetchFigureRequest request, StreamObserver<FetchFigureResponse> responseObserver) {
         GrpcUtil.rpcWrapper(log, responseObserver, () -> {
             final SessionState session = sessionService.getCurrentSession();
-
-            SessionState.ExportObject<ScriptSession> exportedConsole = session.getExport(request.getConsoleId());
+            final SessionState.ExportObject<Object> figure = ticketRouter.resolve(session, request.getSourceId());
```
niloc132, line 332: since we're using this instead to fetch objects by name, make sure it throws NOT_FOUND

grpc-api/src/test/app.d/01-groovy.groovy
devinrsmith: I'd like to get rid of "free" imports here where we have the chance to do something.

Also, I don't like implicit bindings that scripts just "know" about. Often times, I'd see PQs at customers where they would do:

```groovy
db = (Database)db
```
just so they would have strong typing. I'd prefer to create a pattern where scripts are explicit.

Alternatively, we can have a pattern where instead of a "script" the scripts provide a function with a known signature:

```groovy
def script_eval(app):
   ...
```
But that feels a bit clunkier.

devinrsmith (approved):

Issue #730, opened 2021-06-09
Complete, test JS API reconnects


niloc132:
Since replacing the websocket and gwt-rpc serialization with http2/grpc/protobuf, we haven't had an actual server deployed suitable for testing reconnecting a session after some kind of network loss, so the work is incomplete still. Presently we only are testing with the websocket proxy transport (so as to be useful on localhost or in other ssl-less environments, since http2 can't be used in the browser without ssl).

While we're probably interested in websocket proxy reconnects working correctly, it isn't as much of a priority as actual http2 reconnects, so the first step of this task is to get the stack running on a server with ssl and no websockets, then bounce envoy or the network connection to periodically break the streams and make sure that we are able to restore the session, or in cases where the session cannot correctly come back, restore tables which can be re-fetched.

devinrsmith:
My setup is:

crypto -> kafka -> grpc-api.

Restarting just the crypto service container causes the web ui message to display 

```
"Unable to connect: Response closed without grpc-status (Headers only)"
```

a few seconds *after* crypto restarts (and I notice that *new* data *is* coming in to tables for those few seconds). It's followed by messages on the server of the form 

```
ERROR c-nio-worker-ELG-3-1 | i.d.g.a.ArrowFlightUtil   | DoExchangeMarshaller{4e942be2}: unexpected error; force closing subscription: caused by io.grpc.StatusRuntimeException: CANCELLED: client cancelled
```

but, probably harmless according to https://github.com/deephaven/deephaven-core/issues/1205.

PR #1086, opened 2021-08-19
JS API Error handling

### LINKED ISSUE #1065: Table Errors are Ignored
**Description**

If I have a table that produces an error, then the UI does not show it to me.

**Steps to reproduce**

1. Enter a command that will have a failed update into your console.  For example:
```
from deephaven.TableTools import timeTable
tt=timeTable("00:00:01").update("X=ii > 5 ? null : `z`", "Y=X.substring(0)")

```

Wait 6 seconds, until there are enough rows to trigger the NPE.  The table stops ticking, but there is no error indication. 

**Expected results**

On DHE the corresponding code is:
```
tt=db.timeTable("00:00:01").update("X=ii > 5 ? null : `z`", "Y=X.substring(0)")
```
table will have "Unable to open table. Error: Error displaying table." painted over it.

The actual proper behavior would be for the table display to tell me that there was a failure in updating and the original exception.

niloc132:
Added basic error handling, but not proper reconnects or reauth. Three error types are handled for the main streams or polling calls:
 * auth failure which suggests either that the server was restarted or the session timed out somehow - in either case, a new auth is required to resume working
 * Code.INTERNAL, which our grpc-web client library will pass in cases where the response doesn't make sense (broken payloads)
 * Code.UNKNOWN, which our grpc-web client library will pass in cases where the transport doesn't seem to work (no response before connection died)

A new event was added, temporarily, to inform the UI that we are giving up if one of these happens, and the page needs to be reloaded. This will be removed when we have proper reconnect, and inform the UI again that the reconnect succeeded or failed.

Also modified a few errors being sent from the server to hopefully better reflect the intent of the gRPC status codes, and avoid using "INTERNAL" when "something bad" happens.

The UI will need a few small changes to handle this, and to handle a few cases where API calls can fail but not be registered in the UI.

Fixes #1065, but #730 needs to formally address this.

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/QueryConnectable.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/QueryConnectable.java
@@ -2,6 +2,7 @@ package io.deephaven.web.client.api;
 
 import elemental2.core.JsArray;
 import elemental2.core.JsSet;
+import elemental2.dom.CustomEventInit;
 import elemental2.dom.DomGlobal;
 import elemental2.promise.Promise;
 import io.deephaven.ide.shared.IdeSession;
@@ -20,6 +21,7 @@ import io.deephaven.web.shared.fu.RemoverFn;
 import jsinterop.annotations.JsIgnore;
 import jsinterop.annotations.JsMethod;
 import jsinterop.annotations.JsProperty;
+import jsinterop.base.JsPropertyMap;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -65,6 +67,9 @@ public abstract class QueryConnectable <Self extends QueryConnectable<Self>> ext
     @JsProperty(namespace = "dh.QueryInfo")
     public static final String EVENT_CONNECT = "connect";
 
+    @JsProperty(namespace = "dh.IdeConnection")
+    public static final String HACK_CONNECTION_FAILURE = "hack-connection-failure";
+
     private final List<IdeSession> sessions = new ArrayList<>();
     private final JsSet<Ticket> cancelled = new JsSet<>();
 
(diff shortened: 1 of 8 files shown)


niloc132:
Extremely rough patch for @mofojed to finish up nicely in the UI https://github.com/niloc132/web-client-ui/commit/e798052121b5004ca2204d1a7aa7c50f86315b45

web/client-api/src/main/java/io/deephaven/web/client/api/QueryConnectable.java
```diff
@@ -65,6 +67,9 @@ static AuthTokenPromiseSupplier oneShot(ConnectToken initialToken) {
     @JsProperty(namespace = "dh.QueryInfo")
     public static final String EVENT_CONNECT = "connect";
 
+    @JsProperty(namespace = "dh.IdeConnection")
+    public static final String HACK_CONNECTION_FAILURE = "hack-connection-failure";
```
mofojed, line 71: Why is this named `hack-connection-failure`? Why `hack-`?

nbauernfeind (approved):

web/client-api/src/main/java/io/deephaven/web/client/api/QueryConnectable.java
nbauernfeind: I believe 'hack' here is to make it blindly obvious this is a temporary work around to a bigger problem that Colin wants carte blanch to fix correctly.

niloc132: This is temporary, until we have reconnect (#730) which will have this being removed. I wanted to draw a distinction between normal events or internal-only ones, but I'm open to some other way to do that.

I tried to summarize in the description above (and will make it part of the actual commit message), but the old reconnect events are probably what we will use for these cases once the api is smart enough to try reconnecting streams.

mofojed: Okay cool, I've got a change to listen for these new errors pushed on my Application Mode PR: https://github.com/deephaven/web-client-ui/pull/155

mofojed (approved):

PR #1088, opened 2021-08-19
Adding Types.py for wrapping DH column types and table operations around them.

jcferretti:
Initial version contains enough to (1) rewrite Kafka column specifications (in subsequent PR, not here) and (2) provide a way to create table from an array of rows, with a similar API to Panda's DataFrame ctor.

```
import deephaven.Types as dh

data = [ [ "Ashley", 92, 94, 93, 3.9 ],
         [ "Jeff",   78, 88, 93, 2.9 ],
         [ "Rita",   87, 81, 84, 3.0 ],
         [ "Zach",   74, 70, 72, 1.8 ] ]
columns = [ ( "Name",    dh.string ),
            ( "Test1",   dh.int_ ),
            ( "Test2",   dh.int_ ),
            ( "Average", dh.int_ ),
            ( "GPA",     dh.double ) ]
grades = dh.table_of(data, columns)
```

Deephaven - Google Chrome_079

Selection_078


Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,229 @@
...
+    return wrapped(*args, **kwargs)
+
+try:
+    _defineSymbols()
+except Exception as e:
+    pass
+
+@_passThrough
+def asType(type_str):
+    return _table_tools_.typeFromName(type_str)
```
devinrsmith, line 62: So, returning a class is *not* a proper way to get the type information into jpy. You need to use `jpy.get_type(...)`, as it will then be recognized as a proper python "type".

To demonstrate difference in return types between get_type and something that returns Class:

```python
>>> Integer = jpy.get_type('java.lang.Integer')
>>> type(Integer)
<class 'type'>
>>>
>>> type(Integer(1).getClass())
<class 'java.lang.Class'>
```

So:

```python
int_ = jpy.get_type('int')
int_array = jpy.get_type('[I')
stringset = jpy.get_type('io.deephaven.db.tables.libs.StringSet')
```

etc.

Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,229 @@
...
+
+bool_ = asType('java.lang.Boolean')
```
devinrsmith, line 64: So, numba and pandas represent bool_ as a byte; and I don't think they have the concept of a "null" boolean. Does this make our booleans incompatible?

Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,229 @@
...
+float_ = asType('float')
+single = float_   # make life simple for people who are used to NumPy
+float32 = float_  # make life simple for people who are used to pyarrow
+double = asType('double')
+float64 = double  # make life simple for people who are used to pyarrow
+string = asType('java.lang.String')
+bigdecimal = asType('java.math.BigDecimal')
+stringset = asType('io.deephaven.db.tables.libs.StringSet')
+
+byte_array = asType('byte')
```
devinrsmith, line 81: Do you mean for an array here?

Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,229 @@
...
+    for c in range(ncols):
+        col_name = columns[c][0]
+        col_type = columns[c][1]
+        col_data = []
+        for r in range(nrows):
+            row = data[r]
+            if len(row) < c + 1:
+                raise Exception("not enough columns provided in row " + r)
+            col_data.append(row[c])
+        jvalues = jpy.array(_jpyTypeFromDh(col_type), col_data)
```
devinrsmith, line 223: You should be able to use col_type directly here, assuming it's a type returned from `jpy.get_type`.

Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,229 @@
...
+single_array = float_array
+float32_array = float_array
+double_array = asType('double[]')
+float64_array = double_array
+string_array = asType('java.lang.String[]')
+
+datetime = asType('io.deephaven.db.tables.utils.DBDateTime')
+
+@_passThrough
+def _jpyTypeFromDh(t):
```
devinrsmith, line 98: This mapping is unnecessary when the types come from `jpy.get_type`.

Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,229 @@
...
+                    and type is the intended Deephaven column data type.
+    :return: a table with columns and data as specified in the arguments.
+    """
+
+    if columns is None or len(columns) == 0:
+        if data is not None or len(data) > 0:
+            raise Exception("when no column definitions are provided in the 'columns' argument, " +
+                            "only an empty table can be created, and no data can be specified; instead " +
+                            "got a non-empty 'data' argument with " + str(data))
+        return _table_tools_.emptyTable()
```
devinrsmith, line 189: empty tables still need a size for number of rows.

Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,229 @@
...
+    if len(t) == 2:
+        return _col_def_.fromGenericType(col_name, data_type)
+    component_type = t[2]
+    if not isinstance(component_type, _jclazz_type_):
+        raise Exception('Element at index 2 (' + str(component_type) + ') for component type has the wrong type ' +
+                        type(component_type).__name__)
+    return _col_def_.fromGenericType(col_name, data_type, component_type)
+
+@_passThrough
+def cols(ts):
```
devinrsmith, line 150: This method seems unused?

Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,229 @@
...
+
+import jpy
+import wrapt
+import sys
+import collections
+
+# None until the first _defineSymbols() call
+_table_tools_ = None
+_col_def_ = None
+_python_tools_ = None
```
devinrsmith, line 16: This seems unused.

Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,229 @@
...
+            if len(t) != 2:
+                raise Exception("only two element type tuples are supported, instead got " + str(t))
+            try:
+                if col_header is None:
+                    col_header = _qst_col_header_.of(t[0], t[1])
+                else:
+                    col_header = col_header.header(t[0], t[1])
+            except Exception as e:
+                raise Exception("Could not create column definition from " + str(t)) from e
+        return _table_.of(col_header)
```
devinrsmith, line 207: I don't think this is valid, can't call `of` on a col_header. To create a table with zero rows from a ColumnHeader(s) you can do:

`_table_.of(_qst_newtable_.empty(col_header.tableHeader()))`, or something to that effect.

Alternatively, it's perfectly fine to not handle this as a special case, and just fall through to the general case and create Columns with no data in it.

devinrsmith (changes requested):

Integrations/python/deephaven/Types.py
devinrsmith:
```python
>>> import numpy as np
>>> type(np.int32)
<class 'type'>
```

jcferretti: Done, take a peek hopefully now makes more sense.

jcferretti: Done, take a peek, hopefully now makes more sense.

Integrations/python/deephaven/Types.py
jcferretti: Without providing some kind of conversion, probably.  Do we need a dh.bool_from_byte method?

Integrations/python/deephaven/Types.py
devinrsmith: We could make the `NewTable.empty` method a bit more generic, and accept `Iterable<ColumnHeader>` instead of `TableHeader`, in which case we can get rid of one layer of adapting:

```python
_table_.of(_qst_newtable_.empty(col_header))
```

jcferretti:
We added support for creating tables by column, passing in a first argument of dict, like:
```
#
# By column
#

import deephaven.Types as dh

data = {
    "Name"    : ( dh.string, [ "Ashley", "Jeff", "Rita", "Zach"  ] ),
    "Test1"   : ( dh.int_,   [    92,      78,    87,      74,   ] ),
    "Test2"   : ( dh.int_,   [    94,      88,    81,      70,   ] ),
    "Average" : ( dh.int_,   [    93,      93,    84,      72,   ] ),
    "GPA"     : ( dh.double, [     3.9,     2.9,   3.0,     1.8  ] )
}

grades2 = dh.table_of(data)


--

#
# No data.
#

import deephaven.Types as dh

column_definitions = {
    "Name"    : dh.string,
    "Test1"   : dh.int_,
    "Test2"   : dh.int_,
    "Average" : dh.int_,
    "GPA"     : dh.double
}

grades3 = dh.table_of(column_definitions)
```

Deephaven - Google Chrome_086

devinrsmith:
So, I think there may be value in trying to directly reference `io.deephaven.qst.type.Type` as a field in a class that is a python Type. I know I led you down the `jpy.get_type(...)` route, but I think the adapting between different objects would be easier without having to do the map/class lookups.

For example, instead of having to call `fromGenericType` (which is incorrect for primitive types), we want to rely on the exact conversions from the QST when applicable:

```
public static ColumnDefinition<?> from(ColumnHeader<?> header)
```

And then also, I question if we actually need python support for creating/returning ColumnDefinitions? Or, would sticking w/ ColumnHeader serve us better? I'm not sure where `cols(...)` is used from.

devinrsmith:
Oh, I see the cols() calls from kafka tools now. I wonder if we should try to use `ColumnHeader<?>` in KafkaTools instead of `ColumnDefinition`?

Generators/src/main/java/io/deephaven/pythonPreambles/KafkaToolsPreamble.txt
```diff
@@ -211,7 +215,7 @@ def json(col_defs, mapping:dict = None):
         raise Exception("'col_defs' argument needs to be a sequence of tuples, instead got " +
                         str(col_defs) + " of type " + type(col_defs).__name__)
     try:
-        col_defs = _tuplesListToColDefsList(col_defs)
+        col_defs = dh.cols(col_defs)
```
jcferretti, line 218: This is the place where we need cols.  The Java side API for Kafka ingestion takes a `ColumnDefinition[]` for JSON.

jcferretti:
This latest version has python DataType rebuilt on top of qst.type.Type.
Ready for review followup.
Deephaven - Google Chrome_088

Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,311 @@
...
+int_ = DataType(_qst_type_.intType())
+int32 = int_  # make life simple for people who are used to pyarrow
+long_ = DataType(_qst_type_.longType())
+int64 = long_   # make life simple for people who are used to pyarrow
+float_ = DataType(_qst_type_.floatType())
+single = float_   # make life simple for people who are used to NumPy
+float32 = float_  # make life simple for people who are used to pyarrow
+double = DataType(_qst_type_.doubleType())
+float64 = double  # make life simple for people who are used to pyarrow
+string = typeFromJavaClassName('java.lang.String')
```
devinrsmith, line 97: `_qst_type_.stringType()`

Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,311 @@
...
+string_array = typeFromJavaClassName('java.lang.String[]')
+
+@_passThrough
+def _jclassFromType(data_type : DataType):
+    if data_type is None:
+        return None
+    try:
+        jclass = data_type.clazz()
+    except Exception as e:
+        raise Exception("Could not get java class type from " + str(data_type)) from e
```
devinrsmith, line 123: Have you seen this happen? I think `qst.type.Type#clazz` should always return successfully.

Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,311 @@
...
+
+# For more involved types, you can always use the string representation
+# of the Java class (Class.getName()) to get a python type for it.
+@_passThrough
+def typeFromJavaClassName(name : str):
+    """
+    Get the column data type for the corresponding Java type string reprensentation
+    The string provided should match the output in Java for Class.getName()
+    for a class visible to the main ClassLoader in the Deephaven engine in use.
+    """
```
devinrsmith, line 74: These comments are not correct. `String[].class.getName` is *not* "java.lang.String[]". `name` here is a Deephaven-specific "pretty" version.

Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,311 @@
...
+single = float_   # make life simple for people who are used to NumPy
+float32 = float_  # make life simple for people who are used to pyarrow
+double = DataType(_qst_type_.doubleType())
+float64 = double  # make life simple for people who are used to pyarrow
+string = typeFromJavaClassName('java.lang.String')
+bigdecimal = typeFromJavaClassName('java.math.BigDecimal')
+stringset =  typeFromJavaClassName('io.deephaven.db.tables.libs.StringSet')
+datetime = DataType(_qst_type_.instantType())
+
+byte_array = typeFromJavaClassName('byte[]')
```
devinrsmith, line 102: We can go from a non-array type to an array type like `byte.arrayType()`

Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,311 @@
...
+    pass
+
+
+# Represents a Deephaven column data type.
+DataType = NewType('DataType', _qst_type_)
+
+# For more involved types, you can always use the string representation
+# of the Java class (Class.getName()) to get a python type for it.
+@_passThrough
+def typeFromJavaClassName(name : str):
```
devinrsmith, line 69: Do we want to allow python user to define arbitrary types? I might suggest "no", in which case, this can become internal.

Integrations/python/deephaven/Types.py
devinrsmith: Still relevant comment.

Integrations/python/deephaven/Types.py
```diff
@@ -0,0 +1,311 @@
...
+    if col_data is None or len(col_data) < 1:
+        col_header = _qst_col_header_.of(col_name, col_type)
+        return _qst_column_.empty(col_header)
+    jtype = _jpyTypeFromType(col_type)
+    if jtype is None:
+        raise Exception("value for argument 'col_type' " +
+                        str(col_type) + " is not a known data type.")
+    jvalues = jpy.array(jtype, col_data)
+    if _isPrimitive(col_type):
+        return _qst_column_.ofUnsafe(col_name, jvalues)
```
devinrsmith, line 204: This will break for `bool_`, as written.

qst/src/main/java/io/deephaven/qst/column/Column.java
```diff
@@ -49,6 +49,10 @@
         return of(name, Array.of(Type.find(clazz), values));
     }
 
+    public static <T> Column<T> of(String name, Type<T> type, T... values) {
+        return of(name, Array.of(type, values));
+    }
```
devinrsmith, line 54: Nice addition.

Integrations/python/deephaven/Types.py
jcferretti: Fixed.

Integrations/python/deephaven/Types.py
jcferretti: Fixed.

Integrations/python/deephaven/Types.py
jcferretti: Done.

devinrsmith (dismissed):

cpwright (dismissed):

cpwright (approved):

PR #1090, opened 2021-08-19
Add support for converting stream tables to append from python.

jcferretti:


DB/src/main/java/io/deephaven/db/v2/StreamTableTools.java
```diff
@@ -26,7 +26,7 @@
      * @param streamTable the input stream table
      * @return an append-only in-memory table representing all data encountered in the stream
      */
-    public static Table streamToAppendOnlyTable(Table streamTable) {
+    public static Table streamTableToAppendTable(Table streamTable) {
```
rcaudy, line 29: The old name was better.

rcaudy (approved):

PR #1092, opened 2021-08-20
Properly use liveness API for dynamic non-refreshing nodes in SessionState exports

nbauernfeind:
The query:
```groovy
t = emptyTable(10000).update("I = ii")
```

Would display the first page, but any viewport changes would throw an exception due to incorrect usage of LivenessReferent API. It was easily triggered using SSL, could not reproduce using the grpc-web proxy.

### DIFF
--- a/grpc-api/src/main/java/io/deephaven/grpc_api/arrow/ArrowFlightUtil.java
+++ b/grpc-api/src/main/java/io/deephaven/grpc_api/arrow/ArrowFlightUtil.java
@@ -448,7 +448,7 @@ public class ArrowFlightUtil {
                     updateIntervalMs = DEFAULT_UPDATE_INTERVAL_MS;
                 }
                 bmp = table.getResult(operationFactory.create(table, updateIntervalMs));
-                if (DynamicNode.notDynamicOrIsRefreshing(bmp)) {
+                if (bmp.isRefreshing()) {
                     manage(bmp);
                 }
             } else {
--- a/grpc-api/src/main/java/io/deephaven/grpc_api/arrow/ArrowFlightUtil.java
+++ b/grpc-api/src/main/java/io/deephaven/grpc_api/arrow/ArrowFlightUtil.java
@@ -18,7 +18,10 @@ import io.deephaven.configuration.Configuration;
 import io.deephaven.datastructures.util.CollectionUtil;
 import io.deephaven.db.tables.Table;
 import io.deephaven.db.util.LongSizedDataStructure;
+import io.deephaven.db.util.liveness.Liveness;
+import io.deephaven.db.util.liveness.LivenessReferent;
 import io.deephaven.db.util.liveness.SingletonLivenessManager;
+import io.deephaven.db.v2.DynamicNode;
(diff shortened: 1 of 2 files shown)


grpc-api/src/main/java/io/deephaven/grpc_api/arrow/ArrowFlightUtil.java
```diff
@@ -445,7 +448,9 @@ private synchronized void onExportResolved(final SessionState.ExportObject<Objec
                     updateIntervalMs = DEFAULT_UPDATE_INTERVAL_MS;
                 }
                 bmp = table.getResult(operationFactory.create(table, updateIntervalMs));
-                manage(bmp);
+                if (DynamicNode.notDynamicOrIsRefreshing(bmp)) {
```
rcaudy, line 451: This can just be `bmp.isRefreshing()`; bmp is a `DynamicNode`, not some unknown object.

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionState.java
```diff
@@ -526,8 +536,17 @@ private synchronized void setDependencies(final List<ExportObject<?>> parents) {
             this.parents = parents;
             dependentCount = parents.size();
             parents.stream().filter(Objects::nonNull).forEach(this::manage);
-        }
 
+            if (log.isDebugEnabled()) {
+                final Exception e = new RuntimeException();
+                final LogEntry entry = log.debug().append(e).append("\n").append(session.logPrefix).append("export '").append(logIdentity)
```
rcaudy, line 542: `append("\n")` -> `nl()`

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionState.java
```diff
@@ -785,6 +804,11 @@ private void doExport() {
                 } finally {
                     shouldLog = QueryPerformanceRecorder.getInstance().endQuery();
                 }
+
+                if (isNonExport()) {
```
rcaudy, line 808: Why not put this in the `finally`?

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionState.java
```diff
@@ -448,6 +451,8 @@ public static boolean isExportStateTerminal(final ExportNotification.State state
      * @apiNote Non-exports do not publish state changes.
      */
     public final static class ExportObject<T> extends LivenessArtifact {
+        private final Exception creationTag = log.isDebugEnabled() ? new Exception("creationTag") : null;
```
rcaudy, line 454: What is this used for? I don't see anything.

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionState.java
rcaudy: and `append("\n\t")` -> `nl().append('\t')`

rcaudy (approved):

rcaudy (approved):

PR #1111, opened 2021-08-23
Support bidirectional streaming on websockets

### LINKED ISSUE #834: very slow execute response from simple commands from web ui
May or may not have to do with autocomplete. Have repro'd in python and groovy.

python:
```python
print([])
```

groovy:
```groovy
println([])
```

It may have to do w/ specific timing around autocomplete, or backespacing while typing `[]`. But, on occasion, requests will take 8-60 seconds.

### LINKED ISSUE #929: Autocompletion triggers "Internal Error [...] Requirement failed: required offset <= command.length()."
While in the middle of typing a command I get this exception to show up in the web UI:

```
15:57:17.138 ERROR heduler-Concurrent-1 | i.d.g.s.SessionState      | Internal Error '46f08ee3-f4ae-4255-ae9a-63765f34a2c4' io.deephaven.base.verify.RequirementFailure: Requirement failed: required offset <= command.length().
        at io.deephaven.base.verify.Require.fail(Require.java:95)
        at io.deephaven.base.verify.Require.requirement(Require.java:126)
        at io.deephaven.base.verify.Require.requirement(Require.java:131)
        at io.deephaven.lang.completion.CompletionRequest.<init>(CompletionRequest.java:57)
        at io.deephaven.lang.completion.CompletionRequest.<init>(CompletionRequest.java:48)
        at io.deephaven.lang.completion.ChunkerCompleter.runCompletion(ChunkerCompleter.java:110)
        at io.deephaven.grpc_api.console.ConsoleServiceGrpcImpl.lambda$null$24(ConsoleServiceGrpcImpl.java:324)
        at io.deephaven.grpc_api.session.SessionState$ExportBuilder.lambda$submit$1(SessionState.java:1166)
        at io.deephaven.grpc_api.session.SessionState$ExportObject.doExport(SessionState.java:759)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at io.deephaven.grpc_api.runner.DeephavenApiServerModule$ThreadFactory.lambda$newThread$0(DeephavenApiServerModule.java:143)
        at java.lang.Thread.run(Thread.java:748)
```

James believes this is related to an out-of-order issue with autocomplete requests, per slack exchange here:
https://illumon.slack.com/archives/G01K6UHJ00K/p1627502650324000

### LINKED ISSUE #1049: JS API bidi stream abstraction
When we are using websockets for grpc, we can actually support bidi streams, but when using ssl we will use http2/fetch we have to fall back to the server-stream and unary pair, with barrage wrapper to handle sequencing. The client should have an abstraction to let doput/subscription/autocomplete not care what the underlying transport is, but handle sending and receiving in a consistent way.

niloc132:
Fetch/h2-based grpc-web clients do not support bidirectional streaming at this time, but the websocket client we are using from https://github.com/improbable-eng/grpc-web/ does. This lets us keep a single websocket open instead of making many calls. The client has support now to abstract how a bidirectional stream is created, either by a single websocket, or by a pair of fetch calls.

Additionally, autocomplete was seeing problems that appeared related to the server-side grpc handling payloads out of order. These methods have been consolidated into a single bidirectional stream to ensure ordering is sane. The server now has support to only implement a bidirectional stream once, and mark the "open"/"next" rpc methods as part of that, and automatically have their requests enqueued to be handled serially.

Other improvements got slipped into this work as well, such as only interacting with autocomplete from the gRPC service and never from DB, so that dependency was cut (as well as the dependency on gRPC utils, which only existed to get jsr305 annotations into various projects).

The client is incomplete in this patch, we plan on continuing this work as part of #730.

Fixes #929 
Fixes #834 
Fixes #1049


grpc-api/src/main/java/io/deephaven/grpc_api/util/GrpcServiceOverrideBuilder.java
```diff
@@ -175,33 +212,80 @@ public ServerServiceDefinition build() {
...
+
+                final SessionState.ExportObject<BrowserStream<ReqT>> browserStream =
+                        session.getExport(streamData.getRpcTicket());
+
+                session.nonExport()
+                        .require(browserStream)
+                        .onError(responseObserver::onError)
+                        .submit(() -> {
+                            browserStream.get().onMessageReceived(request, streamData);
+                            responseObserver.onNext(null);//TODO simple response payload
```
nbauernfeind, line 269: Perhaps your factory could also give you the "default instance" of a NextRespT?

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -1069,15 +1076,17 @@ private void flush() {
                 BarrageSubscriptionRequest.addTicket(subscriptionReq, tableTicketOffset);
                 subscriptionReq.finish(BarrageSubscriptionRequest.endBarrageSubscriptionRequest(subscriptionReq));
 
-                Uint8Array rpcTicket = config.newTicket();
-
                 FlightData request = new FlightData();
                 //TODO make sure we can set true on halfClose before commit
```
nbauernfeind, line 1080: drop TODO

grpc-api/src/main/java/io/deephaven/grpc_api/browserstreaming/BrowserStream.java
```diff
@@ -0,0 +1,263 @@
...
+            public void onCancel() {
+                StatusRuntimeException canceled = GrpcUtil.statusRuntimeException(Code.CANCELLED, "Stream canceled on the server");
+                GrpcUtil.safelyExecute(() -> {
+                    responseObserver.onError(canceled);
+                });
+
+                //TODO is this necessary or does the above result in this being called automatically?
+                GrpcUtil.safelyExecute(() -> {
+                    requestObserver.onError(canceled);
+                });
```
nbauernfeind, line 80: If this is truly the onCancel from the ServerStream observer, then you will be unable to send a message to the observer. In either case this onError call is duplicated.

grpc-api/src/main/java/io/deephaven/grpc_api/browserstreaming/BrowserStream.java
```diff
@@ -0,0 +1,263 @@
...
+            try {
+                marshaller.onMessageReceived(message);
+            } catch (final RuntimeException e) {
+                onError(e);
+                return;
+            }
+
+            synchronized (this) {
+                if (mode == Mode.IN_ORDER) {
+                    message = pendingSeq == null ? null : pendingSeq.top().getMessage();
```
nbauernfeind, line 198: As discussed this NPEs; and you wanted to make it pretty again. Maybe stretching the logic out like this will make it clean again:

```java
if (pendingSeq == null) {
  break;
}
Message<T> topOfQueue = pendingSeq.top();
if (topOfQueue == null) {
  break;
}
// etc
```

web/client-api/src/main/java/io/deephaven/web/client/api/barrage/stream/BiDiStream.java
```diff
@@ -0,0 +1,226 @@
...
+        public void send(T payload) {
+            if (responseStream == null) {
+                responseStream = responseStreamFactory.apply(payload);
+                pending.forEach((p0, p1, p2) -> {
+                    p0.apply(responseStream);
+                    return null;
+                });
+                pending.length = 0;
+            } else {
+                //TODO handle failure of this call
```
nbauernfeind, line 171: should this (and the one below) get the deep-retry issue number (assuming that we wouldn't be fixing all of these as part of reconnect)?

nbauernfeind (changes requested):

grpc-api/src/main/java/io/deephaven/grpc_api/console/ConsoleServiceGrpcImpl.java
```diff
@@ -253,106 +255,91 @@ public void fetchTable(FetchTableRequest request, StreamObserver<ExportedTableCr
...
-                        scriptSession.getParser().open(doc.getText(), doc.getUri(), Integer.toString(doc.getVersion()));
-                        safelyExecute(() -> {
-                            responseObserver.onNext(OpenDocumentResponse.getDefaultInstance());
-                            responseObserver.onCompleted();
-                        });
-                    });
+    private CompletionParser ensureParserForSession(SessionState session) {
+        return parsers.computeIfAbsent(session, s -> {
+            CompletionParser parser = new CompletionParser();
+            s.addOnCloseCallback(parser);
```
nbauernfeind, line 261: This on close callback also needs to remove from the parsers map.

grpc-api/src/main/java/io/deephaven/grpc_api/arrow/ArrowFlightUtil.java
```diff
@@ -301,7 +295,7 @@ public void onCompleted() {
                     // transfer ownership to submit's liveness scope, drop our extra reference
                     resultTable.manageWithCurrentScope();
                     resultTable.dropReference();
-                    GrpcUtil.safelyExecute(observer::onCompleted);
+                    GrpcUtil.safelyExecuteLocked(observer, observer::onCompleted);
```
nbauernfeind, line 298: Hmm. I forgot about this method. We should have `onErrorSynchronized` and `safelyExecuteLocked` re-use the same suffix. I don't care which; but `safelyExecuteSynchronized` is probably more clear.

nbauernfeind (changes requested):

grpc-api/src/main/java/io/deephaven/grpc_api/util/GrpcServiceOverrideBuilder.java
```diff
@@ -37,65 +41,41 @@ private void addMethod(ServerServiceDefinition.Builder builder) {
...
-        this.serviceName = serviceName;
     }
 
     public static GrpcServiceOverrideBuilder newBuilder(ServerServiceDefinition baseDefinition, String serviceName) {
-        return new GrpcServiceOverrideBuilder(baseDefinition, serviceName);
+        return new GrpcServiceOverrideBuilder(baseDefinition);
     }
 
-    public <ReqT, RespT> GrpcServiceOverrideBuilder override(MethodDescriptor<ReqT, RespT> method, ServerCalls.BidiStreamingMethod<ReqT, RespT> handler) {
+    private <ReqT, RespT> GrpcServiceOverrideBuilder override(MethodDescriptor<ReqT, RespT> method, ServerCalls.BidiStreamingMethod<ReqT, RespT> handler) {
```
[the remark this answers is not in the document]
niloc132: Should I restore the deleted one too? I was trying to reduce the "hmm, which one do i want" surface area, but if we want these still in use (none are used outside this class), we should probably leave them all?

grpc-api/src/main/java/io/deephaven/grpc_api/util/GrpcServiceOverrideBuilder.java
```diff
@@ -175,33 +212,80 @@ public ServerServiceDefinition build() {
...
+                    throw GrpcUtil.statusRuntimeException(Code.INVALID_ARGUMENT, "no x-deephaven-stream headers, cannot handle open request");
+                }
+
+                BrowserStream<ReqT> browserStream = factory.create(session, responseObserver);
+                browserStream.onMessageReceived(request, streamData);
+
+                if (!streamData.isHalfClose()) {
+                    // if this isn't a half-close, we should export it for later calls - if it is, the client won't send more messages
+                    session.newExport(streamData.getRpcTicket())
+                            .onError(responseObserver::onError)
```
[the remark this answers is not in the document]
niloc132: As discussed, latest commit gets close, we're going to make it harder to do it wrong by always locking on something for onError. Will audit existing calls and put up that commit tomorrow.

grpc-api/src/main/java/io/deephaven/grpc_api/util/GrpcServiceOverrideBuilder.java
niloc132: Not necessary right? If we deliberately send null, zero bytes are sent, and the receiver marshaller will use the default value?

Need to drop the TODO though

grpc-api/src/main/java/io/deephaven/grpc_api/browserstreaming/BrowserStream.java
niloc132: onCancel() is only called by close(), which can only happen on session expiry (or possibly explicitly close of the rpc ticket? that would be weird imho...). Instead we should probably rename this method (and test as the TODO says)?

grpc-api/src/main/java/io/deephaven/grpc_api/browserstreaming/BrowserStream.java
niloc132: LMK what you think of the patch I put up compared to your suggested edit above

grpc-api/src/main/java/io/deephaven/grpc_api/console/ConsoleServiceGrpcImpl.java
niloc132: No, this is deliberate (and I'll add a comment before resolving this) - we want to leave the parser and its docs open. The websocket times out (currently after 30s) to avoid eating up a free websocket if there is no activity from the user, but we must maintain the state of the currently open documents.

grpc-api/src/main/java/io/deephaven/grpc_api/arrow/ArrowFlightUtil.java
niloc132: This one I doubt more than the others... should we also make this force a lock? Most of these are very short and involve the observer (being in GrpcUtils and all), thoughts?

grpc-api/src/main/java/io/deephaven/grpc_api/util/GrpcServiceOverrideBuilder.java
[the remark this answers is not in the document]
niloc132: Fixed, PTAL

grpc-api/src/main/java/io/deephaven/grpc_api/console/ConsoleServiceGrpcImpl.java
niloc132: Never mind, I see what you mean.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/console.proto
```diff
@@ -52,7 +65,7 @@ message StartConsoleRequest {
 message StartConsoleResponse {
     io.deephaven.proto.backplane.grpc.Ticket result_id = 1;
 
-    // TOOD i think these are unused without an init script thingie
+    // TODO i think these are unused without an init script thingie
```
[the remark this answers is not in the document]
niloc132: I don't think this applies any more once app mode is merged, since the user will instead be using a stream to get variables that already existed from the console? Or does the app+console mode not address this?

PR #1155, opened 2021-08-30
Initial Kafka python tests.

jcferretti:
This PR has a single test, intended to verify/agree on the way to setup things in gradle for the integration with redpanda.
More tests will be added in a separate PR later, which allows the review burden and the bandwidth requirement for different people to be controlled.


niloc132 (approved):

devinrsmith:
I think we need to add in hooks for
```
composeDown task stops the application and removes the containers, only if 'stopContainers' is set to 'true' (default value).
```?

jcferretti:
Answering Devin above, we discussed via slack:

From the Usage section in the [README.md](https://github.com/avast/gradle-docker-compose-plugin), in the example code:
dockerCompose.isRequiredBy(test) // hooks 'dependsOn composeUp' and 'finalizedBy composeDown', and exposes environment variables and system properties (if possible)

PR #1160, opened 2021-08-30
Add python documentation for Kafka integration.

jcferretti:


jmao-denver (commented):
Looks good to me. Only thing I think would be nice to have is to indicate if a method could raise an exception so a try/except is appropriate when calling it.

jmao-denver (approved):

PR #1166, opened 2021-08-30
ScriptSession#setVariable require subclasses to notify old vs new Value

nbauernfeind:
The pattern of invoking `super.setVariable` is brittle, and the listener assumes the change(s) already happened. This fixes an error call to `getVariable` when the application field listener was trying to process a create via `ConsoleService#bindTableToVariable`.

### DIFF
--- a/DB/src/main/java/io/deephaven/db/util/AbstractScriptSession.java
+++ b/DB/src/main/java/io/deephaven/db/util/AbstractScriptSession.java
@@ -172,14 +172,16 @@ public abstract class AbstractScriptSession extends LivenessScope implements Scr
         }
     }
 
-    @Override
-    public synchronized void setVariable(String name, Object value) {
-        if (changeListener != null) {
-            Changes changes = new Changes();
-            applyVariableChangeToDiff(changes, name, getVariable(name, null), value);
-            if (!changes.isEmpty()) {
-                changeListener.onScopeChanges(this, changes);
-            }
+    protected synchronized void notifyVariableChange(String name, @Nullable Object oldValue,
+            @Nullable Object newValue) {
+        if (changeListener == null) {
+            return;
+        }
+
+        Changes changes = new Changes();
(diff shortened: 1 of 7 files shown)


DB/src/main/java/io/deephaven/db/util/AbstractScriptSession.java
```diff
@@ -172,14 +172,16 @@ public Changes evaluateScript(Path scriptPath) {
...
 
-    @Override
-    public synchronized void setVariable(String name, Object value) {
-        if (changeListener != null) {
-            Changes changes = new Changes();
-            applyVariableChangeToDiff(changes, name, getVariable(name, null), value);
-            if (!changes.isEmpty()) {
-                changeListener.onScopeChanges(this, changes);
-            }
+    protected synchronized void notifyVariableChange(String name, @Nullable Object oldValue,
```
niloc132, line 175: would be nice to propagate these to applyVariableChangesToDiff

DB/src/main/java/io/deephaven/db/util/NoLanguageDeephavenSession.java
```diff
@@ -80,9 +80,10 @@ public boolean hasVariableName(String name) {
     }
 
     @Override
-    public void setVariable(String name, Object value) {
-        super.setVariable(name, value);
-        variables.put(name, value);
+    public void setVariable(String name, @Nullable Object newValue) {
+        Object oldValue = getVariable(name, null);
```
niloc132, line 84: ditto this with @Nullable in getVariable, esp since it delegates to getVariable with no `newValue`, but will throw, so the param isn't "optional if null" as one might suspect...

niloc132 (approved):
Annotation requests, and it seems silly to let all these changes propagate setVariable -> notifyVariableChanges -> applyVariableChangeToDIff and in the third condition of the last link in the chain check if they are actually the same instance

niloc132 (approved):

PR #1168, opened 2021-08-31
Added python tests for every Kafka spec type.  Closes #1123.

### LINKED ISSUE #1123: Add unit tests for KafkaTools
If done from python it will exercise both python and java.

jcferretti:


Integrations/python/test/testKafkaTools.py
```diff
@@ -29,6 +30,15 @@ class TestTableTools(unittest.TestCase):
     def setUpClass(self):
         kt._defineSymbols()
```
jmao-denver, line 31: this is redundant, in fact, it reduces the coverage because 1) this is not how a user would use the module, 2) we lose a chance to test if _defineSymbols() get called automatically behind the scene

Integrations/python/test/testKafkaTools.py
```diff
@@ -58,14 +68,99 @@ def testSimple(self):
...
+            {'bootstrap.servers' : 'redpanda:29092',
+             'schema.registry.url' : 'http://registry:8080/api/ccompat'},
+            'share_price',
+            key = kt.IGNORE,
+            value = kt.avro('share_price_record', schema_version='1'),
+            table_type='append'
+        )
+
+        self.assertIsNotNone(t)
+
```
jmao-denver, line 152: IIRC the constructor would throw an exception when some of the input parameters are invalid, so you have chosen to use exception to report an error. Thus the constructor should never return None. Is this check unnecessary?

Integrations/python/test/testKafkaTools.py
```diff
@@ -58,14 +68,99 @@ def testSimple(self):
...
+                  ('Qty', dh.int_) ],
+                mapping = {
+                    'jsymbol' : 'Symbol',
+                    'jside' : 'Side',
+                    'jprice' : 'Price',
+                    'jqty' : 'Qty'}
+            ),
+            table_type = 'append'
+        )
+        self.assertIsNotNone(t)
```
jmao-denver, line 99: same question as below

jmao-denver (changes requested):

Integrations/python/test/testKafkaTools.py
```diff
@@ -58,14 +68,99 @@ def testSimple(self):
...
-        self.assertEquals(dh.int_.clazz(), cols[0].getDataType())
-        self.assertEquals("KafkaOffset", cols[1].getName())
-        self.assertEquals(dh.long_.clazz(), cols[1].getDataType())
-        self.assertEquals("KafkaTimestamp", cols[2].getName())
-        self.assertEquals(dh.datetime.clazz(), cols[2].getDataType())
+        self._assertCommonCols(cols)
         self.assertEquals("Price", cols[3].getName())
         self.assertEquals(dh.double.clazz(), cols[3].getDataType())
         
         del t
```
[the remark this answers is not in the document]
jcferretti: Other tests do it, and I looked at them as a template/example.  I am guessing DH table objects are heavy.

Integrations/python/test/testKafkaTools.py
jcferretti: It is, it had to do with an earlier version before docker compose was wired. Ty

Integrations/python/test/testKafkaTools.py
jcferretti: Good points.  Solid.

Integrations/python/test/testKafkaTools.py
```diff
@@ -58,14 +68,99 @@ def testSimple(self):
...
+        self.assertEquals(dh.string.clazz(), cols[3].getDataType())
+        self.assertEquals("Side", cols[4].getName())
+        self.assertEquals(dh.string.clazz(), cols[4].getDataType())
+        self.assertEquals("Price", cols[5].getName())
+        self.assertEquals(dh.double.clazz(), cols[5].getDataType())
+        self.assertEquals("Qty", cols[6].getName())
+        self.assertEquals(dh.int_.clazz(), cols[6].getDataType())
+        
+        del t
+
```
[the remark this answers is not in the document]
jcferretti: Removed, see above.

Integrations/python/test/testKafkaTools.py
```diff
@@ -58,14 +68,99 @@ def testSimple(self):
...
+        self.assertEquals("Symbol", cols[3].getName())
+        self.assertEquals(dh.string.clazz(), cols[3].getDataType())
+        self.assertEquals("Side", cols[4].getName())
+        self.assertEquals(dh.string.clazz(), cols[4].getDataType())
+        self.assertEquals("Qty", cols[5].getName())
+        self.assertEquals(dh.int_.clazz(), cols[5].getDataType())
+        self.assertEquals("Price", cols[6].getName())
+        self.assertEquals(dh.double.clazz(), cols[6].getDataType())
+        
+        del t
```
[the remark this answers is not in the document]
jcferretti: Removed, see above.

Integrations/python/test/testKafkaTools.py
jcferretti: Removed

Integrations/python/test/testKafkaTools.py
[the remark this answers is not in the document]
jmao-denver: right, I used the term GC loosely in this case.

jmao-denver (approved):

Issue #1167, opened 2021-08-31
Provide better error message to user when table API request's result_id is missing


nbauernfeind:
Currently if the user forgets to set a result ticket for a one-shot table rpc, they get an awfully obscure error:
```
grpc-api_1    | 2021-08-31T02:02:55.007Z | c-nio-worker-ELG-3-2 | ERROR | .g.t.TableServiceGrpcImpl | io.grpc.StatusRuntimeException: FAILED_PRECONDITION: Cannot parse ticket: found 0x (hex)
grpc-api_1    |         at io.grpc.Status.asRuntimeException(Status.java:535)
grpc-api_1    |         at io.grpc.protobuf.StatusProto.toStatusRuntimeException(StatusProto.java:52)
grpc-api_1    |         at io.deephaven.grpc_api.util.Exceptions.statusRuntimeException(Exceptions.java:15)
grpc-api_1    |         at io.deephaven.grpc_api.util.ExportTicketHelper.ticketToExportIdInternal(ExportTicketHelper.java:267)
grpc-api_1    |         at io.deephaven.grpc_api.util.ExportTicketHelper.ticketToExportId(ExportTicketHelper.java:66)
grpc-api_1    |         at io.deephaven.grpc_api.session.SessionState.newExport(SessionState.java:330)
grpc-api_1    |         at io.deephaven.grpc_api.table.TableServiceGrpcImpl.lambda$oneShotOperationWrapper$15(TableServiceGrpcImpl.java:406)
grpc-api_1    |         at io.deephaven.grpc_api.util.GrpcUtil.rpcWrapper(GrpcUtil.java:33)
grpc-api_1    |         at io.deephaven.grpc_api.table.TableServiceGrpcImpl.oneShotOperationWrapper(TableServiceGrpcImpl.java:393)
grpc-api_1    |         at io.deephaven.grpc_api.table.TableServiceGrpcImpl.fetchTable(TableServiceGrpcImpl.java:252)
[ ... ]
```

The batch operation allows the user to do this on purpose, so that they may create what Devin calls "tainted" nodes in the DAG that cannot guaranteed to be recovered. This should just be a better error message "hey, I can do this, but don't you want me to write it some place?".

Issue #997, opened 2021-08-06
integrate with flight auth


nbauernfeind:
We were able to transition to DoExchange and get rid of the BarrageService without actually making Flight's Auth work. We thought that was a good opportunity to split the project off (as the other app_metadata changes have been large enough).

I came across this document describing the intended plan for flight auth the last time it seems to have changed:
https://docs.google.com/document/d/1k3hdbGeE89jiMZ2oKpoIZEs7HAQmVs_RWOmU7fTpLjQ/edit#heading=h.99wjos12oqt

Authorization configuration/enforcement is part of #2563.

PR #1174, opened 2021-08-31
Improve Ticket/Descriptor error messages for all gRPC usages

### LINKED ISSUE #1153: TicketRouter OOB on empty ticket
The ticket resolver assumes that a ticket has at least one byte:

```java
getResolver(ticket.get(ticket.position()))
```

This causes an OOB when the ticket is empty:

```
grpc-api_1    | 2021-08-28T22:57:31.619Z | c-nio-worker-ELG-3-4 | ERROR | .c.ConsoleServiceGrpcImpl | Internal Error '4233fe6f-05b4-463a-8201-9d600b063ef7' java.lang.IndexOutOfBoundsException: null
grpc-api_1    |         at java.nio.Buffer.checkIndex(Buffer.java:545)
grpc-api_1    |         at java.nio.HeapByteBuffer.get(HeapByteBuffer.java:142)
grpc-api_1    |         at io.deephaven.grpc_api.session.TicketRouter.resolve(TicketRouter.java:51)
grpc-api_1    |         at io.deephaven.grpc_api.session.TicketRouter.resolve(TicketRouter.java:79)
grpc-api_1    |         at io.deephaven.grpc_api.console.ConsoleServiceGrpcImpl.lambda$executeCommand$9(ConsoleServiceGrpcImpl.java:163)
grpc-api_1    |         at io.deephaven.grpc_api.util.GrpcUtil.rpcWrapper(GrpcUtil.java:33)
grpc-api_1    |         at io.deephaven.grpc_api.console.ConsoleServiceGrpcImpl.executeCommand(ConsoleServiceGrpcImpl.java:159)
```

We should probably throw the same type of error that a real resolver would throw when it can't resolve a ticket.

We should also try to add higher-level checking on the incoming message. Ie, "Must provide non-empty consoleId for executeCommand".

nbauernfeind:
Fixes #1153, #1167

### DIFF
--- a/grpc-api/src/main/java/io/deephaven/grpc_api/appmode/ApplicationTicketResolver.java
+++ b/grpc-api/src/main/java/io/deephaven/grpc_api/appmode/ApplicationTicketResolver.java
@@ -108,13 +108,15 @@ public class ApplicationTicketResolver extends TicketResolverBase {
     @Override
     public <T> SessionState.ExportBuilder<T> publish(
             SessionState session, ByteBuffer ticket, final String logId) {
-        throw new UnsupportedOperationException("applications cannot be published to");
+        throw GrpcUtil.statusRuntimeException(Code.FAILED_PRECONDITION,
+                "Could not publish '" + logId + "': application tickets cannot be published to");
     }
 
     @Override
     public <T> SessionState.ExportBuilder<T> publish(
             final SessionState session, final Flight.FlightDescriptor descriptor, final String logId) {
-        throw new UnsupportedOperationException("applications cannot be published to");
+        throw GrpcUtil.statusRuntimeException(Code.FAILED_PRECONDITION,
+                "Could not publish '" + logId + "': application flight descriptors cannot be published to");
     }
 
     @Override
--- a/grpc-api/src/main/java/io/deephaven/grpc_api/appmode/ApplicationTicketResolver.java
+++ b/grpc-api/src/main/java/io/deephaven/grpc_api/appmode/ApplicationTicketResolver.java
@@ -51,47 +51,54 @@ public class ApplicationTicketResolver extends TicketResolverBase {
     }
 
     @Override
-    public <T> SessionState.ExportObject<T> resolve(final @Nullable SessionState session, final ByteBuffer ticket) {
-        return resolve(appFieldIdFor(ticket));
+    public <T> SessionState.ExportObject<T> resolve(
+            final @Nullable SessionState session, final ByteBuffer ticket, final String logId) {
+        return resolve(appFieldIdFor(ticket, logId), logId);
     }
 
     @Override
-    public <T> SessionState.ExportObject<T> resolve(final @Nullable SessionState session,
-            final Flight.FlightDescriptor descriptor) {
-        return resolve(appFieldIdFor(descriptor));
+    public <T> SessionState.ExportObject<T> resolve(
+            final @Nullable SessionState session, final Flight.FlightDescriptor descriptor, final String logId) {
+        return resolve(appFieldIdFor(descriptor, logId), logId);
     }
 
-    private <T> SessionState.ExportObject<T> resolve(final AppFieldId id) {
+    private <T> SessionState.ExportObject<T> resolve(final AppFieldId id, final String logId) {
         if (id.app == null) {
-            throw new IllegalStateException("Cannot resolve field without knowing the application");
+            throw GrpcUtil.statusRuntimeException(Code.FAILED_PRECONDITION,
+                    "Could not resolve '" + logId + "': field '" + getLogNameFor(id)
+                            + "' does not belong to an application");
         }
         final Field<Object> field = id.app.getField(id.fieldName);
         if (field == null) {
-            throw GrpcUtil.statusRuntimeException(Code.NOT_FOUND, "ticket not found: " + getLogNameFor(id));
+            throw GrpcUtil.statusRuntimeException(Code.NOT_FOUND,
+                    "Could not resolve '" + logId + "': field '" + getLogNameFor(id) + "' not found");
         }
         // noinspection unchecked
         return SessionState.wrapAsExport((T) field.value());
     }
 
     @Override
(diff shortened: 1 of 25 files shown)


devinrsmith (approved):
Looks tedious, mostly mechanical. Would probably be good to break out strictly mechanical / refactoring work  (`logId`) from the logical changes (catching when consoleId is empty, switch case on TableReference, etc) in the future.

grpc-api/src/main/java/io/deephaven/grpc_api/appmode/ApplicationTicketResolver.java
```diff
@@ -51,66 +51,75 @@ public synchronized void onApplicationLoad(final ApplicationState app) {
...
         }
 
         return SessionState.wrapAsExport(info);
     }
 
     @Override
-    public <T> SessionState.ExportBuilder<T> publish(SessionState session, ByteBuffer ticket) {
+    public <T> SessionState.ExportBuilder<T> publish(
+            SessionState session, ByteBuffer ticket, final String logId) {
         throw new UnsupportedOperationException("applications cannot be published to");
```
niloc132, line 111: any chance these should get a better message, or is this only reachable via server-side machinery?

grpc-api/src/main/java/io/deephaven/grpc_api/appmode/ApplicationTicketResolver.java
```diff
@@ -51,66 +51,75 @@ public synchronized void onApplicationLoad(final ApplicationState app) {
...
         // noinspection unchecked
         return SessionState.wrapAsExport((T) field.value());
     }
 
     @Override
-    public SessionState.ExportObject<Flight.FlightInfo> flightInfoFor(final @Nullable SessionState session,
-            final Flight.FlightDescriptor descriptor) {
-        final AppFieldId id = appFieldIdFor(descriptor);
+    public SessionState.ExportObject<Flight.FlightInfo> flightInfoFor(
+            final @Nullable SessionState session, final Flight.FlightDescriptor descriptor, final String logId) {
```
niloc132, line 82: wrong conv, but under what circumstances are we okay with session-less requests?

grpc-api/src/main/java/io/deephaven/grpc_api/session/ExportTicketResolver.java
```diff
@@ -25,27 +25,29 @@ public ExportTicketResolver() {
...
 
     @Override
-    public SessionState.ExportObject<Flight.FlightInfo> flightInfoFor(@Nullable final SessionState session,
-            final Flight.FlightDescriptor descriptor) {
+    public SessionState.ExportObject<Flight.FlightInfo> flightInfoFor(
+            @Nullable final SessionState session, final Flight.FlightDescriptor descriptor, final String logId) {
         if (session == null) {
-            throw GrpcUtil.statusRuntimeException(Code.UNAUTHENTICATED, "No session to search");
+            throw GrpcUtil.statusRuntimeException(Code.UNAUTHENTICATED,
+                    "Could not resolve '" + logId + "': no exports can exist without a session to search");
```
niloc132, line 37: 'search' seems like a weird way to describe this operation, why not just "no exports can exist without a session" or "without a valid session", "active session" etc?

grpc-api/src/main/java/io/deephaven/grpc_api/session/TicketRouter.java
```diff
@@ -210,30 +233,30 @@ public void visitFlightInfo(final @Nullable SessionState session, final Consumer
...
         if (descriptor.getPathCount() <= 0) {
             throw GrpcUtil.statusRuntimeException(Code.FAILED_PRECONDITION,
-                    "Cannot find resolver: flight descriptor does not have route path");
+                    "Could not resolve '" + logId + "': flight descriptor does not have route path");
         }
 
         final String route = descriptor.getPath(0);
         final TicketResolver resolver = descriptorResolverMap.get(route);
         if (resolver == null) {
             throw GrpcUtil.statusRuntimeException(Code.FAILED_PRECONDITION,
```
niloc132, line 258: any chance you want these to be NOT_FOUND instead? FAILED_PRECONDITION isn't inaccurate, but NOT_FOUND could make more sense with nonsense tickets, potentially

niloc132 (approved):

grpc-api/src/main/java/io/deephaven/grpc_api/appmode/ApplicationTicketResolver.java
nbauernfeind: These are only reachable if you manually create an application ticket, and try to call a method that writes to it. Such as DoPut. If you DoPut to an application field ticket - that's when you hit this.

grpc-api/src/main/java/io/deephaven/grpc_api/appmode/ApplicationTicketResolver.java
nbauernfeind: We want to be able to expose any content through Flight. Possibly this feature should be gated by a flag and disabled by default.

grpc-api/src/main/java/io/deephaven/grpc_api/session/TicketRouter.java
nbauernfeind: I feel that NOT_FOUND implies that the ticket is structured correctly, but that it just doesn't exist. I've tried to use FAILED_PRECONDITION if the ticket is messed up or unparseable somehow.

nbauernfeind:
@devinrsmith while that often can make sense -- in this case I was driven by the API change (which you called refactoring, but I'm not sure this is a form of refactoring -- this is new functionality to provide better error messages). The call sites to `resolve` (which I had to fix anyways) were the proper places to identify opportunities for missing/empty protobuf fields.

devinrsmith (dismissed):

grpc-api/src/main/java/io/deephaven/grpc_api/appmode/ApplicationTicketResolver.java
nbauernfeind: I added this comment to #997

devinrsmith (approved):

PR #1191, opened 2021-09-01
Propagate query scope changes even if script throws a RuntimeError

nbauernfeind:


devinrsmith (approved):
We should try to capture this eventually as an integration test of sorts.

PR #1193, opened 2021-09-01
Send UI to Unknown Error if auth token cannot be refreshed after a few tries

### LINKED ISSUE #888: Worker should communicate fatal errors to client APIs / Web UI
**Description**

When using DynamicTableWriter, identical mistakes in Java/Groovy cause very different behaviors.  In Python, the expected error is raised.  In Groovy, a TransientFailure occurs which kills the session, forcing a restart of Docker and the GUI.

The error is caused by giving the wrong data type to a particular column (e.g. a string value to an int column).

**Steps to reproduce**

In Python, run the following code to get a working example using DynamicTableWriter to write a row to a table every second. The table has two columns: one with random integers, and one with random ascii characters:

```python
from deephaven import jpy
import random, string, threading, time

DynamicTableWriter = jpy.get_type("io.deephaven.db.v2.utils.DynamicTableWriter")

tableWriter = DynamicTableWriter(["A", "B"], [jpy.get_type("int"), jpy.get_type("java.lang.String")])

t = tableWriter.getTable()

def thread_func():
    for i in range(10):
        a = random.randint(1, 100)
        b = random.choice(string.ascii_letters)
        tableWriter.logRow(a, b)
        time.sleep(1)

thread = threading.Thread(target = thread_func)
thread.start()
```

Now, to get an error, replace `tableWriter.logRow(a, b)` with `tableWriter.logRow("ABC", b)`.  As expected, the console spits out an error telling you you cant cast a Java string to a Java int.

Now, run the following analogue in Groovy:

```groovy
import io.deephaven.db.v2.utils.DynamicTableWriter

chars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890".toCharArray()

columnNames = ["A", "B"] as String[]
columnTypes = [int.class, char.class] as Class[]

tableWriter = new DynamicTableWriter(columnNames, columnTypes)

t = tableWriter.getTable()

def thread = Thread.start {
    for (int i = 0; i < 10; i++) {
        a = (int)Math.abs(new Random().nextInt() % [100]) + 1
        charIndex = (int)Math.abs(new Random().nextInt() % [62]) + 1
        b = (char)chars[charIndex]
        tableWriter.logRow(a, b)
        sleep(1000)
    }
    return
}
```

Once again, the code writes ten new rows to a table.  Make the same replacement as in Python (`tableWriter.logRow(a, b)` to `tableWriter.logRow("ABC", b)`).  Instead of an error with stack trace, you get a TransientError and nothing telling you that you just wrote the wrong data type to a column.

**Expected results**

An error telling you to not write strings to an integer column (this is the case in Python)

**Actual results**

A TransientFailure (in Groovy only)

**Versions**
 - Deephaven: OSS
 - OS: Big Sur
 - Browser: Chrome
 - Docker: 20.10.6

nbauernfeind:
Fixes #888

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
@@ -163,6 +163,9 @@ public class WorkerConnection {
     private Map<String, JsVariableDefinition> knownFields = new HashMap<>();
     private ResponseStreamWrapper<FieldsChangeUpdate> fieldsChangeUpdateStream;
 
+    private long lastSuccessResponseTime = 0;
+    private static final long GIVE_UP_TIMEOUT_MS = 10_000;
+
     public WorkerConnection(QueryConnectable<?> info, Supplier<Promise<ConnectToken>> authTokenPromiseSupplier) {
         this.info = info;
         this.config = new ClientConfiguration();
@@ -305,8 +308,10 @@ public class WorkerConnection {
 
     public void checkStatus(ResponseStreamWrapper.Status status) {
         // TODO provide simpler hooks to retry auth, restart the stream
+        final long now = System.currentTimeMillis();
         if (status.isOk()) {
             // success, ignore
+            lastSuccessResponseTime = now;
         } else if (status.getCode() == Code.Unauthenticated) {
             // TODO re-create session once?
             // for now treating this as fatal, UI should encourage refresh to try again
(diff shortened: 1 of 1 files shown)


devinrsmith (approved):
I'm not the best reviewer for this. Thought there might be a tight CPU loop, but Nate said there is a DOM timeout buffer. I tested the results before/after this change. Likely follow up from Colin, Bender, Don?

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -163,6 +163,9 @@
     private Map<String, JsVariableDefinition> knownFields = new HashMap<>();
     private ResponseStreamWrapper<FieldsChangeUpdate> fieldsChangeUpdateStream;
 
+    private long lastSuccessResponseTime = 0;
+    private static final long GIVE_UP_TIMEOUT_MS = 10_000;
```
mofojed, line 167: Giving up after 10s? Isn't that a little quick?

PR #1204, opened 2021-09-02
auto doc gen for python client

jmao-denver:
added a pyclientdoc job in CI actions

### DIFF
--- a/.github/workflows/docs-ci.yml
+++ b/.github/workflows/docs-ci.yml
@@ -120,8 +120,7 @@ jobs:
         with:
           switches: -avzr --delete
           path: pyclient/docs/build/html
-          remote_path: /core/client-api/python/
-          #remote_path: deephaven-core/pydoc/
+          remote_path: deephaven-core/client-api/python/
           remote_host: ${{ secrets.DOCS_HOST }}
           remote_port: ${{ secrets.DOCS_PORT }}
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-ci.yml
```diff
@@ -101,3 +101,27 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Generate the docs in html
+        run: |
+          pip3 install sphinx sphinx-autodoc-typehints
+          cd pyclient/docs
+          make html
```
devinrsmith, line 115: If we are going to rely on the system python (instead of doing it through gradle), I think we'll want to have more control over the python setup (just like we choose the version of java).

We'll want to use https://github.com/actions/setup-python.

Here are some examples https://docs.github.com/en/actions/guides/building-and-testing-python

devinrsmith (changes requested):

jmao-denver:
After some discussion, we have concluded that the current approach of doing it outside the standardized Gradle + Containerization way will make it harder to run and maintain. @niloc132 will submit a separate PR to utilize the existing sphinx task to cover both the integration Python doc generation and the Python client doc generattion.

PR #1206, opened 2021-09-02
Make docker-compose.yml download example data when run

chipkent:
Make docker-compose.yml download example data when run


containers/groovy-examples/docker-compose.yml
```diff
@@ -41,6 +41,7 @@ services:
     image: ghcr.io/deephaven/examples
     volumes:
       - ./data:/data
+    command: initialize
```
devinrsmith, line 44: I don't see this command on that image...

```
docker run ghcr.io/deephaven/examples initialize
Unknown command 'initialize'
    commands are:
usage: examples_script.sh <command>

        download [<version>] - downloads and mounts all example data
                               gets latest version, unless <version> supplied
        remove - removes all example data
        version - shows current the version
        versions - list available versions
```

devinrsmith:
We need to tag the release for deephaven/examples first, then we can merge this.

devinrsmith (approved):

PR #1210, opened 2021-09-02
Create SessionService#TerminationNotification as best-effort attempt to notify on worker exit

nbauernfeind:
This is follow-up to improve experience in situations like #888, when the worker unexpectedly dies.

Clients subscribe to receive a terminal notification request when the process exits. Included in the message is the best level of detail we can offer describing the error that has just cropped up.

Initial integration with jsapi. The web-client-ui does not quite display this absolutely lovingly. However one can select-all and copy to bring the error to their favorite IDE for analysis.


grpc-api/src/main/java/io/deephaven/grpc_api/console/ConsoleServiceGrpcImpl.java
```diff
@@ -81,6 +90,18 @@ public ConsoleServiceGrpcImpl(final Map<String, Provider<ScriptSession>> scriptT
         if (!scriptTypes.containsKey(WORKER_CONSOLE_TYPE)) {
             throw new IllegalArgumentException("Console type not found: " + WORKER_CONSOLE_TYPE);
         }
+
+        ProcessEnvironment.getGlobalFatalErrorReporter().addInterceptor((message, throwable) -> {
```
rcaudy, line 94: We don't need/want to have `info.notifyConnectionError(status);` here?

grpc-api/src/main/java/io/deephaven/grpc_api/console/ConsoleServiceGrpcImpl.java
```diff
@@ -81,6 +90,18 @@ public ConsoleServiceGrpcImpl(final Map<String, Provider<ScriptSession>> scriptT
...
+            final long now = scheduler.currentTime().getNanos();
+            final LogBufferRecord record = new LogBufferRecord();
+            record.setLevel(LogLevel.ERROR);
+            record.setTimestampMicros(now);
+
+            record.setData(ByteBuffer.wrap(message.getBytes(StandardCharsets.UTF_8)));
+            logBuffer.record(record);
+            record.setData(ByteBuffer.wrap(throwable.getMessage().getBytes(StandardCharsets.UTF_8)));
```
rcaudy, line 102: This pattern seems wrong. Records aren't re-usable in this way, because you'll overwrite the one in the log buffer's circular array.

rcaudy: It's "safe" for your listener, but not for `LogBuffer` itself.

grpc-api/src/main/java/io/deephaven/grpc_api/console/ConsoleServiceGrpcImpl.java
```diff
@@ -402,40 +403,16 @@ public void fetchFigure(FetchFigureRequest request, StreamObserver<FetchFigureRe
         });
     }
 
-    private class LogBufferStreamAdapter implements Closeable, LogBufferRecordListener {
-        private final SessionState session;
+    private static class LogBufferStreamAdapter extends SessionCloseableObserver<LogSubscriptionData>
+            implements LogBufferRecordListener {
         private final LogSubscriptionRequest request;
```
rcaudy, line 408: Why do we care enough to keep track of the request?

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionService.java
```diff
@@ -51,8 +59,53 @@ public SessionService(final Scheduler scheduler, final SessionState.Factory sess
...
         // Protect ourselves from rotation spam, but be loose enough that any reasonable refresh strategy works.
         this.tokenRotateMs = tokenExpireMs / 5;
+
+        ProcessEnvironment.getGlobalFatalErrorReporter().addInterceptor(this::onFatalError);
+    }
+
+    private synchronized void onFatalError(
+            @NotNull String message,
+            @NotNull Throwable throwable,
+            boolean isFromUncaught) {
```
rcaudy, line 69: Forgetting to set this.

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionService.java
```diff
@@ -51,8 +59,53 @@ public SessionService(final Scheduler scheduler, final SessionState.Factory sess
...
+        final TerminationNotificationResponse.Builder builder =
+                TerminationNotificationResponse.newBuilder()
+                        .setAbnormalTermination(true)
+                        .setReason(message);
+
+        builder.addStackTrace(throwable.getMessage());
```
rcaudy, line 75: Should we have the exception class here somewhere?

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionService.java
```diff
@@ -51,8 +59,53 @@ public SessionService(final Scheduler scheduler, final SessionState.Factory sess
...
+        for (final StackTraceElement element : throwable.getStackTrace()) {
+            builder.addStackTrace(element.toString());
+        }
+
+        final TerminationNotificationResponse notification = builder.build();
+        terminationListeners.forEach(listener -> listener.sendMessage(notification));
+        terminationListeners.clear();
+    }
```
rcaudy, line 83: Are we missing a `closeAllSessions()` here?

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/session.proto
```diff
@@ -67,6 +67,12 @@ service SessionService {
    * updated exports. An export id of zero will be sent to indicate all pre-existing exports have been sent.
    */
   rpc ExportNotifications(ExportNotificationRequest) returns (stream ExportNotification) {}
+
+  /*
+   * Receive a best-effort message on-exit indicating why this worker is exiting. Reception of this message cannot be
```
rcaudy, line 72: worker->server?

grpc-api/src/main/java/io/deephaven/grpc_api/runner/DeephavenApiServer.java
```diff
@@ -80,7 +80,7 @@ public static void startMain(PrintStream out, PrintStream err)
 
         // Close outstanding sessions to give any gRPCs closure.
         ProcessEnvironment.getGlobalShutdownManager().registerTask(ShutdownManager.OrderingCategory.MIDDLE,
-                sessionService::closeAllSessions);
+                sessionService::onShutdown);
```
rcaudy, line 83: This suggests to me that onShutdown should be a no-op if we've hit onFatalError.

rcaudy (commented):
I think two of my comments are probably nonsense, since you seem to be relying on a "two stage" effect in the case of fatal error.

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionService.java
rcaudy: I see that we are not.

grpc-api/src/main/java/io/deephaven/grpc_api/runner/DeephavenApiServer.java
rcaudy: Never mind, I see what you did here.

rcaudy (dismissed):

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionService.java
```diff
@@ -51,6 +57,64 @@ public SessionService(final Scheduler scheduler, final SessionState.Factory sess
...
+
+    private static TerminationNotificationResponse.StackTrace transformToProtoBuf(@NotNull final Throwable throwable) {
+        return TerminationNotificationResponse.StackTrace.newBuilder()
+                .setType(throwable.getClass().getSimpleName())
+                .setMessage(throwable.getMessage())
+                .addAllElements(Arrays.stream(throwable.getStackTrace())
+                        .map(StackTraceElement::toString)
```
niloc132, line 92: do you want to toss a limit(500) or something in here in case of stackoverflowexception?

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionServiceGrpcImpl.java
```diff
@@ -222,4 +234,5 @@ public AuthServerInterceptor(final SessionService service) {
             return Contexts.interceptCall(newContext, serverCall, metadata, serverCallHandler);
         }
     }
+
```
niloc132, line 237: formatter oddity, or planned on adding something and changed your mind?

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionServiceGrpcImpl.java
```diff
@@ -13,14 +13,17 @@
 import io.deephaven.io.logger.Logger;
 import io.deephaven.proto.backplane.grpc.*;
 import io.deephaven.util.auth.AuthContext;
+import io.deephaven.util.process.ProcessEnvironment;
```
niloc132, line 16: appears to be unused

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionServiceGrpcImpl.java
```diff
@@ -13,14 +13,17 @@
...
 import io.grpc.*;
 import io.grpc.stub.ServerCallStreamObserver;
 import io.grpc.stub.StreamObserver;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.concurrent.CopyOnWriteArrayList;
```
niloc132, line 26: appears to be unused

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionService.java
niloc132: and should we be at all concerned here about finding a cheaper way to do this, given that something is in the process of blowing up (oom perhaps)? I don't see an obvious way to do so, but figured it was worth asking...

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionService.java
```diff
@@ -51,6 +57,64 @@ public SessionService(final Scheduler scheduler, final SessionState.Factory sess
...
+
+        final TerminationNotificationResponse notification = builder.build();
+        terminationListeners.forEach(listener -> listener.sendMessage(notification));
+        terminationListeners.clear();
+    }
+
+    private static TerminationNotificationResponse.StackTrace transformToProtoBuf(@NotNull final Throwable throwable) {
+        return TerminationNotificationResponse.StackTrace.newBuilder()
+                .setType(throwable.getClass().getSimpleName())
+                .setMessage(throwable.getMessage())
```
niloc132, line 90: plenty of exceptions have null messages, this will throw if passed a null, consider making StackTrace.message optional, and only conditionally setting it (empty != null and all)

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionService.java
```diff
@@ -51,6 +57,64 @@ public SessionService(final Scheduler scheduler, final SessionState.Factory sess
...
+        }
+
+        final TerminationNotificationResponse notification = builder.build();
+        terminationListeners.forEach(listener -> listener.sendMessage(notification));
+        terminationListeners.clear();
+    }
+
+    private static TerminationNotificationResponse.StackTrace transformToProtoBuf(@NotNull final Throwable throwable) {
+        return TerminationNotificationResponse.StackTrace.newBuilder()
+                .setType(throwable.getClass().getSimpleName())
```
niloc132, line 89: my mild preference would be that if we only send the simple name, make sure we qualify it with any enclosing element other than a package, to avoid Foo.Error vs Bar.Error.

But if we're already shipping the full stack trace with full frames (and each frame has the fully qualified class name), we should just suck it up and put the fqcn here too

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -25,12 +25,8 @@
 import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.application_pb.FieldInfo;
 import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.application_pb.FieldsChangeUpdate;
 import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.application_pb.ListFieldsRequest;
-import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.session_pb.ExportNotification;
-import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.session_pb.ExportNotificationRequest;
-import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.session_pb.ExportRequest;
-import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.session_pb.HandshakeRequest;
-import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.session_pb.HandshakeResponse;
-import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.session_pb.ReleaseRequest;
+import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.session_pb.*;
```
niloc132, line 28: ?

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -377,6 +380,66 @@ private void authUpdate(HandshakeResponse handshakeResponse) {
...
+                                    retval.append("\n").append(elements.getAt(jj));
+                                }
+                            }
+
+                            return retval.toString();
+                        }
+
+                        @Override
+                        public BrowserHeaders getMetadata() {
+                            return null; // nothing to offer
```
niloc132, line 437: probably a new BrowserHeaders instead - empty, rather than null?

web/client-api/src/main/java/io/deephaven/web/client/api/QueryConnectable.java
```diff
@@ -79,12 +79,18 @@ static AuthTokenPromiseSupplier oneShot(ConnectToken initialToken) {
...
     private boolean closed;
     private boolean hasDisconnected;
+    private boolean notifiedConnectionError = false;
 
     public QueryConnectable(Supplier<Promise<ConnectToken>> authTokenPromiseSupplier) {
         this.connection = JsLazy.of(() -> new WorkerConnection(this, authTokenPromiseSupplier));
     }
 
     public void notifyConnectionError(ResponseStreamWrapper.Status status) {
+        if (notifiedConnectionError) {
```
rcaudy, line 89: What's the threading model here? Should this be an `AtomicBoolean`, or should the method/block be synchronized?

nbauernfeind: In javascript the threading model is ... you get one thread. That's it! That's the main reason it is extremely important to design all browser-executing code as async as possible.

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionService.java
```diff
@@ -83,14 +93,19 @@ private synchronized void onFatalError(
...
+                .setType(throwable.getClass().getName())
+                .setMessage(stringifyIfNull(throwable.getMessage()))
                 .addAllElements(Arrays.stream(throwable.getStackTrace())
+                        .limit(MAX_STACK_TRACE_DEPTH)
                         .map(StackTraceElement::toString)
                         .collect(Collectors.toList()))
                 .build();
     }
 
+    private static String stringifyIfNull(@Nullable String value) {
```
rcaudy, line 105: You should delete this and just use `java.util.Objects#toString(java.lang.Object)` or `java.lang.String#valueOf(java.lang.Object)`.

rcaudy (commented):
I looked at the changes in the last 7 commits for grpc_api classes. I assume Colin owns review for the rest. I have one suggestion, but otherwise this is ready to go.

niloc132:
Generated JS is more or less ignored, as long as it doesn't break the build - that is the scaffolding such that the JS API could eventually call this api.

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionService.java
```diff
@@ -2,28 +2,39 @@
...
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
-import java.util.Deque;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
```
niloc132, line 22: expand

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionService.java
```diff
@@ -2,28 +2,39 @@
...
 import com.github.f4b6a3.uuid.UuidCreator;
 import com.google.protobuf.ByteString;
+import io.deephaven.configuration.Configuration;
 import io.deephaven.db.tables.utils.DBDateTime;
 import io.deephaven.db.tables.utils.DBTimeUtils;
+import io.deephaven.grpc_api.util.GrpcUtil;
+import io.deephaven.proto.backplane.grpc.TerminationNotificationResponse;
 import io.deephaven.util.auth.AuthContext;
 import io.deephaven.grpc_api.util.Scheduler;
+import io.deephaven.util.process.ProcessEnvironment;
```
niloc132, line 12: sort imports

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -377,6 +404,66 @@ private void authUpdate(HandshakeResponse handshakeResponse) {
...
+                            subscribeToTerminationNotification();
+                            return;
+                        }
+                    }
+
+                    // welp; the server is gone -- let everyone know
+                    info.notifyConnectionError(new ResponseStreamWrapper.Status() {
+                        @Override
+                        public double getCode() {
+                            return 14; // UNAVAILABLE
```
niloc132, line 422: Code.AVAILABLE

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -377,6 +404,66 @@ private void authUpdate(HandshakeResponse handshakeResponse) {
...
+                        @Override
+                        public String getDetails() {
+                            final TerminationNotificationResponse response = (TerminationNotificationResponse) success;
+                            if (!response.getAbnormalTermination()) {
+                                return "Server exited normally.";
+                            }
+
+                            final StringBuilder retval;
+                            if (!response.getReason().isEmpty()) {
+                                retval = new StringBuilder(response.getReason());
```
niloc132, line 434: fun fact, the gwt impl of stringbuilder/buffer is just a string that gets stuff +='d to it, its more efficient in all modern browsers than joining an array after you know the full size

niloc132 (approved):

PR #1211, opened 2021-09-03
TableService's gRPC lazyUpdate/update/select impls must acquire LTM shared lock

nbauernfeind:


devinrsmith:
I found this LTM bug during some client/server testing. Cherry-picking this change fixed my issue.

devinrsmith (approved):

PR #1220, opened 2021-09-08
Convert JS API examples to python

### LINKED ISSUE #904: Convert JSAPI examples from Groovy
Right now there are a couple JSAPI examples (table_viewport.html, table_filter.html, table_sort.html) that execute snippets of Groovy code. However, we default to Python on startup now.

Could do one of the following:
1) Convert examples to Python
2) Convert examples to use JSAPI commands instead of running a snippet of code (requires some other JSAPI changes)
3) Display error if wrong language selected, prompting user to start up in the correct language

mofojed:
- By default, deephaven-core runs Python
- Examples were in Groovy. Convert them to Python to work with default config
- Add a new basic example that simple opens a static table and outputs the data. This super simple example will be referenced from the tutorials in the docs.
- Fixes #904

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/public/table_basic.html
+++ b/web/client-api/src/main/java/io/deephaven/web/public/table_basic.html
@@ -2,7 +2,7 @@
 <html>
 <head>
     <meta charset="utf-8" />
-    <title>Deephaven JS API table viewport and updates</title>
+    <title>Deephaven JS API display basic table</title>
     <script src="dh-internal.js" type="text/javascript"></script>
     <script src="dh-core.js" type="text/javascript"></script>
     <link href="basic.css" rel="stylesheet" type="text/css" />
@@ -15,7 +15,7 @@
 <body>
 
 <h3>Table data</h3>
-<table id="simpleTable" />
+<table id="simpleTable"></table>
 
 <script>
 (async () => {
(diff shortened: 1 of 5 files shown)


web/client-api/src/main/java/io/deephaven/web/public/table_basic.html
```diff
@@ -0,0 +1,71 @@
...
+    // Open a connection to the server
+    var connection = new dh.IdeConnection(window.location.protocol + "//" + window.location.host);
+
+    // Start a python session
+    var ide = await connection.startSession("python");
+
+    // Run code that will create a static table with 10 rows and three columns, I, J, K
+    await ide.runCode(`
+from deephaven.TableTools import emptyTable
+remoteTable = emptyTable(10).updateView("I=i", "J=I*I", "K=i%2==0?\`Hello\`:\`World\`")
```
niloc132, line 31: i hate the escaping quoting in quoting - python supports single quotes, maybe use that instead? ` -> ' -> "

web/client-api/src/main/java/io/deephaven/web/public/table_basic.html
```diff
@@ -0,0 +1,71 @@
+<!doctype html>
+<html>
+<head>
+    <meta charset="utf-8" />
+    <title>Deephaven JS API table viewport and updates</title>
```
niloc132, line 5: change to match the example

web/client-api/src/main/java/io/deephaven/web/public/table_basic.html
```diff
@@ -0,0 +1,71 @@
...
+    <style>
+        table { border-spacing: 0; }
+        thead td { font-weight: bold; border-bottom: 1px solid black; }
+        tbody td { border-bottom: 1px solid gray; padding: 4px; }
+    </style>
+</head>
+<body>
+
+<h3>Table data</h3>
+<table id="simpleTable" />
```
niloc132, line 18: unless there has been a recent change, tables can't be self closing

niloc132 (approved):

Issue #733, opened 2021-06-09
Finish server validation/normalization of filters


niloc132:
Follow-up to #78.

The issue is complete to provide the feature, but there are some improvements that are possible - this issue is to follow up on the TODOs left from the first patch.

- [ ] Validate the passed filters so we avoid potential problems later. The `anyof` implementation we are using now makes it a bit easier to express the actual expected structure than the hierarchy we previously were using.
- [ ] Replace any EQUALS/NOT_EQUALS CompareCondition with a comparable InCondition, so that it correctly is transformed into a MatchFilter
- [ ] Within an OR, find any compatible "in" referring to the same column, within an AND find any compatible "not in" referring to the same column
- [ ] If the resulting item has a top-level AND, unwrap those items and pass them directly to table.where. If the resulting item has a top-level OR, unwrap those items and pass them directly to table.whereIn.

PR #1229, opened 2021-09-08
Fix case insensitive string filters

mofojed:
- When doing a egIgnoreCase filter, the filter gets created on a different path. In that path, strings were not being escaped as they should be.
- Partial #733
- Fixes web-client-ui#190


nbauernfeind (dismissed):

grpc-api/src/main/java/io/deephaven/grpc_api/table/ops/filter/FilterFactory.java
```diff
@@ -173,7 +173,7 @@ public SelectFilter onIn(Value target, List<Value> candidatesList, CaseSensitivi
             if (literal.getValueCase() == Literal.ValueCase.NANO_TIME_VALUE) {
                 values[i] = "'" + new DBDateTime(literal.getNanoTimeValue()).toString(DBTimeZone.TZ_DEFAULT) + "'";
             } else {
-                values[i] = FilterPrinter.printNoEscape(literal);
+                values[i] = FilterPrinter.print(literal);
```
niloc132, line 176: This will break with characters that must be escaped, the noescape variant is expected to wrap in double quotes, but still leave the output unescaped. For better or worse, MatchFilter expects double quote wrapping the strings, but ignores any other escaping inside the string, we should consider changing this behavior.

niloc132 (dismissed):

nbauernfeind (approved):

PR #1234, opened 2021-09-09
correctly set sort direction enum values, fixes #1233

### LINKED ISSUE #1233: The sort direction enum in the Python client isn't correctly assigned values
1. This makes the sort operation on tables return data in the wrong order.
2. Also sort direction should be optional and should be ASCENDING by default.
3. need a convenient way to check if an arrow type is currently supported by DHC

jmao-denver:
the util function makes it easier to prepare external data before 'doPut'. It might not be needed if the DH error reporting were more usable for developers of all skill levels

### DIFF
--- a/pyclient/pydeephaven/utils.py
+++ b/pyclient/pydeephaven/utils.py
@@ -1,0 +1,16 @@
+#
+#  Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
+#
+import pyarrow
+
+from .dherror import DHError
+from ._arrow_flight_service import _map_arrow_type
+
+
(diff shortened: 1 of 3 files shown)


pyclient/pydeephaven/utils.py
```diff
@@ -0,0 +1,16 @@
+#
+#  Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
+#
+import pyarrow
+
+from .dherror import DHError
+from ._arrow_flight_service import _map_arrow_type
+
+
+def is_deephaven_compatible(data_type: pyarrow.DataType) -> bool:
```
niloc132, line 10: doesn't seem to fit in the PR?

niloc132 (approved):

PR #1238, opened 2021-09-09
Reduce verbosity of UNAUTHENTICATED calls from alternative GrpcUtil.rpcWrapper

nbauernfeind:
We reduced the log spam from the more commonly used GrpcUtil.rcpWrapper path, but I missed this one for some reason.

### DIFF
--- a/grpc-api/src/main/java/io/deephaven/grpc_api/util/GrpcUtil.java
+++ b/grpc-api/src/main/java/io/deephaven/grpc_api/util/GrpcUtil.java
@@ -21,7 +21,7 @@ public class GrpcUtil {
     /**
      * Utility to avoid errors escaping to the stream, to make sure the server log and client both see the message if
      * there is an error, and if the error was not meant to propagate to a gRPC client, obfuscates it.
-     * 
+     *
      * @param log the current class's logger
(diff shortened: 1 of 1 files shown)


grpc-api/src/main/java/io/deephaven/grpc_api/util/GrpcUtil.java
```diff
@@ -57,7 +57,11 @@
         try (final SafeCloseable ignored = LivenessScopeStack.open()) {
             return lambda.call();
         } catch (final StatusRuntimeException err) {
-            log.error().append(err).endl();
+            if (err.getStatus().equals(Status.UNAUTHENTICATED)) {
+                log.debug().append("ignoring unauthenticated request: ").append(err).endl();
```
niloc132, line 61: thoughts on INFO so it shows up (and isnt just dropped entirely), but isn't visible in the UI? unless i'm mistaken, debug/trace logging are really only meant for "i'm debugging a problem" not "suppress this until i go looking for it", since it will just get dropped, won't be available for someone to look later.

Would be nice if we could junk the stack trace though, and just log where we're being called from...

niloc132 (approved):

PR #1239, opened 2021-09-09
Fix logical types for unions.

jcferretti:
Logical type processing fails for types embedded in a union.  Thanks to Devin for reporting and for testing the fix.


devinrsmith (approved):
This fixed my problems!

PR #1246, opened 2021-09-10
Added the examples to demo how to use the major functions of pydeephaven

jmao-denver:
The location of the examples might need to be documented/mentioned on DHC docs site under 'how to' or some other appropriate section.


pyclient/examples/demo_run_script.py
```diff
@@ -0,0 +1,36 @@
...
+"""Demo how to run a Python script on the server and open a table generated by the script."""
+import pandas as pd
+
+from examples.import_test_data import import_taxi_records
+from pydeephaven import Session, Table
+
+
+def run_script(dh_session: Session) -> Table:
+    server_script = '''
+t2 = t.where("VendorID > 0")\
```
jcferretti, line 13: If it doesn't affect the execution, would it make sense to indent the script a bit?  I think if we end the line in a way that it makes it evident for python that the script has not finished (like after an open parenthesis) we may be able to avoid the use of backslash and also indent the code in the script to make it a bit more readable.  I am thinking something like

```
t2 = t.where(
     "VendorID > 0"
).sort(
     "VendorID", "fare_amount"
).headBy(
     5, "VendorID"
)
```

I am not actually sure this is better or not.   Really just throwing it out there as an idea.  I used a style like that a bit in a doc I wrote to split long queries in more legible pieces.

pyclient/examples/import_test_data.py
```diff
@@ -0,0 +1,37 @@
...
+
+from examples.csv_downloader import download_csv
+from pydeephaven import Session, Table
+from pydeephaven.utils import is_deephaven_compatible
+
+
+def import_taxi_records(dh_session: Session) -> Table:
+
+    # download the CSV data and read it into a pyarrow table and prepare it for uploading into DH
+    csv_file_name = download_csv(url="https://nyc-tlc.s3.amazonaws.com/trip+data/yellow_tripdata_2020-12.csv")
```
jcferretti, line 22: I am a bit confused here, the variable is called `csv_file_name`, but the name of the function that is being called is `download_csv`.  It sounds like `csv_file_name` is the csv contents?

jcferretti (approved):

pyclient/examples/import_test_data.py
jcferretti: I really like how readable this code ends up being.   In general the argument names help a lot too, but is pretty straightforward to follow in terms of intent.

pyclient/examples/demo_run_script.py
jmao-denver: Yes, definitely

jcferretti (approved):

PR #1248, opened 2021-09-10
DBLanguageParser allow unboxing and widening when finding available methods

### LINKED ISSUE #1247: Python Issues with pre-defined constants
**Description**
There is an issue with using predefined constants. For example, we can't use HOUR in formulas.

```
from deephaven.TableTools import newTable, stringCol, dateTimeCol
from deephaven.DBTimeUtils import *

time1 = convertDateTime("2020-04-01T09:00:00 NY")
time2 = convertDateTime("2020-04-01T10:28:32 NY")
time3 = convertDateTime("2020-04-01T12:00:00 NY")
time4 = convertDateTime("2020-04-01T16:59:59 NY")

events = ttools.newTable(
   dateTimeCol("EventDateTime", time1, time2, time3, time4),
   stringCol("Level", "Info", "Error", "Warning", "Info"),
   stringCol("Event", "System starting", "Something bad happened", "Invalid login", "System stopping"),
)

baseTime = convertDateTime("2020-04-01T10:00:00 NY")
eventsMiddleDay3 = events.where("EventDateTime >= baseTime && EventDateTime <= baseTime + HOUR")
Exception indicates it's having trouble on the formula:
io.deephaven.db.v2.select.FormulaCompilationException: Formula compilation error for: EventDateTime >= baseTime && EventDateTime <= baseTime + HOUR
```
gives
```Having trouble with the following expression:
Full expression           : EventDateTime >= baseTime && EventDateTime <= baseTime + HOUR
Exception message         : Cannot find method lessEquals(io.deephaven.db.tables.utils.DBDateTime, java.lang.Object) 
```

**Steps to reproduce**
See above. Per Colin:
```
# These work:
x = TableTools.emptyTable(1).update("HR = currentTime() + (long)HOUR")
x = TableTools.emptyTable(1).update("HR = currentTime() + (int)HOUR")
x = TableTools.emptyTable(1).update("HR = currentTime() + java.lang.Long.valueOf((int)HOUR)")

# These fail:
x = TableTools.emptyTable(1).update("HR = currentTime() + HOUR")
x = TableTools.emptyTable(1).update("HR = currentTime() + (java.lang.Integer)HOUR")
```

**Expected results**
Formulas like this should work.

**Actual results**
See above exception.

From slack discussion;

This actually works:
```from deephaven.DBTimeUtils import *
x = TableTools.emptyTable(1).update("HR = currentTime() + (long)HOUR")
```

Colin: if i am tracing this properly, we go through PyObject.getObjectValue()
we should be checking this by calling pyobject.isLong()
PyObject.getObjectValue() is the culprit
we must stop using that, as it necessarily boxes, and in native code
the incorrect method is in PythonScopeJpyImp.convert

**Versions**
 - Deephaven: 0.3.1
 - OS: mac 10.15.6
 - Browser: Chrome
 - Docker: ...

nbauernfeind:
Fixes #1247

### DIFF
--- a/DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageParser.java
+++ b/DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageParser.java
@@ -25,6 +25,7 @@ import com.github.javaparser.ast.type.WildcardType;
 import com.github.javaparser.ast.visitor.GenericVisitorAdapter;
 import io.deephaven.db.tables.dbarrays.*;
 import io.deephaven.DeephavenException;
+import org.apache.commons.lang3.ClassUtils;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.jpy.PyObject;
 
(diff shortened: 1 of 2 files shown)


DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageParser.java
```diff
@@ -527,68 +528,70 @@ private String paramsTypesToString(Class<?>[] paramTypes) {
...
+            }
 
-                if (candidateParamTypes.length == paramTypes.length && paramType.isArray()) {
-                    if (!isAssignableFrom(candidateParamTypes[candidateParamTypes.length - 1], paramType)) {
-                        acceptable = false;
-                    }
-                } else {
-                    final Class<?> lastClass = candidateParamTypes[candidateParamTypes.length - 1].getComponentType();
+            if (candidateParamTypes.length == paramTypes.length && paramType.isArray()) {
+                if (!isAssignableFrom(candidateParamTypes[candidateParamTypes.length - 1], paramType)) {
```
nbauernfeind, line 573: I suspect I should move the new logic into a helper, and allow unboxing and widening on varargs, too.

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageParser.java
```diff
@@ -527,68 +528,70 @@ private String paramsTypesToString(Class<?>[] paramTypes) {
...
+                    if (isDbArray(paramType) && lastClass.isArray()) {
+                        paramType = convertDBArray(paramType,
+                                parameterizedTypes[i] == null ? null : parameterizedTypes[i][0]);
+                    }
 
-                        if (!isAssignableFrom(lastClass, paramType)) {
-                            acceptable = false;
-                            break;
-                        }
+                    if (!isAssignableFrom(lastClass, paramType)) {
```
nbauernfeind, line 587: Using it here, as well.

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageParser.java
```diff
@@ -527,68 +528,79 @@ private String paramsTypesToString(Class<?>[] paramTypes) {
...
-            if (acceptable) {
-                accepted.add(candidate);
-            }
+        accepted.add(candidate);
+    }
+
+    private static boolean canAssignType(final Class<?> candidateParamType, final Class<?> paramType) {
+        boolean canAssignWithoutUnbox = isAssignableFrom(candidateParamType, paramType);
+
+        if (paramType.isArray() && candidateParamType.isArray()
```
rcaudy, line 592: This should only work for object arrays, unless or until we add implicit primitive array conversion.

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageParser.java
```diff
@@ -835,24 +847,33 @@ public static boolean isDbArray(Class<?> type) {
...
-                anyExpressionTypesArePrimitive = false;
+                allExpressionTypesArePrimitive = false;
             } else {
                 for (int ei = nArgs - 1; ei < nArgExpressions; ei++) {
                     // iterate over the vararg argument expressions
-                    if (varArgType != expressionTypes[ei] && varArgType.isPrimitive()
-                            && expressionTypes[ei].isPrimitive()) {
+                    if (varArgType == expressionTypes[ei]) {
+                        continue;
+                    } else if (argumentTypes[nArgs - 1] == expressionTypes[ei]) {
```
rcaudy, line 856: If the LHS is assignable to the RHS, we're already good. And we should do that before the loop.

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageParser.java
nbauernfeind: Captured the follow-up conversation in #1301. I added the biggest concern and the biggest benefit. As we discussed offline; it's not obvious if it's worth doing.

rcaudy (approved):

PR #1318, opened 2021-09-20
Make QueryTableWhereTest#testInterFilterInterruption more robust for CI

### LINKED ISSUE #1317: Nightly Check Transient Failure: QueryTableWhereTest#testInterFilterInterruption
The nightly build has failed four times in a row with a timeout located here:

```
io.deephaven.db.v2.QueryTableWhereTest > testInterFilterInterruption FAILED
    java.lang.IllegalStateException at QueryTableWhereTest.java:821
```

At first glance it's not obvious how to perform this test in a better way. It's possible tuning down the number of iterations will the runtime down to within the timeout.

nbauernfeind:
Fixes #1317

### DIFF
--- a/DB/src/test/java/io/deephaven/db/v2/QueryTableWhereTest.java
+++ b/DB/src/test/java/io/deephaven/db/v2/QueryTableWhereTest.java
@@ -4,7 +4,6 @@
 
 package io.deephaven.db.v2;
 
-import io.deephaven.base.SleepUtil;
 import io.deephaven.db.exceptions.QueryCancellationException;
 import io.deephaven.db.tables.SortingOrder;
 import io.deephaven.db.tables.Table;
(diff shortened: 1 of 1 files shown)


DB/src/test/java/io/deephaven/db/v2/QueryTableWhereTest.java
```diff
@@ -716,32 +713,25 @@ public void testWhereBoolean() {
...
         public int applyAsInt(int value) {
-            if (sleepDurationNanos > 0) {
-                final long start = System.nanoTime();
-                final long end = start + sleepDurationNanos;
-                // noinspection StatementWithEmptyBody
-                while (System.nanoTime() < end);
-            }
-            if (++invokes == 1) {
-                latch.countDown();
+            if (++invokes == 1 && interrupt) {
```
rcaudy, line 726: This might be better if we specified an interval to interrupt after, but what you've done is more than adequate for the current use case.

rcaudy (approved):

Issue #1325, opened 2021-09-21
Improve KafkaIngester / KafkaConsumer error handling


jcferretti:
When we get a deserializer error while consuming records, we fail the ingester (which is better that what we were doing before, leaving behind a borked ingester).  But we can do better, we can deliver on the idea (implemented in other parts of the code for other errors) of tolerating a certain number of failed events while still processing.

See KafkaIngester.java around line 320:

```
$ git diff
diff --git a/Kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java b/Kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
index c04c08ea4..cd3e44417 100644
--- a/Kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
+++ b/Kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
@@ -320,7 +320,7 @@ public class KafkaIngester {
         } catch (Exception ex) {
             log.error().append(logPrefix).append("Exception while polling for Kafka messages:").append(ex)
                     .append(", aborting.");
-            return false;
+            return true;
         }
 
         for (final TopicPartition topicPartition : records.partitions()) {
```

PR #1326, opened 2021-09-21
Avoid leaving behind a failed consumer.

jcferretti:
See code; if we return false at this point, the consumer never recovers without further intervention in the KafkaConsumer state; having one failed message should be recoverable but we need to explore how to make that happen, I created a separate issue for it https://github.com/deephaven/deephaven-core/issues/1325


Kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
```diff
@@ -320,7 +320,7 @@ private boolean pollOnce(final Duration timeout) {
         } catch (Exception ex) {
             log.error().append(logPrefix).append("Exception while polling for Kafka messages:").append(ex)
                     .append(", aborting.");
-            return false;
+            return true;
```
rcaudy, line 323: Error message no longer matches the return. Also, there's no feedback here. What do we expect to be the result?

rcaudy: My confusion.

rcaudy (approved):

Issue #1312, opened 2021-09-18
Configure log4j to prevent dependent libraries from spamming log4j warnings on the REPL


jcferretti:
Our use of the http client library seems to require a log4j configuration to prevent some undesirable warnings to spam the REPL.  Other libraries may require the same.
![Selection_108](https://user-images.githubusercontent.com/37232625/133866802-6ddd1c29-3b16-4ed6-9c22-7ff1f86fe277.png)

devinrsmith:
We should likely add a bridge `log4j-over-slf4j` so we don't need to configure log4j itself - http://www.slf4j.org/legacy.html#log4j-over-slf4j; our configuration domain can still lie w/ slf4j/logback.

devinrsmith:
Note: my statement about is *not* a bridge - it's a *jar* replacement. This is slightly different than a bridge. The replacement will take a little bit more build logic (ie, whenever a transitive dependency needs log4j.jar use log4j-over-slf4j.jar instead). It was likely implemented this way due to the lack of extension mechanisms built into log4j.

devinrsmith:
Actually, this logging is further abstracted by apache commons logging, org.apache.commons.logging.Log - luckily, that part is more configurable, and so we should be able to use a bridge directly to slf4j... http://www.slf4j.org/legacy.html#jclOverSLF4J - in which case, we should be including the `jcl-over-slf4j` jar.

PR #1329, opened 2021-09-21
Avoid log4j dropping warnings in the console for dependent libraries.

jcferretti:
Thanks to Devin for the idea behind how to fix this.   In the example below, our dependency to Apache HttpClient from the Kafka module triggers a warning the first time it is used when we are making the first call to `consumeToTable` on the console.

Without this fix:

Deephaven - Google Chrome_107

With this fix:

Deephaven - Google Chrome_106

### DIFF
--- a/log-factory/sinks/log-to-slf4j/build.gradle
+++ b/log-factory/sinks/log-to-slf4j/build.gradle
@@ -7,5 +7,6 @@ dependencies {
 	// internal logger, that is in service of the ServiceLoader - and if consumers should not need
 	// access to the internal logger.
 	implementation project(':log-factory')
+	compileOnly 'org.slf4j:slf4j-api:1.7.30' // users must provide their own impl and slf4j config
 	compileOnly 'ch.qos.logback:logback-classic:1.2.3' // so we can attach an interceptor, if supported
 }


log-factory/sinks/log-to-slf4j/build.gradle
```diff
@@ -7,6 +7,5 @@ dependencies {
 	// internal logger, that is in service of the ServiceLoader - and if consumers should not need
 	// access to the internal logger.
 	implementation project(':log-factory')
-	compileOnly 'org.slf4j:slf4j-api:1.7.30' // users must provide their own impl and slf4j config
```
devinrsmith, line 10: We shouldn't be removing a compile dependency from here. The code still compiles b/c logback-classic transitively depends on slf4j-api, but we are best to be explicit here.

If you want to make a follow up ticket to consolidate slf4j versions to make sure they all match, go ahead and assign it to me.

devinrsmith (changes requested):

log-factory/sinks/log-to-slf4j/build.gradle
jcferretti: Done and done.

jcferretti:
Fixes https://github.com/deephaven/deephaven-core/issues/1312

devinrsmith (approved):

PR #1338, opened 2021-09-22
Move default nginx root outside of first location

nbauernfeind:
According to https://www.nginx.com/resources/wiki/start/topics/tutorials/config_pitfalls/#root-inside-location-block

It is bad practice to declare the primary `root` inside of each `location` of an nginx.conf.

Note that this block does not specify a location, but should share the same location as the `/` location.
```
    location ~* \.chunk.(?:css|js)$ {
        expires 1y;
        add_header Cache-Control 'public, immutable, max-age=31536000';
    }
```

Took me a hot minute to find out why I couldn't fetch static resources from a non-default location.

### DIFF
--- a/web/client-ide/nginx/default.conf
+++ b/web/client-ide/nginx/default.conf
@@ -8,8 +8,8 @@ server {
     # Redirect from / to the IDE
     rewrite ^/$ /ide/ redirect;
 
+    root   /usr/share/nginx/html;
     location / {
-        root   /usr/share/nginx/html;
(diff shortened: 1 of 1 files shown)


devinrsmith:
LGTM, but I'll defer to @mofojed

mofojed (commented):
What error are you getting right now? What's the use case for serving up static files outside of our specified locations?

mofojed (approved):

PR #1344, opened 2021-09-23
Simple Kafka publishing, not cleaned up yet.

cpwright:


cpwright:
The commands I used to exercise this:
result = newTable(
    doubleCol("Doubles", 3.1, 5.45, -1.0),
    stringCol("Strings", "Creating", "New", "Tables"),
    intCol("Ints", 1, 2, 3)
)
stringSer = new io.deephaven.kafka.publish.SimpleKeyOrValueSerializer(result, "Strings")
jsonSer = new io.deephaven.kafka.publish.TableToJSONAdapter.Builder().excludeColumn("Strings").buildFactory().apply(result)
props = new Properties();
 props.put("bootstrap.servers", "redpanda:29092");
 props.put("key.serializer", "org.apache.kafka.common.serialization.StringSerializer");
 props.put("value.serializer", "org.apache.kafka.common.serialization.StringSerializer");
ptk=new io.deephaven.kafka.publish.PublishToKafka(props, result, "foo", stringSer, jsonSer)


And then from my host exec'ing docker:
docker exec -it 4f7a7c378e68 rpk topic consume foo --brokers=localhost:9092

Kafka/src/main/java/io/deephaven/kafka/publish/JsonKeyOrValueSerializer.java
```diff
@@ -0,0 +1,910 @@
...
+            }
+        }
+
+        @Override
+        FieldContext makeContext(int size) {
+            return new CharContext(size);
+        }
+
+        @Override
+        void processField(FieldContext fieldContext, WritableObjectChunk<ObjectNode, Attributes.Values> jsonChunk,
```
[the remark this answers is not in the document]
cpwright: So we could use replication, but not generics if we want to read the chunk nicely and do the comparisons.  We could also say screw it "JSON is bloated" why not just box the crap and have a single version.

One issue with replication is the inner classes, which would be less beautiful but are obviously doable by just passing in other parameters and having them be somewhat incestuous.

Kafka/src/main/java/io/deephaven/kafka/publish/PublishToKafka.java
```diff
@@ -0,0 +1,138 @@
...
+    private final Table table;
+    private final KafkaProducer<K, V> producer;
+    private final String topic;
+    private final KeyOrValueSerializer<K> keySerializer;
+    private final KeyOrValueSerializer<V> valueSerializer;
+
+    @ReferentialIntegrity
+    private final PublishListener publishListener;
+
+    public PublishToKafka(final Properties props,
```
jcferretti, line 38: I think we want a partition number too;
The issue of publishing subsets of keys by partition can depend on the use for defining tables with proper filtering per key; but the mapping to particular partition seems missing.

cpwright: You don't think that should be up to the partitioner specified in properties?
https://jaceklaskowski.gitbooks.io/apache-kafka/content/kafka-producer-internals-DefaultPartitioner.html

jcferretti: You are right, I completely forgot about how partitions work on the publishing side.  Nevermind.

Kafka/src/main/java/io/deephaven/kafka/publish/JsonKeyOrValueSerializer.java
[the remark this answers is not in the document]
jcferretti: Sounds hard and painful.  Let's just leave it as it is.  I did my reviewer duty of asking "maybe this can be avoided without too much work?", the answer was NO.

Kafka/src/main/java/io/deephaven/kafka/publish/GenericRecordKeyOrValueSerializer.java
```diff
@@ -0,0 +1,723 @@
...
+            if (!columnsToOutputFields.isEmpty()) {
+                sb.append(", mapped to fields=").append(columnsToOutputFields.keySet());
+            }
+            throw new KafkaPublisherException(sb.toString());
+        }
+
+        // Now create all the processors for specifically-named fields
+        columnsToOutputFields.forEach(this::makeFieldProcessor);
+
+        if (!StringUtils.isNullOrEmpty(timestampFieldName)) {
```
jcferretti, line 489: I don't think is obvious to me how this feature works.   We seem to be adding support for an additional "timestamp me this at the time of generation".  Which can come on either key, value, or both (by nature of our support).
The utility of that versus the Kafka native timestamp field, again, not obvious to me.

Kafka/src/main/java/io/deephaven/kafka/publish/GenericRecordKeyOrValueSerializer.java
```diff
@@ -0,0 +1,723 @@
...
+
+        @Override
+        FieldContext makeContext(int size) {
+            return null;
+        }
+
+        @Override
+        public void processField(FieldContext fieldContext,
+                WritableObjectChunk<GenericRecord, Attributes.Values> avroChunk, OrderedKeys keys, boolean isRemoval) {
+            // do we really want nanos instead of micros; there are avro things for micros/millis?
```
jcferretti, line 438: There are, to do this right seems like we would need to consider the field type in the schema?
https://avro.apache.org/docs/1.8.2/spec.html#Timestamp+%28millisecond+precision%29

jcferretti: Also, we need to handle DBDateTime outside of the auto generated at publish time (ie, as an actual existing column of type DBDateTime) as one of those avro types too?

Kafka/src/main/java/io/deephaven/kafka/publish/PublishToKafka.java
```diff
@@ -0,0 +1,138 @@
...
+        if (table.isLive()) {
+            ((DynamicTable) table.coalesce()).listenForUpdates(publishListener = new PublishListener());
+        } else {
+            publishListener = null;
+        }
+
+        // Publish the initial table state
+        publishMessages(table.getIndex(), false, true);
+    }
+
```
jcferretti, line 59: We need a way to shutdown, which I think I need to use inside a callback-type object returned from `KafkaTools.produceFromTable`.  In my branch I am using the code below, not sure if that is right, but maybe.

```
    public PublishToKafka(final Properties props,
            final Table table,
            final String topic,
            final KeyOrValueSerializer<K> keySerializer,
            final KeyOrValueSerializer<V> valueSerializer) {
        this.producer = new KafkaProducer<>(props);
        this.topic = topic;
        this.keySerializer = keySerializer;
        this.valueSerializer = valueSerializer;


        if (table.isLive()) {
            final DynamicTable dynTable = ((DynamicTable) table.coalesce());
            this.table = dynTable;
            dynTable.listenForUpdates(publishListener = new PublishListener());
        } else {
            this.table = table;
            publishListener = null;
        }

        // Publish the initial table state
        publishMessages(table.getIndex(), false, true);
    }

    public void shutdown() {
        if (!table.isLive()) {
            return;
        }
        final DynamicTable dynTable = ((DynamicTable) table;
        dynTable.removeUpdateListener(publishListener);
    }
```

Kafka/src/main/java/io/deephaven/kafka/publish/PublishToKafka.java
```diff
@@ -0,0 +1,138 @@
...
+            this.isStream = StreamTableTools.isStream(table);
+        }
+
+        @Override
+        public void onUpdate(Update upstream) {
+            Assert.assertion(upstream.shifted.empty(), "upstream.shifted.empty()");
+
+            if (isStream) {
+                Assert.assertion(upstream.modified.empty(), "upstream.modified.empty()");
+                publishMessages(upstream.added, false, true);
```
jcferretti, line 129: This seems to assume value is always going to be generated?  What about publishing just in Key and ignoring Value?

cpwright:

PR #1363, opened 2021-09-28
GitHub Action for CLA compliance

chipkent:
Added a GitHub Action to maintain signatures for CLA compliance.  See https://github.com/deephaven/cla for more details.

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -1,0 +1,49 @@
+#
+# A GitHub Action to manage Contributor License Agreement (CLA) compliance.
+# This action uses CLA Assistant (https://github.com/marketplace/actions/cla-assistant-lite) to ensure that
+# all contributors have signed the Deephaven CLA before their PRs are merged.  
+# The CLA can be signed by commenting on the PR.
+#
+# The CLA and all signatures are maintained at:  https://github.com/deephaven/cla
+# The "main" branch of this repository must not be protected.
+#
+# To use this action:
+# 
+# 1) Copy this file to `<your_project>/.github/workflows/cla.yml`
+#
+# 2) PERSONAL_ACCESS_TOKEN
+#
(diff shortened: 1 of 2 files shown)


.github/workflows/cla.yml
```diff
@@ -0,0 +1,49 @@
...
+jobs:
+  CLAssistant:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "CLA Assistant"
+        if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have read the CLA Document and I hereby sign the CLA') || github.event_name == 'pull_request_target'
+        uses: cla-assistant/github-action@v2.1.3-beta
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PERSONAL_ACCESS_TOKEN : ${{ secrets.PERSONAL_ACCESS_TOKEN }}
```
devinrsmith, line 42: This secret doesn't exist in this REPO...

.github/workflows/cla.yml
```diff
@@ -0,0 +1,49 @@
...
+#
+# A personal access token must be set up to allow this bot to commit to the CLA repository.  To do this:
+# a) Create a personal access token with repo scope
+# b) Manually add the token to your repository's secrets as PERSONAL_ACCESS_TOKEN
+#
+#
+# 3) allowlist
+#
+# If a GitHub username is included in the allowlist, they will not be required to sign the CLA.  This is useful for allowing bots to contribute without a CLA.
+#
```
devinrsmith, line 24: I don't think we'll want to have this comment in our public code - it's mostly documentation of how we are supposed to use it...

devinrsmith (changes requested):
We have an internal user, `deephaven-internal`, that we may want to get a personal access token for. Or, we may want to create a `deephaven-cla` user that only has permissions against the CLA repo...

devinrsmith (approved):

Issue #1324, opened 2021-09-21
Expose conditional formatting methods in JS API


vbabich:
In order to implement conditional formatting menu in the Web UI we need to expose the following formatting methods in JS API: `formatColumns`, `formatColumnWhere `, `formatRowWhere`, `setColumnRenderers` (for progress bar format).

mofojed:
@vbabich / @dsmmcken : What do we need these APIs for? I'm looking at the Conditional Formatting impl in Swing, and it appears to be entirely client side (unless I'm missing something). Why do we need this API?

dsmmcken:
It's client side in swing? Probably easier to accomplish client side in java then in js. We don't do client side anything, and would need to re-implement all the comparisons and query parsing for all column types, including datetimes, bigints, etc. Also, it formatting can support any valid query condition, like `Spread / Ask * 100.0000 > Last`, do you really want to be computing that client side?

mofojed:
I misunderstood some things previously, now I see what's going on. Basically need to wire some functionality through `JsTable` to call `Table.formatColumnWhere`, `Table.formatRowWhere`, `Table.formatColumns`. We could do this a couple ways (could use @niloc132 's and @vbabich 's input here). 

**Option 1 - Passing string array**
- Add `JsTable.applyFormatting(string[])`. Calls `Table.formatColumns` with the string array provided. 
- `Table.formatColumnWhere` and `Table.formatRowWhere` call `Table.formatColumns` anyway, so really only need to call `Table.formatColumns`
- One complication is getting the `formatRowWhere` functionality - would require knowledge of the value of `ColumnFormattingValues.ROW_FORMAT_NAME`, which is `__ROWFORMATTED`.
- Also ignores how to handle `setColumnRenders` as well, might be difficult/hacky to add that in later (cannot just be expressed as string)

**Option 2 - Building it similar to filters:**
- Add `FormatValue format()` method on `Column`
- Create `FormatValue` class, with a ref to the column
- Add `JsTable.applyFormatting(FormatValue[])`, which ends up converting all the args to text and then passing to server to call `Table.formatColumns`
- Could be extended to support progress bar - just add a constant `PROGRESS` that format can be set to?
```java
class Column {
  ...
  FormatValue format(String format) {
    return new FormatValue(this.name, format);
  }
  ...
}

class JsTable {
  ...
  void applyFormatting(FormatValue[] formats) { ... }
  ...
}

class FormatValue {
  static rowFormat(String format) {
[...]

PR #1375, opened 2021-09-30
WIP Update WebDevelopersGuide.md with outline of changes for conditional formatting

mofojed:
- Will get Colin and Vlad to review API changes before impl
- Partial #1324

### DIFF
--- a/web/WebDevelopersGuide.md
+++ b/web/WebDevelopersGuide.md
@@ -476,8 +476,10 @@ and also will permit some changes not to inform the UI right away that they have
  Note that the filter property will immediately return the new value, but you may receive update events using the old
  filter before the new one is applied, and the `filterchanged` event fires. Reusing existing, applied filters may enable
  this to perform better on the server.  The `updated` event will also fire, but `rowadded` and `rowremoved` will not.
- * `applyCustomColumns(String[]):String[]` - Replace the current custom columns with a new set. These columns can be
- used when adding new filter and sort operations to the table, as long as they are present.
+ * `applyCustomColumns(String[]):String[]` - Deprecated, use `applyCustomColumns(CustomColumn[])` instead - Replace the current custom columns with a new set. These columns can be
+  used when adding new filter and sort operations to the table, as long as they are present.
+ * `applyCustomColumns(CustomColumn[]):CustomColumn[]` - Replace the current custom columns with a new set. These columns can be
+  used when adding new filter and sort operations to the table, as long as they are present.
  * `setViewport(Number firstRow, Number lastRow, Column[]= columns, Number= updateIntervalMs):TableViewportSubscription` - 
  If the columns parameter is not provided, all columns will be used. If the updateIntervalMs parameter is not provided, 
  a default of one second will be used. Until this is called, no data will be available. Invoking this will result in events
@@ -575,6 +577,8 @@ column.
  Column.
  * `filter():FilterValue` - Creates a new value for use in filters based on this column. Used either as a parameter to
  another filter operation, or as a builder to create a filter operation.
+ * `formatColor(String condition, String color): CustomColumnColor` - Return a `CustomColumnColor` object to apply using `applyCustomColumns` with the parameters specified.
+ * `formatRenderer(String renderer): CustomColumnRenderer` - Return a `CustomColumnRenderer` object to pass the `applyCustomColumns`  
  * `sort():Sort` - Creates a sort builder object, to be used when sorting by this column.
 
 ###### Properties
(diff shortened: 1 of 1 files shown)


vbabich (approved):

web/WebDevelopersGuide.md
```diff
@@ -599,7 +603,23 @@ return a new Sort instance.
 ###### Properties
  * `Column column` - The column which is sorted.
  * `String direction` - The direction of this sort, either `ASC`, `DESC`, or `REVERSE`.
- * `boolean isAbs` - True if the absolute value of the column should be used when sorting; defaults to false.
+ * `boolean isAbs` - True if the absolute value of the column should be used when sorting; defaults to false. 
+
+##### Abstract Class `CustomColumn`
```
niloc132, line 608: do we need another subtype for date/number formats?

mofojed:

web/WebDevelopersGuide.md
```diff
@@ -599,7 +603,23 @@ return a new Sort instance.
...
+ * `String method` - The type of select method to use, either `VIEW` or `SELECT`. Defaults to `VIEW`.
+
+##### Class `CustomColumnFormula` extends `CustomColumn`
+ * `String name` - The name of the custom column to create.
+ * `String formula` - The formula to use for this custom column.
+
+##### Class `CustomColumnColor` extends `CustomColumn`
+ * `String name` - The name of the column to color. Set to `null` to apply to all columns (color entire row).
+ * `String condition` - The condition in which to apply this color.
+ * `String color` - The hex representation of the color to apply.
```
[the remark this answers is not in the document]
mofojed: Hmm looking at conditional formatting docs, it looks like you can only set the background color using the UI? https://docs.deephaven.io/latest/Content/User/workingWithTables/conditionalFormatting.htm?Highlight=conditional%20formatting
Assigning foreground color can only be done through code: https://docs.deephaven.io/latest/Content/writeQueries/formatTables.htm?Highlight=conditional%20formatting

But I agree, we should have the API allow setting both. I guess a `type` enum on CustomColumnColor, FOREGROUND/BACKGROUND as the options, and then on `Column`, add `.formatBackgroundColor` (in addition to the `formatColor` I have in this draft already).

For your first question - yes, I envisioned N CustomColumnColor instances to be set for the same column name, and order is the order of the instances. I guess then the API needs to pull all that into one call to `Table.formatColumns`?

web/WebDevelopersGuide.md
```diff
@@ -599,7 +605,32 @@ return a new Sort instance.
 ###### Properties
  * `Column column` - The column which is sorted.
  * `String direction` - The direction of this sort, either `ASC`, `DESC`, or `REVERSE`.
- * `boolean isAbs` - True if the absolute value of the column should be used when sorting; defaults to false.
+ * `boolean isAbs` - True if the absolute value of the column should be used when sorting; defaults to false. 
+
+##### Abstract Class `CustomColumn`
+ * `String method` - The type of select method to use, either `VIEW` or `SELECT`. Defaults to `VIEW`.
```
niloc132, line 611: this only applies to formulas, any format/style should always use view i _think_

niloc132: OTOH, all subtypes have name, so perhaps name belongs in this type?

web/WebDevelopersGuide.md
```diff
@@ -599,7 +605,32 @@ return a new Sort instance.
...
+
+##### Class `CustomColumnColor` extends `CustomColumn`
+ * `String name` - The name of the column to color. Set to `null` to apply to all columns (color entire row).
+ * `String condition` - The condition in which to apply this color.
+ * `String color` - The hex representation of the color to apply, or case-insensitive CSS color module name (see https://drafts.csswg.org/css-color/#named-colors)
+ * `String type` - The type of cell color formatting, `BACKGROUND` or `FOREGROUND`.
+
+##### Class `CustomColumnDateFormat` extends `CustomColumn`
+ * `String name` - The name of the column to apply date formatting.
+ * `String dateFormat` - Any of the [Java DateFormat](https://docs.oracle.com/javase/7/docs/api/java/text/DateFormat.html) strings.
```
niloc132, line 625: pointer to dh.i18n too? since it is going to be reflected back to the JS API for use locally, won't be evaluated on the server

niloc132 (commented):
Can't reply, so quoting here:

> Hmm looking at conditional formatting docs, it looks like you can only set the background color using the UI? https://docs.deephaven.io/latest/Content/User/workingWithTables/conditionalFormatting.htm?Highlight=conditional%20formatting

Perhaps it isn't documented, but it works - DBColorUtil supports specific methods for this like `DBColorUtil.backgroundForeground(bg, fg)` which will return the expected long value based on both colors. This functions just by bitwise ORing the colors together.

> For your first question - yes, I envisioned N CustomColumnColor instances to be set for the same column name, and order is the order of the instances. I guess then the API needs to pull all that into one call to Table.formatColumns?

Or just updateView, so we can use the existing apis.

mofojed:
Right - I know DBColorUtil supports foreground/background, but it's not shown as an option from the UI. In any case we will support it. And yes, calling updateView instead of formatColumns, my bad.

web/WebDevelopersGuide.md
mofojed: Yea, the only hesitation with putting `name` in this type is that it's a little different in each context... either referencing an existing column (eg. `CustomColumnColor`), or creating a new column (`CustomColumnFormula`). Does that matter?

mofojed: I've made it clear from the docs what the `name` field means in each case. I've also added `Table.formatRow` and `Table.customColumn` methods to create `CustomColumnColor` (with `name=*`) and `CustomColumnFormat` respectively

PR #1380, opened 2021-10-01
Upgrade to Java 11 compiler

devinrsmith:


buildSrc/src/main/groovy/io.deephaven.java-conventions.gradle
```diff
@@ -4,20 +4,62 @@ plugins {
...
-    javac.options.incremental = true
+    //javac.options.incremental = true
     javac.options.compilerArgs << '-parameters'
+
+
+    if (compilerVersion >= 9) {
+      javac.options.release = languageLevel
+
+      // This is for annotation processing
+      javac.options.forkOptions.jvmArgs << '--add-opens=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED'
```
niloc132, line 41: what annotation processors are we using that need access to javac internals?

java-client/session/src/main/java/io/deephaven/client/impl/TableHandleManagerSerial.java
```diff
@@ -21,6 +21,7 @@
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.function.Function;
```
niloc132, line 24: wrong patch? changes in this file don't look like they belong here either, unless java 11-17's type inference got a lot worse?

devinrsmith: This method import can go away, but the changes I made to the lambdas (w/ type hints) seem to be necessary.

PR #1381, opened 2021-10-02
Add a ProcessMemoryLog

jcferretti:
For better usability / introspection regarding Mem utilization and GC pressure.

Deephaven - Google Chrome_109

### DIFF
--- a/DB/src/main/java/io/deephaven/db/tablelogger/ProcessMemoryLogLogger.java
+++ b/DB/src/main/java/io/deephaven/db/tablelogger/ProcessMemoryLogLogger.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright (c) 2016-2019 Deephaven Data Labs and Patent Pending
+ */
+
+package io.deephaven.db.tablelogger;
+
+import io.deephaven.db.tables.TableDefinition;
+import io.deephaven.db.tables.utils.ColumnsSpecHelper;
+import io.deephaven.db.tables.utils.DBDateTime;
+import io.deephaven.db.tables.utils.DBTimeUtils;
+import io.deephaven.tablelogger.Row;
+import io.deephaven.tablelogger.RowSetter;
+import io.deephaven.tablelogger.TableLoggerImpl2;
+import io.deephaven.tablelogger.WritableRowContainer;
+
+import java.io.IOException;
+
+public class ProcessMemoryLogLogger extends TableLoggerImpl2<ProcessMemoryLogLogger.ISetter> {
+    private static final String TABLE_NAME = "ProcessMemoryLog";
+
+    public ProcessMemoryLogLogger() {
+        super(TABLE_NAME);
+    }
+
+    public static String getDefaultTableName() {
+        return TABLE_NAME;
+    }
+
+    interface ISetter extends WritableRowContainer {
+        void log(
+                final Row.Flags flags,
+                final long intervalStartTime, final long intervalDurationMs,
+                final long totalMemory, final long freeMemory,
+                final long intervalCollection, final long intervalCollectionTimeMs) throws IOException;
+    }
+
+    class DirectSetter extends TableLoggerImpl2.BaseSetter implements ISetter {
+        RowSetter<DBDateTime> IntervalStartTime;
+        RowSetter<Long> IntervalDurationMs;
+        RowSetter<Long> TotalMemory;
+        RowSetter<Long> FreeMemory;
+        RowSetter<Long> IntervalCollections;
+        RowSetter<Long> IntervalCollectionTimeMs;
+
+        DirectSetter() {
+            IntervalStartTime = row.getSetter("IntervalStartTime", DBDateTime.class);
+            IntervalDurationMs = row.getSetter("IntervalDurationMs", long.class);
+            TotalMemory = row.getSetter("TotalMemory", long.class);
+            FreeMemory = row.getSetter("FreeMemory", long.class);
+            IntervalCollections = row.getSetter("IntervalCollections", long.class);
+            IntervalCollectionTimeMs = row.getSetter("IntervalCollectionTimeMs", long.class);
+        }
+
+        @Override
+        public void log(
+                final Row.Flags flags,
+                final long intervalStartTime, final long intervalDurationMs,
+                final long totalMemory, final long freeMemory,
+                final long intervalCollections, final long intervalCollectionTimeMs) throws IOException {
+            setRowFlags(flags);
+            this.IntervalStartTime.set(DBTimeUtils.millisToTime(intervalStartTime));
+            this.IntervalDurationMs.set(intervalDurationMs);
+            this.TotalMemory.set(totalMemory);
+            this.FreeMemory.set(freeMemory);
+            this.IntervalCollections.set(intervalCollections);
+            this.IntervalCollectionTimeMs.set(intervalCollectionTimeMs);
+        }
+    }
+
+    @Override
+    protected String threadName() {
+        return TABLE_NAME;
+    }
+
+    private static final String[] columnNames;
+    private static final Class<?>[] columnDbTypes;
+
+    static {
+        final ColumnsSpecHelper cols = new ColumnsSpecHelper()
+                .add("IntervalStartTime", DBDateTime.class)
+                .add("IntervalDurationMs", long.class)
+                .add("TotalMemory", long.class)
+                .add("FreeMemory", long.class)
+                .add("IntervalCollections", long.class)
+                .add("IntervalCollectionTimeMs", long.class);
+        columnNames = cols.getColumnNames();
+        columnDbTypes = cols.getDbTypes();
+    }
+
+    @Override
+    protected ProcessMemoryLogLogger.ISetter createSetter() {
+        outstandingSetters.getAndIncrement();
+        return new ProcessMemoryLogLogger.DirectSetter();
+    }
+
+    public void log(
+            final long intervalStartTime,
+            final long intervalDurationMs,
+            final long totalMemory,
+            final long freeMemory,
+            final long intervalCollections,
+            final long intervalCollectionTimeMs) throws IOException {
+        log(DEFAULT_INTRADAY_LOGGER_FLAGS,
+                intervalStartTime, intervalDurationMs,
+                totalMemory, freeMemory,
+                intervalCollections, intervalCollectionTimeMs);
+    }
+
+    public void log(
+            final Row.Flags flags,
+            final long intervalStartTime, final long intervalDurationMs,
+            final long totalMemory, final long freeMemory,
+            final long intervalCollections, final long intervalCollectionTimeMs) throws IOException {
+        verifyCondition(isInitialized(), "init() must be called before calling log()");
+        verifyCondition(!isClosed, "cannot call log() after the logger is closed");
+        verifyCondition(!isShuttingDown, "cannot call log() while the logger is shutting down");
+        final ProcessMemoryLogLogger.ISetter setter = setterPool.take();
+        try {
+            setter.log(flags,
+                    intervalStartTime, intervalDurationMs,
+                    totalMemory, freeMemory,
+                    intervalCollections, intervalCollectionTimeMs);
+        } catch (Exception e) {
+            setterPool.give(setter);
+            throw e;
+        }
+        flush(setter);
+    }
+
+    private static final TableDefinition TABLE_DEFINITION = TableDefinition.tableDefinition(columnDbTypes, columnNames);
+
+    public static TableDefinition getTableDefinition() {
+        return TABLE_DEFINITION;
+    }
+
+    public static String[] getColumnNames() {
+        return columnNames;
+    }
+}
--- a/DB/src/main/java/io/deephaven/db/tablelogger/UpdatePerformanceLogLogger.java
+++ b/DB/src/main/java/io/deephaven/db/tablelogger/UpdatePerformanceLogLogger.java
@@ -52,8 +52,6 @@ public class UpdatePerformanceLogLogger
         RowSetter<Long> EntryIntervalModified;
         RowSetter<Long> EntryIntervalShifted;
         RowSetter<Long> EntryIntervalInvocationCount;
-        RowSetter<Long> TotalMemoryFree;
-        RowSetter<Long> TotalMemoryUsed;
         RowSetter<Long> EntryIntervalAllocatedBytes;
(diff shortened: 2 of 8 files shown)


rcaudy:
I don't understand the intent behind separating this from the UPL. I do like the added columns. They should be added to the query and operation performance log, too.

It might make sense to me to have a separate log table on a fixed interval, rather than tied to to the UPL, in addition to the UPL.

jcferretti:
Fixes #1382

DB/src/main/java/io/deephaven/db/v2/utils/TableLoggers.java
```diff
@@ -7,6 +7,12 @@
  * Tools to obtain internal, Deephaven logs as tables. These tables include query logs and performance logs.
  */
 public class TableLoggers {
+
+    static {
+        UpdatePerformanceTracker.start();
```
rcaudy, line 12: Why would we want the UPL on by default, but not the QPL or QOPL?  I thought `ProcessMemoryTracker` was the one we want as a baseline? I would put UPL back the way it was.

DB/src/main/java/io/deephaven/db/v2/utils/UpdatePerformanceTracker.java
```diff
@@ -362,9 +366,7 @@ public final void onUpdateEnd() {
             startCpuNanos = 0;
             startTimeNanos = 0;
 
-            final RuntimeMemory runtimeMemory = RuntimeMemory.getInstance();
-            totalFreeMemory = runtimeMemory.freeMemory();
-            totalUsedMemory = runtimeMemory.totalMemory();
+            RuntimeMemory.getInstance().read(endMemSample);
```
rcaudy, line 369: Put this before we start zeroing things out. Arguably it should be the first thing in the method.

DB/src/main/java/io/deephaven/db/v2/utils/UpdatePerformanceTracker.java
```diff
@@ -395,21 +397,33 @@ public String toString() {
 
         @Override
         public LogOutput append(final LogOutput logOutput) {
-            return logOutput.append("Entry{").append(", id=").append(id).append(", evaluationNumber=")
-                    .append(evaluationNumber).append(", operationNumber=").append(operationNumber)
+            return logOutput.append("Entry{").append(", id=").append(id)
+                    .append(", evaluationNumber=").append(evaluationNumber)
+                    .append(", operationNumber=").append(operationNumber)
                     .append(", description='").append(description).append('\'').append(", callerLine='")
```
rcaudy, line 403: Why did you correct this for everything except `callerLine`?

DB/src/main/java/io/deephaven/db/v2/utils/UpdatePerformanceTracker.java
```diff
@@ -178,6 +178,8 @@ private void finishInterval(final long intervalStartTimeMillis, final long inter
 
         boolean encounteredErrorLoggingToMemory = false;
 
+        long maxStartTimeNanos = 0;
```
rcaudy, line 181: These are never updated. It suggests to me that some part of this change is incomplete.

DB/src/main/java/io/deephaven/db/v2/utils/UpdatePerformanceTracker.java
```diff
@@ -190,12 +192,11 @@ private void finishInterval(final long intervalStartTimeMillis, final long inter
                 encounteredErrorLoggingToMemory =
                         logToMemory(intervalLevelDetails, entry, encounteredErrorLoggingToMemory);
             } else if (entry.intervalInvocationCount > 0) {
-                if (entry.totalUsedMemory > aggregatedSmallUpdatesEntry.totalUsedMemory) {
-                    aggregatedSmallUpdatesEntry.totalUsedMemory = entry.totalUsedMemory;
+                if (entry.startTimeNanos > maxStartTimeNanos) {
```
rcaudy, line 195: Related to the above, I do not think I correctly understand this code.

DB/src/main/java/io/deephaven/db/v2/utils/UpdatePerformanceTracker.java
```diff
@@ -300,8 +301,8 @@ public long getIntervalDurationNanos() {
         private long startAllocatedBytes;
         private long startPoolAllocatedBytes;
 
-        private long totalFreeMemory;
-        private long totalUsedMemory;
+        private RuntimeMemory.Sample startMemSample;
```
rcaudy, line 304: `final`.

DB/src/main/java/io/deephaven/db/v2/utils/ProcessMemoryTracker.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private void startThread() {
+        Thread driverThread = new Thread(new ProcessMemoryTracker.Driver(), "ProcessMemoryTracker.Driver");
+        driverThread.setDaemon(true);
+        driverThread.start();
+    }
+
+    public static void start() {
+        synchronized (UpdatePerformanceTracker.class) {
```
rcaudy, line 46: Wrong lock.

Also, I think you should just synchronize the method; it would be best if it didn't return until the instance has really started.

DB/src/main/java/io/deephaven/db/v2/utils/RuntimeMemory.java
```diff
@@ -88,6 +104,43 @@ public long totalMemory() {
...
+        void copy(final Sample s) {
+            freeMemory = s.freeMemory;
+            totalMemory = s.totalMemory;
+            totalCollections = s.totalCollections;
+            totalCollectionTimeMs = s.totalCollectionTimeMs;
+        }
+
+    }
+
+    public void read(Sample buf) {
```
rcaudy, line 136: I'm pretty sure we should hold a lock that prevents other calls to `maybeUpdateValues()` for the duration of our copy.

DB/src/main/java/io/deephaven/db/v2/utils/ProcessMemoryTracker.java
jcferretti: Instance?  This method is static.
I'll change both this and the UPT to have the return inside the synchronized block

DB/src/main/java/io/deephaven/db/v2/utils/RuntimeMemory.java
jcferretti: All these data members are volatile.  If you are concerned about getting a consistent snapshot, we didn't care before for Free/Total, so why would we care now?

DB/src/main/java/io/deephaven/db/v2/utils/UpdatePerformanceTracker.java
jcferretti: Look again now, fixed to original intention (made a mistake the first time around).

DB/src/main/java/io/deephaven/db/v2/utils/UpdatePerformanceTracker.java
jcferretti: You are correct, see next answer.

DB/src/main/java/io/deephaven/db/tables/utils/QueryPerformanceNugget.java
```diff
@@ -43,21 +43,19 @@
     private final long startTimeNanos;
     private final long startCpuNanos;
     private final long startUserCpuNanos;
-    private final long startFreeMemory;
-    private final long startTotalMemory;
+    private final RuntimeMemory.Sample startMemorySample = new RuntimeMemory.Sample();
```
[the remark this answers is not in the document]
jcferretti: Putting them in the constructorS (plural) is 2 lines of more code, seems redundant.  But, done.
What "end" fields?  You mean diff fields?  I moved them to their own section later, what about that?

DB/src/main/java/io/deephaven/db/tablelogger/UpdatePerformanceLogLogger.java
```diff
@@ -52,8 +52,12 @@ public static String getDefaultTableName() {
         RowSetter<Long> EntryIntervalModified;
         RowSetter<Long> EntryIntervalShifted;
         RowSetter<Long> EntryIntervalInvocationCount;
-        RowSetter<Long> TotalMemoryFree;
-        RowSetter<Long> TotalMemoryUsed;
+        RowSetter<Long> FreeMemory;
+        RowSetter<Long> TotalMemory;
+        RowSetter<Long> FreeMemoryChange;
```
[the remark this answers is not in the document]
jcferretti: We don't, the UPL does not aggregate the FreeMemory and TotalMemory, they show the value for the whole interval (or try to, and have tried, look at the previous code and the new code).   So to the extent that FreeMemory and TotalMemory make any sense in the UPL (I could agree with you that they don't), then Change* make the same sense.   None of these fields are needed really as they exist, and that was my whole point at removing them earlier.  But maybe that was never clear enough for you in our discussion.

DB/src/main/java/io/deephaven/db/v2/utils/RuntimeMemory.java
jcferretti: Nevermind, removed all those methods and made a truly-snapshotting version of the code.

DB/src/main/java/io/deephaven/db/tablelogger/UpdatePerformanceLogLogger.java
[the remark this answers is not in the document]
rcaudy: Free memory and total memory are sort of nice as a snapshot in context of the UPL. They tell you "at the end of the interval, this is how things stood". It's not totally useless. I don't think we should bother introducing the "change" columns.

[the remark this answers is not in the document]
jcferretti: I don't agree, but I don't care enough about this to argue more.  Yield.

DB/src/main/java/io/deephaven/db/tablelogger/ProcessMemoryLogLogger.java
```diff
@@ -0,0 +1,139 @@
...
+        RowSetter<DBDateTime> IntervalStartTime;
+        RowSetter<Long> IntervalDurationMs;
+        RowSetter<Long> TotalMemory;
+        RowSetter<Long> FreeMemory;
+        RowSetter<Long> IntervalCollections;
+        RowSetter<Long> IntervalCollectionTimeMs;
+
+        DirectSetter() {
+            IntervalStartTime = row.getSetter("IntervalStartTime", DBDateTime.class);
+            IntervalDurationMs = row.getSetter("IntervalDurationMs", long.class);
```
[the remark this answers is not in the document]
jcferretti: I care about the usability of the new table.   My whole point trying to do this is to make something easy to use for median user to be able to be self-served answering the question "Am I having a GC issue".   That column being in nanos gets in the way of that.  So I want to keep it in milliseconds.

If you /really/ care about this this much (I hope not, because I don't see something with specific units as such consistency issue, but you might), we could turn this into a percentage.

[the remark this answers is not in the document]
jcferretti: Btw what we obviously need here is a duration type that has nice rendering (like DBDateTime has nice rendering), and whatever underlying unit.

DB/src/main/java/io/deephaven/db/v2/utils/TableLoggers.java
rcaudy: I didn't think we started those either. I may be mistaken. If someone isn't going to do any performance introspection, we could leave these off by default.

DB/src/main/java/io/deephaven/db/v2/utils/RuntimeMemory.java
```diff
@@ -88,91 +107,84 @@ public static RuntimeMemory getInstance() {
...
+            currSnapshot = snapshot;
         }
-        if (logInterval > 0 && now >= nextLog) {
-            log.info().append("Jvm Heap: ").append(commaFormat.format(lastFreeMemory)).append(" Free / ")
-                    .append(commaFormat.format(lastTotalMemory)).append(" Total (")
+        buf.freeMemory = snapshot.lastFreeMemory;
+        buf.totalMemory = snapshot.lastTotalMemory;
+        buf.totalCollections = snapshot.totalCollections;
+        buf.totalCollectionTimeMs = snapshot.totalCollectionTimeMs;
+        if (logInterval > 0 && now >= snapshot.nextLog) {
```
rcaudy, line 176: I wonder if we should just delete this spammy crap now that we have a log table. It shows up at unpredictable intervals, anyway.

DB/src/main/java/io/deephaven/db/v2/utils/TableLoggers.java
```diff
@@ -75,4 +74,12 @@ public static QueryTable processInfoLog() {
     public static QueryTable processMemoryLog() {
         return ProcessMemoryTracker.getInstance().getQueryTable();
     }
+
+    /**
+     * Start collecting data for query update performance.
+     */
+    @ScriptApi
+    public static void startUpdatePerformanceLog() {
```
rcaudy: You are right, the QPL and QOPL are effectively started automatically the first time they have data.
I think we should start the UPT from `LiveTableMonitor.start()`; if it becomes a memory problem, we will make it configurable later.

DB/src/main/java/io/deephaven/db/v2/utils/TableLoggers.java
```diff
@@ -9,7 +9,6 @@
 public class TableLoggers {
 
     static {
-        UpdatePerformanceTracker.start();
         ProcessMemoryTracker.start();
```
rcaudy, line 12: Doing this via static initializer feels hacky. Should we explicitly make it part of init scripts for python and groovy? Or does this happen as a side effect of those?

DB/src/main/java/io/deephaven/db/v2/utils/TableLoggers.java
rcaudy: See new discussions. I admit that I was definitely wrong.

jcferretti:
Colors!

Deephaven - Google Chrome_112

DecimalFormat is not working, discussed with Bender and Colin, filed a bug: https://github.com/deephaven/deephaven-core/issues/1399

PR #1392, opened 2021-10-05
Format the client and server pydocs the same

chipkent:
The client and server pydocs were not formatted the same.  This resolves the inconsistencies.  

The best features were taken from both the client and server docs.  Source code and parameter types were taken from the client API, and the isolated module layout was taken from the server API.


sphinx/sphinx.gradle
```diff
@@ -8,12 +8,22 @@ description = 'Generates docs for the python libraries provided in Deephaven Cor
 
 evaluationDependsOn ':Integrations'
 
+tasks.register('copyPyClientWhl', Copy) {
+    dependsOn ':pyclient:buildPyClient'
```
niloc132, line 12: must have an `evaluationDependsOn ":pyclient"` if we do this, so that the task is sure to exist in time, otherwise there is a risk that pyclient could depend on us and just fail with NPEs when we build

sphinx/sphinx.gradle
```diff
@@ -8,12 +8,22 @@ description = 'Generates docs for the python libraries provided in Deephaven Cor
...
+    from "$rootDir/pyclient/build/wheel/"
+    into layout.buildDirectory.file('sphinx-image/wheel/')
+}
+
 def sphinxDockerfile = tasks.register('sphinxDockerfile', Dockerfile) {
+    dependsOn 'copyPyClientWhl'
```
niloc132, line 18: this should be in the sphinx image, it doesn't need to be here, and should be an input.files instead of just a dependsOn, roughly line 32 or so should get this:

```
inputs.files tasks.named('copyPyClientWhl').get().outputs.files
```

Moving this would let the dockerfile task be left if up to date, and ensure only the image gets rebuilt (this has also no measurable benefit, but a few ms here and a few ms there...)

Preferring the inputs... and ...outputs... lets gradle know why we depend on that task, and so if the upstream task re-runs but actually has no changes that we care about, this downstream task can avoid re-running. That is more of a win, since we get to avoid talking to docker, and risking re-running even more downstream work.

sphinx/sphinx.gradle
```diff
@@ -8,12 +8,22 @@ description = 'Generates docs for the python libraries provided in Deephaven Cor
 
 evaluationDependsOn ':Integrations'
 
+tasks.register('copyPyClientWhl', Copy) {
```
niloc132, line 11: unless anything else is writing to build/wheel, prefer Sync to Copy, so that when version numbers change we don't require a `./gradlew clean` to mop up old files

sphinx/sphinx.gradle
```diff
@@ -8,12 +8,22 @@ description = 'Generates docs for the python libraries provided in Deephaven Cor
...
+    dependsOn ':pyclient:buildPyClient'
+    from "$rootDir/pyclient/build/wheel/"
```
niloc132, line 13: consider replacing this and dependsOn with 
```
from(project(':pyclient').tasks.named('buildPyClient').get().outputs.files)
```
This one line describes both "I should rebuild when that task changes" and "I want to know the correct path to that task's outputs without hardcoding it here"

sphinx/sphinx.gradle
niloc132: this dependsOn is now unnecessary, the next line implicitly sets up that dependency

sphinx/lib/dh_sphinx.py
```diff
@@ -0,0 +1,96 @@
...
+        toctree = ""
+    else:
+        toctree = ".. toctree::\n"
+        for k in tree:
+            p = package.copy()
+            p.append(k)
+            pn = ".".join(p)
+            toctree += "%s%s <%s>\n"%(" "*4,k,pn)
+
+    rst = "%s\n%s\n\n%s\n.. automodule:: %s\n    :members:\n    :show-inheritance:\n    :special-members: __init__\n    :undoc-members:\n\n"%(package_name,"="*len(package_name),toctree,package_name)
```
jmao-denver, line 50: i think we should remove the  'special-members: __init__' option, at least for pydeephaven. I only want anything with a docstring to be documented, nothing else.

jmao-denver (approved):

PR #1393, opened 2021-10-06
Fix Kafka streaming referential integrity and shutdown processing

### LINKED ISSUE #1352: Stream table referential integrity bug
```
from deephaven import KafkaTools as kt

trades = kt.consumeToTable(
        { 'bootstrap.servers' : 'demo-kafka.c.deephaven-oss.internal:9092',
          'schema.registry.url' : 'http://demo-kafka.c.deephaven-oss.internal:8081' },
        'io.deephaven.crypto.kafka.TradesTopic',
        key = kt.IGNORE,
        value = kt.avro('io.deephaven.crypto.kafka.TradesTopic-io.deephaven.crypto.Trade'),
        table_type='stream')

quotes = kt.consumeToTable(
        { 'bootstrap.servers' : 'demo-kafka.c.deephaven-oss.internal:9092',
          'schema.registry.url' : 'http://demo-kafka.c.deephaven-oss.internal:8081' },
        'io.deephaven.crypto.kafka.QuotesTopic',
        key = kt.IGNORE,
        value = kt.avro('io.deephaven.crypto.kafka.QuotesTopic-io.deephaven.crypto.Quote'),
        table_type='stream')

trades_last_by = trades.lastBy("Exchange", "Instrument")

quotes_last_by = quotes.lastBy("Exchange", "Instrument")

del trades, quotes
```
@devinrsmith :
the last by tables eventually stopped ticking
potentially, a liveness issue in the dag?

I am confident that this is actually a referential integrity bug: the stream `Table` result does not maintain a reference to its parent, the `StreamToTableAdapter` (which is the `LiveTable` in this case).

rcaudy:
Fixes #1352


DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -228,9 +247,37 @@ public DynamicTable table() {
...
+            alive = false;
+            log.info().append("Deregistering ").append(StreamToTableAdapter.class.getSimpleName()).append('-').append(name)
+                    .endl();
+            liveTableRegistrar.removeTable(this);
+            final Runnable localShutdownCallback = shutdownCallback;
+            if (localShutdownCallback != null) {
+                localShutdownCallback.run();
+            }
+        }
+        // CODE REVIEW DISCUSSION: SHOULD WE ATTEMPT THIS CLEANUP? I THINK NOT.
```
cpwright, line 266: I think leaving the switch column sources alone makes sense; because we don't want to change the data of the table off the LTM refresh loop (or without our singularly allowed notification).  We could throw out our bufferChunkSources though.

Kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
```diff
@@ -161,19 +152,22 @@ public String toString() {
...
      *
-     * @param log a log for output
-     * @param props the properties used to create the {@link KafkaConsumer}
-     * @param topic the topic to replicate
-     * @param partitionToConsumer a function implementing a mapping from partition to its consumer of records.
-     * @param partitionToInitialSeekOffset a function implementing a mapping from partition to its intial seek offset,
+     * @param log A log for output
+     * @param props The properties used to create the {@link KafkaConsumer}
+     * @param topic The topic to replicate
+     * @param partitionToStreamConsumer A function implementing a mapping from partition to its consumer of records. The
```
cpwright, line 158: You are saying the implementations should not allocate resources, but I do not think that is actually true of what the implementations do.

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
rcaudy: I was worried they might be in use concurrently. Do you think that's a non-issue?

Kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
cpwright: I mean they definitely do stuff by creating the adapter and inserting stuff into a table.   The question about whether the comment or code is "right" is somewhat philosophical.

                final Pair<StreamToTableAdapter, ConsumerRecordToStreamPublisherAdapter> partitionAdapterPair =
                        adapterFactory.get();
                final Table partitionTable = tableConversion.apply(partitionAdapterPair.getFirst().table());
                streamTableMap.enqueueUpdate(() -> Assert.eqNull(streamTableMap.put(partition, partitionTable),
                        "streamTableMap.put(partition, partitionTable)"));
                return new SimpleKafkaStreamConsumer(partitionAdapterPair.getSecond(), partitionAdapterPair.getFirst());

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
cpwright: So the buffered sources are captured under a sync block; but are filled before that.  We could solve that by making a final refresh release resources if we are done; but then we can't deregister.  It is really a question of whether we care to release these things back to the pool, and if we actually had unit/component tests that could execute w/ chunk tracking then we would blow up.

Kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
rcaudy: Well, the comment has an "if appropriate" hedge. In this case, it's my intent to create the adapters and do the table map puts. Do you think I should re-word it somehow to make it clearer that its programmer's-choice? Existing comment pasted here for reference.
```
     * @param partitionToStreamConsumer A function implementing a mapping from partition to its consumer of records. The
     *        function will be invoked once per partition at construction; implementations should internally defer
     *        resource allocation until first call to {@link KafkaStreamConsumer#accept(Object)} or
     *        {@link KafkaStreamConsumer#acceptFailure(Exception)} if appropriate.
```

DB/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
rcaudy: I had decided that intentionally leaking the chunks and letting GC have them was "fine" in this case, since it's an occasional occurrence on shutdown and not a per-operation thing.
Either way we're leaking some chunks, since there will be some in the switch sources.

cpwright (dismissed):

PR #1414, opened 2021-10-07
Configure app mode to /app.d directory

devinrsmith:


devinrsmith:
James already made this change elsewhere.

PR #1432, opened 2021-10-11
Build and publish grpc-api-tensorflow

devinrsmith:


jjbrosnan (dismissed):

jjbrosnan:
I tested this locally and had no issues.  The image with TensorFlow is ~2.5 GB, so we would likely want to instruct users to increase the memory available to them if they use this image.  As far as "if this will break things", I don't know the answer related to the required testing and the repository itself.

TensorFlow v2.6.0 requires "numpy~=1.19.2".  I believe that running `pip install tensorflow` goes to 1.19.5 because it's the latest 1.19 release version.  I've tested TensorFlow with the older NumPy version more extensively with some local images previously and ran into no issues.  There are some threads as to why TensorFlow doesn't work with NumPy 1.20 or later, but there is little information on when this will be fixed.

grpc-api/server/docker/build.gradle
```diff
@@ -113,4 +115,29 @@ dockerCreateDockerfile.dependsOn grpcApiBase
...
 
+// Create a simple file for docker/build-push-action to reference, tag, and publish
+tasks.register('publishDockerfile', Dockerfile) {
+  def input = dockerBuildImage.imageIdFile
+  def output = layout.buildDirectory.file('publish/Dockerfile')
+
+  it.inputs.file input
+  it.outputs.file output
+
+  it.destFile.set output
```
niloc132, line 126: i think the outputs.file is not needed above when you have this line here, as destFile is marked as an output already

grpc-api/server/docker/build.gradle
```diff
@@ -113,4 +115,29 @@ dockerCreateDockerfile.dependsOn grpcApiBase
 assemble.dependsOn dockerBuildImage
 dockerBuildImage.outputs.upToDateWhen { t -> return Docker.isImageUpToDate(t) }
 
+// Create a simple file for docker/build-push-action to reference, tag, and publish
+tasks.register('publishDockerfile', Dockerfile) {
+  def input = dockerBuildImage.imageIdFile
```
niloc132, line 120: why an intermediate variable?

grpc-api/server/docker/build.gradle
```diff
@@ -113,4 +115,29 @@ dockerCreateDockerfile.dependsOn grpcApiBase
...
+  def output = layout.buildDirectory.file('publish/Dockerfile')
```
niloc132, line 121: gitignore this, and maybe mark it for clean to mop up?

niloc132 (commented):
I realize this is deliberately going outside gradle's usual control over the process - probably worth spending time later seeing if we can tell the github action to just use the images already built, or see if the DockerPushImage gradle task can be leveraged instead.

For now, just a little feedback to simplify/clarify the specific gradle changes, can iterate on the rest later.

grpc-api/server/docker/build.gradle
devinrsmith: It should already be ignored, given it's under the build directory, I think.

devinrsmith:
Superceded by https://github.com/deephaven/deephaven-core/pull/1602

PR #1440, opened 2021-10-12
Provide gRPC api for input tables

niloc132:
This pull request is broken into three commits to start (with JS API client code to follow):
 * ClassUtil improvements
 * Proposed MutableInputTable api enhancements
 * Proposed gRPC api and implementation

ClassUtil got its lookupClass from the dataobjects ClassUtil, but this implementation isn't thread safe, and while it supports both Class.getName() and Class.getCanonicalName() array notation, it only supports getName() inner type notation. Additionally, it initializes the class on first mention, which could mean for example that an arbitrary doPut could force the server to load a class, even if not serializable. This commit still supports (ignores) generic arguments provided to the class, but otherwise delegates to common-lang3 to actually find the class, and provide the class instance without initializing it.

The MutableInputTable changes are intended to make it easier to use from gRPC on the server, and removes some details where the DHE openapi was used to provide implementation details of the input table.

The gRPC API adds a createInputTable call in table service (still has some TODOs for discussion), and provides a separate gRPC service to interact with the input table itself.


grpc-api/src/main/java/io/deephaven/grpc_api/table/ops/CreateInputTableGrpcImpl.java
```diff
@@ -0,0 +1,84 @@
...
+    public Table create(CreateInputTableRequest request, List<SessionState.ExportObject<Table>> sourceTables) {
+        TableDefinition tableDefinitionFromSchema;
+
+        if (request.hasSchema()) {
+            Message message = Message.getRootAsMessage(request.getSchema().asReadOnlyByteBuffer());
+            if (message.headerType() != MessageHeader.Schema) {
+                throw GrpcUtil.statusRuntimeException(Code.INVALID_ARGUMENT,
+                        "Must specify schema header in schema message");
+            }
+            tableDefinitionFromSchema = BarrageUtil.convertArrowSchema((Schema) message.header(new Schema())).tableDef;
```
devinrsmith, line 65: Looks like from this context we don't have to worry about the magic 8 bytes?

I wonder if we should try to use the utility code, or generalize the utility code, in io.deephaven.grpc_api.util.SchemaHelper and/or BarrageUtil instead of doing it inline here?

nbauernfeind (dismissed):

DB/src/main/java/io/deephaven/db/util/config/MutableInputTable.java
```diff
@@ -36,7 +46,50 @@
...
+            if (colDef == null) {
+                error.append("Key Column \"").append(keyColumn).append("\" does not exist.\n");
+            } else if (!colDef.isCompatible(thisColDef)) {
+                error.append("Key Column \"").append(keyColumn).append("\" is not compatible.\n");
+            }
+        }
+        final List<String> extraKeys = keyDefinition.getColumnNames().stream().filter(kd -> !keyNames.contains(kd))
+                .collect(Collectors.toList());
+        if (!extraKeys.isEmpty()) {
+            error.append("Unknown key columns: ").append(extraKeys);
```
devinrsmith, line 83: Do lists append correctly?

niloc132: To my knowledge, all Lists have a toString that calls toString on children (unlike java arrays), so yes, this will list the contents? Is that not the case?

This is the same code as before, just moved out to the interface so that consumers like the grpcapi can directly ask if an operation is legal rather than relying on "guess and find out"

grpc-api/src/main/java/io/deephaven/grpc_api/table/ops/CreateInputTableGrpcImpl.java
niloc132: Added a new method to SchemaHelper to unwrap an arbitrary ByteBuffer, then pass to BarrageUtil to convert to the TableDefinition.

devinrsmith (approved):

nbauernfeind (dismissed):

devinrsmith (approved):

PR #1490, opened 2021-10-27
Enable casting PyObject's to primitives; provide type param info for py lists

### LINKED ISSUE #1009: ClassCastException when casting a value retrieved from a Python list
When creating a table using `emptyTable`, we want the resulting X column to be an int column. We've tried a few variations of queries, but either it executes and the column does NOT become an int column (verified w/ `getMeta`), or we get an exception.

Jianfeng tried the following: 

```
empty = emptyTable(5)
a = [1, 2, 3, 4, 10]
result2 = empty.update("X = (int)a[ (int) i ]")
```
I tried casting it to int and got this exception

```
caused by java.lang.ClassCastException: org.jpy.PyObject cannot be cast to java.lang.Integer
	at io.deephaven.db.tables.lang.DBLanguageFunctionUtil.intCast(DBLanguageFunctionUtil.java:18609)
	at io.deephaven.temp.cm302811563987228486v52_0.Formula.applyFormulaPerItem(Formula.java:177)
	... 33 more
```

nbauernfeind:
Also did this for String/boolean/Boolean:
```python
x = ['0', '1', '42', None]
y = TableTools.emptyTable(1).update("X = (String)x[i % x.size()]")
print(y.getDefinition())
```

Note that after some discussion we preferred to NPE when casting to boolean and the value is a None. We also decided that a String cast should throw an exception if not already a String (instead of converting via `PyObject::str`).

Fixes #1009


DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageFunctionUtil.java
```diff
@@ -19289,6 +19290,121 @@ public static short shortCast(Object a) {
...
+            throw new IllegalArgumentException("provided value is not a PyObject");
+        }
+        PyObject o = (PyObject) a;
+        if (o.isNone()) {
+            return null;
+        }
+        return o.getStringValue();
+    }
+
+    public static boolean booleanPyCast(Object a) {
```
rcaudy, line 19384: Ask Chip and Jianfeng about this. I think `null` and `isNone()` should probably NPE.

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageParser.java
```diff
@@ -997,11 +997,22 @@ public static boolean isDbArray(Class<?> type) {
             Class<?> paramType = n.getIndex().accept(this, printer);
             printer.append(')');
 
-            if (DbArray.class.isAssignableFrom(type) && (n.getName() instanceof NameExpr)) {
-                Class<?> ret = variableParameterizedTypes.get(((NameExpr) n.getName()).getNameAsString())[0];
+            // We'll check for a known component type if this a NameExpr.
+            if (n.getName() instanceof NameExpr) {
```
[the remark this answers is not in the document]
rcaudy: Not too terrible. Let's ship as-is and revisit if we get usability complaints.
```
Full expression           : x[ii]
Expression having trouble : io.deephaven.db.tables.lang.DBLanguageParser$VisitArgs@551360e4
Exception message         : Cannot find method get(long) in class io.deephaven.dynamic.Script_6$1
```

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageParser.java
```diff
@@ -1166,14 +1177,27 @@ else if (fromBoxedType) {
...
+         * There is no "booleanCast()" function. However, we do try to cast to String and Boolean from PyObjects.
          *
          * There are also no special functions for the identity conversion -- e.g. "intCast(int)"
          */
-        if (toPrimitive && !ret.equals(boolean.class) && !ret.equals(exprType)) {
+        final boolean isPyUpgrade =
+                ((ret.equals(boolean.class) || ret.equals(Boolean.class) || ret.equals(String.class))
+                        && exprType.equals(PyObject.class));
+
+        if (toPrimitive && !ret.equals(boolean.class) && !ret.equals(exprType) || isPyUpgrade) {
```
rcaudy, line 1188: Maybe add clarifying parentheses.

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageFunctionUtil.java
```diff
@@ -19375,7 +19375,7 @@ public static String doStringPyCast(Object a) {
         if (o == null || o.isNone()) {
             return null;
         }
-        return o.getStringValue();
+        return o.str();
```
rcaudy, line 19378: I'm a little nervous about this change. The previous impl was more like a String cast, this is more like a toString(). Semantically, the former feels more correct for this use case. Maybe @chipkent or @jmao-denver can offer guidance as to whether it's more Pythonic to do a string conversion instead of cast.

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageFunctionGenerator.java
```diff
@@ -558,6 +560,118 @@ public static void main(String args[]) {
...
+
+        buf.append("        public static float floatPyCast(Object a) {\n");
+        buf.append("            if (a != null && !(a instanceof PyObject)) {\n");
+        buf.append("                throw new IllegalArgumentException(\"Provided value is not a PyObject\");\n");
+        buf.append("            }\n");
+        buf.append("            PyObject o = (PyObject) a;\n");
+        buf.append("            if (o == null || o.isNone()) {\n");
+        buf.append("                return QueryConstants.NULL_FLOAT;\n");
+        buf.append("            }\n");
+        buf.append("            return (float) o.getDoubleValue();\n");
```
chipkent, line 605: This raises a question.  How should these casts behave if they contain the `NULL_<TYPE>` value, instead of just `None`?  I assume this value should be treated the same as `null` or `None`.

```
x = NULL_DOUBLE
t = emptyTable(3).update("X = (float) x")
```

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageFunctionUtil.java
```diff
@@ -19289,6 +19290,116 @@ public static short shortCast(Object a) {
...
+    public static double doublePyCast(Object a) {
+        if (a != null && !(a instanceof PyObject)) {
+            throw new IllegalArgumentException("Provided value is not a PyObject");
+        }
+        PyObject o = (PyObject) a;
+        if (o == null || o.isNone()) {
+            return QueryConstants.NULL_DOUBLE;
+        }
+        return o.getDoubleValue();
+    }
```
chipkent, line 19313: See comment in `DBLanguageFunctionGenerator`.

chipkent (changes requested):

rcaudy (dismissed):

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageFunctionGenerator.java
```diff
@@ -558,6 +560,118 @@ public static void main(String args[]) {
             append(buf, castFromObjFormatter, BinaryExpr.Operator.PLUS, c, Object.class);
         }
 
+        // Special casts for PyObject to primitive
+        buf.append("        public static int intPyCast(Object a) {\n");
```
chipkent, line 564: There is some weirdness here.  Let me try to summarize.
* Python 3 has numeric type changes vs Python 2.  In Python 3, float is double, and int is of unlimited size.
* Our code is frequently using getIntValue , instead of getLongValue.  I assume this is a python 2 relic.  I'm guessing that getIntValue is doing a cast to the int32 range.  The cast to the int32 range could potentially drop null values in the int64 range, which is closer to the python `int`.
* For float, NULL_DOUBLE makes sense for the null value in python.  For int, we could argue that the null value could be NULL_LONG , since that is the closest reasonable value... but this null value is in the middle of the int range.  Ugh.
* Now for more diceyness.  In our casts, we don't treat objects differently if they are float or int.  To properly handle special null values, you would need to call getDoubleValue on float and getLongValue on int.
This leaves us in a weird place, if we try to do any NULL_<TYPE> value handling.

chipkent: Maybe the handling of special values for python types is too hard, given difference in the type system.

chipkent: We discussed this and think that we can't properly handle the special values in a clean way in Python, so Python will use the `None` value as the null identifier.

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageFunctionGenerator.java
```diff
@@ -558,6 +560,118 @@ public static void main(String args[]) {
...
+
+        buf.append("        public static float floatPyCast(Object a) {\n");
+        buf.append("            if (a != null && !(a instanceof PyObject)) {\n");
+        buf.append("                throw new IllegalArgumentException(\"Provided value is not a PyObject\");\n");
+        buf.append("            }\n");
+        buf.append("            PyObject o = (PyObject) a;\n");
+        buf.append("            if (o == null || o.isNone()) {\n");
+        buf.append("                return QueryConstants.NULL_FLOAT;\n");
+        buf.append("            }\n");
+        buf.append("            return floatCast(o.getDoubleValue());\n");
```
chipkent, line 605:
```suggestion
        buf.append("            return (float) o.getDoubleValue();\n");
```
The change is needed to avoid the spurious handling of special null values.

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageFunctionGenerator.java
```diff
@@ -558,6 +560,118 @@ public static void main(String args[]) {
...
+
+        buf.append("        public static char charPyCast(Object a) {\n");
+        buf.append("            if (a != null && !(a instanceof PyObject)) {\n");
+        buf.append("                throw new IllegalArgumentException(\"Provided value is not a PyObject\");\n");
+        buf.append("            }\n");
+        buf.append("            PyObject o = (PyObject) a;\n");
+        buf.append("            if (o == null || o.isNone()) {\n");
+        buf.append("                return QueryConstants.NULL_CHAR;\n");
+        buf.append("            }\n");
+        buf.append("            return charCast(o.getIntValue());\n");
```
chipkent, line 616:
```suggestion
        buf.append("            return (char) o.getIntValue();\n");
```

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageFunctionGenerator.java
```diff
@@ -558,6 +560,118 @@ public static void main(String args[]) {
...
+
+        buf.append("        public static byte bytePyCast(Object a) {\n");
+        buf.append("            if (a != null && !(a instanceof PyObject)) {\n");
+        buf.append("                throw new IllegalArgumentException(\"Provided value is not a PyObject\");\n");
+        buf.append("            }\n");
+        buf.append("            PyObject o = (PyObject) a;\n");
+        buf.append("            if (o == null || o.isNone()) {\n");
+        buf.append("                return QueryConstants.NULL_BYTE;\n");
+        buf.append("            }\n");
+        buf.append("            return byteCast(o.getIntValue());\n");
```
chipkent, line 627:
```suggestion
        buf.append("            return (byte) o.getIntValue();\n");
```

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageFunctionGenerator.java
```diff
@@ -558,6 +560,118 @@ public static void main(String args[]) {
...
+
+        buf.append("        public static short shortPyCast(Object a) {\n");
+        buf.append("            if (a != null && !(a instanceof PyObject)) {\n");
+        buf.append("                throw new IllegalArgumentException(\"Provided value is not a PyObject\");\n");
+        buf.append("            }\n");
+        buf.append("            PyObject o = (PyObject) a;\n");
+        buf.append("            if (o == null || o.isNone()) {\n");
+        buf.append("                return QueryConstants.NULL_SHORT;\n");
+        buf.append("            }\n");
+        buf.append("            return shortCast(o.getIntValue());\n");
```
chipkent, line 638:
```suggestion
        buf.append("            return (short) o.getIntValue();\n");
```

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageFunctionUtil.java
```diff
@@ -19289,6 +19290,116 @@ public static short shortCast(Object a) {
...
+
+    public static float floatPyCast(Object a) {
+        if (a != null && !(a instanceof PyObject)) {
+            throw new IllegalArgumentException("Provided value is not a PyObject");
+        }
+        PyObject o = (PyObject) a;
+        if (o == null || o.isNone()) {
+            return QueryConstants.NULL_FLOAT;
+        }
+        return floatCast(o.getDoubleValue());
```
chipkent, line 19334:
```suggestion
        return (float) o.getDoubleValue());
```

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageFunctionUtil.java
```diff
@@ -19289,6 +19290,116 @@ public static short shortCast(Object a) {
...
+
+    public static char charPyCast(Object a) {
+        if (a != null && !(a instanceof PyObject)) {
+            throw new IllegalArgumentException("Provided value is not a PyObject");
+        }
+        PyObject o = (PyObject) a;
+        if (o == null || o.isNone()) {
+            return QueryConstants.NULL_CHAR;
+        }
+        return charCast(o.getIntValue());
```
chipkent, line 19345:
```suggestion
        return (char) o.getIntValue();
```

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageFunctionUtil.java
```diff
@@ -19289,6 +19290,116 @@ public static short shortCast(Object a) {
...
+
+    public static byte bytePyCast(Object a) {
+        if (a != null && !(a instanceof PyObject)) {
+            throw new IllegalArgumentException("Provided value is not a PyObject");
+        }
+        PyObject o = (PyObject) a;
+        if (o == null || o.isNone()) {
+            return QueryConstants.NULL_BYTE;
+        }
+        return byteCast(o.getIntValue());
```
chipkent, line 19356:
```suggestion
        return (byte) o.getIntValue();
```

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageFunctionUtil.java
```diff
@@ -19289,6 +19290,116 @@ public static short shortCast(Object a) {
...
+
+    public static short shortPyCast(Object a) {
+        if (a != null && !(a instanceof PyObject)) {
+            throw new IllegalArgumentException("Provided value is not a PyObject");
+        }
+        PyObject o = (PyObject) a;
+        if (o == null || o.isNone()) {
+            return QueryConstants.NULL_SHORT;
+        }
+        return shortCast(o.getIntValue());
```
chipkent, line 19367:
```suggestion
        return (short) o.getIntValue();
```

chipkent (changes requested):

DB/src/main/java/io/deephaven/db/tables/lang/DBLanguageFunctionGenerator.java
nbauernfeind: We noted that we may need to verify that the data accessors prefer to give python None over NULL_INT etc. Otherwise, agreed that None is best.

chipkent (approved):

rcaudy (approved):

Issue #1493, opened 2021-10-27
add-opens Docker application based on plugin conventions


devinrsmith:
Broken off from https://github.com/deephaven/deephaven-core/issues/1369.

Our `apply plugin: 'io.deephaven.java-open-nio'` logic doesn't currently apply to the docker application plugin, so we need to manually configure instead of relying on the plugin. It would be nice if we could automatically hook in.

Alternatively, we may be moving away from the built-in docker application plugin and creating a more bespoke entrypoint in the future (https://github.com/deephaven/deephaven-core/discussions/1105) in which case this might become more actionable as well.

PR #1494, opened 2021-10-27
Upgrade runtime images to JDK17, fixes #1369

### LINKED ISSUE #1369: Upgrade runtime images to JDK17
Part of #1364 

The docker base images should change from JDK8 -> JDK17. We may have to be careful exactly which base image we choose, as we likely need python3.7 support explicitly. https://github.com/deephaven/deephaven-core/pull/1213

This will also have the effect of updating our integration testing.

Should happen after #1368

devinrsmith:
Notes:
 * Follow up #1493: add-opens Docker application based on plugin conventions
 * The base buster maven package does not correctly handle JDK17, needed to install a newer version from buster-backports
 * The zulu17-jdk-headless base image is installed instead of relying on Azul; this allows us to stay on buster (important atm for python3.7), but to also fix a locales issue (https://github.com/zulu-openjdk/zulu-openjdk/issues/71).

### DIFF
--- a/DB-test/build.gradle
+++ b/DB-test/build.gradle
@@ -38,7 +38,9 @@ dependencies {
 def gradleWrapper = tasks.register("dockerGradleInit", Wrapper.class) { wrapper ->
   wrapper.scriptFile "${buildDir}/template-project/gradlew"
   wrapper.jarFile "${buildDir}/template-project/gradle/wrapper/gradle-wrapper.jar"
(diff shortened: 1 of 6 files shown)


niloc132 (approved):
Suggest removing that version check entirely, but otherwise looks good. 51 minutes locally to build via qemu for arch with some very minor changes, 11 minutes to build for amd64, both seem to work great.

PR #1498, opened 2021-10-28
Upgrade to Arrow 6.0.1

nbauernfeind:
Need to run through some more formal testing; looking to merge this after release.


devinrsmith (approved):

devinrsmith:
Should we look into upgrading pyarrow too?

Couple places in the source:

pyclient/requirements.txt, pyclient/build.gradle, sphinx/sphinx.gradle

@jmao-denver the versions also seem to not be the same.

jmao-denver:
It is definitely a good idea to upgrade and make all versions the same.

nbauernfeind:
@jmao-denver can't think of a good reason not to update them all at once. Updating now to kick off that CI build.

devinrsmith (approved):

Issue #1370, opened 2021-09-29
Upgrade application target from JDK8 -> JDK11


devinrsmith:
Part of #1364 

This will likely be one of the last parts of the epic. Depending on how we've solved the transition to the JDK17 compiler, we may have additional code updating work here, or new compiler flags, or both.

PR #1502, opened 2021-10-28
Fix deprecated compile and similar gradle invocations

devinrsmith:
Related to #1370


devinrsmith:
Upgrading gradle will likely need a completely new pass, given the engine and jetty changes.

PR #1516, opened 2021-11-01
Fix up char filters

mofojed:
In MatchFilter, just accept a char that is passed with double quotes in addition to single quotes


nbauernfeind (approved):
I restarted the long check which originally segfaulted. You might need to rebase if it fails again.

PR #1523, opened 2021-11-03
Java gatherer

jjbrosnan:
The java gatherer transfers table data to a Python object.  It works in Python using the `frombuffer` method, for which NumPy and Torch have methods.  This is acceptable, because these are the two most common data containers that will be used with AI.

Speed testing between between Chip's v1 and v2 showed mixed results.  This code is his v2, because it showed better speedups more often than v1.  There was enough of both to justify including two methods, but that seems unnecessary given both showed speedups between 5x and 250x over the legacy method.

I have tested for memory leaks and differences between outputs of legacy and new code and have been able to find none.


jjbrosnan:
Speedup using `learn` can be checked:

```
from deephaven import learn
import numpy as np
import time
import sys

end, start = 0, 0
legacy_elapsed, new_elapsed = 0, 0

test_samples = [1001, 10001, 100001, 500001, 1000001]

for n_samples in test_samples:

    print(str(n_samples) + " rows.")

    source = emptyTable(n_samples).update("X = (i / n_samples) * 2 * Math.PI", "Y = X")

    def compute_sin(x):
        return np.sin(x)

    def table_to_numpy(idx, cols):

        global end, start, legacy_elapsed
        start = time.time()
    
        return_array = np.empty([idx.getSize(), len(cols)], dtype = float)
        iter = idx.iterator()
        i = 0
        while (iter.hasNext()):
            it = iter.next()
            j = 0
            for col in cols:
                return_array[i, j] = col.get(it)
                j += 1
            i += 1

        end = time.time()
        legacy_elapsed = end - start

        return np.squeeze(return_array)

    def new_table_to_numpy(idx, cols):

        global end, start, new_elapsed

        start = time.time()
        buffer = learn.gatherer.create_2d_tensor(idx, cols)
        tensor = np.frombuffer(buffer, dtype = float)
        tensor.shape = (idx.getSize(), len(cols))
        end = time.time()

        new_elapsed = end - start

        return np.squeeze(tensor)

    def numpy_to_table(data, idx):
        return data[idx]

    result_legacy = learn.learn(
        table = source,
        model_func = compute_sin,
        inputs = [learn.Input("X", table_to_numpy)],
        outputs = [learn.Output("SinX", numpy_to_table)],
        batch_size = n_samples
    )
    print("LEGACY: Time elapsed: " + str(legacy_elapsed) + " seconds.")

    result_v1 = learn.learn(
        table = source,
        model_func = compute_sin,
        inputs = [learn.Input("X", new_table_to_numpy)],
        outputs = [learn.Output("SinX", numpy_to_table)],
        batch_size = n_samples
    )
    speedup = np.round(legacy_elapsed / new_elapsed, 2)
    print("NEW: Time elapsed: " + str(new_elapsed) + " seconds.  This is " + str(speedup) + "x faster.")
```

First run shows speedups:

1,001 rows: 13.2x
10,001 rows: 36.14x
100,001 rows: 110.48x
500,001 rows: 91.23x
1,000,001 rows: 171.12x

Integrations/python/deephaven/learn/__init__.py
```diff
@@ -14,6 +14,7 @@
 _Output_ = None
 _Computer_ = None
 _Scatterer_ = None
+gatherer = None
```
chipkent, line 17: I think what you have here functions, but I don't think it is what we want to deliver.  Let me try to explain some of the vision.

1) We want the UX to be very pythonic.  By exposing the Java `Gatherer`, we have Java naming conventions and other mess that Jianfeng is trying hard to get rid of.

2) This only supports `float64` tensors.  We need to support the spectrum of tensor types.

3) From the Python side, there is way too mutch futzing around to do what you want.  I think we need a module or modules that contain key helper functions.  In this case, maybe the module is `gather` or `input`, and it would contain various utilities for dealing with input data.  

I would guess that these helper functions look something like the following.
```
def buffer_2d(idx, cols, dtype):
    ...

def numpy_2d(idx, cols, dtype):
    buffer =buffer_2d(idx, cols, dtype)
    tensor = np.frombuffer(buffer, dtype = float)
    tensor.shape = (idx.getSize(), len(cols))
    return tensor
```

Then the user code is as simple as:
```
from deephaven.learn.gather import numpy_2d

my_np_gatherer = lambda idx, cols: numpy_2d(idx, cols, np.int32)
my_tf_gatherer = labda idx, cols: tf.convert_to_tensor(numpy_2d(idx, cols, np.float64))
```

Many different `dtype` specifications should be supported.  numpy types?  DH types?  etc. 

4) The java method currently returns a method shaped like the table.  Should there be an argument to return the transpose for efficiency or convenience?  Does `numpy.transpose` or the equivalent TensorFlow or PyTorch command result in extra memory allocation or slowness from indirection?

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Copy data from a table into a 2d tensor.
+     *
+     * @param indexSet indices of the rows of the table to put into the tensor
+     * @param columnSources columns of data to put into the tensor
+     * @return contiguous RAM allocated for the tensor. When a numpy tensor is passed in for this argument, jpy will
+     *         handle passing the memory reference as a 1d java array here.
+     */
+    public static double[] create_2d_tensor(final IndexSet indexSet, final ColumnSource<?>[] columnSources) {
```
chipkent, line 15: 1) since this is Java, should it be camel case?
2) What other arguments should be provided here?  e.g. transpose
3) Better name?  `tensor_buffer_2d`?
4) Support other output types

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -0,0 +1,37 @@
+package io.deephaven.integrations.learn;
+
+import io.deephaven.db.v2.sources.ColumnSource;
+
+public class Gatherer {
```
chipkent, line 5: Everything here needs to be 100% unit tested.

chipkent (changes requested):

jjbrosnan:
So, here's an example of usage in a Python query:

```
# Deephaven imports
from deephaven import transferrer
from deephaven import learn

# Standard imports
import numpy as np

num_rows = 1000

# Create a table with some data
source = emptyTable(num_rows).update("X = (float)0.01 * i", "Y = (float)sqrt(X)")

my_np_gatherer = lambda idx, cols: transferrer.table_to_numpy_float(idx, cols)

def my_np_scatterer(data, idx):
    return data[idx]

# Define a function that sums rows in a NumPy array
def add_columns(columns):
    results = np.array([])
    for row in columns:
        results = np.append(results, np.sum(row))
    return results

# Apply the add_columns function to source with learn
result = learn.learn(
    table = source,
    model_func = add_columns,
    inputs = [learn.Input(["X", "Y"], my_np_gatherer)],
    outputs = [learn.Output("Z", my_np_scatterer)],
    batch_size = num_rows
)
```

It's more in line with what Chip specified, and a step in the right direction.

jjbrosnan:
Here's an example of a Python query with the current code:

```
# Deephaven imports
from deephaven.transferrer import table_to_numpy_2d
from deephaven import transferrer
from deephaven import learn

# Standard imports
import numpy as np

num_rows = 1000

# Create a table with some data
source = emptyTable(num_rows).update("X = (double)i", "Y = (double)(2 * i)")

my_np_gatherer = lambda idx, cols: table_to_numpy_2d(idx, cols, np.double)

def my_np_scatterer(data, idx):
    return data[idx]

# Define a function that sums rows in a NumPy array
def add_columns(columns):
    results = np.array([])
    for row in columns:
        results = np.append(results, np.sum(row))
    return results

# Apply the add_columns function to source with learn
result = learn.learn(
    table = source,
    model_func = add_columns,
    inputs = [learn.Input(["X", "Y"], my_np_gatherer)],
    outputs = [learn.Output("Z", my_np_scatterer)],
    batch_size = num_rows
)

```

jjbrosnan:
In the current instantiation of the `transferrer` submodule (I'm not married to that name, I just can't think of anything better), the user will explicitly set the numpy dtype using a lambda function.

I have tested this for all data types and have only found issues with boolean values.  I believe this is because the `getBoolean` method returns a `java.lang.Boolean` and not the primitive `boolean` type.

Right now, the following NumPy and Python data types are supported:

Python built in: int, float, boolean
NumPy: bool, byte, float, double, int, short, long (and all of the respective aliases)

Integrations/python/deephaven/learn/__init__.py
jjbrosnan: The latest push has everything discussed here except for point 4.  I'm not sure what exactly how we would want to implement a transpose or other operation.  Let's discuss it soon

chipkent: Transpose is very easy.  It would be something like:

```
                final int ij = transpose ? nRows * j + i: nCols * i + j;
```
And then you would also need to set the transposed shape.

chipkent: I think this `gatherer` line is not needed.

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -0,0 +1,223 @@
...
```
chipkent, line 5: Javadoc

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -0,0 +1,223 @@
...
+    /**
+     * Copy data from a table into a 2d tensor of Booleans.
+     *
+     * @param indexSet indices of the rows of the table to put into the tensor
+     * @param columnSources columns of data to put into the tensor
+     * @return contiguous RAM allocated for the tensor. When a numpy tensor is passed in for this argument, jpy will
+     *         handle passing the memory reference as a 1d java array here.
+     */
+
+    public static Boolean[] booleanTensorBuffer2D(final IndexSet indexSet, final ColumnSource<?>[] columnSources) {
```
chipkent, line 16: 1) I think you want this to return `boolean[]`, since `Boolean[]` is going to be a mess in python and won't work with numpy -- for sure.  Should this throw an exception on null values?
2) I suggest naming these `tensorBuffer2D<type>` since they will then be ordered better in an IDE.

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -0,0 +1,223 @@
...
+        final int nRows = indexSet.getSize();
+        final int nCols = columnSources.length;
+        final Boolean[] tensor = new Boolean[nRows * nCols];
+
+        int j = 0;
+        for (ColumnSource<?> cs : columnSources) {
+            int i = 0;
+
+            for (long idx : indexSet) {
+                final int ij = nCols * i + j;
```
chipkent, line 27: See transpose note.

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -0,0 +1,223 @@
...
+        final int nCols = columnSources.length;
+        final byte[] tensor = new byte[nRows * nCols];
+
+        int j = 0;
+        for (ColumnSource<?> cs : columnSources) {
+            int i = 0;
+
+            for (long idx : indexSet) {
+                final int ij = nCols * i + j;
+                tensor[ij] = cs.getByte(idx);
```
chipkent, line 59: For `NULL_<TYPE>` values, should we offer an option to (1) do what we do here, (2) raise an exception or (3) map to a value?

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -0,0 +1,223 @@
...
+    /**
+     * Copy data from a table into a 2d tensor of shorts.
+     *
+     * @param indexSet indices of the rows of the table to put into the tensor
+     * @param columnSources columns of data to put into the tensor
+     * @return contiguous RAM allocated for the tensor. When a numpy tensor is passed in for this argument, jpy will
+     *         handle passing the memory reference as a 1d java array here.
+     */
+
+    public static short[] shortTensorBuffer2D(final IndexSet indexSet, final ColumnSource<?>[] columnSources) {
```
chipkent, line 201: Other than what is noted above, it seems like the only feature we are missing is the ability to cast to types other than the column types.  I think we can handle this in a fairly clean way.  Right now, we are doing `cs.getShort(idx)`.  

Thinking about this for just a minute, my best shooting from the hip design would be something using inner classes like:
```
class Gatherer {
  // interface for getting typed values from column sources
  private interface Getter {
    bool asBool(ColumnSource<?> cs, long idx){...}
    int asInt(ColumnSource<?> cs, long idx){...}
    ...
  }

  // utilities for getting values from boolean columns
  private static class GetterBool {
    bool asBool(ColumnSource<?> cs, long idx){...}
    int asInt(ColumnSource<?> cs, long idx){...}
    ...
  }
}
```

Then inside the tensor buffer functions, you could make a `Getter[]` that is filled in based upon the column source types.

Other designs could also work.  This was just the first one that came to me.

Integrations/src/main/java/io/deephaven/integrations/learn/Scatterer.java
```diff
@@ -48,9 +48,9 @@ public Object scatter(int idx, FutureOffset fo) {
 
         for (int i = 0; i < outputs.length; i++) {
             // TODO: TICKET #1026: Replace typeString with the following once ticket #1009 is resolved:
-            // final String typeString = outputs[i].getType() == null ? "" : "(" + outputs[i].getType() + ")";
+            final String typeString = outputs[i].getType() == null ? "" : "(" + outputs[i].getType() + ")";
 
-            final String typeString = "";
+            // final String typeString = "";
```
chipkent, line 53: Before merging, this should get cleaned up.

Integrations/python/deephaven/transferrer/__init__.py
```diff
@@ -0,0 +1,81 @@
+#
```
chipkent, line 1: I would probably move this to `learn/getter` or something like that.  I don't know that these are general-purpose methods, since they work on column sources and `learn.IndexSet`.  Also the name `transfer` is confusing when we use `getter` in the documentation.

Integrations/python/deephaven/transferrer/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+"""
+Utilities for transfer between Python objects and Deephaven Tables
+"""
+
+import numpy as np
+import jpy
+import wrapt
+
+# None until the first _defineSymbols() call
+gatherer = None
```
chipkent, line 13: any reason to be public?

Integrations/python/deephaven/transferrer/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+        buffer = gatherer.doubleTensorBuffer2D(idx, cols)
+    elif np_dtype == np.intc:
+        buffer = gatherer.intTensorBuffer2D(idx, cols)
+    elif np_dtype == np.int_:
+        buffer = gatherer.longTensorBuffer2D(idx, cols)
+    elif np_dtype == np.short:
+        buffer = gatherer.shortTensorBuffer2D(idx, cols)
+    elif np_dtype == np.single:
+        buffer = gatherer.floatTensorBuffer2D(idx, cols)
+
```
chipkent, line 77: not handling the error condition.  Also, clearly not testing the condition.  We need to target very high code coverage in the Java and python.

Integrations/python/deephaven/transferrer/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+    tensor = np.frombuffer(buffer, dtype = np_dtype)
+    tensor.shape = (idx.getSize(), len(cols))
+
+    return np.squeeze(tensor)
```
chipkent, line 81: What is the logic for having squeeze here vs requiring the user to add it, if they really want it?  This feels like we're making a design decision for them.

chipkent (changes requested):

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
jjbrosnan: I'm not sure how we want to do null handling.  I didn't include it on this update, though it can be added soon

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
jjbrosnan: Leaving this comment open for now if we decide to go with a design like this in the future.

jjbrosnan:
Here's Python code to test the `getter` submodule:

```
import numpy as np
from deephaven import getter
gatherer = jpy.get_type("io.deephaven.integrations.learn.Gatherer")

def make_test_inputs(n_rows):
    t = emptyTable(n_rows).update("X = (double)i", "Y = (double)(2 * i)")
    col_sources = [ t.getColumnSource(col) for col in ["X", "Y"] ]
    index_set = gatherer.makeIndexSet(t)
    return t, col_sources, index_set

t, col_src, idx_set = make_test_inputs(10)

test_array = getter.table_to_numpy_2d(idx_set, col_src, np.double)
print(test_array)
test_array_2 = getter.table_to_numpy_2d(idx_set, col_src, np.double, transpose = True)
print(test_array_2)
```

Something interesting happens if you change the `np.dtype` to something other than the corresponding Java primitive.  For instance, if you attempt to use `np.single` (float) in place of `np.double` in the code above, Python thinks there are 40 data points instead of 20 due to the doubled memory allocation for a double vs a single.  Similar behavior happens with other data types as well.  Thus, any type conversions should happen on the Java end if we decide to implement them.

Integrations/src/main/java/io/deephaven/integrations/learn/Scatterer.java
```diff
@@ -47,10 +47,8 @@ public Object scatter(int idx, FutureOffset fo) {
         String[] queryStrings = new String[outputs.length];
 
         for (int i = 0; i < outputs.length; i++) {
-            // TODO: TICKET #1026: Replace typeString with the following once ticket #1009 is resolved:
             final String typeString = outputs[i].getType() == null ? "" : "(" + outputs[i].getType() + ")";
 
-            // final String typeString = "";
             queryStrings[i] = String.format("%s = %s (__scatterer.scatter(%d, %s))", outputs[i].getColName(),
                     typeString, i, futureOffsetColName);
```
chipkent: I think this is out of date with the fix I put into main.  I'm guessing you need a rebase.

Integrations/python/deephaven/transferrer/__init__.py
```diff
@@ -10,16 +10,16 @@
 import wrapt
```
chipkent: I think I have noted this in earlier comments.  Since this package is very specific to gatherers, I suggest moving the `deephaven.transferrer` to be `deephaven.learn.gather`.

Integrations/python/deephaven/transferrer/__init__.py
```diff
@@ -61,21 +61,29 @@ def table_to_numpy_2d(idx, cols, np_dtype = None):
         np_dtype = np.intc
 
     if np_dtype == np.bool_:
```
chipkent: I would reorder this section to be ordered by type (e.g. int) from small to large (e.g. float, double).  It just makes reading the code easier later.

Integrations/python/deephaven/transferrer/__init__.py
```diff
@@ -61,21 +61,29 @@ def table_to_numpy_2d(idx, cols, np_dtype = None):
...
     tensor = np.frombuffer(buffer, dtype = np_dtype)
-    tensor.shape = (idx.getSize(), len(cols))
+    if not transpose:
+        tensor.shape = (idx.getSize(), len(cols))
+    else:
+        tensor.shape = (len(cols), idx.getSize())
 
-    return np.squeeze(tensor)
\ No newline at end of file
+    if squeeze:
```
chipkent: It is good that this is optional.  Does it make sense to even be here?  A user could easily do something like the following without increasing the complexity of the code.  
```
tensor = np.squeeze(table_to_numpy_2d(...))
```
Having the option increases the complexity of the functionality.  It also causes a subtle bit of confusion.  The method has 2D in the name.  Without squeeze, we get a 2D tensor.  If you use the squeeze flag, now you get a 1D tensor.  Now the docs are not right.

Did the benefit of the argument make up for the problems?

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -1,30 +1,55 @@
...
+     */
+    public static IndexSet makeIndexSet(final Table t) {
+        final IndexSet is = new IndexSet(t.intSize());
+
+        for(long idx : t.getIndex()) {
+            is.add(idx);
+        }
+
+        return is;
+    }
```
chipkent: The comment here makes this sound like something critical for testing that is just in the wrong permanent location.  Does this need to be in some unit test related functionality?

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -1,30 +1,55 @@
...
+
     /**
      * Copy data from a table into a 2d tensor of Booleans.
      *
      * @param indexSet indices of the rows of the table to put into the tensor
      * @param columnSources columns of data to put into the tensor
+     * @param transpose a boolean indicating the user's intent to return a transposed Python object
```
chipkent, line 34:
```suggestion
     * @param transpose true to return a tensor buffer in transposed order; false to return the tensor buffer in regular order.
```
Apply the change everywhere.
1) Javadoc knows the types, so you don't need that in doc strings.
2) Every parameter is expressing the user's intent, so that does not need to be said.
3) No code here is python specific.  We're just making a general memory buffer.

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -1,30 +1,55 @@
...
+        final boolean[] tensor = new boolean[nRows * nCols];
 
         int j = 0;
-        for (ColumnSource<?> cs : columnSources) {
+        for (ColumnSource<?> cs: columnSources) {
             int i = 0;
 
             for (long idx : indexSet) {
-                final int ij = nCols * i + j;
+                final int ij = transpose ? nRows * j + i : nCols * i + j;
```
chipkent, line 51: Have you printed some output to verify that this does create the transposed tensor?

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -1,30 +1,55 @@
...
      *
      * @param indexSet indices of the rows of the table to put into the tensor
      * @param columnSources columns of data to put into the tensor
+     * @param transpose a boolean indicating the user's intent to return a transposed Python object
      * @return contiguous RAM allocated for the tensor. When a numpy tensor is passed in for this argument, jpy will
      *         handle passing the memory reference as a 1d java array here.
      */
-
-    public static Boolean[] booleanTensorBuffer2D(final IndexSet indexSet, final ColumnSource<?>[] columnSources) {
+    public static boolean[] tensorBuffer2Dboolean(final IndexSet indexSet,
```
chipkent:
```suggestion
    public static boolean[] tensorBuffer2DBoolean(final IndexSet indexSet,
```
Camel case.  You don't need to do lower case for primitive.

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -40,11 +65,13 @@
      *
      * @param indexSet indices of the rows of the table to put into the tensor
      * @param columnSources columns of data to put into the tensor
+     * @param transpose a boolean indicating the user's intent to return a transposed Python object
      * @return contiguous RAM allocated for the tensor. When a numpy tensor is passed in for this argument, jpy will
      *         handle passing the memory reference as a 1d java array here.
      */
-
-    public static byte[] byteTensorBuffer2D(final IndexSet indexSet, final ColumnSource<?>[] columnSources) {
+    public static byte[] TensorBuffer2Dbyte(final IndexSet indexSet,
```
chipkent:
```suggestion
    public static byte[] tensorBuffer2DByte(final IndexSet indexSet,
```
Functions start lower case.  Apply renamings everywhere.

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
chipkent: It seems like there are a few options:
1) just cram our null values into the tensor.  This will be fast but relies on the user to have thought out how they want nulls to behave.
2) raise an exception on null values to force the user to deal with them before this function.
3) provide a default value to use for null values.  Somewhat of a compromise.

The three options are not exclusive.  The function could potentially support them all reasonably efficiently if you are careful.  Thoughts on which ones we should support?

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
chipkent: I just looked at all of the changed files.  There are no unit tests.  That seems related to this note.  We should have 100% test coverage of the Java code that has been added.  Then we should target reasonably high coverage (>50%) for the python.

chipkent (changes requested):

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
jjbrosnan: I have run tests in Python to verify that the transpose works properly.  It does.  I did, however, just run some interesting tests that didn't give the results I expected.  These results support not even bothering with this, and dealing with it only in Python.

NumPy arrays have a built in transpose:

```
a = np.array([[1, 2, 3], [4, 5, 6]])
a_transpose = a.T
```

Using the *.T method to transpose is showing slightly faster performance, and that's without setting the transpose flag in Java, which means it's using the base `ij` index.

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -60,31 +60,24 @@ def table_to_numpy_2d(idx, cols, np_dtype = None, transpose = False, squeeze = F
     elif np_dtype == int:
         np_dtype = np.intc
 
-    if np_dtype == np.bool_:
-        buffer = _gatherer.tensorBuffer2Dboolean(transpose, idx, cols,)
+    if np_dtype == bool:
```
chipkent, line 63: should this be `np.bool_` as it was before?

Integrations/python/test/test_learn.py
```diff
@@ -0,0 +1,304 @@
...
+
+import numpy as np
+import unittest
+import sys
+import os
+
+from deephaven import learn, tableToDataFrame, TableTools
+from deephaven.learn import gather
+
+class TestLearn(unittest.TestCase):
```
chipkent, line 17: I have some general comments on this Python test.  I'm making them as teaching comments to make better tests and code going forward.
1) The code is not very local, in the sense that data is spread far from where it is used -- variable scope is large.  It would be cleaner if the data was defined where it is used (e.g. in the test function), since I think it is only used in one place.
2) There is a ton of code duplication.  The test methods could be refactored to something like:
```
def base_test(...):
    ...
def test_byte():
    table = ...
    ...
    base_test(table, ...)
```
3) `learn` should not appear in these tests.  A unit test should be testing that a function is returning what it should.  By injecting `learn`, this is becoming more of an integration test... and also increasing complexity.
4) I assume this test should be called `test_learn_gather.py` or something, since that is what you are really trying to test, and not learn.  You could have a `test_learn.py` to just test `learn`, but it should be very simple and should not be testing `gather`.
5) There is transpose stuff in the unit tests.  Not sure why that is here, since transpose was removed.

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
chipkent: To easily see what is currently going on with nulls, write a unit test.

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -93,29 +89,28 @@ public static IndexSet makeIndexSet(final Table t) {
...
      * @param columnSources columns of data to put into the tensor
-     * @param transpose a boolean indicating the user's intent to return a transposed Python object
      * @return contiguous RAM allocated for the tensor. When a numpy tensor is passed in for this argument, jpy will
      *         handle passing the memory reference as a 1d java array here.
      */
-    public static double[] tensorBuffer2Ddouble(final boolean transpose,
-            final IndexSet indexSet,
+
+    public static short[] tensorBuffer2DShort(final IndexSet indexSet,
             final ColumnSource<?>[] columnSources) {
```
chipkent: move these onto one line?

Integrations/src/test/java/io/deephaven/integrations/learn/GathererTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+        for (long idx : t.getIndex()) {
+            is.add(idx);
+        }
+
+        return is;
+    }
+
+    @Test
+    public void gatherTestMethod() {
```
chipkent, line 54: There should be a separate test method for each method.  Right now, a failure in one type will mask failures in other types, which increases the time to create a fix.

Integrations/src/test/java/io/deephaven/integrations/learn/GathererTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+        boolean[] booleanArray = {true, true, false, false, true, false, true, false};
+        byte[] byteArray = {(byte)1, (byte)2, (byte)3, (byte)4, (byte)5, (byte)6, (byte)7, (byte)8};
+        short[] shortArray = {(short)-1, (short)-2, (short)-3, (short)-4, (short)-5, (short)-6, (short)-7, (short)-8};
+        int[] intArray = {100, 200, -100, -200, -300, -400, 300, 400};
+        long[] longArray = {1L, 100L, 10000L, 1000000L, 9L, 999L, 99999L, 9999999L};
+        float[] floatArray = {3.14F, 2.73F, 1.5F, 0.63F, 0.1F, 0.2F, 0.3F, 0.4F};
+        double[] doubleArray = {3.14, 2.73, 1.5, 0.63, 0.1, 0.2, 0.3, 0.4};
```
chipkent, line 62: This is all duplicated above.  DRY.
https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

Integrations/src/test/java/io/deephaven/integrations/learn/GathererTest.java
```diff
@@ -0,0 +1,113 @@
+package io.deephaven.integrations.learn;
```
chipkent, line 1: When you are developing these tests, make sure you run them in IJ with code coverage.  It will let you see if there are any parts of your code that are not tested well.

Integrations/src/test/java/io/deephaven/integrations/learn/GathererTest.java
```diff
@@ -0,0 +1,113 @@
...
+
```
chipkent, line 2: Tests are failing becuase the autoformatter was not run.
```
Execution failed for task ':Integrations:spotlessJavaCheck'.
```

chipkent (changes requested):

Integrations/src/test/java/io/deephaven/integrations/learn/GathererTest.java
jjbrosnan: I've run the tests with code coverage, but I don't really know how to interpret the results.  I'll spend some more time on it tomorrow as well

Integrations/python/test/test_learn_gather.py
```diff
@@ -0,0 +1,196 @@
...
+    def boolean_model(self, features):
+        return np.count_nonzero(features, axis = 1) < 2
+
+    # Model for learn to use when dtype = [np.byte, np.short, np.intc, np.int_]
+    def whole_model(self, features):
+        return np.sum(features, axis = 1)
+
+    # Model for learn to use when dtype = [np.single, np.double]
+    def decimal_model(self, features):
+        return np.prod(features, axis = 1)
```
chipkent, line 120: not used?

Integrations/python/test/test_learn_gather.py
```diff
@@ -0,0 +1,196 @@
...
+
+    # The base test, which other tests will be built from
+    def base_test(self, np_dtype):
+        if np_dtype == np.bool_:
+            source = self.bool_table
+            model = self.boolean_model
```
chipkent, line 126: I don't think any of these are used.

chipkent (changes requested):

chipkent (approved):

jjbrosnan:
Currently, `deephaven.learn` uses an `IndexSet` as the first argument to the function that gathers data from a Table to a Python object.  This PR creates a flat Java array using a `RowSequence`.  I'm not sure if we should make the change in this PR or another one, but `deephaven.learn` needs to be updated to use a `RowSequence`, and not an `IndexSet`.  Either that, or convert an `IndexSet` to a `RowSequence` under the hood (provided that's possible and not horribly inefficient).

Integrations/python/test/test_learn_gather.py
```diff
@@ -0,0 +1,200 @@
...
+        return np.sum(features, axis = 1)
+
+    # Model for learn to use when dtype = [np.single, np.double]
+    def decimal_model(self, features):
+        return np.prod(features, axis = 1)
+
+    # Test boolean data types
+    def test_boolean(self):
+        # self.base_test(np_dtype = np.bool_)
+        return
```
chipkent, line 115: This test should either be commented out or removed.  As it is, it will run and pass, which gives the false sense that booleans are being tested.

Also, see the comment above.  If `bool` is broken, the python method should not run as if it is fine.  If it is fine, the unit test should not be broken.

Integrations/python/test/test_learn_gather.py
```diff
@@ -0,0 +1,200 @@
...
+            model = self.whole_model
+        elif np_dtype == np.int_:
+            source = self.long_table
+            model = self.whole_model
+        elif np_dtype == np.single:
+            source = self.float_table
+            model = self.decimal_model
+        elif np_dtype == np.double:
+            source = self.double_table
+            model = self.decimal_model
```
chipkent, line 163: This if-block is funky.  It seems like the code would be much more simple if this function just accepted the source and model as input.  By making that small change, you can ditch 20 lines of convoluted code.  e.g. `test_double` knows that the type is a double, but you then call a function that re-determines that the type is a double.

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+    elif dtype == float:
+        dtype = np.double
+    elif dtype == int:
+        dtype = np.intc
+
+    if dtype == np.bool_:
+        if order == 0:
+            buffer = _gatherer.tensorBuffer2DBooleanColumns(row_set, col_set)
+        elif order == 1:
+            buffer = _gatherer.tensorBuffer2DBooleanRows(row_set, col_set)
```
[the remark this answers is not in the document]
chipkent: After some thinking about this, and the `order` constant, I think you can massively simplify things by:
1) Creating a Java function called `tensorBuffer2DBoolean` with a bool argument for row vs column major.  This function would then call either `tensorBuffer2DBooleanColumnMajor` or `tensorBuffer2DBooleanRowMajor`.

2) Use a Python enum with values to for `order`.  This link below gives some details.
https://docs.python.org/3/library/enum.html#:~:text=Planet-,%C2%B6,-If%20__new__()%20or

It would look something like:
```
class MemoryLayout(Enum):
    ROW_MAJOR = (True)
    COLUMN_MAJOR = (False)
    C = (True) 
    FORTRAN = (False)

    def __init(self, is_row_major)
        self.is_row_major = is_row_major

def table_to_numpy_2d(row_set, col_set, order:MemoryLayout=MemoryLayout.ROW_MAJOR, dtype:numpy.dtype = numpy.float64) -> numpy.ndarray
```

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+    :param dtype: The desired NumPy data type of the output NumPy array
+    :return: A NumPy ndarray
+    """
+
+    if dtype == bool:
+        dtype = np.bool_
+    elif dtype == float:
+        dtype = np.double
+    elif dtype == int:
+        dtype = np.intc
```
chipkent, line 62: Is this block needed, or does numpy automatically handle this for us?  I have seen some comments that imply things will work without this block, but I'm not sure.

If we do need this conversion, I would make it into a function, because I'm sure we'll need it elsewhere as well.

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -0,0 +1,543 @@
+package io.deephaven.integrations.learn;
```
chipkent, line 1: See note above about renaming/moving this class to help furtureproof.

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -0,0 +1,543 @@
...
+
+    private static final int COPY_CHUNK_SIZE = 2048;
+
+    /**
+     * Copy data from a table into a 2d tensor of Booleans in column-major order.
+     *
+     * @param rowSeq indices of the rows of the table to put into the tensor
+     * @param columnSources columns of data to put into the tensor
+     * @return contiguous RAM allocated for the tensor. When a numpy tensor is passed in for this argument, jpy will
+     *         handle passing the memory reference as a 1d java array here.
```
chipkent, line 25:
```suggestion
     * @return contiguous RAM allocated for the tensor.
```
No way to pass in an array.  Apply this to all other locations.

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -0,0 +1,543 @@
...
+
+        return tensor;
+    }
+
+    /**
+     * Copy data from a table into a 2d tensor of Booleans in row-major order.
+     *
+     * @param rowSeq indices of the rows of the table to put into the tensor
+     * @param columnSources columns of data to put into the tensor
+     * @return contiguous RAM allocated for the tensor. When a numpy tensor is passed in for this argument, jpy will
```
chipkent, line 56: fix

Integrations/src/main/java/io/deephaven/integrations/learn/Gatherer.java
```diff
@@ -0,0 +1,543 @@
...
+                        tensor[ti++] = inputColumnValues[ci].get(ri);
+                    }
+                }
+                inputSharedContext.reset();
+            }
+        }
+
+        return tensor;
+    }
+
```
chipkent, line 542: Just a note.  I'm redoing the primitives library.  This has some new code autogeneration stuff that will help you avoid having to hard-code these functions that only differ by types.  This is for future reference only.

Integrations/src/test/java/io/deephaven/integrations/learn/GathererTest.java
```diff
@@ -0,0 +1,167 @@
...
+        // Data should be stored in column-major order
+        int idx = 0;
+        for (int i = 0; i < 2; i++) {
+            for (int j = 0; j < 4; j++) {
+                Assert.assertEquals("i=" + i + " j=" + j, expected.apply(i, j), actual.apply(idx));
+                idx++;
+            }
+        }
+    }
+
```
chipkent, line 106: We either need to remove/comment-out the boolean method or test it here.  We don't want a user to use it if we think it is broken or untested.

chipkent (changes requested):

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+        _gatherer = jpy.get_type("io.deephaven.integrations.learn.gather.NumPy")
+    
+class MemoryLayout(enum.Enum):
+    ROW_MAJOR = True
+    COLUMN_MAJOR = False
+    C = True
+    FORTRAN = False
+    def __init__(self, is_row_major):
+        self.is_row_major = is_row_major
+Layout = MemoryLayout(True)
```
chipkent, line 33:
```suggestion
```
Looks like it shouldn't be here.

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+    if _gatherer is None:
+        _gatherer = jpy.get_type("io.deephaven.integrations.learn.gather.NumPy")
+    
+class MemoryLayout(enum.Enum):
+    ROW_MAJOR = True
+    COLUMN_MAJOR = False
+    C = True
+    FORTRAN = False
+    def __init__(self, is_row_major):
+        self.is_row_major = is_row_major
```
chipkent, line 32: Missing pydocs

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+@_passThrough
+def convert_to_numpy_dtype(dtype):
+    if dtype == bool:
+        dtype = np.bool_
+    elif dtype == float:
+        dtype = np.double
+    elif dtype == int:
+        dtype = np.intc
+    else:
+        raise ValueError("{} is not a data type that can be converted to a NumPy dtype.".format(dtype))
```
chipkent, line 66: If it is already a numpy dtype, just return it.  This will simplify other code.

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+    return dtype
+
+@_passThrough
+def table_to_numpy_2d(row_set, col_set, order = Layout.ROW_MAJOR, dtype:np.dtype = np.intc):
```
chipkent, line 70:
```suggestion
def table_to_numpy_2d(row_set, col_set, order:MemoryLayout = MemoryLayout.ROW_MAJOR, dtype:np.dtype = np.double):
```

chipkent: I think you can also add the return type to the signature.

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+
+    if order in [Layout.ROW_MAJOR, Layout.C, True]:
+        flag = True
+    elif order in [Layout.COLUMN_MAJOR, Layout.FORTRAN, False]:
+        flag = False
+    else:
+        raise ValueError("Invalid major order.  Please use an True, False, or an enum value from MemoryLayout.")
+
+    if dtype == bool or dtype == float or dtype == int:
+        dtype = convert_to_numpy_dtype(dtype)
```
chipkent, line 91:
```suggestion
    dtype = convert_to_numpy_dtype(dtype)
```
See comment on this function.

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+
+    global Layout
+
+    if order in [Layout.ROW_MAJOR, Layout.C, True]:
+        flag = True
+    elif order in [Layout.COLUMN_MAJOR, Layout.FORTRAN, False]:
+        flag = False
+    else:
+        raise ValueError("Invalid major order.  Please use an True, False, or an enum value from MemoryLayout.")
```
chipkent, line 88: 1) We should use typing so that order must be a `MemoryLayout`.  The `True` / `False` stuff is asking for user bugs.
2) We don't need the `flag` value.  We can just use `order.is_row_major` in the function calls below.
3) With those two changes, we don't need this if-block.

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+    elif dtype == np.single:
+
+        buffer = _gatherer.tensorBuffer2DFloat(row_set, col_set, flag)
+
+    elif dtype == np.double:
+
+        buffer = _gatherer.tensorBuffer2DDouble(row_set, col_set, flag)
+
+    else:
+
```
chipkent, line 118: Take out excessive white space.

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+        raise ValueError("Data type {input_type} is not supported.".format(input_type = dtype))
```
chipkent, line 119:
```suggestion
        raise ValueError(f"Data type {dtype} is not supported.")
```
f-strings are easier to read than format strings.

Integrations/src/main/java/io/deephaven/integrations/learn/gather/NumPy.java
```diff
@@ -0,0 +1,668 @@
...
+    /**
+     * Copy data from a table into a 2d tensor of Booleans.
+     *
+     * @param rowSeq indices of the rows of the table to put into the tensor
+     * @param columnSources columns of data to put into the tensor
+     * @param ColumnMajorOrder the major order of the operation
+     * @return contiguous RAM allocated for the tensor
+     */
+    public static boolean[] tensorBuffer2DBoolean(final RowSequence rowSeq, final ColumnSource<?>[] columnSources,
+            boolean ColumnMajorOrder) {
```
chipkent, line 28: `ColumnMajorOrder` should be `columnMajorOrder` to comply with our naming conventions.  Same change elsewhere.

Integrations/src/main/java/io/deephaven/integrations/learn/gather/NumPy.java
```diff
@@ -0,0 +1,668 @@
...
+    }
+
+    /**
+     * Copy data from a table into a 2d tensor of Booleans in column-major order.
+     *
+     * @param rowSeq indices of the rows of the table to put into the tensor
+     * @param columnSources columns of data to put into the tensor
+     * @return contiguous RAM allocated for the tensor
+     */
+    public static boolean[] tensorBuffer2DBooleanColumnMajor(final RowSequence rowSeq,
```
chipkent, line 45: Should the `ColumnMajor` and `RowMajor` functions be public or private?  I don't have a strong view, but have a slight lean to minimizing the exposed API.

Integrations/src/main/java/io/deephaven/integrations/learn/gather/NumPy.java
```diff
@@ -0,0 +1,668 @@
...
+public class NumPy {
+
+    private static final int COPY_CHUNK_SIZE = 2048;
+
+    /**
+     * Copy data from a table into a 2d tensor of Booleans.
+     *
+     * @param rowSeq indices of the rows of the table to put into the tensor
+     * @param columnSources columns of data to put into the tensor
+     * @param ColumnMajorOrder the major order of the operation
```
chipkent, line 24:
```suggestion
     * @param columnMajorOrder true to return a column-major array; false to return a row-major array.
```
apply everywhere.

Integrations/src/test/java/io/deephaven/integrations/learn/gather/GathererTest.java
```diff
@@ -0,0 +1,167 @@
+package io.deephaven.integrations.learn.gather;
```
chipkent, line 1: Reanme to `NumPyTest`.  File and class need renaming.

chipkent (changes requested):

Integrations/python/deephaven/learn/__init__.py
```diff
@@ -55,6 +55,10 @@ def _passThrough(wrapped, instance, args, kwargs):
     _defineSymbols()
     return wrapped(*args, **kwargs)
 
+try:
+    _defineSymbols()
+except Exception as e:
+    pass
```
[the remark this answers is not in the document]
jmao-denver: there are other modules that are doing the same. It is probably not needed but harmless nonetheless with minimal performance penalty. In v2, I have removed the use of it.

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -56,18 +64,25 @@ def _passThrough(wrapped, instance, args, kwargs):
...
+        return dtype
+    elif dtype == bool:
         dtype = np.bool_
     elif dtype == float:
         dtype = np.double
     elif dtype == int:
         dtype = np.intc
     else:
-        raise ValueError("{} is not a data type that can be converted to a NumPy dtype.".format(dtype))
+        raise ValueError("{dtype} is not a data type that can be converted to a NumPy dtype.")
```
chipkent, line 81:
```suggestion
        raise ValueError(f"{dtype} is not a data type that can be converted to a NumPy dtype.")
```

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -78,45 +93,25 @@ def table_to_numpy_2d(row_set, col_set, order = Layout.ROW_MAJOR, dtype:np.dtype
...
-    global Layout
-
-    if order in [Layout.ROW_MAJOR, Layout.C, True]:
-        flag = True
-    elif order in [Layout.COLUMN_MAJOR, Layout.FORTRAN, False]:
-        flag = False
-    else:
-        raise ValueError("Invalid major order.  Please use an True, False, or an enum value from MemoryLayout.")
+    if not(MemoryLayout.has_value(order)):
+        raise ValueError("Invalid major order {order}.  Please use an enum value from MemoryLayout.")
```
chipkent, line 97:
```suggestion
        raise ValueError(f"Invalid major order {order}.  Please use an enum value from MemoryLayout.")
```

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -78,45 +93,25 @@ def table_to_numpy_2d(row_set, col_set, order = Layout.ROW_MAJOR, dtype:np.dtype
...
 
-    if dtype == bool or dtype == float or dtype == int:
-        dtype = convert_to_numpy_dtype(dtype)
+    dtype = convert_to_numpy_dtype(dtype)
 
     if dtype == np.byte:
-
-        buffer = _gatherer.tensorBuffer2DByte(row_set, col_set, flag)
-
+        buffer = _gatherer.tensorBuffer2DByte(row_set, col_set, order)
```
chipkent, line 102: I think this might not be doing what you want.  I'm guessing python is doing truthiness on `order`, and since order is always non-null, I think it is defaulting to `True`.  I think you want:

```suggestion
        buffer = _gatherer.tensorBuffer2DByte(row_set, col_set, order.is_row_major)
```

I could be wrong on this one.  If I'm right, apply the change everywhere.

https://www.freecodecamp.org/news/truthy-and-falsy-values-in-python/

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -23,14 +23,22 @@ def _defineSymbols():
...
 class MemoryLayout(enum.Enum):
+    """
+    Class with boolean values for memory layout of an array
+
+    :param enum.Enum: Base class for creating enumerated constants
+    """
     ROW_MAJOR = True
     COLUMN_MAJOR = False
     C = True
     FORTRAN = False
```
chipkent: Need docs

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -23,14 +23,22 @@ def _defineSymbols():
     if _gatherer is None:
         _gatherer = jpy.get_type("io.deephaven.integrations.learn.gather.NumPy")
     
+# A class defining the memory layout of an array
```
chipkent, line 26:
```suggestion
```
Docstring should contain all of the key info.

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -23,14 +23,22 @@ def _defineSymbols():
...
 class MemoryLayout(enum.Enum):
+    """
+    Class with boolean values for memory layout of an array
```
chipkent, line 29:
```suggestion
    Memory layouts for an array.
```

Integrations/python/deephaven/learn/gather/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+    :param enum.Enum: Base class for creating enumerated constants
+    """
+    ROW_MAJOR = True
+    COLUMN_MAJOR = False
+    C = True
+    FORTRAN = False
+    def __init__(self, is_row_major):
+        self.is_row_major = is_row_major
+    @classmethod
+    def has_value(cls, value):
```
chipkent, line 40:
```suggestion
    FORTRAN = False
    
    def __init__(self, is_row_major):
        self.is_row_major = is_row_major
        
    @classmethod
    def has_value(cls, value):
```
Whitespace for readability.

chipkent: Document `has_value`

chipkent (changes requested):

chipkent (approved):

PR #1526, opened 2021-11-03
Update to latest gradle-docker plugin

niloc132:
This will add support for the platform flag to be passed to docker build and run. This is now possible to upgrade since we have Java 11 as the baseline to run our gradle build at all.


niloc132:
Whoops, https://github.com/deephaven/deephaven-core/pull/1515 already exists.

PR #1541, opened 2021-11-09
Implemented the Table class wrapper, fixes #1486

### LINKED ISSUE #1486: Hand-wrap the Table class in Python
Works to be done:
1. wrap all the Table ops, properties
2. wrap Table metadata
3. create unit tests
4. Docstrings for everything public

jmao-denver:
1. Wrapped  the Table class and those of its methods that are most usable by Python
2. Added agg.py, constants.py
2. Added test_table.py, test_table.csv to the test suite
3. Added Python docstrings for all public methods/properties in the new code
4. Reorganized the source to make it more pep8 compliant
5. Fixed a bug in test environment configuration to enable ops on time_tables
6. Minor changes to the Python client API to keep it consistent with the server package


configs/dh-defaults.prop
```diff
@@ -133,7 +133,7 @@ Webapi.persistentquery.heap.default=4
 OneClick.allowedPattern.USym=^[A-Za-z0-9/.]+$
 OneClick.enforceUpperCaseColumns=USym
 
-LiveTableMonitor.checkTableOperations=true
+LiveTableMonitor.checkTableOperations=false
```
devinrsmith, line 136: Why is this necessary?

pyintegration/deephaven2/table.py
```diff
@@ -2,28 +2,85 @@
...
 
+import jpy
+
 from deephaven2 import DHError
 from deephaven2.column import Column
+from deephaven2.combo_agg import ComboAggregation
+from deephaven2.constants import SortDirection
+
+TableTools = jpy.get_type("io.deephaven.db.tables.utils.TableTools")
+SortPair = jpy.get_type("io.deephaven.db.tables.SortPair")
```
devinrsmith, line 17: I have a general preference to use the newer java qst / table APIs where possible - ala, io.deephaven.api.SortColumn here. @rcaudy ?

pyintegration/deephaven2/table.py
```diff
@@ -55,6 +222,801 @@ def update(self, formulas: List[str]):
...
+        """
+        try:
+            if by:
+                return Table(j_table=self._j_table.countBy(col, *by))
+            else:
+                return Table(j_table=self._j_table.countBy(col))
+        except Exception as e:
+            raise DHError(e, "table count_by operation failed.") from e
+
+    def combo_by(self, combo_agg: ComboAggregation, by: List[str]) -> Table:
```
devinrsmith, line 998: I think the switching parameter order is better.

pyintegration/deephaven2/__init__.py
```diff
@@ -1,12 +1,23 @@
...
 #  Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
 #
-"""Deephaven Python Integration Package provides the ability to access the Deephaven's query engine natively and thus
+"""Deephaven Python Integration Package provides the ability to access the Deephaven"s query engine natively and thus
 unlocks the unique and tremendous power Deephaven to Python community.
 
 """
+import jpy
+
+__version__ = "0.5.2"
```
devinrsmith, line 10: Needs to be updated.

pyintegration/deephaven2/combo_agg.py
```diff
@@ -0,0 +1,277 @@
+#
+#  Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
+#
+
+from typing import List, Any
+
+import jpy
+
+ComboAggregateFactory = jpy.get_type("io.deephaven.db.v2.by.ComboAggregateFactory")
```
devinrsmith, line 9: The usage of this looks like:

```python
from deephaven import ComboAggregation
x = table.combo_by(ComboAggregation().sum("Bar", "Baz").median("Baz"), ["Foo"])
```
if I'm looking correctly?

I think `io.deephaven.api.agg.Aggregation` may be the better interface to use, and we can wrap the Aggregation instead of the `ComboAggregation`. Then we could have: 

```python
def combo_by(self, by: List[str], aggs: List[Aggregation])
```

```python
from deephaven.Aggregation import sum, median

x = table.combo_by(["Foo"], [sum("Bar", "Baz"), median("Baz")])
```

pyintegration/deephaven2/table.py
```diff
@@ -55,6 +222,801 @@ def update(self, formulas: List[str]):
...
+        """
+        try:
+            if not filters:
+                return Table(j_table=self._j_table.where())
+            else:
+                return Table(j_table=self._j_table.where(*filters))
+        except Exception as e:
+            raise DHError(e, "table where operation failed.") from e
+
+    def where_in(self, filter_table, cols: List[str]) -> Table:
```
devinrsmith, line 421: We can provide typing for filter_table, yeah?

pyintegration/deephaven2/table.py
```diff
@@ -55,6 +222,801 @@ def update(self, formulas: List[str]):
...
+
+        Raises:
+            DHError
+        """
+        try:
+            return Table(j_table=self._j_table.whereIn(filter_table._j_table, *cols))
+        except Exception as e:
+            raise DHError(e, "table where_in operation failed.") from e
+
+    def where_not_in(self, filter_table, cols: List[str]) -> Table:
```
devinrsmith, line 440: Ditto.

pyintegration/deephaven2/table.py
```diff
@@ -55,6 +222,801 @@ def update(self, formulas: List[str]):
...
+        Raises:
+            DHError
+        """
+        try:
+            return Table(j_table=self._j_table.headBy(num_rows, *by))
+        except Exception as e:
+            raise DHError(e, "table head_by operation failed.") from e
+
+    def tail_by(self, num_rows: int, by: List[str]) -> Table:
+        """ The head_by method creates a new table containing the last number of rows for each group.
```
devinrsmith, line 737: You reference head_by

pyintegration/deephaven2/table.py
```diff
@@ -55,6 +222,801 @@ def update(self, formulas: List[str]):
...
+
+        Args:
+            num_rows (int): the number of rows at the beginning of each group
```
devinrsmith, line 740: at the end

pyintegration/deephaven2/table.py
```diff
@@ -42,11 +99,121 @@ def columns(self):
...
+        """
+        try:
+            return TableTools.string(self._j_table, num_rows, *cols)
+        except Exception as e:
+            raise DHError(e, "table to_string failed") from e
+
+    #
+    # Table operation category: Select
+    #
+    def drop_columns(self, cols: List[str]) -> Table:
```
devinrsmith, line 123: I think we should move the "_columns" methods to the end of the file. Personally I feel like they are somewhat superficial; I think we should generally order the methods like: selects, filters, joins, sorts, bys, other.

pyintegration/deephaven2/__init__.py
```diff
@@ -1,12 +1,23 @@
 #
 #  Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
 #
-"""Deephaven Python Integration Package provides the ability to access the Deephaven's query engine natively and thus
+"""Deephaven Python Integration Package provides the ability to access the Deephaven"s query engine natively and thus
```
devinrsmith, line 4: why the double quote?

devinrsmith (dismissed):
Quick pass. Didn't pay too much attention to the pydocs.

configs/dh-defaults.prop
jmao-denver: This is to avoid a LTM exclusive lock check error when running the tests involving updating a time table in the test mode.

pyintegration/deephaven2/combo_agg.py
jmao-denver: yes, that's the correct usage, and it is the same interface as the one in the client (pydeephaven).  I do like the simpler interface but I also like the explicitness and slightly better readability of ComboAggregation.  Need to hear @chipkent's thoughts.

configs/dh-defaults.prop
jmao-denver: This is to avoid a LTM exclusive lock check error when running the Python tests involving updating a time table in the test mode.

pyintegration/deephaven2/__init__.py
jmao-denver: done

pyintegration/deephaven2/combo_agg.py
jmao-denver: yes, that's the correct usage, and it is the same interface as the one in the client (pydeephaven). I do like the simpler interface but I also like the explicitness and slightly better readability of ComboAggregation. Need to hear @chipkent's thoughts.

pyintegration/deephaven2/table.py
jmao-denver: yes, done

pyintegration/deephaven2/table.py
jmao-denver: done

pyintegration/deephaven2/table.py
jmao-denver: I mostly followed the new doc https://deephaven.io/core/docs/reference/table-operations/select/drop-columns/ on the order of groups and methods. @chipkent ?

pyintegration/deephaven2/table.py
jmao-denver: I have no strong opinion. @chipkent ?

configs/dh-defaults.prop
jmao-denver: This flag causes Python test cases on time table update etc to error out due to the exclusive lock check failure. This file is only used to set up the test environment for running standalone Python tests in Integrations and for py code generations in the Generators project. I have consulted @niloc132 and we both believe that it was a mistake to set it to true.

pyintegration/deephaven2/combo_agg.py
chipkent: 1) I think the jpy type should be private if it is there.
2) I think the `Aggregation` syntax is cleaner than the `ComboAggregation` syntax.
3) I assume the module would be `deephaven.aggregation` and not `deephaven.Aggregation`.

chipkent (changes requested):

pyclient/pydeephaven/_table_interface.py
```diff
@@ -564,12 +564,12 @@ def count(self, col: str):
         table_op = DedicatedAggOp(AggType.COUNT, count_column=col)
         return self.table_op_handler(table_op)
 
-    def combo_by(self, by: List[str], agg: ComboAggregation):
+    def combo_by(self, agg: ComboAggregation, by: List[str]):
```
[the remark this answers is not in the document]
jmao-denver: @chipkent  this is actually the client side change to make it consistent with the server.

pyintegration/deephaven2/constants.py
```diff
@@ -0,0 +1,15 @@
+#
+#  Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
+#
+
+from enum import Enum, auto
+
+
+class SortDirection(Enum):
+    """An enum defining the sorting orders."""
+    DESCENDING = auto()
```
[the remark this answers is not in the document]
jmao-denver: No because there is no reason for the users to use the enums' values directly but only the enums themselves.

pyintegration/deephaven2/table.py
```diff
@@ -2,28 +2,85 @@
...
-        self._db_table = db_table
-        self._definition = self._db_table.getDefinition()
+    def __init__(self, j_table):
+        self._j_table = j_table
+        self._definition = self._j_table.getDefinition()
         self._schema = None
 
-    # to make the table visible to DH script session
+    # to make the table visible to DH script session, internal use only
     def get_dh_table(self):
```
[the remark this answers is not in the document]
jmao-denver: this is actually dictated by the server code, any other names would defeat the purpose unless we change the server code as well.

pyintegration/deephaven2/table.py
```diff
@@ -42,11 +99,121 @@ def columns(self):
 
         return self._schema
 
-    def update(self, formulas: List[str]):
+    def to_string(self, num_rows: int = 10, cols: List[str] = []) -> str:
```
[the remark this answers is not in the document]
jmao-denver: __repr__()

configs/dh-defaults.prop
jmao-denver: @rcaudy do you think the change is warranted?

jmao-denver: test

jmao-denver: @rcaudy do you think the change is warranted?

pyintegration/deephaven2/table.py
jmao-denver: Tried to use SortColumn instead but ran into a potential JPY overload matching issue, pretty sure I don't want to touch JPY just yet.

pyintegration/deephaven2/agg.py
```diff
@@ -0,0 +1,256 @@
...
+    """
+
+    def __init__(self, j_aggregation):
+        self._j_aggregation = j_aggregation
+
+    @property
+    def j_agg(self):
+        return self._j_aggregation
+
+    @staticmethod
```
chipkent, line 25: I expected these to be module-level functions.  Having them as static functions seems to add complexity without obvious benefit in this design.  With these as static methods, users can not import them individually, which may lead to nastier code in some cases.  What am I missing?
https://stackoverflow.com/questions/12603541/import-only-a-class-static-method

chipkent (changes requested):

pyintegration/deephaven2/table.py
```diff
@@ -55,6 +244,806 @@ def update(self, formulas: List[str]):
...
+        """
+        try:
+            if by:
+                return Table(j_table=self._j_table.countBy(col, *by))
+            else:
+                return Table(j_table=self._j_table.countBy(col))
+        except Exception as e:
+            raise DHError(e, "table count_by operation failed.") from e
+
+    def combo_by(self, aggs: List[Aggregation], by: List[str]) -> Table:
```
chipkent, line 1031: I think ryan is naming this `agg_by`.

chipkent (changes requested):

pyintegration/deephaven2/agg.py
```diff
@@ -0,0 +1,256 @@
...
+    Args:
+        cols (List[str]): the columns to aggregate on, can be renaming expressions, i.e. "new_col = col"
+
+    Returns:
+        an aggregation
+    """
+    return Aggregation(j_aggregation=_JAggregation.AggSum(*cols))
+
+
+def abs_sum(cols: List[str]):
```
chipkent, line 38: Most of these aggregations do not have the return type

chipkent (changes requested):

chipkent (approved):

Issue #1026, opened 2021-08-11
Learn: Implement user-specified column types


alexpeters1208:
**Description**

The Learn library cannot create new table columns with user-specified output types until ticket #1009 is resolved.

**Steps to reproduce**

Execute the following code in the Deephaven IDE:

```
from deephaven.TableTools import newTable, intCol
from deephaven import learn
import numpy as np

table = newTable(intCol("Column1", 1, 2, 3, 4, 5))

def gather(idx, col):
    rst = np.empty(idx.getSize(), dtype=np.int)
    iter = idx.iterator()
    i = 0
    while(iter.hasNext()):
        it = iter.next()
        rst[i] = col[0].get(it)
        i=i+1
    return rst

def model_func(args):
    return args

def scatter(data, i):
    return data[i]

outputTable = learn.eval(
    table = table,
    model_func = model_func,
    inputs = [learn.Input("Column1", gather)],
    outputs = [learn.Output("NewColumn", scatter, "int")]
)
```

**Expected results**

NewColumn should be of type `int`.

**Actual results**

NewColumn is of type `String`, as it cannot be cast to an `int` per #1009.

**Additional information**

The correct code is already written as a TODO in io.deephaven.integrations/learn/Scatterer.java. Simply uncomment and get rid of current for loop body (~line 56) once #1009 is resolved.

**Versions**
 - Deephaven: 0.2.1
 - OS: macOS Big Sur 11.4
 - Browser: Safari 14.1.1
 - Docker: 3.3.3

PR #1543, opened 2021-11-09
DHLearn: Scatterer: For Python, cast scatter results to PyObject before casting to provided type

chipkent:
JJ's DHLearn example doesn't work because this `scatter` method returns `Object` instead of `PyObject`.  The query language has rules for casting.  The `Object -> (double)` cast uses a `(Double)` cast, while `PyObject -> (double)` uses a python-specific `doubleValue` cast.  Because the query language does not know that the `scatter` result is a `PyObject`, it uses the `Object -> (Double)` casting rule and fails.

This fix forcefully casts results from Python scatter functions to `PyObject` before casting to the desired type.

The important bits of JJ's example is this:
```python
# A function to extract a result
def get_results(data, idx):
    return data[idx]

result = learn.learn(
    table = source,
    model_func = sum_columns,
    inputs = [learn.Input(["X", "Y"], table_to_array)],
    outputs = [learn.Output("Z", get_results, "double")],
    batch_size = 1000
```

Resolves: #1026 
Closes: #1539


Integrations/src/main/java/io/deephaven/integrations/learn/Output.java
```diff
@@ -36,13 +37,25 @@ public Output(String colName, PyObject scatterFunc, String type) {
...
+    }
+
+    /**
+     * Creates a new Output.
+     *
+     * @param colName name of new column to store results.
+     * @param scatterFunc function to scatter the results of a Python object into the table column.
+     * @param type desired datatype of the new column.
+     */
+    public Output(final String colName, final boolean isPythonScatterFunc, final Function<Object[], Object> scatterFunc, final String type) {
```
nbauernfeind, line 50: javadoc the new boolean, please

nbauernfeind:
LGTM. Looks like spotless is failing. Can be fixed with:
```
./gradlew spotlessApply
```

And then committing the generated changes.

jjbrosnan:
I have tested this with combinations of types in `deephaven.learn.learn` function calls for both inputs and outputs:

byte, short, int, long, float, double

I have done operations on (I think?) every combination of what's listed above, and returned all of the types above, and the code has worked as expected.  I tested with `scikit-learn`, `pytorch`, and `tensorflow`, and the code still worked.

I'm wrapping up my testing.  If there's anything I missed, please let me know and I'll check it before this gets merged.

nbauernfeind (dismissed):

jjbrosnan (approved):

jjbrosnan (approved):

nbauernfeind (approved):

PR #1547, opened 2021-11-09
Move wheel building to own projects

devinrsmith:


niloc132 (approved):
Approved. I don't care for a lot of pieces of it, but we're moving in the right direction, as discussed in slack.

Specific squeaky wheels (ha.) that call my name:
 * Integrations:buildDeephavenPython should feed in the actual files to the image, not just depend on the images and assume that the dockerfile happens to also name the images by tag. We have no other use for the output of a "registerDockerImage" task here, we just use it as a vehicle to store a wheel.
 * Going further, we should probably have configurations/dependencies for wheels, consider linkedin/pygradle at least as inspiration, if not actually the right way to do it. Some way to set up "hey here's an output" without stringly-typed projectname:taskname tuples scattered across projects, using NPEs to find missing links.

None of this is made worse by this patch, and I'm explicitly signing off that we should merge now, but it isn't getting better either :p.

PR #1556, opened 2021-11-11
Extract deephaven/java-and-python

devinrsmith:


niloc132 (approved):
No specific objections, just would prefer that java-and-python etc live with our other base images, so they are already tagged and ready to go.

PR #1557, opened 2021-11-11
Extract deephaven/runtime-base

devinrsmith:


docker/java-and-python/src/main/docker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+      python3.7 \
+      libpython3.7 \
+      python3-pip \
+      python3-setuptools \
+      python3-wheel \
+      ; \
+    rm -rf /var/lib/apt/lists/*; \
+    python3 -m pip install -q --no-cache-dir --upgrade pip; \
+    python3 -m pip install -q --no-cache-dir --upgrade setuptools; \
+    python3 -m pip install -q --no-cache-dir --upgrade wheel
```
niloc132, line 39: where possible, version these pip/apt-get packages in this file

docker/runtime-base/build.gradle
```diff
@@ -0,0 +1,49 @@
+plugins {
+    id 'com.bmuschko.docker-remote-api'
+}
+
+evaluationDependsOn ':docker-java-and-python'
+evaluationDependsOn ':deephaven-jpy'
+evaluationDependsOn ':deephaven-wheel'
+evaluationDependsOn ':deephaven2-wheel'
```
niloc132, line 8: eesh i wish we had dependencies instead of this stringy goo you hope you never forget...

docker/java-and-python/src/main/docker/Dockerfile
devinrsmith: I've got a new structure for requirements.txt, let me know what you think.

docker/runtime-base/build.gradle
devinrsmith: I *think* the good news is that the gradle build will just fail to work if an evaluationDependsOn is missing. That said, it's not necessarily consistently failing - but hopefully it's obvious enough to us doing "things" in gradle how to fix it.

niloc132 (approved):

PR #1562, opened 2021-11-12
docker registry

devinrsmith:


docker/java-and-python/build.gradle
```diff
@@ -0,0 +1,13 @@
+plugins {
+    id 'com.bmuschko.docker-remote-api'
+}
+
+evaluationDependsOn Docker.registryProject('debian')
+
+def buildDocker = Docker.registerDockerImage(project, 'buildDocker') {
+    inputDir.set layout.projectDirectory.dir('src/main/docker')
+    inputs.files Docker.registryFiles(project, 'debian')
```
niloc132, line 9: if we're going so far as to providing all our own tagged uptsreams, why not take the next step and provide a model for these, let registerDockerImage have a reference to the string in the registry? then it can invoke evaluationDependsOn and find the files itself?

docker/registry/alpine/gradle.properties
```diff
@@ -0,0 +1,4 @@
+# Deliberately using old alpine, see issue #1034 in deephaven-core
+# See also alpinelinux/docker-alpine#155
+deephaven.registry.imageName=alpine:3.12
+deephaven.registry.imageId=alpine@sha256:a296b4c6f6ee2b88f095b61e95c7ef4f51ba25598835b4978c9256d8c8ace48a
```
niloc132, line 4: Will need to branch this to support more than one sha per tag for os/arch pairs (see https://hub.docker.com/r/envoyproxy/envoy/tags for example, but notice too that if you click on v1.20.0 or whatever, even though it lists two digests, it only takes you to the linux/amd64 variant...)

devinrsmith: That's not correct - a repo digest (assuming it supports multi arch) will work across any architectures it supports. For example:

```
$ docker buildx imagetools inspect envoyproxy/envoy@sha256:1e4de079793a62a1ed7023bc44228b80d753773dbf24dcf7b318f360ca595b31 
Name:      docker.io/envoyproxy/envoy@sha256:1e4de079793a62a1ed7023bc44228b80d753773dbf24dcf7b318f360ca595b31
MediaType: application/vnd.docker.distribution.manifest.list.v2+json
Digest:    sha256:1e4de079793a62a1ed7023bc44228b80d753773dbf24dcf7b318f360ca595b31
           
Manifests: 
  Name:      docker.io/envoyproxy/envoy@sha256:11cd0b7aefd2d6b5abc2ad9c2937b35dc9b2f10c67e19eca9af780cdc7e9892b
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/arm64
             
  Name:      docker.io/envoyproxy/envoy@sha256:e02931a6470916fd1f468d807cded3840062f2effa586ab90b901d7f494cc91f
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/amd64
```

```
$docker buildx imagetools inspect alpine@sha256:a296b4c6f6ee2b88f095b61e95c7ef4f51ba25598835b4978c9256d8c8ace48a           
Name:      docker.io/library/alpine@sha256:a296b4c6f6ee2b88f095b61e95c7ef4f51ba25598835b4978c9256d8c8ace48a
MediaType: application/vnd.docker.distribution.manifest.list.v2+json
Digest:    sha256:a296b4c6f6ee2b88f095b61e95c7ef4f51ba25598835b4978c9256d8c8ace48a
           
Manifests: 
  Name:      docker.io/library/alpine@sha256:a721d672be3cdaaf1285c2b60b76d81ab4a8a3cf5a7180a966f2c545305ad6de
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/amd64
             
  Name:      docker.io/library/alpine@sha256:b7b182ccc0811af3fc00754c3cc744e45bc3d7d4f6ef9ee47071173f893dbfb0
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/arm/v6
             
  Name:      docker.io/library/alpine@sha256:16aea3fc0648559c47f974d7f31096e71b9ec7013224bb29e35cf3db375af976
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/arm/v7
             
  Name:      docker.io/library/alpine@sha256:df5a62c11342b503e692ac719f9569da3518c8f1d5a7f072ad76f1b5ef8d3d43
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/arm64/v8
             
  Name:      docker.io/library/alpine@sha256:38f483c3c603557ed5f9a287d13dd9ca8daea07435ad3f06b038621a2f9fc1d9
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/386
             
  Name:      docker.io/library/alpine@sha256:372de563e36c3989cee334c3cedf72d0e80879ed989ebd8dee42bdbff5fe4184
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/ppc64le
             
  Name:      docker.io/library/alpine@sha256:5391bba453c1474d620fd092c96342e09e0a8ee7f1c2e95ae2dc6e3ed44408c9
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/s390x
```

docker/java-and-python/build.gradle
devinrsmith: This seems like an ok idea - it's just a deeper level of integration that I didn't want to go down at this time. Giving the PR a "lite" touch.

niloc132: To verify, do we need to pin any of the entries currently present? It looks like after you run `pullImage` (without cleaning docker...) you will be reminded each time about alpine if you run `bumpImage`, with no way to suppress it?

I know we're accidentally out of date on this branch and so expecting the command to fail for most of them, but the new github actions check added in this commit will fail on the deliberately-out-of-date alpine, right?

devinrsmith: I *think* you mean `compareImage` and not `bumpImage`? You'll only get the failure when you are specifically running docker registry "release maintenance" tasks; not as part of a prepareCompose.

I think it makes sense to add a "I know this is out of date, don't fail" flag. It will allow us to make sure that our failure notification to DHC notify is new information.

niloc132: Ah yes, compareImage, sorry. Will re-run with the latest and make sure I follow.

niloc132: Going through my bash history, I actually ran compareImage, so that's good at least... 

To confirm, you added the new ignoreOutOfDate on non-alpine images, rather than update them - is that what you meant to do? Without the last commit, wouldn't slack have started messaging nightly to fix these?

Also, maybe change the note when ignoreOutOfDate is set, to the effect of "this is out of date, ignoring, remove ignoreOutOfDate or run bumpImage and test" so it is clear which images are "really out of date" and which ones are "expected to be out of date"?

Or do we want to start from the default position of ignoreOutOfDate=true?

devinrsmith: I'm happy to bump all the digests to their latest repo digest version. The reason I didn't (yet - was going to be an immediate follow up to this PR) was to keep the build images as close as possible to the "existing" build images (at the time the PR first started).

The "out-of-date" refers to human-friendly name being out-of-date wrt the repo digest; it doesn't refer to the human-friendly name being out-of-date.  We should even bump alpine:3.12; there are likely security concerns sticking to old 3.12s.

Without my last commit, we would be getting notifications in the DHC notification channel nightly. With my latest commit, the relevant CI tasks would not fail, and would not cause notifications (but it would still print out warnings).

The default should be ignoreOutOfDate=false; if we are stuck on an explicit *repo-digest* (because the newer digest breaks our build, or because we are "baking" QA for a release), that is the time to set ignoreOutOfDate=false and likely attache a ticket number.

I'll push out what was going to be a follow-up PR to bump all the versions.

buildSrc/src/main/groovy/Docker.groovy
```diff
@@ -473,10 +478,121 @@ class Docker {
...
+                                    "deephaven.registry.imageId=${repoDigest}\n"
+                    inspect.logger.quiet("Updated imageId for '${imageName}' to '${repoDigest}' from '${imageId}'.")
+                } else {
+                    inspect.logger.quiet("ImageId for '${imageName}' already up-to-date.")
+                }
+            }
+        }
+
+        project.tasks.register('compareImage', DockerInspectImage) {inspect ->
+            inspect.group = 'Docker Registry'
```
niloc132, line 532: probably should make this a constant

niloc132 (approved):

jcferretti (approved):

PR #1565, opened 2021-11-12
Update JS API to support gRPC input tables

niloc132:
Building on https://github.com/deephaven/deephaven-core/pull/1440, this adds support in the JS API for input tables.

CSV upload is rebuilt in this as well, and expanded to be more generic, so that input tables can reuse the same logic. This should be at least as flexible as the UI previously required (though the ability to handle arrays of data has been removed temporarily, but to my knowledge it was never in active use).


web/client-api/src/main/java/io/deephaven/web/client/api/barrage/def/TableAttributesDefinition.java
```diff
@@ -0,0 +1,81 @@
...
+    public TableAttributesDefinition(Map<String, String> keys, Set<String> remainingAttributes) {
+        map = keys;
+        this.remainingAttributeKeys = remainingAttributes;
+        if (map.containsKey(HIERARCHICAL_SOURCE_INFO_ATTRIBUTE + "." + HIERARCHICAL_COLUMN_NAME)) {
+            // marker present for tree table metadata
+            rollupDefinition = new RollupDefinition();
+            rollupDefinition.setByColumns(map.get(HIERARCHICAL_COLUMN_NAME + "." + "byColumns").split(","));
+            rollupDefinition.setLeafType(
+                    RollupDefinition.LeafType.valueOf(map.get(HIERARCHICAL_COLUMN_NAME + "." + "leafType")));
+        }
```
mofojed, line 35: I'm a little confused about the constants being used here... `HIERARCHICAL_COLUMN_NAME` is `drinkingCamelCase`, and here on the first line you get `HIERARCHICAL_SOURCE_INFO_ATTRIBUTE + "." + HIERARCHICAL_COLUMN_NAME`, or getting `"HierarchicalSourceTableInfo.hierarchicalColumnName"`, which makes sense... but then below, you're getting `"hierarchicalColumnName.byColumns"` - does that not need the `HIERACRCHICAL_SOURCE_INFO_ATTRIBUTE` prefix? If not why not?

I think you should just define these all as constants so it's not as confusing:
```
SOURCE_INFO_COLUMN_NAME = HIERARCHICAL_SOURCE_INFO_ATTRIBUTE + "." + HIERARCHICAL_COLUMN_NAME;
BY_COLUMN_ATTRIBUTE = HIERARCHICAL_COLUMN_NAME + ".byColumns";
LEAF_TYPE_ATTRIBUTE = HIERARCHICAL_COLUMN_NAME + ".leafType"

```

web/client-api/src/main/java/io/deephaven/web/client/api/input/JsInputTable.java
```diff
@@ -60,93 +62,136 @@ public JsInputTable(JsTable from, String[] keys, String[] values) {
...
+        // for each table, make a view on that table of only key columns, then union the tables and drop together
+        final List<JsRunnable> cleanups = new ArrayList<>();
+        final Ticket ticketToDelete;
+        if (tablesToDelete.length == 1) {
+            JsTable onlyTable = tablesToDelete[0];
+            // don't try too hard to find matching columns, if it looks like we have a match go for it
+            if (onlyTable.getColumns().length == keys.length && onlyTable.findColumns(keys).length == keys.length) {
+                ticketToDelete = onlyTable.getHandle().makeTicket();
+            } else {
+                // view the only table
```
mofojed, line 149: Shouldn't it throw if the tablesToDelete doesn't match the schema of the input table you're deleting from?

mofojed (changes requested):
A bit of minor cleanup, and a couple of questions

web/client-api/src/main/java/io/deephaven/web/client/api/input/JsInputTable.java
niloc132: It will throw, but not until later. I can add a local check here to give a clearer message.

web/client-api/src/main/java/io/deephaven/web/client/api/barrage/def/TableAttributesDefinition.java
niloc132: Yeah the real headache is that they get defined twice, once on the server and once on the client. As discussed in the custom object call, let's redo this again once that is merged? I will give this a quick look though to make sure you're right, but trees don't work anyway...

web/client-api/src/main/java/io/deephaven/web/client/api/barrage/def/TableAttributesDefinition.java
```diff
@@ -29,9 +32,8 @@ public TableAttributesDefinition(Map<String, String> keys, Set<String> remaining
         if (map.containsKey(HIERARCHICAL_SOURCE_INFO_ATTRIBUTE + "." + HIERARCHICAL_COLUMN_NAME)) {
```
mofojed, line 32: No need to add the prefix again, it's already added:
```suggestion
        if (map.containsKey(HIERARCHICAL_COLUMN_NAME)) {
```

web/client-api/src/main/java/io/deephaven/web/client/api/input/JsInputTable.java
```diff
@@ -182,7 +185,17 @@ public JsInputTable(JsTable from, String[] keys, String[] values) {
             mergeRequest.setResultId(ticketToDelete);
             batch.addOps(new Operation()).setMerge(mergeRequest);
 
-            table.getConnection().tableServiceClient().batch(batch, table.getConnection().metadata());
+            failureToReport = new Promise<>((resolve, reject) -> {
+                ResponseStreamWrapper<ExportedTableCreationResponse> wrapper = ResponseStreamWrapper.of(
+                        table.getConnection().tableServiceClient().batch(batch, table.getConnection().metadata()));
+                wrapper.onData(response -> {
+                    // kill the promise on the first failure we see
```
mofojed, line 192: Is it possible to get more than one failure? Then we might be double rejecting the promise

mofojed (changes requested):

web/client-api/src/main/java/io/deephaven/web/client/api/input/JsInputTable.java
niloc132: Double rejecting just means you lose a later reject (or resolve), no? This code is operating on that assumption for now, rather than accumulate errors or something - is that okay?

mofojed: Yea I guess that's fine

mofojed (approved):

web/client-api/src/main/java/io/deephaven/web/client/api/parse/JsDataHandler.java
```diff
@@ -0,0 +1,488 @@
...
+                    }
+                } else {
+                    // fall back to whatever we can get - this is null safe, since we already tested for null above
+                    value = (long) (double) JsDataHandler.doubleFromData(data[i]);
+                }
+                if (value == NULL_LONG) {
+                    nullCount++;
+                } else {
+                    nulls.set(i);
+                }
```
nbauernfeind, line 195: It's ok actually to count all of these as not null (for the count). If the null count is zero, then the validity buffer can be skipped (sent as empty buffer). Server side, whether the client enables sending deephaven nulls or not, these longs will simply be the null value. So to save bandwidth, I recommend we do this.

web/client-api/src/main/java/io/deephaven/web/client/api/parse/JsDataHandler.java
```diff
@@ -0,0 +1,488 @@
...
+
+                if (boolValue != FALSE_BOOLEAN_AS_BYTE && boolValue != TRUE_BOOLEAN_AS_BYTE
+                        && boolValue != NULL_BOOLEAN_AS_BYTE) {
+                    throw new IllegalArgumentException("Can't handle " + val + " as a boolean value");
+                }
+
+                // write the value, and mark non-null if necessary
+                if (boolValue != NULL_BOOLEAN_AS_BYTE) {
+                    nulls.set(i);
```
nbauernfeind, line 311: this is inconsistent w.r.t. a corresponding nullCount -- however, I think this use case also applies; that we explicitly want to send the empty validity buffer.

web/client-api/src/main/java/io/deephaven/web/client/api/parse/JsDataHandler.java
```diff
@@ -0,0 +1,488 @@
...
+     * @param addNode a callback to append a node to the arrow payload
+     * @param addBuffer a callback to append a buffer to the arrow payload
+     * @param bytesPerElement the width of the arrow/deephaven type being created
+     * @param nullValue the value to write in case of null data
+     * @param bufferConstructor a constructor to produce a typedarray for the data being created
+     */
+    private static void writeSimpleNumbers(Object[] data, JsConsumer<Node> addNode, JsConsumer<Uint8Array> addBuffer,
+            double bytesPerElement, double nullValue, JsFunction<ArrayBuffer, ? extends TypedArray> bufferConstructor) {
+        int nullCount = 0;
+        BitSet nulls = new BitSet(data.length);
```
nbauernfeind, line 372: same comment here on the null counting; I'd send the empty validity buffer explicitly

web/client-api/src/main/java/io/deephaven/web/client/api/input/JsInputTable.java
```diff
@@ -60,99 +64,155 @@ public JsInputTable(JsTable from, String[] keys, String[] values) {
...
+                .then(merged -> {
+                    // noinspection CodeBlock2Expr - easier readability for chained then()
+                    return Callbacks.grpcUnaryPromise(c -> {
+                        AddTableRequest addTableRequest = new AddTableRequest();
+                        addTableRequest.setInputTable(table.getHeadHandle().makeTicket());
+                        addTableRequest.setTableToAdd(merged.getHeadHandle().makeTicket());
+                        table.getConnection().inputTableServiceClient().addTableToInputTable(addTableRequest,
+                                table.getConnection().metadata(), c::apply);
+                    }).then(success -> {
+                        if (merged != tablesToAdd[0]) {
```
nbauernfeind, line 117: I understand the check; but correctness might be more obvious if you stick to the length check like on line 101.

web/client-api/src/main/java/io/deephaven/web/client/api/parse/JsDataHandler.java
```diff
@@ -0,0 +1,488 @@
...
+            Int32Array positions = makeBuffer(data.length + 1, 4, Int32Array::new);
+            // work out the total length we'll need for the payload, plus padding
+            int payloadLength =
+                    Arrays.stream(data).filter(Objects::nonNull).map(Object::toString).mapToInt(String::length).sum();
+            Uint8Array payload = makeBuffer(payloadLength);
+
+            int lastOffset = 0;
+            for (int i = 0; i < data.length; i++) {
+                if (data[i] == null) {
+                    continue;
```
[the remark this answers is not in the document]
niloc132: Yep, that got lost in the refactor

web/client-api/src/main/java/io/deephaven/web/client/api/parse/JsDataHandler.java
niloc132: At this time, I am not sending a dhNulls=true, but doing a vanilla DoPut, just with deephaven:type metadata.

web/client-api/src/main/java/io/deephaven/web/client/api/input/JsInputTable.java
niloc132: The goal in having a different check was to acknowledge "this is an intermediate instance so it must be closed" - if the logic between here and 101 changes, I would want to allow for this still to be correct.

What do you think of something instead like `Arrays.stream(tablesToAdd).allMatch(t -> t != merged)`?

nbauernfeind (dismissed):

nbauernfeind (approved):

PR #1581, opened 2021-11-16
Consolidate test setup/teardown

### LINKED ISSUE #1549: QueryScope: Unit tests that install a new one should clean up in a finally block
A failure to perform proper cleanup was the likely cause of the failure in:
https://gist.github.com/niloc132/4432d7a005f9300b1f5d82474f65e190

In that example, `io.deephaven.db.v2.TableMapTest.testCrossDependencies()` failed because a thread in the update graph processor couldn't compile a formula that referenced a query scope variable. The likely explanation is that the UGP thread was using the default query scope, while the test thread was using a different query scope installed by a sloppy test.

We should audit and address all cases of tests failing to uninstall a custom scope they install. Likely we should make the common base test classes clean up the current and default query scope in `tearDown` as a matter of course.

niloc132:
This patch consolidates various setup/teardown code into LiveTableTestCase for JUnit 3 tests, and DbCleanup for JUnit4 tests, and adds the DbCleanup test rule to other tests that were currently missing it. Those tests that were previously missing it often had incomplete cleanup, which has mostly been removed through this change.

Fixes #1549


DB/src/main/java/io/deephaven/db/v2/utils/KeyedArrayBackedMutableTable.java
```diff
@@ -204,8 +203,10 @@ protected void processPendingDelete(Table table, IndexChangeRecorder indexChange
         try (final WritableLongChunk<Attributes.KeyIndices> destinations =
                 WritableLongChunk.makeWritableChunk(chunkCapacity)) {
             try (final ChunkSource.GetContext getContext = keySource.makeGetContext(chunkCapacity, sharedContext);
-                    final ChunkBoxer.BoxerKernel boxer = ChunkBoxer.getBoxer(keySource.getChunkType(), chunkCapacity)) {
-                final Chunk<? extends Attributes.Values> keys = keySource.getChunk(getContext, table.getIndex());
+                    final ChunkBoxer.BoxerKernel boxer = ChunkBoxer.getBoxer(keySource.getChunkType(), chunkCapacity);
+                    final Index tableIndex = table.getIndex().clone();) {
```
rcaudy, line 207: Same question: why do we need to clone an index we aren't mutating or keeping long-term ownership of?

DB/src/test/java/io/deephaven/db/v2/LiveTableTestCase.java
```diff
@@ -30,33 +32,51 @@
...
     static public boolean printTableUpdates = Configuration.getInstance()
             .getBooleanForClassWithDefault(LiveTableTestCase.class, "printTableUpdates", false);
+    private static final boolean ENABLE_COMPILER_TOOLS_LOGGING = Configuration.getInstance()
+            .getBooleanForClassWithDefault(QueryTableTestBase.class, "CompilerTools.logEnabled", false);
 
     private boolean oldMemoize;
     private UpdateErrorReporter oldReporter;
     private boolean expectError = false;
     private SafeCloseable scopeCloseable;
+    private QueryScope originalScope;
```
rcaudy, line 42: Maybe call this `originalQueryScope` and rename `scopeCloseable` to `livenessScopeCloseable` for clarity?

DB/src/test/java/io/deephaven/db/v2/select/TestFormulaColumn.java
```diff
@@ -65,6 +67,9 @@ public TestFormulaColumn(boolean useKernelFormulas) {
         availableColumns = testDataTable.getDefinition().getColumnNameMap();
     }
 
+    @Rule
+    public final DbCleanup base = new DbCleanup();
+
     @Before
     public void setUp() throws Exception {
```
rcaudy, line 74: Did you mean to remove more of the setup and teardown, here?

DB/src/test/java/io/deephaven/util/calendar/StaticCalendarMethodsTest.java
```diff
@@ -24,6 +26,20 @@
     private final String date1 = "2017-08-01";
     private final String date2 = "2017-08-05";
 
+    private final DbCleanup base = new DbCleanup();
+
+    @Override
+    protected void setUp() throws Exception {
+        super.setUp();
+        base.setUp();
```
rcaudy, line 34: Why do you need to do this in this class, but not in other ones?

extensions/barrage/src/test/java/io/deephaven/extensions/barrage/chunk/BarrageColumnRoundTripTest.java
```diff
@@ -450,13 +450,19 @@ public void assertExpected(final WritableChunk<Attributes.Values> untypedOrigina
...
             }
 
             // empty subset
             try (final BarrageProtoUtil.ExposedByteArrayOutputStream baos =
-                    new BarrageProtoUtil.ExposedByteArrayOutputStream()) {
-                final ChunkInputStreamGenerator.DrainableColumn column =
-                        generator.getInputStream(options, Index.CURRENT_FACTORY.getEmptyIndex());
+                    new BarrageProtoUtil.ExposedByteArrayOutputStream();
+                    final ChunkInputStreamGenerator.DrainableColumn column =
+                            generator.getInputStream(options, Index.CURRENT_FACTORY.getEmptyIndex());) {
```
rcaudy, line 464: @nbauernfeind Is the Index going to be mutated, too? Should we be using it as a resource?

extensions/barrage/src/test/java/io/deephaven/extensions/barrage/chunk/BarrageColumnRoundTripTest.java
```diff
@@ -425,18 +425,18 @@ public void assertExpected(final WritableChunk<Attributes.Values> untypedOrigina
         final ChunkType chunkType = ChunkType.fromElementType(type);
 
         WritableChunk<Attributes.Values> rtData = null;
-        final WritableChunk<Attributes.Values> data = chunkType.makeWritableChunk(4096);
+        WritableChunk<Attributes.Values> data = chunkType.makeWritableChunk(4096);
```
rcaudy, line 428: Isn't data going to leak, too? Why did it become non-final?

DB/src/main/java/io/deephaven/db/v2/utils/KeyedArrayBackedMutableTable.java
nbauernfeind: The problem is that this index is the table's actual index. There is a call that creates an ordered keys chunk of ranges which is then cached in the Index. But that Chunk is never free'd in the test because it isn't closed until the index itself is closed. So it looks like a chunk-leak to the test.

It seemed like the only way around this was to clone and close -- enabling that orderedkeys cached chunk to be cleaned up.

extensions/barrage/src/test/java/io/deephaven/extensions/barrage/chunk/BarrageColumnRoundTripTest.java
nbauernfeind: That index is not mutated. It does get cloned by the DrainableColumn and that subset is closed when the column itself is closed.

We might need to tryWithResources it if empty indexes are no longer allowed to leak to GC.

rcaudy: We're leaving it for the GC regardless of whether we close it. This is fine; I was only concerned about making sure we close it if we mutate it and then clone it in a more persistent structure.

extensions/barrage/src/test/java/io/deephaven/extensions/barrage/chunk/BarrageColumnRoundTripTest.java
nbauernfeind: That data chunk is passed off to the ChunkInputStreamGenerator as a matter of ownership. One line 432 that StreamGenerator is now correctly being tried with resources, which enables this chunk to be closed correctly. The StreamGenerator needs it for its entire life. 

Final vs non-final though was probably just an iterative change while Colin was looking for the source of the issue with this test's failure.

rcaudy: OK. Thanks for the clarification on ownership.

DB/src/test/java/io/deephaven/db/v2/select/TestFormulaColumn.java
niloc132: Yes - I didn't hit every other test class that could be slimmed down yet, trying to keep the churn down, but this one I should have removed the CompilerTools line.

DB/src/test/java/io/deephaven/util/calendar/StaticCalendarMethodsTest.java
niloc132: This is a junit3 test, I only intended to upgrade the ones that seemed already to be meant to be junit4

extensions/barrage/src/test/java/io/deephaven/extensions/barrage/chunk/BarrageColumnRoundTripTest.java
niloc132: fixed final, and i made rtData more "local" so that it is easier to clean up

niloc132:
One more fix that needs to be added that got lost: the various ChunkInputStreamGenerator impls need to stop closing the stream they are given, i'll add that patch and test it again.

nbauernfeind (approved):

rcaudy (approved):

PR #1585, opened 2021-11-17
Add Replayer to TableManipulation

jjbrosnan:
Also alphabetizes the contents of the TableManipulation submodule


Integrations/python/deephaven/TableManipulation/__init__.py
```diff
@@ -9,18 +9,18 @@
 import jpy
```
chipkent, line 9: I did a quick search of https://github.com/jjbrosnan/deephaven-core/blob/add-replayer-to-python/Integrations/python/deephaven/__init__.py for `TableManipulation`.  I see a number of places in that file that look like they might need changing as well.

chipkent (changes requested):

chipkent (approved):

jmao-denver (approved):

PR #1593, opened 2021-11-19
Update redpanda version and enable embedded schema registry.

jcferretti:


Integrations/docker-compose.yml
```diff
@@ -9,13 +9,4 @@ services:
...
-  registry:
-    image: apicurio/apicurio-registry-mem:1.2.2.Final
-    expose:
-      - 8080
-    environment:
-      QUARKUS_PROFILE: prod
-      KAFKA_BOOTSTRAP_SERVERS: redpanda:9092
-      APPLICATION_ID: registry_id
-      APPLICATION_SERVER: localhost:9000
+      - 8081
```
niloc132, line 12: personal opinion, stop using docker-compose for the test if we can, the plugin is not great, and i think we'll be happier to just do plain docker commands.

this isnt a requirement for merge or even a follow up, just a preference

devinrsmith (approved):

Integrations/docker-compose.yml
jcferretti: I hear you and makes sense but that is a non-trivial amount of work, so not doing it on this pass.

PR #1598, opened 2021-11-19
Use debian:buster-slim base image with node/python

devinrsmith:


devinrsmith:
Fixed in https://github.com/deephaven/deephaven-base-images/pull/5

PR #1602, opened 2021-11-23
Add docker-server and docker-server-pytorch

devinrsmith:
I ran a successful build in my repo - you should be able to pull:

```
docker pull ghcr.io/devinrsmith/server-pytorch:build-docker-server-pytorch
```

Note: these are large images.


grpc-api/server/native/src/main/resources/grpc-api-native.prop
```diff
@@ -1 +1,5 @@
 includefiles=grpc-api.prop
+
+jpy.jpyLib=/usr/local/lib/python3.7/dist-packages/jpy.cpython-37m-x86_64-linux-gnu.so
+jpy.jdlLib=/usr/local/lib/python3.7/dist-packages/jdl.cpython-37m-x86_64-linux-gnu.so
+jpy.pythonLib=/usr/lib/x86_64-linux-gnu/libpython3.7m.so.1.0
```
niloc132, line 5: If this file now exclusively only works inside docker, it should be renamed and moved. By design, the -native and -docker projects have different configs (though these are the only substantial changes, the name of this file, the contents, and the reference to it from the bootstrap properties), and this change should only exist on the docker side.

Perhaps remove this file for native and add it only when docker/server copies sources from this project, likewise with its own bootstrap (or command line arg sysprop to pass to the docker startup)? Or set these sysprops from docker command line?

grpc-api/server/native/build.gradle
```diff
@@ -22,14 +23,45 @@ dependencies {
   // runtimeOnly project(':log-to-logbuffer')
 }
 
-application {
+distributions {
```
niloc132, line 26: With the removal of the grpc-api-server-docker, it probably makes sense to move this up to grpc-api-server itself?

I'm content to have this change happen later, probably in the jetty branch where I'm renaming grpc-api to server anyway, and adding jetty vs netty variants, but just want to be sure this makes sense as a plan.

docker/server-pytorch/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+}
+
+def buildDocker = Docker.registerDockerImage(project, 'buildDocker') {
+    dependsOn prepareDocker
+    inputDir.set dockerContext
+    inputs.files project.project(':docker-server').tasks.findByName('buildDocker').outputs.files
+    images.add('deephaven/server-pytorch:local-build')
+}
+
+assemble.dependsOn buildDocker
```
niloc132, line 22: newline nit

docker/server-pytorch/src/main/docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+COPY extra-requirements.txt extra-requirements.txt
+
+# Note: by running both requirements in a single invocation, we'll make sure that there aren't any conflicting requirements
+RUN set -eux; \
+    python3 -m pip install --no-cache-dir -r requirements.txt -r extra-requirements.txt; \
+    rm requirements.txt extra-requirements.txt
+
+FROM deephaven/server:local-build
+COPY --from=install / /
```
niloc132, line 13: newline nit

docker/server/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+
+def buildDocker = Docker.registerDockerImage(project, 'buildDocker') {
+    dependsOn prepareDocker
+    inputDir.set dockerContext
+    inputs.files project.project(':docker-runtime-base').tasks.findByName('buildDocker').outputs.files
+    buildArgs.put('DEEPHAVEN_VERSION', project.version)
+    images.add('deephaven/server:local-build')
+}
+
+assemble.dependsOn buildDocker
```
niloc132, line 43: newline nit

docker/server/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+
+evaluationDependsOn ':docker-runtime-base'
+evaluationDependsOn ':grpc-api-server-native'
+
+def dockerContext = project.layout.buildDirectory.dir('context')
+
+def dockerLicenses = License.createFrom(project).syncDockerLicense().get()
+
+def prepareDocker = project.tasks.register('prepareDocker', Sync) {
+    def distTar = project.project(':grpc-api-server-native').tasks.findByName('distTar')
```
niloc132, line 16: instead of task output, we can just depend on it as an artifact in the dependencies {} block, and skip the evaluationDependsOn - we don't care about the _task_, we care about the artifact.

grpc-api/server/native/README.md
```diff
@@ -8,11 +8,5 @@
 
 produces
 
-* `grpc-api/server/native/build/distributions/grpc-api-native-<version>.tar`
-* `grpc-api/server/native/build/distributions/grpc-api-native-<version>.zip`
-
-### Run
-
-```shell
-./gradlew grpc-api-server-native:run
```
niloc132, line 17: this is still runnable, no?

docker/server-pytorch/build.gradle
```diff
@@ -0,0 +1,22 @@
+plugins {
+    id 'com.bmuschko.docker-remote-api'
+}
+
+evaluationDependsOn ':docker-server'
+
+def dockerContext = project.layout.buildDirectory.dir('context')
+
+def prepareDocker = project.tasks.register('prepareDocker', Sync) {
```
niloc132, line 9: wouldn't it make more sense to put these contents in docker/server itself? we don't have a plugin/extension which can only happen once per project, these are just plain tasks, so we can stack them up (or make them generic and declare several sets of them). This removes the need for evaluationDependsOn and project().tasks.findByName() calls, since the tasks are right there and can be strongly referenced

build.gradle
```diff
@@ -89,15 +89,28 @@ idea {
...
+        'envoy',
+        'grpc-proxy',
+        'web-client-ui',
+        'protoc',
+        'pyclient',
+        'sphinx',
+        'docker-java-and-python',
+        'docker-runtime-base',
+        'docker-server',
+        'docker-server-pytorch'] }
```
niloc132, line 113: newline before `] }` (and possibly trailing comma) so it is cleaner to add new projects?

build.gradle
```diff
@@ -89,15 +89,28 @@ idea {
     }
 }
 
+project.version = '0.7.0'
 ext.versionSource = versionSource
-ext.globalVersion = "0.7.0"
+ext.globalVersion = project.version
 
 // To set fishlibVersion, go to project-local gradle.properties file
```
niloc132, line 96: suggestion, while you're in there, kill this comment block,  which doesnt make sense in dhcc

niloc132:
I'll follow up with trying to implement this same basic pattern for the web image's contents - breaking it up into "here's an archive with contents that jetty can serve statically" and "here's a docker image based on nginx" will fit well into the work i'm doing.

grpc-api/server/native/src/main/resources/grpc-api-native.prop
devinrsmith: I think we'll want to completely re-do some bootstrap and properties stuff if we need to extract this out - right now, there isn't a good extensible way for packagers to add their own configuration on top. Potentially related to https://github.com/deephaven/deephaven-core/discussions/1105

grpc-api/server/native/build.gradle
devinrsmith: Yeah, I'd support a rename.

docker/server/build.gradle
devinrsmith: I think I've got that working now.

docker/server-pytorch/build.gradle
devinrsmith: I've been tending to prefer one docker image per project; and maybe there is a way in the future where we can use the docker images as artifacts (like you've suggested with the application.tar), and we can do away w/ the eval depends on. Even though it's a bit more setup, I think it's easier to understand as a self contained unit (something that JJ or others can grok more easily having it separated).

niloc132:
* arch support in bootstrap.properties - and anticipate py vers changes too
* tagging on upstream images when dhcc ships

file for future work:
 * wheels as gradle artifacts (as i'm doing with js)
 * parameterizing the dockerfiles with upstream names, since commands are the same otherwise

docker/server-pytorch/build.gradle
devinrsmith: Done

grpc-api/server/native/src/main/resources/grpc-api-native.prop
devinrsmith: Fixed

devinrsmith:
This doesn't have "arch support in bootstrap.properties" - happy to extend support, but I've not tried to build arm64. I'm assuming it may be better as a follow up to get full stack arm64 support?

docker/server/build.gradle
```diff
@@ -0,0 +1,92 @@
...
+        def deephavenJpyWheel = project(':deephaven-jpy').tasks.getByName('buildWheel')
+
+        // deephaven.whl
+        def deephavenWheel = project(':deephaven-wheel').tasks.getByName('buildWheel')
+
+        // deephaven2.whl
+        def deephaven2Wheel = project(':deephaven2-wheel').tasks.getByName('buildWheel')
+
+        from 'src/main/common'
+        from "src/main/${baseMap[base]}"
```
niloc132, line 56: this looks like the only different 'from', everything else coming into this sync is the same, right? could we take advantage of that somehow, and not end up with both so many copies of these files (the applicationDist for example isn't small) and also dependencies such that any chance to any of them require syncing again (i.e. push them down into the DockerBuildImage task dependencies?

I can't obviously see a way to do this with very small changes, but consider this - just one sync task, with this rough structure
```
docker/
  common/
    *.whl
    serverApplicationDist/server-1.2.3.tar
    licenses/
    ... (unroll common here too)
  server/
  server-nltk/
  server-pytorch/
  (etc)
```
Then in the dockerfile, since we already have $BASE, can we point at the one directory as the docker context for each build? granted, we still have to upload each time to docker as the context, but we had to do that already anyway.

docker/server/src/main/common/image-bootstrap.properties
```diff
@@ -0,0 +1,11 @@
+Configuration.rootFile=grpc-api.prop
+io.deephaven.configuration.PropertyInputStreamLoader.override=io.deephaven.configuration.PropertyInputStreamLoaderTraditional
+LoggerFactory.silenceOnProcessEnvironment=true
+workspace=.
+devroot=.
+stdout.toLogBuffer=true
+stderr.toLogBuffer=true
+
+jpy.jpyLib=/usr/local/lib/python3.7/dist-packages/jpy.cpython-37m-x86_64-linux-gnu.so
```
niloc132, line 9: this file could be modified when copied, such that the arch and py version can be parameterized by a future map.

definitely not requesting that you try arm, just requesting that the arg be made available, plumbed through the gradle/docker barrier so we have one place to change, as long as we're already providing the ability to parameterize the server build.

docker/server-slim/src/main/docker/image-bootstrap.properties
```diff
@@ -1,7 +1,9 @@
-Configuration.rootFile=grpc-api-docker.prop
+Configuration.rootFile=grpc-api.prop
 io.deephaven.configuration.PropertyInputStreamLoader.override=io.deephaven.configuration.PropertyInputStreamLoaderTraditional
 LoggerFactory.silenceOnProcessEnvironment=true
 workspace=.
 devroot=.
 stdout.toLogBuffer=true
 stderr.toLogBuffer=true
+
+deephaven.console.type=groovy
```
niloc132, line 9: deliberately setting this to unconditionally be groovy?

grpc-api/server/native/README.md
niloc132: :+1: Jianfeng and I have a short-ish set of steps to make py work from either netty or jetty run locally, without docker.

docker/server/build.gradle
```diff
@@ -0,0 +1,92 @@
...
+def buildDockerClosure = { String base ->
+    def prepareDockerBase = prepareDockerClosure(base)
+    prepareDocker.dependsOn prepareDockerBase
+    Docker.registerDockerImage(project, "buildDocker-${baseMap[base]}") {
+        dependsOn prepareDockerBase
+        inputDir.set context(base)
+        inputs.files Docker.registryFiles(project, base)
+        buildArgs.put('BASE', "deephaven/${base}:local-build")
+        buildArgs.put('DEEPHAVEN_VERSION', project.version)
+        images.add("deephaven/${baseMap[base]}:local-build")
```
niloc132, line 87: to avoid changing this yet again in a few days, would you consider `deephaven/${baseMap[base]}-netty:local-build` as the docker image template, so that a jetty variant can exist for those who a) want docker (because reasons), but b) don't want to copy each in-transit payload through the websocket and envoy proxies?

Happy to consider another naming convention that achieves the same goal

docker/server/src/main/common/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN set -eux; \
+    ln -s /opt/deephaven/server-${DEEPHAVEN_VERSION} /opt/deephaven/server
+
+COPY image-bootstrap.properties image-bootstrap.properties
+
+FROM $BASE
+COPY --from=install / /
+VOLUME /data
+VOLUME /cache
+HEALTHCHECK --interval=3s --retries=3 --timeout=11s CMD /bin/grpc_health_probe -addr=localhost:8080 -connect-timeout=10s || exit 1
```
niloc132, line 43: irritating that we can't just FROM this in, and will have to change the settings in more than one place if we alter our expectations

docker/server/build.gradle
```diff
@@ -0,0 +1,92 @@
...
+    }
+}
+
+assemble.dependsOn baseMap.keySet().collect { base -> buildDockerClosure(base) }
```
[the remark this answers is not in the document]
devinrsmith: Yes, will fix.

docker/server/build.gradle
devinrsmith: Good idea. I think this will be easy.

docker/server-slim/src/main/docker/image-bootstrap.properties
devinrsmith: In `deephaven/server-slim` we don't have python installed; so setting it unconditionally here. I think a better solution would be for the software layer to choose the best value when it's unset (right now it defaults to python), but there are some unclean configuration implementation details that I'd rather not bake in to the software layer w/o some cleaner refactoring.

docker/server/src/main/common/Dockerfile
devinrsmith: I think we can generalize a bit in the future, and create a wrapper script around `grpc_health_probe` that does the right thing against the configurations we set (TLS, timeouts, etc).

docker/server/src/main/common/image-bootstrap.properties
devinrsmith: I'd actually like to parameterize for architecture first before trying to architecture for python version; and before parameterizing for python version, I think we should work on building up the stack as a virtual environment.

This bootstrap file is meant to aid in just the docker image building process; I expect we'll need to parameterize our native run instructions more explicitly to pick up python.

docker/server/build.gradle
devinrsmith: I think we should discuss with group; but I'm generally in support of:

* `deephaven/server`, `deephaven/server-jetty` be our defaults
* `deephaven/server-netty` be our alternative

Ditto pattern for all the extended images.

Alternatively, it *may* be possible to package netty and jetty jars, and have it be a runtime configuration? ie,

```
docker run --rm -it deephaven/server --netty
```

```
docker run --rm -it deephaven/server --jetty
```
?

docker/server-slim/src/main/docker/image-bootstrap.properties
niloc132: ah i misunderstood, looking locally I see it makes sense.

docker/server/src/main/common/image-bootstrap.properties
niloc132: Parameterizing for arch outside of properties like this is only an extra 1-3 lines in Docker.groovy - with the other images moved out externally (and already building aarch64), this should be the last piece we need to fix.

docker/server/build.gradle
niloc132: i'd vote for -netty/-jetty suffixes over a flag and expanded classpath, but i also like having deephaven/server be an alias for one or the other, and as long as it runs behind envoy and the ws proxy, it won't matter which it is (i.e. clean upgrade path), so we can kick it down the road.

niloc132 (approved):
Approved, with the "file and followup on aarch64 in that properties file" follow-up

.github/workflows/build-ci.yml
```diff
@@ -40,11 +44,14 @@ jobs:
...
           echo "::set-output name=semver::$semver"
 
-      - name: Docker meta
-        id: docker_meta
+      - name: Docker meta server
+        id: docker_meta_server
         uses: docker/metadata-action@v3
         with:
-          images: ghcr.io/${{ github.repository_owner }}/grpc-api
+          # grpc-api deprecated
```
JamesXNelson, line 51: what up w/ this comment?

JamesXNelson (approved):
rubber stamp for the gitlab file; there's one comment that maybe worth deleting, and I'm not uptodate on what crane is doing, but the changes to the file itself look valid.

PR #1603, opened 2021-11-23
Added the 'placeholder' write_csv func, fixes #1542

### LINKED ISSUE #1542: A barebone csv.write method in the new server Python package
Since the new CSV writer isn't available yet, it makes sense to implement a simple, extensible writer in the new Python package to complete the story first. It should be designed in a way that won't break compatibility when a full functional wrapper is needed for the new CSV writer.

jmao-denver:
A write_csv func with miniaml number of params as a temporary fillgap
for future full-featured implementation

### DIFF
--- a/pyintegration/deephaven2/table.py
+++ b/pyintegration/deephaven2/table.py
@@ -93,6 +93,12 @@ class Table:
     def get_dh_table(self):
         return self._j_table
 
+    def write_csv(self: Table, path: str, cols: List[str] = []) -> None:
+        try:
+            _JTableTools.writeCsv(self._j_table, path, *cols)
+        except Exception as e:
+            raise DHError("write csv failed.") from e
+
     @property
     def size(self) -> int:
         """ The current number of rows in the table. """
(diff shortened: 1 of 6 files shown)


pyintegration/deephaven2/table.py
```diff
@@ -89,6 +89,12 @@ def __repr__(self):
     def get_dh_table(self):
         return self._j_table
 
+    def write_csv(self: Table, path: str, cols: List[str] = []) -> None:
+        try:
+            _JTableTools.writeCsv(self._j_table, path, *cols)
+        except Exception as e:
+            raise DHError("write csv failed.") from e
+
```
chipkent, line 97: I don't think we want this method on the table.  We support many data formats and have finally started packaging each format into its own module.  By having this method here, we have decided that CSV is somehow more special than the other formats, and we are willing to add it to the table API.  I think the method in the CSV module is sufficient.  If it isn't, we can easily add a method here in the future.  It is harder to take a method away from here.

chipkent (changes requested):

chipkent (approved):

PR #1604, opened 2021-11-24
Conditional Formatting - Throw an error if multiple formats of the same type are set

mofojed:
When multiple formats for the same column are set, it only takes the last one. Instead of allowing the user to do something wrong and be confused why it doesn't work, throw.

Related to #1375

### DIFF
--- a/web/shared-beans/src/main/java/io/deephaven/web/shared/data/CustomColumnDescriptor.java
+++ b/web/shared-beans/src/main/java/io/deephaven/web/shared/data/CustomColumnDescriptor.java
@@ -1,10 +1,7 @@
 package io.deephaven.web.shared.data;
 
 import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
 
 /**
  * We still send plain strings to the server and receive plain strings from the client.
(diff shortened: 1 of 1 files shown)


web/shared-beans/src/main/java/io/deephaven/web/shared/data/CustomColumnDescriptor.java
```diff
@@ -95,9 +92,15 @@ public static boolean isCompatible(List<CustomColumnDescriptor> was, List<Custom
     }
 
     public static List<CustomColumnDescriptor> from(String[] newCustomColumns) {
+        final Set<String> descriptorNames = new HashSet<>();
         final List<CustomColumnDescriptor> list = new ArrayList<>();
         for (String col : newCustomColumns) {
-            list.add(new CustomColumnDescriptor().setExpression(col));
+            CustomColumnDescriptor descriptor = new CustomColumnDescriptor().setExpression(col);
+            if (descriptorNames.contains(descriptor.getName())) {
```
niloc132, line 99: you could put the `descriptorNames.add(descriptor.getName())` in here instead - Set.add returns a boolean, true if the set didn't contain the item already:

```
            if (!descriptorNames.add(descriptor.getName())) {
```
and then not add it below

niloc132 (approved):
Approved, could make it a line shorter, up to you.

niloc132 (approved):

PR #1620, opened 2021-12-01
Remove race conditions from TestModelFarm

### LINKED ISSUE #1150: TestModelFarm: Racy unit test creates spurious failures in CI
```
io.deephaven.base.verify.RequirementFailure: Requirement failed: required terminated == true, instead terminated == false, testShutdownTimeoutSecs == 1.0.
	at io.deephaven.base.verify.Require.requirement(Require.java:247)
	at io.deephaven.base.verify.Require.requirement(Require.java:254)
	at io.deephaven.modelfarm.TestModelFarm.testModelFarmNoWorkShutdown(TestModelFarm.java:44)
```
The test in question sleeps for 100 milliseconds and assumes that's enough time for a condition to be met. This isn't a good strategy for reliable, consistent unit testing.

chipkent:
Remove race conditions from TestModelFarm

Resolves #1150

### DIFF
--- a/ModelFarm/src/test/java/io/deephaven/modelfarm/TestModelFarm.java
+++ b/ModelFarm/src/test/java/io/deephaven/modelfarm/TestModelFarm.java
@@ -32,7 +32,7 @@ public class TestModelFarm extends TestCase {
         final CountDownLatch latchThreadStart = new CountDownLatch(0);
         final CountDownLatch latchThreadComplete = new CountDownLatch(0);
         final ModelFarmTick<Long, MutableLong, RowDataManager<Long, MutableLong>> modelFarmTick =
-                getModelFarmTick(nModelFarmThreadsDefault, 0, latchThreadStart, latchThreadComplete,null);
+                getModelFarmTick(nModelFarmThreadsDefault, 0, latchThreadStart, latchThreadComplete, null);
 
         Require.eq(modelFarmTick.getState(), "modelFarmTick.getState()", ModelFarmBase.State.WAITING);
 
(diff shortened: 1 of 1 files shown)


ModelFarm/src/test/java/io/deephaven/modelfarm/TestModelFarm.java
```diff
@@ -51,32 +54,33 @@ public void testModelFarmNoWorkShutdown() throws Exception {
...
         // the model farm will still be busy -- so its state should be SHUTDOWN, not TERMINATED.
         modelFarmTick.shutdown();
         Require.eq(modelFarmTick.getState(), "modelFarmTick.getState()", ModelFarmBase.State.SHUTDOWN);
 
         // Test awaitTermination(), without calling terminate().
-        // This should return true, we are waiting for (workerSleepMs * 2) milliseconds but all work should be finished
-        // within (workerSleepMs) milliseconds.
-        boolean terminated = modelFarmTick.awaitTermination(workerSleepMs * 2, TimeUnit.MILLISECONDS);
+        // This should return true, since threads are finished.
+        latchThreadComplete.countDown();
```
rcaudy, line 82: This doesn't make sense to me. You're making it so the ModelFarm shutdown won't stop the threads until some external latch is counted down. That seems contrary to what you would want.

chipkent: The latch is used to control when the ModelFarm threads are complete.  This allows me to test shutdown vs terminate.  Shutdown is a soft kill that lets work complete.  Terminate is a hard kill that kills running threads.  It is better than the multiple sleeps in the prior version.

rcaudy (dismissed):

ModelFarm/src/test/java/io/deephaven/modelfarm/TestModelFarm.java
```diff
@@ -51,32 +54,33 @@ public void testModelFarmNoWorkShutdown() throws Exception {
...
+        boolean terminated = modelFarmTick.awaitTermination(5000, TimeUnit.MILLISECONDS);
```
rcaudy, line 83: While this is far more likely to succeed, there's still a timing component to it. I'm not sure you can do better under the circumstances, besides waiting forever and potentially blocking tests if you have a bug.

rcaudy (approved):

Issue #1101, opened 2021-08-21
Python not matching by aggregation with lists


devinrsmith:
Jpy method matching does not work in this case when provided with a python list. It should match

`Table.by(Collection<? extends Selectable> groupByColumns, Collection<? extends Aggregation> aggregations)`

```python
from deephaven import TableTools
import jpy

ColumnName = jpy.get_type('io.deephaven.api.ColumnName').of
AggMax = jpy.get_type('io.deephaven.api.agg.Aggregation').AggMax
AsList = jpy.get_type('java.util.Arrays').asList

works = TableTools.timeTable("00:00:01")\
    .updateView("I=i%4")\
    .by(AsList(ColumnName("I")), AsList(AggMax("Max=Timestamp")))

broken = TableTools.timeTable("00:00:01")\
    .updateView("I=i%4")\
    .by([ ColumnName("I") ], [ AggMax("Max=Timestamp") ])
```

```
Value: no matching Java method overloads found
```

devinrsmith:
Note: we could fix this natively at the jpy layer, or at the Table wrapper layer.

jmao-denver:
his fails earlier than #1102, possibly related to the generics used in the method signature, but need to debug the code to find out exactly why JPY failed to match a overload method, so far the quick code study hasn't been sufficient for a confident diagnosis.

PR #1622, opened 2021-12-01
Worked around a JPY param matching issue, fixes #1102, #1101

### LINKED ISSUE #1102: Python bad matching by with list
Jpy is matching the wrong method when provided with a python list in the by case.

`Table.by(Collection<? extends Selectable> groupByColumns)`

```python
import jpy

ColumnName = jpy.get_type('io.deephaven.api.ColumnName').of
AsList = jpy.get_type('java.util.Arrays').asList

works = TableTools.timeTable("00:00:01")\
    .updateView("I=i%4")\
    .by(AsList(ColumnName("I")))

broken = TableTools.timeTable("00:00:01")\
    .updateView("I=i%4")\
    .by([ ColumnName("I") ])
```

```
Type: <class 'ValueError'>
Value: cannot convert a Python 'io.deephaven.api.ColumnName' to a Java 'io.deephaven.db.v2.select.SelectColumn'
```

This error suggests that jpy is trying to call `Table.by(SelectColumn... groupByColumns)`

jmao-denver:
1. created ArrayList in Python to hold collections of objects to pass to Java methods
2. removed temp hacky code in the server

### DIFF
--- a/pyintegration/deephaven2/table.py
+++ b/pyintegration/deephaven2/table.py
@@ -18,8 +18,7 @@ _JColumnName = jpy.get_type("io.deephaven.api.ColumnName")
 _JSortColumn = jpy.get_type("io.deephaven.api.SortColumn")
 _JFilter = jpy.get_type("io.deephaven.api.filter.Filter")
 _JFilterOr = jpy.get_type("io.deephaven.api.filter.FilterOr")
-_JTWD = jpy.get_type("io.deephaven.engine.table.impl.TableWithDefaults")
-_JAggHolder = jpy.get_type("io.deephaven.engine.table.impl.TableWithDefaults$AggHolder")
+_JArrayList = jpy.get_type("java.util.ArrayList")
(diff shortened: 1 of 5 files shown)


pyintegration/deephaven2/table.py
```diff
@@ -570,7 +569,10 @@ def sort_column(col, dir_):
         try:
             if order:
                 sort_columns = [sort_column(col, dir_) for col, dir_ in zip(order_by, order)]
-                return Table(j_table=_JTWD.sort(self.j_table, *sort_columns))
+                j_sc_list = _JArrayList()
+                for sc in sort_columns:
```
rcaudy, line 573: Why don't we write a method "convert_list" or similar? Also, since we know the size of the pylist, we should size the arraylist.

rcaudy (approved):
This is good. It might be nice to use the new function in the two unit tests, too.

PR #1637, opened 2021-12-04
Replace the IndexSet in the deephaven.learn package with a RowSet

### LINKED ISSUE #1636: Replace the IndexSet in the deephaven.learn package with a RowSet
The IndexSet in the deephaven.learn package can be replaced with a RowSet to yield a more efficient implementation by allowing the chunk interface to fill arrays.

chipkent:
Replace the IndexSet in the deephaven.learn package with a RowSet.

Resolves #1636.


jjbrosnan (dismissed):
Looks good to me.  I'll make sure to pull this into my Gatherer PR once it's merged into main and make any necessary changes so we can get that merged ASAP as well.

rcaudy (commented):
I've commented on the code as it is. That said, I'd like to understand a little bit more. Using a `RowSet` this way is fine, but it's quite distinct from the code @jjbrosnan showed me the other day.
The main concern I have is that if the use case is "two phase," e.g.
1. Specify a set of rows.
2. Do work for those rows.

If that's the case, we should be using builders (sequential or random, depending) instead of serial `RowSet.insert`.

chipkent:
Let me try to explain the use case.  This is the complex model we came up with to avoid inventing a bunch of complex engine logic.

The key high level logic is in the `deephaven.learn.learn` python function.  
https://github.com/chipkent/deephaven-core/blob/5c9bc8711c715cecf1d6f9cd6362c926c1b627b3/Integrations/python/deephaven/learn/__init__.py#L230

```
        return table.update(f"{futureOffset} = __computer.compute(k)")\
                .update(__scatterer.generateQueryStrings(f"{futureOffset}"))\
                .update(f"{clean} = __computer.clear()")\
                .dropColumns(f"{futureOffset}", f"{clean}")
```

Here we create a `Computer` and call `compute(k)` for each new or modified row.  `Computer.compute(k)` returns a `FutureOffset` which contains an index into the computed data plus a `Future`.  The `Future` is a future calculation to perform.  It contains a `RowSet` of rows being processed in a batch, a functions to apply, plus relevant data sources.

Once `batchSize` rows have been added to `Computer.compute(k)`, the `Computer` creates a new `Future` for the next batch.

...so, your assessment above about the two-phase model of (1) specifying a set of rows and (2) doing the work on those rows is correct.

How should we approach this in the most efficient way?

rcaudy (approved):

PR #1642, opened 2021-12-06
Fix ModelFarm race conditions and unit tests

### LINKED ISSUE #1150: TestModelFarm: Racy unit test creates spurious failures in CI

### LINKED ISSUE #1641: Race condition in ModelFarm shutdown
The ModelFarm shutdown methods have a race condition bug.

chipkent:
ModelFarm shutdown methods have a race condition that triggered occasional unit test failures.  Bugs were also found in the unit tests.

Resolves #1641 
Resolves #1150

### DIFF
--- a/ModelFarm/src/main/java/io/deephaven/modelfarm/ModelFarmBase.java
+++ b/ModelFarm/src/main/java/io/deephaven/modelfarm/ModelFarmBase.java
@@ -56,7 +56,7 @@ public abstract class ModelFarmBase<DATATYPE> implements ModelFarm {
          * Performs an operation using data from a query.
          *
          * @param usePrev Whether to use the previous data at a given row key when retrieving data (i.e. if
-         *                {@code true}, use {@link ColumnSource#getPrev} instead of {@link ColumnSource#get}).
+         *        {@code true}, use {@link ColumnSource#getPrev} instead of {@link ColumnSource#get}).
          */
         void retrieveData(boolean usePrev);
 
@@ -151,7 +151,7 @@ public abstract class ModelFarmBase<DATATYPE> implements ModelFarm {
      * Create a multithreaded resource to execute data driven models.
      *
      * @param nThreads number of worker threads.
-     * @param model    model to execute.
+     * @param model model to execute.
      */
     @SuppressWarnings("WeakerAccess")
     protected ModelFarmBase(final int nThreads, final Model<DATATYPE> model) {
@@ -183,7 +183,7 @@ public abstract class ModelFarmBase<DATATYPE> implements ModelFarm {
     /**
      * Interface for getting the most recent row data for a unique identifier.
      *
-     * @param <KEYTYPE>  unique ID key type
+     * @param <KEYTYPE> unique ID key type
      * @param <DATATYPE> data type
      */
     @FunctionalInterface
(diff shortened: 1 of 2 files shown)


ModelFarm/src/main/java/io/deephaven/modelfarm/ModelFarmBase.java
```diff
@@ -340,8 +340,8 @@ public final boolean awaitTermination(final long timeout, final TimeUnit unit) {
                 timeout == Long.MAX_VALUE ? Long.MAX_VALUE : System.currentTimeMillis() + unit.toMillis(timeout);
         boolean allThreadsTerminated = false;
 
-        while (!allThreadsTerminated && System.currentTimeMillis() < timeoutMillis) {
-            synchronized (ModelFarmBase.this) {
+        synchronized (ModelFarmBase.this) {
+            while (!allThreadsTerminated && System.currentTimeMillis() < timeoutMillis) {
```
rcaudy, line 344: You should record the current time (or the remaining time) from this line and use that in the `wait` below on L348. Currently you run the risk of an `IllegalArgumentException` if you ask to wait for a negative time, or blocking indefinitely if you ask to wait for zero milliseconds.

ModelFarm/src/test/java/io/deephaven/modelfarm/TestModelFarm.java
```diff
@@ -22,7 +22,7 @@
 public class TestModelFarm extends TestCase {
 
     private final long testShutdownTimeoutSecs =
-            Configuration.getInstance().getIntegerWithDefault("TestModelFarm.testShutdownTimeoutSecs", 1);
+            Configuration.getInstance().getIntegerWithDefault("TestModelFarm.testShutdownTimeoutSecs", 5);
```
rcaudy, line 25: Were you planning to commit a bigger default? Did CI in fact time out, or was something else wrong?

ModelFarm/src/main/java/io/deephaven/modelfarm/ModelFarmBase.java
chipkent: Good catch.

ModelFarm/src/test/java/io/deephaven/modelfarm/TestModelFarm.java
chipkent: I wasn't planning to increase the constant.  The bug was a branch that wasn't getting handled.  I just ran with the old constant and had zero failures for the first two (faster) test cases in >50k tries.

ModelFarm/src/main/java/io/deephaven/modelfarm/ModelFarmBase.java
```diff
@@ -360,6 +362,8 @@ public final boolean awaitTermination(final long timeout, final TimeUnit unit) {
                 } else {
                     allThreadsTerminated = true;
                 }
+
+                currentTime = System.currentTimeMillis();
```
rcaudy, line 366: Approving, but if you wanted this code to be as tight as you can make it I would either move to `System.currentTimeMillis()` to the where (that is, `while (!allThreadsTerminated && (currentTime = System.currenttimeMillis()) < timeoutMillis) {`) or add a `break` after the lines where `allThreadsTerminate = true;` is set.

rcaudy (approved):

PR #1652, opened 2021-12-07
Ensure the wrong number of parameters in a call to deephaven.ParquetTools.writeTable generates an error.

jcferretti:
As it is right now, and due to a mistake in an earlier change on that file, a call to `deephaven.ParquetTools.writeTable` with the wrong number of parameters (eg, 1, like just the filename or just the table name), silently does nothing instead of alerting the user
that something is wrong; this was not intentional and is misleading, eg, a user can mistake the (wrongful) no output for success.
I noticed this recently after having fell myself to that trap in some debugging session: I forgot one argument and for a while couldn't understand what was going on when the file was not showing up.


jmao-denver (dismissed):
Because writeTable() is generated, I understand the need for adding such 'unpythonic' check. It will not be necessary if we hand-wrap the class.

jmao-denver (approved):

PR #1654, opened 2021-12-07
Move `dephaven.learn` RowSet generation to use a builder.

### LINKED ISSUE #1653: `deephaven.learn` RowSet generation can be made more efficient by using builders
`deephaven.learn` RowSet generation can be made more efficient by using builders.

chipkent:
RowSet generation can be more efficient by using builders.

Resolves #1653

### DIFF
--- a/Integrations/src/main/java/io/deephaven/integrations/learn/Future.java
+++ b/Integrations/src/main/java/io/deephaven/integrations/learn/Future.java
@@ -2,8 +2,8 @@ package io.deephaven.integrations.learn;
 
 import io.deephaven.base.verify.Assert;
 import io.deephaven.engine.rowset.RowSet;
+import io.deephaven.engine.rowset.RowSetBuilderRandom;
 import io.deephaven.engine.rowset.RowSetFactory;
-import io.deephaven.engine.rowset.WritableRowSet;
 import io.deephaven.engine.table.ColumnSource;
 
 import java.util.function.Function;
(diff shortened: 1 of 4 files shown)


Integrations/src/main/java/io/deephaven/integrations/learn/Future.java
```diff
@@ -47,17 +49,16 @@
...
                 for (int i = 0; i < inputs.length; i++) {
-                    gathered[i] = gather(inputs[i], colSets[i]);
+                    gathered[i] = gather(inputs[i], colSets[i], rowSet);
                 }
 
                 result = func.apply(gathered);
             } finally {
-                rowSet.close();
-                rowSet = null;
+                rowSetBuilder = null;
```
rcaudy, line 61: I would do this inside of `makeRowSet()`. I also would probably use `rowSetBuilder == null` instead of `rowSetBuilt` (and get rid of the extra boolean).

Integrations/src/main/java/io/deephaven/integrations/learn/Future.java
```diff
@@ -34,7 +36,7 @@
         this.inputs = inputs;
         this.colSets = colSets;
         this.batchSize = batchSize;
-        this.rowSet = RowSetFactory.empty();
+        this.rowSetBuilder = RowSetFactory.builderRandom();
```
rcaudy, line 39: Are you certain you need random-order support, rather than `builderSequential()`?

rcaudy (commented):
Mostly looked at the production files, didn't study the tests too carefully.

Integrations/src/main/java/io/deephaven/integrations/learn/Future.java
rcaudy: Discussed, answered, needs random.

rcaudy (approved):
Approving in case you don't want to make the proposed change.

PR #1665, opened 2021-12-09
Added tests for avro mappings in consume. Reworked somewhat the impl.

jcferretti:
I had to eliminate the use of `IDENTITY` as a constant imported from `conversion_utils` in `ConsumeKafka.py`.  The issue I ran into is that in test mode, the value of `IDENTITY` ended up being `None`, instead of a distinct object that would be initialized if `_defineSymbols` was called correctly; somehow in testing this setup is not working.   Essentially I think using symbols defined initially as `None` and later redefined via `_defineSymbols` doesn't work in testing when done cross-imports.

Anyway, I eliminated the need for the `IDENTITY` constant by separating `_dictToFun` in two methods, one that would have an identity default (with a single parameter) and another one that takes the desired default as a second argument.

### DIFF
--- a/Integrations/python/deephaven/ConsumeKafka.py
+++ b/Integrations/python/deephaven/ConsumeKafka.py
@@ -11,7 +11,7 @@ import wrapt
 import deephaven.Types as dh
 
 from deephaven.conversion_utils import \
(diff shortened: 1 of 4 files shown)


jmao-denver (approved):
The changes are meant to improve the testability of the feature and they LGTM.

PR #1670, opened 2021-12-09
Check the docker registry directories for gradle.properties.

devinrsmith:
Otherwise, project may become unbuildable if you check out a branch that has a new registry, and then checkout a branch that doesn't have that registry.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -249,9 +249,11 @@ project(':docker-java-and-python').projectDir = file('docker/java-and-python')
 include(':docker-runtime-base')
 project(':docker-runtime-base').projectDir = file('docker/runtime-base')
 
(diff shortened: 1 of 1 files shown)


mofojed (approved):

devinrsmith:
I'm not against making the project structure better, but in the immediate term, this is a strict improvement to what already exists IMO (as we saw, Charles ran into an issues that would have been solved if he had this patch). We can get into conventions discussions `foo:bar` vs `foo-bar`, opt-in vs opt-out, etc... but I don't want that to hold up a useful fix.

devinrsmith:
Added discussion: https://github.com/deephaven/deephaven-core/discussions/1684

PR #1673, opened 2021-12-09
Update python requirements.txt

devinrsmith:
Ran instructions in

* docker/java-and-python/src/main/docker/Dockerfile
* docker/runtime-base/src/main/docker/Dockerfile

Note: there are no changes in the runtime-base requirements - even though numpy is out of date, it looks like the latest version of numba places the requirement on an earlier version of numpy.


devinrsmith:
Irrelevant given https://github.com/deephaven/deephaven-core/pull/1602

Issue #1664, opened 2021-12-08
Start using official nginx release


devinrsmith:
It turns out, we are downgrading the official nginx image to the package maintainers version in an effort to get webdav support.

```
$ docker run --rm -it ghcr.io/deephaven/web:0.6.0 nginx -V 
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: IPv6 listen already enabled
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/99-init-notebooks.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
nginx version: nginx/1.14.2
built with OpenSSL 1.1.1d  10 Sep 2019
TLS SNI support enabled
[...]

PR #1675, opened 2021-12-09
Transition from nginx to deephaven-base-image's nginx-base.

devinrsmith:
Follow up #1664


devinrsmith:
Follow up later this month: https://github.com/deephaven/deephaven-core/issues/1676

devinrsmith:
Fixed by #1681

PR #1692, opened 2021-12-14
Add deephaven/server-slim to build-ci.yml

devinrsmith:


devinrsmith:
My new favorite tool is crane:

```shell
docker run --rm -it gcr.io/go-containerregistry/crane config ghcr.io/devinrsmith/server-slim:build-docker-publish-slim | jq '.'
```

niloc132 (approved):

PR #1697, opened 2021-12-15
Move docker LABELs into Dockerfiles

devinrsmith:
Allows us to parameterize the labels via build args in the Dockerfile instead of just during CI


.github/workflows/build-ci.yml
```diff
@@ -350,12 +315,19 @@ jobs:
...
         with:
           job-id: build-grpc-proxy
-          arguments: --scan grpc-proxy:prepareDocker
+          arguments: --scan outputVersion grpc-proxy:prepareDocker
           gradle-version: wrapper
 
+      - name: Get Deephaven Version
+        id: deephaven_version
+        run: |
+          echo "::set-output name=deephaven_version::$(cat build/version)"
```
mofojed, line 324: Interesting that some examples use `SCREAMING_SNAKE_CASE` and others just use `snake_case` for the outputs from set-output. I think I'd prefer screaming but meh doesn't really matter.

.github/workflows/build-ci.yml
```diff
@@ -434,12 +401,19 @@ jobs:
         uses: burrunan/gradle-cache-action@v1
         with:
           job-id: build-envoy
-          arguments: --scan envoy:prepareDocker
+          arguments: --scan outputVersion envoy:prepareDocker
           gradle-version: wrapper
 
+      - name: Get Deephaven Version
+        id: deephaven_version
```
mofojed, line 408: Do you need this again? The deephaven_version action is already above, doing the same thing...

docker/server/src/main/common/Dockerfile
```diff
@@ -42,4 +42,14 @@ COPY --from=install / /
 VOLUME /data
 VOLUME /cache
 HEALTHCHECK --interval=3s --retries=3 --timeout=11s CMD /bin/grpc_health_probe -addr=localhost:8080 -connect-timeout=10s || exit 1
-ENTRYPOINT [ "/opt/deephaven/server/bin/start", "image-bootstrap.properties" ]
\ No newline at end of file
+ENTRYPOINT [ "/opt/deephaven/server/bin/start", "image-bootstrap.properties" ]
+ARG DEEPHAVEN_VERSION
+ARG SERVER
```
mofojed, line 47: I might be missing this, but where is `SERVER` defined?

.github/workflows/build-ci.yml
devinrsmith: It's in a separate job

devinrsmith:
My new favorite command, to inspect remote repo image info without needing to pull:

```
docker run --rm -it gcr.io/go-containerregistry/crane config ghcr.io/devinrsmith/server:build-docker-labels | jq '.config.Labels'
```

mofojed (approved):

niloc132 (approved):

mofojed (approved):

PR #1698, opened 2021-12-15
Add support for local arm64 builds

devinrsmith:
@rcaudy has tested CLI and IJ


devinrsmith:
Note: I did not pull over the protoc (now, protoc-base in deephaven-base-images) changes after short chat with Colin that it's was mainly a build speed issues (M1 can emulate running AMD64 protoc-base), nor the Parquet configuration change.

buildSrc/src/main/groovy/io/deephaven/tools/docker/Architecture.groovy
```diff
@@ -0,0 +1,31 @@
...
+
+    Architecture(String dockerName) {
+        this.dockerName = dockerName
+    }
+
+    @Override
+    String toString() {
+        dockerName
+    }
+}
```
niloc132, line 31: missing newline

docker/server/build.gradle
```diff
@@ -1,17 +1,25 @@
...
     id 'com.bmuschko.docker-remote-api'
 }
 
-def baseMap = [
-    'server-base': 'server',
-    'nltk-base': 'server-nltk',
-    'pytorch-base': 'server-pytorch',
-    'sklearn-base': 'server-sklearn',
-    'tensorflow-base': 'server-tensorflow',
+def hostArch = Architecture.fromHost()
```
niloc132, line 8: make this configurable from a property so others can test locally (and so github actions can produce images via ci)?

docker/server/build.gradle
```diff
@@ -1,17 +1,25 @@
...
+def baseMapAmd64 = [
+        'server-base': 'server',
+        'nltk-base': 'server-nltk',
+        'pytorch-base': 'server-pytorch',
+        'sklearn-base': 'server-sklearn',
+        'tensorflow-base': 'server-tensorflow',
 ]
 
+// Only the server image is supported on arm64
+def baseMapArm64 = baseMapAmd64.findAll { key, value -> value == 'server' }
```
niloc132, line 19: this is very unfortunate - do none of these libraries support arm yet?

docker/server/src/main/arch/arm64/image-bootstrap.properties
```diff
@@ -0,0 +1,11 @@
+Configuration.rootFile=grpc-api.prop
+io.deephaven.configuration.PropertyInputStreamLoader.override=io.deephaven.configuration.PropertyInputStreamLoaderTraditional
+LoggerFactory.silenceOnProcessEnvironment=true
+workspace=.
+devroot=.
+stdout.toLogBuffer=true
+stderr.toLogBuffer=true
+
+jpy.jpyLib=/usr/local/lib/python3.7/dist-packages/jpy.cpython-37m-aarch64-linux-gnu.so
```
niloc132, line 9: Rather than duplicating this file just to set one property (of two we know we'll need) can we not template it? Either gradle work out the file necessary for each build context, or let some RUN make it out of env vars, or compute the paths based on the image's contents?

grpc-proxy/build.gradle
```diff
@@ -9,23 +10,30 @@ evaluationDependsOn Docker.registryProject('alpine')
...
     }
-    into 'build/docker'
+    into context
 }
 
 def buildDocker = Docker.registerDockerImage(project, 'buildDocker') {
+    inputDir.set context
     inputs.files prepareDocker.get().outputs.files
     inputs.files Docker.registryFiles(project, 'alpine')
+    buildArgs.put('TARGETARCH', Architecture.fromHost().toString())
```
niloc132, line 36: parameterize so we can produce all images in a build for the specified arch

docker/server/build.gradle
devinrsmith: I haven't spent the time yet trying to matrix them out yet.

niloc132: TODO and issue for followup?

docker/server/src/main/docker/Dockerfile
```diff
@@ -29,13 +29,17 @@ RUN set -eux; \
...
 ARG DEEPHAVEN_VERSION
-
 ADD serverApplicationDist/server-${DEEPHAVEN_VERSION}.tar /opt/deephaven
 
 RUN set -eux; \
     ln -s /opt/deephaven/server-${DEEPHAVEN_VERSION} /opt/deephaven/server
 
-COPY image-bootstrap.properties image-bootstrap.properties
+COPY configure/ /configure
+RUN set -eux; \
```
niloc132, line 38: as discussed, it makes more sense to me to either have a common startup script that passes this to the jvm, or have the jvm call out to python and ask for this, so that docker and local jvm behave the same

niloc132: as discussed on call, lets just file an issue and TODO this for later, we have a working option, and we think we can make it magically work at runtime.

niloc132 (approved):

PR #1707, opened 2021-12-16
Reduce scope of dependencies for faster/simplier rebuilds

niloc132:
This is an attempt to use implementation when we can over api (or compile), to better force projects to explicitly declare the dependencies they use and export. The effect of this will be fewer recompiles when not necessary, smaller classpaths for those compiles (which can make them faster), and better parallelization (due to a sparser dependency graph).

The "new" dependencies introduced in this commit were ones that were transitive before with no need to be (i.e. not part of the actual API, but exposed anyway). Some of these are negotiable, like what should the grpc-api project actually expose to downstream projects - this is just a proposal that appears to build as-is, cherry-picked from the jetty work, to make that eventual merge smaller.

Partial #1270

### DIFF
--- a/buildSrc/src/main/groovy/io.deephaven.java-conventions.gradle
+++ b/buildSrc/src/main/groovy/io.deephaven.java-conventions.gradle
@@ -106,5 +106,5 @@ jar {
 }
 
 configurations {
-  testOutput.extendsFrom testRuntime
+  testOutput.extendsFrom testRuntimeClasspath
 }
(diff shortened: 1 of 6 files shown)


niloc132:
Some dependencies are not propagating to grpc-api flight test - in the future this will be its own project, testing by the netty and jetty server implementations separately, but should be fixed for this commit.

devinrsmith:
Happy to approve once passing tests, and continue limiting dependencies in the future, as we need to upgrade gradle sometime.

devinrsmith (approved):

PR #1708, opened 2021-12-16
Add implicit jpy configuration based on the python environment

### LINKED ISSUE #1703: Native-compatible python startup improvements
We'd like to better support python startup in native (and docker) environments.

devinrsmith:
Fixes #1703


devinrsmith:
@jmao-denver has used this PR to get jetty+py running natively.

engine/table/src/main/java/io/deephaven/engine/util/PythonDeephavenSession.java
```diff
@@ -77,8 +78,14 @@ public PythonDeephavenSession(
             @Nullable final Listener listener, boolean runInitScripts, boolean isDefaultScriptSession)
             throws IOException {
         super(listener, isDefaultScriptSession);
-
-        JpyInit.init(log);
+        try {
+            JpyInit.init(log);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
```
rcaudy, line 85: `UncheckedDeephavenException` or a subclass, pls.

engine/table/src/main/java/io/deephaven/engine/util/PythonDeephavenSession.java
```diff
@@ -77,8 +78,14 @@ public PythonDeephavenSession(
             @Nullable final Listener listener, boolean runInitScripts, boolean isDefaultScriptSession)
             throws IOException {
         super(listener, isDefaultScriptSession);
-
-        JpyInit.init(log);
+        try {
+            JpyInit.init(log);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
rcaudy, line 84: This just sets a landmine for other code.

py/jpy-config/src/main/java/io/deephaven/jpy/PropertiesFromSubprocess.java
```diff
@@ -0,0 +1,61 @@
...
+        final Process process;
+        try (final InputStream in = PropertiesFromSubprocess.class.getResourceAsStream("introspect.py")) {
+            if (in == null) {
+                throw new IllegalStateException("Expected to find introspect.py resource");
+            }
+            // The "-" means that python will read script from stdin
+            try {
+                process = new ProcessBuilder(pythonName, "-").start();
+            } catch (IOException e) {
+                if (e.getMessage().contains("error=2, No such file or directory")) {
```
rcaudy, line 26: This looks unreliable.
Maybe a case-insensitive check for "no such" ? Also a check for FNFE?

py/jpy-config/src/main/java/io/deephaven/jpy/PropertiesFromSubprocess.java
```diff
@@ -0,0 +1,61 @@
...
+        } catch (Throwable t) {
+            process.destroy();
+            throw t;
+        }
+        final int exitValue = process.exitValue();
+        if (exitValue != 0) {
+            final String error = new String(process.getErrorStream().readAllBytes(), StandardCharsets.UTF_8);
+            if (error.contains("ModuleNotFoundError: No module named 'jpyutil'")) {
+                throw new IllegalStateException(
+                        String.format("A Deephaven python environment has not been configured for '%s'", pythonName));
```
rcaudy, line 52: This could use more help.

engine/table/src/main/java/io/deephaven/engine/util/PythonDeephavenSession.java
devinrsmith: Most code that does not _expect_ an interrupt  should either 1) rethrow the interrupted exception, or 2) mark the thread as interupted and return, or mark and throw an unchecked exception. Failing to do that, higher level code up the stack won't get the control they need - it's usually logic that grpc, thread pools, or thread-aware code is responsible for. Unless the piece of code is specifically dealing with a concurrency regime where it knows how to handle the interupts itself, this is the pattern to use. See article by Brian Goetz

https://itblackbelt.wordpress.com/2006/05/24/dealing-with-interruptedexception-by-brian-goetz/

Don’t swallow interrupts

Sometimes throwing InterruptedException is not an option, such as when a task defined by Runnable calls an interruptible method. In this case, you can’t rethrow InterruptedException, but you also do not want to do nothing. When a blocking method detects interruption and throws InterruptedException, it clears the interrupted status. If you catch InterruptedException but cannot rethrow it, you should preserve evidence that the interruption occurred so that code higher up on the call stack can learn of the interruption and respond to it if it wants to. This task is accomplished by calling interrupt() to “reinterrupt” the current thread, as shown in Listing 3. At the very least, whenever you catch InterruptedException and don’t rethrow it, reinterrupt the current thread before returning.

py/jpy-config/src/main/java/io/deephaven/jpy/PropertiesFromSubprocess.java
devinrsmith: I think maybe just re-wrapping the IOException makes sense.

py/jpy-config/src/main/java/io/deephaven/jpy/PropertiesFromSubprocess.java
devinrsmith: Added a bit more. I think once we have a solid article on virtual environments and official Deephaven native install procedure, we can add a link to better documentation.

engine/table/src/main/java/io/deephaven/engine/util/PythonDeephavenSession.java
niloc132: Yeah I'm firmly with Devin on this, though we could change the contract for script sessions to throw InterruptedException as another option, so it is clear that any session impl could do this.

py/jpy-config/src/main/java/io/deephaven/jpy/PropertiesFromSubprocess.java
niloc132: Testing on mac/windows seems important to make sure we get the same error. This code will run (that is, "fail") cleanly without jpy being built for the platform, so we don't even need to do that work.

engine/table/src/main/java/io/deephaven/engine/util/PythonDeephavenSession.java
rcaudy: There are really only two possibilities, here:
1. The higher level code expects an interrupt might happen. In that case, re-throwing InterruptedException makes the most sense.
2. The higher level code doesn't. It will never check the interrupt status, leaving it set. This is what I mean about "a landmine" - completely unrelated code that expects interruption might see the interrupted status left over in dirty state.

If you're planning to let the thread die anyway, it doesn't matter much. Otherwise, you aren't doing anyone any favors. If you think this should be handled by calling code, re-throw it.

rcaudy: I do see the point about thread pools, etc. My major concern is "any random thread we control". If we expect to outlive the exception, it's a liability to have that flag set.

devinrsmith: Arguably, we may want to change how startup occurs, so the session constructor isn't responsible for starting jpy.

rcaudy: Regardless, we don't expect to outlive the exception, right? At that point, this is just a moot discussion. The code can stay as-is.

rcaudy (approved):

niloc132 (approved):

PR #1712, opened 2021-12-16
Configure cache dir

devinrsmith:
We should not be writing to a container's `/tmp` directory in docker, unless it's a docker volume.

Our docker images declare a `/cache` volume, so we should:

1. Set `-Djava.io.tmpdir=/cache/tmp` so that any java code using temporary files will inherit a docker volume backed tmp space
2. Introduce `-Ddeephaven.cache.dir` so that those who want more explicit "cache" space (instead of "tmp" space), can take advantage of it. (Only `AbstractScriptSession` atm, but others could use it in the future.)

### DIFF
--- a/docker/server-slim/src/main/docker/image-bootstrap.properties
+++ b/docker/server-slim/src/main/docker/image-bootstrap.properties
@@ -8,4 +8,3 @@ stderr.toLogBuffer=true
 
 deephaven.console.type=groovy
 deephaven.cache.dir=/cache
-java.io.tmpdir=/cache/tmp
--- a/docker/server-slim/src/main/docker/image-bootstrap.properties
+++ b/docker/server-slim/src/main/docker/image-bootstrap.properties
@@ -7,3 +7,5 @@ stdout.toLogBuffer=true
 stderr.toLogBuffer=true
 
 deephaven.console.type=groovy
+deephaven.cache.dir=/cache
+java.io.tmpdir=/cache/tmp
--- a/engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
@@ -9,7 +9,6 @@ import io.deephaven.UncheckedDeephavenException;
 import io.deephaven.api.util.NameValidator;
 import io.deephaven.base.FileUtils;
 import io.deephaven.compilertools.CompilerTools;
-import io.deephaven.configuration.Configuration;
 import io.deephaven.engine.table.Table;
 import io.deephaven.engine.table.TableDefinition;
 import io.deephaven.engine.table.lang.QueryLibrary;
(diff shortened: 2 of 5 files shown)


docker/server-slim/src/main/docker/image-bootstrap.properties
```diff
@@ -6,4 +6,6 @@ devroot=.
 stdout.toLogBuffer=true
 stderr.toLogBuffer=true
 
-deephaven.console.type=groovy
\ No newline at end of file
+deephaven.console.type=groovy
+deephaven.cache.dir=/cache
+java.io.tmpdir=/cache/tmp
```
rcaudy, line 11: Can we make this a sub-dir of the host's `/tmp`?

engine/table/src/main/java/io/deephaven/engine/util/CacheDir.java
```diff
@@ -0,0 +1,32 @@
...
+import java.nio.file.Path;
+
+public final class CacheDir {
+    private static final String DEEPHAVEN_CACHE_DIR = "deephaven.cache.dir";
+    private static final String JAVA_IO_TMPDIR = "java.io.tmpdir";
+
+    private static final Path cacheDir;
+
+    static {
+        final String deephavenCacheDir = System.getProperty(DEEPHAVEN_CACHE_DIR);
```
rcaudy, line 12: `Configuration.getInstance().getPropertyWithDefault`?

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -61,7 +60,7 @@ protected AbstractScriptSession(@Nullable Listener changeListener, boolean isDef
         this.changeListener = changeListener;
 
         final UUID scriptCacheId = UuidCreator.getRandomBased();
-        classCacheDirectory = new File(CLASS_CACHE_LOCATION, UuidCreator.toString(scriptCacheId));
+        classCacheDirectory = CLASS_CACHE_LOCATION.resolve(UuidCreator.toString(scriptCacheId)).toFile();
```
rcaudy, line 63: We discussed introducing props:
INSTANCE_ID (if set, this is the ID)
INSTANCE_ID_FILE (if set and INSTANCE_ID isn't this is the file we load from or generate)
else, we randomly allocate an instance ID and be sure to clean up on exit.

engine/table/src/main/java/io/deephaven/engine/util/CacheDir.java
devinrsmith: ^ this is not an existing method. But I can use Configuration, sure.

docker/server-slim/src/main/docker/image-bootstrap.properties
devinrsmith: Punting - https://github.com/deephaven/deephaven-core/issues/1714

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
devinrsmith: Punting - https://github.com/deephaven/deephaven-core/issues/1713

engine/table/src/main/java/io/deephaven/engine/util/CacheDir.java
rcaudy: Sorry, method is `io.deephaven.configuration.PropertyFile#getStringWithDefault`.

rcaudy (approved):

rcaudy (approved):

PR #1719, opened 2021-12-20
Upgade grpc version, use bom instead of manually specifying each time

niloc132:
Dependencies are distributed to where they are used now, and
now-unnecessary proto/build.gradle is removed entirely.

Partial #1270

This does not resolve flight-core's excessive dependency problem (but does document it).


devinrsmith (approved):

buildSrc/src/main/groovy/Classpaths.groovy
```diff
@@ -81,6 +81,9 @@ class Classpaths {
     static final String GROOVY_GROUP = 'org.codehaus.groovy'
     static final String GROOVY_VERSION = '3.0.9'
 
+    static final String GRPC_GROUP = 'io.grpc'
+    static final String GRPC_VERSION = '1.42.0'
```
niloc132, line 85: looks like 1.43.0 is out now, probably should move to that

devinrsmith (approved):

PR #1720, opened 2021-12-20
Change dependencies on java-client dagger modules to implementation

devinrsmith:
POC


devinrsmith:
Not as applicable any more w/ jetty split, needs to be API.

PR #1727, opened 2021-12-22
Bump base images

devinrsmith:
Pandas dependency updated from 1.3.4 to 1.3.5


niloc132 (approved):
Approved, provided you tested our numpy integration, which we know to be brittle.

niloc132 (approved):

PR #1729, opened 2021-12-22
Add runtime dependency to grpc-netty for java-client-session-examples

devinrsmith:
Fixes this error: io.grpc.ManagedChannelProvider$ProviderNotFoundException: No functional channel service provider found. Try adding a dependency on the grpc-okhttp, grpc-netty, or grpc-netty-shaded artifact


devinrsmith:
Cloning in favor of https://github.com/deephaven/deephaven-core/pull/1738

Issue #1657, opened 2021-12-08
Tracking issue for running JPY and Deephaven wheels outside of docker


niloc132:
With #1270 in progress we're considering the other doors this will open for us, and one of them is using local python virtual environments (instead of ones in the docker image). This ticket is a placeholder to make sure this idea will work, and to consider other tasks.

Basic user story that this presents is that a python developer or data scientist could create a python venv with jpy and deephaven installed, and install/develop other wheels and other python code locally, then run the deephaven server-jetty binary pointing to that venv. No docker setup would be required to run it, just a zip/tar download and configuration to point at the venv.

jmao-denver:
As a step in this exploration, we want to make sure the deephaven fork of JPY can be built locally on MacOS. There is a compiler option change we need to the build script (setup.py):
```
diff --git a/py/jpy/setup.py b/py/jpy/setup.py
index bd91d382..3921347f 100644
--- a/py/jpy/setup.py
+++ b/py/jpy/setup.py
@@ -139,6 +139,11 @@ libraries = [jpyutil.JVM_LIB_NAME]
 define_macros = []
 extra_link_args = []
 extra_compile_args = ["-std=c99"]
+# making a big but probably safe assumption here that the compiler used to build Python is the same one we will be using
+compiler_name, compiler_version, *_ = platform.python_compiler().split()
+if compiler_name.upper() == "CLANG" and compiler_version > "10.0.0":
+    extra_compile_args.append("-Wno-error-implicit-function-declaration")
+

 if platform.system() == 'Windows':
     define_macros += [('WIN32', '1')]
```

With this change, we are able to build the wheel by running 'python setup.py bdist_wheel' in the py/jpy directory.
```
cp dist/deephaven_jpy-Unknown-cp37-cp37m-macosx_11_0_x86_64.whl /tmp
```
and install it with pip in a python venv as follows:
```
 ➜ /tmp python -m venv dockerless
 ➜ /tmp source dockerless/bin/activate
(dockerless) ➜  /tmp python -m pip install --upgrade pip
Collecting pip
  Using cached pip-21.3.1-py3-none-any.whl (1.7 MB)
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 20.1.1
    Uninstalling pip-20.1.1:
      Successfully uninstalled pip-20.1.1
Successfully installed pip-21.3.1
[...]

PR #1731, opened 2021-12-22
Support jetty as a server for deephaven

### LINKED ISSUE #1270: EPIC: Evaluate Jetty as replacement for Netty

niloc132:
The default gRPC-java server implementation is backed by Netty, which has some limitations for us, and has required us to distribute the application as a set of docker containers. This patch adds support for running the server inside of Jetty, a java servlet container, and also bundles in the static html/js/css required for the JS API and UI, as well as a websocket handler that can allow the browser to connect without full grpc support.

Fixes #1270


server/src/test/app.d/02-python.py
```diff
@@ -23,7 +23,7 @@ def demo_implicit():
   global world_imp; world_imp = TableTools.timeTable("00:00:01")
 
 def initialize_implicitly(func: Callable[[], None]):
-  app = jpy.get_type("io.deephaven.grpc_api.appmode.ApplicationContext").get()
+  app = jpy.get_type("io.deephaven.appmode.ApplicationContext").get()
```
niloc132, line 26: Odd that this isn't failing in main right now, I already checked in this change. Will validate that these tests actually can fail...

server/test/src/main/java/io/deephaven/server/test/FlightMessageRoundTripTest.java
```diff
@@ -215,7 +232,7 @@ protected void after() {
     @Test
     public void testSimpleEmptyTableDoGet() {
         Flight.Ticket simpleTableTicket = FlightExportTicketHelper.exportIdToFlightTicket(1);
-        currentSession.newExport(simpleTableTicket, "test")
+        currentSession.newExport(simpleTableTicket, "")
```
niloc132, line 235: probably rebase damage, will revert these string changes

web/client-api/src/main/java/io/deephaven/web/public/table_viewport.html
```diff
@@ -33,12 +33,12 @@ <h3>Table data</h3>
...
-    await ide.runCode(`
-from deephaven.TableTools import timeTable
-remoteTable = timeTable("00:00:01").updateView("I=i", "J=I*I", "K=I%100").lastBy("K")
-    `)
+    ide = await connection.startSession("groovy");
+await ide.runCode('remoteTable = timeTable("00:00:01").updateView("I=i", "J=I*I", "K=I%100").lastBy("K")')
+//     await ide.runCode(`
+// from deephaven.TableTools import timeTable
+// remoteTable = timeTable("00:00:01").updateView("I=i", "J=I*I", "K=I%100").lastBy("K")
+//     `)
```
niloc132, line 41: something like this should exist so that we can handle groovy or python from html samples in jetty

server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
```diff
@@ -0,0 +1,114 @@
...
+        sc.addConnectionFactory(factory);
+
+        final WebAppContext context =
+                new WebAppContext(null, "/", null, null, null, new ErrorPageErrorHandler(), SESSIONS);
+        try {
+            String knownFile = "/ide/index.html";
+            URL ide = JettyBackedGrpcServer.class.getResource(knownFile);
+            context.setBaseResource(Resource.newResource(ide.toExternalForm().replace("!" + knownFile, "!/")));
+        } catch (IOException ioException) {
+            ioException.printStackTrace();
```
devinrsmith, line 50: Should we throw unchecked io?

server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    public void shutdown() throws Exception {
+        jetty.stop();
+    }
+
+    @Override
+    public void shutdownNow() {
+        // jetty.stop();
```
devinrsmith, line 101: Not applicable?

server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    public void awaitTermination() throws InterruptedException {
+        jetty.join();
+    }
+
+    @Override
+    public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
+        jetty.join();
```
devinrsmith, line 111: Not taking into account timeout?

build.gradle
```diff
@@ -472,14 +473,14 @@ tasks.register('quick') {
...
-    it.dependsOn project(':grpc-api').tasks.findByName(LifecycleBasePlugin.CHECK_TASK_NAME)
+    it.dependsOn project(':server').tasks.findByName(LifecycleBasePlugin.CHECK_TASK_NAME)
     it.dependsOn project(':Generators').tasks.findByName(LifecycleBasePlugin.CHECK_TASK_NAME)
 }
 
 tasks.register('prepareCompose') {
     it.group 'Deephaven lifecycle'
     it.description 'A lifecycle task that prepares prerequisites for local docker-compose builds'
-    it.dependsOn project(':docker-server-slim').tasks.findByName('buildDocker-server-slim'),
+    it.dependsOn project(':docker-server-slim').tasks.findByName('buildDocker-server-slim'),//docker-server-slim-netty
```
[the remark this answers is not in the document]
devinrsmith: I'm very excited for jetty native and jetty dockerized without the need for extra containers.

Naming is OK for now IMO, but I suspect as jetty version becomes tested and is the easiest to use, we'll transition and/or continue with more renames.

devinrsmith (commented):
Good stuff. Some commented out stuff that we should probably cleanup?

Running locally, I see that server-jetty is serving up a META-INF listing from `http://localhost:10000`. Should we filter to only include ide/ and jsapi/?

I'm guessing it was this way before, but some of our "Feature Demos" in jsapi/ seem to be dependent on python? Might be good to generalize and use non-console apis? (I think this is the same as your comment on table_viewport file).

Should we auto-redirect to `ide/`?

server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
niloc132: This is fixed now, the whole GrpcServer interface and impls have been rewritten, doc'd.

server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
niloc132: Fixed now

niloc132:
> Should we auto-redirect to ide/?

Fixed.

server/src/test/app.d/02-python.py
niloc132: @devinrsmith can you confirm that these tests don't actually care what is in these files, as long as they exist? That seems to be the case, I can mangle their contents without failing any tests

devinrsmith: It looks like we have tests for them, but they are ignored.

See `io.deephaven.grpc_api.appmode.ApplicationTest#app02`.

niloc132: Do we have an issue for that? As this reflects, this seems subject to bitrot

server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+
+        // For the Web UI, cache everything in the static folder
+        // https://create-react-app.dev/docs/production-build/#static-file-caching
+        context.addFilter(NoCacheFilter.class, "/iriside/*", EnumSet.noneOf(DispatcherType.class));
+        context.addFilter(CacheFilter.class, "/iriside/static/*", EnumSet.noneOf(DispatcherType.class));
+
+        // Always add eTags
+        context.setInitParameter("org.eclipse.jetty.servlet.Default.etags", "true");
```
devinrsmith, line 61: Ah, here are the etags. Still interested in if we are over-caching with our cache contrrol directive.

server/src/test/app.d/02-python.py
devinrsmith: Doesn't look like it...

devinrsmith (commented):
I'll check it out locally soon and go through some QA.

server/jetty/README.md
```diff
@@ -0,0 +1,38 @@
+# Native packaging for Deephaven Jetty server
```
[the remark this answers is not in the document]
niloc132: Added notes on python dev setup, will punt on the rest for now

server/jetty/README.md
```diff
@@ -1,5 +1,44 @@
...
+    $ cd -
+   ```
+1. Build and install wheels for deephaven-jpy and deephaven:
+    ```shell
+    $ python -m pip install --upgrade pip # First upgrade pip
+    $ pip install wheel
+    $ export DEEPHAVEN_VERSION=0.8.0 # this should match the current version of your git repo
+
+    $ cd py/jpy
+    $ JAVA_HOME=/path/to/your/java/home # Customize this to fit your computer
```
devinrsmith, line 30: needs to be exported, or included as part of the setup.py command.

server/jetty/src/main/java/io/deephaven/server/jetty/CacheFilter.java
```diff
@@ -0,0 +1,36 @@
...
+        // set expiry to one year in the future
+        final Calendar calendar = Calendar.getInstance();
+        calendar.setTime(new Date());
+        calendar.add(Calendar.YEAR, 1);
+        httpResponse.setDateHeader("Expires", calendar.getTime().getTime());
+        // Note: immutable tells firefox to never revalidate as data will never change
+        httpResponse.setHeader("Cache-control", "max-age=" + YEAR_IN_SECONDS + ", public, immutable");
+        httpResponse.setHeader("Pragma", "");
+
+        filterChain.doFilter(request, response);
```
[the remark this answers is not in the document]
niloc132: Take a look at where we plug in the CacheFilter over in JettyBackedGrpcServer - those files in the static dir are all generated with their hash as part of their name, so that without a name change, the contents are (in theory) unchanged.

These filters and their configuration are the same as in DHE - to a fault, I didn't correct the path to be /ide/...

server/jetty/README.md
```diff
@@ -1,5 +1,44 @@
...
+there will be errors on startup indicating that the Python environment is not suitable.
+
+See https://github.com/deephaven/deephaven-core/issues/1657 for more discussion
+
+1. On MacOS there is an extra patch to apply at this time, to add an extra argument to clang,
+see the first code snippet of the comment at
+https://github.com/deephaven/deephaven-core/issues/1657#issuecomment-989040798 for specifics.
+1. Make sure Python is installed with a shared library. For example, if using `pyenv install`,
+be sure to first set `PYTHON_CONFIGURE_OPTS="--enabled-shared"`.
+1. Make a new directory for a virtual environment, and set it up:
```
devinrsmith, line 16: Many `1.`s.

server/jetty/README.md
```diff
@@ -1,5 +1,44 @@
...
 
+### Setting up a Python virtual environment
+
+This is an optional prerequisite, but lets the later commands to run the server work without
+specifying groovy as the console language. If you skip this step, be sure to use groovy, or
+there will be errors on startup indicating that the Python environment is not suitable.
+
+See https://github.com/deephaven/deephaven-core/issues/1657 for more discussion
+
+1. On MacOS there is an extra patch to apply at this time, to add an extra argument to clang,
```
devinrsmith, line 11: On Fedora, I also needed python development headers installed. That looked something like

```
sudo dnf install python-devel
```

for me. Likely want a note about headers, and how to get them for ubuntu, fedora, maybe others?

server/jetty/README.md
niloc132: This renders cleanly in markdown - using 1. each time ensures you don't rely on the text to be correct, can easily insert numbers, but the rendered html will be numbered correctly.

server/jetty/README.md
```diff
@@ -1,5 +1,44 @@
...
+   ```
+1. Build and install wheels for deephaven-jpy and deephaven:
+    ```shell
+    $ python -m pip install --upgrade pip # First upgrade pip
+    $ pip install wheel
+    $ export DEEPHAVEN_VERSION=0.8.0 # this should match the current version of your git repo
+
+    $ cd py/jpy
+    $ JAVA_HOME=/path/to/your/java/home # Customize this to fit your computer
+   $ python setup.py bdist_wheel
```
devinrsmith, line 31: Unfortunately, this is breaking for me (Fedora 35 running on python 3.10.1), so I'm not able to test locally.

```
src/main/c/jpy_jobj.c: In function ‘JType_InitSlots’:
src/main/c/jpy_jobj.c:714:24: error: lvalue required as left operand of assignment
  714 |     Py_REFCNT(typeObj) = 1;
```

Probably time to incorporate https://github.com/jpy-consortium/jpy/pull/14?

server/jetty/README.md
niloc132: Probably, but better put it #1657 if we require that, this isn't meant to be exhaustive at this time.

server/jetty/README.md
niloc132: We don't explicitly support any python version that we don't package at this time (and probably we should actually test the matrix of packages we ship before getting into testing cross OS, py vers, arch...)

devinrsmith: We only "package" python 3.7, right? In your example though, it looks like you are building w/ 3.9? 

It would be nice to get back into more native packaging / releasing of artifacts as wheels with each release; in which case we could be very explicit about what we expect should work.

niloc132: We only package 3.7 yes, and it happens to work with 3.9 too as far as I can tell. `pyenv` probably should be used for now to try out a different version.

devinrsmith (approved):
Looking good. We know about the follow up issues. I've done some light QA of w2w across a matrix of different server types.

PR #1737, opened 2021-12-23
Special cases that need no permutation.

cpwright:


replication/util/src/main/java/io/deephaven/replication/ReplicatePrimitiveCode.java
```diff
@@ -460,7 +460,9 @@ public static String replaceAll(String sourceClassJavaPath, Map<String, Long> se
         out.println(" * AUTO-GENERATED CLASS - DO NOT EDIT MANUALLY - for any changes edit "
                 + sourceClassName + " and regenerate");
         out.println(
-                " * ------------------------------------------------------------------------------------------------------------------ */");
+                " * ------------------------------------------------------------------------------------------------------------------");
```
rcaudy: Please add 3 more dashes. It's totally lame that the beginning and end lines aren't the same length:
```
/* ---------------------------------------------------------------------------------------------------------------------
 * AUTO-GENERATED CLASS - DO NOT EDIT MANUALLY - for any changes edit ParquetColumnRegionChar and regenerate
 * ------------------------------------------------------------------------------------------------------------------
 */
```

engine/table/src/test/java/io/deephaven/engine/table/impl/FuzzerTest.java
```diff
@@ -200,7 +200,7 @@ public void testLargeSetOfFuzzerQueriesRealtime() throws IOException, Interrupte
     public void testLargeSetOfFuzzerQueriesSimTime() throws IOException, InterruptedException {
         final long seed1 = DateTime.now().getNanos();
         for (long iteration = 0; iteration < 5; ++iteration) {
-            for (int segment = 0; segment < 10; segment++) {
+            for (int segment = 8; segment < 10; segment++) {
```
rcaudy, line 203: Is this intended for commit?

engine/table/src/test/java/io/deephaven/engine/table/impl/TestSumByProfile.java
```diff
@@ -25,22 +28,74 @@ public static void main(String[] args) {
...
-        System.out.println("Sum Elapsed Time: "
+        System.out.println("Operation Elapsed Time: "
                 + new DecimalFormat("###,###.000").format(((double) (endTimeSum - endTimeSelect) / 1_000_000_000.0)));
-        TableTools.show(sumBy);
+        TableTools.show(result);
     }
 
     private static void usage() {
-        System.err.println("TestSumByProfile select|view filename");
+        System.err.println("TestSumByProfile select|view sum|count filename");
```
rcaudy, line 98: Feel like a rename for this class is in order.
Unrelated: does the test suite try to run this? It's not really a unit test, but it follows the junit3 naming convention.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/ChunkedOperatorAggregationHelper.java
```diff
@@ -608,14 +608,14 @@ private void doRemovesForChunk(@NotNull final RowSequence keyIndicesToRemoveChun
...
-
+            final boolean permute = findSlotRuns(sortKernelContext, hashedRunContext, runStarts, runLengths,
+                    chunkPositions, slotsToRemoveFrom,
+                    ac.requiresRunFinds(SKIP_RUN_FIND));
 
             if (ac.requiresIndices()) {
-                if (findRuns) {
+                if (permute) {
                     final LongChunk<OrderedRowKeys> keyIndices = keyIndicesToRemoveChunk.asRowKeyChunk();
+                    permutedKeyIndices.setSize(keyIndices.size());
```
rcaudy, line 618: Was the absence of this `setSize` a bug, or are you just trying to be more hygienic?

engine/table/src/main/java/io/deephaven/engine/table/impl/by/ChunkedOperatorAggregationHelper.java
```diff
@@ -626,7 +626,7 @@ private void propagateRemovesToOperators(@NotNull final RowSequence keyIndicesTo
             boolean firstOperator = true;
             setFalse(modifiedSlots, runStarts.size());
 
-            final Chunk<? extends Values>[] valueChunks = findRuns ? workingChunks : new Chunk[ac.size()];
+            final Chunk<? extends Values>[] valueChunks = new Chunk[ac.size()];
```
rcaudy, line 629: Why change this? I don't see anything wrong with the previous, lower-allocation version. I do see that in other parts of the code you couldn't do this due to where the `findSlotRuns` call happens... are you just trying to be consistent everywhere?
Maybe would be better to just allocate this on the context?

engine/table/src/main/java/io/deephaven/engine/table/impl/by/ChunkedOperatorAggregationHelper.java
```diff
@@ -1619,12 +1654,13 @@ private static void initialBucketedKeyAddition(QueryTable withView,
...
                     final int inputSlot = ac.inputSlot(ii);
                     if (ii == inputSlot) {
-                        if (!findRuns) {
+                        if (!permute) {
                             valueChunks[inputSlot] = getChunk(ac.inputColumns[ii], getContexts[ii], chunkOk, usePrev);
                         } else {
-                            getAndPermuteChunk(ac.inputColumns[ii], getContexts[ii], chunkOk, usePrev,
-                                    permuteKernels[ii],
-                                    chunkPosition, workingChunks[ii]);
+                            assert workingChunks != null;
```
rcaudy, line 1660: Is this assertion intended for commit? Should be it be an `Assert.assertion` if so?

engine/table/src/main/java/io/deephaven/engine/table/impl/by/ChunkedOperatorAggregationHelper.java
cpwright: Permute input does not set the size of the output; so without this the permutedKeyIndices size is undefined.  I don't recall if it manifested itself anywhere.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/ChunkedOperatorAggregationHelper.java
cpwright: It makes IntelliJ happy that workingChunks[ii] isn't going to be an NPE to reduce a yellow warning.  Our own assertion doesn't do that, because IJ doesn't know about it.  I can take it out if you prefer.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/ChunkedOperatorAggregationHelper.java
cpwright: It was for consistency, but I moved it to the context.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/ChunkedOperatorAggregationHelper.java
```diff
@@ -1656,7 +1657,7 @@ private static void initialBucketedKeyAddition(QueryTable withView,
                     if (ii == inputSlot) {
                         if (!permute) {
                             valueChunks[inputSlot] = getChunk(ac.inputColumns[ii], getContexts[ii], chunkOk, usePrev);
-                        } else {
+                         } else {
```
rcaudy: Stray space?

rcaudy (approved):

PR #1738, opened 2021-12-23
Add ApplicationService with subscribeToFields to java-client-session

devinrsmith:


java-client/session-examples/build.gradle
```diff
@@ -10,6 +10,7 @@ dependencies {
 
     Classpaths.inheritGrpcPlatform(project, 'implementation')
     implementation 'io.grpc:grpc-core'
+    implementation 'io.grpc:grpc-netty'
```
niloc132, line 13: It seems odd that this wasn't required before now, in that it might be a hole in the java-client-session api to not convey this, or this might be a more in-depth example than normally goes in this project?

niloc132 (dismissed):
approved, with a thought to consider

java-client/session-examples/build.gradle
devinrsmith: Sorry, I need to change this to `runtimeOnly`.

devinrsmith: During your BOM refactoring, the session-examples broke - I haven't dug too deep, but this is the quick fix to get it back working.

niloc132 (approved):

PR #1746, opened 2021-12-28
Avoid removing session close callbacks if session already expired

### LINKED ISSUE #1651: SessionState ConcurrentModificationException
I'm seeing the following with somewhat regular consistency (by starting and stopping remote java client apps).

```
grpc-api_1    | 2021-12-02T17:27:52.917Z | EFAULT.refreshThread |  INFO | .e.u.UpdateGraphProcessor | Minimal Live Table Monitor cycle times: 25.18194ms / 34 cycles = 0.7406452941176471ms/cycle average
grpc-api_1    | 2021-12-02T17:27:54.416Z | heduler-Concurrent-3 |  INFO | i.d.g.s.SessionState      | SessionState{fcb6b9fb-1734-436d-8703-898c066267cf}: releasing outstanding exports
grpc-api_1    | 2021-12-02T17:27:54.417Z | heduler-Concurrent-3 |  INFO | i.d.g.s.SessionState      | SessionState{fcb6b9fb-1734-436d-8703-898c066267cf}: outstanding exports released
grpc-api_1    | 2021-12-02T17:27:54.418Z | heduler-Concurrent-3 |  INFO | portedTableUpdateListener | ExportedTableUpdateListener(4bbcb167) is complete
grpc-api_1    | Dec 02, 2021 5:27:54 PM io.grpc.internal.ServerCallImpl internalClose
grpc-api_1    | WARNING: Cancelling the stream with status Status{code=INTERNAL, description=Completed without a response, cause=null}
grpc-api_1    | Initiating shutdown due to: Exception while processing concurrentExecutor task
grpc-api_1    | java.util.ConcurrentModificationException
grpc-api_1    |         at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1043)
grpc-api_1    |         at java.base/java.util.ArrayList$Itr.next(ArrayList.java:997)
grpc-api_1    |         at io.deephaven.util.datastructures.SimpleReferenceManager.forEach(SimpleReferenceManager.java:117)
grpc-api_1    |         at io.deephaven.grpc_api.session.SessionState.onExpired(SessionState.java:426)
grpc-api_1    |         at io.deephaven.grpc_api.session.SessionService$SessionCleanupJob.run(SessionService.java:309)
grpc-api_1    |         at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
grpc-api_1    |         at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
grpc-api_1    |         at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
grpc-api_1    |         at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
grpc-api_1    |         at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
grpc-api_1    |         at io.deephaven.grpc_api.runner.DeephavenApiServerModule$ThreadFactory.lambda$newThread$0(DeephavenApiServerModule.java:156)
grpc-api_1    |         at java.base/java.lang.Thread.run(Thread.java:829)
```

niloc132:
This should eliminate a ConcurrentModificationException for some clients
that have enough objects in their sessions in the right order.

Also added a bit more documentation to clearly acknowledge that while
Closeable.close() is expected to be idempotent, the SessionState will
not call it more than once, and provides removeOnCloseCallback to help
downstream code make its own guarantees about this.

Fixes #1651

### DIFF
--- a/server/src/main/java/io/deephaven/server/session/SessionState.java
+++ b/server/src/main/java/io/deephaven/server/session/SessionState.java
@@ -367,13 +367,21 @@ public class SessionState {
     }
 
     /**
-     * Attach an on-close callback bound to the life of the session.
+     * Attach an on-close callback bound to the life of the session. Note that {@link Closeable} does not require that
+     * the close() method be idempotent, but when combined with {@link #removeOnCloseCallback(Closeable)}, close() will
+     * only be called once from this class.
+     * <p>
+     * </p>
+     * If called after the session has expired, this will throw, and the close() method on the provided instance will
+     * not be called.
      *
      * @param onClose the callback to invoke at end-of-life
      */
     public void addOnCloseCallback(final Closeable onClose) {
         synchronized (onCloseCallbacks) {
             if (isExpired()) {
+                // After the session has expired, nothing new can be added to the collection, so throw an exception (and
+                // release the lock, allowing each item already in the collection to be released)
                 throw GrpcUtil.statusRuntimeException(Code.UNAUTHENTICATED, "session has expired");
             }
             onCloseCallbacks.add(onClose);
@@ -382,12 +390,23 @@ public class SessionState {
 
     /**
      * Remove an on-close callback bound to the life of the session.
+     * <p />
+     * A common pattern to use this will be for an object to try to remove itself, and if it succeeds, to call its own
+     * {@link Closeable#close()}. If it fails, it can expect to have close() be called automatically.
      *
      * @param onClose the callback to no longer invoke at end-of-life
      * @return true iff the callback was removed
+     * @apiNote If this SessionState has already begun expiration processing, {@code onClose} will not be removed by
+     *          this method. This means that if {@code onClose} was previously added and not removed, it either has
+     *          already been invoked or will be invoked by the SessionState.
      */
     public boolean removeOnCloseCallback(final Closeable onClose) {
         synchronized (onCloseCallbacks) {
+            if (isExpired()) {
+                // After the session has expired, nothing can be removed from the collection. We return false here,
+                // preventing a closeable from trying to remove itself during its own close()
+                return false;
+            }
             return onCloseCallbacks.remove(onClose) != null;
         }
     }


rcaudy (approved):

grpc-api/src/main/java/io/deephaven/grpc_api/session/SessionState.java
```diff
@@ -383,12 +385,21 @@ public void addOnCloseCallback(final Closeable onClose) {
...
+     * @apiNote If this SessionState has already begun expiration processing, {@code onClose} will not be removed by
+     *          this method. This means that if {@code onClose} was previously added and not removed, it either has
+     *          already been invoked or will be invoked by the SessionState.
      */
     public boolean removeOnCloseCallback(final Closeable onClose) {
         synchronized (onCloseCallbacks) {
+            if (isExpired()) {
+                return false;
+            }
             return onCloseCallbacks.remove(onClose) != null;
```
devinrsmith, line 403: So, this isn't really a guard for against multiple threads trying to remove the closeable (and invoke the close themselves) - this is a guard against *re-entrancy* (the synchronize already provides multi-threaded safety support). I feel like we are missing some context if we don't add a comment a "re-entrancy" comment above isExpired().

I'm happy to approve this, pending above; but I think there is room for a follow up where SimpleReferenceManager interface could be improved - at least with how we are trying to use it from SessionState:

```java
        synchronized (onCloseCallbacks) {
            onCloseCallbacks.forEach((ref, callback) -> {
                ...
                callback.close();
                ...
            });
            onCloseCallbacks.clear();
        }
```

The above is clumsy IMO, and if we could remove the callback before calling close (instead of afterwards during the clear) there wouldn't be any need for a re-entrancy check. I'm happy to file a ticket and tackle it later.

niloc132: Well, we already had a guard against different threads removing the closable, the synchronized block. This is specifically a guard against something removing an item while the same thread is already in the process of _iterating_ the list (at which point it will clear the list). Reentrancy is not itself a problem, except because we're storing "code" instead of objects - it is sort of a weird use of this to invoke arbitrary callbacks from inside the lock.

I think @rcaudy agrees that SimpleReferenceManager could do with some improvement.

devinrsmith (approved):

PR #1749, opened 2021-12-28
Port Parallel Select Implementation from Deephaven Enterprise.

cpwright:


engine/table/src/main/java/io/deephaven/engine/table/impl/perf/UpdatePerformanceTracker.java
```diff
@@ -420,31 +365,23 @@ public String toString() {
 
         @Override
         public LogOutput append(final LogOutput logOutput) {
-            return logOutput.append("Entry{").append(", id=").append(id)
+            final LogOutput beginning = logOutput.append("Entry{").append(", id=").append(id)
                     .append(", evaluationNumber=").append(evaluationNumber)
                     .append(", operationNumber=").append(operationNumber)
                     .append(", description='").append(description).append('\'')
                     .append(", callerLine='").append(callerLine).append('\'')
-                    .append(", intervalUsageNanos=").append(intervalUsageNanos)
```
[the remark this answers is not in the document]
cpwright: Bad merge.

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/UpdatePerformanceTracker.java
```diff
@@ -420,31 +365,23 @@ public String toString() {
...
-                    .append(", startTimeNanos=").append(startTimeNanos)
-                    .append(", startAllocatedBytes=").append(startAllocatedBytes)
-                    .append(", startPoolAllocatedBytes=").append(startPoolAllocatedBytes)
                     .append(", maxTotalMemory=").append(maxTotalMemory)
                     .append(", minFreeMemory=").append(minFreeMemory)
                     .append(", collections=").append(collections)
                     .append(", collectionTimeNanos=").append(DateTimeUtils.millisToNanos(collectionTimeMs))
                     .append('}');
+            return appendStart(beginning).append(", totalMemory=").append(endSample.totalMemory)
+                    .append(", totalFreeMemory=").append(endSample.freeMemory).append('}');
```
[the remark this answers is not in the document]
cpwright: Why would you not want them, they are useful information about how that entry impacted memory usage;  stored and logged in the UPL but not logged in this output.  The counter point is why do we bother with the start fields at all, if we want to trim it.

engine/table/src/main/java/io/deephaven/engine/table/impl/SelectOrUpdateListener.java
```diff
@@ -0,0 +1,138 @@
...
+        final ModifiedColumnSet[] mcss = new ModifiedColumnSet[effects.size()];
+        int nextIndex = 0;
+        for (Map.Entry<String, String[]> entry : effects.entrySet()) {
+            parentNames[nextIndex] = entry.getKey();
+            mcss[nextIndex] = dependent.newModifiedColumnSet(entry.getValue());
+            ++nextIndex;
+        }
+        transformer = parent.newModifiedColumnSetTransformer(parentNames, mcss);
+        this.analyzer = analyzer;
+        this.enableParallelUpdate =
```
rcaudy, line 58: 1. Missing force
2. Why not just pass this in, since the calling code already knows the right answer? Oh, you answered that. Fine. Just fine.

engine/table/src/main/java/io/deephaven/engine/table/impl/SelectOrUpdateListener.java
```diff
@@ -0,0 +1,138 @@
...
+ * columns, then creates a column set transformer which will be used by onUpdate to transform updates.
+ */
+class SelectOrUpdateListener extends BaseTable.ListenerImpl {
+    private final QueryTable dependent;
+    private final TrackingRowSet resultIndex;
+    private final ModifiedColumnSet.Transformer transformer;
+    private final SelectAndViewAnalyzer analyzer;
+
+    private long initialNotificationStep = NotificationStepReceiver.NULL_NOTIFICATION_STEP;
+    private long finalNotificationStep = NotificationStepReceiver.NULL_NOTIFICATION_STEP;
```
rcaudy, line 30: This is more complicated than it needs to be. Use a `boolean updateInProgress` or similar. Set it to one value when work begins, the other when work ends, check for the idle value.

engine/table/src/main/java/io/deephaven/engine/table/impl/SelectOrUpdateListener.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    private void handleException(Exception e) {
+        dependent.notifyListenersOnError(e, getEntry());
+        try {
+            if (SystemicObjectTracker.isSystemic(dependent)) {
+                AsyncClientErrorNotifier.reportError(e);
+            }
+        } catch (IOException ioe) {
+            throw new RuntimeException("Exception in " + getEntry().toString(), e);
```
rcaudy, line 107: Some flavor or `UncheckedDeephavenException`?

engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/SelectAndViewAnalyzer.java
```diff
@@ -353,16 +353,22 @@ public SelectLayerCompletionHandler(BitSet completedColumns, BitSet requiredColu
...
-                completedColumns.set(completedColumn);
-                allRequiredColumnsSet = !fired && requiredColumns.stream().allMatch(completedColumns::get);
-                if (allRequiredColumnsSet) {
-                    fired = true;
+            if (!fired) {
+                final boolean readyToFire;
+                synchronized (completedColumns) {
+                    completedColumns.set(completedColumn);
+                    if (requiredColumns.get(completedColumn)) {
+                        readyToFire = !fired && requiredColumns.stream().allMatch(completedColumns::get);
```
rcaudy, line 361: We already know `!fired &&`.

rcaudy (approved):

PR #1752, opened 2021-12-29
Unify prop files for builds and production

### LINKED ISSUE #1732: dh-defaults.prop either needs to be replaced by grpc-api.prop or cleaned up for DHC
dh-defaults.prop contains a lot of properties that don't apply to DHC and is used in certain projects and tests. grpc-api.prop is much cleaner and more relevant.

niloc132:
Fixes #1732


configs/dh-defaults.prop
```diff
@@ -1,40 +1,9 @@
...
-allocation.stats.enabled=false
-measurement.comm_layer_timing=false
-measurement.partitioned_distributed_hashtables=false
+CompilerTools.logEnabledDefault=false
+UpdatePerformanceTracker.reportingMode=LISTENER_ONLY
+UpdatePerformanceTracker.reportIntervalMillis=60000
 measurement.per_thread_cpu=false
-measurement.log_and_reset=false
+allocation.stats.enabled=true
+statsdriver.enabled=true
```
devinrsmith, line 6: GH diff not very helpful here. For reference, both previously `false`.

configs/dh-defaults.prop
```diff
@@ -45,123 +14,41 @@ measurement.log_and_reset=false
...
-
-WorkerPythonEnvironment.defaultScriptPath=<devroot>
-
-########## Deephaven UpdatePerformanceTracker Properties ##########
-UpdatePerformanceTracker.reportingMode=LISTENER_ONLY
-UpdatePerformanceTracker.reportIntervalMillis=60000
-
-########## Deephaven QueryPerformanceRecorder Properties ##########
-QueryPerformanceRecorder.packageFilter.internal=defaultPackageFilters.qpr
+StringUtils.cacheSize=0
```
devinrsmith, line 25: Previously 65536

configs/dh-defaults.prop
```diff
@@ -45,123 +14,41 @@ measurement.log_and_reset=false
...
 
-########## Deephaven Local Data Refresh Properties ##########
-TableDataRefreshService.tableLocationsRefreshMillis=10000
-TableDataRefreshService.tableSizeRefreshMillis=1000
+TrackedFileHandleFactory.maxOpenFiles=1024
```
devinrsmith, line 27: Previously 4096

configs/dh-defaults.prop
```diff
@@ -45,123 +14,41 @@ measurement.log_and_reset=false
 # TODO: all of the created threads are waiting on RPCs or locks.
 #
 NIO.driver.workTimeout=100
-NIO.driver.initialThreadCount=2
-NIO.driver.maxThreadCount=400
+NIO.driver.initialThreadCount=4
+NIO.driver.maxThreadCount=16
```
devinrsmith, line 18: Remind me, what happens at exhaustion? Do we throw exception, or schedule for next available thread.

devinrsmith (commented):
Happy with structural changes. Happy to change defaults, as long as we are aware of what's changing.

niloc132:
So, the important thing here is that dh-defaults.prop was only used for some tests and some codegen, but was never used at runtime by our running containers. 

I have re-run all generators and there is no changes as a result of this to our .java or .py files.

The py tests were inconsistent before and some had broken paths, which is what prompted this change to begin with.

My assumption is that since workers have always* used the grpc-api.prop file's contents and we've never used dh-defaults.prop in any engine unit test or running container, that we would prefer the contents of grpc-api.prop, though we like the name of dh-defaults.prop better.

* since we ran a grpc server at all

devinrsmith (approved):
Ah - I missed the part where you were taking the existing values from grpc-api.prop.

I think our defaults are due for an audit, but no reason to hold up this good structural change.

configs/dh-defaults.prop
rcaudy: We shutdown the process.

rcaudy: That said, we shouldn't be using this thread pool much anymore.

configs/dh-defaults.prop
```diff
@@ -45,123 +14,41 @@ measurement.log_and_reset=false
...
 NIO.driver.useFast=true
 
-#
-# NIO wire-lag clock
-#
-NIO.wireLagClock.native=false
```
rcaudy, line 55: Still used.

configs/dh-defaults.prop
```diff
@@ -45,123 +14,41 @@ measurement.log_and_reset=false
...
-#Webapi.plugins=/full/path/plugins
-# A comma separate list of names for an App level JavaScript Plugins
-#Webapi.app.plugins=AppPluginName1,AppPluginName2
-# Set to true to display system badge
-#Webapi.user.defaults.systemBadge=false
-
-OneClick.allowedPattern.USym=^[A-Za-z0-9/.]+$
-OneClick.enforceUpperCaseColumns=USym
-
-UpdateGraphProcessor.checkTableOperations=true
```
rcaudy, line 132: Please put this back.

configs/dh-defaults.prop
```diff
@@ -45,123 +14,41 @@ measurement.log_and_reset=false
...
-QueryPerformanceRecorder.packageFilter.internal=defaultPackageFilters.qpr
+StringUtils.cacheSize=0
 
-########## Deephaven Local Data Refresh Properties ##########
-TableDataRefreshService.tableLocationsRefreshMillis=10000
-TableDataRefreshService.tableSizeRefreshMillis=1000
+TrackedFileHandleFactory.maxOpenFiles=1024
 
-# Defines the directory for pidfiles
-pidFileDirectory=/etc/deephaven/run
```
rcaudy, line 92: Can you restore this, but as `/var/run/deephaven` for now? If someone tries to use PidFileUtil it will blow up.

configs/dh-defaults.prop
```diff
@@ -45,123 +14,41 @@ measurement.log_and_reset=false
...
+Calendar.default=USNYSE
+Calendar.importPath=/default_calendar_imports.txt
 
-default.processEnvironmentFactory=io.deephaven.util.process.DefaultProcessEnvironment$Factory
+Plot.theme.default=LIGHT
+Plot.theme.internalPath=<devroot>/configs/chartthemes
 
 columnsFile=DeephavenColumns.xml
```
rcaudy, line 35: Delete this, unused now.

configs/dh-defaults.prop
```diff
@@ -45,123 +14,41 @@ measurement.log_and_reset=false
...
-#Webapi.app.plugins=AppPluginName1,AppPluginName2
-# Set to true to display system badge
-#Webapi.user.defaults.systemBadge=false
-
-OneClick.allowedPattern.USym=^[A-Za-z0-9/.]+$
-OneClick.enforceUpperCaseColumns=USym
-
-UpdateGraphProcessor.checkTableOperations=true
-
-logroot=/var/log/deephaven
```
rcaudy, line 134: We should put these (`logroot` and `defaultLogDir`) back, unless we delete the accessor for it in Configuration.

configs/dh-defaults.prop
devinrsmith: I *think* Colin did a copy/paste and/or rename, but since there were file renames, the diff isn't very good.

rcaudy: I think he just moved `grpc-api.prop` to `dh-defaults.prop`. Either way, I would like `UpdateGraphProcessor.checkTableOperations=true` in our production applications.

configs/dh-defaults.prop
rcaudy: You know, I think I'm wrong about this. It would probably be better to blow up and make someone consciously choose a value.

configs/dh-defaults.prop
rcaudy: You know, I think I'm wrong about this. It would probably be better to blow up and make someone consciously choose a value.

niloc132: If the server doesn't use it though, should we reference it here? Especially if we are set up to run the server instances as plain user processes? We would otherwise need to also add logic to create the directory structure, and probably as root (and "not at all" on windows).

Relative dir to execution okay?

configs/dh-defaults.prop
niloc132: Are you okay with a relative path (same as pid)?

configs/dh-defaults.prop
niloc132: This was included in grpc-api.prop to begin with because the server wouldnt start without it. Startup today without that:

```
Initiating shutdown due to: Uncaught exception in thread UpdateGraphProcessor.DEFAULT.refreshThread
io.deephaven.configuration.PropertyException: Missing property in config file: NIO.driver.maxThreadCount
        at io.deephaven.configuration.PropertyFile.getProperty(PropertyFile.java:115)
        at io.deephaven.configuration.PropertyFile.getInteger(PropertyFile.java:126)
        at io.deephaven.net.impl.nio.NIODriver.init(NIODriver.java:106)
        at io.deephaven.net.impl.nio.NIODriver.init(NIODriver.java:96)
        at io.deephaven.net.CommBase.getScheduler(CommBase.java:68)
        at io.deephaven.engine.updategraph.UpdateGraphProcessor.refreshTablesAndFlushNotifications(UpdateGraphProcessor.java:1457)
        at io.deephaven.engine.updategraph.UpdateGraphProcessor$1.run(UpdateGraphProcessor.java:208)

```

rcaudy (approved):

PR #1765, opened 2021-12-29
Consolidate application, query scope, and raw ticket parameters

devinrsmith:
The java client examples CLI input for tickets is now re-usable across the different ticket types:

```
 ([--variable=<variable>] | [--app-id=<applicationId>
 --app-field=<fieldName>] | [--ticket=<ticket>])
```


java-client/example-utilities/src/main/java/io/deephaven/client/examples/RawTicket.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.nio.charset.StandardCharsets;
+
+public class RawTicket implements HasTicketId {
+    @Option(names = {"--ticket"}, required = true, description = "The raw ticket.")
+    String ticket;
+
+    @Override
+    public TicketId ticketId() {
+        return new TicketId(ticket.getBytes(StandardCharsets.UTF_8));
```
niloc132, line 15: to confirm, only bytes-as-utf8 from the command line is allowed to be passed in this way? if I had to guess, i'd suggest this should ask for ASCII instead, or use base64 to be clear that tickets are binary

java-client/example-utilities/src/main/java/io/deephaven/client/examples/Ticket.java
```diff
@@ -0,0 +1,34 @@
...
+    @ArgGroup(exclusive = false)
+    ScopeField scopeField;
+
+    @ArgGroup(exclusive = false)
+    ApplicationField applicationField;
+
+    @ArgGroup(exclusive = false)
+    RawTicket rawTicket;
+
+    private HasTicketId get() {
```
niloc132, line 17: personal preference, a private method that is only called once and only saves a tiny bit of repetition isn't usually worth it - this could be easily inlined with no effect on readability.

java-client/flight-examples/build.gradle
```diff
@@ -51,13 +50,9 @@ applicationDistribution.into('bin') {
     from(createApplication('add-to-input-table', 'io.deephaven.client.examples.AddToInputTable'))
     from(createApplication('kv-input-table', 'io.deephaven.client.examples.KeyValueInputTable'))
 
-    from(createApplication('get-ticket-table', 'io.deephaven.client.examples.GetTicketTable'))
+    from(createApplication('get-table', 'io.deephaven.client.examples.GetDirectTable'))
+    from(createApplication('get-schema', 'io.deephaven.client.examples.GetDirectSchema'))
```
niloc132, line 54: next chance you get, consider generalizing the CreateStartScripts so these mains run cleanly from IJ clicking on the green triangle next to the name - should be a matter of assigning the global application fields like classpath, jvm args, etc as the jetty/netty modules do, and have the scripts reference those rather than have each task have its own copy and none be global

niloc132 (commented):
Package/project names should be remade to be clear that these are cli tools (and as such, documentation written). 

If possible, attach a issue id?

devinrsmith:
In support of `fetch-object` part of https://github.com/deephaven/deephaven-core/issues/1766

java-client/flight-examples/build.gradle
devinrsmith: So, these are runnable running from IJ (maybe they don't inherit the same default jvm args as we specify in gradle, I'll grant that).

Clicking the green run for JettyMain - that doesn't inherit applicationDefaultJvmArgs, does it?

java-client/example-utilities/src/main/java/io/deephaven/client/examples/RawTicket.java
devinrsmith: I'm not sure the best way to do this - I'm not necessarily trying to support arbitrary bytes via CLI arguments, but just make it so that "tickets that look like strings" can be passed this way. The picocli library will have some way of creating `String ticket`, maybe with unicode support (maybe it depends on the terminal).

java-client/flight-examples/build.gradle
niloc132: ...you are right, that's irritating, I had understood that to use the application plugin configuration to exec, but instead it is some synthetic task, `':server-jetty:JettyMain.main()'`, which isn't in the actual gradle model (but IJ makes it work anyway...).

That task is apparently a JavaExec, so the java-open-nio applies, and the same trick can be used for consistent jvm args (like forcing groovy):

```
tasks.withType(JavaExec) {
    jvmArgs applicationDefaultJvmArgs
}
```
(this is in server/jetty/build.gradle, and just grabs the application args, etc).

java-client/example-utilities/src/main/java/io/deephaven/client/examples/RawTicket.java
niloc132: Probably should ask the terminal what charset it uses (mine isn't defaulting to utf8 on at least one box), or assume ascii so we can fail in the general case?

niloc132 (approved):

PR #1793, opened 2021-12-31
 Flat Array Sources for Static Select, Parquet Cleanups

cpwright:


engine/table/src/main/java/io/deephaven/engine/table/impl/util/FunctionGeneratedTableFactory.java
```diff
@@ -117,21 +111,22 @@ private long updateTable() {
...
-                long current = sourceIt.nextLong();
-                // noinspection unchecked
-                destColumn.copy((ColumnSource) entry.getValue(), current, position++);
-            }
+            sourceColumnsArray[cc] = entry.getValue();
+            destColumnsArray[cc++] = destColumn;
         }
+
+        // noinspection unchecked
+        ChunkUtils.copyData(sourceColumnsArray, sourceRowSet, destColumnsArray, RowSetFactory.flat(sourceRowSet.size()),
```
rcaudy, line 128: `RowSequence` instead of `RowSet`?

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/flat/FlatByteArraySource.java
```diff
@@ -0,0 +1,235 @@
...
+// endregion boxing imports
+
+/**
+ * Simple flat array source that supports fillFromChunk for initial creation.
+ *
+ * No previous value tracking is permitted, so this column source is only useful as a flat static source.
+ *
+ * A single array backs the result, so getChunk calls with contiguous ranges should always be able to return a
+ * reference to the backing store without an array copy.  The immediate consequence is that you may not create
+ * sources that have a capacity larger than Integer.MAX_VALUE.
```
rcaudy, line 32: "Max array size"

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/flat/FlatByteArraySource.java
```diff
@@ -0,0 +1,235 @@
...
+        return getUnsafe(index);
+    }
+
+    public final byte getUnsafe(long index) {
+        return data[(int)index];
+    }
+
+    @Override
+    public final void set(long key, byte value) {
+        data[Math.toIntExact(key)] = value;
```
rcaudy, line 70: Just cast it to an `int`.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/flat/FlatByteArraySource.java
```diff
@@ -0,0 +1,235 @@
...
+    public Chunk<? extends Values> getPrevChunk(@NotNull GetContext context, long firstKey, long lastKey) {
+        return getChunk(context, firstKey, lastKey);
+    }
+
+    @Override
+    public boolean providesFillUnordered() {
+        return true;
+    }
+
+    // region reinterpret
```
rcaudy, line 233: Need to reinterpret as `Boolean`.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/flat/FlatByteArraySource.java
```diff
@@ -0,0 +1,235 @@
...
+ * reference to the backing store without an array copy.  The immediate consequence is that you may not create
+ * sources that have a capacity larger than Integer.MAX_VALUE.
+ *
+ * If your size is greater than Integer.MAX_VALUE, prefer {@link Flat2DByteArraySource}.
+ */
+public class FlatByteArraySource extends AbstractDeferredGroupingColumnSource<Byte> implements ImmutableColumnSourceGetDefaults.ForByte, WritableColumnSource<Byte>, FillUnordered, InMemoryColumnSource, ChunkedBackingStoreExposedWritableSource {
+    private byte[] data;
+
+    // region constructor
+    public FlatByteArraySource(long size) {
```
rcaudy, line 40: Should we stop taking `size` here? Or save it and validate in `ensureCapacity`?

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/flat/Flat2DByteArraySource.java
```diff
@@ -0,0 +1,285 @@
...
+/**
+ * Simple almost flat array source that supports fillFromChunk for initial creation.
+ *
+ * No previous value tracking is permitted, so this column source is only useful as a flat static source.
+ *
+ * A two-dimension array single array backs the result, with by default segments of 2^30 elements.  This is so that
+ * getChunk calls with contiguous ranges are often able to return a reference to the backing store without an array
+ * copy.
+ *
+ * If your size is smaller than Integer.MAX_VALUE, prefer {@link FlatByteArraySource}.
```
rcaudy, line 33: "max array size"

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/flat/Flat2DByteArraySource.java
```diff
@@ -0,0 +1,285 @@
...
+    public GetContext makeGetContext(int chunkCapacity, SharedContext sharedContext) {
+        return new GetContextWithResettable(super.makeGetContext(chunkCapacity, sharedContext));
+    }
+
+    @Override
+    public Chunk<? extends Values> getChunk(@NotNull GetContext context, long firstKey, long lastKey) {
+        final GetContextWithResettable contextWithResettable = (GetContextWithResettable) context;
+        final int segment = keyToSegment(firstKey);
+        if (segment != keyToSegment(lastKey)) {
+            super.getChunk(contextWithResettable.inner, firstKey, lastKey);
```
rcaudy, line 202: Return...

engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/SelectAndViewAnalyzer.java
```diff
@@ -47,6 +54,10 @@ public static SelectAndViewAnalyzer create(Mode mode, Map<String, ColumnSource<?
             rowRedirection = null;
         }
 
+        boolean flattenedResult = !rowSet.isFlat() && allowFlattening && mode == Mode.SELECT_STATIC;
+        final boolean flatResult = rowSet.isFlat();
```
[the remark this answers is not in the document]
cpwright: As mentioned on the phone, we need to do the work to initialize the definition w/ the previous analyzers before we can say shouldPreserve; so this doesn't actually work.

Integrations/python/test/testTableTools.py
```diff
@@ -465,7 +465,9 @@ def testPrimitiveColSourceCases(self):
             # print("table from string varargs = \n{}".format(TableTools.html(TableTools.newTable(3, {'string2': col2}))))
 
         with self.subTest(msg="colSource with datetime"):
+            print("Trying date time")
```
rcaudy, line 468: Please delete these.

engine/table/src/test/java/io/deephaven/engine/table/impl/QueryTableSelectUpdateBigTest.java
```diff
@@ -0,0 +1,54 @@
...
+@Category(OutOfBandTest.class)
+public class QueryTableSelectUpdateBigTest {
+    @Rule
+    public final EngineCleanup base = new EngineCleanup();
+
+    @Test
+    public void test2DSelect() {
+        final Table input = emptyTable(3_000_000_000L).updateView("A=(byte)7").where("ii >= 10");
+        final Table selected = input.select();
+        assertTableEquals(input.tail(2_999_999_990L), selected);
```
rcaudy, line 52: Why do we need the `tail`?

cpwright: Notice the ii >= 10; the point was to force things to not quite line up and make sure that the new code for the iteration of the destination ordered keys inside the exposed backing store path is exercised.

rcaudy (approved):

Issue #1162, opened 2021-08-30
Adopt java-platform to manage versions


niloc132:
This will replace `classpaths.gradle` and `Classpaths.groovy`.

https://docs.gradle.org/current/userguide/java_platform_plugin.html

cc @JamesXNelson @nbauernfeind

devinrsmith:
It seems like the latest recommendation is to use a "catalog".

 https://docs.gradle.org/current/userguide/platforms.html#sub:platforms-vs-catalog

```
Because platforms and catalogs both talk about dependency versions and can both be used to share dependency versions in a project, there might be a confusion regarding what to use and if one is preferable to the other.

The answer is that you should always use catalogs and maybe use platforms in addition.
```

devinrsmith:
I think this can be closed out given https://github.com/deephaven/deephaven-core/pull/5683?

PR #1799, opened 2022-01-04
Use the same version of barrage in all parts of the project

niloc132:
This only affects the JS API - I manually tested to confirm that subscriptions still work correctly.

https://github.com/deephaven/deephaven-core/issues/1162 should make this mistake harder to happen in the future.


lbooker42 (approved):
This looks correct to me.

PR #1803, opened 2022-01-05
Add base+ images to repo and info to README

jjbrosnan:
Will require updates to the docs repo.


margaretkennedy (dismissed):

containers/python-examples/NLTK/docker-compose.yml
```diff
@@ -0,0 +1,48 @@
+version: "3.4"
+
+services:
+  grpc-api:
```
devinrsmith, line 4: We shouldn't be using `grpc-api` anymore. We should likely call this simply `server`. Ditto for all the other compose files.

containers/python-examples/NLTK/docker-compose.yml
```diff
@@ -0,0 +1,48 @@
...
+    image: ghcr.io/deephaven/nltk-base:${VERSION:-latest}
```
devinrsmith, line 5: This is not the correct image. It should be `server-nltk`. See packages available at https://github.com/orgs/deephaven/packages?repo_name=deephaven-core, this comment also applies to all the other compose files. Note: it looks like we need to tag the extended images with latest, filed an issue for myself https://github.com/deephaven/deephaven-core/issues/1808

devinrsmith (changes requested):

containers/python-examples/NLTK/docker-compose.yml
jjbrosnan: All naming has been updated.  Thanks for pointing that out

margaretkennedy (approved):

README.md
```diff
@@ -24,8 +24,8 @@ which functions as the data backbone for prominent hedge funds, banks, and finan
 
 | Language      | Server Application | Client Application (OpenAPI) |
```
devinrsmith, line 25: We probably shouldn't have any references to `OpenAPI` anywhere... we don't really call it that.

containers/python-examples/PyTorch/docker-compose.yml
```diff
@@ -0,0 +1,48 @@
...
+    image: ghcr.io/deephaven/envoy:${VERSION:-latest}
+    depends_on:
+      - web
+      - grpc-proxy
+      - grpc-api
+    ports:
+      - "${PORT:-10000}:10000"
+
+  examples:
+    image: ghcr.io/deephaven/examples
```
devinrsmith, line 41: We should be explicit and use the `:latest` tag, instead of leaving it unspecified. Ditto in the other compose files.

containers/python-examples/NLTK/docker-compose.yml
```diff
@@ -0,0 +1,48 @@
...
+    expose:
+      - '80'
+    volumes:
+      - ./data:/data
+      - web-tmp:/tmp
+
+  grpc-proxy:
+    image: ghcr.io/deephaven/grpc-proxy:${VERSION:-latest}
+    environment:
+      - BACKEND_ADDR=grpc-api:8080
```
devinrsmith, line 25: This won't work, we don't have `grpc-api` anymore. This makes me think that these compose files haven't been tested? All docker container references to `grpc-api` should be changed as appropriate to `server` where applicable. Same goes for other compose files.

containers/python/NLTK/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
+version: "3.4"
+
+services:
+  grpc-api:
+    image: ghcr.io/deephaven/nltk-base:${VERSION:-latest}
```
devinrsmith, line 5: This is not the correct image, which makes me think this compose file hasn't been tested.

containers/python/NLTK/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
+version: "3.4"
+
+services:
+  grpc-api:
```
devinrsmith, line 4: We shouldn't be calling the service `grpc-api` anymore.

containers/python/PyTorch/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
+version: "3.4"
+
+services:
+  grpc-api:
+    image: ghcr.io/deephaven/pytorch-base:${VERSION:-latest}
```
devinrsmith, line 5: Wrong image, old service name.

containers/python/TensorFlow/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
+version: "3.4"
+
+services:
+  grpc-api:
+    image: ghcr.io/deephaven/tensorflow-base:${VERSION:-latest}
```
devinrsmith, line 5: Wrong image, old service name.

devinrsmith (changes requested):
I think we need to, at a very minimum, have a testing script we run through with each of the compose compositions to make sure it all works. As it stands, I think a large number of them will not work.

In addition, I think we should try and work towards a new strategy instead of having the matrix of python packages and examples/no-examples. We can work to parameterize these compose definitions in the future (using environment variables, or .env files), and it may reduce complexity.

margaretkennedy (approved):

jjbrosnan:
I have tested every docker-compose file in this PR.  They all work.  There is one thing to note:

The two NLTK images cause warnings in the web console when starting up Deephaven because a couple of other Python packages aren't installed: tkinter and one other.

devinrsmith (approved):

PR #1805, opened 2022-01-05
Mark grpc dependencies as api

### LINKED ISSUE #1804: Proper API dependencies for consumers of deephaven-java-client-session
Per @jcferretti

> I updated to 0.8.0 (with the new name), but now I am missing grpc.  I am guessing the old version was pulling grpc in, and this one doesn't. 

Consumers of deephaven-java-client-session should inherit grpc-api.

devinrsmith:
Fixes #1804

### DIFF
--- a/java-client/example-utilities/build.gradle
+++ b/java-client/example-utilities/build.gradle
@@ -5,7 +5,4 @@ plugins {
 dependencies {
     api 'info.picocli:picocli:4.6.1'
     api project(':java-client-session')
-
-    Classpaths.inheritGrpcPlatform(project, 'api')
-    api 'io.grpc:grpc-core'
 }
--- a/java-client/example-utilities/build.gradle
+++ b/java-client/example-utilities/build.gradle
@@ -5,4 +5,5 @@ plugins {
 dependencies {
     api 'info.picocli:picocli:4.6.1'
     api project(':java-client-session')
+    api 'io.grpc:grpc-api'
 }
--- a/java-client/session-dagger/build.gradle
+++ b/java-client/session-dagger/build.gradle
@@ -6,6 +6,8 @@ description = 'The Deephaven client session dagger library'
 
 dependencies {
     api project(':java-client-session')
+    api 'io.grpc:grpc-api'
+    implementation project(':proto:proto-backplane-grpc')
 
     Classpaths.inheritDagger(project)
 
--- a/java-client/session-dagger/build.gradle
+++ b/java-client/session-dagger/build.gradle
@@ -7,9 +7,6 @@ description = 'The Deephaven client session dagger library'
 dependencies {
     api project(':java-client-session')
 
-    Classpaths.inheritGrpcPlatform(project, 'implementation')
-    implementation 'io.grpc:grpc-api'
-
     Classpaths.inheritDagger(project)
 
     Classpaths.inheritSlf4j(project, 'slf4j-api', 'implementation')
--- a/java-client/session/build.gradle
+++ b/java-client/session/build.gradle
@@ -6,8 +6,10 @@ description = 'The Deephaven client session library'
 
 dependencies {
   api project(':qst')
-  api project(':proto:proto-backplane-grpc')
   api project(':uri')
+  api 'io.grpc:grpc-api'
+  // TODO(deephaven-core#1783): Make proto-backplane-grpc implementation dependency
+  api project(':proto:proto-backplane-grpc')
 
   Classpaths.inheritImmutables(project)
 
--- a/java-client/session/build.gradle
+++ b/java-client/session/build.gradle
@@ -9,9 +9,6 @@ dependencies {
   api project(':proto:proto-backplane-grpc')
   api project(':uri')
 
-  Classpaths.inheritGrpcPlatform(project, 'implementation')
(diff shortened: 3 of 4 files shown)


java-client/session-dagger/build.gradle
```diff
@@ -7,9 +7,6 @@ description = 'The Deephaven client session dagger library'
 dependencies {
     api project(':java-client-session')
 
-    Classpaths.inheritGrpcPlatform(project, 'implementation')
-    implementation 'io.grpc:grpc-api'
```
niloc132, line 11: This should still be here, as a matter of hygiene, no? if proto-backplane-grpc were to change its return type to "Object" for some method that currently returns a grpc-api type, would this project still compile?

Ditto for other projects.

java-client/example-utilities/build.gradle
```diff
@@ -5,7 +5,4 @@ plugins {
 dependencies {
     api 'info.picocli:picocli:4.6.1'
     api project(':java-client-session')
-
-    Classpaths.inheritGrpcPlatform(project, 'api')
-    api 'io.grpc:grpc-core'
```
niloc132, line 10: as an API dependency, this should still be here because _this_ project includes grpc-core in its API - if upstream changed to be implementation, that would break this project's _downstream consumers_, we might not even notice here...

java-client/session-dagger/build.gradle
devinrsmith: I'm not sure where to draw the line. Should we lean on transitive api dependencies, or do we want to be as specific as possible (it may be the case when the upstream implementation changes, we want the downstream consumers to break?).

For example, session-dagger also has `proto-backplane-grpc` as part of it's public interface (by way of DeephavenChannel) - should we declare:

```gradle
    api project(':java-client-session')
    api project(':proto:proto-backplane-grpc')
    api 'io.grpc:grpc-api'
```

Or maybe, we only consider Module return types as part of the "public" interface, and the input values can be left out, in which case:

```gradle
    api project(':java-client-session')
    api 'io.grpc:grpc-api'
```
is ok?

niloc132: The "obviously correct"/"pedantically conservative" approach would be to re-declare anything you actually use (implementation) or expose (api) by your classes, the "brittle yet works (until it doesn't)" is to rely on transitive dependencies. I don't really either.

I also wish that gradle would be a little more opinionated about this, and voice preferences when we bend the rules, so we can't actually break it (I can see a point for not actually generating it for us, so that we have to deliberately think about it).

With all that in mind, session-dagger uses DeephavenChannel from proto-backplane-grpc and exposes it as part of the api, so it should depend on it as an api dependency. This doesn't obviously mean to me that grpc-api should also be an api dependency _unless_ session-dagger itself exposes any of its types (and it doesn't need to be an implementation dependency, unless session-dagger uses a grpc-api class). That is, under the maximally conservative approach, ProjectA  depends on LibraryB and will re-depend on some of LibraryB's API dependencies if it actually uses that part of the API. Any API dependencies of LibraryB will only be explicitly re-depended on as `implementation` if they are used, and only explicitly re-depended on as `api` if ProjectA explicitly exposes them. It is _not_ necessary to re-expose some transitive dependency as `api` if it is only exposed by LibraryB's own types, whcih ProjectA exposes:

```
ProjectA has ClassA
  ClassA.foo() returns ClassB 
    --> LibraryB is an api dependency
  ClassA.bar() returns ClassC (from a ClassB instance)
    --> TransitiveApiC is an api dependency, re-dependended
  ClassA.baz() returns void, but uses ClassB.blah() which returns ClassD)
    --> TransitiveApiD is implementation dependency, re-depended
  (note that TransitiveApiE/TransitiveApiF is not re-depended at all)

LibraryB has ClassB
  ClassB.junk() returns ClassC
    --> TransitiveApiC is an api dependency
  ClassB.blah() returns ClassD
    --> TransitiveApiD is an api dependency
  ClassB.other() returns ClassE
    --> TransitiveApiE is an api dependency
  ClassB.last() returns void, but uses ClassF to do it
    --> TransitiveImplF is an implementation dependency

TransitiveApiC has ClassC

TransitiveApiD has ClassD

TransitiveApiE has ClassE

TransitiveImplF has ClassF # never "re-depended on"
```

--

I think I agree that only subclasses and return types cause a dependency to be marked as `api`, internal usage and params seem clear to be implementation.

I'll poke around for any other reading materials on this...

devinrsmith: It also gets a bit murkier with the submodules - for example, the user-written flight-dagger sources don't have a dependency on DeephavenChannel, but if you look at the generated DaggerDeephavenFlightRoot, it *does* (inherited through the subcomponent modules). I'm assuming we *don't* want to explicitly declare an impl (nor api) dependency on proto-backplane-grpc from flight-dagger, as I expect this is something we *want* to be transitively provided via session-dagger?

niloc132: As I understand it, if components are kept in a project and only concerned with that project, the only dependencies that dagger requires are to resolve sub/child components - nested components are only checked to ensure they aren't cyclical (and this is optional), they are not required to be able to resolve all of their referenced types.

niloc132 (approved):

PR #1811, opened 2022-01-07
Add commented out debugging settings to docker-compose-common, to ease out turning debugging on.

jcferretti:


devinrsmith (commented):
I don't think we can remove -Xmx from our dev configuration and have it just work across all our dev environments, there are some caveats that may depend on docker engine version and OS cgroup version.

For production configurations in environments that are known in advance, I think this is a good change.

Chatting offline.

docker-compose-common.yml
```diff
@@ -26,6 +28,7 @@ services:
       resources:
         limits:
           cpus: '4.0'
+          # The JVM memory is configured from gradle to use *RAMPercentage options.
```
devinrsmith: This comment doesn't make sense here anymore

devinrsmith (approved):

devinrsmith (approved):

PR #1812, opened 2022-01-07
hand-wrapped the datetimeutils module, fixes #1640

### LINKED ISSUE #1640: Hand craft  a DateTimeUtils wrapper
The new wrapper needs to:
1. make the names of the static methods PEP8 compliant
2. may need to move some methods to other modules (dtypes for example) to be more cohesive.

related to #1263

jmao-denver:
1. refactored the dtypes module into a package
2. created DateTime and Period wrapper classes under dtypes
3. created datetimeutils.py
4. docstrings are added in datetimeutils.py
5. unit tests in test_datetimeutils.py
6. minor changes here and there to tide other modules up


pyintegration/tests/test_datetimeutils.py
```diff
@@ -0,0 +1,228 @@
...
+from deephaven2.constants import NULL_LONG
+from deephaven2.datetimeutils import *
+
+
+class DateTimeUtilsTestCase(unittest.TestCase):
+    def test_convert_datetime(self):
+        datetime_str = "2021-12-10T23:59:59"
+        timezone_str = "NY"
+        dt = convert_datetime(f"{datetime_str} {timezone_str}")
+        print(dt)
```
devinrsmith, line 17: Print necessary?

pyintegration/deephaven2/dtypes/_datetime.py
```diff
@@ -0,0 +1,108 @@
...
+        if isinstance(v, _DHDateTime.j_type):
+            self.j_datetime = v
+        else:
+            # TODO conversion support from Python built-in datetime, string datetime representation etc.
+            self.j_datetime = _DHDateTime(v)
+
+    def __eq__(self, other):
+        if other is None:
+            if self.j_datetime is None:
+                return True
```
devinrsmith, line 28: This is confusing to me - we want `DateTime(None) == None`?

devinrsmith: Can / should `j_datetime` even be allowed to be `None`?

Larger question in general - should we allow wrapped `None` for any types? It means we have to handle:

`None`, `Wrapped(None)`, and `Wrapped(<real>)`  everywhere instead of
`None` and `Wrapped(<real>)`

pyintegration/deephaven2/constants.py
```diff
@@ -92,3 +93,56 @@ class SortDirection(Enum):
 MIN_POS_DOUBLE = _JQueryConstants.MIN_POS_DOUBLE
 """ Minimum positive value of type double. """
 
+
+class TimeZone(Enum):
```
rcaudy, line 97: Should we have a way to define arbitrary time zones? Maybe needs an answer in Java, first.

pyintegration/deephaven2/dtypes/_datetime.py
jmao-denver: Good catch. I actually went back and forth on this a bit and decided against allowing j_datetime to be None. A null Java object is always translated to None in Python by JPY. I modified the constructor but overlooked __eq__.

pyintegration/tests/test_datetimeutils.py
jmao-denver: will remove.

pyintegration/deephaven2/constants.py
jmao-denver: definitely can wait until requested by the users

pyintegration/tests/test_datetimeutils.py
```diff
@@ -0,0 +1,227 @@
...
+        period_str = "T1M"
+        period = convert_period(period_str)
+        self.assertEqual(repr(period).upper(), period_str)
+
+        with self.assertRaises(DHError) as cm:
+            period_str = "T1Y"
+            period = convert_period(period_str)
+        self.assertIn("RuntimeException", str(cm.exception))
+
+    def test_convert_time(self):
```
chipkent, line 39: looks like this is the function that creates nanos from a string.  Does it do anything else?

pyintegration/tests/test_datetimeutils.py
```diff
@@ -0,0 +1,227 @@
...
+        dt = current_time()
+        self.assertIn(day_of_week(dt, TimeZone.MT), range(1, 8))
+
+    def test_day_of_year(self):
+        dt = current_time()
+        self.assertIn(day_of_year(dt, TimeZone.MT), range(1, 366))
+
+    def test_format_date(self):
+        dt = current_time()
+        self.assertIn(TimeZone.SYD.name, format_datetime(dt, TimeZone.SYD))
```
chipkent, line 79: test is named `test_format_date`, but the tested method is `format_datetime`.

pyintegration/tests/test_datetimeutils.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def test_format_nanos(self):
+        dt = current_time()
+        ns = nanos(dt)
+        ns_str1 = format_nanos(ns).split(".")[-1]
+        ns_str2 = format_datetime(dt, TimeZone.UTC).split(".")[-1]
+        self.assertTrue(ns_str2.startswith(ns_str1))
```
chipkent, line 86: Shouldn't this be equal?

pyintegration/deephaven2/datetimeutils.py
```diff
@@ -0,0 +1,803 @@
+#
+#   Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
```
chipkent, line 2: Below, I made some comments about possibly doing a default `TimeZone`.  I'm not sure that's a good idea.  Being explicit has value.

chipkent (changes requested):

pyintegration/deephaven2/dtypes/_datetime.py
```diff
@@ -0,0 +1,107 @@
...
+        new_seq = []
+        for v in seq:
+            if v is None:
+                new_seq.append(None)
+            elif isinstance(v, DateTime):
+                new_seq.append(v.j_datetime)
+            elif isinstance(v, _DHDateTime.j_type):
+                new_seq.append(v)
+            else:
+                raise DHError(message="Not a valid datetime")
```
devinrsmith, line 83: I wonder if it makes sense to have this as a top level utility method - in fact, it's probably a pattern we want to establish for all handwrapped types?

pyintegration/deephaven2/dtypes/_datetime.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __init__(self, v: Any):
+        if isinstance(v, _DHPeriod.j_type):
+            self.j_period = v
+        else:
+            # TODO? conversion support from Python timedelta, etc.
+            self.j_period = _DHPeriod(v)
+
+    def __repr__(self):
+        return str(self.j_period)
```
devinrsmith, line 102: If we setup a HandWrappedABC class, we might be able to get a lot of this boilerplate for free, and have common access point for `self.j_object`.

pyintegration/deephaven2/column.py
```diff
@@ -33,8 +32,8 @@ def __repr__(self):
 class Column:
     """ A Column object represents a column definition in a Deephaven Table. """
     name: str
-    data_type: DType
-    component_type: DType = None
+    data_type: Any
+    component_type: Any = None
```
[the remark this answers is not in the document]
jmao-denver: Here is the problem:
The majority of the types defined in dtypes are instances of the DType class, they are not subclasses of DType.  We could make them into classes by defining them as subclasses of DType, but then we need to deal with constructors which don't make sense for the primitive types. Even for DH DateTime etc. - the non-primitive classes, we have a problem with whether to wrap them or not. In the case of receiving a column of DateTime as a Java array, it is best that the data is represented as the jpy type. However a wrapper of that also makes sense because we want to support translation between DH DateTime and the built-in datetime in Python or numpy/pandas specific implementation or some other implementation in other popular Python packages. (I don't know for sure, making that up just for making a point here.)

pyintegration/deephaven2/dtypes/_datetime.py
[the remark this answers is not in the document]
jmao-denver: this is intentional, to make __str__ and __repr__ return the same info

jmao-denver: I am leaning towards the direction of not wrapping the Java classes but instead of relying on utility function to wrap the static methods on these Java classes.

pyintegration/deephaven2/datetimeutils.py
```diff
@@ -0,0 +1,803 @@
...
+        return Period(_JDateTimeUtils.convertPeriod(s))
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def convert_time(s, quiet: bool = False) -> int:
+    """ Converts a string time to nanoseconds from Epoch.
+
+    Args:
+        s (str): in the format of: hh:mm:ss[.nnnnnnnnn]
```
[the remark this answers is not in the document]
jmao-denver: Took it straight from the Java code
```
   /**
     * Converts a String time to nanoseconds from Epoch. The format for the String is:
     * <p>
     * hh:mm:ss[.nnnnnnnnn].
     *
     * @param s The String to be evaluated and converted.
     * @return A long value representing an Epoch offset in nanoseconds. Throws {@link RuntimeException} if the String
     *         cannot be parsed.
     */
    public static long convertTime(String s) {
        long ret = convertTimeQuiet(s);

        if (ret == io.deephaven.util.QueryConstants.NULL_LONG) {
            throw new RuntimeException("Cannot parse time : " + s);
        }

        return ret;
    }
```

pyintegration/deephaven2/datetimeutils.py
```diff
@@ -0,0 +1,803 @@
...
+        DHError
+    """
+    try:
+        j_datetime = dt.j_datetime if dt else None
+        return _JDateTimeUtils.format(j_datetime, tz.value)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def format_nanos(ns: int) -> str:
```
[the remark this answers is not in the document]
jmao-denver: I had the same question but that's what's in the Java class.

pyintegration/deephaven2/datetimeutils.py
```diff
@@ -0,0 +1,803 @@
...
+    Raises:
+        DHError
+    """
+    try:
+        return _JDateTimeUtils.format(ns)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def format_as_date(dt: DateTime, tz: TimeZone) -> str:
```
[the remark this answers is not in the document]
jmao-denver: not available in the Java code,  guess you could always call nanos_to_time() first

pyintegration/deephaven2/datetimeutils.py
```diff
@@ -0,0 +1,803 @@
...
+    """
+    try:
+        j_datetime1 = dt1.j_datetime if dt1 else None
+        j_datetime2 = dt2.j_datetime if dt2 else None
+        return _JDateTimeUtils.isBefore(j_datetime1, j_datetime2)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def lower_bin(dt: DateTime, interval: int, offset: int = 0) -> DateTime:
```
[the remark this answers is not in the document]
jmao-denver: can't find an overloaded Java method with that signature, some magic in the query engine?

pyintegration/deephaven2/datetimeutils.py
```diff
@@ -0,0 +1,803 @@
...
+        DHError
+    """
+    try:
+        j_datetime = dt.j_datetime if dt else None
+        return _JDateTimeUtils.millisOfSecond(j_datetime, tz.value)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def millis_to_nanos(ms: int) -> int:
```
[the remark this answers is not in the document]
jmao-denver: the typehint signals that None is not a valid argument. I believe type checkers like mypy would complain if you do

pyintegration/tests/test_datetimeutils.py
jmao-denver: no, they are not, ns_str2 has 'UTC' at the end.

pyintegration/tests/test_datetimeutils.py
jmao-denver: no

pyintegration/deephaven2/constants.py
chipkent, line 97: Should `TimeZone` go here, or in `time`?  Maybe `time`?

pyintegration/deephaven2/hmtl.py
```diff
@@ -0,0 +1,24 @@
+#
```
chipkent, line 1: file should be renamed to `html.py`

pyintegration/deephaven2/hmtl.py
```diff
@@ -0,0 +1,24 @@
...
+#   Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
+#
+import jpy
```
chipkent, line 4: Module needs docstring.

pyintegration/deephaven2/hmtl.py
```diff
@@ -0,0 +1,24 @@
...
+
+def to_html(table: Table) -> str:
+    """  Returns a printout of a table formatted as HTML. Limit use to small tables to avoid running out of memory.
+
+    Returns:
+        a String of the table printout formatted as HTML
+
+    Raises:
+        DHError
+    """
```
chipkent, line 20:
```suggestion
    """  Returns a table formatted as an HTML string. Limit use to small tables to avoid running out of memory.

    Returns:
        a table formatted as an HTML string

    Raises:
        DHError
    """
```

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
+#
+#   Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
+#
+""" This module defines functions for handling Deephaven date/time data. """
+
+import jpy
+
+from deephaven2 import DHError
+from deephaven2.constants import TimeZone
```
chipkent, line 9: I think this import suggests that `TimeZone` should be here and not in `constants`.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+HOUR = 60 * MINUTE  #: One hour in nanoseconds.
+DAY = 24 * HOUR  #: One day in nanoseconds.
+WEEK = 7 * DAY  #: One week in nanoseconds.
+YEAR = 52 * WEEK  #: One year in nanoseconds.
+
+_JDateTimeUtils = jpy.get_type("io.deephaven.time.DateTimeUtils")
+
+
+def to_datetime(s: str, quiet: bool = False):
+    """ Converts a datetime string to a DateTime object
```
chipkent, line 23:
```suggestion
    """ Converts a datetime string to a DateTime object.
```

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+
+    Args:
+        s (str): in the form yyyy-MM-ddThh:mm:ss and with optional sub-seconds after an
+            optional decimal point, followed by a mandatory time zone character code
```
chipkent, line 27:
```suggestion
        s (str): in the form "yyyy-MM-ddThh:mm:ss[.SSSSSSSSS] TZ".
```
Is this more or less clear?

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+        quiet (bool, optional): when True, if the datetime string can't be parsed, this function returns None, otherwise
+            it raises an exception; default is False
```
chipkent, line 29:
```suggestion
        quiet (bool): when True, if the datetime string can't be parsed, this function returns None; otherwise
            it raises an exception.  The default is False
```
You haven't documented optional like this anywhere else in your code.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+MINUTE = 60 * SECOND  #: One minute in nanoseconds.
+HOUR = 60 * MINUTE  #: One hour in nanoseconds.
+DAY = 24 * HOUR  #: One day in nanoseconds.
+WEEK = 7 * DAY  #: One week in nanoseconds.
+YEAR = 52 * WEEK  #: One year in nanoseconds.
+
+_JDateTimeUtils = jpy.get_type("io.deephaven.time.DateTimeUtils")
+
+
+def to_datetime(s: str, quiet: bool = False):
```
chipkent, line 22: No return type.  Check this on all functions.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+        raise DHError(e) from e
+
+
+def to_period(s: str, quiet: bool = False) -> Period:
+    """ Converts a period string into a Period object.
+
+    Args:
+        s (str): a string in the form of nYnMnWnDTnHnMnS, with n being numeric values, e.g. 1W for one week, T1M for
+            one minute, 1WT1H for one week plus one hour
+        quiet (bool, optional): when True, if the period string can't be parsed, this function returns None, otherwise
```
chipkent, line 52: Copy the quite change from above everywhere it matters.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def to_nanos(s, quiet: bool = False) -> int:
+    """ Converts a time string to nanoseconds from the Epoch.
+
+    Args:
+        s (str): in the format of: hh:mm:ss[.nnnnnnnnn]
+        quiet (bool): to return None or raise an exception if the string can't be parsed, default is False
```
chipkent, line 75: This quiet string is different than the prior.  More cut and paste.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+        return _JDateTimeUtils.convertPeriod(s)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def to_nanos(s, quiet: bool = False) -> int:
+    """ Converts a time string to nanoseconds from the Epoch.
+
+    Args:
+        s (str): in the format of: hh:mm:ss[.nnnnnnnnn]
```
chipkent, line 74: See my suggested comment above.  Do we want to use `[.nnnnnnnnnn]` or `[.SSSSSSSSS]`?  Java uses the latter, but I don't know if there is another standard that uses the former.
https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+    if quiet:
+        return _JDateTimeUtils.convertTimeQuiet(s)
+
+    try:
+        return _JDateTimeUtils.convertTime(s)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def now():
```
chipkent, line 92: return type

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+    """ Provides the current date/time.
```
chipkent, line 93:
```suggestion
    """ Provides the current datetime.
```
As it is written, it looks like it returns one or the other.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+    Raises:
+        DHError
+    """
+    try:
+        return _JDateTimeUtils.currentTime()
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def date_at_midnight(dt, tz: TimeZone):
```
chipkent, line 107:
```suggestion
def datetime_at_midnight(dt, tz: TimeZone):
```
Is this more clear?

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+    """
+    try:
+
+        return _JDateTimeUtils.dateAtMidnight(dt, tz.value)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def day_of_month(dt, tz: TimeZone) -> int:
+    """ Returns an 1-based int value of the day of the month for a DateTime and specified time zone.
```
chipkent, line 128:
```suggestion
    """ Returns an 1-based int value of the day of the month for a DateTime and specified time zone.  January is 1 and December is 12.
```

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+        DHError
+    """
+    try:
+        return _JDateTimeUtils.currentTime()
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def date_at_midnight(dt, tz: TimeZone):
+    """ Returns a DateTime for the requested DateTime at midnight in the specified time zone.
```
chipkent, line 108: What happens if an input is None?

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+
+        return _JDateTimeUtils.dateAtMidnight(dt, tz.value)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def day_of_month(dt, tz: TimeZone) -> int:
+    """ Returns an 1-based int value of the day of the month for a DateTime and specified time zone.
+
+    Args:
```
chipkent, line 130: What happens if an input is None?

... same comment on all docs in this module.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+        DHError
+    """
+    try:
+        return _JDateTimeUtils.diffNanos(dt1, dt2)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def format_datetime(dt, tz: TimeZone) -> str:
+    """ Returns a string date/time representation formatted as yyyy-MM-ddThh:mm:ss.nnnnnnnnn TZ.
```
chipkent, line 205: 1) I would put the format in quotes.
2) See comments above for `n` vs `S`.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+
+    Args:
+        dt (DateTime): the DateTime to format as a string
+        tz (TimeZone): the TimeZone to use when interpreting the date/time
```
chipkent, line 209: I think you want `datetime` or `DateTime` and not `date/time`.  The last makes it look like a date or a time, while datetime variants are clear.  Check all of the docks for this.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+    Returns:
+        str
+
+    Raises:
+        DHError
+    """
+    try:
+        return _JDateTimeUtils.format(ns)
+    except Exception as e:
+        raise DHError(e) from e
```
chipkent, line 238: 1) See above comments about docs.
2) The method should take a timezone.
3) I assume the format string also includes a timezone.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+
+
+def format_date(dt, tz: TimeZone) -> str:
+    """ Returns a string date representation of a DateTime interpreted for a specified time zone formatted as yyy-MM-dd.
```
chipkent, line 242: I would quote the format.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+        DHError
+    """
+    try:
+        return _JDateTimeUtils.hourOfDay(dt, tz.value)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def is_after(dt1: DateTime, dt2: DateTime) -> bool:
+    """ Evaluates whether one DateTime value is later than a second DateTime value.
```
chipkent, line 281: In the docs `date/time`, `datetime`, and `DateTime` are used.  It seems like we probably want to use `DateTime` most places.  The existing docs use a hyphenated `date-time` when not using the specific `DateTime`.
https://deephaven.io/core/docs/how-to-guides/work-with-date-time/

I also see in that link that we are using `f` for fractions of a second.  Yet another variation.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+    Raises:
+        DHError
+    """
+    try:
+        return _JDateTimeUtils.isBefore(dt1, dt2)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def lower_bin(dt, interval: int, offset: int = 0):
```
chipkent, line 320: return type

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def minus_nanos(dt, ns: int):
+    """ Subtracts nanoseconds from a DateTime.
+
+    Args:
+        dt (DateTime): the starting DateTime value
+        ns (int): the long number of nanoseconds to subtract from dateTime
```
chipkent, line 459:
```suggestion
        ns (int): the number of nanoseconds to subtract from DateTime
```

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+        DHError
+    """
+    try:
+        return _JDateTimeUtils.minus(dt, period)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def minute_of_day(dt, tz: TimeZone) -> int:
+    """ Returns an int value of minutes since midnight for a DateTime in the specified time zone.
```
chipkent, line 493:
```suggestion
    """ Returns the number of minutes since midnight for a DateTime in the specified time zone.
```

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+        DHError
+    """
+    try:
+        return _JDateTimeUtils.minuteOfDay(dt, tz.value)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def minute_of_hour(dt, tz: TimeZone) -> int:
+    """ Returns an int value of minutes since the top of the hour for a DateTime in the specified time zone.
```
chipkent, line 512:
```suggestion
    """ Returns the number of minutes since the top of the hour for a DateTime in the specified time zone.
```

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+    """
+    try:
+
+        return _JDateTimeUtils.minuteOfHour(dt, tz.value)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def month_of_year(dt, tz: TimeZone) -> int:
+    """ Returns an 1-based int value for the month of a DateTime in the specified time zone.
```
chipkent, line 532:
```suggestion
    """ Returns a 1-based int value for the month of a DateTime in the specified time zone.  January is 1, and December is 12.
```

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+        DHError
+    """
+    try:
+        return _JDateTimeUtils.nanosOfDay(dt, tz.value)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def nanos_of_second(dt, tz: TimeZone) -> int:
+    """ Returns a long value of nanoseconds since the top of the second for a DateTime in the specified time zone.
```
chipkent, line 589:
```suggestion
    """ Returns the number of nanoseconds since the top of the second for a DateTime in the specified time zone.
```

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+        return _JDateTimeUtils.nanosToMillis(ns)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def nanos_to_datetime(ns: int):
+    """ Converts a value of nanoseconds from Epoch to a DateTime.
+
+    Args:
+        ns (long): the long nanoseconds since Epoch value to convert
```
chipkent, line 630:
```suggestion
        ns (int): nanoseconds since Epoch value to convert
```
Search the docs for "int" and "long" to see what needs to get cleaned up.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+    Returns:
+        DateTime
+    """
+    try:
+        return _JDateTimeUtils.nanosToTime(ns)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def plus_period(dt, period):
```
chipkent, line 641: type annotation

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def plus_nanos(dt, ns: int):
+    """ Adds nanoseconds to a DateTime.
+
+    Args:
+        dt (DateTime): the starting DateTime value
+        ns (int): the long number of nanoseconds to add to DateTime
```
chipkent, line 665:
```suggestion
        ns (int): the number of nanoseconds to add to DateTime
```

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+        DHError
+    """
+    try:
+        return _JDateTimeUtils.plus(dt, ns)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def second_of_day(dt, tz: TimeZone) -> int:
+    """ Returns an int value of seconds since midnight for a DateTime in the specified time zone.
```
chipkent, line 680:
```suggestion
    """ Returns the number of seconds since midnight for a DateTime in the specified time zone.
```
We don't need to say that things return an int, because it is listed below.  We just need to get across what it is doing as clearly as possible. in this sentence.  Check other methods.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,775 @@
...
+        DHError
+    """
+    try:
+        return _JDateTimeUtils.secondOfDay(dt, tz.value)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def second_of_minute(dt, tz: TimeZone) -> int:
+    """ Returns an int value of seconds since the top of the minute for a DateTime in the specified time zone.
```
chipkent, line 699:
```suggestion
    """ Returns the number of seconds since the top of the minute for a DateTime in the specified time zone.
```

pyintegration/tests/test_time.py
```diff
@@ -0,0 +1,236 @@
...
+        timezone_str = "IN"
+        dt = to_datetime(f"{datetime_str} {timezone_str}")
+        self.assertEqual(2021, year(dt, TimeZone.IN))
+
+    def test_year(self):
+        datetime_str = "2021-12-10T00:01:05"
+        timezone_str = "JP"
+        dt = to_datetime(f"{datetime_str} {timezone_str}")
+        self.assertEqual(21, year_of_century(dt, TimeZone.JP))
+
```
chipkent, line 233: For all of the methods that take inputs, I would test `None` inputs so that we can properly document the behavior.

chipkent (changes requested):

pyintegration/deephaven2/constants.py
jmao-denver: agree

pyintegration/deephaven2/time.py
jmao-denver: I like it better

pyintegration/deephaven2/time.py
jmao-denver: dtypes.DateTime isn't a Python class but a DType instance. So return type hint isn't possible.

pyintegration/deephaven2/time.py
jmao-denver: Not month_of_year

pyintegration/deephaven2/time.py
```diff
@@ -143,16 +197,16 @@ def day_of_month(dt, tz: TimeZone) -> int:
...
 
-def day_of_week(dt, tz: TimeZone) -> int:
+def day_of_week(dt: DateTime, tz: TimeZone) -> int:
     """ Returns an 1-based int value of the day of the week for a DateTime in the specified time zone, with 1 being
      Monday and 7 being Sunday.
 
     Args:
         dt (DateTime): the DateTime for which to find the day of the week.
-        tz (TimeZone): the TimeZone to use when interpreting the date/time.
+        tz (TimeZone): the TimeZone to use when interpreting the Datetime.
```
chipkent, line 206:
```suggestion
        tz (TimeZone): the TimeZone to use when interpreting the DateTime.
```

pyintegration/deephaven2/time.py
```diff
@@ -221,7 +275,7 @@ def format_datetime(dt, tz: TimeZone) -> str:
 
 
 def format_nanos(ns: int) -> str:
-    """ Returns a string date/time representation formatted as yyyy-MM-ddThh:mm:ss.nnnnnnnnn.
+    """ Returns a string DateTime representation formatted as "yyyy-MM-ddThh:mm:ss.SSSSSSSSS".
```
chipkent, line 278: This function should accept a timezone and return a timezone in the string.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,830 @@
...
+        return _JDateTimeUtils.convertPeriodQuiet(s)
+
+    try:
+        return _JDateTimeUtils.convertPeriod(s)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def to_nanos(s, quiet: bool = False) -> int:
+    """ Converts a time string to nanoseconds from the Epoch.
```
chipkent, line 125: I think this description is wrong.  I think we are taking a string like "01:02:03" and converting it to HOUR+2*MINUTE+3*SECOND.  This is not time since the Epoch.  Am I wrong?

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,830 @@
...
+        DHError
+    """
+    try:
+        return _JDateTimeUtils.format(dt, tz.value)
+    except Exception as e:
+        raise DHError(e) from e
+
+
+def format_nanos(ns: int) -> str:
+    """ Returns a string DateTime representation formatted as "yyyy-MM-ddThh:mm:ss.SSSSSSSSS".
```
chipkent, line 278: Is this formatting epoch nanos to a string date time?  If so, it should take in a timezone and output a full "yyyy-MM-ddThh:mm:ss.SSSSSSSSS TZ" format.  On the other hand, if this takes in nanos like 1*HOUR+2*MINUTE+3*SECOND and returns "hh:mm:ss.SSSSSSSSS", then the function signature is fine, but the documentation isn't.

pyintegration/deephaven2/time.py
```diff
@@ -0,0 +1,830 @@
...
+
+    Args:
+        ns (int): the number of nanoseconds from the Epoch
```
chipkent, line 281: See comment above.  Is it really Echoch nanos or just nanos?

pyintegration/tests/test_time.py
```diff
@@ -0,0 +1,258 @@
...
+    def test_format_datetime(self):
+        dt = now()
+        self.assertIn(TimeZone.SYD.name, format_datetime(dt, TimeZone.SYD))
+
+    def test_format_nanos(self):
+        dt = now()
+        ns = nanos(dt)
+        ns_str1 = format_nanos(ns).split(".")[-1]
+        ns_str2 = format_datetime(dt, TimeZone.UTC).split(".")[-1]
+        self.assertTrue(ns_str2.startswith(ns_str1))
```
chipkent, line 93: If this is a clean test depends upon the answers to the questions I raised about the nanos methods.

chipkent (changes requested):

pyintegration/deephaven2/time.py
jmao-denver: you are right. The description is wrong, just straight conversion not from Epoch.

pyintegration/deephaven2/time.py
jmao-denver: yes, the doc is wrong.

pyintegration/deephaven2/time.py
jmao-denver: Just nanos.

chipkent (approved):

PR #1813, opened 2022-01-10
Added AggFormula in io.deephaven.api.agg.Aggregation, fixes #1699

### LINKED ISSUE #1699: Add aggFormula support in deephaven2
AggFormula is added back after the db module refactoring. Needs to add it in the wrapper.

related to #1263

jmao-denver:
also made it avaiable in deephaven2 via the agg.py wrapper

### DIFF
--- a/pyintegration/tests/test_table.py
+++ b/pyintegration/tests/test_table.py
@@ -5,7 +5,7 @@ import unittest
 
 from deephaven2 import DHError, read_csv, empty_table, SortDirection
 from deephaven2.agg import sum_, weighted_avg, avg, pct, group, count_, first, last, max_, median, min_, std, abs_sum, \
-    var
+    var, formula
 from deephaven2.table import Table
 from tests.testbase import BaseTestCase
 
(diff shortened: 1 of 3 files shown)


pyintegration/tests/test_table.py
```diff
@@ -283,7 +284,6 @@ def test_agg_by_2(self):
                 weighted_avg("var", ["weights"])]
 
         result_table = test_table.agg_by(aggs, ["dumb"])
```
devinrsmith, line 286: Not part of this review, but I think the variable name `dumb` is "improper". I'd advise choosing completely "neutral" names commonly used (`foo`, `bar`, `baz`), or choose something "meaningful" such as `key`, or `my_group`. I'd suggest an audit of the client for other potentially "improper" names.

devinrsmith (approved):
Changes look good.

pyintegration/tests/test_table.py
jmao-denver: will do.

devinrsmith (approved):

PR #1814, opened 2022-01-10
Support unions with null in Avro nested record traversals; BigDecimals via ByteBuffer.

jcferretti:


extensions/kafka/src/main/java/io/deephaven/kafka/ingest/GenericRecordBigDecimalFieldCopier.java
```diff
@@ -34,14 +35,31 @@ public void copyField(
...
+                final BigDecimal bd = new BigDecimal(bi, scale);
+                output.set(ii + destOffset, bd);
+            } else if (bytesObj instanceof ByteBuffer) {
+                final ByteBuffer bb = (ByteBuffer) bytesObj;
+                final BigInteger bi;
+                if (bb.hasArray()) {
+                    bi = new BigInteger(bb.array(), bb.position() + bb.arrayOffset(), bb.remaining());
+                } else {
+                    final byte[] bytes = new byte[bb.remaining()];
+                    bb.get(bytes);
```
niloc132, line 55: are you allowed/expected to move the position of the bytebuffer?

jcferretti: That is a very good question; I did not stop to think about that, good catch.

I had to look at the source code to confirm what was my (implicit) intuition here, that an object returned has to be defensively copied so it can't be shared context etc (a different API could/should be used to avoid the copy and still give the user a way to copy over from a state buffer etc).

The source in question in the Avro implementation is GenericData.java.   It does object copy before return across the board, so we are safe.

niloc132 (approved):

PR #1819, opened 2022-01-12
Initial CDC support for DHC

jcferretti:
Deephaven - Google Chrome_140

### DIFF
--- a/Integrations/python/deephaven/ConsumeCdc.py
+++ b/Integrations/python/deephaven/ConsumeCdc.py
@@ -68,7 +68,7 @@ def consumeToTable(
     :param cdc_spec:  A CDC Spec opaque object obtained from calling either the cdc_explict_spec method
                       or the cdc_short_spec method
     :param partitions: Either a sequence of integer partition numbers or the predefined constant
-        ALL_PARTITIONS for all partitions.  Defaults to ALL_PARTITIONS if unespecified.
+        ALL_PARTITIONS for all partitions.  Defaults to ALL_PARTITIONS if unspecified.
     :param ignore_key: Whether to ignore the key related columns for the CDC stream.
         If true, and the source is not append only, the source table will be treated as if
         the primary key contained all columns.  Defaults to FALSE.
--- a/Integrations/python/deephaven/ConsumeCdc.py
+++ b/Integrations/python/deephaven/ConsumeCdc.py
@@ -117,7 +117,7 @@ def consumeRawToTable(
     return _java_type_.consumeRawToTable(kafka_config, cdc_spec, partitions, table_type_enum)
 
 @_passThrough
-def cdc_explicit_spec(
+def cdc_long_spec(
         topic:str,
         key_schema_name:str,
         key_schema_version:str,
@@ -125,6 +125,9 @@ def cdc_explicit_spec(
         value_schema_version:str,
 ):
     """
+    Create a CdcSpec opaque object (necessary for one argument in a call to consume*ToTable)
+    via explicitly specifying all configuration options.
+
     :param topic:  The Kafka topic for the CDC events associated to the desired table data.
     :param key_schema_name:  The schema name for the Key Kafka field in the CDC events for the topic.
         This schema should include definitions for the columns forming the PRIMARY KEY of the underlying table.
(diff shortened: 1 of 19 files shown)


docker-compose-common.yml
```diff
@@ -14,10 +14,11 @@ services:
       # For remote debugging switch the line above for the one below (and also change the ports below)
       # - JAVA_TOOL_OPTIONS=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005 -Xmx4g -Ddeephaven.console.type=${DEEPHAVEN_CONSOLE_TYPE} -Ddeephaven.application.dir=${DEEPHAVEN_APPLICATION_DIR}
 
+
     expose:
       - '8080'
-#    ports:
-#      - '5005:5005'        # For remote debugging (change if using different port)
```
cpwright, line 20: meant to be left open?

Integrations/python/deephaven/ConsumeCdc.py
```diff
@@ -0,0 +1,111 @@
...
+        KafkaConsumer specific desired configuration here.  Note this should include the relevant property
+        for a schema server URL where the key and/or value Avro necessary schemas are stored.
+    :param server_name:  The CDC server name.  This is used to compute the Kafka topic name (first component),
+        and also the Avro schema names expected in schema server for either key or value.
+    :param db_name:  The CDC database name.  This is used to compute the Kafka topic name (second component),
+        and also the Avro schema names expected in schema server for either key or value.
+    :param table_name:  The CDC table name.  This is used to compute the Kafka topic name (third component),
+        and also the Avro schema names expected in schema server for either key or value.
+    :param partitions: Either a sequence of integer partition numbers or the predefined constant
+        ALL_PARTITIONS for all partitions.  Defaults to ALL_PARTITIONS if unespecified.
```
cpwright, line 80: s/unespecified/unspecified/

cpwright (commented):
I think you've basically taken away a stream table as a valid output.  You can:
(1) ignore the keys and get a lastBy
(2) just make it append only, so you have a huge table
(3) what about if I know what I'm doing and want the raw stream table instead?  Like maybe I don't actually care about these individual things, but I want to aggregate them with counts and sums.  Or in an advanced use case maybe I can instead do something like:
purchases = streamTable.where(op = add).view("Dollars").sumBy()
returns = streamTable.where(op = del).view("Dollars").sumBy()
netRevenue = purchases.naturalJoin(returns, "", "ReturnDollars=Dollars").update("NetRevenue=Dollars-ReturnDollars")

jcferretti:
Ref https://github.com/deephaven/deephaven-core/pull/1819#pullrequestreview-849885407
(I can't answer directly under that comment, sadly, very poor github UI design).
Makes sense and it becomes obvious when looking at the implementation code; just returning the streaming table before wrapping it in a call to streamingToAppendTable is obviously more general and the user themselves can do the wrapping if they so desire.

I've reworked some other parts of the change too; there was a bug in that the after__X columns are not present on a delete, so it is important to take the primary key columns always from the Key component (those are present in a delete).

Something else I learned from debezium documentation, when an update in the underlying database table modifies a primary key, debezium generates two events, one to delete the old row and one to add the new row.  Is very important for our algorithm that they do this, otherwise the trick for 

```
.where("op = `d`")
```

wouldn't work; I thought about this case,  went into panic, went to read the debezium docs, and was relieved.

extensions/kafka/src/main/java/io/deephaven/kafka/CdcTools.java
```diff
@@ -2,99 +2,370 @@
...
-                        ? KafkaTools.Consume.IGNORE
-                        : KafkaTools.Consume.avroSpec(keySchema),
+                KafkaTools.Consume.avroSpec(keySchema),
                 KafkaTools.Consume.avroSpec(valueSchema),
                 KafkaTools.TableType.Stream);
         final List<String> dbTableColumnNames = dbTableColumnNames(streamingIn);
+        final String[] allDroppedColumns;
+        if (dropColumns != null && dropColumns.size() > 0) {
+            int i = 0;
+            allDroppedColumns = new String[dropColumns.size() + (asStreamTable ? 0 : 1)];
```
cpwright, line 307: maybe just Arrays.copyOf?

cpwright (dismissed):

cpwright (dismissed):

cpwright (approved):

Issue #1818, opened 2022-01-11
Quick Filters char filters do not work with do not equal


mofojed:
**Description**

**Steps to reproduce**
1. Create a table with a char column:
```
from deephaven.TableTools import emptyTable
charTable = emptyTable(26).update("A=(char)(65+i)")
```
2. In the Quick Filter bar, enter `!A`

**Expected results**
2. `A` should be filtered out of the table

**Actual results**
2. `A` is still visible in the table

**Additional details and attachments**
![image](https://user-images.githubusercontent.com/4505624/149034677-e34144e4-5e90-4a4a-8143-3e19020476cb.png)

rcaudy:
Probably fixed by https://github.com/deephaven/deephaven-core/pull/1820

PR #1820, opened 2022-01-12
Fix missing "!=" support for WhereFilterFactory, and tighten up the char literal matching regex

rcaudy:


engine/table/src/main/java/io/deephaven/engine/table/impl/select/WhereFilterFactory.java
```diff
@@ -56,12 +60,20 @@ public WhereFilter getExpression(String expression, Matcher matcher, Object... a
                 }
                 log.debug().append("WhereFilterFactory creating MatchFilter for expression: ").append(expression)
                         .endl();
-                return new MatchFilter(MatchFilter.CaseSensitivity.MatchCase, columnName, matcher.group(2));
+                // @formatter:off
+                return new MatchFilter(
+                        MatchFilter.CaseSensitivity.MatchCase,
+                        matcher.group(3) == null // We must have matched EQUALS_PTRN, rather than NOT_EQUALS_PTRN
```
cpwright, line 66: this is very subtle

cpwright (approved):

engine/table/src/main/java/io/deephaven/engine/table/impl/select/WhereFilterFactory.java
rcaudy: That's why it has a comment. It's the same way as ICASE and NOT work.

cpwright (approved):

engine/table/src/main/java/io/deephaven/engine/table/impl/select/WhereFilterFactory.java
rcaudy: I re-did the whole class' handling of matcher groups, see if this is better.

cpwright (approved):

cpwright (approved):

rcaudy:
Fixes #1818

cpwright (approved):

PR #1824, opened 2022-01-13
Add AutoService

devinrsmith:
From a suggestion from @niloc132 in https://github.com/deephaven/deephaven-core/pull/1787, I decided to plumb AutoService for better maintainability and visibility into ServiceLoader services. This reduces the developer burden for creating the services files manually by using https://github.com/google/auto/tree/master/service - this is a compile/annotation-time-only dependency, no additional runtime dependencies. In addition, it likely fares better during refactoring, and provides immediate context to anybody looking at the code that this will be auto-discovered during runtime.

As part of the PR, I noticed that `open-api/lang-parser/src/main/resources/META-INF/services/io.deephaven.lang.parse.api.CompletionParseService` is not a real service (maybe it used to be?).


fishconfig-local/build.gradle
```diff
@@ -6,4 +6,8 @@ description = 'Provides a io.deephaven.configuration.PropertyInputStreamLoader t
 
 configurations {
   compile.extendsFrom fishConfig
+}
+
+dependencies {
+  Classpaths.inheritAutoService(project)
 }
```
rcaudy, line 13: Missing trailing newline.

rcaudy (commented):
What about `deephaven-core/py/jpy/src/test/resources/META-INF/services/javax.script.ScriptEngineFactory` ?

devinrsmith:
Preferring to not push down into jpy just to minimize churn there.

rcaudy (approved):

JamesXNelson:
Indeed, it started as a service, then colin refactored it to not need to be a service anymore.  Feel free to rename / delete as you see fit.

Issue #1546, opened 2021-11-09
Remake conversion_utils.py and change NULL_CHAR to MAX_VALUE


jmao-denver:
Since we expect users to test special DH values in their code, we need to present or wrap them appropriately in the new Python package.

related to #1535

jmao-denver:
closed by #1825

PR #1825, opened 2022-01-13
Hand wrapped KafkaTools, ParquetTools, fixes #1643, #1546, #1892

### LINKED ISSUE #1643: hand-written wrappers  for Kafka/Parquet tools
to make the interface more pythonic

relates to #1263

jmao-denver:
1. created stream/kafka package which included the consumer.py and producer.py modules
2. added the parquet.py. module
3. converted and added docstrings to the new standards
4. refitted the existing test code to the new interfaces and added some new cases


pyintegration/deephaven2/dtypes.py
```diff
@@ -146,3 +147,20 @@ def j_array_list(values: Iterable):
...
+    instances = re.findall("<class '(.*?)'>", str(type(obj)))
+    if len(instances) > 0:
+        return instances[0]
+    else:
+        return None
+
+
+def is_java_type(obj, return_name=False):
+    class_name = _get_full_type_name(obj)
+    value = (class_name is not None) and (class_name in jpy.types)
```
devinrsmith, line 162: why not use `isinstance(obj, jpy.JType)`?

devinrsmith: And then, if we know it's a java type, we can do:

```
obj.jclass.getName()
```

jmao-denver: good idea, this is from the old conversion_utils.py, I probably shouldn't have just copied without questioning it.

pyintegration/deephaven2/stream/kafka/consumer.py
```diff
@@ -0,0 +1,289 @@
...
+
+
+def consume(
+        kafka_config: dict,
+        topic: str,
+        partitions=None,
+        offsets=None,
+        key=None,
+        value=None,
+        table_type='stream'
```
jcferretti, line 59: No `:str` type declared?

pyintegration/deephaven2/stream/kafka/_utils.py
```diff
@@ -0,0 +1,47 @@
...
+def _dict_to_j_properties(d):
+    r = _JProperties()
+    for key, value in d.items():
+        if value is None:
+            value = ''
+        r.setProperty(key, value)
+    return r
+
+
+def _dict_to_j_map(d):
```
jcferretti, line 23: These utilities will be useful for more modules I think.  At least the upcoming stream.cdc, so perhaps a better place for them would be at a higher directory level.

pyintegration/deephaven2/__init__.py
```diff
@@ -5,17 +5,21 @@
 unlocks the unique and tremendous power of Deephaven to the Python community.
```
jcferretti, line 5: To me this sounds like too much of a blatant marketing plug to belong in the source code.

jcferretti (commented):
I reviewed Kafka and Parquet parts of the diff; can't comment to changes to pre-existing files in pyintegration/deephaven2

pyintegration/deephaven2/dtypes.py
jcferretti: My recollection is that I tried this `isinstance(obj, jpy.JType)` and it did not work consistently.  Be sure to test thoroughly if you try this change.

devinrsmith: If we find any exceptional cases where we think something should be JType, but isn't, we need to capture that information in a ticket.

We do need some sort of ABC eventually (if we don't already) for all of our handwrapped python classes, such that we know they are handwrapped java types.

Come across that a few times, and I've made notes about it.

* https://github.com/deephaven/deephaven-plugin-json/blob/v0.0.1.dev3/src/deephaven/plugin/json/__init__.py#L49-L51
* https://github.com/devinrsmith/deephaven-core/blob/a7fc020d58401ba231f8fcf81f27d9b50fe9e43b/pyintegration/deephaven2/server/script_session/__init__.py#L32-L46

pyintegration/tests/test_parquet.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+class ParquetTestCase(BaseTestCase):
+    """ Test cases for the deephaven.ParquetTools module (performed locally) """
+    root_dir = ""
+
+    @classmethod
+    def setUpClass(cls):
+        # define a junk table workspace directory
+        cls.root_dir = os.path.join(get_workspace_root(), 'TestParquetTools')
```
chipkent, line 23: This looks like the bad old days of the DHE workspace.  Is there any reason you can't just use a python temp dir
https://docs.python.org/3/library/tempfile.html

pyintegration/tests/test_parquet.py
```diff
@@ -0,0 +1,144 @@
...
+            self.assertTrue(os.path.exists(file_location))
+            shutil.rmtree(base_dir)
+        with self.subTest(msg="write_tables(Table[], destinations, col_definitions"):
+            write_tables([table, table], [file_location, file_location2], definition)
+            self.assertTrue(os.path.exists(file_location))
+            self.assertTrue(os.path.exists(file_location2))
+
+        # Reading
+        with self.subTest(msg="read_table(str)"):
+            table2 = read_table(file_location)
```
chipkent, line 52: Should you check that the contents are equal?

pyintegration/tests/test_parquet.py
```diff
@@ -0,0 +1,144 @@
...
+            shutil.rmtree(base_dir)
+        with self.subTest(msg="write_tables(Table[], destinations, col_definitions, "):
+            write_tables([table, table], [file_location, file_location2], col_definitions,
+                         col_instructions=[col_inst, col_inst1])
+            self.assertTrue(os.path.exists(file_location))
+            self.assertTrue(os.path.exists(file_location2))
+
+        # Reading
+        with self.subTest(msg="read_table(str)"):
+            table2 = read_table(path=file_location, col_instructions=[col_inst, col_inst1])
```
chipkent, line 99: assert equal?

pyintegration/tests/test_parquet.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @classmethod
+    def tearDownClass(cls):
+        # remove the junk definitions created in the tests, if they exist...
+        if os.path.exists(cls.root_dir):
+            try:
+                shutil.rmtree(cls.root_dir)
+            except Exception as e:
+                print("Tried removing directory {}, but failed with error {}. "
+                      "Manual clean-up may be necessary".format(cls.root_dir, e))
```
chipkent, line 140: I would put this at the top by the setup method, so that it is very clear what is happening.  Until I got down here, I assumed there was no teardown.

pyintegration/tests/test_produce_kafka.py
```diff
@@ -0,0 +1,138 @@
...
+    def test_simple(self):
+        """
+        Check a simple Kafka subscription creates the right table.
+        """
+        t = new_table(cols=[double_col('Price', [10.0, 10.5, 11.0, 11.5])])
+        cleanup = pk.produce(
+            t,
+            {'bootstrap.servers': 'redpanda:29092'},
+            'orders',
+            key=pk.IGNORE,
```
chipkent, line 43: Looking at this, I think an `ignore()` function may be the way to go.  That way, a user knows they always need to call a function, instead of switching between a function and a constant.

chipkent (changes requested):

pyintegration/deephaven2/parquet.py
```diff
@@ -0,0 +1,206 @@
...
+    except Exception as e:
+        raise DHError(e, "failed to write to parquet data.") from e
+
+
+def write_tables(tables: List[Table], destinations: List[str], col_definitions: List[Column],
+                 col_instructions: List[ColumnInstruction] = None, compression_codec_name: str = None,
+                 max_dictionary_keys: int = None, grouping_cols: List[str] = None):
+    """ Writes tables to disk in parquet format to a supplied set of destinations. If you specify grouping columns,
+    there must already be grouping information for those columns in the sources. This can be accomplished with
+    .groupBy(<grouping columns>).ungroup() or .sort(<grouping column>).
```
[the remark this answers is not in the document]
jcferretti: These is just carrying over the javadoc for the java code as pydoc:

https://github.com/deephaven/deephaven-core/blob/1dc07f3c939e4a46c6ec698eb3cafcf0b0a4a254/extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java#L286

Any changes we make should be reflected there first (and sadly, copied over here since this is not auto-generated).

But either way, this PR doesn't look like the right place for that change.

pyintegration/deephaven2/parquet.py
```diff
@@ -0,0 +1,206 @@
...
+        else:
+            if write_instructions:
+                _JParquetTools.writeTable(table.j_table, _get_file_object(destination), write_instructions)
+            else:
+                _JParquetTools.writeTable(table.j_table, destination)
+    except Exception as e:
+        raise DHError(e, "failed to write to parquet data.") from e
+
+
+def write_tables(tables: List[Table], destinations: List[str], col_definitions: List[Column],
```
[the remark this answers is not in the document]
jcferretti: For 2, 3, and 4, similar comment as before: this is just bringing over javadoc from the java side, so if any changes are necessary, we should do them there:

https://github.com/deephaven/deephaven-core/blob/1dc07f3c939e4a46c6ec698eb3cafcf0b0a4a254/extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java#L286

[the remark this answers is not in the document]
jcferretti: On (2), after examining the java source, I think one value we are getting out of it is the fault tolerance of an "all or nothing" approach to writing.   Say we have 10 tables that together form some cohesive whole (a model for something, they are related to each other etc).  So we want to save all of them; if this was mysql, they would probably be in a separate "database" (a scope of sorts for tables).   Anyway, by using this method to write them, we ensure that we can't have a situation where, we wrote tables 1, 2, 3, 4 and it failed in table 5, so we have a "partial" result.   There is no ambiguity, you either get all your files, if everything worked, or you get no files of output, if anything at any point failed.  It is not parallel writes, it is sequential, but an error triggers an exception handler that cleans up any partial work up to that point when there is an error.  Not sure that is worth a HUGE lot, but it probably is worth something.  HTH.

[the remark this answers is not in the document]
jcferretti: On (3), the java implementation of `writeTable` (mapped here from `write_table`) get the `groupingColumns`, if any, from the table definition of that one table.   In `writeTables`, the code passes the provided groupingColumns down to the logic that writes parquet.  That logic btw creates a separate `X_grouping.parquet` file.
I am not sure on the details of how this works or why it makes sense, I think we need Ryan or Charles to figure this out.  Again, not sure if this conversation belongs in this PR tho.

pyintegration/tests/test_parquet.py
jmao-denver: no, but using the default DH workspace also makes sense as we won't potentially spill garbage outside our isolated env. If you feel too strongly about this, I'd be happy to change to use temp.

pyintegration/deephaven2/utils/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+import jpy
+
+
+_JDHConfig = jpy.get_type("io.deephaven.configuration.Configuration")
+
+
+def get_workspace_root():
+    """
+    Helper function for extracting the root directory for the workspace configuration
+    """
```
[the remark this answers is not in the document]
jmao-denver: @rcaudy said workspace isn't completely gone, but users shouldn't need to use it. This relates to your other comment about using Python temp files in the Parquet unit tests. Will make the change.

pyintegration/deephaven2/parquet.py
```diff
@@ -0,0 +1,206 @@
...
+                 max_dictionary_keys: int = None, grouping_cols: List[str] = None):
+    """ Writes tables to disk in parquet format to a supplied set of destinations. If you specify grouping columns,
+    there must already be grouping information for those columns in the sources. This can be accomplished with
+    .groupBy(<grouping columns>).ungroup() or .sort(<grouping column>).
+
+    Args:
+        tables (List[Table]): the source tables
+        destinations (List[str]): the destinations paths. Any non existing directories in the paths provided are
+            created. If there is an error any intermediate directories previously created are removed; note this makes
+            this method unsafe for concurrent use
```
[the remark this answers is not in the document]
jcferretti: On (1), I don't think we have enough context to understand what you mean, the comment above is multi-pronged (and addressed there).
On (2), yes, the java code will fail with an exception (a not particularly friendly one, we should fix that) if they don't.  So we should check, both here and in the java code.
On (3), this is just a thin wrapper for the java call, and as I mentioned in an earlier comment, this is not concurrent.  That is an interesting feature on its own, but is orthogonal with this PR I think.

pyintegration/deephaven2/stream/kafka/consumer.py
```diff
@@ -0,0 +1,289 @@
...
+_JKafkaTools_Consume = jpy.get_type("io.deephaven.kafka.KafkaTools$Consume")
+SEEK_TO_BEGINNING = getattr(_JKafkaTools, 'SEEK_TO_BEGINNING')
+DONT_SEEK = getattr(_JKafkaTools, 'DONT_SEEK')
+SEEK_TO_END = getattr(_JKafkaTools, 'SEEK_TO_END')
+FROM_PROPERTIES = getattr(_JKafkaTools, 'FROM_PROPERTIES')
+IGNORE = getattr(_JKafkaTools_Consume, 'IGNORE')
+ALL_PARTITIONS = getattr(_JKafkaTools, 'ALL_PARTITIONS')
+ALL_PARTITIONS_SEEK_TO_BEGINNING = getattr(_JKafkaTools, 'ALL_PARTITIONS_SEEK_TO_BEGINNING')
+ALL_PARTITIONS_DONT_SEEK = getattr(_JKafkaTools, 'ALL_PARTITIONS_DONT_SEEK')
+ALL_PARTITIONS_SEEK_TO_END = getattr(_JKafkaTools, 'ALL_PARTITIONS_SEEK_TO_END')
```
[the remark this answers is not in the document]
jmao-denver: using Enum here actually complicates using the typehints.

pyintegration/deephaven2/stream/kafka/producer.py
```diff
@@ -0,0 +1,195 @@
...
+        kafka_config: dict,
+        topic: str,
+        key,
+        value,
+        last_by_key_columns: bool = False
+):
+    """ Produce to Kafka from a Deephaven table.
+
+    Args:
+        table: a Deephaven table used as a source of rows to publish to Kafka.
```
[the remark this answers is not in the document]
jmao-denver: according to Cristian, any  table would work

pyintegration/deephaven2/stream/kafka/producer.py
```diff
@@ -0,0 +1,195 @@
...
+        key: a specification for how to map table column(s) to the Key field in produced
+            Kafka messages.  This should be the result of calling one of the methods
+            simple, avro or json in this module, or the constant IGNORE.
+        value: a specification for how to map table column(s) to the Value field in produced
+            Kafka messages.  This should be the result of calling one of the methods
+            simple, avro or json in this module, or the constant IGNORE.
+        last_by_key_columns:  Whether to publish only the last record for each unique key.
+            Ignored if key is IGNORE.  If key is not IGNORE and last_by_key_columns is false,
+            it is expected that table updates will not produce any row shifts; that is, the publisher
+            expects keyed tables to be streams, add-only, or aggregated.
```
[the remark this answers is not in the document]
jcferretti: That comment is carried over the java side and by looking at the latest version there, I believe is wrong.  I think it probably got updated in a sloppy way (likely by me) after some change, so the original comment was likely "vaguely meaningful if you try really hard", this is just pure nonsense.  I am going to send a separate PR to fix it.

The important thing to know in terms of how this works is this:
* If Key is IGNORE, `last_by_key_columns` is meaningless.
* When Key is not IGNORE and `last_by_key_columns` is true, the table we actually publish is: (quoting the java code)
```
            final Table effectiveTable = (!ignoreKey && lastByKeyColumns)
                    ? table.lastBy(keyColumns)
                    : table.coalesce();
```
The idea is that we are generating what in the Kafka world would be a stream of value updates based on keys.  Let's say that the table has a history of prices keyed by stock symbol.  Let's say that in the last refresh cycle, three distinct updates for `MSFT` where produced.   `last_by_key_columns` will only make us publish the last one.

Full code link:
https://github.com/deephaven/deephaven-core/blob/ac8970d817cb668857e9447210f7026089d74274/extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java#L1419

pyintegration/deephaven2/stream/kafka/producer.py
```diff
@@ -0,0 +1,195 @@
...
+            associated kafka_config dict in the call to consumeToTable should include the key
+            'schema.registry.url' with the associated value of the Schema Server URL for fetching the schema
+            definition.
+        schema_version:  if a string schema name is provided, the version to fetch from schema
+            service; if not specified, a default of 'latest' is assumed.
+        field_to_col_mapping: a dict mapping field names in the schema to column names in the Deephaven table.
+            Any fields in the schema not present in the dict as keys are mapped to columns of the same name.
+            If this argument is None, all schema fields are mapped to columns of the same name.
+        timestamp_field: a string for the name of an additional timestamp field to include,
+            or None for no such field.
```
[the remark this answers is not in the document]
cpwright: Generally, we always want to have some kind of timestamp for the stages we are processing.  Having timestamps in the payload instead of just things in a header is also sometimes beneficial when headers get stripped away, because your Avro data may be destined for places other than Kafka.

pyintegration/deephaven2/stream/kafka/producer.py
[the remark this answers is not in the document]
jcferretti: Reworded here: https://github.com/deephaven/deephaven-core/pull/1935
PTAL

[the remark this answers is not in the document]
rcaudy: Replaced by #1955

pyintegration/deephaven2/parquet.py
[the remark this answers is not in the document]
chipkent: This PR is addressing the Python v2 API.  If the API or API docs are weird, those comments belong in this PR.  Once v2 goes out, we are more stuck with what we have.

pyintegration/deephaven2/parquet.py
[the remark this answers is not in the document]
chipkent: This PR is addressing the Python v2 API.  If the API or API docs are weird, those comments belong in this PR.  Once v2 goes out, we are more stuck with what we have.

pyintegration/deephaven2/parquet.py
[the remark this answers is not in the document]
chipkent: On (3) I was just seeing if the pydocs should be telling a user to do ABC if they need to do something in parallel.  Apparently, there is no solution, so there is nothing to say.

pyintegration/deephaven2/stream/kafka/producer.py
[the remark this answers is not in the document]
chipkent: Sounds like we just need very clear pydocs.

pyintegration/deephaven2/stream/kafka/producer.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def __init__(self, j_spec):
+        self._j_spec = j_spec
+
+    @property
+    def j_object(self) -> jpy.JType:
+        return self._j_spec
+
+
+KeyValueSpec.IGNORE = KeyValueSpec(_JKafkaTools_Produce.IGNORE)
```
chipkent, line 29: See comments from consumer

pyintegration/deephaven2/dtypes.py
```diff
@@ -136,11 +133,65 @@ def array_from(self, seq: Sequence, remap: Callable[[Any], Any] = None):
...
             j_list.add(v)
         return j_list
-    except Exception as e:
-        raise DHError(e, "failed to create a Java ArrayList from the Python collection.") from e
+
+
+Properties = PropertiesDType(j_name="java.util.Properties")
+HashMap = HashMapDType(j_name="java.util.HashMap")
+HashSet = HashSetDType(j_name="java.util.HashSet")
+ArrayList = ArrayListDType(j_name="java.util.ArrayList")
```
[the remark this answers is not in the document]
chipkent: Also, I notice that these new types are just being used internally.  Are they things we expect users to want to do?  Maybe if they are integrating with java?

pyintegration/deephaven2/stream/kafka/producer.py
```diff
@@ -0,0 +1,183 @@
...
+            the 'exclude_columns' must be None.
+        exclude_columns (List[str]): the list of Deephaven column names to omit in the JSON output as fields, default
+            is None, meaning no column is omitted. If not None, include_columns must be None.
+        mapping (Dict[str, str]): a mapping from column names to JSON field names.  Any column name implied by earlier
+            arguments and not included as a key in the map implies a field of the same name. default is None,
+            meaning all columns will be mapped to JSON fields of the same name.
+        nested_delim: if nested JSON fields are desired, the field separator that is used for the field names
+            parameter, or None for no nesting (default). For instance, if a particular column should be mapped
+            to JSON field X nested inside field Y, the corresponding field name value for the column key
+            in the mapping dict can be the string "X.Y", in which case the value for nested_delim should be "."
```
chipkent, line 145: I can figure out the nesting stuff, I think, but a paragraph before args could be helpful.

pyintegration/tests/test_parquet.py
```diff
@@ -0,0 +1,146 @@
+#
```
chipkent, line 1: Just sticking this comment here...
Changes that don't seem to be tested:
1) `column.py`
2) `dtypes.py`
3) `stream/__init__.py`

chipkent (changes requested):

pyintegration/deephaven2/dtypes.py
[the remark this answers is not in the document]
jmao-denver: Good thing that you caught this 'inconsistency'. I wasn't very happy about it but in the spirit of getting the most commonly used Java data structures into dtypes module and having a consistent way among all the types to construct an instance of them, I came up with this design. I later found a better/more consistent way to do that in the pandas module and was planning to update this PR with it but completely forgot about it. Let's see if the new approach agrees with you more.

pyintegration/deephaven2/stream/kafka/consumer.py
```diff
@@ -0,0 +1,226 @@
...
+    def __init__(self, j_spec):
+        self._j_spec = j_spec
+
+    @property
+    def j_object(self) -> jpy.JType:
+        return self._j_spec
+
+
+KeyValueSpec.IGNORE = KeyValueSpec(_JKafkaTools_Consume.IGNORE)
+KeyValueSpec.NONE = KeyValueSpec(_JKafkaTools.FROM_PROPERTIES)
```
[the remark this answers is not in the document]
jmao-denver: I think using a no-arg function could be confusing here. The fact is these are truly constants. Enum could be appropriate but seems to be an overkill in this case.

pyintegration/deephaven2/dtypes.py
```diff
@@ -24,174 +27,161 @@ def _qst_custom_type(cls_name: str):
...
+string = DType(j_name="java.lang.String", qst_type=_JQstType.stringType(), np_type=np.object_)
+BigDecimal = DType(j_name="java.math.BigDecimal", np_type=np.object_)
+StringSet = DType(j_name="io.deephaven.stringset.StringSet", np_type=np.object_)
+DateTime = DType(j_name="io.deephaven.time.DateTime", np_type=np.dtype("datetime64[ns]"))
+Period = DType(j_name="io.deephaven.time.Period", np_type=np.object_)
+PyObject = DType(j_name="org.jpy.PyObject", np_type=np.object_)
+JObject = DType(j_name="java.lang.Object", np_type=np.object_)
+
+
+def array(dtype: DType, seq: Sequence, remap: Callable[[Any], Any] = None) -> jpy.JType:
```
chipkent, line 89: Do we need the ability to create an array that is not filled in?

pyintegration/deephaven2/dtypes.py
```diff
@@ -24,174 +27,161 @@ def _qst_custom_type(cls_name: str):
...
-    def __call__(self, s: Set):
-        if s is None:
-            return None
-        r = self.j_type()
-        for v in s:
-            r.add(v)
-        return r
+def j_array_list(values: Iterable):
+    if values is None:
+        return None
```
chipkent, line 150: Should this return `None` or an empty list?

pyintegration/deephaven2/dtypes.py
```diff
@@ -24,174 +27,161 @@ def _qst_custom_type(cls_name: str):
...
+
 
-    def __call__(self, s: Set):
-        if s is None:
-            return None
-        r = self.j_type()
-        for v in s:
-            r.add(v)
-        return r
+def j_array_list(values: Iterable):
```
chipkent, line 148: pydoc

pyintegration/deephaven2/dtypes.py
```diff
@@ -24,174 +27,161 @@ def _qst_custom_type(cls_name: str):
...
+        return None
+    r = jpy.get_type("java.util.ArrayList")(len(list(values)))
+    for v in values:
+        r.add(v)
+    return r
 
 
-class ArrayListDType(DType):
-    """ The case the wraps java Set. """
+def j_hashmap(d: Dict):
```
chipkent, line 157: pydoc

pyintegration/deephaven2/dtypes.py
```diff
@@ -24,174 +27,161 @@ def _qst_custom_type(cls_name: str):
...
+    if d is None:
+        return None
```
chipkent, line 159: `None` or an empty map?

pyintegration/deephaven2/dtypes.py
```diff
@@ -24,174 +27,161 @@ def _qst_custom_type(cls_name: str):
...
+            value = ''
+        r.put(key, value)
+    return r
 
-        j_list = self.j_type()
-        for v in it:
-            j_list.add(v)
-        return j_list
 
+def j_hashset(s: Set):
```
chipkent, line 169: pydoc

pyintegration/deephaven2/dtypes.py
```diff
@@ -24,174 +27,161 @@ def _qst_custom_type(cls_name: str):
...
+    if s is None:
+        return None
```
chipkent, line 171: `None` or an empty set?

pyintegration/deephaven2/dtypes.py
```diff
@@ -24,174 +27,161 @@ def _qst_custom_type(cls_name: str):
...
+    r = jpy.get_type("java.util.HashSet")()
+    for v in s:
+        r.add(v)
+    return r
 
 
-def is_java_type(obj: Any) -> bool:
-    """ Returns True if the object is originated in Java. """
-    return isinstance(obj, jpy.JType)
+def j_properties(d: Dict):
```
chipkent, line 179: pydoc

pyintegration/deephaven2/dtypes.py
```diff
@@ -24,174 +27,161 @@ def _qst_custom_type(cls_name: str):
...
+    if d is None:
+        return None
```
chipkent, line 181: `None` or empty properties?

pyintegration/deephaven2/dtypes.py
chipkent, line 148: Should line 148 and below go in the types function, or should there be a separate module for it? e.g. `jcompat`.  None of it refers to anything else in `dtypes`.

pyintegration/deephaven2/parquet.py
```diff
@@ -167,7 +169,7 @@ def write_tables(tables: List[Table], destinations: List[str], col_definitions:
         col_definitions (List[Column]): the column definitions to use
         col_instructions (List[ColumnInstruction]): instructions for customizations while writing
         compression_codec_name (str): the compression codec to use, if not specified, defaults to SNAPPY
-        max_dictionary_keys (int): the maximum dictionary keys allowed, if not specified, defaults to 2^20
+        max_dictionary_keys (int): the maximum dictionary keys allowed, if not specified, defaults to 2^20(1,048,576)
```
chipkent, line 172:
```suggestion
        max_dictionary_keys (int): the maximum dictionary keys allowed, if not specified, defaults to 2^20 (1,048,576)
```

chipkent (changes requested):

pyintegration/deephaven2/table.py
```diff
@@ -9,6 +9,7 @@
 import jpy
 
 from deephaven2 import DHError, dtypes
+from deephaven2._wrapper_abc import JObjectWrapper
```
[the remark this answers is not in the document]
jmao-denver: We have a todo item for Sphinx, this should pop up if Sphinx can't handle it.

pyintegration/deephaven2/dtypes.py
jmao-denver: Java code KafkaTools expects null

pyintegration/deephaven2/dtypes.py
jmao-denver: I thought on this for a while, couldn't find any use case

pyintegration/deephaven2/_jcompat.py
```diff
@@ -0,0 +1,59 @@
+#
+#   Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+#
+""" This module provides Java compatibility support including convenience functions to create some widely used Java
+data structures from corresponding Python ones in order to be able to call Deephaven Java methods. """
```
chipkent, line 5:
```suggestion
""" This module provides Java compatibility support including convenience functions to create some widely used Java
data structures from corresponding Python ones in order to be able to call Java methods. """
```

pyintegration/deephaven2/stream/kafka/producer.py
```diff
@@ -99,8 +101,8 @@ def avro_spec(schema: str, schema_version: str = "latest", field_to_col_mapping:
             schema generated from the table definition, for the columns and fields implied by field_to_col_mapping,
             include_only_columns, and exclude_columns; if a schema_version is provided and the resulting version after
             publishing does not match, an exception results. The default is False.
-        schema_namespace (str): when 'publish_schema' is True, the namespace for the generated schema to be registered in
-            Schema Registry Server.
+        schema_namespace (str): when 'publish_schema' is True, the namespace for the generated schema to be registered
+            in Schema Registry Server.
```
chipkent, line 105:
```suggestion
            in the Schema Registry Server.
```

pyintegration/deephaven2/stream/kafka/producer.py
```diff
@@ -139,12 +145,10 @@ def json_spec(include_columns: List[str] = None, exclude_columns: List[str] = No
         mapping (Dict[str, str]): a mapping from column names to JSON field names.  Any column name implied by earlier
             arguments and not included as a key in the map implies a field of the same name. default is None,
             meaning all columns will be mapped to JSON fields of the same name.
-        nested_delim: if nested JSON fields are desired, the field separator that is used for the field names
-            parameter, or None for no nesting (default). For instance, if a particular column should be mapped
-            to JSON field X nested inside field Y, the corresponding field name value for the column key
-            in the mapping dict can be the string "X.Y", in which case the value for nested_delim should be "."
-        output_nulls: when False (default), do not output a field for null column values
-        timestamp_field: the name of a publishing timestamp field to include, default is None
+        nested_delim (str): the filed separator for nested JSON fields
```
chipkent, line 148: I think my comment about nested delimiters was misinterpreted.  Here, you pulled most of the pydoc for `nested_delim` out of the variable area and moved it to the paragraph area.  I think that wording belongs in the variable area.  In the paragraph area, we need some high-level wording that explains the concept of nested Kafka json, so that the user can understand the wording for the `nested_delim` variable.

chipkent (changes requested):

chipkent (approved):

PR #1841, opened 2022-01-18
Bump in prep for end of month release of 0.9.0

### LINKED ISSUE #1808: The extended images should be tagged with `latest`
The extended images aren't tagged with `latest` (like `server` is). They should be.

devinrsmith:
Fixes #1808

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -55,7 +55,7 @@ jobs:
           tags: |
             type=edge,branch=main
             type=raw,priority=950,enable=${{ startsWith(github.ref, 'refs/heads/release/v') }},value=${{ steps.semver.outputs.semver }}
-            type=raw,enable=${{ startsWith(github.ref, 'refs/heads/release/v0.8') }},value=latest
+            type=raw,enable=${{ startsWith(github.ref, 'refs/heads/release/v0.9') }},value=latest
             type=ref,event=branch
 
       - name: Docker meta server nltk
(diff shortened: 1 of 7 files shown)


.github/workflows/build-ci.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           tags: |
             type=edge,branch=main
             type=raw,priority=950,enable=${{ startsWith(github.ref, 'refs/heads/release/v') }},value=${{ steps.semver.outputs.semver }}
-            type=raw,enable=${{ startsWith(github.ref, 'refs/heads/release/v0.7') }},value=latest
+            type=raw,enable=${{ startsWith(github.ref, 'refs/heads/release/v0.9') }},value=latest
```
mofojed, line 72: Some of these are switching from 0.7 to 0.9, while others are switching from 0.8. Is there a reason why? And no way to share a constant using a [matrix](https://docs.github.com/en/actions/learn-github-actions/contexts#matrix-context) or an [inputs](https://docs.github.com/en/actions/learn-github-actions/contexts#inputs-context) context?

jcferretti (approved):

.github/workflows/build-ci.yml
devinrsmith: I think it was due to the previous PR for these straddling a release bump. There may be some better ways to use github action outputs, or scripting, but nothing built into the yaml afaict. https://github.com/deephaven/deephaven-core/issues/1808

devinrsmith: Oh, maybe inputs would work. Haven't used them before.

devinrsmith: https://github.com/deephaven/deephaven-core/issues/1842

Issue #1775, opened 2021-12-30
multivariate jpy (unwrapped) return type into java


devinrsmith:
We're unable to proxy from java into python and return an unwrapped java object if the underlying item is a JType (python wrapper for java object), but otherwise return the PyObject. This would allow us to be more efficient with some operations that we can only do right now by repeatedly crossing the JNI boundary.

If you tell jpy you want an Object.class back, it *will* unwrap JType objects, but will fail on native python objects.

If applicable, we can have a specific target type for this use case `ObjectOrPyObject.class`; or allow more parameters into the jpy `call(...)` methods to say we are ok returning a PyObject.class for Object.class.

devinrsmith:
This is also applicable when python is calling into a java method.

devinrsmith:
`io.deephaven.engine.util.PythonScopeJpyImpl#convert` may likely benefit if we could solve this issue, and we could push the logic down into python.

PR #1843, opened 2022-01-18
A workaround to a JPY issue with long values, fixes #1461

### LINKED ISSUE #1461: [Python] lowerBin() computes incorrect result when used in a query
**Description**

The following python code is supposed to make a table with 2 columns, one containing a date-time and the other containing the date-time floored via the `lowerBin()` method, and print the equivalent result of `lowerBin()` to console.

```
from deephaven.DBTimeUtils import expressionToNanos, convertDateTime, lowerBin
from deephaven.TableTools import newTable, dateTimeCol

nanosBin = expressionToNanos("T15M")
dateTime = convertDateTime("2020-01-01T00:35:00 NY")

source = newTable(
    dateTimeCol("DateTimes", dateTime)
)

result = source.update("DateTimesBinned = lowerBin(DateTimes, nanosBin)")
print(lowerBin(dateTime, nanosBin))
```

However, the print statement and the table result in 2 different results

`2020-01-01T00:30:00.000000000 NY` vs `2020-01-01T00:34:59.016`

Screen Shot 2021-10-18 at 10 57 06 PM

**Steps to reproduce**

Described above

**Expected results**

The output of the print statement for `lowerBin()` should match the column computed by `lowerBin()`.

**Actual results**

The output of the print statement for `lowerBin()` appears correct, while the column computed by `lowerBin()` is not correct.

This does not happen in groovy. This may be related to https://github.com/deephaven/deephaven-core/issues/1348

**Versions**
 - Deephaven: latest
 - OS: OSX
 - Browser: Chrome
 - Docker: Docker version 20.10.8, build 3967b7d

jmao-denver:
In JPY, PyObject.getObjectValue() seems always to select Java Integer
for a Python int object  which is not restricted by the number
of bits in Python 3

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
@@ -192,7 +192,7 @@ public class PythonScopeJpyImpl implements PythonScope<PyObject> {
             Object objValue = pyObject.getObjectValue();
             // workaround a JPY issue with determining the correct Java type a Python 'int' should be
             // mapped to (int or long) without causing an implicit overflow
-            if (objValue instanceof Integer || objValue.getClass() == int.class) {
+            if (objValue instanceof Integer) {
                 long longValue;
                 if ((longValue = pyObject.getLongValue()) != (int) objValue) {
                     return longValue;
--- a/engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
@@ -194,7 +194,7 @@ public class PythonScopeJpyImpl implements PythonScope<PyObject> {
             // mapped to (int or long) without causing an implicit overflow
             if (objValue instanceof  Integer || objValue.getClass() == int.class) {
                 long longValue;
-                if ((longValue = pyObject.getLongValue()) > (int) objValue) {
+                if ((longValue = pyObject.getLongValue()) !=  (int) objValue) {
                     return longValue;
                 }
             }
--- a/engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
@@ -189,7 +189,16 @@ public class PythonScopeJpyImpl implements PythonScope<PyObject> {
         } else if (pyObject.isCallable()) {
             return wrapCallable(pyObject);
         } else if (pyObject.isConvertible()) {
-            return pyObject.getObjectValue();
+            Object objValue = pyObject.getObjectValue();
+            // workaround a JPY issue with determining the correct Java type a Python 'int' should be
+            // mapped to (int or long) without causing an implicit overflow
+            if (objValue instanceof  Integer || objValue.getClass() == int.class) {
+                long longValue;
+                if ((longValue = pyObject.getLongValue()) > (int) objValue) {
+                    return longValue;
+                }
+            }
+            return objValue;
         } else {
             return pyObject;
         }
--- a/engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
@@ -192,9 +192,9 @@ public class PythonScopeJpyImpl implements PythonScope<PyObject> {
             Object objValue = pyObject.getObjectValue();
             // workaround a JPY issue with determining the correct Java type a Python 'int' should be
             // mapped to (int or long) without causing an implicit overflow
-            if (objValue instanceof  Integer || objValue.getClass() == int.class) {
+            if (objValue instanceof Integer || objValue.getClass() == int.class) {
                 long longValue;
-                if ((longValue = pyObject.getLongValue()) !=  (int) objValue) {
+                if ((longValue = pyObject.getLongValue()) != (int) objValue) {
                     return longValue;
                 }
             }
--- a/Integrations/python/functional-tests/pylong_conv_test.py
+++ b/Integrations/python/functional-tests/pylong_conv_test.py
@@ -1,0 +1,22 @@
+#
+# Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+#
+import unittest
+from deephaven import TableTools
+import bootstrap
+
+long_value = 2 ** 32 + 5
+
+
+class TestClass(unittest.TestCase):
+
+    def test_long_number_conversion(self):
+        t = TableTools.emptyTable(1)
+        result = TableTools.string(t.update("X = long_value"), 1)
+        self.assertEqual(long_value, int(result.split()[2]))
+
+
+if __name__ == "__main__":
+    bootstrap.build_py_session()
+
+    unittest.main(verbosity=2)


jmao-denver:
Not sure if DHE suffers from the same problem

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
```diff
@@ -189,7 +189,16 @@ public static Object convert(PyObject pyObject) {
         } else if (pyObject.isCallable()) {
             return wrapCallable(pyObject);
         } else if (pyObject.isConvertible()) {
-            return pyObject.getObjectValue();
+            Object objValue = pyObject.getObjectValue();
+            // workaround a JPY issue with determining the correct Java type a Python 'int' should be
+            // mapped to (int or long) without causing an implicit overflow
+            if (objValue instanceof  Integer || objValue.getClass() == int.class) {
+                long longValue;
+                if ((longValue = pyObject.getLongValue()) > (int) objValue) {
```
devinrsmith, line 197: I think `!=` is more appropriate?

Note: we'll still have issues w/ python "ints" that really should be BigInteger (> 64 bits)

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
```diff
@@ -189,7 +189,16 @@ public static Object convert(PyObject pyObject) {
         } else if (pyObject.isCallable()) {
             return wrapCallable(pyObject);
         } else if (pyObject.isConvertible()) {
-            return pyObject.getObjectValue();
+            Object objValue = pyObject.getObjectValue();
+            // workaround a JPY issue with determining the correct Java type a Python 'int' should be
+            // mapped to (int or long) without causing an implicit overflow
+            if (objValue instanceof  Integer || objValue.getClass() == int.class) {
```
devinrsmith, line 195: Do we think the second case will ever trigger? I would expect that the instanceof is sufficient.

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
jmao-denver: != is definitely the correct way, covering both negative and positive values; and yes, we need to address the BigDecimal translation in the future.

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
jmao-denver: I am not 100% sure, and that's why I threw it in there just in case there are some edges cases that are not obvious to us now

cpwright:
Should we simply look further up the chain at JPy and fix that.  I expect the problem relates to these lines in 
JType_ConvertJavaToPythonObject.

    } else if (JPy_IS_CLONG(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Integer_JClass, type->classRef)))) {
        return JType_CreateJavaIntegerObject(jenv, type, pyArg, objectRef);
    } else if (JPy_IS_CLONG(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Long_JClass, type->classRef)))) {
        return JType_CreateJavaLongObject(jenv, type, pyArg, objectRef);

devinrsmith:
> Should we simply look further up the chain at JPy and fix that. I expect the problem relates to these lines in JType_ConvertJavaToPythonObject.
> 
> ```
> } else if (JPy_IS_CLONG(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Integer_JClass, type->classRef)))) {
>     return JType_CreateJavaIntegerObject(jenv, type, pyArg, objectRef);
> } else if (JPy_IS_CLONG(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Long_JClass, type->classRef)))) {
>     return JType_CreateJavaLongObject(jenv, type, pyArg, objectRef);
> ```

I made this same comment here https://github.com/deephaven/deephaven-core/issues/1461#issuecomment-1015624546. Ultimately, I think the implicit type conversion layer of python objects to to java objects needs to be examined, and we may want to provide the (java) caller with better hooks on how to handle the different cases.

These types of problems may also be better solved by pushing application logic further down into python instead of trying to be as general as we can w/ `PythonScope`. I've made some similar changes for FetchObject (landing soon) where script session changes are handled with .py logic instead of numerous JNI boundary crossings.

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
rcaudy: Why don't we use `PyObject.isLong()` or `PyObject.isInt()`?

devinrsmith: isLong / isInt are the same (python 3), but it won't actually tell you how many bits are in the python int.

devinrsmith: An alternative would be to do `PyObject.isLong()`, `PyObject.getLongValue()`, and downcast back to `int` if appropriate (or, choose to have `convert` never return an int and always return a long)

jmao-denver:
@devinrsmith @cpwright 

I think you meant JType_ConvertPythonToJavaObject (not JType_ConvertJavaToPythonObject).  

The straightforward changes I could see are:

1. to switch the order of the two blocks (or simply remove the one check assignability of Integer, which amounts to the alternative Devin mentioned about, and with that, potentially double the memory used for a column unnecessarily

2. to examine the actual value and select the appropriate Java type, which is almost the same (hacky) as the solution in this PR but with one fewer JNI/Python boundary crossing

I am not strongly inclined towards  any one of the 3 solutions, but this PR is still my favorite simply because it is higher level and less likely to break other things (?)

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
```diff
@@ -189,7 +189,16 @@ public static Object convert(PyObject pyObject) {
         } else if (pyObject.isCallable()) {
             return wrapCallable(pyObject);
         } else if (pyObject.isConvertible()) {
-            return pyObject.getObjectValue();
+            Object objValue = pyObject.getObjectValue();
+            // workaround a JPY issue with determining the correct Java type a Python 'int' should be
+            // mapped to (int or long) without causing an implicit overflow
+            if (objValue instanceof Integer || objValue.getClass() == int.class) {
```
devinrsmith, line 195: I'm not a fan of `.getClass()` check here, I don't think it adds anything of value.

Happy to approve without.

I think this convert logic could be pushed all down into python (and save us numerous JNI calls), potentially pending https://github.com/deephaven/deephaven-core/issues/1775. Obviously not part of this simple fix, but something we may want to do in the future. Going to add a note to the linked issue.

devinrsmith (changes requested):

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
rcaudy: I like that.

rcaudy: @cpwright do you like the above? Preference for always `long`, or `int` if it fits?

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
cpwright: objValue.getClass() being int.class seems impossible; as you would not have a class object.

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
cpwright: I would suggest int if it fits.

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
jmao-denver: thanks, removed!

devinrsmith (dismissed):
I'm OK with this approach. Based on @rcaudy feedback thought, he may prefer the `pyObject.isLong()`, `pyObject.getLongValue()`, and seeing if it fits into int. It's an extra case, but it fits the pattern established here.

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
```diff
@@ -189,7 +189,16 @@ public static Object convert(PyObject pyObject) {
         } else if (pyObject.isCallable()) {
             return wrapCallable(pyObject);
         } else if (pyObject.isConvertible()) {
-            return pyObject.getObjectValue();
+            Object objValue = pyObject.getObjectValue();
+            // workaround a JPY issue with determining the correct Java type a Python 'int' should be
+            // mapped to (int or long) without causing an implicit overflow
+            if (objValue instanceof Integer) {
+                long longValue;
+                if ((longValue = pyObject.getLongValue()) != (int) objValue) {
```
rcaudy, line 197: Please add a unit test for the fix.

rcaudy (dismissed):

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
jmao-denver: done

devinrsmith (approved):

PR #1854, opened 2022-01-20
Implemented viewport snapshot functionality over DoExchange

### LINKED ISSUE #925: (Viewport) Support DoExchange equivalent of ConstructSnapshot over Flight
It's a little silly, but Flight's DoGet does not allow a user to supply app_metadata. We can still support a DoGet w/column and viewport filters, but has to be tucked nicely inside of DoExchange.

lbooker42:
Closes #925


lbooker42:

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
```diff
@@ -0,0 +1,325 @@
...
+    private final CountDownLatch resultLatch = new CountDownLatch(1);
+    private Throwable exceptionWhileSealing = null;
+
+    private volatile boolean connected = true;
+
+    /**
+     * Represents a BarrageSnapshot.
+     *
+     * @param session the Deephaven session that this export belongs to
+     * @param tableHandle the tableHandle to snapshot (ownership is transferred to the snapshot)
```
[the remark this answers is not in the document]
lbooker42: @rcaudy Is this really gaining ownership of the server-side table?  What calls are should change to ensure this doesn't happen?

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSession.java
```diff
@@ -51,6 +51,19 @@ public BarrageSubscription subscribe(final TableHandle tableHandle, final Barrag
         return new BarrageSubscriptionImpl(this, tableHandle.newRef(), options);
     }
 
+    @Override
+    public BarrageSnapshot snapshot(final TableSpec tableSpec, final BarrageSubscriptionOptions options)
```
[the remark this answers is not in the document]
lbooker42: Should this be renamed to `BarrangeExhangeOptions` and remain shared between `BarrageSnapshotRequest` and `BarrageSubscriptionRequest` or should we clone this and have two structures that can diverge from each other as needed?

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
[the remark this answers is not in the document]
rcaudy: `io.deephaven.client.impl.BarrageSnapshotImpl#handleDisconnect`
That method probably needs to call `cleanup()` the same as the one in the subscription impl does.

server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
```diff
@@ -326,50 +338,123 @@ public void onNext(final InputStream request) {
...
 
-                    if (message.app_metadata.msgPayloadVector() == null) {
-                        throw GrpcUtil.statusRuntimeException(Code.INVALID_ARGUMENT,
-                                "Subscription request not supplied");
+                    // handle the different message types that can come over DoExchange
+                    switch (message.app_metadata.msgType()) {
+                        case BarrageMessageType.BarrageSubscriptionRequest:
+                            serviceBarrageSubscriptionRequest(message);
+                            break;
+                        case BarrageMessageType.DoGetRequest: // rename to SnapshotRequest?
```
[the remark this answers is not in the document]
niloc132: These names are defined in github.com/deephaven/barrage/

server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
```diff
@@ -326,50 +338,123 @@ public void onNext(final InputStream request) {
...
+                                        : null;
+
+                        // get ourselves some data!
+                        final BarrageMessage msg = ConstructSnapshot.constructBackplaneSnapshotInPositionSpace(this,
+                                table, columns, viewport);
+                        msg.modColumnData = new BarrageMessage.ModColumnData[0]; // actually no mod column data for
+
+                        // translate the viewport to keyspace and make the call
+                        try (final BarrageStreamGenerator bsg = new BarrageStreamGenerator(msg);
+                                final RowSet keySpaceViewport =
```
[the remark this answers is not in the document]
niloc132: Unless I very much misunderstand, it is flat when sent over the wire - this index is just needed for the getSubView to know what contents to copy to the flight message

server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
```diff
@@ -326,50 +338,123 @@ public void onNext(final InputStream request) {
...
+
+                        // leverage the stream generator SchemaView constructor
+                        final BarrageStreamGenerator.SchemaView schemaView =
+                                new BarrageStreamGenerator.SchemaView(serializedMessage);
+
+                        // push the schema to the listener
+                        listener.onNext(schemaView);
+
+                        // collect the viewport and columnsets (if provided)
+                        final boolean hasColumns = subscriptionRequest.columnsVector() != null;
```
niloc132: aside from the columns and rowset, can we share more of the implementation with customDoGet? That is, let customDoGet call into some shared code, with no columns/rows set?

this probably makes sense to do after splitting the subscribe/doget.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSession.java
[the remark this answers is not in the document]
lbooker42: Per discussion, cloning and renaming  to `BarrageSnapshotOptions`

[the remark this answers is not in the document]
lbooker42: This renaming needs to initially take place in the barrage companion repo before we can follow with changes in these files.

server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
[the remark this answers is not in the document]
rcaudy:
```
        final int rowsAddedOffset;
        if (isSnapshot && !view.isInitialSnapshot) {
            // client's don't need/want to receive the full RowSet on every snapshot
            rowsAddedOffset = EmptyIndexGenerator.INSTANCE.addToFlatBuffer(metadata);
        } else {
            rowsAddedOffset = rowsAdded.addToFlatBuffer(metadata);
        }

        final int rowsRemovedOffset = rowsRemoved.addToFlatBuffer(metadata);
        final int shiftDataOffset = shifted.addToFlatBuffer(metadata);

        // Added Chunk Data:
        int addedRowsIncludedOffset = 0;
        if (view.isViewport()) {
            addedRowsIncludedOffset = rowsIncluded.addToFlatBuffer(view.keyspaceViewport, metadata);
        }
```
This code shows how the DoGet handling is replacing `rowsAdded` with an empty RowSet (not flat, but empty, even more concise). That is, in `FlightServiceGrpcImpl` it is somewhat falsely specifying that the SubView is not an initial snapshot.
`rowsIncluded` is not getting the same treatment, because the DoGet use case doesn't specify viewports.
We likely want to do something similar for our new request. Maybe time to add a boolean option "don't send row sets"?
Or maybe we can skip metadata altogether?

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -251,6 +263,87 @@ public void forEachStream(Consumer<InputStream> visitor) throws IOException {
...
+            this.viewport = viewport;
+            this.keyspaceViewport = keyspaceViewport;
+            this.subscribedColumns = subscribedColumns;
+            this.hasModBatch = false;
+            this.hasAddBatch = generator.rowsIncluded.original.isNonempty();
+        }
+
+        @Override
+        public void forEachStream(Consumer<InputStream> visitor) throws IOException {
+            ByteBuffer metadata = generator.getSnapshotMetadata(this);
```
rcaudy, line 330: In our discussion the other day, I think we discussed getting rid of metadata on snapshots altogether. If we don't, we should at least narrow the snapshot metadata down to a meaningful subset of existing fields.

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -558,6 +649,56 @@ private ByteBuffer getMetadata(final SubView view) throws IOException {
         return header.dataBuffer().slice();
     }
 
+    private ByteBuffer getSnapshotMetadata(final SnapshotView view) throws IOException {
```
rcaudy, line 652: If we *do* get rid of this (or narrow it), we'll need changes to `BarrageTable` to supply empty values or flat row sets for the essential fields.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/BarrageSnapshotOptions.java
```diff
@@ -0,0 +1,100 @@
...
+    public ColumnConversionMode columnConversionMode() {
+        return ColumnConversionMode.Stringify;
+    }
+
+    public int appendTo(FlatBufferBuilder builder) {
+        return io.deephaven.barrage.flatbuf.BarrageSnapshotOptions.createBarrageSnapshotOptions(
+                builder, conversionModeEnumToFb(columnConversionMode()), useDeephavenNulls(), batchSize());
+    }
+
+    private static ColumnConversionMode conversionModeFbToEnum(final byte mode) {
```
rcaudy, line 64: Please migrate these methods from `BarrageSnapshotOptions` and `BarrageSubscriptionOptions` to `ColumnConversionMode` and make them package-private.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/ColumnConversionMode.java
```diff
@@ -0,0 +1,5 @@
+package io.deephaven.extensions.barrage;
+
+public enum ColumnConversionMode {
+    Stringify, JavaSerialization, ThrowError
```
rcaudy, line 4:
```suggestion
    // @formatter:offf
    Stringify,
    JavaSerialization,
    ThrowError
    // @formatter:on
```

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
```diff
@@ -0,0 +1,325 @@
...
+
+        this.logName = tableHandle.exportId().toString();
+        this.options = options;
+        this.tableHandle = tableHandle;
+
+        final TableDefinition tableDefinition = BarrageUtil.convertArrowSchema(tableHandle.response()).tableDef;
+        resultTable = BarrageTable.make(tableDefinition, false);
+        resultTable.addParentReference(this);
+
+        final MethodDescriptor<FlightData, BarrageMessage> snapshotDescriptor =
```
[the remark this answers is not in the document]
rcaudy: The more I think about it, the more I think it's worth the effort to use `doGet` here in order to make sure we don't break `doGet`. That, or we need `doGet`-specific test cases. Either way, that can be a separate issue we file.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/StreamReaderOptions.java
```diff
@@ -0,0 +1,25 @@
...
+import io.deephaven.annotations.SimpleStyle;
+import io.deephaven.extensions.barrage.BarrageSnapshotOptions;
+import io.deephaven.extensions.barrage.BarrageSubscriptionOptions;
+import io.deephaven.extensions.barrage.ColumnConversionMode;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@SimpleStyle
+public abstract class StreamReaderOptions {
+    public static StreamReaderOptions of(BarrageSnapshotOptions options) {
```
rcaudy, line 12: Let's make this interface just an interface, and have the Barrage*Options classes implement it?

server/src/main/java/io/deephaven/server/barrage/BarrageMessageProducer.java
```diff
@@ -158,8 +158,29 @@
          * @param subscribedColumns are the columns subscribed for this view
          * @return a MessageView filtered by the subscription properties that can be sent to that subscriber
          */
-        MessageView getSubView(Options options, boolean isInitialSnapshot, @Nullable RowSet viewport,
+        MessageView getSubView(BarrageSubscriptionOptions options, boolean isInitialSnapshot, @Nullable RowSet viewport,
                 @Nullable RowSet keyspaceViewport, BitSet subscribedColumns);
+
+        /**
+         * Obtain a Full-Subscription View of this StreamGenerator that can be sent to a single subscriber.
```
rcaudy, line 165: Both of these methods need a bit of javadoc and parameter cleanup to be snapshot not subscription.

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -267,18 +353,33 @@ public SchemaView(final ByteBuffer buffer) {
...
+        public RowSet keyspaceViewport() {
+            return null;
+        }
     }
 
     /**
      * Treats the visitor with FlightData InputStream's to fulfill a DoGet.
      */
-    public void forEachDoGetStream(final SubView view, final Consumer<InputStream> visitor) throws IOException {
+    public void forEachDoGetStream(final SnapshotView view, final Consumer<InputStream> visitor) throws IOException {
```
rcaudy, line 376: We can delete this method now.

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
rcaudy: This needs to be deleted, per Larry.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageStreamReader.java
```diff
@@ -151,8 +151,34 @@ public BarrageMessage safelyParseFrom(final StreamReaderOptions options,
...
+                        // set the column selector active by default
+                        msg.snapshotColumns.set(ci);
+                    }
+
+                    // no mod column data
+                    msg.modColumnData = new BarrageMessage.ModColumnData[0];
+
+                    // generate empty row sets
+                    msg.rowsRemoved = RowSetFactory.empty();
+                    msg.shifted = new RowSetShiftData.Builder().build();
```
rcaudy, line 177: There's an EMPTY singleton for this. (`io.deephaven.engine.rowset.RowSetShiftData#EMPTY`)

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageStreamReader.java
```diff
@@ -193,6 +219,11 @@ public BarrageMessage safelyParseFrom(final StreamReaderOptions options,
                         for (int ci = 0; ci < msg.addColumnData.length; ++ci) {
                             msg.addColumnData[ci].data = ChunkInputStreamGenerator.extractChunkFromInputStream(options,
                                     columnChunkTypes[ci], columnTypes[ci], fieldNodeIter, bufferInfoIter, ois);
+
+                            // test for zero-data columns and remove from snapshot columns bitset
+                            if (msg.addColumnData[ci].data.size() == 0 && msg.snapshotColumns != null) {
```
rcaudy, line 224: Does it matter if there are `0` rows?

lbooker42: I have tested this using options that return `0` rows and the correct behavior is obtained.

server/src/main/java/io/deephaven/server/barrage/BarrageMessageProducer.java
```diff
@@ -162,21 +162,21 @@ MessageView getSubView(BarrageSubscriptionOptions options, boolean isInitialSnap
...
          *
          * @param options serialization options for this specific view
          * @param viewport is the position-space viewport
          * @param keyspaceViewport is the key-space viewport
          * @param subscribedColumns are the columns subscribed for this view
-         * @return a MessageView filtered by the subscription properties that can be sent to that subscriber
+         * @return a MessageView filtered by the snapshot properties that can be sent to that requestor
          */
         MessageView getSnapshotView(BarrageSnapshotOptions options, @Nullable RowSet viewport,
                 @Nullable RowSet keyspaceViewport, BitSet subscribedColumns);
```
rcaudy, line 182: Rename `subscribedColumns`?

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -259,7 +259,7 @@ public boolean isViewport() {
         }
 
         public final StreamReaderOptions options() {
-            return StreamReaderOptions.of(this.options);
+            return this.options;
```
rcaudy, line 262: You can just return `options`.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/StreamReaderOptions.java
```diff
@@ -6,17 +6,7 @@
...
 public abstract class StreamReaderOptions {
-    public static StreamReaderOptions of(BarrageSnapshotOptions options) {
-        return ImmutableStreamReaderOptions.of(options.useDeephavenNulls(), options.columnConversionMode());
-    }
-
-    public static StreamReaderOptions of(BarrageSubscriptionOptions options) {
-        return ImmutableStreamReaderOptions.of(options.useDeephavenNulls(), options.columnConversionMode());
-    }
-
     @Value.Parameter
```
rcaudy, line 10: I think this should be an interface, and you can drop the remaining annotations from immutables.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/BarrageSubscriptionOptions.java
```diff
@@ -7,12 +7,13 @@
...
 import io.deephaven.annotations.BuildableStyle;
 import io.deephaven.barrage.flatbuf.BarrageSubscriptionRequest;
+import io.deephaven.extensions.barrage.util.StreamReaderOptions;
 import org.immutables.value.Value.Default;
 import org.immutables.value.Value.Immutable;
 
 @Immutable
 @BuildableStyle
-public abstract class BarrageSubscriptionOptions {
+public abstract class BarrageSubscriptionOptions extends StreamReaderOptions {
```
rcaudy, line 16: Annotate overridden methods with `Override`.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageStreamReader.java
lbooker42: OBE by refactoring `StreamReader` to include the expected columns

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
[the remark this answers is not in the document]
lbooker42: Issue #1921 created to log this issue

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/StreamReader.java
```diff
@@ -14,14 +15,15 @@
      * Converts an InputStream to a BarrageMessage in the context of the provided parameters.
      *
      * @param options the options related to parsing this message
+     * @param expectedColumns the columns expected to appear in the stream, null implies all columns
```
rcaudy, line 18: If and only if you have to edit this code again for some other reason, I would apply this minor cosmetic cleanup:
```
     * @param expectedColumns the columns expected to appear in the stream; null implies all columns
     * @param columnChunkTypes the types to use for each column chunk
     * @param columnTypes the actual type for the column
     * @param componentTypes the actual component type for the column
     * @param stream the input stream that holds the message to be parsed
     * @return a BarrageMessage filled out by the stream's payload
     */
    BarrageMessage safelyParseFrom(StreamReaderOptions options,
            BitSet expectedColumns,
            ChunkType[] columnChunkTypes,
            Class<?>[] columnTypes,
            Class<?>[] componentTypes,
            InputStream stream);
```

rcaudy (approved):
Approving, but please don't merge until you've verified with some end-to-end testing.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
```diff
@@ -147,6 +152,23 @@ public BarrageTable entireTable() throws InterruptedException {
 
     @Override
     public synchronized BarrageTable partialTable(RowSet viewport, BitSet columns) throws InterruptedException {
+        // notify user when connection has already been used and closed
+        if (prevUsed) {
+            throw new UnsupportedOperationException("Cannot create multiple snapshots on a single connection");
```
rcaudy, line 157: "connection" feels out of place here. Not worth delaying merge for, though.
```suggestion
            throw new UnsupportedOperationException("Snapshot object already used");
```

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
```diff
@@ -46,14 +48,17 @@
     private final ClientCallStreamObserver<FlightData> observer;
 
     private final BarrageTable resultTable;
+    private volatile Condition sealedCondition;
```
rcaudy, line 51: I might put this directly above `sealed`. Fit better with that, organizationally.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
```diff
@@ -134,14 +134,19 @@ public void onCompleted() {
 
     @Override
     public synchronized BarrageTable entireTable() {
```
rcaudy, line 136: No need to mark `entireTable` as `synchronized`. Harmless, however.

server/src/main/java/io/deephaven/server/arrow/Demo.groovy
```diff
@@ -0,0 +1,6 @@
+import io.deephaven.server.uri.BarrageTableResolver;
```
rcaudy, line 1: This file looks like something you intended to finish, or delete?

server/src/main/java/io/deephaven/server/uri/BarrageTableResolver.java
```diff
@@ -134,6 +132,101 @@ public Table subscribe(DeephavenTarget target, TableSpec table, BarrageSubscript
         return sub.entireTable();
     }
 
+    /**
+     * Create a partial table subscription to the {@code table} via the {@code targetUri}. Uses {@link #SNAP_OPTIONS}.
```
rcaudy, line 136:
```suggestion
     * Create a partial table subscription to the {@code table} via the {@code targetUri}. Uses {@link #SUB_OPTIONS}.
```

server/src/main/java/io/deephaven/server/uri/BarrageTableResolver.java
```diff
@@ -134,6 +132,101 @@ public Table subscribe(DeephavenTarget target, TableSpec table, BarrageSubscript
...
+    public Table subscribe(DeephavenTarget target, TableSpec table, BarrageSubscriptionOptions options, RowSet viewport,
+            BitSet columns)
+            throws TableHandleException, InterruptedException {
+        final BarrageSession session = session(target);
+        final BarrageSubscription sub = session.subscribe(table, options);
+        return sub.partialTable(viewport, columns);
+    }
+
+    /**
+     * Create a full snapshot of the remote URI. Uses {@link #SUB_OPTIONS}.
```
rcaudy, line 164:
```suggestion
     * Create a full snapshot of the remote URI. Uses {@link #SNAP_OPTIONS}.
```

rcaudy (approved):

PR #1856, opened 2022-01-21
Full version of derived tables for debezium demo.

jcferretti:
The original SQL definitions for the tables in `demo.py` are here:

  https://github.com/MaterializeInc/ecommerce-demo/blob/main/README_RPM.md

Starting on point 7 (search for "define all of the tables in mysql.shop as Kafka sources").

### DIFF
--- a/debezium-demo/demo.py
+++ b/debezium-demo/demo.py
@@ -4,6 +4,7 @@ import deephaven.ProduceKafka as pk
 import deephaven.Types as dh
 from deephaven import Aggregation as agg, as_list
 from deephaven import PythonFunction as pyfun
+import deephaven.TableManipulation.WindowCheck as wck
 import jpy
 
 def tmapfun(x):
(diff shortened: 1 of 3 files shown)


debezium-demo/demo.py
```diff
@@ -0,0 +1,163 @@
...
+# Once deephaven-core PR #1843 is merged, we can remove the need for explicitly using L below
+# by defining a minute_in_nanos python variable, which would be interpolated with the right type,
+# and replacing `60L*second_in_nanos` by `minute_in_nanos`.
+profile_views_per_minute_last_10 = pageviews_stg \
+    .updateView(
+        'received_at_nanos = nanos(received_at)',
+        'received_at_minutes = received_at_nanos - received_at_nanos % minute_in_nanos'
+    ).where(
+        'pageview_type = `profiles`',
+        'nanos(DateTime.now()) < received_at_nanos + 10*minute_in_nanos'
```
cpwright, line 95: If you want to have things that actually update with time, you need to use window check; otherwise this is not a generally reliable way to use the engine, because now() can change out form under it.

debezium-demo/demo.py
```diff
@@ -0,0 +1,163 @@
...
+
+dd_flagged_profile_view = dd_flagged_profiles \
+    .naturalJoin(pageviews_stg, 'user_id')
+
+jbigd_type = jpy.get_type('java.math.BigDecimal')
+def mult(jbigd_value, long_value):
+    return jbigd_value.multiply(jbigd_type(long_value))
+
+high_value_users = purchases \
+    .updateView('purchase_total = (java.math.BigDecimal) mult(purchase_price, quantity)') \
```
cpwright, line 140: you have a python function with boundary crossing, why don't you just trust purchase_price.multiply(java.math.BigDecimal.valueOf(quantity)) which would also eliminate the necessity of the return cast because Python is hiding our type.

jcferretti:
Part of #1753 (needed for published tables in demo)

cpwright (approved):

jcferretti:
Part of #1753

Issue #1858, opened 2022-01-21
Better error handling for pyclient around release


devinrsmith:
There was a case where the pyclient integration tests caused a bunch of server side errors due to malformed release request, but it didn't effect the client because it was ignoring the errors:

```python
    def __del__(self):
        try:
            # only table objects that are explicitly exported have schema info and only the tickets associated with
            # such tables should be released.
            if self.ticket and self.schema and self.session.is_alive:
                self.session.release(self.ticket)
        except Exception as e:
            # TODO: log the exception
            ...
```

```
io.grpc.StatusRuntimeException: FAILED_PRECONDITION: Could not resolve ticket 'id': ticket was not provided                                                        
        at io.grpc.Status.asRuntimeException(Status.java:535)                                                                                                                                                                                            
        at io.grpc.protobuf.StatusProto.toStatusRuntimeException(StatusProto.java:52)                                                                                                                                                                    
        at io.deephaven.proto.util.Exceptions.statusRuntimeException(Exceptions.java:15)                                                                                                                                                                 
        at io.deephaven.proto.util.ExportTicketHelper.ticketToExportIdInternal(ExportTicketHelper.java:193)                                                                                                                                              
        at io.deephaven.proto.util.ExportTicketHelper.ticketToExportId(ExportTicketHelper.java:55)                                                                                                                                                       
        at io.deephaven.server.session.SessionState.getExportIfExists(SessionState.java:291)                                                                                                                                                             
        at io.deephaven.server.session.SessionServiceGrpcImpl.lambda$release$3(SessionServiceGrpcImpl.java:153)                                                                                                                                          
        at io.deephaven.extensions.barrage.util.GrpcUtil.rpcWrapper(GrpcUtil.java:38)                                                                                                                                                                    
        at io.deephaven.server.session.SessionServiceGrpcImpl.release(SessionServiceGrpcImpl.java:146)         
```

It turned out that pyclient was not handling the Batch, ExportedTableCreationResponse.result_id correctly, always assuming the ticket case, and not the batch_offset case. (Will have link to fix here.)

We should prefer to handle exceptional cases we expect, but should ensure we propagate exceptions we don't expect (ie, "ticket was not provided" should result in a hard error client side).

PR #1859, opened 2022-01-21
Fix batch export response handling pyclient

devinrsmith:
Follow up #1858


jmao-denver (approved):
LGTM, the error handling improvement might better be done as part of an overall better logging/configuration support, e.g. instead of swallowing the exception, log a warning to the log file and the terminal, ...

Issue #1860, opened 2022-01-21
Update GWT to 2.9


mofojed:
As a developer, I want gwt updated so that I don't run into compiler issues like https://github.com/google/elemental2/issues/28

niloc132:
https://github.com/deephaven/deephaven-core/pull/1888/files#r793089934 should support easy decoding of bytes as utf8 into a  string.

PR #1888, opened 2022-01-25
JSAPI for fetching custom object

### LINKED ISSUE #1874: JS API to fetch custom object from Deephaven plugin framework

mofojed:
Methods for fetching the JsWidget for custom types. Tested by installing deephaven-plugin and deephaven-plugin-matplotlib, using web-client-ui server-widgets branch that I've created, and running the following code snippets:

```
import matplotlib.pyplot as plt

x = [0, 2, 4, 6]
y = [1, 3, 4, 8]
plt.plot(x, y)
plt.xlabel('x values')
plt.ylabel('y values')
plt.title('plotted x and y values')
plt.legend(['line 1'])

m_figure=plt.gcf()
```

```
from deephaven.plugin.json import Node

t1 = emptyTable(10).update("A=i")
t2 = emptyTable(100).update("B=2*i")
t3 = emptyTable(1000).update("C=i*i")

z = Node({'x': 1, 't1': t1, 't2': t2, 't3': t3})
```

Fixes #1874

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
@@ -830,7 +830,10 @@ public class WorkerConnection {
         return whenServerReady("get a widget")
                 .then(server -> Callbacks.<FetchObjectResponse, Object>grpcUnaryPromise(c -> {
                     FetchObjectRequest request = new FetchObjectRequest();
-                    request.setSourceId(TableTicket.createTicket(varDef));
+                    TypedTicket typedTicket = new TypedTicket();
+                    typedTicket.setTicket(TableTicket.createTicket(varDef));
+                    typedTicket.setType(varDef.getType());
+                    request.setSourceId(typedTicket);
                     objectServiceClient().fetchObject(request, metadata(), c::apply);
                 })).then(response -> Promise.resolve(new JsWidget(this, response)));
     }
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
@@ -73,6 +73,7 @@ import io.deephaven.web.client.api.parse.JsDataHandler;
 import io.deephaven.web.client.api.state.StateCache;
 import io.deephaven.web.client.api.tree.JsTreeTable;
 import io.deephaven.web.client.api.widget.plot.JsFigure;
+import io.deephaven.web.client.api.widget.JsWidget;
 import io.deephaven.web.client.fu.JsItr;
 import io.deephaven.web.client.fu.JsLog;
 import io.deephaven.web.client.fu.LazyPromise;
@@ -82,7 +83,6 @@ import io.deephaven.web.client.state.TableReviver;
 import io.deephaven.web.shared.data.*;
 import io.deephaven.web.shared.fu.JsConsumer;
 import io.deephaven.web.shared.fu.JsRunnable;
-import io.deephaven.web.shared.ide.VariableType;
 import jsinterop.annotations.JsMethod;
 import jsinterop.annotations.JsOptional;
 import jsinterop.base.Js;
@@ -704,7 +704,7 @@ public class WorkerConnection {
         } else if ("Figure".equals(definition.getType())) {
             return (Promise) getFigure(definition);
         } else {
-            return (Promise) getPluginObject(definition);
+            return (Promise) getWidget(definition);
         }
     }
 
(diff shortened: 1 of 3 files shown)


web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidget.java
```diff
@@ -0,0 +1,46 @@
...
+        this.response = response;
+    }
+
+    @JsProperty
+    public String getType() {
+        return response.getType();
+    }
+
+    @JsMethod
+    public String getDataAsBase64() {
```
niloc132, line 27: for json at least, i'd still vote a getDataAsUtf8 or something like that? just to do it internally nicely, rather than have to read the bytes and decode externally

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidget.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @JsProperty
+    public JsWidgetExportedObject[] getExportedObjects() {
+        if (this.exportedObjects == null) {
+            this.exportedObjects = response.getTypedExportIdList().asList().stream()
+                    .map(ticket -> new JsWidgetExportedObject(connection, ticket))
+                    .toArray(JsWidgetExportedObject[]::new);
+        }
+
+        return this.exportedObjects;
```
niloc132, line 44: `this.` is implicit in java (except when ambiguous, like on the constructor), here and earlier in this method

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @JsProperty
+    String getType() {
+        return ticket.getType();
+    }
+
+    @JsMethod
+    Promise<Object> fetch() {
+        if (!getType().equals("Table")) {
+            // TODO (deephaven-core#62) implement fetch for tablemaps
```
niloc132, line 31: seems like non-tables could be fetched _in this case_ as objects, just call connection.getObject(...)? We have the full typed ticket, we can make the var def (or make a util method on WorkerConnection to handle an existing typed ticket).

niloc132 (approved):

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidget.java
mofojed: My JS is showing

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
mofojed: Not sure we could? We don't have the name or id of the object I don't think?

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidget.java
mofojed: TextDecoder isn't in current version of GWT, I think need to wait for #1860 to merge first?

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -698,22 +700,12 @@ public void onError(Throwable throwable) {
...
-            // case OtherWidget:
-            // return (Promise) getWidget(definition.getName());
-            // case TableMap:
-            // return (Promise) getTableMap(definition.getName());
-            default:
-                return Promise.reject(
-                        new Error("Object " + definition.getTitle() + " unknown type " + definition.getType() + "."));
+        if (definition.getType().equals(VariableType.Table.toString())) {
+            return (Promise) getTable(definition, null);
+        } else if (definition.getType().equals(VariableType.Figure.toString())) {
```
niloc132, line 705: Can we use JsVariableChanges.FIGURE instead, so we know the string is the same that JS might have passed in?

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidget.java
niloc132: Yeah let's wait, note added.

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
mofojed: Yup makes sense

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
mofojed: I'm dumb, should be good now.

niloc132 (approved):

Issue #1774, opened 2021-12-30
#1753 Debezium Kafka Demo Derived Tables to DHC


jcferretti:
Port the Business Intelligence real time queries.

Under #1753
Depends on #1773

Issue #1226, opened 2021-09-08
Kafka: support logicalType decimal


devinrsmith:
It would be very convenient to handle logical decimals, https://avro.apache.org/docs/current/spec.html#Decimal

Part of #293 ?

```
    {
      "name": "bid",
      "type": {"type": "bytes", "logicalType": "decimal", "precision": 16, "scale": 8}
    }
```

PR #1894, opened 2022-01-27
Support publishing BigDecimal columns to Kafka Avro.

jcferretti:
Also update `debezium-demo/demo.py` for it.  With this the coverage of the original SQL/materialize script is complete.

Example of publishing a static table:
image

Example of publishing a refreshing table (from last 2 statements in `debezium-demo/demo.py`)
image

### DIFF
--- a/extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
+++ b/extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
@@ -27,6 +27,7 @@ import io.deephaven.engine.table.impl.LocalTableMap;
 import io.deephaven.engine.table.impl.StreamTableTools;
 import io.deephaven.engine.table.TableMap;
 import io.deephaven.engine.table.TransformableTableMap;
+import io.deephaven.engine.util.BigDecimalUtils;
 import io.deephaven.internal.log.LoggerFactory;
 import io.deephaven.io.logger.Logger;
 import io.deephaven.kafka.ingest.*;
@@ -192,13 +193,16 @@ public class KafkaTools {
     }
 
     public static Schema columnDefinitionsToAvroSchema(
-            final ColumnDefinition<?>[] colDefs,
+            final Table t,
             final String schemaName,
             final String namespace,
             final Properties colProps,
             final Predicate<String> includeOnly,
-            final Predicate<String> exclude) {
+            final Predicate<String> exclude,
+            final MutableObject<Properties> colPropsOut) {
         SchemaBuilder.FieldAssembler<Schema> fass = SchemaBuilder.record(schemaName).namespace(namespace).fields();
+        final ColumnDefinition<?>[] colDefs = t.getDefinition().getColumns();
+        colPropsOut.setValue(colProps);
         for (final ColumnDefinition<?> colDef : colDefs) {
             if (includeOnly != null && !includeOnly.test(colDef.getName())) {
                 continue;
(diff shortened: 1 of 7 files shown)


jcferretti:
Part of #1753

jcferretti:
Fixes #1774, #1789

jcferretti:
Fixes #1226

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -236,24 +243,59 @@ public static Schema columnDefinitionsToAvroSchema(
...
-                precision = colProps.getProperty(colName + colNameToPropSeparator + precisionName);
-                scale = colProps.getProperty(colName + colNameToPropSeparator + scaleName);
-            }
-            if (precision == null || scale == null) {
-                throw new IllegalArgumentException(
-                        "When " + BigDecimal.class.getSimpleName() + " type columns are included, " +
-                                " column properties '" + precisionName + "' and '" + scaleName
-                                + "' for the given columns should be provided.");
+            final BigDecimalUtils.PrecisionAndScalePropertyNames propertyNames =
+                    new BigDecimalUtils.PrecisionAndScalePropertyNames(colName);
```
cpwright, line 247: This seems a little over teched rather than just telling the following function the name of the properties.  Or if you are going to have that; then the sets on 291 should also take the object instead of just doing it by hand.

jcferretti: The issue is handling errors and giving the user meaningful exception text, while avoiding the repetition in the code for variable names and values of "precision" and "scale" in the two places this is used from.  I converged to this from having all this done manually so that I have a place to put the property names so that they can be used for the exception strings etc.  I would be happy to entertain a simpler way, I don't love it, but it is better than what I had before.

I don't understand you comment about line 291; we have two objects, one that contains the property names, and one that contains values.  You think we should have a function that takes those two and a properties object and set those?  The function would do exactly what is done in line 291, and it would be used from just one place (everything else in BigDecimalUtils is used from two places).

cpwright: What I am referring to is that on 291 you have all this manual code:
                    precision = precisionAndScale.precision;
                    scale = precisionAndScale.scale;
                    final Properties toSet;
                    if (colProps == null) {
                        toSet = new Properties();
                        colPropsMu.setValue(toSet);
                    } else {
                        toSet = colProps;
                    }
                    toSet.setProperty(propertyNames.precisionProperty, Integer.toString(precision));
                    toSet.setProperty(propertyNames.scaleProperty, Integer.toString(scale));

It seems like if you decided there should be a function to extract things from a property object, you should also have a function to set those from the property object.

It definitely feels like you have this division between your property name object, the holder object, and this code.  But then they are intertwined anyway.

jcferretti: Yeah I see your point, plus the naming I don't think is the most happy.  Is hard for me to strike a balance for clarity without falling over to the "overly verbose (and lots of helper classes)" side.

I reworked the code, I think you will find it a better compromise.

cpwright (dismissed):

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
jcferretti: Charles and I discussed over slack, he considers this to have gone above the good enough threshold, so I am resolving this comment.

cpwright (approved):

PR #1896, opened 2022-01-27
Make the quick check fail fast

devinrsmith:
Removes server:check from the quick check (was taking 20 minutes), and the quick ci task will error out immediately instead of `--continue`.


cpwright (approved):

devinrsmith:
Closed in favor of https://github.com/deephaven/deephaven-core/pull/1897

PR #1899, opened 2022-01-27
Quality of life improvements for formula handling around BigDecimal columns.

jcferretti:
Deephaven - Google Chrome_143


engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageFunctionUtils.java
```diff
@@ -19428,4 +19430,436 @@ public static int negate(short a) {
...
+
+    public static BigDecimal multiply(BigDecimal a, long b) {
+        if (a == null || b == QueryConstants.NULL_LONG) {
+            return null;
+        }
+        return a.multiply(BigDecimal.valueOf(b));
+    }
+
+    public static BigDecimal multiply(long a, BigDecimal b) {
+        if (a == QueryConstants.NULL_LONG || b == null) {
```
cpwright, line 19498: should you really repeat checks if you are just going to use the commutative operator?

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageFunctionUtils.java
```diff
@@ -19428,4 +19430,436 @@ public static int negate(short a) {
...
+
+        if (b == QueryConstants.NULL_DOUBLE) {
+            return 1;
+        }
+
+        return a.compareTo(BigDecimal.valueOf(b));
+    }
+
+    public static int compareTo(double a, BigDecimal b) {
+        if (a == QueryConstants.NULL_DOUBLE) {
```
cpwright, line 19703: Deephaven normally has special comparisons for Double.NaN.

cpwright (changes requested):
This all needs unit test coverage; we can't just merge it unless we are sure that we cover all these cases from an update or where statement.

cpwright (commented):
So I think you did a good job covering the functions.  There is a question as to whether it makes sense to verify a few uses in a formula.  I know you have it in the demo, but a handful of unit tests (less exhaustive) is still probably worth it to ensure we don't break e.g. the parser's handling of these things.

cpwright (dismissed):

cpwright (approved):

jcferretti:
Part of #1753

PR #1907, opened 2022-01-31
Remove unused gradle.properties options

devinrsmith:


rcaudy (approved):

devinrsmith:
Alternatively, should we have a single entrypoint into our `npm pack` that specifies code-studio and embed-grid?

Issue #187, opened 2021-03-23
Remove openapi websocket classes


niloc132:
With the rewrite to gRPC, we no longer need web-shared-beans and web-shared-socket, but these classes are still used from a few places. Any code that uses web-shared-socket no longer works, since the websocket itself was removed.

- [x]  c# is built from this, should all be deleted until the API is solidified in enterprise, and the gRPC api is solidified in open source. This doesn't work presently anyway.
- [x]  jupyter probably also makes sense to be deleted since it uses websockets not only for openapi but also for forwarding sockets, and that isn't rebuilt to work in envoy yet either.
- [ ]  js api uses the web-shared-beans classes to manage state, and to simplify updating to gRPC many protobuf objects are being copied to the old beans before being exposed to JS.

PR #1929, opened 2022-02-02
Publish all jars

### LINKED ISSUE #1939: Publish DH jars
Currently, there are a handful of jars that we opt into publishing. Namely, those needed for the java client.

As part of #1938, we need to publish any jars that are relevant for building and running a DH server from the perspective of a 3rd party.

It will likely be best to adopt an "opt-out" approach instead of an "opt-in" approach. Alternatively, we may want to adopt a stricter approach where each subproject must be explicit (and error if a new subproject doesn't specify).

It could be good to have some sort of sanity check as well, where we can examine the dependencies for server-netty/server-jetty, and ensure that all of those projects are set for publishing (likely a new ticket, but TBD on how hard it would be). It may be easier (and desirable from other perspectives) to maintain a separate repository and CI that consumes the latest published jars for an example of how to build your own server.

There may be significant work that needs to be done around javadocs for each project; lots of errors in trying to do a first pass at javadocs. https://central.sonatype.org/publish/requirements/

devinrsmith:
Publish DH jars

Includes a large refactoring, pushing root build.gradle logic into plugin conventions.

Each gradle project must declare a single io.deephaven.project.X plugin type.

Created combined-javadoc project (with configuration combinedJavadoc as output artifact).

Better lifecycle management of 'quick' as a subset of 'check'.

Fixes #1939


devinrsmith:
To audit the publication:

```shell
$ ./gradlew publishToMavenLocal -x sign
$ find ~/.m2/repository/io/deephaven -type f -wholename "**/0.10.0/**" -name "*.pom"
```

buildSrc/src/main/groovy/io.deephaven.java-classpath-conventions.gradle
```diff
@@ -1,4 +1,20 @@
...
+    depAllocation = 'com.google.code.java-allocation-instrumenter:java-allocation-instrumenter:2.0'
+    depCommonsCodec = 'commons-codec:commons-codec:1.9'
+    depCommonsCompress = 'org.apache.commons:commons-compress:1.8'
+    depCommonsLang3 = 'org.apache.commons:commons-lang3:3.9'
+    depCommonsNet = 'commons-net:commons-net:3.2'
+    depCommonsEmail = 'org.apache.commons:commons-email:1.4'
+    depCommonsIo = 'commons-io:commons-io:2.5'
+    depJdom2 = 'org.jdom:jdom2:2.0.6'
+    depTrove3 = 'net.sf.trove4j:trove4j:3.0.3'
+}
```
JamesXNelson, line 17: this is still really old-school / shady.  `java-platform` plugin is better, imho.

buildSrc/src/main/groovy/io.deephaven.java-common-conventions.gradle
```diff
@@ -0,0 +1,35 @@
...
+    c.dependencies.all({
+        Dependency dep ->
+            if (dep.version && dep.version.endsWith('+')) {
+                throw new GradleException("Dynamic versions not allowed ($dep found in $path)")
+            }
+    })
+    c.resolutionStrategy {
+        // by default, we don't use any changing modules.
+        // however, if a developer is working on something where jars
+        // are published to a local filesystem (like mavenLocal()),
```
JamesXNelson, line 25: I think I was wrong on this original comment: filesystem repos are _always_ "rechecked" b/c they are just files on disk.

buildSrc/src/main/groovy/io.deephaven.java-jar-conventions.gradle
```diff
@@ -0,0 +1,17 @@
...
+    manifest {
+        attributes 'Implementation-Title': 'Deephaven Database',
+                'Implementation-Version': project.version,
+                'Provider': 'gradle'
+    }
+}
+
+tasks.withType(Jar).configureEach {
+    Jar jar ->
+        jar.preserveFileTimestamps = false
```
JamesXNelson, line 15: Maybe a comment about how this makes the timestamps all ~epoch, so that jar files are binary-stable regardless of when they were built / edited / etc.

buildSrc/src/main/groovy/io/deephaven/project/Type.groovy
```diff
@@ -0,0 +1,89 @@
...
+import groovy.transform.CompileStatic
+import org.gradle.api.Project
+import org.gradle.api.Task
+import org.gradle.api.artifacts.ProjectDependency
+import org.gradle.api.plugins.JavaPlugin
+import org.gradle.api.tasks.TaskProvider
+import org.gradle.language.base.plugins.LifecycleBasePlugin
+
+@CompileStatic
+class Type {
```
JamesXNelson, line 12: I think `Type` is one of the least descriptive name possible...  what _type_ of Type is it? ^-^

JamesXNelson (commented):
In general, I very much approve of these changes, and wish we did the same thing in enterprise.

buildSrc/src/main/groovy/io.deephaven.java-classpath-conventions.gradle
devinrsmith: https://github.com/deephaven/deephaven-core/issues/1976

buildSrc/src/main/groovy/io.deephaven.java-common-conventions.gradle
devinrsmith: Does this mean that resolutionStrategy closure should be deleted?

buildSrc/src/main/groovy/io.deephaven.java-classpath-conventions.gradle
niloc132: https://github.com/deephaven/deephaven-core/issues/1162

buildSrc/src/main/groovy/io.deephaven.java-jacoco-conventions.gradle
```diff
@@ -0,0 +1,21 @@
...
+jacoco {
+    toolVersion = '0.8.7'
+}
+
+jacocoTestReport {
+    reports {
+        xml.enabled true
+        csv.enabled true
+        html.enabled true
+        //            html.destination "${buildDir}/jacocoHtml"
```
niloc132, line 15: if not used, remove?

buildSrc/src/main/groovy/io.deephaven.java-test-conventions.gradle
```diff
@@ -0,0 +1,88 @@
...
+        if (!maxHeapSize) {
+            maxHeapSize = '3g'
+        }
+
+        if (!maxParallelForks) {
+            maxParallelForks = 4
+        }
+
+        if (!forkEvery) {
+            // NB: This comment may be outdated:
```
niloc132, line 48: worth removing rather than copy/pasting if we aren't sure it is accurate?

buildSrc/src/main/groovy/io.deephaven.java-test-conventions.gradle
```diff
@@ -0,0 +1,88 @@
...
+        useJUnit()
+
+        new File("$rootDir/tmp/workspace".toString()).mkdirs()
+        new File("$rootDir/tmp/logs".toString()).mkdirs()
+
+        // pass -PforceTest=true to run test again after they've passed once,
+        // and input source has not changed (i.e. invalidate caching)
+        if (findProperty('forceTest') == 'true') {
+            // not gonna stay uptodate very well like this: :-)
+            inputs.property("forceTest", UUID.randomUUID().toString())
```
niloc132, line 35: dumb question, but isn't `outputs.upToDateWhen { -> false }` more concise? Either optionally add that lambda, or read forceTest in there, rather than create a garbage value?

buildSrc/src/main/groovy/io.deephaven.java-test-conventions.gradle
```diff
@@ -0,0 +1,88 @@
...
+        if (findProperty('shortTests') == 'true') {
+            systemProperty 'TstUtils.shortTests', 'true'
+        }
+        systemProperty 'Configuration.rootFile', 'dh-tests.prop'
+        systemProperty 'devroot', "$rootDir"
+        systemProperty 'workspace', "$rootDir/tmp/workspace"
+        systemProperty 'configuration.quiet', 'true'
+        systemProperty 'java.awt.headless', 'true'
+
+        //testLogging {
```
niloc132, line 64: if not used, remove?

combined-javadoc/build.gradle
```diff
@@ -0,0 +1,67 @@
...
+        jdoc.javadocTool = javaToolchains.javadocToolFor{it ->
+            // Javadoc version >=11 is needed for search
+            // Javadoc version >12 is needed to avoid javadoc bugs in linking with modules
+            languageVersion = JavaLanguageVersion.of(15)
+        }
+        jdoc.inputs.file javaDocOverviewLocation
+        jdoc.options.encoding = 'UTF-8'
+        jdoc.options.tags = ['apiNote', 'implNote']
+        // include a note on the front Javadoc page telling us what version the Javadoc was generated from.
+//        jdoc.options.overview = new File("$rootDir", javaDocOverviewLocation)
```
niloc132, line 38: remove i think?

buildSrc/src/main/groovy/io.deephaven.java-test-conventions.gradle
devinrsmith: I'm running a test to see if this hack is still necessary - https://github.com/devinrsmith/deephaven-core/actions/runs/1849663443 . If these tests fail, or a lot longer than normal to run, I'd say the comment is valid and we need to leave the comment/logic (likely, follow up w/ a ticketed issue).

buildSrc/src/main/groovy/io.deephaven.java-test-conventions.gradle
devinrsmith: Sure, happy to adopt that.

buildSrc/src/main/groovy/io.deephaven.lifecycle-conventions.gradle
```diff
@@ -0,0 +1,8 @@
+def quickTask = project.tasks.register('quick') {
+    group = 'Deephaven lifecycle'
```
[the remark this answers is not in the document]
devinrsmith: I see that we use `~Deephaven` only in `Integrations:generateAll` - as far as I can tell, nothing calls that explicitly (maybe it's useful for manual purposes?).

That said, `generateAll` isn't a "Lifecycle" task - it *is* a grouping task, but only applicable for Integrations. As such, I don't think re-using either of the existing group names in the other place makes sense.

buildSrc/src/main/groovy/io.deephaven.java-test-conventions.gradle
devinrsmith: Sure, will clean that up.

combined-javadoc/build.gradle
```diff
@@ -0,0 +1,66 @@
...
+            // Javadoc version >=11 is needed for search
+            // Javadoc version >12 is needed to avoid javadoc bugs in linking with modules
+            languageVersion = JavaLanguageVersion.of(15)
+        }
+        jdoc.inputs.file javaDocOverviewLocation
+        jdoc.options.encoding = 'UTF-8'
+        jdoc.options.tags = ['apiNote', 'implNote']
+        // include a note on the front Javadoc page telling us what version the Javadoc was generated from.
+        jdoc.options.overview = new File(javaDocOverviewLocation)
+        // Link to javadoc 11 API, because it is closer to java 8, which we actually use to build and run
```
rcaudy, line 39: This comment seems wrong, and maybe we want to not hard-code 11?

buildSrc/src/main/groovy/io/deephaven/project/ProjectType.groovy
```diff
@@ -0,0 +1,88 @@
...
+import org.gradle.api.tasks.TaskProvider
+
+@CompileStatic
+class ProjectType {
+
+    public static final String VERIFY_ALL_PROJECTS_REGISTERED_TASK_NAME = 'verifyAllProjectsRegistered'
+
+    public static final String VERIFY_RUNTIME_CLASSPATH_IS_PUBLIC_TASK_NAME = 'verifyRuntimeClasspathIsPublic'
+
+    static void registerRoot(Project project) {
```
rcaudy, line 17: Can we have more than one root?

buildSrc/src/main/groovy/io.deephaven.bin-conventions.gradle
```diff
@@ -0,0 +1,23 @@
+plugins {
+    id 'java'
+}
+
+// our "bin" projects are a little special; the resources that go into the jar
```
rcaudy, line 5: Should we rename this "configs"? "resources"? Should we try to conform to IntelliJ's expectations?

buildSrc/src/main/groovy/io.deephaven.java-header-conventions.gradle
```diff
@@ -0,0 +1,13 @@
+plugins {
+    // This is old, and should be removed eventually
```
rcaudy, line 2: Make a todo + ticket.

buildSrc/src/main/groovy/io.deephaven.java-license-conventions.gradle
```diff
@@ -0,0 +1,8 @@
+plugins {
+    // io.deephaven.tools.License depends on java plugin
```
rcaudy, line 2: Seems like an unnecessary comment.

buildSrc/src/main/groovy/io.deephaven.java-test-conventions.gradle
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        if (findProperty('shortTests') == 'true') {
+            systemProperty 'TstUtils.shortTests', 'true'
+        }
+        systemProperty 'Configuration.rootFile', 'dh-tests.prop'
+        systemProperty 'devroot', "$rootDir"
+        systemProperty 'workspace', "$rootDir/tmp/workspace"
+        systemProperty 'configuration.quiet', 'true'
+        systemProperty 'java.awt.headless', 'true'
```
rcaudy, line 61: We probably don't need this anymore.

buildSrc/src/main/groovy/io.deephaven.java-jar-conventions.gradle
```diff
@@ -0,0 +1,18 @@
+plugins {
+    id 'java'
+}
+
+jar {
+    manifest {
+        attributes 'Implementation-Title': 'Deephaven Database',
```
rcaudy, line 7: Not Database.

server/jetty/build.gradle
```diff
@@ -1,5 +1,6 @@
 plugins {
     id 'application'
+    id 'io.deephaven.project.java-local'
```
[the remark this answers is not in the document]
rcaudy: @niloc132 might need to weigh in.

buildSrc/src/main/groovy/io/deephaven/project/ProjectType.groovy
devinrsmith: Technically yes, some other project could register itself as a "root" project. I *could* provide some safety rails, but I think it's sort of a harmless misconfiguration that should be immediately obvious.

buildSrc/src/main/groovy/io.deephaven.bin-conventions.gradle
devinrsmith: I'm doing one better. I'm restructuring the "bin" projects, and deleting the custom logic that was previously used for them.

open-api/shared-ide/shared-ide.gradle
```diff
@@ -1,3 +1,7 @@
+plugins {
+    id 'io.deephaven.project.java-public'
```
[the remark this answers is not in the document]
devinrsmith: afaict, this stuff needs to be public. @niloc132 for further clarification. I think some of this stuff could be cleaned up, but out of scope for this PR IMO.

py/jpy/build.gradle
```diff
@@ -1,33 +1,33 @@
...
+    id 'com.bmuschko.docker-remote-api'
+}
 
-  apply plugin: 'com.bmuschko.docker-remote-api'
+evaluationDependsOn Docker.registryProject('python-dev-base')
 
-  test {
+test {
     onlyIf { false } // don't run tests, it depends on python assets, which we aren't building ATM
     exclude "**/*" // make sure intellij doesn't try to use this disabled test class7
```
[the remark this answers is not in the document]
devinrsmith: Deleted stray. Not sure about comment...

web/client-api/client-api.gradle
```diff
@@ -1,3 +1,7 @@
+plugins {
+    id 'io.deephaven.project.java-local'
```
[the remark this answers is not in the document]
devinrsmith: I think the jar that client-api produces is only consumed internally, and not explicitly depended on from our servers.

projects/test-configs/build.gradle
```diff
@@ -0,0 +1,10 @@
+plugins {
+    id 'io.deephaven.project.bin-local'
+}
+
+dependencies {
+    // For unit-tests that explicitly (or transitively) depend on Configuration.getInstance(),
+    // we want to make sure that they continue working without having to spin up an etcd backed
```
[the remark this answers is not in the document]
devinrsmith: Moved logic into test-configs, deleted dep.

buildSrc/src/main/groovy/io.deephaven.bin-conventions.gradle
```diff
@@ -0,0 +1,23 @@
...
+// The projectDir is $rootDir/projects/$b.name, so we avoid using any relative files.
+String resources = "$rootDir/$project.name".toString()
+// Set the sourceSet resources directory.
+project.sourceSets.main.resources.srcDirs = [resources]
+project.sourceSets.main.java.srcDirs = [] // no java source for these modules.
+// This is technically superfluous, but we'll leave it just in case
+(project.tasks.jar as Jar).exclude 'build', 'out', 'build.gradle'
+
+// This is a one-time cleanup; it works better than trying to tell IntelliJ / gradle to ignore the files.
+// We'll leave it in case developers switch to release branches then back to develop.
```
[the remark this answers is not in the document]
devinrsmith: Now deleted :)

open-api/shared-ide/shared-ide.gradle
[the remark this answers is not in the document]
niloc132: Some of these are shared, and depended on (eventually) by the server so it can do autocomplete work. In DHE we actually shared that code with the client in a way that would pass objects between client and server, but in DHC we use gRPC instead. https://github.com/deephaven/deephaven-core/issues/187 is tracking the removal of those actual shared types, we probably need to merge some of the empty/pointless projects too.

web/client-api/client-api.gradle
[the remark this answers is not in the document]
niloc132: we're just publishing the web jar then with the various JS in it?

combined-javadoc/build.gradle
```diff
@@ -0,0 +1,65 @@
...
+}
+
+configurations {
+    combinedJavadoc
+}
+
+String javaDocOverviewLocation = 'build/docs/overview.html'
+
+def gitHash
+gitHash = "${-> gitHash = 'git rev-list --max-count=1 HEAD'.execute([], rootDir).text.trim()}"
```
niloc132, line 13: Just in case, might want to fall back on some default strings, such as if someone downloads a zip of the project from github rather than clone.

buildSrc/src/main/groovy/io.deephaven.java-test-conventions.gradle
niloc132: conclusion here? the linked build is stuck on the enterprise plugin

buildSrc/src/main/groovy/io.deephaven.bin-conventions.gradle
```diff
@@ -0,0 +1,23 @@
...
+//// Set the sourceSet resources directory.
+//project.sourceSets.main.resources.srcDirs = [resources]
+//project.sourceSets.main.java.srcDirs = [] // no java source for these modules.
+//// This is technically superfluous, but we'll leave it just in case
+//(project.tasks.jar as Jar).exclude 'build', 'out', 'build.gradle'
+//
+//// This is a one-time cleanup; it works better than trying to tell IntelliJ / gradle to ignore the files.
+//// We'll leave it in case developers switch to release branches then back to develop.
+//// We really _shouldn't_ do file deletion during configuration phase, but this should be very fast once already clean.
+//files("$resources/build", "$resources/out").files*.deleteDir()
```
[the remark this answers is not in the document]
devinrsmith: oops, deleted.

combined-javadoc/build.gradle
devinrsmith: Yeah, it's a good point - the good news is that it's not part of the prepareCompose dependencies, and typically only run in CI. I can create an issue for future.

devinrsmith: #2000

server/jetty/build.gradle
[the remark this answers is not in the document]
devinrsmith: The reason not public *yet* is just some extra wiring to publish the js jar; hoping it can be a follow-up?

settings.gradle
```diff
@@ -60,12 +57,17 @@ pyMods.each {
         project(":$name").projectDir = file(dir)
 }
 
+include(':configs')
+project(':configs').projectDir = file('projects/configs')
```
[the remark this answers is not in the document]
devinrsmith: This was the path of the configs "project". I just got rid of the extra layer of indirection that we had. If we want to move configs, test-configs to a different directory, or top level, we can.

server/jetty/build.gradle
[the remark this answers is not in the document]
devinrsmith: #2035

web/client-api/client-api.gradle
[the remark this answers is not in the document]
devinrsmith: #2035

devinrsmith:
Fork every follow up #2037

rcaudy (approved):

jcferretti (approved):

PR #1935, opened 2022-02-04
Reworded the documentation for lastByKeyColumns.

jcferretti:
Original discussion here:
https://github.com/deephaven/deephaven-core/pull/1825#discussion_r799118672


jcferretti:
I don't think I am adding any value here; this change is so small that is easily replicated as a new PR, and Charles and/or Ryan are in a much better position to decide what this javadoc should say, I add not value being in the middle of it.

PR #1942, opened 2022-02-04
Plugin for pandas dataframes and JS API/UI support

### LINKED ISSUE #741: Display Pandas DataFrames in the Web UI
It would be nice to be able to examine the contents of Pandas dataframes in the Web UI. DHE supports that.

```
import pandas as pd
import numpy as np
df = pd.DataFrame({'A': [1,2,3], 'B': [1.1,np.nan,3.3]})
df1 = df.isna()
df2 = df.dropna()
df3 = df.fillna(value=7.3)

from deephaven.TableTools import col, newTable
from deephaven.conversion_utils import NULL_DOUBLE
t = newTable(col('A',1,2,3), col('B',1.1,NULL_DOUBLE,3.3))
t1 = t.update("IsMissing = isNull(B)")
t2 = t.where("!isNull(B)")
t3 = t.update("B=isNull(B) ? 7.3 : B")
```

It would demo that DH truly plays well with Python.

niloc132:
This is a draft for now, hoping for feedback on how to allow python plugins to be built-in.

Fixes #741


docker/server/src/main/docker/Dockerfile
```diff
@@ -20,7 +20,7 @@ RUN set -eux; \
 
 COPY deephaven-wheel/ /deephaven-wheel
 RUN set -eux; \
-    python -m pip install -q --no-index --no-cache-dir /deephaven-wheel/*.whl; \
+    python -m pip install -q --no-cache-dir /deephaven-wheel/*.whl; \
```
devinrsmith, line 23: For final version, we shouldn't be checking in this change. If deephaven-plugin remains external, we'd need to update the deephaven base images as appropriate.

Integrations/python/deephaven/pandas/pandas_as_table.py
```diff
@@ -0,0 +1,18 @@
+from deephaven.python_to_java import dataFrameToTable
+from deephaven.plugin.object import Exporter, ObjectType
+from pandas import DataFrame
```
devinrsmith, line 3: We have a choice - this pandas plugin *could* be completely external (like how the matplotlib plugin lives today). The bad part about that right now is that it has knowledge about "deephaven.python_to_java", and potentially other aspects of Integrations. That said, if we are trying to move to be better "python as a deephaven library", we *should* be trying to move towards publishing the python integration points as PyPi packages.

Integrations/python/setup.py
```diff
@@ -55,6 +55,12 @@ def normalize_version(ver):
       keywords='Deephaven Development',
       packages=find_packages(exclude=['docs', 'test']),
       install_requires=['deephaven-jpy=={}'.format(__normalized_version__),
-                        'numpy', 'dill>=0.2.8', 'wrapt', 'pandas', 'numba;python_version>"3.0"',
-                        'enum34;python_version<"3.4"'],
+                        'deephaven-plugin==0.0.1.dev5',
```
devinrsmith, line 58: If we had deephaven-plugin-pandas external, we could specify it here instead of deephaven-plugin.

Either way, we've tended to put the hard version requirements as part of our docker/**/requirements.txt files as opposed to here in the setup.py.

devinrsmith (commented):
Think there is room for further discussion in team channels.

devinrsmith (dismissed):

jmao-denver (approved):
maybe we can use a dict for entry_points

niloc132:
Testing process, run this python code in the java-run server:
```
import pandas as pd
d = {'IT':[1,20,35,44,51,1000]}
df = pd.DataFrame({'dept': 'IT', 'count': d['IT']})
```
expected, a table shows up with `df` as its title, and a bar across it showing that it is a pandas dataframe, with a refresh button to reload the content (for when it is modified directly, without binding a new instance.)

jmao-denver (approved):

devinrsmith (approved):
LGTM, assuming you've tested in console with your latest change.

Issue #1730, opened 2021-12-22
NULL query constants and most likely other ones don't work well when carried over to Python


jmao-denver:
```
from deephaven2.constants import NULL_SHORT
from deephaven2 import empty_table
null_short = NULL_SHORT
t = empty_table(10).update(["X = i%2==0 ? NULL_SHORT : 1", "Y = NULL_SHORT", "Z = null_short", "X=(short)X"])
```
This code will result in X column being of short type, but both Y and Z int type, because NULL_SHORT is a Python int when mapped from QueryConstants.NULL_SHORT.

Some questions:
1. making NULL_SHORT a Java boxed type might work in this case, but what happens when it is mixed with non special short values?
2. what's NULL_SHORT + 5? If NULL_SHORT is java.lang.Short, then it will raise an exception, otherwise, it is perfectly valid operation
3. the more generic question is how we can safely input and output special DH values in Python?

jmao-denver:
Fixed by #1949

PR #1949, opened 2022-02-08
Hand-wrap TableReplayer/DynamicTableWriter/TableLoggers, fixes #1944

### LINKED ISSUE #1944: Migrate the TableManipulation/TableLoggers modules to deephaven v2 (hand-wrapping/refactoring)
Not a simple migration but more like curating the existing functionality and find a new, natural home for them in v2. 

related to #1263

jmao-denver:
1. fold the TableReplayer and TableWriter into table_factory.py
2. renamed TableLoggers to perfmon.py
3. fixed some unclear documentation in the TableLoggers.java

fixes #1944 #2004 #1730


engine/table/src/main/java/io/deephaven/engine/table/impl/util/DynamicTableWriter.java
```diff
@@ -481,7 +481,7 @@ void writeToColumnSource() {
 
         @Override
         public void set(Object value) {
-            setByte(value == null ? QueryConstants.NULL_BYTE : (Byte) value);
+            setByte(value == null ? QueryConstants.NULL_BYTE : ((Number) value).byteValue());
```
rcaudy, line 484: I kind of hate this set of changes. That said, I also kind of hate that we've documented using the weakly-typed `DynamicTableWriter#logRow` method as if it were advisable; it will never be a high-performance solution to anything.

Maybe a compromise would be to add a new `addPermissive` interface method or similar, that lets the user explicitly send weakly-typed inputs that we'll convert on a best-effort basis.

rcaudy: The more I think about it, the more I think adding a weakly-typed parallel code path for Python is the way to go. That way we have Java code that owns the translation from Python.

engine/table/src/main/java/io/deephaven/engine/table/impl/util/DynamicTableWriter.java
```diff
@@ -244,7 +253,48 @@ public void logRow(Object... values) {
...
+     * This is a convenience function so that you can log an entire row at a time using a Map. You must specify all
+     * values in the setters map (and can't have any extras). The type of the value must be convertible (safely or
+     * unsafely) to the type of the permissive setter.
+     *
+     * @param values a map from column name to value for the row to be logged
+     */
+    @SuppressWarnings("unused")
+    public void logRowPermissive(Map<String, Object> values) {
+        if (values.size() != factoryMap.size()) {
+            throw new RuntimeException("Incompatible logRow call: " + values.keySet() + " != " + factoryMap.keySet());
```
rcaudy, line 272: Wrong method name. Also at line 293.

rcaudy (commented):
Just reviewed DyamicTableWriter.java.

pyintegration/tests/test_table_factory.py
```diff
@@ -100,11 +109,71 @@ def test_input_column_error(self):
...
+                "Double": dtypes.double,
+                "Float": dtypes.float32,
+                "Long": dtypes.long,
+                "Int32": dtypes.int32,
+                "Short": dtypes.short,
+                "Byte": dtypes.byte
+            }
+            with DynamicTableWriter(col_defs) as table_writer:
+                table_writer.write_row(10, 10, 10, 10, 10, 10)
+                table_writer.write_row(10.1, 10.1, 10.1, 10.1, 10.1, 10.1)
```
chipkent, line 139: I would assert the output on this guy as a sanity check.

pyintegration/tests/test_table_factory.py
```diff
@@ -100,11 +109,71 @@ def test_input_column_error(self):
...
+
+        with self.subTest("Incorrect value types"):
+            with DynamicTableWriter(col_defs) as table_writer, self.assertRaises(DHError) as cm:
+                table_writer.write_row(10, '10', 10, 10, 10, '10')
+            self.assertIn("RuntimeError", cm.exception.root_cause)
+
+    def test_historical_tale_replayer(self):
```
chipkent, line 146:
```suggestion
    def test_historical_table_replayer(self):
```

pyintegration/tests/test_table_factory.py
```diff
@@ -100,11 +109,71 @@ def test_input_column_error(self):
...
+
+        replayer = TableReplayer(start_time, end_time)
+        replay_table = replayer.add_table(hist_table, "DateTime")
+        replay_table2 = replayer.add_table(hist_table, "DateTime")
+        self.assertEqual(replay_table, replay_table2)
+
+        replayer.start()
+        self.assertTrue(replay_table.is_refreshing)
+        self.assertTrue(replay_table2.is_refreshing)
+        sleep(1)
```
chipkent, line 167: @rcaudy may have an opinion on if this sleep is the best option.

pyintegration/tests/test_table_factory.py
```diff
@@ -100,11 +109,71 @@ def test_input_column_error(self):
...
+        replayer.shutdown()
 
-@dataclass
-class CustomClass:
-    f1: int
-    f2: str
+        with self.assertRaises(DHError) as cm:
+            replay_table3 = replayer.add_table(hist_table, "DateTime")
+        self.assertIn("RuntimeError", cm.exception.root_cause)
```
chipkent, line 172: Looks like you are expecting a failure if you register after starting.  This should be reflected in the javadoc above.

chipkent (changes requested):

pyintegration/deephaven2/table_factory.py
```diff
@@ -2,16 +2,19 @@
...
 
-from deephaven2 import DHError
+from deephaven2 import DHError, dtypes
 from deephaven2.column import InputColumn
 from deephaven2.table import Table
 
 _JTableFactory = jpy.get_type("io.deephaven.engine.table.TableFactory")
 _JTableTools = jpy.get_type("io.deephaven.engine.util.TableTools")
+_JDynamicTableWriter = jpy.get_type('io.deephaven.engine.table.impl.util.DynamicTableWriter')
+_JReplayer = jpy.get_type('io.deephaven.engine.table.impl.replay.Replayer')
```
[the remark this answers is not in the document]
rcaudy: I think `table_factory` is OK. It's not a wrapper around `io.deephaven.engine.table.TableFactory`, but it's somewhat equivalent.
I agree that I wouldn't put replayers or writers there. I don't have a proposed name for where to put them, maybe their own individual modules?

pyintegration/tests/test_table_factory.py
rcaudy: I have an opinion that I don't even know why it's there.

pyintegration/deephaven2/table_factory.py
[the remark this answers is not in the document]
jmao-denver: my reasoning  was all the functions in table_factory are moved up to the package level anyways and I didn't want to create another module or package to host those single purpose functions, also I couldn't think of a good name for the modules. Ironically I think TableTools is a decent name but we risk making it a dumping ground again. @chipkent what do you think?

pyintegration/deephaven2/table_factory.py
```diff
@@ -111,3 +114,120 @@ def merge_sorted(tables: List[Table], order_by: str) -> Table:
         return Table(j_table=_JTableTools.mergeSorted(order_by, *[t.j_table for t in tables]))
     except Exception as e:
         raise DHError(e, "merge sorted operation failed.") from e
+
+
+class DynamicTableWriter:
```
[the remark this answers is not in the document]
jmao-denver: yes, although not strictly necessary

pyintegration/tests/test_table_factory.py
jmao-denver: Since we can't verify it deterministically, sleep(1) would hopefully allow the server replay the table for 1s.

pyintegration/tests/test_table_factory.py
jmao-denver: content check won't be deterministically since the table is refreshing, or will they?

pyintegration/deephaven2/table_factory.py
```diff
@@ -111,3 +114,120 @@ def merge_sorted(tables: List[Table], order_by: str) -> Table:
...
+
+        Args:
+            *values (Any): the values of the new row, the data types of these values must match the column definitions
+                of the table
+
+        Raises:
+             DHError
+        """
+        try:
+            self._j_table_writer.logRowPermissive(*values)
```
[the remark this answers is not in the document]
jmao-denver: I would say no. But one can argue either way.

pyintegration/deephaven2/table_factory.py
```diff
@@ -111,3 +114,120 @@ def merge_sorted(tables: List[Table], order_by: str) -> Table:
...
+        Args:
+            table (Table): the historical table
+            col (str): column in the table containing timestamps
+
+        Returns:
+            a new Table
+
+        Raises:
+            DHError
+        """
```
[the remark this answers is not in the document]
jmao-denver: added a test and tables can be added after replay started. add_table is slight better than just add.

pyintegration/deephaven2/table_factory.py
```diff
@@ -111,3 +114,120 @@ def merge_sorted(tables: List[Table], order_by: str) -> Table:
...
     except Exception as e:
         raise DHError(e, "merge sorted operation failed.") from e
+
+
+class DynamicTableWriter:
+    """ The DynamicTableWriter creates a new in-memory table and supports writing data to it. This class implements
+    the context manager protocol and thus can be used in the with statement.
+    """
+
+    def __init__(self, col_defs: Dict[str, DType]):
```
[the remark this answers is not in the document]
jmao-denver: yes, Python dict now is guaranteed to ordered based on insertion. Unless we are worried about race conditions (another thread modifying the dict at the same time) here I don't think we need to worry about the ordering.

engine/table/src/main/java/io/deephaven/engine/table/impl/util/DynamicTableWriter.java
```diff
@@ -20,10 +20,12 @@
 import io.deephaven.engine.table.impl.sources.ArrayBackedColumnSource;
```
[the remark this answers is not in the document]
jmao-denver: The Python tests should have provided enough coverage.

py/jpy/src/main/c/jpy_jtype.c
```diff
@@ -423,6 +423,40 @@ int JType_CreateJavaShortObject(JNIEnv* jenv, JPy_JType* type, PyObject* pyArg,
     return JType_CreateJavaObject(jenv, type, pyArg, JPy_Short_JClass, JPy_Short_Init_MID, value, objectRef);
 }
 
+int JType_CreateJavaIntegerObjectEx(JNIEnv* jenv, JPy_JType* type, PyObject* pyArg, jobject* objectRef)
```
devinrsmith, line 426: I don't like the name of this method, it suggest the return type is as always a java `java.lang.Integer`.

JType_CreateJavaByteObject returns java.lang.Byte
JType_CreateJavaShortObject returns java.lang.Short
JType_CreateJavaIntegerObject returns java.lang.Integer

etc.

py/jpy/src/main/c/jpy_jtype.c
```diff
@@ -862,7 +896,7 @@ int JType_ConvertPythonToJavaObject(JNIEnv* jenv, JPy_JType* type, PyObject* pyA
     } else if (PyBool_Check(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Boolean_JClass, type->classRef)))) {
         return JType_CreateJavaBooleanObject(jenv, type, pyArg, objectRef);
     } else if (JPy_IS_CLONG(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Integer_JClass, type->classRef)))) {
```
devinrsmith, line 898: `(*jenv)->IsAssignableFrom(jenv, JPy_Integer_JClass, type->classRef)` means:

"Can a java.lang.Integer be safely cast to classRef?"

If so, we could return a java.lang.Integer and know that a user expecting classRef would be able to safely use it.

Now, `JType_CreateJavaIntegerObjectEx` has a variable return type - and the same check we are doing doesn't exhaustively check that it's a safe return type.

At the very least, we need to add some more if-cases if we want to support JType_CreateJavaIntegerObjectEx - but, I'd suggest each of the internal cases you have in JType_CreateJavaIntegerObjectEx may need to be moved to individual if statements in this method.

devinrsmith (dismissed):
Only looked at the jpy / c stuff for now, but I don't think it's safe as-is.

py/jpy/src/main/c/jpy_jtype.c
jmao-denver: Changed to JType_CreateJavaIntegerTypeObjectByInference

py/jpy/src/main/c/jpy_jtype.c
```diff
@@ -423,6 +423,40 @@ int JType_CreateJavaShortObject(JNIEnv* jenv, JPy_JType* type, PyObject* pyArg,
...
 }
 
+int JType_CreateJavaIntegerTypeObjectByInference(JNIEnv* jenv, JPy_JType* type, PyObject* pyArg, jobject* objectRef)
+{
+    jvalue value;
+    char b;
+    short s;
+    long i;
+    long long j;
+    if (JPy_IS_CLONG(pyArg)) {
```
devinrsmith, line 433: We've already verified this condition is true, yeah?

py/jpy/src/main/c/jpy_jtype.c
```diff
@@ -861,7 +895,9 @@ int JType_ConvertPythonToJavaObject(JNIEnv* jenv, JPy_JType* type, PyObject* pyA
         return JPy_AsJString(jenv, pyArg, objectRef);
     } else if (PyBool_Check(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Boolean_JClass, type->classRef)))) {
         return JType_CreateJavaBooleanObject(jenv, type, pyArg, objectRef);
-    } else if (JPy_IS_CLONG(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Integer_JClass, type->classRef)))) {
+    } else if (JPy_IS_CLONG(pyArg) && (type == JPy_JObject )) {
+        return JType_CreateJavaIntegerTypeObjectByInference(jenv, type, pyArg, objectRef);
```
devinrsmith, line 899: I think `JavaIntegerType` is still not the best name. I *think* something like:

```c
    } else if (JPy_IS_CLONG(pyArg) && ((*jenv)->IsAssignableFrom(jenv, JPy_Number_JClass, type->classRef))) {
        return JType_CreateJavaNumberObject(jenv, type, pyArg, objectRef);
```

Not sure if `JPy_Number_JClass` exists, but would point to `java.lang.Number`.

pyintegration/deephaven2/_jcompat.py
```diff
@@ -56,4 +56,18 @@ def j_properties(d: Dict = None) -> jpy.JType:
...
+def j_map_to_dict(m):
+    """  Converts a java map to a python dictionary. """
+    if not m:
+        return None
+
+    r = {}
+    for e in m.entrySet().toArray():
+        k = e.getKey()
+        v = e.getValue()
+        r[k] = v
```
chipkent, line 71: I think this can be done with an explicit iterator to avoid array creation.  Efficiency change.

chipkent (changes requested):
This is a partial review.

py/jpy-integration/src/javaToPython/java/io/deephaven/jpy/integration/MyNumbersBoxed.java
```diff
@@ -0,0 +1,12 @@
+package io.deephaven.jpy.integration;
+
+public interface MyNumbersBoxed {
```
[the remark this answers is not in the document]
devinrsmith: Everything in py/jpy-integration is testing code.

py/jpy/src/main/c/jpy_jtype.c
```diff
@@ -861,7 +891,9 @@ int JType_ConvertPythonToJavaObject(JNIEnv* jenv, JPy_JType* type, PyObject* pyA
         return JPy_AsJString(jenv, pyArg, objectRef);
     } else if (PyBool_Check(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Boolean_JClass, type->classRef)))) {
         return JType_CreateJavaBooleanObject(jenv, type, pyArg, objectRef);
-    } else if (JPy_IS_CLONG(pyArg) && (type == JPy_JObject || ((*jenv)->IsAssignableFrom(jenv, JPy_Integer_JClass, type->classRef)))) {
+    } else if (JPy_IS_CLONG(pyArg) && (type == JPy_JObject || (*jenv)->IsAssignableFrom(jenv, JPy_Number_JClass, type->classRef))) {
+        return JType_CreateJavaNumberFromPythonInt(jenv, type, pyArg, objectRef);
+    } else if (JPy_IS_CLONG(pyArg) && ((*jenv)->IsAssignableFrom(jenv, JPy_Integer_JClass, type->classRef))) {
```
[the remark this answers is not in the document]
devinrsmith: This code path changes a few things:

`org.jpy.PyObject#getObjectValue`, and anything that calls into jpy and says they want a return type of `Object` or `Number`, `org.jpy.PyObject#call(java.lang.Class<T>, java.lang.String, java.lang.Class<?>[], java.lang.Object[])`, ie `pyObject.call(Number.class, "my_method")`, etc.

My guess is that there aren't any callers currently that ask for `Number`, but we do have conversion code that expects an `Object` back. Namely, how we bring python script variable changes into scope for java. I think we are really trying to solve for the long vs int issue; but there may be some existing script code that could break, if it starts getting a byte or a short back. It may be reasonable to say that we always want to convert to java int, or java long; not into java byte or java short.

As currently coded,
```python
foo = 1
```
will create a java `byte` type.

engine/table/src/main/java/io/deephaven/engine/table/impl/util/DynamicTableWriter.java
```diff
@@ -162,13 +164,20 @@ public Row getRowWriter() {
      * @return a RowSetter for the given column
      */
     @Override
-    public RowSetter getSetter(String name) {
+    public PermissiveRowSetter getSetter(String name) {
```
devinrsmith, line 167: Not relevant explicitly for this review, but I think we are planning to deprecate DynamicTableWriter eventually and push the input table interface. Hopefully we can provide nice wrappers for that soon.

engine/table/src/main/java/io/deephaven/engine/table/impl/util/DynamicTableWriter.java
```diff
@@ -244,7 +253,50 @@ public void logRow(Object... values) {
...
+
+    /**
+     * This is a convenience function so that you can log an entire row at a time using a Map. You must specify all
+     * values in the setters map (and can't have any extras). The type of the value must be convertible (safely or
+     * unsafely) to the type of the permissive setter.
+     *
+     * @param values a map from column name to value for the row to be logged
+     */
+    @SuppressWarnings("unused")
+    public void logRowPermissive(Map<String, Object> values) {
```
devinrsmith, line 270: Does it make sense to provide `logRow(Map)` as well?

engine/table/src/main/java/io/deephaven/engine/table/impl/util/DynamicTableWriter.java
```diff
@@ -527,7 +594,12 @@ void writeToColumnSource() {
 
         @Override
         public void set(Object value) {
-            setInt(value == null ? QueryConstants.NULL_INT : (Integer) value);
+            setInt(value == null ? QueryConstants.NULL_INT : (int) value);
```
devinrsmith, line 597: Does casting as a primitive `int` make us more permissive here than before? I'm not sure if there might be some boxed -> primitive logic that is different than casting to explicit box.

I could check myself, but what happens in the following cases:

```java
int x1 = (int)Long.valueOf(1);
int x2 = (Integer)Long.valueOf(1);
int x3 = (int)Short.valueOf(1);
int x4 = (Integer)Short.valueOf(1);
```

I'm *assuming* that `x2` and `x4` should fail - does `x1` and `x3` fail similarly?

Ditto for all other `set` changes.

pyintegration/deephaven2/_jcompat.py
chipkent: I think jpy might handle it for some but not all types.

pyintegration/deephaven2/perfmon.py
```diff
@@ -0,0 +1,256 @@
...
+        DHError
+    """
+    try:
+        return _JPerformanceQueries.processInfo(proc_id, proc_type, key)
+    except Exception as e:
+        raise DHError(e, "failed to obtain the process info.") from e
+
+
+def process_memory():
+    """ Presents a user friendly view with basic memory and GC data samples for the current engine process.
```
chipkent, line 150: view of what?

chipkent (changes requested):

pyintegration/deephaven2/replay.py
```diff
@@ -0,0 +1,80 @@
...
+        Returns:
+            a replay Table
+
+        Raises:
+            DHError
+        """
+        try:
+            replay_table = Table(j_table=self._j_replayer.replay(table.j_table, col))
+            self._hist_tables.append((table, col))
+            self._replay_tables.append(replay_table)
```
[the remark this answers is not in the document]
jmao-denver: Shouldn't interfere with engine's table GC logic but would with Python's GC. Initially wanted to give the user a way to inspect the replayer but probably not necessary, will remove

pyintegration/deephaven2/table_factory.py
```diff
@@ -111,3 +116,68 @@ def merge_sorted(tables: List[Table], order_by: str) -> Table:
...
+
+        Raises:
+            DHError
+        """
+        try:
+            self._j_table_writer.close()
+        except Exception as e:
+            raise DHError(e, "failed to close the writer.") from e
+
+    def write_row(self, *values: Any) -> None:
```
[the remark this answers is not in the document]
jmao-denver: I could go either way too, But since this is probably going away (replaced by the better StreamPublisher). I wouldn't bother with it now.

py/jpy/src/main/c/jpy_jtype.c
```diff
@@ -423,6 +423,36 @@ int JType_CreateJavaShortObject(JNIEnv* jenv, JPy_JType* type, PyObject* pyArg,
...
+        return JType_CreateJavaObject(jenv, type, pyArg, JPy_Integer_JClass, JPy_Integer_Init_MID, value,
+                                     objectRef);
+    }
+    if (b != s) {
+        value.s = s;
+        return JType_CreateJavaObject(jenv, type, pyArg, JPy_Short_JClass, JPy_Short_Init_MID, value,
+                                      objectRef);
+    }
+    value.b = b;
+    return JType_CreateJavaObject(jenv, type, pyArg, JPy_Byte_JClass, JPy_Byte_Init_MID, value, objectRef);
```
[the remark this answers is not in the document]
jmao-denver: In our attempt to ensure the changes we made won't break existing behaviors or rather improve, we found another issue with how we determine the value types of Python objects. A new issue was filed for that: https://github.com/deephaven/deephaven-core/issues/2044. This is related in that the proper mapping of Python value type to Java's may not be possible to be handled in JPY alone, some changes in the engine might be needed.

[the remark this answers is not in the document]
chipkent: We have tried to break this multiple ways, and the only significant problem we found is #2044.  I think what is here should get merged, but we may need to reassess it at some future point.

pyintegration/deephaven2/replay.py
```diff
@@ -75,6 +73,6 @@ def start(self) -> None:
     def shutdown(self) -> None:
         """ Shuts down and invalidates the replayer. After this call, the replayer can no longer be used. """
         try:
-            self._j_replayer.shutdown()
+            self._hist_tables = []
```
chipkent, line 76: Shouldn't the shutdown call still be here, but the `self._hist_tables` be gone?

chipkent (changes requested):

PR #1958, opened 2022-02-09
install tenserflow

hythloda:


demo/grpc-api/src/main/app.d/crypto-application.app
```diff
@@ -4,3 +4,5 @@ id=app.deephaven.demo.crypto.Application
 name=Crypto Application
 enabled=true
 file_0=crypto-application.py
+file_1=tensorflow-install.py
```
mofojed, line 7: Should we add the install to the Dockerfile instead? As per https://github.com/deephaven/deephaven-core/blob/demo/demo/grpc-api/src/main/app.d/crypto-application.py
My reasoning:
- Not related to the crypto app (I don't think), so shouldn't be referenced from crypto-application.app. Would need to make a separate app just for this
- Package should be installed with the image, so that it doesn't need to be installed whenever the session starts up (which could slow down startup)

@devinrsmith may have input as well

hythloda: agreed. I was going for the quick and dirty. removed that for the better option which is more extendable. now have a requirements file to pip install in the Dockerfile.

demo/grpc-api/src/main/docker/Dockerfile
```diff
@@ -4,4 +4,9 @@ FROM deephaven/grpc-api:local-build
...
 
 # Add the application directory
-COPY app.d/ /app.d
\ No newline at end of file
+COPY app.d/ /app.d
+
+
+from python:3.8
+COPY requirements.txt /requirements.txt
+RUN pip install -r /requirements.txt
```
devinrsmith, line 12: This isn't going to work - this is just creating a new image with python instead of basing off of deephaven/grpc-api.

devinrsmith (changes requested):
Re-`FROM`ing against python won't work - we are replacing the image. Now, we could probably just do a `RUN pip install tensorflow==2.8.0`, but I can't say what will happen without testing...

demo/grpc-api/src/main/docker/Dockerfile
hythloda: I thought I removed line 10 and 11 and changed it to :RUN pip3 install -r /app.d/requirements.txt

JamesXNelson:
To be clear, what I'm suggesting is that we should have a `docker-python-all` image, where we install everything, and that image should likely be a) what is used in the demo, and b) the default that we suggest to new users running from docker.

demo/grpc-api/src/main/docker/Dockerfile
JamesXNelson: This is also misleading, the branch I'm deploying no longer uses this Dockerfile (though, arguably, it should again; I think it was removed when I was away and mike 'n devin were trying to get a working deploy...). My repo+branch is JamesXNelson/deephaven-core jxn/demo_refactored and in it, the base `docker/server/src/main/docker/Dockerfile` is what is getting deployed.  This is very messy, and needs to be cleaned up....  but, for now, we need to give _that_ image the correctly installed py dependencies, preferably using pre-baked base images that already have all python dependencies setup correctly.

JamesXNelson: note, I select the image in demo/demo.gradle via
`t.dependsOn demoGrpcApi.tasks.findByName('buildDocker-server')`

If we setup a "has everything" base image in `docker/server`, then I can just point to it via task name in demo.gradle, and we should be done here.

devinrsmith:
The "base" images aren't the final product. You can built the equivalent Deephaven server image via the appropriate task:

```
docker-server:buildDocker-server
docker-server:buildDocker-server-nltk
docker-server:buildDocker-server-pytorch
docker-server:buildDocker-server-sklearn
docker-server:buildDocker-server-tensorflow
```

devinrsmith:
One issue trying to built an all-encompassing AI image is that they have competing explicit dependencies; some that are outdated and have CVEs. We need to have a QA process around the combined dependencies, b/c otherwise there may be incompatibilities we introduce

PR #1964, opened 2022-02-10
Increased compatibility with `Flight` reference implementation of `DoExchange`

lbooker42:
`ArrowFlight` implementation of `DoExchange` does not allow `app_metadata` to be populated on the first message but our current implementation requires every message to contain valid metadata to be processed.


server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
```diff
@@ -356,9 +364,35 @@ public void onNext(final InputStream request) {
...
+                    // handle the possible error cases
+                    if (!isFirstMsg) {
+                        // only the first messages is allowed to have null metadata
+                        throw GrpcUtil.statusRuntimeException(Code.INVALID_ARGUMENT,
+                                myPrefix + "failed to receive Barrage request metadata");
+                    }
+
+                    isFirstMsg = false;
+
+                    // test whether the FlightDescriptor cmd field contains a valid Barrage object
```
lbooker42, line 380: For maximum compatibility, we should make the `FlightDescriptor` cmd field optional and allow nulls/empty rather than require a barrage-style message that serves no purpose.  If we want to add action commands later, we can test for that but allowing nulls is the most user-friendly.

lbooker42: After discussion, decided to enforce `FlightDescriptor` cmd field being equal to `BarrageUtil.FLATBUFFER_MAGIC` which is '0x6E687064' (the numerical representation of the ASCII "dphn".)

java-client/flight/src/main/java/io/deephaven/client/impl/FlightSession.java
```diff
@@ -79,6 +75,18 @@ public FlightStream stream(HasTicketId ticketId) {
         return FlightClientHelper.get(client, ticketId);
     }
 
+    /**
+     * Creates a new server side DoExchange session .
+     *
+     * @param descriptor the FlightDescriptor objcet to include on the first FlightData message (other fields will
```
nbauernfeind, line 81: s/objcet/object

java-client/flight-examples/src/main/java/io/deephaven/client/examples/DoExchange.java
```diff
@@ -0,0 +1,113 @@
...
+        // equivalent to '0x6E687064' which is the numerical representation of the ASCII "dphn".
+        byte[] cmd = new byte[] {100, 112, 104, 110};
+
+        FlightDescriptor fd = FlightDescriptor.command(cmd);
+
+        try (FlightClient.ExchangeReaderWriter erw = flight.startExchange(fd);
+                final RootAllocator allocator = new RootAllocator(Integer.MAX_VALUE)) {
+
+            // BarrageSnapshotRequest for ticket 's/timetable'
+            cmd = new byte[] {16, 0, 0, 0, 0, 0, 10, 0, 16, 0, 8, 0, 7, 0, 12, 0, 10, 0, 0, 0, 0, 0, 0, 7, 100, 112,
```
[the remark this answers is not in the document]
lbooker42: I agree, but wasn't sure if I could add a dependency in this sub-project to the `barrage` libraries.  They weren't referenced before.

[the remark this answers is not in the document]
nbauernfeind: flight-examples is fair game to add dependencies to

java-client/flight-examples/build.gradle
```diff
@@ -1,12 +1,15 @@
 plugins {
     id 'application'
+    id 'java-library'
```
nbauernfeind, line 3: If this is required for `api` then I would change it to `implementation` and leave this line out. Otherwise double check with Devin that this is an OK add.

java-client/flight-examples/src/main/java/io/deephaven/client/examples/DoExchange.java
```diff
@@ -0,0 +1,89 @@
...
+            /////////////////////////////////////////////////////////////
+
+            // inner metadata for the snapshot request
+            final FlatBufferBuilder metadata = new FlatBufferBuilder();
+
+            int optOffset =
+                    BarrageSnapshotOptions.createBarrageSnapshotOptions(metadata, ColumnConversionMode.Stringify,
+                            false, 1000);
+
+            final int ticOffset = BarrageSnapshotRequest.createTicketVector(metadata, "s/timetable".getBytes());
```
nbauernfeind, line 46: ideally the scoped field name would be a flag

server/test/src/main/java/io/deephaven/server/test/FlightMessageRoundTripTest.java
```diff
@@ -387,6 +380,133 @@ public void testGetSchema() {
...
+                byte[] msg = wrapper.sizedByteArray();
+                ArrowBuf data = allocator.buffer(msg.length);
+                data.writeBytes(msg);
+
+                erw.getWriter().putMetadata(data);
+                erw.getWriter().completed();
+
+                // read everything from the server
+                while (erw.getReader().next()) {
+                    // NOP
```
nbauernfeind, line 434: I suggest counting the messages. I see below that you expect the root to match the expected row count. I think a given root is applicable immediately after `next()` returns. I'm kind of surprised it is not nulled out after `next()` returns `false`. IMO, I would label that a flight bug (maybe it doesn't matter if they are using their own allocator). Either way, it probably makes sense to error if the number of messages is unexpected before erroring if the number of rows is unexpected. (Fragmentation, for example, is likely to break big snapshots up.)

java-client/flight-examples/build.gradle
devinrsmith: This is not typically appropriate, unless you want to have other consumers depend on flight-examples... I don't think that's the case here.

lbooker42: per @devinrsmith, `implementation` is correct, will remove this line

nbauernfeind (approved):

PR #1970, opened 2022-02-12
Cleanup of .env content; Add debezium/perf.

jcferretti:
On `.env` files:

* Remove unused definitions.
* Rename PORT to DEEPHAVEN_PORT; since environment variables override, it is desirable to have something more specific to avoid somebody's personal environment definitions mixing things for them.

Simplify memory configuration for the server definition in `docker-compose-common.yml`.


devinrsmith (changes requested):
I was under the assumption you'd be removing our docker deploy resource limits - was that going to be separate?

I do want to take the time Monday and run the compose file myself.

jcferretti:
> I was under the assumption you'd be removing our docker deploy resource limits - was that going to be separate?
> 
> I do want to take the time Monday and run the compose file myself.

Somehow I lost that change; done.

devinrsmith (approved):
Let's give @niloc132 a chance; and I still want to take a pass at running it soon.

PR #1985, opened 2022-02-15
Exclude certain engine methods from Tier 4 compilation

### LINKED ISSUE #1529: PhaseAggressiveCoalesce SIGSEGV
We've run into DB:test triggering SIGSEGV twice now in the same relative spot, on two different JDK 17 JVMs.

```
2021-10-30T06:24:33.1443208Z > Task :DB:test
2021-10-30T06:35:03.7824211Z #
2021-10-30T06:35:03.7924936Z # A fatal error has been detected by the Java Runtime Environment:
2021-10-30T06:35:03.8004601Z #
2021-10-30T06:35:03.8100735Z #  SIGSEGV (0xb) at pc=0x00007f641e67d27c, pid=9248, tid=9262
2021-10-30T06:35:03.8180524Z #
2021-10-30T06:35:03.8626465Z # JRE version: OpenJDK Runtime Environment Zulu17.28+13-CA (17.0+35) (build 17+35-LTS)
2021-10-30T06:35:03.8649483Z # Java VM: OpenJDK 64-Bit Server VM Zulu17.28+13-CA (17+35-LTS, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
2021-10-30T06:35:03.8659672Z # Problematic frame:
2021-10-30T06:35:03.8781830Z # V  [libjvm.so+0x5a327c]  PhaseAggressiveCoalesce::coalesce(Block*)+0x5bc
2021-10-30T06:35:03.8813477Z #
2021-10-30T06:35:03.8929788Z # Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport %p %s %c %d %P %E" (or dumping to /home/runner/work/deephaven-core/core.9248)
2021-10-30T06:35:03.8949855Z #
2021-10-30T06:35:04.0609087Z # An error report file with more information is saved as:
2021-10-30T06:35:04.0635799Z # /home/runner/work/deephaven-core/hs_err_pid9248.log
2021-10-30T06:35:04.0664505Z #
2021-10-30T06:35:04.0712513Z # Compiler replay data is saved as:
2021-10-30T06:35:04.0759258Z # /home/runner/work/deephaven-core/replay_pid9248.log
2021-10-30T06:35:04.0821478Z #
2021-10-30T06:35:04.0898339Z # If you would like to submit a bug report, please visit:
2021-10-30T06:35:04.1025821Z #   http://www.azul.com/support/
2021-10-30T06:35:04.1072909Z #
```

and

```
2021-11-04T06:21:38.1020387Z > Task :DB:test
2021-11-04T06:32:54.1827737Z #
2021-11-04T06:32:54.1843776Z # A fatal error has been detected by the Java Runtime Environment:
2021-11-04T06:32:54.1851434Z #
2021-11-04T06:32:54.1862188Z #  SIGSEGV (0xb) at pc=0x00007f2dd7200dbc, pid=9554, tid=9568
2021-11-04T06:32:54.1869803Z #
2021-11-04T06:32:54.2121001Z # JRE version: OpenJDK Runtime Environment Temurin-17.0.1+12 (17.0.1+12) (build 17.0.1+12)
2021-11-04T06:32:54.2133835Z # Java VM: OpenJDK 64-Bit Server VM Temurin-17.0.1+12 (17.0.1+12, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
2021-11-04T06:32:54.2142508Z # Problematic frame:
2021-11-04T06:32:54.2178345Z # V  [libjvm.so+0x5a2dbc]  PhaseAggressiveCoalesce::coalesce(Block*)+0x5bc
2021-11-04T06:32:54.2189617Z #
2021-11-04T06:32:54.2200460Z # Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport %p %s %c %d %P %E" (or dumping to /home/runner/work/deephaven-core/core.9554)
2021-11-04T06:32:54.2208082Z #
2021-11-04T06:32:54.2220071Z # An error report file with more information is saved as:
2021-11-04T06:32:54.2229771Z # /home/runner/work/deephaven-core/hs_err_pid9554.log
2021-11-04T06:32:54.2239442Z #
2021-11-04T06:32:54.2247048Z # Compiler replay data is saved as:
2021-11-04T06:32:54.2258419Z # /home/runner/work/deephaven-core/replay_pid9554.log
2021-11-04T06:32:54.2268199Z #
2021-11-04T06:32:54.2276148Z # If you would like to submit a bug report, please visit:
2021-11-04T06:32:54.2291548Z #   https://github.com/adoptium/adoptium-support/issues
2021-11-04T06:32:54.2299594Z #
```

Unfortunately, we have been unable to grab the JVM err logs thus far - hopefully once #1528 we'll be able to capture the logs successfully.

niloc132:
This commit excludes ~11 methods from hotspot's Tier 4 (aka C2) JIT compilation. Each of these has been found crash the certain JVM implementations (across vendors, across OSes, and across hardware archetectures) when it compiles this code, but not consistently. By excluding only C2 in this way, we allow C1 (aka Tier 1-3) to still run, and get better performance than interpreted bytecode.

All existing ways to run code should be affected by this - running main()s or tests through either gradle or IntelliJ, and running our shell scripts to launch the server will all end up with a compiler directives file to disable c2 for these methods. Using the engine as a library is _not_ fixed in this way, but this is a problem that can only be solved through documentation, to inform users that they should make sure to start the JVM with a specific set of flags.

An extra array of "devJvmArgs" is left in the io.deephaven.java-conventions plugin to make it easier to add more flags later for this kind of problem - it will automatically be included for all of the above categories.

Fixes #1529


niloc132:
Tested in Java11 and 17, see https://github.com/deephaven/deephaven-core/runs/5205478306?check_suite_focus=true - I will keep re-running this to ensure that 17 doesn't run into surprise issues. I'll put up a later PR to introduce the changes that lead to that nightly build as well.

We might want to consider the c2-excludes.txt as a build output, or at least publish its contents with each release. It might also make sense to rename it, as we might refine this e.g. to also disable inlining if we find that a specific method is the actual problem, etc.

buildSrc/src/main/groovy/io.deephaven.java-conventions.gradle
```diff
@@ -74,18 +76,85 @@ tasks.withType(JavaCompile).configureEach {
...
+def tier4CompilationExcludes = [
+  'io.deephaven.engine.table.impl.by.ssmpercentile.DoublePercentileTypeMedianHelper':['setResult'],
+  'io.deephaven.engine.table.impl.by.ssmpercentile.FloatPercentileTypeMedianHelper':['setResult'],
+  'io.deephaven.engine.table.impl.by.ssmpercentile.SsmChunkedPercentileOperator':['addChunk'],
+  'io.deephaven.engine.table.impl.by.IncrementalChunkedOperatorAggregationStateManager':['buildTable'],
+  'io.deephaven.engine.table.impl.RightIncrementalChunkedCrossJoinStateManager':['addToIndex', 'decorationProbe'],
+  'io.deephaven.engine.table.impl.IncrementalChunkedNaturalJoinStateManager':['buildTable', 'decorationProbe'],
+]
+def createCompilerDirectives = tasks.register('createCompilerDirectives') {
+  def c2ExcludesFile = project.layout.buildDirectory.file('c2-excludes.txt')
```
devinrsmith, line 88: Agree w/ your comment elsewhere, I think we should name this something else - maybe `dh-compiler-directives.txt`?

buildSrc/src/main/groovy/io.deephaven.java-conventions.gradle
```diff
@@ -74,18 +76,85 @@ tasks.withType(JavaCompile).configureEach {
...
   javaLauncher.set testRuntimeLauncher
+  jvmArgs += c2ExcludesJvmArgs(c2ExcludesFile.singleFile.path) + devJvmArgs
 }
 
 tasks.withType(GroovyCompile).configureEach {
   javaLauncher.set groovyCompilerLauncher
 }
 
+plugins.withType(ApplicationPlugin) {
+  applicationDistribution.into('lib') {
```
devinrsmith, line 143: Do we have strong opinions about where the file should live?

I could argue for the directory `lib`, `files` (or some other generic name), or even into the root directory.

buildSrc/src/main/resources/windowsStartScript.txt
```diff
@@ -0,0 +1,93 @@
...
+set APP_HOME=%DIRNAME%${appHomeRelativePath}
+
+@rem Resolve any "." and ".." in APP_HOME to make it shorter.
+for %%i in ("%APP_HOME%") do set APP_HOME=%%~fi
+
+@rem Add default JVM options here. You can also use JAVA_OPTS and ${optsEnvironmentVar} to pass JVM options to this script.
+set DEFAULT_JVM_OPTS=${defaultJvmOpts}
+
+@rem Customization for deephaven-core to reference a compiler directives file
+set DEFAULT_JVM_OPTS=%DEFAULT_JVM_OPTS% -XX:CompilerDirectivesFile=%APP_HOME%/c2-excludes.txt
```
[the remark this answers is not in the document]
niloc132: I'll need to double check, I moved it a few times in my hopes of parameterize this file. I'll test on windows with Don tomorrow.

buildSrc/src/main/resources/unixStartScript.txt
```diff
@@ -0,0 +1,188 @@
...
+cd "\$SAVED" >/dev/null
+
+APP_NAME="${applicationName}"
+APP_BASE_NAME=`basename "\$0"`
+
+# Add default JVM options here. You can also use JAVA_OPTS and ${optsEnvironmentVar} to pass JVM options to this script.
+DEFAULT_JVM_OPTS=${defaultJvmOpts}
+
+# Customization for deephaven-core to reference a compiler directives file
+DEFAULT_JVM_OPTS="\${DEFAULT_JVM_OPTS} -XX:+UnlockDiagnosticVMOptions -XX:CompilerDirectivesFile=\"\${APP_HOME}/lib/c2-excludes.txt\""
```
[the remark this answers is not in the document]
niloc132: Correct, one line in each file, with a comment, that calls out "deephaven" by name.

buildSrc/src/main/groovy/io.deephaven.java-conventions.gradle
```diff
@@ -74,18 +76,85 @@ tasks.withType(JavaCompile).configureEach {
...
+    from(createCompilerDirectives.get().outputs.files)
+  }
+}
+tasks.withType(CreateStartScripts).configureEach {
+  inputs.properties(['scriptVers': Math.random()])
```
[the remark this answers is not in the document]
niloc132: Some form is, this is the dumb way of "never up to date". Ideally I show it that the classpath changes when the file changes, or we just put a static value here that we change when the file changes. I'll see if i can at least make buildSrc's own build.gradle emit some property or something that represents the contents of these files, since actually changing the file doesnt invalidate classpath (unlike changing _this_ file).

buildSrc/src/main/groovy/io.deephaven.java-conventions.gradle
niloc132: Perfect, I'll use that - was mostly looking for someone else to make a suggestion...

buildSrc/src/main/groovy/io.deephaven.java-conventions.gradle
niloc132: lib or root dir was my guess, rather than make a new dir. I don't have a strong opinion here, will take any suggestion/advice you offer. At the end of the day, anything in the archive will work, and will be "owned" by the dh-core install.

If I had to make a new dir, it was going to be `etc/`, as a "these are default configs, which you might change, but the defaults are necessary".

niloc132: Do you have a preference, or are you okay with lib/?

buildSrc/src/main/groovy/io.deephaven.java-conventions.gradle
```diff
@@ -74,18 +76,85 @@ tasks.withType(JavaCompile).configureEach {
...
 
+def tier4CompilationExcludes = [
+  'io.deephaven.engine.table.impl.by.ssmpercentile.DoublePercentileTypeMedianHelper':['setResult'],
+  'io.deephaven.engine.table.impl.by.ssmpercentile.FloatPercentileTypeMedianHelper':['setResult'],
+  'io.deephaven.engine.table.impl.by.ssmpercentile.SsmChunkedPercentileOperator':['addChunk'],
+  'io.deephaven.engine.table.impl.by.IncrementalChunkedOperatorAggregationStateManager':['buildTable'],
+  'io.deephaven.engine.table.impl.RightIncrementalChunkedCrossJoinStateManager':['addToIndex', 'decorationProbe'],
+  'io.deephaven.engine.table.impl.IncrementalChunkedNaturalJoinStateManager':['buildTable', 'decorationProbe'],
+]
+def createCompilerDirectives = tasks.register('createCompilerDirectives') {
```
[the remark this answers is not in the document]
niloc132: By making the map here, we can ensure there aren't duplicate class/method entries (IJ will warn), and IJ will be more likely to notice the class/method name string when you do refactors.

Also given the issues with the start scripts, we want to be certain that gradle notices changes as we tweak this, and it doesn't seem to understand that when it is in buildSrc/src/main/resources, unlike buildSrc/src/main/groovy.

Between those it seemed like enough of a benefit to generate it as needed.

buildSrc/src/main/groovy/io.deephaven.java-conventions.gradle
devinrsmith: I'm fine with `lib/` ATM. If we find ourselves needing more special files in the future, I think something like `etc/` may be called for.

buildSrc/src/main/groovy/io.deephaven.java-conventions.gradle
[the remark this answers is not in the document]
devinrsmith: Got it.

buildSrc/src/main/groovy/io.deephaven.java-conventions.gradle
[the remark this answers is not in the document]
niloc132: Fixed, files can be read off the classpath as input, same as i'm providing them to the generator.

buildSrc/src/main/resources/windowsStartScript.txt
```diff
@@ -0,0 +1,93 @@
...
+set APP_HOME=%DIRNAME%${appHomeRelativePath}
+
+@rem Resolve any "." and ".." in APP_HOME to make it shorter.
+for %%i in ("%APP_HOME%") do set APP_HOME=%%~fi
+
+@rem Add default JVM options here. You can also use JAVA_OPTS and ${optsEnvironmentVar} to pass JVM options to this script.
+set DEFAULT_JVM_OPTS=${defaultJvmOpts}
+
+@rem Customization for deephaven-core to reference a compiler directives file
+set DEFAULT_JVM_OPTS=%DEFAULT_JVM_OPTS% -XX:+UnlockDiagnosticVMOptions "-XX:CompilerDirectivesFile=%APP_HOME%/dh-compiler-directives.txt"
```
devinrsmith, line 40: missing `lib` dir?

niloc132: And slashes pointing the wrong way...

buildSrc/src/main/resources/windowsStartScript.txt
```diff
@@ -0,0 +1,93 @@
+@rem
```
devinrsmith, line 1: Do we still want to check in this file?

niloc132: I could go either way - if we leave it in, maybe i add another comment in the edited region to note that this isn't usable yet, though I think the syntax at least is right.

Removing the file would mean we'd need to re-implement it - I could copy the code and put it somewhere else, but "where it belongs" seems better?

Will go either way, based on your opinion.

devinrsmith:
> > Given our issues with long CLI arguments (on Windows), I wonder if we should try to craft our shell launchers to use star on the `lib` directory, instead of trying to list them all out?
> 
> Downside of star is relying on the order that the shell uses, instead of deterministic classpath order from the specified list of jars.

I don't think it actually relies on shell-globbing - I *think* that java is able to accept arguments and expand the star itself. But, this is a concern potentially for later on.

buildSrc/src/main/resources/windowsStartScript.txt
devinrsmith: Let's take it out for now - I think it's easy to re-impl when necessary.

PR #1989, opened 2022-02-17
Encode Vectors similar to Arrays over Barrage

### LINKED ISSUE #755: DBArrays are being toString as Object instead of primitive array wrappers
DBArrays are functionally equivalent to primitive arrays (most of them are). The Arrow IPC treats array columns as nested columns where the outer column is offsets into the inner column of the primitive.

nbauernfeind:
- Fixes #755.
- Also fixes ungroup() on an object array column that was received via barrage; such as a `long []`.
- Also adds support for `DateTime[]` and `ObjectVector` of `DateTime` (such as one created via a `groupBy`).
- Also fixes columns of Boxed arrays.

### DIFF
--- a/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/CharChunkInputStreamGenerator.java
+++ b/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/CharChunkInputStreamGenerator.java
@@ -6,19 +6,18 @@ package io.deephaven.extensions.barrage.chunk;
 
 import gnu.trove.iterator.TLongIterator;
 import io.deephaven.chunk.ObjectChunk;
-import io.deephaven.chunk.WritableObjectChunk;
 import io.deephaven.chunk.attributes.Values;
 import io.deephaven.chunk.util.pools.PoolableChunk;
 import io.deephaven.engine.rowset.RowSet;
 import com.google.common.io.LittleEndianDataOutputStream;
 import io.deephaven.UncheckedDeephavenException;
 import io.deephaven.extensions.barrage.util.StreamReaderOptions;
-import io.deephaven.util.QueryConstants;
 import io.deephaven.util.datastructures.LongSizedDataStructure;
 import io.deephaven.chunk.CharChunk;
 import io.deephaven.chunk.Chunk;
 import io.deephaven.chunk.WritableCharChunk;
 import io.deephaven.chunk.WritableLongChunk;
+import io.deephaven.util.type.TypeUtils;
 import org.jetbrains.annotations.Nullable;
 
 import java.io.DataInput;
@@ -36,7 +35,7 @@ public class CharChunkInputStreamGenerator extends BaseChunkInputStreamGenerator
         WritableCharChunk<Values> outChunk = WritableCharChunk.makeWritableChunk(inChunk.size());
         for (int i = 0; i < inChunk.size(); ++i) {
             final Character value = inChunk.get(i);
-            outChunk.set(i, value == null ? QueryConstants.NULL_CHAR : value);
+            outChunk.set(i, TypeUtils.unbox(value));
         }
         if (inChunk instanceof PoolableChunk) {
             ((PoolableChunk) inChunk).close();
@@ -157,12 +156,6 @@ public class CharChunkInputStreamGenerator extends BaseChunkInputStreamGenerator
         CharConversion IDENTITY = (char a) -> a;
     }
 
-    @FunctionalInterface
-    public interface CharToTypeConversion<T> {
-        T apply(char in);
-        CharToTypeConversion<Character> BOXED = (char a) -> a == NULL_CHAR ? null : (Character) a;
-    }
-
     static Chunk<Values> extractChunkFromInputStream(
             final int elementSize,
             final StreamReaderOptions options,
@@ -216,38 +209,9 @@ public class CharChunkInputStreamGenerator extends BaseChunkInputStreamGenerator
             }
 
             if (options.useDeephavenNulls()) {
-                if (conversion == CharChunkInputStreamGenerator.CharConversion.IDENTITY) {
-                    for (int ii = 0; ii < nodeInfo.numElements; ++ii) {
-                        chunk.set(ii, is.readChar());
-                    }
-                } else {
-                    for (int ii = 0; ii < nodeInfo.numElements; ++ii) {
-                        final char in = is.readChar();
-                        final char out;
-                        if (in == NULL_CHAR) {
-                            out = in;
-                        } else {
-                            out = conversion.apply(in);
-                        }
-                        chunk.set(ii, out);
-                    }
-                }
+                useDeephavenNulls(conversion, is, nodeInfo, chunk);
             } else {
-                long nextValid = 0;
-                for (int ii = 0; ii < nodeInfo.numElements; ++ii) {
-                    if ((ii % 64) == 0) {
-                        nextValid = isValid.get(ii / 64);
-                    }
-                    final char value;
-                    if ((nextValid & 0x1) == 0x0) {
-                        value = NULL_CHAR;
-                        is.skipBytes(elementSize);
-                    } else {
-                        value = conversion.apply(is.readChar());
-                    }
-                    nextValid >>= 1;
-                    chunk.set(ii, value);
-                }
+                useValidityBuffer(elementSize, conversion, is, nodeInfo, chunk, isValid);
             }
 
             final long overhangPayload = payloadBuffer - payloadRead;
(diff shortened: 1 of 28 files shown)


extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/CharChunkInputStreamGenerator.java
```diff
@@ -27,6 +31,19 @@
 public class CharChunkInputStreamGenerator extends BaseChunkInputStreamGenerator<CharChunk<Values>> {
     private static final String DEBUG_NAME = "CharChunkInputStreamGenerator";
 
+    public static CharChunkInputStreamGenerator convertBoxed(final ObjectChunk<Character, Values> inChunk) {
+        // This code path is utilized for arrays and vectors of DateTimes, which cannot be reinterpreted.
+        WritableCharChunk<Values> outChunk = WritableCharChunk.makeWritableChunk(inChunk.size());
+        for (int i = 0; i < inChunk.size(); ++i) {
+            final Character value = inChunk.get(i);
+            outChunk.set(i, value == null ? QueryConstants.NULL_CHAR : value);
```
rcaudy, line 39: `TypeUtils.unbox`

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/ChunkInputStreamGenerator.java
```diff
@@ -156,6 +205,53 @@
                             }
                     );
                 }
+                if (type == DateTime.class) {
+                    return LongChunkInputStreamGenerator.extractChunkFromInputStreamWithTypeConversion(
+                            Long.BYTES, options, DateTime::new, fieldNodeIter, bufferInfoIter, is
```
rcaudy, line 210: `io.deephaven.time.DateTimeUtils#nanosToTime`

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/CharChunkInputStreamGenerator.java
```diff
@@ -139,6 +157,12 @@ public int drainTo(final OutputStream outputStream) throws IOException {
         CharConversion IDENTITY = (char a) -> a;
     }
 
+    @FunctionalInterface
+    public interface CharToTypeConversion<T> {
+        T apply(char in);
+        CharToTypeConversion<Character> BOXED = (char a) -> a == NULL_CHAR ? null : (Character) a;
```
rcaudy, line 163: `TypeUtils.box`

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/CharChunkInputStreamGenerator.java
```diff
@@ -235,4 +259,79 @@ public int drainTo(final OutputStream outputStream) throws IOException {
...
+
+            final long payloadRead = (long) nodeInfo.numElements * elementSize;
+            if (payloadBuffer < payloadRead) {
+                throw new IllegalStateException("payload buffer is too short for expected number of elements");
+            }
+
+            if (options.useDeephavenNulls()) {
+                for (int ii = 0; ii < nodeInfo.numElements; ++ii) {
+                    final char in = is.readChar();
+                    chunk.set(ii, in == NULL_CHAR ? null : conversion.apply(in));
```
rcaudy, line 308: We should give this to the conversion even when `null`. Or not, since that would differ from the behavior for non-DH nulls, unless we want to pass the dh null value in that case?

Or you can document the difference, etc.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/CharChunkInputStreamGenerator.java
```diff
@@ -235,4 +259,79 @@ public int drainTo(final OutputStream outputStream) throws IOException {
...
+                }
+            } else {
+                long nextValid = 0;
+                for (int ii = 0; ii < nodeInfo.numElements; ++ii) {
+                    if ((ii % 64) == 0) {
+                        nextValid = isValid.get(ii / 64);
+                    }
+                    final T value;
+                    if ((nextValid & 0x1) == 0x0) {
```
rcaudy, line 317: I wonder (and this is not worth changing unless easy) if it would be best to determine the lowest one-bit and skip to that.

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -419,35 +419,40 @@ private InputStream getInputStream(final View view, final ByteBuffer metadata, f
...
+                bufferInfos.get().add(length);
             };
             numRows = columnVisitor.visit(view, addStream, fieldNodeListener, bufferListener);
 
-            RecordBatch.startNodesVector(header, nodeOffsets.size());
-            for (int i = nodeOffsets.size() - 1; i >= 0; --i) {
-                final ChunkInputStreamGenerator.FieldNodeInfo node = nodeOffsets.get(i);
+            RecordBatch.startNodesVector(header, nodeOffsets.get().size());
+            for (int i = nodeOffsets.get().size() - 1; i >= 0; --i) {
+                final ChunkInputStreamGenerator.FieldNodeInfo node =
```
rcaudy, line 446: Worth doing the repeated part outside the loop (get() through cast)? Current chunks are fixed after the visit is done.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/vector/CharVectorExpansionKernel.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        final ObjectChunk<CharVector, A> typedSource = source.asObjectChunk();
+        final SizedCharChunk<A> resultWrapper = new SizedCharChunk<>();
+
+        int lenWritten = 0;
+        perElementLengthDest.setSize(source.size() + 1);
+        for (int i = 0; i < typedSource.size(); ++i) {
+            final CharVector row = typedSource.get(i);
+            final int len = row == null ? 0 : row.intSize();
```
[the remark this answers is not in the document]
rcaudy: Should fix ObjectVectorExpansionKernel the same way.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/FixedWidthChunkInputStreamGenerator.java
```diff
@@ -0,0 +1,130 @@
...
+import io.deephaven.chunk.WritableObjectChunk;
+import io.deephaven.chunk.attributes.Values;
+import io.deephaven.extensions.barrage.util.StreamReaderOptions;
+import io.deephaven.util.datastructures.LongSizedDataStructure;
+
+import java.io.DataInput;
+import java.io.IOException;
+import java.util.Iterator;
+
+import static io.deephaven.util.QueryConstants.NULL_CHAR;
```
rcaudy, line 15: Unused.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/FixedWidthChunkInputStreamGenerator.java
```diff
@@ -0,0 +1,130 @@
...
+            if (overhangPayload > 0) {
+                is.skipBytes(LongSizedDataStructure.intSize(DEBUG_NAME, overhangPayload));
+            }
+        }
+
+        chunk.setSize(nodeInfo.numElements);
+        return chunk;
+    }
+
+    private static <T> void useValidityBuffer(
```
rcaudy, line 99: Be sure to apply suggestions from `CharChunkInputStreamGenerator` here.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/FixedWidthChunkInputStreamGenerator.java
```diff
@@ -0,0 +1,130 @@
...
+            return chunk;
+        }
+
+        final int numValidityLongs = options.useDeephavenNulls() ? 0 : (nodeInfo.numElements + 63) / 64;
+        try (final WritableLongChunk<Values> isValid = WritableLongChunk.makeWritableChunk(numValidityLongs)) {
+            if (options.useDeephavenNulls() && validityBuffer != 0) {
+                throw new IllegalStateException("validity buffer is non-empty, but is unnecessary");
+            }
+            int jj = 0;
+            for (; jj < Math.min(numValidityLongs, validityBuffer / 8); ++jj) {
```
rcaudy, line 63: I have a (maybe superstitious) belief that it's better to write down the bounds when there's any computation to be done.
```suggestion
            final int numValidityLongsPresent = Math.min(numValidityLongs, validityBuffer / 8);
            for (; jj < numValidityLongsPresent; ++jj) {
```

rcaudy (approved):

PR #1990, opened 2022-02-17
Enable formatting of BigDecimal and BigInteger in the web console.

jcferretti:
This change by itself does not make a difference, but together with this change on the `web-client-ui` repo does the trick:
```
cfs@erke 12:58:09 ~/dh/oss3/web-client-ui
$ git diff
diff --git a/packages/iris-grid/src/TableUtils.ts b/packages/iris-grid/src/TableUtils.ts
index cdfb3d1..b0f5e9b 100644
--- a/packages/iris-grid/src/TableUtils.ts
+++ b/packages/iris-grid/src/TableUtils.ts
@@ -294,6 +294,7 @@ class TableUtils {
       case 'java.lang.Double':
       case 'float':
       case 'java.lang.Float':
+      case 'java.math.BigDecimal':
       case TableUtils.dataType.DECIMAL:
         return TableUtils.dataType.DECIMAL;
       case 'int':
@@ -304,6 +305,7 @@ class TableUtils {
       case 'java.lang.Short':
       case 'byte':
       case 'java.lang.Byte':
+      case 'java.math.BigInteger':
       case TableUtils.dataType.INT:
         return TableUtils.dataType.INT;
       default:
```
Selection_195


niloc132 (approved):

jcferretti:
PR for doc here (merged already): https://github.com/deephaven/deephaven.io/pull/1189

PR #2001, opened 2022-02-18
Update demo layout for web v0.9.3

mattrunyon:
Markdown notebooks no longer have their own component and instead are supported as part of a normal notebook panel. This updates the layout to use the proper component and adds some props that were likely added in those updates somewhere.

Layout was created and exported in latest web version


mattrunyon:

mofojed (approved):

PR #2012, opened 2022-02-22
Fix an issue with MetricsManager.resetCounters()

jcferretti:
Thanks to Jianfeng for finding this during Python testing.


jmao-denver (approved):
The change looks good and the added test case ensures the bug is fixed.

PR #2014, opened 2022-02-23
Improve javadoc for PerformanceQueries.java.

jcferretti:
Specially around query evaluation numbers.


jmao-denver (changes requested):

jmao-denver (approved):
The updated JavaDocs have clarified the API much more.

PR #2029, opened 2022-02-25
Clean up deephaven_jpy_init.py

### LINKED ISSUE #2027: Clean up deephaven_jpy_init.py
This module was carried over almost unchanged from DHE so there might be things in it that don't apply anymore.

jmao-denver:
Removed code that doesn't apply to DHC
Add /data/notebooks to sys.path to support importing other notebooks

Fixes #2027


engine/table/python/core/deephaven_jpy_init.py
```diff
@@ -18,188 +16,5 @@
...
-    :param use_intraday:
-    :param server_host:
-    :param as_of_time:
-    :return:
-    """
-
-    _jtype_ = jpy.get_type("io.deephaven.engine.table.impl.util.PerformanceQueries")
-
-    tableMap = _jtype_.queryUpdatePerformanceMap(evaluationNumber)
-    add_tables_from_map_to_binding(tableMap)
```
chipkent, line 69: If we are taking this out, we need a ticket to wrap `io.deephaven.engine.table.impl.util.PerformanceQueries`.  In enterprise, these are queries that make the perf analysis of queries much easier.  I assume the same may still be true in DHC.

engine/table/python/core/deephaven_jpy_init.py
```diff
@@ -18,188 +16,5 @@
...
-            # noinspection PyBroadException
-            try:
-                importjava(line)
-            except Exception:
-                logging.error("Could not import java class: {}".format(line))
-
-# clean up the workspace, since this is forwarded as the console namespace
-del _empty_or_comment_, _exists_and_is_file_, default_imports
+# make it possible to import other python modules saved in notebooks
+sys.path.append("/data/notebooks")
```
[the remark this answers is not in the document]
mofojed: @devinrsmith  What's the risk in doing this automatically? I think it is somewhat expected, as this is the directory where our notebooks are hosted.

mofojed (approved):

engine/table/python/core/deephaven_jpy_init.py
```diff
@@ -18,188 +16,5 @@
...
-    javaclass = jpy.get_type(clazz)
-    for key, value in javaclass.__dict__.items():
-        if not key.startswith("__"):
-            if hasattr(value, 'methods'):
-                methods = value.methods
-                if methods:
-                    if methods[0].is_static:
-                        globals()[value.name] = value
-
-    ql.importClass(_jclass_.forName(clazz))
```
chipkent, line 100: `QueryLibrary` allows java functions to be added to the query language.  We might want to wrap this to allow more powerful multi-language integrations.  It should be in its own module and not here.  Create a ticket for it.

engine/table/python/core/deephaven_jpy_init.py
```diff
@@ -18,188 +16,5 @@
...
-
-            try:
-                if import_pattern.match(line):
-                    exec(line)
-                else:
-                    logging.error("Could not run line \n{}\n from python.default.imports, "
-                                  "does not match import pattern" .format(line))
-            except ImportError:
-                logging.error("Could not import module: {}".format(line))
-    del import_re, import_pattern
```
chipkent, line 189: I'm not sure about the deletion of this section.  In Groovy, we have the ability to automatically exec some code at the start of a session, so that console users don't have to type as much.  In Python, we have backed away from that and have required users to be explicit.  Should we retain the functionality?  Maybe?  Maybe not.

engine/table/python/core/deephaven_jpy_init.py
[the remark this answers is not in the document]
devinrsmith: It's a hack - ostensibly, we are trying to *clean* up our old hacks, not add new ones. This is easy enough to document that a user could run it if the want - or, if there is a feature request for this, there are better places to add it.

engine/table/python/core/deephaven_jpy_init.py
```diff
@@ -18,188 +16,5 @@
...
-    with open(default_imports) as f:
-        for line in f:
-            if _empty_or_comment_():
-                continue
-
-            # noinspection PyBroadException
-            try:
-                importjava(line)
-            except Exception:
-                logging.error("Could not import java class: {}".format(line))
```
chipkent, line 202: This section comes back to importing java methods into the query language.  DHE has the ability to preconfigure user-specified functions to be in the query language.  It is a reasonable feature.

How best to handle this and the imports is an interesting question.  Simply supporting a (1) preexec script and (2) a `QueryLibrary` wrapper, you could cover both of the use cases here, with less specific code.

engine/table/python/core/deephaven_jpy_init.py
chipkent, line 20: Is this the best way to do this?  You can imagine scenarios where a user would want to add other directories to the path as well.  Maybe you leave this as it is and then assume that the other directories could be added as part of a preexec script or via the Python path.  If the `PYTHON_PATH` is the right choice, should `/data/notebooks` be in the `PYTHON_PATH`, instead of here?  Also, hard-coding this here may lead to problems with python-as-a-library.

chipkent (changes requested):

engine/table/python/core/deephaven_jpy_init.py
jmao-denver: the new perfmon module in #1949 wraps it and that's why I removed it.

engine/table/python/core/deephaven_jpy_init.py
```diff
@@ -18,188 +16,5 @@
...
-from deephaven.python_to_java import dataFrameToTable
-
-# NOTE: **THIS IS REQUIRED** for WorkerPythonEnvironment - don't take it out...
-import __main__
-
-
-###############################################################################################
-
-# can't use IsWidget.get_dh_table or the data frame will be interpreted as a table instead of a widget?
-pandas.DataFrame.to_dh_table = lambda self: dataFrameToTable(self, True)
```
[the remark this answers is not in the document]
niloc132: Yes, this will still work, it will use a different mechanism, see https://github.com/deephaven/deephaven-core/pull/1942

engine/table/python/core/deephaven_jpy_init.py
niloc132: Doesn't application mode cover this use case, allowing the server to start up with some pre-defined script?

jmao-denver:
#2053 is created to restore the removed QL customization feature in the future

engine/table/python/core/deephaven_jpy_init.py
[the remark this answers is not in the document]
niloc132: Two specific reasons I hate this:

 * it assumes that `/data/notebooks` is a valid directory even outside of docker, and that it is safe to load content from there.
 * it assumes that `/data/notebooks` is safely user-controlled, which even inside of docker, this isn't valid - we use a wide-open webdav setup to allow any user on the internet to write to these directories. For our demo instances we deem this to be safe (instances are short lived, can't connect to the server, and are hard to discover), but we wouldn't want to make this part of a client installation.

Instead I propose making it a suggested idiom in an application mode script to append the desired path to sys.path, and let the app mode script author control it that way (as opposed to some configuration property etc)

chipkent (approved):

PR #2049, opened 2022-02-28
Add BOM

devinrsmith:


bom/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+
+description = 'Deephaven Bill of Materials'
+
+dependencies {
+    constraints { constraint ->
+        project.rootProject
+            .subprojects
+            .findAll {  p -> io.deephaven.project.ProjectType.isPublic(p) }
+            .each { p ->
+                constraint.api p
```
JamesXNelson, line 13: if there's any runtime-only dependencies, they should get runtime constraints too.

JamesXNelson (approved):
looks great!

might want to look at specifically-runtime-only dependencies in the next PR, this looks good as-is.

bom/build.gradle
devinrsmith: Good ticket to track - https://github.com/deephaven/deephaven-core/issues/2050

rcaudy (approved):

PR #2052, opened 2022-02-28
WIP: Create server-standard

devinrsmith:


server/standard/src/main/resources/logback-debug.xml
```diff
@@ -1,29 +1,27 @@
-<configuration debug="false">
+<configuration debug="true">
```
devinrsmith, line 1: I'm not sure why (maybe b/c I made some minor spacing edits), but this file was a simpler rename than this diff suggests...

PR #2058, opened 2022-03-01
Add deephaven2 to sphinx

chipkent:


sphinx/sphinx.gradle
```diff
@@ -8,6 +8,7 @@ plugins {
 description = 'Generates docs for the python libraries provided in Deephaven Core'
 
 evaluationDependsOn ':docker-runtime-base'
+evaluationDependsOn ':docker-server'
 evaluationDependsOn ':pyclient'
```
devinrsmith, line 12: I'm wondering if we can get rid of the `evaluationDependsOn ':docker-runtime-base'` if we move the doxygen to deephaven/server as well?

sphinx/sphinx.gradle
```diff
@@ -38,7 +49,7 @@ def sphinxImage = Docker.registerDockerImage(project, 'sphinx') {
     inputs.files copyPyClientWhl.get().outputs.files
     inputs.files copySphinxLib.get().outputs.files
     inputDir.set layout.buildDirectory.dir('sphinx-image')
-    inputs.files project(':docker-runtime-base').tasks.findByName('buildDocker').outputs.files // deephaven/runtime-base
+    inputs.files project(':docker-server').tasks.findByName('build').outputs.files // deephaven/docker/server
```
devinrsmith, line 52: We shouldn't depend on the `build` task here - it's doing a lot more than we need for deephaven/server (also, the comment is wrong). We should depend on the `buildDocker-server` task.

devinrsmith (changes requested):

sphinx/sphinx.gradle
```diff
@@ -23,12 +23,22 @@ def copySphinxLib = tasks.register('copySphinxLib', Sync) {
 def sphinxDockerfile = tasks.register('sphinxDockerfile', Dockerfile) {
     destFile.set layout.buildDirectory.file('sphinx-image/Dockerfile')
     // Deephaven server python API requires that the wheel be installed to build, so we share this image
-    from 'deephaven/runtime-base:local-build'
+    from 'deephaven/server:local-build'
 
     copyFile "./wheel", "/wheel"
     copyFile "./lib", '${VIRTUAL_ENV}/lib/python3.7/site-packages/'
 
+    environmentVariable 'DEEPHAVEN_CLASSPATH', '/opt/deephaven/server-0.10.0/lib/*:/opt/deephaven/server-0.10.0/lib'
```
devinrsmith, line 31: We should template in `${project.version}`; and, I think we only need the `*` version.

sphinx/sphinx.gradle
```diff
@@ -23,12 +23,22 @@ def copySphinxLib = tasks.register('copySphinxLib', Sync) {
...
+    environmentVariable 'DEEPHAVEN_WORKSPACE', '/tmp/workspace'
+    environmentVariable 'DEEPHAVEN_DEVROOT', '/tmp/devroot'
+    environmentVariable 'DEEPHAVEN_PROPFILE', 'dh-defaults.prop'
+    environmentVariable 'JDK_HOME', '/usr/lib/jvm/zulu8/jre/'
+    environmentVariable 'JAVA_VERSION', '1.8'
```
devinrsmith, line 36: I don't think these environment variables are true wrt deephaven/server. If they aren't actually used / important, we should see if we can delete them?

devinrsmith (approved):

PR #2065, opened 2022-03-03
Debezium perf improvements: better message rate tracking; send to multiple kafka partitions.

jcferretti:


hythloda (approved):
much cleaner. nice change

cpwright (approved):

PR #2066, opened 2022-03-03
Change the default Deephaven java client TLS port to 443

devinrsmith:


java-client/session/src/main/java/io/deephaven/client/impl/ChannelHelper.java
```diff
@@ -1,37 +1,33 @@
...
 
-    public static final int DEFAULT_TLS_PORT = 8080;
+    public static final int DEFAULT_TLS_PORT = 443;
 
-    public static final int DEFAULT_PLAINTEXT_PORT = 8080;
-
-    public static ManagedChannel channel(DeephavenTarget target) {
-        return channelBuilder(target).build();
-    }
+    public static final int DEFAULT_PLAINTEXT_PORT = 10000;
```
devinrsmith, line 14: Note: this was functionally the default before from the context of the examples (ConnectOptions.DEFAULT_PORT).

nbauernfeind (approved):

PR #2067, opened 2022-03-03
Migrated debezium perf to process executors; can send 500k msg/s now.

jcferretti:


debezium/loadgen/generate_load.py
```diff
@@ -441,104 +446,128 @@ async def run_command_server():
...
+# We use process-based executors to run purchase and pageview actions;
+# Each has its own kafka producer; purchase action executors also need
+# a database connection.
+producer = None
+cursor = None
+connection = None
+item_prices = None
+def init_resources(kafka_only = False):
+    global producer, cursor, connection, item_prices
+    if not kafka_only:
```
cpwright, line 458: There is still the cursor stuff on line 334 that I don't see gated by this.

debezium/loadgen/generate_load.py
```diff
@@ -441,104 +446,128 @@ async def run_command_server():
...
+                name="pageviews",
+                num_partitions=kafka_partitions,
+                replication_factor=1
+            )
+        ])
+
+        # With python 3.10 we will be able to use parenthesis around
+        # 'with' with multiple 'as', and get rid
+        # of the backslashes.
+        with \
```
cpwright, line 513: Do you need to have this gated on Kafka only?

debezium/loadgen/generate_load.py
jcferretti: If you look at the initialization code for the executors, the purchases executor is initialized with init_resources(False), which means it will have a cursor.  Line 334 using the global cursor is the purchases loop.  So I believe this is fine; also, from output running there are no exceptions showing up for `cursor` being `None`.

debezium/loadgen/generate_load.py
jcferretti: This is database initialization code in main; is independent of what the executors (which are separate processes) do.
The initialization code that gates on kafka is run on the forked sub-processes of the process pool executor.
I've changed the names of the methods and global variables that are executor specific to all start with `executor_` in the hope that that makes it more clear.

cpwright (approved):

PR #2074, opened 2022-03-08
Nightly builds to run on both Java11 and 17

niloc132:
This patch is the result of the bug hunt around c2 compilation issues, and while it adds support for building each night on each Java 11 and 17, it also simplifies the nightly build yml to run each expected task using the matrix strategy instead of repeating sections of the build over and over.

Build output such as test reports are also somewhat consolidated, to reduce the number of zips available for download, and contain related material in the same file.

One potential downside is that the "Publish Test Results" is now once per matrix run rather than collecting results across all runs.

Another potential downside (or upside) is that when one task fails, the others don't continue. This could also be an upside, save some build minutes, as these are not short tasks. This is configurable, I just set the default for now.

### DIFF
--- a/.github/workflows/nightly-check-ci.yml
+++ b/.github/workflows/nightly-check-ci.yml
@@ -52,22 +52,19 @@ jobs:
         uses: actions/upload-artifact@v2
         if: always()
         with:
-          name: nightly-check-ci-results
-          path: '**/build/test-results/*/TEST-*.xml'
-
-      - name: Upload Test Reports
-        uses: actions/upload-artifact@v2
-        if: always()
-        with:
-          name: nightly-check-ci-reports
-          path: '**/build/reports/tests/**'
+          name: nightly-check-ci-test-results
+          path: |
+            **/build/test-results/*/TEST-*.xml
+            **/build/reports/tests/**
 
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-check-ci.yml
```diff
@@ -41,111 +45,30 @@ jobs:
           echo "org.gradle.java.installations.paths=${{ steps.setup-java-11.outputs.path }},${{ steps.setup-java-17.outputs.path }}," >> gradle.properties
           cat gradle.properties
 
-      - name: Run gradle
+      - name: Run gradle ${{ matrix.gradle-check }} on java ${{ matrix.test-jvm-version }}
         uses: burrunan/gradle-cache-action@v1
         with:
           job-id: check
```
devinrsmith, line 51: Rename, `gradle-run`?

.github/workflows/nightly-check-ci.yml
```diff
@@ -11,6 +11,10 @@ on:
 
 jobs:
   check:
```
devinrsmith, line 13: I think this could be renamed, `nightly`?

devinrsmith (approved):

PR #2075, opened 2022-03-08
Publish deephaven-web-js and deephaven-server-jetty.

### LINKED ISSUE #2035: Publish web-js jar and server-jetty
Part of #1938

The way we are currently depending on the jsJar makes it non-publishable.

https://docs.gradle.org/current/userguide/cross_project_publications.html#sec:simple-sharing-artifacts-between-projects

> Declaring a dependency on an explicit target configuration is not recommended if you plan to publish the component which has this dependency: this would likely lead to broken metadata. If you need to publish the component on a remote repository, follow the instructions of the [variant-aware cross publication documentation](https://docs.gradle.org/current/userguide/cross_project_publications.html#sec:variant-aware-sharing).

devinrsmith:
Removes the jsJar configuration in favor of attaching directly to the main (publishable) jar.

Fixes #2035


settings.gradle
```diff
@@ -26,7 +26,7 @@ String[] mods = [
         // new web projects; these modules are intended to form a complete, modular web application,
         // with heavy dependency isolation that should enable very fast rebuilds.
 String[] webMods = [
-        'web',
+        'web-js',
```
niloc132, line 29: why rename this (here and elsewhere) and make a new project, directory? This is the aggregation of the public artifacts produced by child projects - should "web" have other meaning than this?

rename aside, this isn't just js, but the html/css to run the page, and html samples to show the api

web/js/js.gradle
```diff
@@ -15,13 +12,12 @@ dependencies {
...
 
 /**
- * Provides a jar full of static js/html/css that can be served from jetty
+ * Attach the static js/html/css to the main jar so that can be served from jetty
  */
-def ideClientJsJar = tasks.register('ideClientJsJar', Jar) {
-    from configurations.js
-}
+tasks.named(JavaPlugin.JAR_TASK_NAME, Jar, { jar ->
+    jar.from(configurations.js)
```
niloc132, line 18: I assume that when built this way, gradle doesn't consider those dependencies to be transitive? That was what I was trying to avoid to begin with by making this new configuration.

settings.gradle
devinrsmith: Previously we were targetting the `web` `jsJar` configuration, so I figured the descriptor `js` was appropriate. If we think `web` suffices, or there's a better name, happy to change it.

web/js/js.gradle
devinrsmith: We are "skipping" the dependency transitivity by attach the outputs directly to the jar afaict. Maybe @JamesXNelson has a better way of doing this? I'm trying to avoid declaring a dependency on new configuration, https://docs.gradle.org/current/userguide/cross_project_publications.html#sec:simple-sharing-artifacts-between-projects:

> Declaring a dependency on an explicit target configuration is not recommended if you plan to publish the component which has this dependency: this would likely lead to broken metadata.

(Variant-aware sharing of artifacts between projects seems complex.)

```
$ ./gradlew web-js:jar web-js:generateMetadataFileForMavenJavaPublication web-js:generatePomFileForMavenJavaPublication
$ jar tf web/js/build/libs/deephaven-web-js-0.10.0.jar
$ cat web/js/build/publications/mavenJava/pom-default.xml
$ cat web/js/build/publications/mavenJava/module.json
```

Is how I examined the output.

JamesXNelson: This will also include the .java files.  For better or worse...

Since we're open source it doesn't _really_ matter. A couple KB of noise in the file.

If we wanted to _replace_ the java jar with the js jar, we'd have to get a little more creative with the publication configuration... possible, but I don't really see any value in doing so.  _this is fine_.

docker/web/build.gradle
```diff
@@ -9,7 +9,7 @@ configurations {
     js
 }
 dependencies {
-    js project(path: ':web', targetConfiguration: 'js')
+    js project(path: ':web-js', targetConfiguration: 'js')
```
JamesXNelson, line 12: Assuming this project does NOT produce java artifacts, this is fine.  If this does in any way get published w/ jars, then we can't use `targetConfiguration` and expect transitivity to work.

settings.gradle
JamesXNelson: I'd lean towards "no rename", strictly for the sake of not moving files unless needed, but it really doesn't matter.

JamesXNelson (dismissed):
Tentatively approved; I left comments but I don't _need_ any more work done to consider this good.  @niloc132 should probably sign off too when he's back in action.

settings.gradle
devinrsmith: Undid rename.

web/js/js.gradle
devinrsmith: Note: there *aren't* any .java files :)

niloc132 (approved):

PR #2091, opened 2022-03-11
Fix #2087 bug in SingleRange.ixSubIndexByPosOnNew.

### LINKED ISSUE #2087: RowSet.subIndexByPos bug
This was discovered when exercised by a demo @devinrsmith is building. The following (*groovy*) reproducer simulates what he encountered:
```
t = emptyTable(6).update("II=ii")
t.setRefreshing(true)
c=t.countBy("Count", "II").sortDescending("Count").update("K=k")
s = c.slice(5, Long.MAX_VALUE)
```
(Ticking sort uses a large offset from 0 when building its initial row set in order to allow for data to tick in before existing data without shifting immediately.)

The following should illustrate the specifics of the bug as exercised by the above table operations:
```
println c.getRowSet()
{1073741821-1073741826}

println c.getRowSet().innerSet
io.deephaven.engine.rowset.impl.singlerange.IntStartIntDeltaSingleRange@16626b51
println c.getRowSet().innerSet.unsignedIntStart
1073741821
println c.getRowSet().innerSet.unsignedIntDelta
5

println s.getRowSet().innerSet
io.deephaven.engine.rowset.impl.singlerange.IntStartLongDeltaSingleRange@63cd4dbb
println s.getRowSet().innerSet.unsignedIntStart
1073741826
println s.getRowSet().innerSet.delta
9223372036854775801
```

The operation that is going awry is:
```
c.getRowSet().subsetByPositionRange(5, Long.MAX_VALUE)
```
Meaning there must be a bug in:
`io.deephaven.engine.rowset.impl.singlerange.SingleRange#ixSubindexByPosOnNew`

I think the bug is at the end:
```
        return make(
                Math.max(rangeStart() + startPos, rangeStart()),
                Math.min(rangeStart() + endPos, rangeEnd()));
```

This doesn't contemplate wraparound. We could fix it with this change:
```
        return make(
                Math.max(rangeStart() + startPos, rangeStart()),
                Math.min(Math.max(rangeStart() + endPos, Long.MAX_VALUE), rangeEnd()));
```
That is, if `rangeStart() + endPos` wraps around, just clamp to `Long.MAX_VALUE`.

jcferretti:
The new unit test added fails without the change in SingleRange.java.


Util/src/main/java/io/deephaven/util/SaturationArithmetic.java
```diff
@@ -0,0 +1,25 @@
...
+        return res;
+    }
+
+    public static long minusSaturated(final long x, final long y) {
+        final long res = x - y;
+        if (res <= x) {
+            return Long.MIN_VALUE;
+        }
+        return res;
+    }
```
rcaudy, line 20: We agreed that this is wrong, and since it's also unused, that it should be deleted.

Util/src/main/java/io/deephaven/util/SaturationArithmetic.java
```diff
@@ -0,0 +1,25 @@
...
+// See http://locklessinc.com/articles/sat_arithmetic/
+
+public class SaturationArithmetic {
+    public static long addSaturated(final long x, final long y) {
+        final long res = x + y;
+        if (res < x) {
+            return Long.MAX_VALUE;
+        }
+        return res;
+    }
```
rcaudy, line 12: We said maybe we should just move this to `SingleRange` for now.

Util/src/main/java/io/deephaven/util/SaturationArithmetic.java
```diff
@@ -0,0 +1,25 @@
+package io.deephaven.util;
+
+// See http://locklessinc.com/articles/sat_arithmetic/
```
rcaudy, line 3: If we're going to reference this site, we should be clear that we aren't using any code from there, just using it to understand the concepts of saturated arithmetic.

rcaudy (approved):

engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/singlerange/SingleRange.java
```diff
@@ -204,6 +204,14 @@ public final OrderedLongSet ixRemove(final long key) {
         return OrderedLongSet.twoRanges(rangeStart(), key - 1, key + 1, rangeEnd());
     }
 
+    private static long addSaturated(final long x, final long y) {
+        final long res = x + y;
+        if (res < x) {
+            return Long.MAX_VALUE;
+        }
```
devinrsmith, line 211: I doubt it matters - but we might be able to have a simpler check `res < 0`?

engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/singlerange/SingleRange.java
```diff
@@ -219,8 +227,8 @@ public final OrderedLongSet ixSubindexByPosOnNew(final long startPos, final long
             return ixCowRef();
         }
         return make(
-                Math.max(rangeStart() + startPos, rangeStart()),
-                Math.min(rangeStart() + endPos, rangeEnd()));
+                Math.max(addSaturated(rangeStart(), startPos), rangeStart()),
```
devinrsmith, line 230: is max necessary?

jcferretti: Good catch, thanks.

engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/singlerange/SingleRange.java
jcferretti: We discussed over slack and we think it is better.
We can do a "math proof" that it will be < 0 if it overlaps based on the fact that there are more negative values (1 more) than positive values, and that we know x and y are both non-negative.

rcaudy (approved):

rcaudy (approved):

PR #2096, opened 2022-03-14
Correctly close http2-based emulated bidi streams from JS

### LINKED ISSUE #2095: HTTP2 corrupt table data
Scrolling down on a (ticking?) table, and then back up, the data seems to get corrupted. Colin somewhat triaged as an http2 issue.

image

niloc132:
The earlier iteration of this assumed that just because one empty JS PB object was the same as another from JS's perspective, that GWT/Java wouldn't mind, and this isn't the case. It isn't possible to send null, as the TS gRPC library will not accept a null value, though from a regular Java gRPC library, that is probably what we would do here.

Tested by Devin on his demo server.

Fixes #2095

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
@@ -945,9 +945,9 @@ public class WorkerConnection {
             // table's creation, which can race this
             BiDiStream<FlightData, FlightData> stream = this.<FlightData, FlightData>streamFactory().create(
                     headers -> flightServiceClient.doPut(headers),
-                    (firstPayload, headers) -> browserFlightServiceClient.openDoPut(firstPayload, headers),
-                    (nextPayload, headers, callback) -> browserFlightServiceClient.nextDoPut(nextPayload, headers,
-                            callback::apply));
+                    (first, headers) -> browserFlightServiceClient.openDoPut(first, headers),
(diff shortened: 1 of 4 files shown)


mofojed:
@niloc132 I think you mean Fixes #2095 ?

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -945,9 +945,9 @@ public BrowserHeaders metadata() {
             // table's creation, which can race this
             BiDiStream<FlightData, FlightData> stream = this.<FlightData, FlightData>streamFactory().create(
                     headers -> flightServiceClient.doPut(headers),
-                    (firstPayload, headers) -> browserFlightServiceClient.openDoPut(firstPayload, headers),
-                    (nextPayload, headers, callback) -> browserFlightServiceClient.nextDoPut(nextPayload, headers,
-                            callback::apply));
+                    (first, headers) -> browserFlightServiceClient.openDoPut(first, headers),
+                    (next, headers, callback) -> browserFlightServiceClient.nextDoPut(next, headers, callback::apply),
```
mofojed, line 949: Minor inconsistency with later, could just call `callback` `c` instead

mofojed (approved):

niloc132:
Thanks, fixing.

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
niloc132: `c` is already defined in line 891, can't declare it again

PR #2100, opened 2022-03-14
Java Client & W2W: increase default inbound message size

### LINKED ISSUE #2083: Support > 4MB messages in the Java Barrage Client
As a user, I want to be able to snapshot or subscribe to large tables that exceed the standard 4MB limit of the Java gRPC client.  Currently can receive the following error when the data returned is too large:

`Error detected in subscription: io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: gRPC message exceeds maximum size 4194304: 289836370`

lbooker42:
Implemented a command line parameter `--max-inbound-message-size` in the Java Client examples to allow limiting the size of inbound messages.  When omitted, the maximum message size is set to `100MB`

Also increased the default limits in `BarrageTableResolver` to 100MB and added a configurable option `BarrageTableResolver.maxInboundMessageSize` to allow changes from default.

Closes #2083


nbauernfeind (dismissed):

devinrsmith:
Happy to make this change, but I don't think it's exhaustive - it only effects the examples. If we want w2w as well, we'll need to make sure to apply a limit at `io.deephaven.server.uri.BarrageTableResolver#newSession(io.deephaven.uri.DeephavenTarget)` is well.

niloc132:
I think we might possibly also consider a smaller limit? Granted, this limits what the server can send the client, and once you connect to a server, you _more or less_ trust it, but it is possible we should have some kind of limit to that trust?

We currently limit the server to only accepting 100mb max size payloads from client.

lbooker42:
I think @niloc132 is correct, should limit the default size to 100MB but we should probably enable `Barrage` batching by default (`BarrageStreamGenerator.batchSize` defaults to `Integer.MAX_SIZE`) or we have no way to get data from a large table

rcaudy:
I'm concerned that constraining the size of the payload in bytes is just going to create issues for users. We don't have a good plan for how to constrain the server to stay under a target size in *bytes*.
We can have default batch sizes that are small enough to *hopefully* stay under a size constraint, but even that is best effort.

rcaudy:
I wonder if the code we need to use to determine if we're exceeding array size constraints (and batch accordingly) can be adapted to also track payload size in bytes.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
```diff
@@ -118,7 +118,7 @@ public void onNext(final BarrageMessage barrageMessage) {
                 }
 
                 // override server-supplied data and regenerate local rowsets
-                barrageMessage.rowsAdded = RowSetFactory.flat(barrageMessage.length);
+                barrageMessage.rowsAdded = RowSetFactory.flat(barrageMessage.rowsIncluded.size());
```
rcaudy, line 121: Don't we need to close the original `rowsAdded` and `rowsIncluded` if we're going to overwrite the pointers here?

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -797,6 +799,60 @@ private ByteBuffer getSubscriptionMetadata(final SubView view) throws IOExceptio
...
+
+        int effectiveViewportOffset = 0;
+        if (view.isViewport()) {
+            try (final RowSetGenerator viewportGen = new RowSetGenerator(view.viewport)) {
+                effectiveViewportOffset = viewportGen.addToFlatBuffer(metadata);
+            }
+        }
+
+        int effectiveColumnSetOffset = 0;
+        if (isSnapshot && view.subscribedColumns != null) {
```
rcaudy, line 813: `isSnapshot` is kind of implied here.

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -797,6 +799,60 @@ private ByteBuffer getSubscriptionMetadata(final SubView view) throws IOExceptio
...
+            effectiveColumnSetOffset = new BitSetGenerator(view.subscribedColumns).addToFlatBuffer(metadata);
+        }
+
+        final int rowsAddedOffset = rowsAdded.addToFlatBuffer(metadata);
+        final int rowsRemovedOffset = 0;
+
+        final int shiftDataOffset = shifted.addToFlatBuffer(metadata);
```
rcaudy, line 820: We can't have shifts, right?

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -797,6 +799,60 @@ private ByteBuffer getSubscriptionMetadata(final SubView view) throws IOExceptio
...
+        // Added Chunk Data:
+        int addedRowsIncludedOffset = rowsIncluded.addToFlatBuffer(metadata);
+
+        final int nodesOffset = 0;
+
+        BarrageUpdateMetadata.startBarrageUpdateMetadata(metadata);
+        BarrageUpdateMetadata.addNumAddBatches(metadata,
+                LongSizedDataStructure.intSize("BarrageStreamGenerator", view.numAddBatches));
+        BarrageUpdateMetadata.addNumModBatches(metadata,
+                LongSizedDataStructure.intSize("BarrageStreamGenerator", 0));
```
rcaudy, line 831: Don't need the `intSize` constraint here.

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -797,6 +799,60 @@ private ByteBuffer getSubscriptionMetadata(final SubView view) throws IOExceptio
...
+        BarrageUpdateMetadata.addIsSnapshot(metadata, isSnapshot);
```
rcaudy, line 832: I think there's something to be said for hard-coding constants that can never be changed.

rcaudy (approved):

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -810,36 +810,31 @@ private ByteBuffer getSnapshotMetadata(final SnapshotView view) throws IOExcepti
...
 
         int effectiveColumnSetOffset = 0;
-        if (isSnapshot && view.subscribedColumns != null) {
+        if (view.subscribedColumns != null) {
             effectiveColumnSetOffset = new BitSetGenerator(view.subscribedColumns).addToFlatBuffer(metadata);
         }
 
         final int rowsAddedOffset = rowsAdded.addToFlatBuffer(metadata);
-        final int rowsRemovedOffset = 0;
+        final int addedRowsIncludedOffset = rowsIncluded.addToFlatBuffer(metadata);
```
rcaudy, line 818: Relative to `getSubscriptionMetadata`, we are adding `rowsIncluded` out-of-order w.r.t. `shifted`. Is this a bug?
Related, do we need to do this if `!view.isViewport()`?

niloc132: out-of-order: I believe this is correct (or at least permissible) - the data just needs to have been put in the flatbuffer before we first reference it, the actual order of the data doesn't matter.

lbooker42: Made changes to `getSnapshotMetadata()` to re-order shifts / rowIncluded to match exactly the `barrage.fbs` definition.  Also re-ordered `BarrageUpdateMetadata.addEffectiveReverseViewport()` calls.  Now we are consistent with the definition.

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -829,12 +834,12 @@ private ByteBuffer getSnapshotMetadata(final SnapshotView view) throws IOExcepti
         BarrageUpdateMetadata.addLastSeq(metadata, lastSeq);
         BarrageUpdateMetadata.addEffectiveViewport(metadata, effectiveViewportOffset);
         BarrageUpdateMetadata.addEffectiveColumnSet(metadata, effectiveColumnSetOffset);
-        BarrageUpdateMetadata.addEffectiveReverseViewport(metadata, view.reverseViewport);
         BarrageUpdateMetadata.addAddedRows(metadata, rowsAddedOffset);
         BarrageUpdateMetadata.addRemovedRows(metadata, 0);
         BarrageUpdateMetadata.addShiftData(metadata, shiftDataOffset);
         BarrageUpdateMetadata.addAddedRowsIncluded(metadata, addedRowsIncludedOffset);
         BarrageUpdateMetadata.addModColumnNodes(metadata, 0);
+        BarrageUpdateMetadata.addEffectiveReverseViewport(metadata, view.reverseViewport);
```
rcaudy, line 842: Is this pair of changes just to line up with message field order?

rcaudy (approved):

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
lbooker42: Yes, we might as well be accurate in the order of the FlatBuffers definition.

PR #2105, opened 2022-03-18
jpy vararg leak

devinrsmith:
DHC targeted version of https://github.com/jpy-consortium/jpy/pull/22


cpwright (approved):

devinrsmith:
QA script:

```python
system=jpy.get_type("java.lang.System")
objects=jpy.get_type("java.util.Objects")
import gc
a=jpy.array("int", 100000000)
objects.hash(a)
a=None
gc.collect()
system.gc()
```

On a 4G VM, I was able to run this a few times and notice the heap grow, until it eventually ran out of memory. After this patch, I can qualitatively confirm the same leak does not appear to exist.

jmao-denver (approved):
LGTM

PR #2107, opened 2022-03-21
Pyclient groovy support

jakemulf:
This unblocks https://github.com/deephaven/action-assert-dh-tables-present/issues/3 and other similar issues with trying to use the pyclient with a groovy deephaven instance

### DIFF
--- a/pyclient/pydeephaven/session.py
+++ b/pyclient/pydeephaven/session.py
@@ -71,7 +71,7 @@ class Session:
 
         self.session_type = session_type
         if session_type is None:
-            session_type = 'python'
+            self.session_type = 'python'
 
         self._connect()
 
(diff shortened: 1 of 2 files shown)


jmao-denver (dismissed):
This change makes sense and makes it possible to support using other languages that DHC chooses to support in the future.

pyclient/pydeephaven/session.py
```diff
@@ -67,6 +68,7 @@ def __init__(self, host: str = None, port: int = None, never_timeout: bool = Tru
         self._tables = {}
         self._never_timeout = never_timeout
         self._keep_alive_timer = None
+        self.session_type = session_type
```
devinrsmith, line 71: Should we try to stick to an underscore for internal variables coding convention?

jakemulf: Yeah that's fine, updated

devinrsmith (approved):

PR #2111, opened 2022-03-22
Revert "jpy vararg leak"

niloc132:
This fix seems to be causing JVM crashes in at least one deephaven2 test.
Reverts deephaven/deephaven-core#2105


devinrsmith:
Note: this was reverted due to some NPEs. Diagnosing now.

PR #2119, opened 2022-03-24
Remove an empty duplicate file

niloc132:
Due to classpath hiding, this could prevent the real copy of
1-performance.groovy in :engine-table from loading its contents
correctly at runtime.


devinrsmith:
Looks like we need to kick this PR to get CI to run?

niloc132:
That's weird, I'll rebase and force push.

devinrsmith (approved):

PR #2123, opened 2022-03-25
Add SafePoint times to UpdateGraphProcessor cycle times logging.

jcferretti:
Given Charles comments (below), I've re-done the implementation based on `HotspotRuntimeMBean` instead of `GarbageCollectorMXBean`.

The information from `HotspotRuntimeMBean` contains count and times for stop the world pauses due to safepointing; every GC uses a safepoint, not every safepoint is GC.

See
* On GC Pauses: https://stackoverflow.com/questions/50380836/what-is-the-best-way-to-measure-gc-pause-times
* On `HotspotRuntimeMBean`: https://stackoverflow.com/questions/42514817/printgcapplicationstoppedtime
* On Safepoints: https://stackoverflow.com/questions/29666057/analyzing-gc-logs
* On all causes for Safepoints: http://hg.openjdk.java.net/jdk8u/jdk8u/hotspot/file/fc3cd1db10e2/src/share/vm/runtime/vm_operations.hpp#l39.

--

Logging HotSpot Safepoint times using data from HotSpot introspection (see https://stackoverflow.com/questions/42514817/printgcapplicationstoppedtime/42515743#42515743)
Thanks to Devin for the help hooking up this in a way that
(1) Works for both jdk 11 and 17.
(2) Does not overly burden the system with additional dependencies: we are adding as an optional a service; if not present is not used.

Also format millis output to a fixed number of decimal places.

This is how it looked before:

```
server_1        | 2022-03-24T20:51:10.289Z | EFAULT.refreshThread |  INFO | .e.u.UpdateGraphProcessor | Minimal Update Graph Processor cycle times: 18.871795ms / 27 cycles = 0.6989553703703704ms/cycle average
server_1        | 2022-03-24T20:51:10.289Z | EFAULT.refreshThread |  INFO | .e.u.UpdateGraphProcessor | Update Graph Processor cycleTime=3341.797276ms, lockWaitTime=3112.12555ms, yieldTime=0.0ms, sleepTime=0.0ms
```

This is how it looks now:

```
server_1        | 2022-03-30T03:17:42.926Z | EFAULT.refreshThread |  INFO | .e.u.UpdateGraphProcessor | Minimal Update Graph Processor cycle times: 18.203ms / 27 cycles = 0.674ms/cycle average), safePointTime=0ms
server_1        | 2022-03-30T03:17:42.926Z | EFAULT.refreshThread |  INFO | .e.u.UpdateGraphProcessor | Update Graph Processor cycleTime=3296.973ms, safePointTime=136ms, safePointTimePct=4.12%, safePointSyncTime=4ms, lockWaitTime=3051.758ms, yieldTime=0ms, sleepTime=0ms
server_1        | 2022-03-30T03:17:43.940Z | EFAULT.refreshThread |  INFO | .e.u.UpdateGraphProcessor | Minimal Update Graph Processor cycle times: 30.103ms / 2 cycles = 15.051ms/cycle average), safePointTime=0ms
server_1        | 2022-03-30T03:17:45.778Z | EFAULT.refreshThread |  INFO | .e.u.UpdateGraphProcessor | Update Graph Processor cycleTime=852.066ms, safePointTime=86ms, safePointTimePct=10.09%, safePointSyncTime=3ms, lockWaitTime=0.132ms, yieldTime=0ms, sleepTime=0ms
server_1        | 2022-03-30T03:17:46.586Z | EFAULT.refreshThread |  INFO | .e.u.UpdateGraphProcessor | Update Graph Processor cycleTime=659.362ms, safePointTime=70ms, safePointTimePct=10.62%, safePointSyncTime=3ms, lockWaitTime=0.188ms, yieldTime=0ms, sleepTime=0ms
server_1        | 2022-03-30T03:17:47.433Z | EFAULT.refreshThread |  INFO | .e.u.UpdateGraphProcessor | Update Graph Processor cycleTime=507.558ms, safePointTime=22ms, safePointTimePct=4.33%, safePointSyncTime=0ms, lockWaitTime=0.198ms, yieldTime=0ms, sleepTime=0ms
server_1        | 2022-03-30T03:17:48.464Z | EFAULT.refreshThread |  INFO | .e.u.UpdateGraphProcessor | Update Graph Processor cycleTime=538.352ms, safePointTime=13ms, safePointTimePct=2.41%, safePointSyncTime=0ms, lockWaitTime=0.103ms, yieldTime=0ms, sleepTime=0ms
server_1        | 2022-03-30T03:17:49.465Z | EFAULT.refreshThread |  INFO | .e.u.UpdateGraphProcessor | Update Graph Processor cycleTime=538.972ms, safePointTime=24ms, safePointTimePct=4.45%, safePointSyncTime=0ms, lockWaitTime=0.182ms, yieldTime=0ms, sleepTime=0ms
```

The idea for computing the number of base 10 digits for an integer came from here: https://stackoverflow.com/questions/25892665/performance-of-log10-function-returning-an-int


engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
```diff
@@ -1485,19 +1501,35 @@ private void refreshTablesAndFlushNotifications() {
...
+                        .append("Update Graph Processor cycleTime=").appendPositiveDouble(cycleTimeMillis, 3)
+                        .append("ms, gcTime=").append(gcTimeMillis)
+                        .append("ms, gcTimePct=");
+                if (gcTimeMillis > 0 && cycleTimeMillis > 0.0) {
+                    final double gcTimePct = 100.0 * gcTimeMillis / cycleTimeMillis;
+                    entry = entry.appendPositiveDouble(gcTimePct, 2);
+                } else {
+                    entry = entry.append("0");
+                }
+                entry = entry.append("%, lockWaitTime=");
```
cpwright, line 1522: After your if up there, I think I would go back to the standard chained append calls.

cpwright (commented):
The decimal printout stuff looks nice and improves the output.  What I am not sure that I have my head around is if we are actually reporting the right thing.  We will report the amount of GC that occurred during a cycle.  However, I don't know if that is the most useful metric or really what we should report at all.

I think that separating STW vs. other collection is important.  I also think that we're ignoring everything outside the cycle; which may or may not matter.  For non-STW collections,  I don't know how the things will accumulate when something starts or ends cross a cycle boundary.  Will we sometimes have zero if we start but don't finish; and other times have > 100% when we finish but had already started.

Tracking GC separately using the existing GC logs may be a better mechanism than what is being done here.  @rcaudy is better at GC tuning and investigation than I am.  Also Raffi.

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
jcferretti: We can't call appendAsMillisFromNanos as part of a chain.

Util/src/main/java/io/deephaven/util/JvmIntrospectionContext.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Utility class to facilitate obtaining data for safepoint pauses count and time between two points in code. A
+ * safepoint pause is a "stop the world, pause all threads" event in the HotSpot JVM. Note full Garbage Collection
+ * pauses are a dominant cause of safepoint pauses, but there are other triggers like: <il>
+ * <li>Deoptimization</li>
+ * <li>Biased lock revocation</li>
+ * <li>Thread dump</li>
+ * <li>Heap inspection< /li>
+ * <li>Class redefinition</li> </il> And others; you can see a full list <a href=
+ * "http://hg.openjdk.java.net/jdk8u/jdk8u/hotspot/file/fc3cd1db10e2/src/share/vm/runtime/vm_operations.hpp#l39"> here
```
devinrsmith, line 14: Seems weird to have a reference to java 8 source; maybe find equivalent for 11?

hotspot/src/main/java/io/deephaven/hotspot/HotSpotImpl.java
```diff
@@ -0,0 +1,19 @@
...
+
+    @Override
+    public long getSafepointCount() {
+        return ManagementFactoryHelper.getVMManagement().getSafepointCount();
+    }
+
+    @Override
+    public long getTotalSafepointTimeMillis() {
+        return ManagementFactoryHelper.getVMManagement().getTotalSafepointTime();
+    }
```
devinrsmith, line 18: How should we choose to fail when the user doesn't add `add-opens`? As written now, the failure will occur at method-call time. It may be better to capture `getVMManagement()` during the construction of `HotSpotImpl`.

settings.gradle
```diff
@@ -285,6 +285,9 @@ project(':docker-web-plugin-packager').projectDir = file('docker/web-plugin-pack
 include(':docker-server-slim')
 project(':docker-server-slim').projectDir = file('docker/server-slim')
 
+include(':hotspot')
```
devinrsmith, line 288: I'm wondering if we want a more specific package name (I choose "hotspot" without much thought). `hotspot` seems to imply more of a library, rather than an implementation. We may want to extract `HotSpot` interface outside of `engine-util` in the future(or now?), and `hotspot` might be the best name for that. For the implementation, maybe `hotspot-impl`?

devinrsmith (changes requested):
I didn't review any of the logical changes to UGP. I think we need some more polish here before merging.

Util/src/main/java/io/deephaven/util/hotspot/HotSpot.java
```diff
@@ -0,0 +1,25 @@
...
+        final HotSpot impl = it.next();
+        if (it.hasNext()) {
+            throw new IllegalStateException("Found multiple implementations for " + HotSpot.class.getSimpleName());
+        }
+        return Optional.of(impl);
+    }
+
+    long getSafepointCount();
+
+    long getTotalSafepointTimeMillis();
```
rcaudy, line 24: Should these have JavaDoc?

Util/src/main/java/io/deephaven/util/JvmIntrospectionContext.java
```diff
@@ -0,0 +1,69 @@
...
+    public void startSample() {
+        if (hotspot == null) {
+            return;
+        }
+        lastStartPausesCount = hotspot.getSafepointCount();
+        lastStartPausesTimeMillis = hotspot.getTotalSafepointTimeMillis();
+    }
+
+    /**
+     * Sample garbage collection count and times at the point of call.
```
rcaudy, line 42: The JavaDocs in this class are inaccurate, maybe they stayed the same while the class evolved?

hotspot/src/main/java/io/deephaven/hotspot/HotSpotImpl.java
```diff
@@ -0,0 +1,19 @@
...
+import com.google.auto.service.AutoService;
+import io.deephaven.util.hotspot.HotSpot;
+import sun.management.ManagementFactoryHelper;
+
+@AutoService(HotSpot.class)
+public class HotSpotImpl implements HotSpot {
+
+    @Override
+    public long getSafepointCount() {
+        return ManagementFactoryHelper.getVMManagement().getSafepointCount();
```
rcaudy, line 12: I spent a little bit of time looking at the (internal, Sun-specific) APIs here, and I think we should do this slightly differently. See:
`sun.management.ManagementFactoryHelper#getHotspotRuntimeMBean`
It's labeled "This method is for testing only."
That said, it produces an MBean with an actual documented interface (`sun.management.HotspotRuntimeMBean`). It delegates internally to the same object, but using this class gets us on board with actual code samples on the Internet (I can't find any that are directly using `VMManagement`).

I assume we have adequate guards to make sure we don't try to do this on non-Hotspot VMs, and that @devinrsmith is looking at that.

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
```diff
@@ -1485,19 +1501,42 @@ private void refreshTablesAndFlushNotifications() {
...
                 }
-                log.info().append("Update Graph Processor cycleTime=").appendDouble(cycleTime / 1_000_000.0)
-                        .append("ms, lockWaitTime=").appendDouble(currentCycleLockWaitTotalNanos / 1_000_000.0)
-                        .append("ms, yieldTime=").appendDouble(currentCycleYieldTotalNanos / 1_000_000.0)
-                        .append("ms, sleepTime=").appendDouble(currentCycleSleepTotalNanos / 1_000_000.0)
-                        .append("ms").endl();
-            } else if (cycleTime > 0) {
-                suppressedCycles++;
-                suppressedCyclesTotalNanos += cycleTime;
+                final long safePointPauseTimeMillis = jvmIntrospectionContext.deltaSafePointPausesTimeMillis();
```
rcaudy, line 1510: We might have reached the point where this logging code adds enough bulk that it should be pulled into a new method, as `logSuppressedCycles` was long ago.

rcaudy (commented):
My review is mostly at a high level. Contrary to expectations, I need not do a line-by-line review. I'll address overall context in the Slack channel.

IO/src/main/java/io/deephaven/io/log/LogEntry.java
```diff
@@ -93,54 +93,90 @@ default LogEntry appendDouble(Double f) {
...
+                --actualDecimalPlaces;
+            }
+            if (fractional == 0) {
+                return entry;
+            }
+        }
+        entry = entry.append(".");
         final int base10FractionalDigits = MathUtil.base10digits(fractional);
-        final int leadingZeroes = decimalPlaces - base10FractionalDigits;
+        final int leadingZeroes = actualDecimalPlaces - base10FractionalDigits;
```
cpwright, line 147: What prevents someone from being a jerk and telling you they want 20 decimal places and then getting zero of them?

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
```diff
@@ -1485,19 +1501,42 @@ private void refreshTablesAndFlushNotifications() {
...
+                final double cycleTimeMillis = cycleTimeNanos / 1_000_000.0;
+                LogEntry entry = log.info()
+                        .append("Update Graph Processor cycleTime=").appendDoubleToDecimalPlaces(cycleTimeMillis, 3);
+                if (JvmIntrospectionContext.hasSafePointData()) {
+                    entry = entry
+                            .append("ms, safePointTime=")
+                            .append(safePointPauseTimeMillis)
+                            .append("ms, safePointTimePct=");
+                    if (safePointPauseTimeMillis > 0 && cycleTimeMillis > 0.0) {
+                        final double gcTimePct = 100.0 * safePointPauseTimeMillis / cycleTimeMillis;
```
cpwright, line 1520: It is not gcTimePct anymore.

buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
```diff
@@ -64,15 +64,12 @@ tasks.withType(JavaCompile).configureEach {
   options.compilerArgs << '-parameters'
 
   if (!excludeReleaseFlag) {
-    if (name == 'compileTestJava') {
-      if (compilerVersion != testLanguageLevel) {
+    def checkLevel = (name == 'compileTestJava') ? testLanguageLevel : languageLevel
+    if (compilerVersion != checkLevel) {
         options.release.set testLanguageLevel
```
devinrsmith, line 69: Assuming you mean `options.release.set checkLevel`?

buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
```diff
@@ -64,15 +64,12 @@ tasks.withType(JavaCompile).configureEach {
...
-      }
-    } else {
-      if (compilerVersion != languageLevel) {
-        options.release.set languageLevel
-      }
     }
+  } else {
+    sourceCompatibility = languageLevel
```
devinrsmith, line 72: Let's make a comment here that this is a special case - I'm not sure why add-exports doesn't work w/o this (while the failure w/ --release makes a bit of sense).

hotspot/build.gradle
```diff
@@ -0,0 +1,18 @@
...
+dependencies {
+    compile project(':log-factory')
+
+    testRuntime project(path: ':configs'), project(path: ':test-configs')
+
+    testRuntimeOnly project(':log-to-slf4j')
+    Classpaths.inheritSlf4j(project, 'slf4j-simple', 'testRuntimeOnly')
+
+    Classpaths.inheritJUnitPlatform(project)
+}
```
devinrsmith, line 18: We should not use compile or testRuntime dependencies for new projects. I don't think these dependencies are necessary?

server/jetty/build.gradle
```diff
@@ -10,6 +10,8 @@ dependencies {
     Classpaths.inheritDagger(project, /* test */ true)
 
     Classpaths.inheritGrpcPlatform(project)
+    runtimeOnly project(':hotspot-impl')
+
```
devinrsmith, line 14: We can't have this unguarded - this was the part that was supposed to be in the hasProperty guard.

server/netty/build.gradle
```diff
@@ -17,6 +17,7 @@ dependencies {
     Classpaths.inheritGrpcPlatform(project)
     implementation 'io.grpc:grpc-netty'
 
+    runtimeOnly project(':hotspot-impl')
```
devinrsmith, line 20: ditto.

devinrsmith (changes requested):
I'll be going through this to try and support our hotspot-impl in docker use case.

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
```diff
@@ -1550,15 +1513,57 @@ private void refreshTablesAndFlushNotifications() {
         waitForNextCycle(startTime, sched);
     }
 
+    private void logCycle(final long cycleTimeNanos) {
+        if (cycleTimeNanos >= minimumCycleDurationToLogNanos) {
+            if (suppressedCycles > 0) {
+                logSuppressedCycles();
+            }
+            final long safePointPauseTimeMillis = jvmIntrospectionContext.deltaSafePointPausesTimeMillis();
+            final long safePointSyncTimeMillis = jvmIntrospectionContext.deltaSafePointSyncTimeMillis();
```
rcaudy, line 1522: Should these two lines go inside the `if (jvmIntrospectionContext.hasSafePointData()) {` block?

hotspot-impl/src/main/java/io/deephaven/hotspot/impl/HotSpotImpl.java
```diff
@@ -0,0 +1,30 @@
+package io.deephaven.hotspot.impl;
+
+import com.google.auto.service.AutoService;
+import io.deephaven.hotspot.HotSpot;
+import sun.management.ManagementFactoryHelper;
+import sun.management.HotspotRuntimeMBean;
+
+@AutoService(HotSpot.class)
+public class HotSpotImpl implements HotSpot {
+    private HotspotRuntimeMBean hotspotRuntimeMBean;
```
rcaudy, line 10: `final`

niloc132 (approved):
Thoughts on instead renaming the existing server-jetty/netty to server-jetty-lib etc, and leaving server-jetty as the app itself?

I could see either way here, and have just about convinced myself it is better with this explicit rename (so that the libraries are easier to use downstream), only real complaint is that ./gradlew :server-jetty-app:run is long to get it running - but we really should just make a top-level task if we decide this is important and something people will do.

I also think we should re-add the readmes for netty/jetty, but seems like we should do that separately.

Approved with one comment that I think would improve readability.

devinrsmith (approved):

PR #2131, opened 2022-03-28
Add nested JSON support to KafkaTools.consumeToTable.

jcferretti:
We keep the API fully backwards compatible by extending the support in the `mapping` argument to the JSON spec: before the mapping was a map from a key of JSON (toplevel) field name to a value of DH table column name.  Now the key can be either a toplevel field name or it can be a JSON Pointer to a node (the [spec](https://www.rfc-editor.org/rfc/rfc6901) has some nuance, but essentially a JSON pointer string looks like an absolute path in a filesystem, eg, `/User/FirstName`).

Example:

```
cfs@erke 17:39:22 ~/dh/oss1/deephaven-core/redpanda/examples/python
$ python3 kafka-produce.py orders 0 '' 'str:{ "Symbol" : "MSFT", "Side" : "BUY", "Price" : "278.85", "Qty" : "200", "User" : { "Name" : "Pepo", "Id" : "1234" } }'
Message key=|b''|, value=|b'{ "Symbol" : "MSFT", "Side" : "BUY", "Price" : "278.85", "Qty" : "200", "User" : { "Name" : "Pepo", "Id" : "1234" } }'| delivered to topic orders partition 0
(confluent-kafka)
```
dh_nested_json

### DIFF
--- a/extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
+++ b/extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
@@ -101,9 +101,10 @@ public class KafkaTools {
     public static final String BYTE_BUFFER_SERIALIZER = ByteBufferSerializer.class.getName();
     public static final String AVRO_SERIALIZER = KafkaAvroSerializer.class.getName();
     public static final String SERIALIZER_FOR_IGNORE = BYTE_BUFFER_SERIALIZER;
-    public static final String NESTED_FIELD_NAME_SEPARATOR = "__";
+    public static final String NESTED_FIELD_NAME_SEPARATOR = ".";
     private static final Pattern NESTED_FIELD_NAME_SEPARATOR_PATTERN =
             Pattern.compile(Pattern.quote(NESTED_FIELD_NAME_SEPARATOR));
+    public static final String NESTED_FIELD_COLUMN_NAME_SEPARATOR = "__";
     public static final String AVRO_LATEST_VERSION = "latest";
 
     private static final Logger log = LoggerFactory.getLogger(KafkaTools.class);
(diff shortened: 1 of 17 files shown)


extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -575,19 +576,50 @@ DataFormat dataFormat() {
...
                 @Override
                 DataFormat dataFormat() {
                     return DataFormat.JSON;
                 }
+
+                private static Map<String, String> mapNonPointers(final Map<String, String> fieldNameToColumnName) {
+                    if (fieldNameToColumnName == null) {
+                        return null;
+                    }
+                    boolean needsMapping = false;
```
cpwright, line 597: This actually seems like a case streams would be clearer:
final boolean needsMapping = fieldNameToColumnName.keySet().anymatch(key -> key.startsWith("/"));

This is not necessary, but we have very similar loops to parse out.

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -575,19 +576,50 @@ DataFormat dataFormat() {
...
+                        if (key.startsWith("/")) {
+                            ans.put(key, entry.getValue());
+                        } else {
+                            ans.put(mapTopLevelFieldNameToJsonPointerStr(key), entry.getValue());
+                        }
+                    }
+                    return ans;
+                }
+
+                public static String mapTopLevelFieldNameToJsonPointerStr(final String key) {
```
cpwright, line 620: Because the characters '~' (%x7E) and '/' (%x2F) have special
   meanings in JSON Pointer, '~' needs to be encoded as '~0' and '/'
   needs to be encoded as '~1' when these characters appear in a
   reference token.

cpwright (commented):
The comment about streams can be ignored, but we should add the comment in the comment.

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
jcferretti: Done.

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
jcferretti: Documented.

cpwright (dismissed):

PR #2132, opened 2022-03-28
Clean up tables correctly.

### LINKED ISSUE #2130: Kafka subscription / table not cleaned up after deleting table
It appears that deleting a kafka table does not free the resources of the underlying kafka subscription or table. Demonstrated in python (will try in groovy soon).

devinrsmith:
Fixes #2130

Co-authored-by: Ryan Caudy <ryan@deephaven.io>


devinrsmith:
I'm not confident there aren't other bugs in the ApplicationServiceGrpcImpl.

There are bifurcated paths out of schedulePropagationOrClearIncrementalState, and they don't seem to be equivalent. The job side seems to be missing out on clearing out knownFieldMap; and the other side seems to be missing out on maybeUnmanage. I've just added an extra remove for knownFieldMap, but there may need to be deeper look into liveness for both paths to ensure equivalent. (@nbauernfeind)

devinrsmith:
On another note, it seems like we need *two* full GCs to actually clear out Kafka tables.

devinrsmith:
Superseded by #2158

PR #2146, opened 2022-03-30
Update object allocation stats so it works, but disabled by default

### LINKED ISSUE #2120: Investigate java-allocation-instrumenter, ensure it works as expected
While the server cannot startup without this dependency, we are not setting it up on the bootclasspath as instructed, so it probably doesn't function.

We should disable the `allocation.stats.enabled` flag by default if it isn't actually functioning, and make the dependency compileOnly if if actually doesn't function.

If it does function, we should force an upgrade from the asm groupId to the modern org.ow2.asm groupId.

niloc132:
Also adds guava as an explicit dependency to other modules, which
previously implicitly picked up guava from the old
java-allocation-tracker library version.

Fixes #2120

### DIFF
--- a/Plot/Plot.gradle
+++ b/Plot/Plot.gradle
@@ -20,6 +20,8 @@ dependencies {
 
     compileOnly 'javax.inject:javax.inject:1'
 
+    implementation 'com.google.guava:guava:19.0'
+
     testCompile TestTools.projectDependency(project, 'Util'),
                 TestTools.projectDependency(project, 'engine-table')
 
--- a/Stats/src/main/java/io/deephaven/stats/ObjectAllocationCollector.java
+++ b/Stats/src/main/java/io/deephaven/stats/ObjectAllocationCollector.java
@@ -12,13 +12,14 @@ import io.deephaven.base.stats.State;
 import io.deephaven.base.stats.Stats;
 import io.deephaven.base.stats.Value;
 
+import java.util.Objects;
+
 /**
- * Use the allocation instrumenter from http://code.google.com/p/java-allocation-instrumenter/ to produce stats
+ * Use the allocation instrumenter from https://github.com/google/allocation-instrumenter to produce stats
  *
  * To use make sure you set
  *
- * -j -Dallocation.stats.enabled=true -j -Xbootclasspath/a:java_lib/allocation-2.0.jar -j
- * -javaagent:java_lib/allocation-2.0.jar (maybe even use -j -noverify)
+ * -Dallocation.stats.enabled=true -javaagent:/opt/deephaven/server/lib/java-allocation-instrumenter-3.3.0.jar
  */
 public class ObjectAllocationCollector {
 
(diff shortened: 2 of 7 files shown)


Plot/Plot.gradle
```diff
@@ -20,6 +20,8 @@ dependencies {
 
     compileOnly 'javax.inject:javax.inject:1'
 
+    implementation 'com.google.guava:guava:19.0'
```
devinrsmith, line 23: I know this was depended on implicitly before, but I'm wondering if we want to try and remove our dependency on guava, or update to a more recent version? (19.0 from 2015)

https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-8908
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-10237

At the very least, we should have a follow-up ticket.

Stats/src/main/java/io/deephaven/stats/ObjectAllocationCollector.java
```diff
@@ -12,13 +12,14 @@
...
+
 /**
- * Use the allocation instrumenter from http://code.google.com/p/java-allocation-instrumenter/ to produce stats
+ * Use the allocation instrumenter from https://github.com/google/allocation-instrumenter to produce stats
  *
  * To use make sure you set
  *
- * -j -Dallocation.stats.enabled=true -j -Xbootclasspath/a:java_lib/allocation-2.0.jar -j
- * -javaagent:java_lib/allocation-2.0.jar (maybe even use -j -noverify)
+ * -Dallocation.stats.enabled=true -javaagent:/opt/deephaven/server/lib/java-allocation-instrumenter-3.3.0.jar
```
devinrsmith, line 22: This comment implies we are adding the jar to our classpath - but since it's compileOnly, we *aren't*; yeah?

devinrsmith (commented):
May approve as-is, depending on follow-up comments.

Plot/Plot.gradle
niloc132: Yep, this was picked to ensure we don't change our current version. 

We are not susceptible to either cve, I actually checked this while waiting for a build - we don't use the classes required by either, and for CVE-2018-10237 we don't allow guava objects to be serialized and sent to us at all at this time (w2w might support this in theory some day, but java serialization is already a minefield, and we would want to only enable that feature for trusted other servers).

Stats/src/main/java/io/deephaven/stats/ObjectAllocationCollector.java
niloc132: Making this compileOnly makes the feature opt-in, and we want it off by default - without the javaagent set, it doesnt matter if you turn enable `allocation.stats.enabled` or not, nothing will be recorded).

The act of passing javaagent both a) adds that (shaded) jar to the classpath and b) enables the jar's premain to run and instrument the real classpath. As a side effect of b), we will see the jar from our own classpath.

The path in that documentation string is just a hypothetical, but we don't ship the jar (or its ... incorrect dependency tree) at all - I have a comment on #2120 that describes how to actually enable this feature.

devinrsmith (approved):
:+1:

PR #2157, opened 2022-03-30
Don't remove subscriptions while iterating them

### LINKED ISSUE #2070: propagateUpdates CME
```
java.util.ConcurrentModificationException: null
        at java.base/java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:719)
        at java.base/java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:741)
        at java.base/java.lang.Iterable.forEach(Iterable.java:74)
        at io.deephaven.server.appmode.ApplicationServiceGrpcImpl.propagateUpdates(ApplicationServiceGrpcImpl.java:212)
        at io.deephaven.server.appmode.ApplicationServiceGrpcImpl$FieldUpdatePropagationJob.run(ApplicationServiceGrpcImpl.java:298)
        at io.deephaven.server.runner.DeephavenApiServerModule$ThreadFactory.lambda$newThread$0(DeephavenApiServerModule.java:143)
```

niloc132:
Fixes #2070

### DIFF
--- a/server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
+++ b/server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
@@ -23,6 +23,7 @@ import io.grpc.stub.StreamObserver;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import java.io.Closeable;
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -30,7 +31,6 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 @Singleton
 public class ApplicationServiceGrpcImpl extends ApplicationServiceGrpc.ApplicationServiceImplBase
@@ -122,8 +122,10 @@ public class ApplicationServiceGrpcImpl extends ApplicationServiceGrpc.Applicati
         accumulated.clear();
         if (!updater.isEmpty() && !subscriptions.isEmpty()) {
             final FieldsChangeUpdate update = updater.build();
-            List<Subscription> subscriptionsToCancel = subscriptions.stream().filter(sub -> !sub.send(update)).collect(Collectors.toList());
-            subscriptionsToCancel.forEach(Subscription::onCancel);
+            // Send updates to all subscriptions, if they fail to handle the update, cancel the subscription
+            List<Subscription> toCancel = new ArrayList<>(subscriptions);
+            toCancel.removeIf(s -> s.send(update));
+            toCancel.forEach(Subscription::onCancel);
         }
     }
 
--- a/server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
+++ b/server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
@@ -25,10 +25,12 @@ import javax.inject.Singleton;
 import java.io.Closeable;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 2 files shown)


nbauernfeind (dismissed):

server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
```diff
@@ -209,7 +210,8 @@ private synchronized void propagateUpdates() {
         updatedFields.clear();
         final FieldsChangeUpdate update = builder.build();
 
-        subscriptions.forEach(sub -> sub.send(update));
+        List<Subscription> subscriptionsToCancel = subscriptions.stream().filter(sub -> sub.send(update)).collect(Collectors.toList());
```
devinrsmith, line 213: filter predicates *should* be stateless ...

devinrsmith (changes requested):

server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
niloc132: Yeah - removeIf is so much clearer, thanks for the kick to just do it right.

I almost went with removeIf right away after seeing your comment, then assume that it would be equally "inappropriate" to do so, even though, like Stream.filter with a collect on the end, I know it will traverse every item. Instead I went with 
```
        for (Iterator<Subscription> iterator = subscriptions.iterator(); iterator.hasNext(); ) {
            Subscription subscription = iterator.next();
            if (!subscription.send(update)) {
                iterator.remove();
            }
        }
```
And IJ gave me a "why don't you just use Collection.removeIf" popup. Fixed.

server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
```diff
@@ -209,7 +210,7 @@ private synchronized void propagateUpdates() {
         updatedFields.clear();
         final FieldsChangeUpdate update = builder.build();
 
-        subscriptions.forEach(sub -> sub.send(update));
+        subscriptions.removeIf(subscription -> !subscription.send(update));
```
devinrsmith, line 213: Do we need any of the side-effects that onClose previously provided? (We aren't calling onClose now.)

* `session.removeOnCloseCallback(subscription)`
* `subscription.notifyObserverAborted()`

niloc132: No you're right, though the broken behavior won't be visible externally

 * the leaked callback will just leak until the session cleans it up, which isnt ideal
 * the observer notification doesnt matter, since the observer is already by definition failed

I really dont care for this close-is-remove-but-not-cancel pattern we have used here and elsewhere. I am also not sure that what we seem to have here makes sense at all (why does a close() call necessarily mean sending the aborted message, but doesnt do the cleanup from the session?), but I think that cleanup is outside the scope for this set of changes.

server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
```diff
@@ -363,12 +367,17 @@ public void close() {
             remove(this);
         }
 
-        // must be sync wrt parent
+        /**
+         * Sends an update to the subscribed client. Returns true if successful - if false, the client is no longer
+         * listening and this subscription should be canceled after iteration.
```
devinrsmith, line 372: Looking at callers of `send` - do we also need to do an onCancel call in `listFields`?

niloc132: Probably, though a) it is a weak ref anyway, and b) the subscription won't be in the list so doesnt need to be removed. I'll make the change in this patch, though it probably should be incorporated into a bigger change that generally cleans up these patterns, rather than fixing each piecemeal as we've been doing.

devinrsmith (approved):

PR #2158, opened 2022-03-30
Don't keep hard references to Table

### LINKED ISSUE #2130: Kafka subscription / table not cleaned up after deleting table

devinrsmith:
Fixes #2130

Co-authored-by: Ryan Caudy <ryan@deephaven.io>

### DIFF
--- a/engine/table/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
+++ b/engine/table/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
@@ -5,6 +5,7 @@ import io.deephaven.UncheckedDeephavenException;
 import io.deephaven.base.verify.Assert;
 import io.deephaven.chunk.attributes.Any;
 import io.deephaven.chunk.attributes.Values;
+import io.deephaven.engine.liveness.ReferenceCountedLivenessNode;
 import io.deephaven.engine.table.ColumnSource;
 import io.deephaven.engine.table.ColumnDefinition;
 import io.deephaven.engine.table.Table;
@@ -27,6 +28,7 @@ import io.deephaven.util.MultiException;
 import io.deephaven.util.SafeCloseable;
 import org.jetbrains.annotations.NotNull;
 
+import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -35,7 +37,7 @@ import java.util.Map;
 /**
  * Adapter for converting streams of data into columnar Deephaven {@link Table tables}.
  */
-public class StreamToTableAdapter implements SafeCloseable, StreamConsumer, Runnable {
+public class StreamToTableAdapter extends ReferenceCountedLivenessNode implements SafeCloseable, StreamConsumer, Runnable {
 
     private static final Logger log = LoggerFactory.getLogger(StreamToTableAdapter.class);
 
@@ -44,7 +46,7 @@ public class StreamToTableAdapter implements SafeCloseable, StreamConsumer, Runn
     private final UpdateSourceRegistrar updateSourceRegistrar;
     private final String name;
 
-    private final QueryTable table;
+    private final WeakReference<QueryTable> tableRef;
     private final TrackingWritableRowSet rowSet;
     private final SwitchColumnSource<?>[] switchSources;
 
@@ -61,6 +63,7 @@ public class StreamToTableAdapter implements SafeCloseable, StreamConsumer, Runn
     /** A list of failures that have occurred. */
     private List<Exception> enqueuedFailure;
 
+    private volatile QueryTable table;
     private volatile Runnable shutdownCallback;
     private volatile boolean alive = true;
 
@@ -68,6 +71,7 @@ public class StreamToTableAdapter implements SafeCloseable, StreamConsumer, Runn
             @NotNull final StreamPublisher streamPublisher,
             @NotNull final UpdateSourceRegistrar updateSourceRegistrar,
             @NotNull final String name) {
+        super(false);
         this.tableDefinition = tableDefinition;
         this.streamPublisher = streamPublisher;
         this.updateSourceRegistrar = updateSourceRegistrar;
@@ -91,12 +95,8 @@ public class StreamToTableAdapter implements SafeCloseable, StreamConsumer, Runn
                 setAttribute(Table.STREAM_TABLE_ATTRIBUTE, Boolean.TRUE);
                 addParentReference(StreamToTableAdapter.this);
             }
-
-            @Override
-            public void destroy() {
-                StreamToTableAdapter.this.close();
-            }
         };
+        tableRef = new WeakReference<>(table);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
```diff
@@ -67,149 +67,264 @@ public ApplicationServiceGrpcImpl(final AppMode mode,
...
 
-            // let's not duplicate information when a client does actually join
-            addedFields.clear();
-            removedFields.clear();
-            updatedFields.clear();
+        boolean removed = false;
+
+        @Override
+        CUR out() {
+            // If an object does initially exists, there are two outcomes:
```
rcaudy, line 205: IntelliJ is complaining that "exists" is plural.

server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
```diff
@@ -67,149 +67,264 @@ public ApplicationServiceGrpcImpl(final AppMode mode,
...
     }
 
     @Override
-    public synchronized void onNewField(final ApplicationState app, final Field<?> field) {
+    public synchronized void onRemoveField(ApplicationState app, Field<?> oldField) {
         if (!mode.hasVisibilityToAppExports()) {
             return;
         }
+        final AppFieldId id = AppFieldId.from(app, oldField.name());
+        // todo: or should we rely on the known type
```
rcaudy, line 135: I think that is "safest". Same for description. Might be annoying given the current code pattern, though.

server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
```diff
@@ -304,7 +378,7 @@ public void run() {
         // must be sync wrt parent
         private void markRunning() {
             if (!isScheduled) {
-                throw new IllegalStateException("Job is running without being scheduled");
+                // throw new IllegalStateException("Job is running without being scheduled");
```
rcaudy, line 381: Why did you need to comment this?

rcaudy (commented):
Seems like a lot of code for the state machines, etc. That said, I think it does a pretty good job of cleaning up the reachability and liveness issues, and it's overall simpler/more obviously correct.

I'm a little concerned that @nbauernfeind mentioned we might have liveness issues with Java applications. I assume Groovy or Python get executed as scripts using the session's `LivenessScope`. That is arguably out-of-scope for this PR, however.

server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
devinrsmith: This is actually a bug to not use the known type. Updated code is much cleaner.

server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
devinrsmith: It didn't work w/ calling propogateUpdates on thread. I could make the state management better for on-thread vs scheduler tracking, but I think it's a potentially unnecessary safety belt.

server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
```diff
@@ -381,18 +251,159 @@ private void notifyObserverAborted() {
...
+
+    private void update(AppFieldId id, String description, String type) {
+        computeIfAbsent(id).update(description, type);
+    }
+
+    private void remove(AppFieldId id) {
+        computeIfAbsent(id).remove();
+    }
+
+    private State computeIfAbsent(AppFieldId id) {
```
rcaudy, line 266: I would have called this something like `accumulate` or `accumulated`.

server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
```diff
@@ -381,18 +251,159 @@ private void notifyObserverAborted() {
...
+            // If we send a remove, we'll be basing it off of our existing info
+            this.description = null;
+            this.type = null;
+        }
 
-    private static class LivenessTracker extends LivenessArtifact {
-        private <T> void maybeManage(T object) {
-            if (object instanceof LivenessReferent && DynamicNode.notDynamicOrIsRefreshing(object)) {
-                manage((LivenessReferent) object);
+        public void append(Updater builder) {
```
rcaudy, line 335: Maybe better to rename `builder` to `updater`?

server/src/main/java/io/deephaven/server/appmode/ApplicationServiceGrpcImpl.java
```diff
@@ -381,18 +251,159 @@ private void notifyObserverAborted() {
...
+                    .setTypedTicket(typedTicket(id, type))
+                    .setFieldName(id.fieldName)
+                    .setFieldDescription(description == null ? "" : description)
+                    .setApplicationId(id.applicationId())
+                    .setApplicationName(id.applicationName())
+                    .build();
+        }
+    }
+
+    // An "updater" as opposed to a "builder" - modifies state and is builder.
```
rcaudy, line 378: Why not use JavaDoc format?

rcaudy (approved):
I made some strictly cosmetic comments. I like this change a lot. Merge it as-is, or make the small changes, either way it's good.

rcaudy (dismissed):

nbauernfeind (approved):

engine/table/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -329,9 +344,12 @@ private void doRefresh() {
...
 
-        table.notifyListeners(new TableUpdateImpl(RowSetFactory.flat(newSize),
-                RowSetFactory.flat(oldSize), RowSetFactory.empty(),
-                RowSetShiftData.EMPTY, ModifiedColumnSet.EMPTY));
+        final QueryTable localTable = tableRef.get();
+        if (localTable != null) {
+            localTable.notifyListeners(new TableUpdateImpl(RowSetFactory.flat(newSize),
+                    RowSetFactory.flat(oldSize), RowSetFactory.empty(),
+                    RowSetShiftData.EMPTY, ModifiedColumnSet.EMPTY));
+        }
```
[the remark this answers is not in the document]
rcaudy: I'm ambivalent. We could, since we think that without the child table we have no reason to continue existing, but it's not strictly necessary. If `tableRef` has been cleared, the cleanup handler thread should drain its `RetainedReferenceTracker` from the reference queue, which should drop our reference count to 0. Probably better to be aggressive, though, in case someone has us in a `LivenessScope` keeping us active when there's no reason for it. I will make two "suggestion" comments.

engine/table/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
```diff
@@ -267,15 +272,23 @@ public void close() {
...
             doRefresh();
         } catch (Exception e) {
             log.error().append("Error refreshing ").append(StreamToTableAdapter.class.getSimpleName()).append('-')
                     .append(name).append(": ").append(e).endl();
-            table.notifyListenersOnError(e, null);
             updateSourceRegistrar.removeSource(this);
+            final QueryTable localTable = tableRef.get();
+            if (localTable != null) {
+                localTable.notifyListenersOnError(e, null);
+            }
```
rcaudy, line 291:
```suggestion
            } else {
                close();
            }
```

engine/table/src/main/java/io/deephaven/stream/StreamToTableAdapter.java
rcaudy, line 352:
```suggestion
        } else {
            close();
        }
```

rcaudy (approved):

PR #2163, opened 2022-03-31
Fix JsFigureFactory ClassCastException

### LINKED ISSUE #2159: ChartBuilder throws an error
**Description**

**Steps to reproduce**
1. Create a ticking table:
```
from deephaven.TableTools import timeTable
tt = timeTable("00:00:01").update("y=Math.sin(i)")
```
2. From the table side bar, open the Chart Builder, and click "Create"

**Expected results**
2. Chart should be created

**Actual results**
2. An error is output
```
[ChartPanel] handleLoadError Error: java.lang.ClassCastException
    at ClassCastException_0.createError (dh-core.js:1190:10)
    at ClassCastException_0.initializeBackingError (dh-core.js:1210:46)
    at ClassCastException_0.Throwable_0 (dh-core.js:1155:8)
    at ClassCastException_0.Exception_1 (dh-core.js:1234:18)
    at ClassCastException_0.RuntimeException_1 (dh-core.js:1247:18)
    at new ClassCastException_0 (dh-core.js:1506:25)
    at Object.checkCriticalType (dh-core.js:4797:16)
    at castTo (dh-core.js:685:6)
    at Function.onInvoke_162 (dh-core.js:26698:46)
    at lambda (dh-core.js:150:22)
```

image

**Versions**
 - Deephaven: v0.11.4

mofojed:
- The asArray doesn't seem to set the castable types correctly
- Just set to a Promise<?> then force the cast later

Fixes #2159

This change looks bigger than it is because of whitespace/indenting. Only changed the Promise type from `Promise<JsTable>[]` to `Promise<?>[]` and then added a `Js.uncheckedCast` on line 47.

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/widget/plot/JsFigureFactory.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/widget/plot/JsFigureFactory.java
@@ -40,66 +40,68 @@ public class JsFigureFactory {
         FigureDescriptor figureDescriptor = convertJsFigureDescriptor(descriptor);
         FetchObjectResponse response = new FetchObjectResponse();
         response.setData(figureDescriptor.serializeBinary());
-        // noinspection unchecked
-        Promise<JsTable>[] tableCopyPromises =
+        Promise<?>[] tableCopyPromises =
                 tables.map((table, index, all) -> table.copy(false)).asArray(new Promise[0]);
         return Promise.all(tableCopyPromises)
-                .then(tableCopies -> new JsFigure(
-                        c -> c.apply(null, response),
-                        (figure, descriptor1) -> {
-                            // We need to listen for disconnects and reconnects
-                            boolean[] isTableDisconnected = new boolean[tableCopies.length];
-                            ArrayList<RemoverFn> removerFns = new ArrayList<>(tableCopies.length * 3);
+                .then(unknownTableCopies -> {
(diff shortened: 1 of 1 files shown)


devinrsmith (commented):
I can't say if this is a reasonable change or not. If we need it approved before Colin can review, let me know.

web/client-api/src/main/java/io/deephaven/web/client/api/widget/plot/JsFigureFactory.java
```diff
@@ -40,66 +40,69 @@
...
-                            for (int i = 0; i < tableCopies.length; i++) {
-                                final int tableIndex = i;
-                                // Tables are closed when the figure is closed, no need to remove listeners later
-                                removerFns.add(tableCopies[i].addEventListener(JsTable.EVENT_DISCONNECT, ignore -> {
-                                    isTableDisconnected[tableIndex] = true;
-                                    for (int j = 0; j < isTableDisconnected.length; j++) {
-                                        if (isTableDisconnected[j] && j != tableIndex) {
-                                            return;
+                .then(unknownTableCopies -> {
+                    JsTable[] tableCopies = Js.uncheckedCast(unknownTableCopies);
```
niloc132, line 47: If you change `tableCopies` to be of type `JsArray<JsTable>`, then you can do a plain Js.cast() instead of unchecked cast.

niloc132 (approved):

PR #2165, opened 2022-03-31
Rename python/deephaven-> python/deephaven_legacy

### LINKED ISSUE #2162: rename the deephaven python package to deephaven_legacy
This is the first step to make the hand-written 'deephavenV2' the default/official python package for DHC. The next step is to rename deephaven2 to the vacant 'deephaven' name.

jmao-denver:
Fixes #2162


devinrsmith (approved):
Not sure there is much I can add to review, except I've looked at all the non-renamed files and they look ok. Of course, I can't look exhaustively at things that should have been changed, but weren't. Best we can do is rely on CI and QA; and I think any downstream fallout from this change can be addressed as needed.

Not sure the timing around when we want to merge. Pre release, or post release? @rcaudy

rcaudy:
@devinrsmith we want to merge this (and the deephaven2->deephaven change) pre-release.
@jmao-denver please run nightlies before merging this.

niloc132:
@rcaudy I'm not sure that this makes sense, since all nightly builds are engine tests right? we don't have other steps that run anything except the serial/parallel/outofband stuff?

I see that extensions/csv and parquet/table both have some outofband stuff, but that's it?

rcaudy (approved):
Deferring to @devinrsmith 's more substantive review; let's merge this.

Issue #2062, opened 2022-03-02
Sphinx generated pydocs for deephaven2 are missing the Kafka modules


jmao-denver:
It appears that the doc generation stopped at the stream sub package level and didn't scan the modules inside it.

jmao-denver:
fixed by #2167

PR #2167, opened 2022-04-01
Rename deephaven2->deephaven

jmao-denver:
1. gradle scripts still use names like deephaven-wheel, deephaven2-wheel, needs help from either @niloc132 or @devinrsmith to rename them properly
2. #2062 will follow after this gets merged


Generators/src/main/java/io/deephaven/plot/util/GeneratePyV2FigureAPI.java
```diff
@@ -24,7 +24,7 @@
     private static final String INDENT = "    ";
     private static final String JCLASS = "io.deephaven.plot.Figure";
     private static final String PYPREAMBLE = "/Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py";
-    private static final String PYMODUlE = "/pyintegration/deephaven2/plot/figure.py";
+    private static final String PYMODUlE = "/pyintegration/deephaven/plot/figure.py";
```
niloc132, line 27: Possibly rename this class too? and also the various gradle tasks that are specific to deephaven vs deephaven2?

niloc132: Description covers this, i'll ignore for now, please make a follow up

niloc132 (approved):
Approved, I ignored all json files, and only read the contents of the .py and .java changes. I had a few small questions about the changes that weren't just import fixes, but Jianfeng has clarified, I'm good with it.

Gradle, docker, and codegen fixes will be a followup that Jianfeng will file (and either handle himself with help from me/devin, or assign to one of us).

Generators/src/main/java/io/deephaven/plot/util/GeneratePyV2FigureAPI.java
jmao-denver: This can be done after we officially kill deephavenv1, most of the generator stuff can be removed

devinrsmith (approved):

PR #2170, opened 2022-04-01
Fix JS API examples after deephaven2 -> deephaven rename

mofojed:
- Fix all python code snippets in the JS API examples
- Tested each page accessible from http://localhost:10000/jsapi/

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/public/table_basic.html
+++ b/web/client-api/src/main/java/io/deephaven/web/public/table_basic.html
@@ -35,7 +35,7 @@ var ide;
         // Run code that will create a static table with 10 rows and three columns, I, J, K
         await ide.runCode(`
 from deephaven import empty_table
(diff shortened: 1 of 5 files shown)


web/client-api/src/main/java/io/deephaven/web/public/table_basic.html
```diff
@@ -34,8 +34,8 @@ <h3>Table data</h3>
         ide = await connection.startSession("python");
         // Run code that will create a static table with 10 rows and three columns, I, J, K
         await ide.runCode(`
-from deephaven.TableTools import emptyTable
-remoteTable = emptyTable(10).updateView('I=i', 'J=I*I', 'K=i%2==0?"Hello":"World"')
```
niloc132, line 38: `formulas = ` is unnecessary, you just need to wrap in an array

devinrsmith (approved):

PR #2177, opened 2022-04-01
removed deephaven2 reference

jakemulf:


hythloda (approved):

Integrations/python/deephaven_legacy/tutorials/crypto_trades.py
```diff
@@ -40,7 +40,7 @@ def ticking_crypto_milliseconds(interval: int):
     return t
 
 def ticking_crypto_milliseconds_v2(interval: int):
-    from deephaven2 import time_table
+    from deephaven import time_table
```
jjbrosnan, line 43: If this is legacy, isn't the method name `timeTable`?

```suggestion
    from deephaven import timeTable
```

jjbrosnan (changes requested):
Legacy naming convention

Integrations/python/deephaven_legacy/tutorials/crypto_trades.py
jakemulf: I don't think this is legacy so it's supposed to be `time_table`

This code works when I manually make the update on the server

```
from deephaven_legacy.tutorials.crypto_trades import *

crypto_streaming = ticking_crypto_milliseconds_v2(25)
```

jakemulf: Without the manual update it fails with the following error message

```
Type: <class 'ModuleNotFoundError'>
Value: No module named 'deephaven2'
```

jjbrosnan (approved):

hythloda (approved):

PR #2178, opened 2022-04-01
Accept both single value/multi values in params

### LINKED ISSUE #2173: Unexpected Error message when using .update with just a string
**Description**

Using the new Python deephaven API, getting an unexpected error when using the .update method incorrectly.

**Steps to reproduce**
1. Run:
```
tt = time_table("00:00:01").update("y=Math.sin(i)")
```

**Expected results**
1. Table is created correctly, or a helpful error message is output

**Actual results**
1. Error message is returned, saying:
```
Value: table update operation failed. : RuntimeError: java.lang.IllegalArgumentException: Unable to parse formula '=', expected form '<newColumn>=<expression>'
Line: 246
```
This is somewhat confusing, because my formula _does_ seem to be that format. But I think it's because it's expecting an array, and it's treating the string as an array of chars.

jmao-denver:
To make transition to the new Python API easier.

Fixes #2173

### DIFF
--- a/pyintegration/tests/test_table.py
+++ b/pyintegration/tests/test_table.py
@@ -54,6 +54,8 @@ class TableTestCase(BaseTestCase):
         column_names = [f.name for f in self.test_table.columns]
         result_table = self.test_table.drop_columns(cols=column_names[:-1])
         self.assertEqual(1, len(result_table.columns))
+        result_table = self.test_table.drop_columns(cols=column_names[-1])
+        self.assertEqual(1, len(self.test_table.columns) - len(result_table.columns))
 
     def test_move_columns(self):
         column_names = [f.name for f in self.test_table.columns]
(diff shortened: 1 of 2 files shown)


mofojed (approved):
Looks like it fixes the case in the ticket, was able to run it successfully now.

rcaudy (approved):

pyintegration/tests/test_table.py
```diff
@@ -139,10 +170,8 @@ def test_where_in(self):
             self.assertLessEqual(unique_table.size, result_table.size)
 
         with self.subTest("where-not-in filter"):
-            result_table2 = self.test_table.where_not_in(unique_table, cols=["c"])
-            self.assertEqual(
-                result_table.size, self.test_table.size - result_table2.size
-            )
```
chipkent, line 145: Did you mean to delete this one?

pyintegration/tests/test_table.py
```diff
@@ -394,7 +463,7 @@ def test_agg_all_by(self):
                 self.assertGreaterEqual(result_table.size, 1)
 
         with self.assertRaises(DHError) as cm:
-            test_table.agg_all_by(count_("aggCount"), ["grp_id"])
+            test_table.agg_all_by(count_("aggCount"), "grp_id")
```
chipkent, line 466: Should both of these be tested?

chipkent (changes requested):

pyintegration/tests/test_table.py
jmao-denver: yes, the single value case covers the original test and this change

pyintegration/tests/test_table.py
jmao-denver: yes, the single value case covers both the original test and this change

chipkent (approved):

PR #2183, opened 2022-04-01
Support casting py created Java boxed number obj

jmao-denver:
Note, from some reason,  QueryLanguageFunctionUtils.java hasn't been generated for some time and has been manually modified by Cristian for the BigDeciaml/BigInteger support. His change is quite extensive and considering that the change to fix this problem is quite limited, we thought it makes senses for me to just change this file directly and he will file a ticket to properly address the 'generated' becoming 'manual'  issue.

Fixes bug-2044

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageFunctionUtils.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageFunctionUtils.java
@@ -15,6 +15,8 @@ import java.math.RoundingMode;
 
 import static java.lang.Math.*;
 
+import java.lang.Number;
+
 @SuppressWarnings({"unused", "WeakerAccess", "SimplifiableIfStatement"})
 public final class QueryLanguageFunctionUtils {
 
@@ -19276,19 +19278,19 @@ public final class QueryLanguageFunctionUtils {
     }
 
     public static int intCast(Object a) {
-        return a == null ? QueryConstants.NULL_INT : (int) a;
+        return a == null ? QueryConstants.NULL_INT : ((Number)a).intValue();
     }
 
     public static double doubleCast(Object a) {
-        return a == null ? QueryConstants.NULL_DOUBLE : (double) a;
+        return a == null ? QueryConstants.NULL_DOUBLE : ((Number)a).doubleValue();
     }
 
     public static long longCast(Object a) {
-        return a == null ? QueryConstants.NULL_LONG : (long) a;
+        return a == null ? QueryConstants.NULL_LONG : ((Number)a).longValue();
     }
 
     public static float floatCast(Object a) {
-        return a == null ? QueryConstants.NULL_FLOAT : (float) a;
+        return a == null ? QueryConstants.NULL_FLOAT : ((Number)a).floatValue();
     }
 
     public static char charCast(Object a) {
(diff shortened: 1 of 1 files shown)


jjbrosnan (dismissed):

jakemulf (dismissed):

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageFunctionUtils.java
```diff
@@ -15,6 +15,8 @@
 
 import static java.lang.Math.*;
 
+import java.lang.Number;
```
niloc132, line 18: This should not be necessary, all java.lang imports are implicit, unless there is a conflicting Number type?

pyintegration/integration-tests/numbers_from_list_and_func.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def get_x(i):
+    return x[i]
+
+
+t_list = empty_table(n).update(["X = x[i]"])
+t_func = empty_table(n).update(["X = get_x(i)"])
+assert t_list.columns[0].data_type == dtypes.PyObject
+assert t_func.columns[0].data_type == dtypes.JObject
```
niloc132, line 19: personally i dont care for these assertions except as a sort of "hey this changed", but we dont actually require this difference, right?

it isn't bad to have the assertions, but maybe add a note that this behavior isnt actually essential, it is just what happens today or something?

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageFunctionUtils.java
```diff
@@ -19276,31 +19278,31 @@ public static byte byteCast(short a) {
     }
 
     public static int intCast(Object a) {
-        return a == null ? QueryConstants.NULL_INT : (int) a;
+        return a == null ? QueryConstants.NULL_INT : ((Number) a).intValue();
```
niloc132, line 19281: please change the query lang func utils generator, not this class (then run it, and it will re-generate this class).

note that @jcferretti is also making changes to that file, perhaps coordinate with him to be sure you aren't changing overlapping code.

niloc132: As discussed, ignore, Cristian is handling this part too.

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageFunctionUtils.java
```diff
@@ -15,6 +15,7 @@
 
 import static java.lang.Math.*;
 
+
```
niloc132, line 18: remove new extra newline

pyintegration/integration-tests/pyobj_field_access_test.py
```diff
@@ -0,0 +1,30 @@
+#
+# Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
```
niloc132, line 2: 2022

pyintegration/integration-tests/numba_vectorized_filter_test.py
```diff
@@ -0,0 +1,36 @@
+#
+# Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
```
niloc132, line 2: 2022

niloc132 (approved):

PR #2186, opened 2022-04-02
Add return type annotations for columns()

### LINKED ISSUE #2185: In deephaven/table.py, the columns function is missing return type annotation.

jmao-denver:
Fixes #2185

### DIFF
--- a/pyintegration/deephaven/table.py
+++ b/pyintegration/deephaven/table.py
@@ -4,7 +4,7 @@
 """ This module implements the Table class and functions that work with Tables. """
 from __future__ import annotations
 
-from typing import Union, TypeVar, Sequence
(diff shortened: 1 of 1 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageFunctionUtils.java
```diff
@@ -15,6 +15,8 @@
 
 import static java.lang.Math.*;
 
+import java.lang.Number;
```
chipkent, line 18: These changes belong in another PR

chipkent (changes requested):

PR #2193, opened 2022-04-04
Rename process memory tables to server state and add UGP stats.

jcferretti:
I'm using the debezium/perf code to generate load to showcase the table below.

Latest version:

Deephaven - Google Chrome_167


hythloda (dismissed):

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
```diff
@@ -147,9 +147,65 @@
...
+            ++totalCycles;
+            final boolean cycleOnBudget = targetCycleDurationMillis * 1000 * 1000 >= cycleTimeNanos;
+            if (cycleOnBudget) {
+                ++totalCyclesOnBudget;
+            }
+            totalCyclesTimeNanos += cycleTimeNanos;
+            totalSafePointPauseTimeMillis += safePointPauseTimeMillis;
+            if (sampledCycleTimeNanos.length <= maxSampledCycles) {
+                if (nSampledCycles + 1 > sampledCycleTimeNanos.length) {
+                    final long[] tmp = new long[sampledCycleTimeNanos.length * 2];
```
cpwright, line 185: maybe just Arrays.copyOf() instead of doing it by hand?

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
```diff
@@ -147,9 +147,65 @@
...
 
+    public static class AccumulatedCycleStats {
+        /**
+         * Total cycles run.
+         */
+        public long totalCycles = 0L;
+        /**
+         * Cycles run no exceeding their time budget.
+         */
+        public long totalCyclesOnBudget = 0L;
```
cpwright, line 160: I sort of figure why bother computing that here, when you can just do that afterwards when you are computing other things.

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
```diff
@@ -1513,7 +1569,12 @@ private void refreshTablesAndFlushNotifications() {
         waitForNextCycle(startTime, sched);
     }
 
-    private void logCycle(final long cycleTimeNanos) {
+    private void computeStatsAndLogCycle(final long cycleTimeNanos) {
+        final long safePointPauseTimeMillis = jvmIntrospectionContext.deltaSafePointPausesTimeMillis();
+        accumulatedCycleStats.accumulate(
+                DEFAULT_TARGET_CYCLE_DURATION_MILLIS,
```
cpwright, line 1575: Why would you use the default instead of the actual value?  That makes it even less worth computing the on/off budget here instead of later.

cpwright (commented):
In general this is fine; but I think it might make sense to simply log all of the cycles that exceed our de-minimus threshold as an array; the PerformanceQueries post processing could then do whatever it wants, including the median, P90 values, average, stdev, whatever w/o the need for us to decide it ahead of time.

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
jcferretti: The idea was that the target cycle time may change (but then I changed it during debugging to the constant and forgot to revert that, also ref your comment below)

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
jcferretti: See comment above.

cpwright (dismissed):

PR #2194, opened 2022-04-05
typo fix for countDistinct

jakemulf:


jmao-denver (approved):
Thanks for catching the error and fixing it.

PR #2196, opened 2022-04-05
Simplify and relocate deephaven py server library

niloc132:
As the deephaven wheel isn't intended to be used from outside the jvm, this patch removes the ability for it to start in that way. Attempting to use the APIs in this way will now result in an error.

Instead, unit tests, integration tests, and document generation will now have their own code to turn on enough of the JVM to handle these cases.

In the future, tests should also function from inside the JVM, so we know they actually use the same environment as a user would be exposed to in production.


niloc132:
After discussion with @jmao-denver it looks like there is a simpler way to handle the unit tests.

pyintegration/setup.py
```diff
@@ -29,7 +29,7 @@ def normalize_version(version):
       version=__normalized_version__,
       description='Deephaven Engine Python Package',
       long_description=README,
-      packages=find_namespace_packages(exclude=("tests",)),
+      packages=find_namespace_packages(exclude=("tests", "integrations-tests", "test_helper")),
```
niloc132, line 32: verify that these are needed, probably old wheels included integration tests...

pyintegration/test_helper/__init__.py
```diff
@@ -1,37 +1,57 @@
...
+DEFAULT_CLASSPATH = os.environ.get('DEEPHAVEN_CLASSPATH', "/opt/deephaven/server/lib/*")
+
+
+def start_jvm():
+    """ This function uses the default DH property file to embed the Deephaven server and starts a Deephaven Python
+    Script session. """
+    if not jpy.has_jvm():
+
+        # we will try to initialize the jvm
+        init_jvm()
```
niloc132, line 27: finish inlining this

pyintegration/tests/__init__.py
```diff
@@ -1,3 +1,6 @@
 #
 #   Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
 #
+
+from test_helper import start_jvm
+start_jvm()
```
niloc132, line 6: newline

pyintegration/test_helper/__init__.py
```diff
@@ -1,37 +1,57 @@
...
-              # The following are the jpyutil.init_jvm options which are passed through after attaching our options
-              java_home=None,
-              jvm_dll=None,
-              jvm_maxmem=None,
-              jvm_classpath=None,
-              jvm_properties=None,
-              jvm_options=None,
-              config_file=None,
-              config=None):
+DEFAULT_DEVROOT = os.environ.get('DEEPHAVEN_DEVROOT', "/tmp/pyintegration")
```
niloc132, line 15: @jmao-denver via call requests that these change to current dir for non-docker testing

jmao-denver (approved):
LGTM

niloc132:
Force pushed, resolving conflicts with main. Retested both deephaven_legacy and deephaven packages in the web ui.

PR #2212, opened 2022-04-06
Add a missing inc ref to PyObject when unwrapping to python

devinrsmith:


devinrsmith:
Demonstrated via:

```python
import jpy
import sys

class CustomClass:
    def __init__(self):
        pass

    def __del__(self):
        print("Deleted")

custom_obj_1 = CustomClass()
print(sys.getrefcount(custom_obj_1) - 1)

py_obj_array = jpy.array('org.jpy.PyObject', [custom_obj_1])
print(sys.getrefcount(custom_obj_1) - 1)

extracted_obj_1 = py_obj_array[0]
print(sys.getrefcount(custom_obj_1) - 1)
```

Before this change, the results were:

```
1
2
2
```

After this change, the results are:

```
1
2
3
```

cpwright (approved):

jmao-denver (approved):

PR #2209, opened 2022-04-06
debexium v2 changes

hythloda:

Issue #2206, opened 2022-04-06
Add WindowCheck in the new deephaven API


jmao-denver:
This is a feature that was available in the legacy deephaven, used in the debezium demo script.

jmao-denver:
fixed by #2271

PR #2215, opened 2022-04-06
debezium  deephaven2

hythloda:
Reverts deephaven/deephaven-core#2209


hythloda:
waiting on https://github.com/deephaven/deephaven-core/issues/2206

PR #2220, opened 2022-04-07
Add a ring table

devinrsmith:


devinrsmith:
Unit tests forthcoming.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
```diff
@@ -0,0 +1,389 @@
...
+     *
+     * <p>
+     * Logically equivalent to the condition {@code firstKey <= key <= lastKey}.
+     *
+     * @param key the key
+     * @return {@code true} if {@code key} is in the index.
+     * @see #firstKey()
+     * @see #lastKey()
+     */
+    public final boolean containsKey(long key) {
```
rcaudy, line 94: See my other comment in `get`. You don't need this method.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
```diff
@@ -0,0 +1,389 @@
...
+ * @param <T> the item type
+ * @param <ARRAY> the array type
+ * @param <SELF> the self type
+ */
+abstract class AbstractRingChunkSource<T, ARRAY, SELF extends AbstractRingChunkSource<T, ARRAY, SELF>>
+        implements DefaultChunkSource<Values> {
+
+    protected final ARRAY ring;
+    protected final int capacity;
+    long nextRingIx;
```
rcaudy, line 36: Can be `private`.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
```diff
@@ -0,0 +1,389 @@
...
+        // Precondition: firstRingIx > lastRingIx
+        final int fillSize1 = capacity - firstRingIx;
+        final int fillSize2 = lastRingIx + 1;
+        destination.copyFromArray(ring, firstRingIx, destOffset, fillSize1);
+        destination.copyFromArray(ring, 0, destOffset + fillSize1, fillSize2);
+        return fillSize1 + fillSize2;
+    }
+
+    final int keyToRingIndex(long key) {
+        return (int) (key % capacity);
```
rcaudy, line 337: Are you sure you don't want to use a power-of-two capacity requirement to make this calc more efficient?

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    final void bringUpToDate(FillContext fillContext, SELF current) {
+        // We *could* try to be smart and get away with a single copy.
```
rcaudy, line 341: I prefer the current strategy (doing 1-2 copies, not re-copying cells that didn't roll off).

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
```diff
@@ -0,0 +1,389 @@
...
+        if (!containsRange(firstKey, lastKey)) {
+            throw new IllegalStateException(
+                    String.format("getChunk precondition broken, invalid range. requested=[%d, %d], available=[%d, %d]",
+                            firstKey, lastKey, firstKey(), lastKey()));
+        }
+        final int firstRingIx = keyToRingIndex(firstKey);
+        final int secondRingIx = keyToRingIndex(lastKey);
+        if (firstRingIx <= secondRingIx) {
+            // Optimization when we can return a contiguous view
+            return ring(firstRingIx, secondRingIx - firstRingIx + 1); // todo: is it bad that this is writable?
```
rcaudy, line 226: Regarding your to-do, it's OK. Callers are *not* allowed to modify the result of a `getChunk`.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
```diff
@@ -0,0 +1,389 @@
...
+        private int destOffset;
+
+        public KeyFiller(WritableChunk<? super Values> destination) {
+            this.destination = Objects.requireNonNull(destination);
+        }
+
+        @Override
+        public void accept(long key) {
+            // this check should not be necessary given precondition on fillChunk
+            if (!containsKey(key)) {
```
rcaudy, line 262: I would not spend the cycles on this check. Better to delete it, or put it under a debug flag. Same with `containsRange` in `RangeFiller`.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
```diff
@@ -0,0 +1,389 @@
...
+                rowSequence.forAllRowKeys(filler);
+            }
+        } else {
+            try (final RangeFiller filler = new RangeFiller(destination)) {
+                rowSequence.forAllRowKeyRanges(filler);
+            }
+        }
+    }
+
+    private final class KeyFiller implements LongConsumer, Closeable {
```
rcaudy, line 251: I previously suggested pushing `fillChunk` down. I would at least push fillers down.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/RingColumnSource.java
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    @Override
+    public Chunk<? extends Values> getPrevChunk(@NotNull GetContext context, long firstKey, long lastKey) {
+        return prev.getChunk(context, firstKey, lastKey);
+    }
+
+    @Override
+    public boolean isImmutable() {
+        return false;
```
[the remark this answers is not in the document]
devinrsmith: Ah, so if it's outside of the key index (and return NULL_X), that doesn't count against immutability?

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AddsToRingsListener.java
```diff
@@ -0,0 +1,153 @@
...
+    private WritableRowSet resultRowSet() {
+        return (WritableRowSet) getDependent().getRowSet();
+    }
+
+    private void init(Init init) {
+        if (init == Init.NONE) {
+            return;
+        }
+        final boolean usePrev = init == Init.FROM_PREVIOUS;
+        try (final RowSet srcKeys = usePrev ? getParent().getRowSet().copyPrev() : getParent().getRowSet().copy()) {
```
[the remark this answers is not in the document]
devinrsmith: neat trick.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/RingColumnSource.java
```diff
@@ -0,0 +1,277 @@
...
+        final long k2 = prev.lastKey();
+        final long k3 = ring.firstKey();
+        final long k4 = ring.lastKey();
+        final RowSet removed;
+        final RowSet added;
+        if (k2 < k3) {
+            // No intersection. Prev may be empty.
+            // Removed: empty or [k1, k2]
+            // Added: [k3, k4]
+            removed = k2 == -1 ? RowSetFactory.empty() : RowSetFactory.fromRange(k1, k2);
```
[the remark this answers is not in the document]
devinrsmith: So, I've got *slightly* different semantics for firstKey() / lastKey() than RowSet does. But I'll move this value into a static constant I can reference.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/ObjectRingChunkSource.java
```diff
@@ -0,0 +1,44 @@
...
+        return ChunkType.Object;
+    }
+
+    @Override
+    void clear() {
+        // todo: when do we call this?
+        Arrays.fill(ring, null);
+    }
+
+    @Override
```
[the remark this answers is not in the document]
devinrsmith: So, the abstract base class contains the full logic for `fillChunk`, and I don't think it's much more (if at all) inefficient than per-type ones would be (at least once JITd). And it's a lot less code. Also, `getChunk`. There is likely JIT advantages as well since all of the math-code is exactly the same; and if you break it out into per-type code, your code size and paths are larger.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/ObjectRingChunkSource.java
```diff
@@ -0,0 +1,44 @@
...
+    void fillKey(@NotNull WritableChunk<? super Values> destination, int destOffset, int ringIx) {
+        destination.asWritableObjectChunk().set(destOffset, ring[ringIx]);
+    }
+
+    @Override
+    T get(long key) {
+        if (!containsKey(key)) {
```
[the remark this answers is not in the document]
devinrsmith: ArraySourceHelper impls do the check:

```java
        if (index < 0 || index > maxIndex) {
```
Is there any javadoc that expresses the requirements for get of single key?

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
devinrsmith: Nope, RangeFiller needs access.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/ObjectRingChunkSource.java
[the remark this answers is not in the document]
devinrsmith: Ah, I do see the note on ColumnSource now.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
devinrsmith: I think some safety checks *may* be good. See STRICT_KEYS - can turn off if desired.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
devinrsmith: I'm able to move this check up to the fillChunk method.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
devinrsmith: Done

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/ObjectRingChunkSource.java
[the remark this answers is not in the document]
devinrsmith: I've pushed down the fillers.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/RingColumnSource.java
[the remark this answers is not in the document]
rcaudy: I'm wrong. Since the values that are removed from the row set become unavailable in current space, this isn't immutable.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
rcaudy: I don't trust JIT to optimize something like this if there are any instances with non-power-of-2 capacity.
I'm not sure about your tail comment; the tail costs a bit of extra row set math once per cycle, vs. the mod that is a tax on every cell access.
That said, this is not necessarily the most important thing to optimize.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AddsToRingsListener.java
```diff
@@ -0,0 +1,153 @@
...
+            rings[ix] = ring;
+            resultMap.put(name, output);
+            ++ix;
+        }
+        final QueryTable result = new QueryTable(RowSetFactory.empty().toTracking(), resultMap);
+        result.setRefreshing(true);
+        result.addParentReference(swapListener);
+        final AddsToRingsListener listener =
+                new AddsToRingsListener("AddsToRingsListener", parent, result, sources, rings);
+        listener.init(init);
```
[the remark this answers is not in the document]
devinrsmith: Extracted init.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AddsToRingsListener.java
```diff
@@ -0,0 +1,153 @@
...
+        if (!(dependent.getRowSet() instanceof WritableRowSet)) {
+            throw new IllegalArgumentException("Expected writable row set");
+        }
+        final int capacity = rings[0].capacity();
+        for (RingColumnSource<?> ring : rings) {
+            if (ring.capacity() != capacity) {
+                throw new IllegalArgumentException();
+            }
+        }
+        fillContexts = Arrays.stream(sources).map(s -> s.makeFillContext(capacity)).toArray(FillContext[]::new);
```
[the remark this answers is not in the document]
devinrsmith: Using fixed chunk size now, created on demand.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
```diff
@@ -0,0 +1,440 @@
...
+            physicalRows = srcKeys.subSetByPositionRange(skipRows, logicalFillSize);
+            physicalStartRingIx = nextRingIx + skipRows;
+        }
+        final int fillStartIx = keyToRingIndex(physicalStartRingIx);
+        try (
+                final RowSet _physicalRowsToClose = hasSkippedRows ? physicalRows : null;
+                final ResettableWritableChunk<Any> chunk = getChunkType().makeResettableWritableChunk();
+                final FillContext fillContext = source.makeFillContext(appendChunkSize);
+                final Iterator it = physicalRows.getRowSequenceIterator()) {
+            // Note: if we could do it.skip(skipRows), that would probably be more efficient than subSetByPositionRange
```
rcaudy, line 221: We discussed. Better to not dot he `subsetByPositionRange()`. Instead do `it.advance(srcKeys.get(skipRows))`. If you want to be fancy, you could have an assertion around `advanceAndGetPositionDistance()`.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
```diff
@@ -31,20 +32,45 @@
 abstract class AbstractRingChunkSource<T, ARRAY, SELF extends AbstractRingChunkSource<T, ARRAY, SELF>>
         implements DefaultChunkSource<Values> {
 
+    /**
+     * The return value of {@link #firstKey()} when {@link #isEmpty()}.
+     */
+    public static final long FIRST_KEY_EMPTY = 0;
```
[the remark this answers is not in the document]
devinrsmith: Yes, as discussed.

engine/table/src/test/java/io/deephaven/engine/table/impl/TstUtils.java
```diff
@@ -2163,6 +2167,43 @@ public static void assertTableEquals(@NotNull final Table expected, @NotNull fin
...
+            RowSet rowsExpected, ElementSource<T> srcExpected) {
+        assertThat(rowsActual.size()).isEqualTo(rowsExpected.size());
+        try (
+                final RowSet.Iterator itActual = rowsActual.iterator();
+                final RowSet.Iterator itExpected = rowsExpected.iterator()) {
+            while (itActual.hasNext()) {
+                if (!itExpected.hasNext()) {
+                    throw new IllegalStateException();
+                }
+                assertThat(srcActual.get(itActual.nextLong())).isEqualTo(srcExpected.get(itExpected.nextLong()));
```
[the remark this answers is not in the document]
devinrsmith: I think it's "OK"

Integrations/python/deephaven_legacy/TableTools/__init__.py
```diff
@@ -373,6 +373,11 @@ def base64Fingerprint(source):
     return _java_type_.base64Fingerprint(source)
 
 
+@_passThrough
+def booleanCol(name, *data):
+    return _java_type_.booleanCol(name, *data)
```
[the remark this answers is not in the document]
devinrsmith: So, this code is actually the correct implementation - the custom ones are for primitives. The missing pydoc is because I didn't follow-up with some .json changes (check fails if the python method booleanCol is missing, but not if the pydoc is "wrong"). Since we are axing legacy shortly, I figure it's better to *not* change hundreds of files (other small doc changes have been accumulating, but nobody has re-run the python generator); and leave as-is?

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
devinrsmith: Cool.

Integrations/python/deephaven_legacy/TableTools/__init__.py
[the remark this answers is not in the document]
rcaudy: Or you could re-run the generator? This is my (paranoid) pre-commit script:
```
./gradlew :replication-static:replicateAllSafe
./gradlew :replication-reflective:replicateAll
./gradlew spotlessApply
rm -r Integrations/python/deephaven/doc/*
./gradlew :Generators:generateAllPython
./gradlew :Generators:groovyStaticImportGenerator
./gradlew spotlessApply
```

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ring/AbstractRingChunkSource.java
```diff
@@ -201,37 +201,37 @@ public final void append(ChunkSource<? extends Values> source, RowSet srcKeys, i
...
+            if (skipRows > 0) {
+                skipRows(it, srcKeys.get(skipRows), skipRows);
+            }
             fillRingFromMiddle(source, chunk, fillContext, it, fillStartIx, appendChunkSize);
             fillRingFromStart(source, chunk, fillContext, it, fillStartIx, appendChunkSize);
         }
         nextRingIx += logicalFillSize;
     }
 
+    private void skipRows(Iterator it, long skipKey, long expectedDistance) {
```
rcaudy, line 219: `static`?

rcaudy (dismissed):

rcaudy (approved):

PR #2228, opened 2022-04-07
Create streaming variant of TimeTable; use synthetic column source for both

nbauernfeind:
Create a stream table version of time table. Make time table's array source synthetic instead of using precious memory.


engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
```diff
@@ -33,37 +47,47 @@
...
+        this(timeProvider, null, period, isStreaming);
     }
 
     public TimeTable(TimeProvider timeProvider, DateTime firstTime, long period) {
-        super(RowSetFactory.fromKeys().toTracking(), initColumn());
+        this(timeProvider, firstTime, period, false);
+    }
+
+    public TimeTable(TimeProvider timeProvider, DateTime firstTime, long period, boolean isStreaming) {
+        super(RowSetFactory.fromKeys().toTracking(), initColumn(period));
```
rcaudy, line 68: I know you didn't change this, but `empty()` would be more idiomatic than `fromKeys()`.

rcaudy, line 68: We're using `period` before we validate it. Likely harmless, though.

engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
```diff
@@ -33,37 +47,47 @@
...
         }
-        this.entry = UpdatePerformanceTracker.getInstance().getEntry("TimeTable(" + firstTime + "," + period + ")");
-        this.lastTime = firstTime == null ? null : new DateTime(firstTime.getNanos() - period);
-        binOffset = firstTime == null ? 0 : lastTime.getNanos() % period;
-        dateTimeArraySource = (DateTimeArraySource) getColumnSourceMap().get(TIMESTAMP);
+        this.isStreaming = isStreaming;
+        final String name = isStreaming ? "TimeTableStream" : "TimeTable";
+        this.entry = UpdatePerformanceTracker.getInstance().getEntry(name + "(" + firstTime + "," + period + ")");
+        columnSource = (SyntheticDateTimeSource) getColumnSourceMap().get(TIMESTAMP);
+        columnSource.baseTime = firstTime;
```
rcaudy, line 76: Couldn't we pass this into the column source constructor along with `period`? Better, we could pass `timeProvider` in, too, and always initialize `baseTime` in the constructor, allowing it to be `final`.

engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
```diff
@@ -75,33 +99,29 @@ private void refresh(final boolean notifyListeners) {
...
+
             if (rangeStart <= lastIndex) {
                 // If we have a period longer than 10s, print out that the timetable has been updated. This can be
                 // useful when analyzing what's gone wrong in the logs. It is capped at periods of 5s, so we don't
                 // end up with too much log spam for short interval time tables. 5s is not so coincidentally the period
                 // of the Jvm Heap: messages.
-                if (period >= 5_000_000_000L) {
-                    log.info().append("TimeTable updated to ").append(lastTime.toString()).endl();
+                if (columnSource.period >= 5_000_000_000L) {
+                    log.info().append("TimeTable updated to ").append(columnSource.get(lastIndex).toString()).endl();
```
rcaudy, line 118: I wonder if the comments and this log should just come out.

engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
```diff
@@ -33,37 +47,47 @@
...
         this.timeProvider = timeProvider;
-        this.period = period;
-        setAttribute(Table.ADD_ONLY_TABLE_ATTRIBUTE, Boolean.TRUE);
-        setFlat();
+        if (isStreaming) {
+            setAttribute(Table.STREAM_TABLE_ATTRIBUTE, Boolean.TRUE);
+        } else {
+            setAttribute(Table.ADD_ONLY_TABLE_ATTRIBUTE, Boolean.TRUE);
+            setFlat();
+        }
```
rcaudy, line 83: I really hate that we don't add ourselves to the UPG in the constructor, but rather rely on external code to do so. Especially given that we remove ourselves on destroy. Is there a reason for this?

engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
```diff
@@ -75,33 +99,29 @@ private void refresh(final boolean notifyListeners) {
...
                 // of the Jvm Heap: messages.
-                if (period >= 5_000_000_000L) {
-                    log.info().append("TimeTable updated to ").append(lastTime.toString()).endl();
+                if (columnSource.period >= 5_000_000_000L) {
+                    log.info().append("TimeTable updated to ").append(columnSource.get(lastIndex).toString()).endl();
                 }
                 final RowSet range = RowSetFactory.fromRange(rangeStart, lastIndex);
-                getRowSet().writableCast().insert(range);
+                final RowSet removedRange = isStreaming ? getRowSet().copy() : RowSetFactory.empty();
+                getRowSet().writableCast().update(range, removedRange);
```
rcaudy, line 122: I would consider just doing a `removeRange()` (if stream), and always doing an `insertRange()`. The outcome should be the same, but this way you tell the `RowSet` what you already know, rather than have it determine everything itself with `instanceof SingleRange` checks and similar.

engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
```diff
@@ -75,33 +99,29 @@ private void refresh(final boolean notifyListeners) {
...
                 // end up with too much log spam for short interval time tables. 5s is not so coincidentally the period
                 // of the Jvm Heap: messages.
-                if (period >= 5_000_000_000L) {
-                    log.info().append("TimeTable updated to ").append(lastTime.toString()).endl();
+                if (columnSource.period >= 5_000_000_000L) {
+                    log.info().append("TimeTable updated to ").append(columnSource.get(lastIndex).toString()).endl();
                 }
                 final RowSet range = RowSetFactory.fromRange(rangeStart, lastIndex);
-                getRowSet().writableCast().insert(range);
+                final RowSet removedRange = isStreaming ? getRowSet().copy() : RowSetFactory.empty();
```
rcaudy, line 121: I think it's better not to `copy()`. Just do a `fromRange()`. I'd rather have a new `WritableRowSetImpl` wrapping a `SingleRange` that is disposed of with the `TableUpdate` than force the primary `RowSet` to have to copy-on-write its internal `OrderedLongSet` (which should also be a `SingleRange`). A little less book-keeping, slightly better object lifecycles for GC.

engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
```diff
@@ -114,4 +134,309 @@ protected void destroy() {
         super.destroy();
         UpdateGraphProcessor.DEFAULT.removeSource(this);
     }
+
+    private static class SyntheticDateTimeSource implements
+            ColumnSource<DateTime>,
+            DefaultChunkSource.WithPrev<Values>,
+            MutableColumnSourceGetDefaults.LongBacked<DateTime>,
```
rcaudy, line 141: This `ColumnSource` is immutable, since you are not ever re-using keys in the streaming case.

That means you should use `io.deephaven.engine.table.impl.ImmutableColumnSourceGetDefaults.LongBacked`, delete your `getPrevX` overrides, and your prev fills and gets should delegate to curr (which they already do).

engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
```diff
@@ -114,4 +134,309 @@ protected void destroy() {
...
+            FillUnordered,
+            InMemoryColumnSource {
```
rcaudy, line 143: This is not an appropriate marker interface to use. This is more like a view column source than an in-memory column source.

engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
```diff
@@ -114,4 +134,309 @@ protected void destroy() {
...
+            return DateTime.class;
+        }
+
+        @Override
+        public Class<?> getComponentType() {
+            return null;
+        }
+
+        @Override
+        public WritableRowSet match(boolean invertMatch, boolean usePrev, boolean caseInsensitive, RowSet selection,
```
rcaudy, line 197: This might be the only override of `match` in the entire code base. Everything else extends `AbstractColumnSource`. I'll need to read this more carefully tomorrow.

rcaudy (commented):
This is an incomplete first pass.

engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
```diff
@@ -114,4 +134,309 @@ protected void destroy() {
...
+        private DateTime baseTime;
+        private final long period;
+
+        private SyntheticDateTimeSource(long period) {
+            this.period = period;
+        }
+
+        @Override
+        public DateTime get(long index) {
+            return DateTimeUtils.plus(baseTime, period * index);
```
rcaudy, line 154: `rowKey` is a better name than `index` these days.
We use this calculation all over the place. Maybe we want a private method for it, or to have everywhere else just call `get`? I'll point out that the default `fillChunk` implementation is just looping on `get`, and that that is what your implementation is doing.

engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
rcaudy: I'm convinced that what you have is correct.

engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
```diff
@@ -114,4 +134,309 @@ protected void destroy() {
...
+
+        @Override
+        public boolean providesFillUnordered() {
+            return true;
+        }
+
+        private class SyntheticDateTimeAsLongSource implements
+                ColumnSource<Long>,
+                DefaultChunkSource.WithPrev<Values>,
+                MutableColumnSourceGetDefaults.LongBacked<Long>,
```
rcaudy, line 296: Same as above: immutable, and not in-memory.

engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
nbauernfeind: As discussed on slack, initializing the base time would change the semantics slightly; this is currently deferred to the first refresh.

engine/table/src/main/java/io/deephaven/engine/table/impl/ImmutableColumnSource.java
```diff
@@ -57,4 +61,9 @@ default short getPrevShort(final long elementIndex) {
     default boolean isImmutable() {
         return true;
     }
+
+    @Override
+    default void fillPrevChunk(@NotNull FillContext context, @NotNull WritableChunk<? super Values> destination, @NotNull RowSequence rowSequence) {
```
rcaudy: Great idea. Let's do it for `getPrevChunk`, too. Be sure to check that nothing is overriding any of these methods without a good reason.

I wonder if this actually does anything, since `AbstractColumnSource` is the parent for most implementations.

engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
```diff
@@ -45,49 +46,91 @@
...
+        private UpdateSourceRegistrar registrar = UpdateGraphProcessor.DEFAULT;
+        private TimeProvider timeProvider;
+        private DateTime startTime;
+        private long period;
+        private boolean isStreamTable;
 
-    public TimeTable(TimeProvider timeProvider, long period) {
-        this(timeProvider, null, period, false);
-    }
+        public Builder withRegistrar(UpdateSourceRegistrar registrar) {
```
rcaudy, line 56: `with` and `as` imply to me that you are doing a copy-on-write of an immutable builder.

engine/table/src/main/java/io/deephaven/engine/table/impl/TimeTable.java
```diff
@@ -100,28 +143,25 @@ private void refresh(final boolean notifyListeners) {
...
 
             if (rangeStart <= lastIndex) {
-                // If we have a period longer than 10s, print out that the timetable has been updated. This can be
-                // useful when analyzing what's gone wrong in the logs. It is capped at periods of 5s, so we don't
-                // end up with too much log spam for short interval time tables. 5s is not so coincidentally the period
-                // of the Jvm Heap: messages.
-                if (columnSource.period >= 5_000_000_000L) {
-                    log.info().append("TimeTable updated to ").append(columnSource.get(lastIndex).toString()).endl();
+                final RowSet addedRange = RowSetFactory.fromRange(rangeStart, lastIndex);
+                final RowSet removedRange = isStreamTable && rangeStart > 0
```
rcaudy, line 157: As you noticed, we need to remove all prior rows if `isStreamTable` unconditionally.

rcaudy (dismissed):

rcaudy (approved):

PR #2229, opened 2022-04-07
Csv parser v3 wrapper

hythloda:


pyintegration/deephaven/csv.py
```diff
@@ -21,6 +21,11 @@ def read(
     path: str,
     header: Dict[str, dht.DType] = None,
     headless: bool = False,
+    skip_rows: int = 0,
+    num_rows: int = sys.maxint,
```
devinrsmith, line 25: Is sys.maxint a real thing?

pyintegration/deephaven/csv.py
```diff
@@ -31,6 +36,10 @@ def read(
     Args:
         path (str): a file path or a URL string
         header (Dict[str, DType]): a dict to define the table columns with key being the name, value being the data type
+        skip_rows (long):Number of data rows to skip before processing data. This is useful when you want to parse data in chunks. Defaults to 0
+        num_rows (long):  Max number of rows to process. This is useful when you want to parse data in chunks. Defaults to {@link Long#MAX_VALUE} 
+        allow_missing_columns (bool): Whether the library should allow missing columns in the input. If this flag is set, then rows that are too short (that have fewer columns than the header row) will be interpreted as if the missing columns contained the empty string. Defaults to false.
+        ignore_excess_columns (bool): Whether the library should allow excess columns in the input. If this flag is set, then rows that are too long (that have more columns than the header row) will have those excess columns dropped. Defaults to false.
         headless (bool): indicates if the CSV data is headless, default is False
```
jmao-denver, line 43: one space after ':'

pyintegration/deephaven/csv.py
jmao-denver: should use deephaven.constants.MAX_LONG?

jmao-denver (changes requested):

pyintegration/deephaven/csv.py
```diff
@@ -15,12 +15,17 @@
 _JCsvTools = jpy.get_type("io.deephaven.csv.CsvTools")
 _JParsers = jpy.get_type("io.deephaven.csv.parsers.Parsers")
 _JArrays = jpy.get_type("java.util.Arrays")
-
+_JQueryConstants = jpy.get_type("io.deephaven.util.QueryConstants")
```
jmao-denver, line 18: duplicate work with the constants module, should use 'from deephaven.constants import MAX_LONG'

pyintegration/deephaven/csv.py
```diff
@@ -15,12 +15,17 @@
...
 
 def read(
     path: str,
     header: Dict[str, dht.DType] = None,
     headless: bool = False,
+    skip_rows: int = 0,
+    num_rows: int = _JQueryConstants.MAX_LONG,
```
jmao-denver, line 25: should this be max_rows?

pyintegration/deephaven/csv.py
```diff
@@ -66,6 +75,11 @@ def read(
 
         csv_specs = (
             csv_specs_builder.hasHeaderRow(not headless)
+            .skipRows(skip_rows)
+            .numRows(num_rows)
+            .ignoreEmptyLines(ignore_empty_lines)
+            .allowMissingColumns(allow_missing_columns)
+            .ignoreExcessColumns(ignore_excess_columns)
             .delimiter(ord(delimiter))
             .quote(ord(quote))
```
jmao-denver, line 84: I am not sure if I like the design, the number of params has increased to 12, what if we need to add more options in the future, the signature of this method might become unwieldy pretty fast. It almost feels like that we should wrap CsvSpec and support building it through a builder pattern to make the code cleaner and more extensible. I think we also need @chipkent's input on this.

jmao-denver (changes requested):

pyintegration/deephaven/csv.py
devinrsmith: As long as all the named arguments have the correct defaults, I think it's "ok" to have a lot of them? IMO the builder pattern is useful in java because we don't have named argument support.

pyintegration/deephaven/csv.py
hythloda: I think max_long so it is consistent with the groovy default.

jmao-denver: i meant 'num_rows'

pyintegration/deephaven/csv.py
jmao-denver: OK, I am convinced because the Pandas version has even more params

jmao-denver (approved):

PR #2230, opened 2022-04-07
Wrap RegexFilter in deephaven V2

### LINKED ISSUE #2225: Regex filter should be wrapped in the new deephaven API
regex filter is a nice tool and available in deephaven legacy.

jmao-denver:
Also laid foundation for supporting the new Filter interface in full.

Fixes #2225

### DIFF
--- a/pyintegration/deephaven/_wrapper.py
+++ b/pyintegration/deephaven/_wrapper.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 import inspect
 from abc import ABC, abstractmethod
-from typing import Set, Union, Optional
+from typing import Set, Union, Optional, Any
 
 import jpy
 
@@ -70,3 +70,13 @@ def wrap_j_object(j_obj: jpy.JType) -> Union[JObjectWrapper, jpy.JType]:
     wc = _lookup_wrapped_class(j_obj)
 
     return wc(j_obj) if wc else j_obj
+
+
+def unwrap(obj: Any) -> Union[jpy.JType, Any]:
+    """ Returns the wrapped raw Java object if this is a wrapped Java object. Otherwise, returns the same object. """
+    if isinstance(obj, JObjectWrapper):
+        return obj.j_object
+
+    return obj
+
+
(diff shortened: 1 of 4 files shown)


pyintegration/deephaven/_wrapper.py
```diff
@@ -70,3 +70,13 @@ def wrap_j_object(j_obj: jpy.JType) -> Union[JObjectWrapper, jpy.JType]:
     wc = _lookup_wrapped_class(j_obj)
 
     return wc(j_obj) if wc else j_obj
+
+
+def unwrap(obj: Any) -> Union[jpy.JType, Any]:
+    """ Returns the wrapped raw Java object if this is a wrapped Java object. Otherwise, returns the same object. """
```
devinrsmith, line 76: Yes, I like this! Were there other places we should use this method instead (I think there might have been a few places that were explicitly checking to unwrap Tables, and maybe some other wrapped types?).

jjbrosnan (approved):

pyintegration/deephaven/filters.py
```diff
@@ -0,0 +1,43 @@
+#
+#   Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+#
+
+#
+#   Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+#
+""" This module implement various filters that can be used in deephaven table's filter operations."""
```
devinrsmith, line 8: I really think the design of this library is going to start paying dividends. I like this.

jakemulf (approved):

hythloda (dismissed):

pyintegration/deephaven/_wrapper.py
jmao-denver: Except in deephaven/server/script_session/__init__.py with includes the support of unwrapping legacy's FigureWrapper for compatibility till we retire deephaven_legacy, this is the first case that we would accept a mix of parameter types including the wrapped one which calls for the generic  unwrapping.

devinrsmith (approved):

PR #2241, opened 2022-04-10
Add AsOfMatchRule and add match rule param in aj

### LINKED ISSUE #2240: python aj missing match rule option
With an `aj`, the user should be able to specify LT (less than) or LTE (less than or equal) (by default it's LTE).

Similarly with `raj`.

jmao-denver:
Fixes #2240

### DIFF
--- a/pyintegration/tests/test_table.py
+++ b/pyintegration/tests/test_table.py
@@ -3,24 +3,9 @@
 #
 import unittest
 
-from deephaven import DHError, read_csv, empty_table, SortDirection
-from deephaven.agg import (
-    sum_,
-    weighted_avg,
-    avg,
-    pct,
-    group,
-    count_,
-    first,
-    last,
-    max_,
-    median,
-    min_,
-    std,
-    abs_sum,
-    var,
-    formula,
-)
+from deephaven import DHError, read_csv, empty_table, SortDirection, AsOfMatchRule
+from deephaven.agg import sum_, weighted_avg, avg, pct, group, count_, first, last, max_, median, min_, std, abs_sum, \
+    var, formula
 from deephaven.table import Table
 from tests.testbase import BaseTestCase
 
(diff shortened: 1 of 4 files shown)


pyintegration/tests/test_table.py
```diff
@@ -289,15 +274,15 @@ def test_as_of_join(self):
             result_table = left_table.aj(right_table, on=["a"])
             self.assertGreater(result_table.size, 0)
             self.assertLessEqual(result_table.size, left_table.size)
-            result_table = left_table.aj(right_table, on="a")
+            result_table = left_table.aj(right_table, on="a", joins="e", match_rule=AsOfMatchRule.LESS_THAN)
```
devinrsmith, line 277: This test should still pass without setting this param, right? The default is LTE.

pyintegration/tests/test_table.py
```diff
@@ -289,15 +274,15 @@ def test_as_of_join(self):
...
             self.assertGreater(result_table.size, 0)
             self.assertLessEqual(result_table.size, left_table.size)
 
         with self.subTest("reverse-as-of join"):
             result_table = left_table.raj(right_table, on=["a"])
             self.assertGreater(result_table.size, 0)
             self.assertLessEqual(result_table.size, left_table.size)
-            result_table = left_table.raj(right_table, on="a")
+            result_table = left_table.raj(right_table, on="a", joins="e", match_rule=AsOfMatchRule.GREATER_THAN)
```
devinrsmith, line 285: ditto

pyintegration/tests/test_table.py
jmao-denver: The modified case tests both if a single string value for 'on' works and if aj can accept the non-default match rule.

pyintegration/tests/test_table.py
jmao-denver: see above.

devinrsmith (approved):

PR #2242, opened 2022-04-10
Handle the sequence type params correctly

### LINKED ISSUE #2239: plot_xy with 'by' argument gets 'no matching Java method overloads found' error.
To reproduce:
```

test_table = read_csv("/data/test_table.csv")
figure = Figure()
# new_f = figure.plot_xy("plot1", test_table, x="a", y="b")
new_f = figure.plot_xy("plot1", test_table, x="a", y="b", by=["d"])
```

jmao-denver:
Fixes #2239

### DIFF
--- a/Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
+++ b/Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
@@ -78,6 +78,10 @@ def _convert_j(name: str, obj: Any, types: List) -> Any:
         else:
             return obj.value
     elif isinstance(obj, Sequence):
+        if not obj or not isinstance(obj[0], (bool, int, float)):
+            return obj
+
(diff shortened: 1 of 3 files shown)


Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
```diff
@@ -78,6 +78,10 @@ def _convert_j(name: str, obj: Any, types: List) -> Any:
         else:
             return obj.value
     elif isinstance(obj, Sequence):
+        if not obj or not isinstance(obj[0], (bool, int, float)):
+            return obj
```
devinrsmith, line 82: I hate checking just the first item of a sequence and calling it "good". I also don't really understand what's going on here... in some cases we are return `obj` or unwrapping it... but why are we doing something with numpy?

devinrsmith (approved):

Issue #2221, opened 2022-04-07
EPIC: Allow deephaven server to be started from python


niloc132:
- [ ] Make the Jetty-based server implementation easier to start with custom options, bindings
 - [x] Move bootstrap.py to test code, to be clear that this shouldn't be used to start the jvm under normal circumstances. It also probably should be pared down to be simpler to read and understand, as we practically never need that degree of customizability.
 - [x] Move existing top-level python projects into `py/`
 - [x] Create a new java application project with basic wiring to launch jetty for static content and gRPC services, designed to be easily called from python. This will involve cleaning up some currently rough edges in the dagger bindings, so that it is easier to build custom applications like this
 - [x] Create a new wheel project called `deephaven_server` which is able to start the jvm and the java components of Deephaven, including the engine and the gRPC service. This will have very little python in it, but will mostly be a container to hold the above java application, and will mostly offer the ability to start different parts of the server, as well as manipulate the script session or application
 - [ ] Ensure that `deephaven` has tools for working with the UGP
 - [ ] Update all wheel projects to publish working binaries to pypi, including jpy-deephaven (or jpy), deephaven, and deephaven_server.
 - [ ] Provide an optional custom UI experience where no console appears in the UI, but newly bound objects appear when the UI hears about them

niloc132:
Incorrectly closed, this issue isn't complete yet.

PR #2249, opened 2022-04-11
Relocate python sources into py/

niloc132:
This also declares wheels as artifacts, making it simpler to declare and install dependencies.

Gradle build/ and python dist/ are now excluded from the python build - if we find this is onerous, buildPyWheel could take a CopySpec argument. With that said, this shouldn't be necessary soon, since deephaven-jpy will hopefully soon be replaced by upstream jpy, and deephaven-legacy will be phased out, leaving only projects that use the new `python-wheel` plugin.

Legacy deephaven-wheel is renamed now in gradle as well, but not relocated.

Partial #2221


docker/server/build.gradle
```diff
@@ -26,31 +26,24 @@ baseMap.keySet().each {base ->
...
-
 configurations {
     serverApplicationDist
+    pythonWheel
 }
 
 dependencies {
     serverApplicationDist project(path: ':server-netty-app', configuration: 'applicationDist')
+
+    pythonWheel project(path: ':deephaven-jpy', targetConfiguration: 'pythonWheel')
```
niloc132, line 37: Technically this isn't required any more, it is transitively picked up by py-server and deephaven-wheel both - but since we expect people will actually directly call `jpy.get_type` and such, I elected to explicitly list it. Thoughts?

py/jpy/build.gradle
```diff
@@ -30,4 +34,10 @@ Docker.registerDockerTask(project, 'buildWheel') {
     copyOut {
         into 'build/deephaven-jpy-wheel'
     }
-}
\ No newline at end of file
+}
+
+artifacts {
+    pythonWheel(buildWheel.get().outputs.files.singleFile) {
```
niloc132, line 40: Note to self, double check this, this might be able to be simplified

py/server/build.gradle
```diff
@@ -0,0 +1,7 @@
+plugins {
```
niloc132, line 1: Note to other reviewers, this replaces `py/deephaven2-wheel/build.gradle`

settings.gradle
```diff
@@ -270,11 +255,28 @@ include(':application-mode')
 include(':util-immutables')
 project(':util-immutables').projectDir = file('Util/util-immutables')
 
+include(':deephaven-jpy')
```
niloc132, line 258: note to other reviewers, I flattened this out since it is a net decrease in actual code, and just as clear as all the other renames we do, but without two different lambdas trying to define these projects

docker/runtime-base/build.gradle
```diff
@@ -4,30 +4,23 @@ plugins {
...
+configurations {
+    pythonWheel
+}
 
-def prepareDocker = project.tasks.register('prepareDocker', Sync) {
-    // deephaven-jpy.whl
-    def deephavenJpyWheel = project(':deephaven-jpy').tasks.getByName('buildWheel')
+dependencies {
+    pythonWheel project(':deephaven-jpy')
+    pythonWheel project(':deephaven-wheel')
```
niloc132, line 14: note to self, actually rename this project in more than its docker image name

niloc132:
There are some conflicts waiting for me, as the moved files were modified after I made the commit that relocated them. Ideally I'd like to get a first round of review done before I redo that move, so as to not keep moving the files over and over as changes happen upstream.

jmao-denver (approved):

docker/server/src/main/docker/Dockerfile
```diff
@@ -13,20 +13,10 @@ RUN set -eux; \
...
-RUN set -eux; \
-    python -m pip install -q --no-index --no-cache-dir /deephaven-wheel/*.whl; \
-    rm -r /deephaven-wheel
-
-COPY deephaven2-wheel/ /deephaven2-wheel
-RUN set -eux; \
-    python -m pip install -q --no-index --no-cache-dir /deephaven2-wheel/*.whl; \
-    rm -r /deephaven2-wheel
+    python -m pip install -q --no-index --no-cache-dir /wheels/*.whl; \
+    rm -r /wheels
```
devinrsmith, line 19: We are "losing" a bit of docker cachability here, but I doubt it's of much consequence.

docker/server/build.gradle
devinrsmith: My preference would be to leave it out, unless we have code in `docker/server` (project) that refers to `jpy`.

devinrsmith (approved):

jmao-denver (approved):
LGTM

devinrsmith (approved):

PR #2251, opened 2022-04-12
Fix gradle not telling us docker tasks are failing, and the actual failure

niloc132:
Gradle was just a bit too clever for us here - the Sync task at the end of the registerDockerTask was set to fail if the actual command had failed, but it never actually ran in some cases, so it never ran the action that would have emitted the failure, so the test seemed to pass.

The underlying python issue then wasn't detected - as part of the refactor, an env var was added to point to the full python directory, but this must be invoked in a particular way. Note that gradle (or docker?) doesn't always seem to detect the `+x` that is now set on the `run-tests.sh`, developers may need to run `:Integrations:clean` and `docker image prune` once to shake the cached images out.


jmao-denver (approved):
LGTM

PR #2263, opened 2022-04-13
Add wrappers for application mode

### LINKED ISSUE #2245: App mode library doesn't work with new python tables
**Description**

The following application mode code does not work with the new python library

```
import jpy
from typing import Callable

ApplicationState = jpy.get_type("io.deephaven.appmode.ApplicationState")

#ApplicationState examples
def start(app: ApplicationState):
    table = empty_table(5)
    table = table.update(["Values = i"])

    app.setField("table", table)

def initialize(func: Callable[[ApplicationState], None]):
  app = jpy.get_type("io.deephaven.appmode.ApplicationContext").get()
  func(app)

initialize(start)
```

Error thrown
```
server_1      | Type: <class 'ValueError'>
server_1      | Value: cannot convert a Python 'Table' to a Java 'java.lang.Object'
```

Changing the setField call to `app.setField("table", table.j_object)` makes it work as expected

**Expected results**

The code above works

**Actual results**

The code above errors out

**Versions**
 - Deephaven: edge
 - OS: osx
 - Browser: chrome
 - Docker: Docker version 20.10.8, build 3967b7d

jmao-denver:
Fixes #2245

### DIFF
--- a/py/server/deephaven/appmode.py
+++ b/py/server/deephaven/appmode.py
@@ -1,0 +1,66 @@
+#
+#   Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+#
+""" This module supports writing Deephaven application mode Python scripts. """
+from typing import Dict
+
+import jpy
+
+from deephaven import DHError
+from deephaven._wrapper import JObjectWrapper, wrap_j_object, unwrap
+
+_JApplicationContext = jpy.get_type("io.deephaven.appmode.ApplicationContext")
+_JApplicationState = jpy.get_type("io.deephaven.appmode.ApplicationState")
+
+
+class ApplicationState(JObjectWrapper):
+    """ The ApplicationState represents the state of an application. """
+    j_object_type = _JApplicationState
+
+    @property
+    def j_object(self) -> jpy.JType:
+        return self.j_app_state
+
+    def __init__(self, j_app_state):
+        self.j_app_state = j_app_state
+
+    def __repr__(self):
+        return f"id: {self.j_app_state.id()}, name: {self.j_app_state.name()}"
+
+    def __getitem__(self, item):
+        item = str(item)
+        value = wrap_j_object(self.j_app_state.getField(item))
+        if not value:
+            raise KeyError(item)
+
+    def __setitem__(self, key, value):
+        key = str(key)
+        self.j_app_state.setField(key, unwrap(value))
+
+    def __delitem__(self, key):
+        key = str(key)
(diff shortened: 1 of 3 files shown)


pyintegration/deephaven/appmode.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __getitem__(self, item):
+        item = str(item)
+        value = wrap_j_object(self.j_app_state.getField(item))
+        if not value:
+            raise KeyError(item)
+
+    def __setitem__(self, key, value):
+        key = str(key)
+        self.j_app_state.setField(key, unwrap(value))
```
devinrsmith, line 38: Note: the item you are returning from get is not the same item you are placing via set. I see you *are* calling `.value()` in fields() - I'm not sure if we want to have JObjectWrapper Field, or return `.value()` from a get.

jakemulf (dismissed):
From the use case that I was looking to replicate, this looks good, but I'd consider Devin's feedback in case there's more use cases that I may have missed

pyintegration/deephaven/appmode.py
jmao-denver: I think Python dict is a good fit, wrapping Field seems a bit overkill.

jakemulf (approved):

pyintegration/deephaven/appmode.py
devinrsmith: Sure - but I just want to triple check:

```python
app = get_app_state()
app['hello'] = 'world'
print(app['hello'])
```
will *not* print `world` b/c getField returns a Field...

devinrsmith (commented):
See earlier comment.

pyintegration/deephaven/appmode.py
jmao-denver: Thanks for finding the bug!

pyintegration/test_helper/__init__.py
```diff
@@ -61,7 +63,13 @@ def start_jvm():
             'workspace': os.path.realpath(workspace),
 
         }
-        jvm_options= {
+
+        if jvm_props:
+            jvm_properties.update(jvm_props)
+
+        print(jvm_properties)
```
devinrsmith, line 70: Was this on purpose, or meant to be debugging only?

jmao-denver: good catch again, I really should have taken a break and not been rushing things.

jakemulf (approved):

PR #2267, opened 2022-04-14
Allow deephaven_legacy to not be installed

niloc132:
This will likely be removed soon, but in the meantime it would permit a user to install the deephaven and deephaven_server wheels to start the server from python without also installing deephaven_legacy.

Partial #2221

### DIFF
--- a/py/server/deephaven/server/script_session/__init__.py
+++ b/py/server/deephaven/server/script_session/__init__.py
@@ -1,5 +1,8 @@
 # Implementation utilities for io.deephaven.engine.util.PythonDeephavenSession
-from deephaven_legacy.Plot import FigureWrapper
+try:
+    from deephaven_legacy.Plot import FigureWrapper
+except ImportError:
+    FigureWrapper = None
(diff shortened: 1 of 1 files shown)


py/server/deephaven/server/script_session/__init__.py
```diff
@@ -1,5 +1,8 @@
 # Implementation utilities for io.deephaven.engine.util.PythonDeephavenSession
-from deephaven_legacy.Plot import FigureWrapper
+try:
+    from deephaven_legacy.Plot import FigureWrapper
+except:
+    FigureWrapper = None
```
jmao-denver, line 5: except ImportError ?

jmao-denver (dismissed):
The change is safe and future-proof.

py/server/deephaven/server/script_session/__init__.py
niloc132: Thanks, I'll correct that.

jmao-denver (approved):

PR #2270, opened 2022-04-14
Defensively include Dockerfile

niloc132:
This is a problem as I make the "build this wheel in docker" more customizable, as it is possible to provide "includes" and "excludes" rules for the wheel that accidentally exclude the Dockerfile itself, thus making the project unbuildable. This commit ensures that the dockerfile itself will always be included when the image is prepared.

Partial #2221

### DIFF
--- a/buildSrc/src/main/groovy/Docker.groovy
+++ b/buildSrc/src/main/groovy/Docker.groovy
@@ -249,9 +249,13 @@ class Docker {
             sync.into dockerWorkspaceContents
 
             if (cfg.dockerfileFile) {
-                sync.from cfg.dockerfileFile
+                sync.from(cfg.dockerfileFile) { CopySpec dockerfileCopy ->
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/Docker.groovy
```diff
@@ -249,9 +249,13 @@ class Docker {
             sync.into dockerWorkspaceContents
 
             if (cfg.dockerfileFile) {
-                sync.from cfg.dockerfileFile
+                sync.from(cfg.dockerfileFile) { CopySpec dockerfileCopy ->
+                    dockerfileCopy.include('Dockerfile')
```
devinrsmith, line 253: Should we take the provided file name instead of assuming Dockerfile?

niloc132: Yes, that's a good idea, I'll correct that.

PR #2287, opened 2022-04-19
Support very large tables

### LINKED ISSUE #2175: Support for subscribing / snapshotting very large tables
As a user, I want to be able to snapshot and subscribe to very large tables for analysis and computation.

Currently if a table contains more than 2^31 rows, an array or string column with accumulated length more than 2^31 values, or a column requiring marshalling with accumulated byte length more than 2^31, then the server will generate an error.  Update the engine to dynamically adjust snapshot and delta sizes to avoid exceeding these limits.

lbooker42:
Adds support for snapshotting  very large tables (i.e. contains more than 2^31 rows, an array or string column with accumulated length more than 2^31 values, or a column requiring marshalling with accumulated byte length more than 2^31)

Closes #2175


server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -104,14 +103,59 @@ public View getSchemaView(final TableDefinition table,
...
-            data = ChunkInputStreamGenerator.makeInputStreamGenerator(
-                    col.data.getChunkType(), col.type, col.componentType, col.data);
+            data = new ChunkListInputStreamGenerator(col);
+
+            // build the row offsets for this column chunks
+            long offset = 0;
+            modChunkRowSets = new RowSet[col.data.size()];
+            for (int chunkIdx = 0; chunkIdx < col.data.size(); ++chunkIdx) {
+                int chunkSize = col.data.get(chunkIdx).size();
+                modChunkRowSets[chunkIdx] = RowSetFactory.fromRange(offset, offset + chunkSize - 1);
```
lbooker42, line 156: The RowSets in `modChunkRowSets` need to be closed.

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -187,7 +249,7 @@ public void close() {
         rowsRemoved.close();
 
         if (addColumnData != null) {
```
lbooker42, line 251: Close the mod/add RowSets here...

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -658,17 +773,40 @@ private long appendAddColumns(final View view, final long startRange, final long
...
-                        col.getInputStream(view.options(), myAddedOffsets);
-                drainableColumn.visitFieldNodes(fieldNodeListener);
-                drainableColumn.visitBuffers(bufferListener);
-
-                // Add the drainable last as it is allowed to immediately close a row set the visitors need
-                addStream.accept(drainableColumn);
+            for (final ChunkListInputStreamGenerator chunkSetGen : addColumnData) {
+
+                if (chunkSetGen.size() == 0) {
+                    System.out.println("oops");
```
lbooker42, line 781: Debugging structure? needs removed

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -658,17 +773,40 @@ private long appendAddColumns(final View view, final long startRange, final long
...
+                    final ChunkInputStreamGenerator generator = chunkSetGen.generators[i];
+
+                    final long shiftAmount = -addChunkRowSets[i].firstRowKey();
+
+                    // get an offset rowset for each chunk in the set
+                    try (final WritableRowSet adjustedOffsets = myAddedOffsets.intersect(addChunkRowSets[i])) {
+                        // normalize this to the chunk offsets
+                        adjustedOffsets.shiftInPlace(shiftAmount);
+
+                        // System.out.println("myAddedOffsets: " + myAddedOffsets + ", adjustedOffsets: " +
```
lbooker42, line 796: remove comments

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -768,10 +918,10 @@ private ByteBuffer getSubscriptionMetadata(final SubView view) throws IOExceptio
         final int nodesOffset = metadata.endVector();
 
         BarrageUpdateMetadata.startBarrageUpdateMetadata(metadata);
-        BarrageUpdateMetadata.addNumAddBatches(metadata,
-                LongSizedDataStructure.intSize("BarrageStreamGenerator", view.numAddBatches));
-        BarrageUpdateMetadata.addNumModBatches(metadata,
-                LongSizedDataStructure.intSize("BarrageStreamGenerator", view.numModBatches));
+        BarrageUpdateMetadata.addNumAddBatches(metadata, LongSizedDataStructure.intSize("BarrageStreamGenerator",
```
lbooker42, line 921: The concept of number of batches is deprecated IMO.  We should change this to number of rows or remove (since num rows is implicit in rowsIncluded / rowsModified).

lbooker42 (commented):
Self-review

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/VarBinaryChunkInputStreamGenerator.java
```diff
@@ -223,11 +355,17 @@ public int drainTo(final OutputStream outputStream) throws IOException {
             dos.writeInt(0);
 
             final MutableInt logicalSize = new MutableInt();
-            subset.forAllRowKeys((rawRow) -> {
+            subset.forAllRowKeys((idx) -> {
```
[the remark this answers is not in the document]
lbooker42: This is what I came up with, can it be improved?
`
                        final int startArrayIndex;
                        final int startOffset;
                        if (lastIdx.longValue() == idx) {
                            // we can optimize
                            startArrayIndex = lastIdxArrayIdx.intValue();
                            startOffset = lastIdxOffset.intValue();
                        } else {
                            startArrayIndex = myByteStorage.getByteArrayIndex((int)idx);
                            startOffset = myByteStorage.offsets.get((int)idx);
                        }

                        final int endArrayIndex = myByteStorage.getByteArrayIndex((int)idx + 1);
                        final int endOffset = myByteStorage.offsets.get((int)idx + 1);

                        if (startArrayIndex == endArrayIndex) { // same byte array, can optimize
                            logicalSize.add(endOffset - startOffset);
                        } else {
                            // need to span multiple byte arrays
                            long byteCount = myByteStorage.getByteArraySize(startArrayIndex) - startOffset;
                            for (int midArrayIndex = startArrayIndex + 1; midArrayIndex < endArrayIndex; midArrayIndex++) {
                                byteCount += myByteStorage.getByteArraySize(midArrayIndex);
                            }
                            byteCount += endOffset;
                            logicalSize.add(byteCount);
                        }

                        // store the data for potential re-use
                        lastIdx.setValue(idx + 1);
                        lastIdxArrayIdx.setValue(endArrayIndex);
                        lastIdxOffset.setValue(endOffset);
`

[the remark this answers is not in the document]
lbooker42: Now that I think about it, the `ByteStorage` class could just save this data (pseudo-memo a single but very likely re-used data point).

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -658,17 +879,49 @@ private long appendAddColumns(final View view, final long startRange, final long
...
+                    try (final RowSet empty = RowSetFactory.empty()) {
+                        final ChunkInputStreamGenerator.DrainableColumn drainableColumn =
+                                chunkSetGen.generators[0].getInputStream(view.options(), empty);
+                        drainableColumn.visitFieldNodes(fieldNodeListener);
+                        drainableColumn.visitBuffers(bufferListener);
+
+                        // Add the drainable last as it is allowed to immediately close a row set the visitors need
+                        addStream.accept(drainableColumn);
+                    }
+                }
```
nbauernfeind, line 924: I think I'd restructure this method a little:
```
// find chunk index for the request
// if chunk DNE then I think you should throw an exception -- this seems unexpected and should not be ignored by writing empty columns
// else write chunk for all columns and return `adjustedOffsets.size()` -- noting that adjustedOffsets should be re-used for all add columns
```

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -699,14 +952,43 @@ private long appendModColumns(final View view, final long startRange, final long
...
+                    try (final RowSet empty = RowSetFactory.empty()) {
+                        final ChunkInputStreamGenerator.DrainableColumn drainableColumn =
+                                mcd.data.generators[0].getInputStream(view.options(), empty);
+                        drainableColumn.visitFieldNodes(fieldNodeListener);
+                        drainableColumn.visitBuffers(bufferListener);
+
+                        // Add the drainable last as it is allowed to immediately close a row set the visitors need
+                        addStream.accept(drainableColumn);
+                    }
+                }
```
nbauernfeind, line 991: similar comments for adds -- but, obviously, the `adjustedOffsets` is unique per column. For return value I would select the max rows written across all columns.

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -658,17 +879,49 @@ private long appendAddColumns(final View view, final long startRange, final long
...
+                            // normalize this to the chunk offsets
+                            adjustedOffsets.shiftInPlace(shiftAmount);
+
+                            final ChunkInputStreamGenerator.DrainableColumn drainableColumn =
+                                    generator.getInputStream(view.options(), adjustedOffsets);
+                            drainableColumn.visitFieldNodes(fieldNodeListener);
+                            drainableColumn.visitBuffers(bufferListener);
+
+                            // Add the drainable last as it is allowed to immediately close a row set the visitors need
+                            addStream.accept(drainableColumn);
```
nbauernfeind, line 907: It should be an error if `endRange` exceeds the rows in this chunk. You are attempting to guarantee this condition by calculating a `chunkOffset` on the way into this method, but there is no debug-friendly message if it is ever miscalculated.

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -658,17 +879,49 @@ private long appendAddColumns(final View view, final long startRange, final long
...
+                                chunkSetGen.generators[0].getInputStream(view.options(), empty);
+                        drainableColumn.visitFieldNodes(fieldNodeListener);
+                        drainableColumn.visitBuffers(bufferListener);
+
+                        // Add the drainable last as it is allowed to immediately close a row set the visitors need
+                        addStream.accept(drainableColumn);
+                    }
+                }
             }
             return myAddedOffsets.size();
```
nbauernfeind, line 926: The return value needs to be the number of rows written in this record batch, which is `adjustedOffsets.size()`.

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -298,34 +379,126 @@ public SubView(final BarrageStreamGenerator generator,
...
+
+                            // let's write the data
+                            visitor.accept(is);
+
+                            offset += batchSize;
+                            chunkOffset += batchSize;
+                            metadata = null;
+                        }
+                        // recompute the batch limit for the next message
+                        int bytesPerRow = bytesToWrite / batchSize;
```
nbauernfeind, line 440: Maybe add a comment that `getInputStream` might append fewer rows if this batch does not overlap completely with the user's subscription. This is likely ok as it just underestimates bytes per row.

I wonder if it would be worth having `appendAllColumns` try to write the intended range as sliced from `myAddedOffsets`. You could pass a `MutableInt` into `getInputStream` that writes how many rows were actually written -- allowing you to short-cut the case where the "batch" crosses two chunk input stream generators.

Actually, what do you think of that idea? It will eliminate sending a bunch of empty batches to viewports when you have a full subscription and a viewport subscribed to the same table. It is complicated for modifications as the chunk boundaries are not guaranteed to line up. Maybe the compromise for mods is to run through all columns to limit the size of the batch before creating the first column's input stream. It feels like a good compromise.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/VarBinaryChunkInputStreamGenerator.java
```diff
@@ -78,30 +228,27 @@ public void close() {
...
 
         private ObjectChunkInputStream(
                 final StreamReaderOptions options,
-                final IntChunk<ChunkPositions> myOffsets,
-                final byte[] myBytes, final RowSet subset) {
+                final ByteStorage myByteStorage, final RowSet subset) {
             super(chunk, options, subset);
-            this.myBytes = myBytes;
-            this.myOffsets = myOffsets;
+            this.myByteStorage = myByteStorage;
```
nbauernfeind, line 251: This is a non-static inner class -- so we have access to the outer `byteStorage`. I don't know why I did that the first time around. Maybe at one point it was a static inner class...?

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/VarBinaryChunkInputStreamGenerator.java
```diff
@@ -155,30 +300,33 @@ public void visitBuffers(final BufferListener listener) {
...
+                        totalCachedSize.add(myByteStorage.getPayloadSize((int)s, (int)e));
                     });
                 }
 
                 if (!subset.isEmpty() && (subset.size() & 0x1) == 0) {
                     // then we must also align offset array
-                    cachedSize += Integer.BYTES;
+                    totalCachedSize.add(Integer.BYTES);
                 }
+                cachedSize = LongSizedDataStructure.intSize("VarBinaryChunkInputStreamGenerator.getRawSize", totalCachedSize.longValue());
```
nbauernfeind, line 329: Technically this int limit is on the payload; not including the cost for offsets. The offset buffer is limited to 2^32 offsets, not 2^32 bytes worth of offsets - so we should verify that the subset can be intSized(). We should sum up the payload separately.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/VarBinaryChunkInputStreamGenerator.java
```diff
@@ -155,30 +300,33 @@ public void visitBuffers(final BufferListener listener) {
...
 
                 // there are n+1 offsets; it is not assumed first offset is zero
-                if (!subset.isEmpty() && subset.size() == myOffsets.size() - 1) {
-                    cachedSize += myOffsets.size() * Integer.BYTES;
-                    cachedSize += myOffsets.get(subset.intSize(DEBUG_NAME)) - myOffsets.get(0);
+                if (!subset.isEmpty() && subset.size() == myByteStorage.offsets.size() - 1) {
+                    totalCachedSize.add(myByteStorage.offsets.size() * Integer.BYTES);
+                    for (int i = 0; i < myByteStorage.size(); i++) {
+                        totalCachedSize.add(myByteStorage.getByteArraySize(i));
+                    }
```
nbauernfeind, line 313: Another benefit of changing byteStorage to think in overall offsets means that this calculation becomes O(1) again.

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -298,34 +379,126 @@ public SubView(final BarrageStreamGenerator generator,
...
+                            int rowLimit = DEFAULT_MESSAGE_SIZE_LIMIT / bytesPerRow;
+
+                            // add some margin for abnormal cell contents
+                            batchSizeLimit = Math.min(batchSize(), Math.max(1, (int) ((double) rowLimit * 0.9)));
+                        }
+                    } catch (SizeException ex) {
+                        // was an overflow in the ChunkInputStream generator (probably VarBinary). We can't compute the
+                        // correct number of rows from this failure, so cut batch size in half and try again. This may
+                        // occur multiple times until the size is restricted properly
+                        batchSizeLimit = Math.max(1, batchSizeLimit / 2);
```
nbauernfeind, line 451: I just realized that this Math.max check leads to an infinite loop. If we can't send a single row (i.e. batchSizeLimit / 2 == 0) because of size constraints, then we are just simply SOL and we'll need to kill affected subscriptions.

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
lbooker42: submitted barrage PR (https://github.com/deephaven/barrage/pull/28)

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/table/BarrageTable.java
```diff
@@ -424,18 +478,26 @@ private void freeRows(final RowSet rowsToFree) {
         }
 
         // Note: these are NOT OrderedRowKeys until after the call to .sort()
-        try (final WritableLongChunk<OrderedRowKeys> redirectedRows =
-                WritableLongChunk.makeWritableChunk(rowsToFree.intSize("BarrageTable"))) {
-            redirectedRows.setSize(0);
+        final int chunkSize = (int) Math.min(rowsToFree.size(), TWO_DIMENSIONAL_COLUMN_SOURCE_THRESHOLD);
```
nbauernfeind, line 481: It is probably ever-so-slightly better to use a pooled chunk here. I recommend swapping the max with `1 << ChunkPoolConstants.LARGEST_POOLED_CHUNK_LOG2_CAPACITY` (as you already do in the add batch).

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageStreamReader.java
```diff
@@ -219,30 +238,77 @@ public BarrageMessage safelyParseFrom(final StreamReaderOptions options,
...
-                                    columnChunkTypes[ci], columnTypes[ci], componentTypes[ci], fieldNodeIter,
-                                    bufferInfoIter, ois, (WritableChunk<Values>) msg.addColumnData[ci].data,
-                                    numAddRowsRead, numRowsTotal);
+                            final BarrageMessage.AddColumnData acd = msg.addColumnData[ci];
+
+                            int lastChunkIndex = acd.data.size() - 1;
+
+                            // need to add the batch row data to the column chunks
+                            WritableChunk<Values> chunk = (WritableChunk<Values>) acd.data.get(lastChunkIndex);
+                            int chunkSize = chunk.size();
```
nbauernfeind, line 252: So, for the snapshot case where you've added `null` won't this NPE? 

It seems incorrect now for `extractChunkFromInputStream` to have the feature to return/replace the provided chunk. As in, it seems like we could make it an error if a large enough chunk is not provided. You can fix the NPE for this PR, and change the extractChunk interface in a follow up?

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -660,31 +711,107 @@ private static int createByteVector(final FlatBufferBuilder builder, final byte[
...
+                if (bytesPerRow > 0) {
+                    int rowLimit = DEFAULT_MESSAGE_SIZE_LIMIT / bytesPerRow;
 
-                // Add the drainable last as it is allowed to immediately close a row set the visitors need
-                addStream.accept(drainableColumn);
+                    // add some margin for abnormal cell contents
+                    batchSize = Math.min(maxBatchSize, Math.max(1, (int) ((double) rowLimit * 0.9)));
+                }
+            } catch (SizeException ex) {
+                // was an overflow in the ChunkInputStream generator (probably VarBinary). We can't compute the
```
nbauernfeind, line 750: You need to handle the case when `batchSize` is already 1; at least for completeness if it is currently impossible. Though I think it is possible if `addStream.accept` can throw a SizeException for the sum of columns exceeding max int.

nbauernfeind (commented):
We'll still need a large-table implementation for VarListChunkInputStreamGenerator (similar to the ByteStorage approach). I'm happy if that is a separate review.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageStreamReader.java
lbooker42: Moved to #2365

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageStreamReader.java
```diff
@@ -219,30 +237,79 @@ public BarrageMessage safelyParseFrom(final StreamReaderOptions options,
...
+                            int lastChunkIndex = acd.data.size() - 1;
+
+                            // need to add the batch row data to the column chunks
+                            WritableChunk<Values> chunk = (WritableChunk<Values>) acd.data.get(lastChunkIndex);
+                            int chunkSize = chunk == null ? 0 : chunk.size();
+
+                            final int chunkOffset;
+                            long rowOffset = numAddRowsRead - lastAddStartIndex;
+                            // reading the rows from this batch might overflow the existing chunk
+                            if (rowOffset + batch.length() > chunkSize) {
```
nbauernfeind, line 256: If you are intending to overwrite the null, then this if-block should not create a new chunk, right? Perhaps you shouldn't add the null chunk and instead allow this block to construct the initial chunk when numAddRowsTotal is zero.

We need to figure out how to get code coverage on this logic. Can you check if there is a way to clear metadata before uploading to the DoPut in the FlightRoundTrip tests?

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageStreamReader.java
```diff
@@ -259,7 +326,7 @@ public BarrageMessage safelyParseFrom(final StreamReaderOptions options,
             }
 
             deserializeTmConsumer.accept(System.nanoTime() - startDeserTm);
-            if (numAddBatchesRemaining + numModBatchesRemaining == 0) {
+            if (numAddRowsRead == numAddRowsTotal && numModRowsRead == numModRowsTotal) {
```
nbauernfeind, line 329: We should also deliver this if `numAddRowsTotal == 0 && numModRowsTotal == 0` - maybe it would be worthwhile creating a `final boolean noDeephavenMetadata` earlier in the logic (to re-use above and here).

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -775,10 +923,10 @@ private ByteBuffer getSubscriptionMetadata(final SubView view) throws IOExceptio
...
 
         BarrageUpdateMetadata.startBarrageUpdateMetadata(metadata);
-        BarrageUpdateMetadata.addNumAddBatches(metadata,
-                LongSizedDataStructure.intSize("BarrageStreamGenerator", view.numAddBatches));
-        BarrageUpdateMetadata.addNumModBatches(metadata,
-                LongSizedDataStructure.intSize("BarrageStreamGenerator", view.numModBatches));
+        BarrageUpdateMetadata.addNumAddBatches(metadata, LongSizedDataStructure.intSize("BarrageStreamGenerator",
+                (view.numAddRows + view.batchSize() - 1) / view.batchSize()));
+        BarrageUpdateMetadata.addNumModBatches(metadata, LongSizedDataStructure.intSize("BarrageStreamGenerator",
+                (view.numModRows + view.batchSize() - 1) / view.batchSize()));
```
nbauernfeind, line 929: This isn't accurate, is it? It's probably better to not set any value for this, what do you think?

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/VarBinaryChunkInputStreamGenerator.java
```diff
@@ -7,35 +7,184 @@
...
+        @Override
+        public void close() {
+            try {
+                super.close();
+            } catch (IOException e) {
+                // ignore this error
+            }
+
+            // close the offset and byte chunks
+            if (offsets instanceof PoolableChunk) {
```
nbauernfeind, line 176: `WriteableChunk`s are Poolable, you do not need the instanceof checks or the cast to PoolableChunk to call `close`.

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
lbooker42: Until we implement the `barrage` repo update, we need to provide this.  It is accurate unless the message size limit is exceeded or we cross a chunk boundary both of which are unlikely in regular use.

nbauernfeind (approved):

PR #2288, opened 2022-04-19
Primitives (engine-functions) library rewrite

### LINKED ISSUE #72: Rewrite primitives library and related code generation
The primitives library is a mess.  We need better code generation, and the library needs to be rewritten with it.

chipkent:
Rewrite of the primitives library.

This is based off of the previously reviewed library for DHE.  Changes were made to avoid function conflicts between generic functions and primitive functions.  The generic functions are now named `<method>Obj`.

Resolves #72.


niloc132 (approved):
+1 on the gradle files

jakemulf (dismissed):
Looks good to me. The groovy static imports file was a bit hard to review because the diff got a bit messed up, but nothing stood out as problematic or concerning to me. I like the `xyz` vs `xyzObj` pattern

engine/function/src/templates/Numeric.ftl
```diff
@@ -0,0 +1,2358 @@
...
+        return ${pt.boxed}.isNaN(value);
+    }
+
+    /**
+     * Returns {@code true} if the value is infinite and {@code false} otherwise.
+     *
+     * @param value value.
+     * @return {@code true} if the value is infinite and {@code false} otherwise.
+     */
+    static public boolean isInf(${pt.boxed} value) {
```
kosak, line 2103: As above I might suggest

```
return value != null && ${pt.boxed}.isInfinite(value);
```

engine/function/src/templates/Numeric.ftl
```diff
@@ -0,0 +1,2358 @@
...
+        return isNormal(castDouble(value));
+    }
+
+    /**
+     * Returns {@code true} if the value is normal, where "normal" is defined as not infinite, not NaN, and not null.
+     *
+     * @param value value.
+     * @return {@code true} if the value is not infinite, NaN, nor null; {@code false} otherwise
+     */
+    static public boolean isNormal(${pt.primitive} value) {
```
kosak, line 2133: If it's not too late I'd probably propose naming the method `isFinite` rather than `isNormal` (thus avoiding the problematic term Normal which means something else to floating-point people) and I'd probably propose writing
```
return ${pt.boxed}.isFinite(value) && !isNull(value)
```

engine/function/src/templates/Cast.ftl
```diff
@@ -0,0 +1,407 @@
...
+     *
+     * @param value value
+     * @param checkFidelity check to see if the cast preserves the value.
+     * @return cast value.
+     */
+    public static int castInt(long value, boolean checkFidelity) {
+        if( value == NULL_LONG ){
+            return NULL_INT;
+        }
+
```
kosak, line 82: I think you're going to need a second check that looks like this

```
        if (checkFidelity && value == NULL_INT) {
            throw new CastDoesNotPreserveValue("Casting is not supported because the value, " + value + ", represents the null integer sentinel and therefore cannot be cast to int.");
        }
```

engine/function/src/templates/Cast.ftl
```diff
@@ -0,0 +1,407 @@
...
+         * Creates a new exception.
+         *
+         * @param message error message.
+         */
+        public CastDoesNotPreserveValue(final String message) {
+            super(message);
+        }
+    }
+
+    private static boolean isLosingPrecisionDouble(final long v) {
```
kosak, line 34: I think we need a more precise test here. There are lots of long values greater than 9007199254740992 that convert to double without losing precision. For example all the powers of two. One example is 2^63.

Maybe a test like this would work: `v != (long)(double)v`
but I'm not sure about the edge cases

kosak (changes requested):

engine/function/src/templates/Cast.ftl
chipkent: hmmm.... When I implemented this originally, I was just looking at a range of acceptable values.  I hadn't considered the powers of two, or anything like that.  I can convince myself that either the range case or the `v != (long)(double)v` is the "right one".  I made the suggested change, but I'm not 100% convinced it is the right way to go... and I'm not 100% convinced it is the wrong way to go.  Thoughts?

kosak: I'm kind of torn myself really. I'm really not sure any more.

I think maybe the *original* way you had it might be less surprising to the user, because it represents a dense range of acceptable values from -9007199254740992L to 9007199254740992L, where every value in the range is legal.

So maybe go back to the way you had it, and if further enlightenment strikes, we can revisit the issue later?

kosak (approved):

engine/function/src/templates/Numeric.ftl
```diff
@@ -2115,48 +2115,48 @@ public class Numeric {
     }
 
     /**
-     * Returns {@code true} if the value is normal, where "normal" is defined as not infinite, not NaN, and not null.
+     * Returns {@code true} if the value is fine, where "finite" is defined as not infinite, not NaN, and not null.
```
kosak, line 2168: "fine" -> "finite"

engine/function/src/templates/Numeric.ftl
```diff
@@ -1820,6 +1841,32 @@ public class Numeric {
...
+    <#if pt.valueType.isFloat >
+
+    /**
+     * Returns the upper bound of the bin containing the value.
+     *
+     * @param value value.
+     * @param interval bin width.
+     * @return upper bound of the bin containing the value.
+     */
+    public static ${pt.primitive} upperBin(${pt.primitive} value, ${pt.primitive} interval) {
```
kosak, line 1853: I'm not exactly sure what the definition of "upperBin" should be, but I felt a little "surprised" by the assymetry between lowerBin and upperBin:

| v | lowerBin(v,3) | upperBin(v,3) |
|--|------------|-------------|
| 0 | 0 | 0 |
| 1 | 0 | 3 |
| 2 | 0 | 3 | 
| 3 | 3 | 3 | 
| 4 | 3 | 6 |
| 5 | 3 | 6 |
| 6 | 6 | 6 |
| 7 | 6 | 9 |
| 8 | 6 | 9 |
| 9 | 9 | 9 |
| 10 |  9 |  12 |
| 11 | 9 |  12 |

I guess you're going for an analogy to floor() and ceil() so maybe it's fine.

In any case, I think maybe it might help if there were a comment with an algebraic relation like

```
lowerBin: if value >= k * interval and value < (k + 1) * interval for some integer k, return k * interval
```

(Also I'm kind of punting on what happens if you pass in negative numbers... I'm not sure that's legal... do we want to check for negative numbers and throw?)

and then the definition for upperBin is something like

```
upperBin: if value > k * interval and value <= (k + 1) * interval for some integer k, return (k + 1) * interval
```
Which I *think* matches the code you've written. Thinking about the sort-of-special-case zero, that works too, if we allow k to be -1.

So to summarize I guess I'm requesting that we have a crisp algebraic formula like the above in the comments. Or some other way of expressing it if you prefer.

kosak (dismissed):

engine/function/src/templates/Numeric.ftl
chipkent: That code was a mess.  I _think_ it is now resolved.

engine/function/src/templates/Numeric.ftl
```diff
@@ -0,0 +1,2446 @@
...
+     *
+     * @param value value.
+     * @param interval bin width.
+     * @return lower bound of the bin containing the value.
+     */
+    public static ${pt.primitive} lowerBin(${pt.primitive} value, ${pt.primitive} interval) {
+        if (value == ${pt.null} || interval == ${pt.null}) {
+            return ${pt.null};
+        }
+
```
kosak, line 1805: Probably want the `interval <= 0` check here as well, as in

```
        if ( interval <= 0 ) {
            throw new IllegalArgumentException("Interval is not positive: " + interval);
        }
```

engine/function/src/templates/Numeric.ftl
```diff
@@ -0,0 +1,2446 @@
...
+     * @param value value.
+     * @param interval bin width.
+     * @return upper bound of the bin containing the value.
+     */
+    public static ${pt.primitive} upperBin(${pt.primitive} value, ${pt.primitive} interval) {
+        if (value == ${pt.null} || interval == ${pt.null}) {
+            return ${pt.null};
+        }
+
+        return (${pt.primitive}) (interval * Math.ceil(value / interval));
```
kosak, line 1865: Also I think we want the `interval <= 0` check here too:

```
        if ( interval <= 0 ) {
            throw new IllegalArgumentException("Interval is not positive: " + interval);
        }
```

engine/function/src/templates/Numeric.ftl
```diff
@@ -0,0 +1,2446 @@
...
+            return ${pt.null};
+        }
+
+        if ( interval <= 0 ) {
+            throw new IllegalArgumentException("Interval is not positive: " + interval);
+        }
+
+        final long d = ((long)value) / ((long)interval);
+        final long m = ((long)value) % ((long)interval);
+        final long r = (m != 0 && value < 0) ? d - 1 : d;
```
kosak, line 1829: `value < 0` implies `m < 0` so we could delete `m != 0` from this if statement.
(I think)

engine/function/src/templates/Numeric.ftl
```diff
@@ -0,0 +1,2446 @@
...
+
+
+    <#if pt.valueType.isFloat >
+
+    /**
+     * Returns the lower bound of the bin containing the value.
+     *
+     * @param value value.
+     * @param interval bin width.
+     * @return lower bound of the bin containing the value.
```
kosak, line 1799: In the previous review I suggested that the Javadocs all explain what the mathematical formula is for calculating lowerBin/upperBin (because callers won't see the code).

If you disagree, that's fine, I'm not insisting. I'm just repeating the suggestion here in case it was overlooked.

kosak (approved):
some nits

engine/function/src/templates/Numeric.ftl
chipkent: Actually, `value < 0` does not imply `m < 0`.  See here:

image

Removing `m != 0` results in unit test failures.

kosak: um, oops. thanks for verifying. sorry.

kosak (dismissed):

kosak (dismissed):
Spotless change approved.

kosak (approved):

PR #2289, opened 2022-04-19
Fix flakiness in test_perfmon.py for test_metrics_get_counters.

jcferretti:


jmao-denver (approved):
I think the change and the explanation make sense.

PR #2290, opened 2022-04-19
ring table for kafka

devinrsmith:


devinrsmith:
Will likely want to follow-up with docs and examples update.

jcferretti (dismissed):

py/server/deephaven/stream/kafka/consumer.py
```diff
@@ -74,6 +58,64 @@ def j_object(self) -> jpy.JType:
...
+        return TableType(TableType.j_object_type.ring(capacity))
+
+    @staticmethod
+    def ring_map(capacity : int):
+        """ Similar to ring(capacity), but each partition is mapped to a distinct in-memory ring table. """
+        return TableType(TableType.j_object_type.ringMap(capacity))
+
+    def __init__(self, j_spec: jpy.JType):
+        self._j_spec = j_spec
+
```
jmao-denver, line 99: is self._j_table_type a better name?

jmao-denver (dismissed):

py/server/deephaven/stream/kafka/consumer.py
devinrsmith: Yep, done.

jcferretti (approved):

PR #2295, opened 2022-04-19
Fix naming inconsistency and default param values

### LINKED ISSUE #2284: Table joins and head_by etc. methods should have the 'on' and 'by' parameters as optional. Also select_distinct should be using the param name 'formulas' instead of cols'
When these parameters are None, it just means performing the operation on all the full table.

jmao-denver:
Fixes #2284

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -387,13 +387,13 @@ class Table(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "table select operation failed.") from e
 
-    def select_distinct(self, cols: Union[str, Sequence[str]] = None) -> Table:
+    def select_distinct(self, formulas: Union[str, Sequence[str]] = None) -> Table:
         """The select_distinct method creates a new table containing all of the unique values for a set of key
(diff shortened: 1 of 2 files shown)


py/server/deephaven/table.py
```diff
@@ -618,13 +618,13 @@ def reverse(self) -> Table:
...
 
-    def sort(self, order_by: Union[str, Sequence[str]],
+    def sort(self, order_by: Union[str, Sequence[str]] = None,
              order: Union[SortDirection, Sequence[SortDirection]] = None) -> Table:
         """The sort method creates a new table where (1) rows are sorted in a smallest to largest order based on the
         order_by column(s) (2) where rows are sorted in the order defined by the order argument.
 
         Args:
-            order_by (Union[str, Sequence[str]]): the column(s) to be sorted on
+            order_by (Union[str, Sequence[str]]): the column(s) to be sorted on, default is None
```
niloc132, line 627: i dont think this can be none, you can't sort on nothing - the `order` won't make sense without an `order_by` right?

niloc132 (approved):

Issue #2273, opened 2022-04-14
Improve layout hint front and end API naming


mattrunyon:
The method names in `LayoutHintBuilder` are inconsistent/unclear to the intention around `atFront` and `atEnd`.

They are inconsistent because `front` and `end` have different connotations. Should probably be `front` and `back` or `start` and `end`. There is also `getFrontCols` and `getBackCols`, so `atEnd` is paired with `getBackCols`

The method could also be more clear to what it does by changing to something like `lockFront` or `lockToFront`. Not sure on this one as there is also `freeze` which is just `atFront` but also the column floats for scrolling purposes

PR #2296, opened 2022-04-19
Deephaven2 Python layout hints

### LINKED ISSUE #2262: Support LayoutHints in Python v2
The Python v2 API does not support LayoutHints.

It should have a `table.setLayoutHints` method and be able to use `LayoutHintBuilder` in some way.

The legacy version gets `LayoutHintBuilder` from jpy

Example Groovy code
```groovy

hints = LayoutHintBuilder.get().freeze("NameOfIntCol");

t = newTable(
   stringCol("NameOfStringCol", "Data String 1", 'Data String 2', "Data String 3"),
   intCol("NameOfIntCol", 4, 5, 6)
).setLayoutHints(hints.build());

println t.getAttribute("LayoutHints")
```

Rough example of Python code that should work
```python
from deephaven import new_table, LayoutHintBuilder
from deephaven.column import string_col, int_col

hints = LayoutHintBuilder.get().freeze("NameOfIntCol")

t = new_table([
   string_col("Name_Of_String_Col", ["Data String 1", 'Data String 2', "Data String 3"]),
   int_col("Name_Of_Int_Col", [4, 5, 6])
]).set_layout_hints(hints.build())

# Alternatively
# hints.applyToTable(t)
# This calls table.setLayoutHints(hints.build()) internally
```

mattrunyon:
Fixes #2262. Also addresses #2273 for the Python API

Adds `table.layout_hints()` in Python. This supports 4 params which are supported by the web UI: `front`, `back`, `freeze`, and `hide`. These are arrays of column names or just a string with a single column name.

The engine does not check the column names exist. The hints are just a string applied as an attribute to the table. The web UI validates the layout hints string.

I used this Python snippet to test

```python
from deephaven.column import string_col, int_col

hints = {
   'front': ['D'],
   'back': ['A'],
}

base = new_table([
   string_col("A", ["A"]),
   int_col("B", [2]),
   string_col("C", ["C"]),
   int_col("D", [4]),
   string_col("E", ["E"]),
   int_col("F", [6])
])


t1 = base.layout_hints(front='D', back=['A'], freeze=["C", 'E'], hide=['F'])
t2 = base.layout_hints(**hints)
```

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -26,6 +26,8 @@ _JPair = jpy.get_type("io.deephaven.api.agg.Pair")
 _JMatchPair = jpy.get_type("io.deephaven.engine.table.MatchPair")
 _JLayoutHintBuilder = jpy.get_type("io.deephaven.engine.util.LayoutHintBuilder")
 
+T = TypeVar('T')
+ValueOrArray = Union[T, List[T]]
 
 class SortDirection(Enum):
     """An enum defining the sorting orders."""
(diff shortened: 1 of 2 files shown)


jmao-denver (changes requested):
1. a single try/except block for all the JPY calls might be better
2. at least one failure test case to cover the most common errors
3. use type alias to make the func signature cleaner

jmao-denver (approved):
I like the approach of separating building the hints and setting the hints on the table into two try/except blocks. This makes the potential exceptions more precise and user friendly

PR #2314, opened 2022-04-21
JPy should release the GIL while in Java, with a special method to deliberately acquire it

niloc132:
There is one exceptional case to deal with, allowing Java to hold and keep the GIL when it wants it, such as for cleanup work. Now Java can grab the GIL if it needs it, without a full trip into python and back again.

In a future patch (or update to this) I might add another similar method to create a PyState from Java but not hold the GIL, so that later py methods can reuse that state instead of making a fresh one per invocation, in case this offers a performance improvement.

### DIFF
--- a/py/jpy-integration/src/pythonToJava/python/test/test_jpy.py
+++ b/py/jpy-integration/src/pythonToJava/python/test/test_jpy.py
@@ -54,7 +54,7 @@ class TestJpy(unittest.TestCase):
 
   def test_has_gil(self):
     PyLib = jpy.get_type('org.jpy.PyLib')
-    self.assertTrue(PyLib.hasGil())
+    self.assertFalse(PyLib.hasGil())
 
   def test_reenter_python(self):
     ReenterPython = jpy.get_type('io.deephaven.jpy.integration.ReenterPython')
--- a/py/jpy/src/main/c/jni/org_jpy_PyLib.c
+++ b/py/jpy/src/main/c/jni/org_jpy_PyLib.c
@@ -2145,6 +2145,21 @@ JNIEXPORT jboolean JNICALL Java_org_jpy_PyLib_hasGil
     return result;
 }
 
+/*
+ * Class:     org_jpy_PyLib
+ * Method:    ensureGil
+ * Signature: (Ljava/util/function/Supplier;)Ljava/lang/Object;
+ */
+JNIEXPORT jobject JNICALL Java_org_jpy_PyLib_ensureGil
+  (JNIEnv* jenv, jclass jLibClass, jobject supplier)
+{
+    jobject result;
+    JPy_BEGIN_GIL_STATE
+    result = (*jenv)->CallObjectMethod(jenv, supplier, JPy_Supplier_get_MID);
+    JPy_END_GIL_STATE
+    return result;
+}
+
 
 /*
  * Class:     org_jpy_python_PyLib
--- a/py/jpy/src/main/c/jpy_jmethod.c
+++ b/py/jpy/src/main/c/jpy_jmethod.c
@@ -268,48 +268,80 @@ PyObject* JMethod_InvokeMethod(JNIEnv* jenv, JPy_JMethod* method, PyObject* pyAr
         JPy_DIAG_PRINT(JPy_DIAG_F_EXEC, "JMethod_InvokeMethod: calling static Java method %s#%s\n", declaringClass->javaName, JPy_AS_UTF8(method->name));
 
         if (returnType == JPy_JVoid) {
+            Py_BEGIN_ALLOW_THREADS;
             (*jenv)->CallStaticVoidMethodA(jenv, classRef, method->mid, jArgs);
+            Py_END_ALLOW_THREADS
             JPy_ON_JAVA_EXCEPTION_GOTO(error);
             returnValue = JPy_FROM_JVOID();
         } else if (returnType == JPy_JBoolean) {
-            jboolean v = (*jenv)->CallStaticBooleanMethodA(jenv, classRef, method->mid, jArgs);
+            jboolean v;
+            Py_BEGIN_ALLOW_THREADS;
+            v = (*jenv)->CallStaticBooleanMethodA(jenv, classRef, method->mid, jArgs);
+            Py_END_ALLOW_THREADS;
             JPy_ON_JAVA_EXCEPTION_GOTO(error);
             returnValue = JPy_FROM_JBOOLEAN(v);
         } else if (returnType == JPy_JChar) {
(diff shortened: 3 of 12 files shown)


py/jpy/src/main/c/jni/org_jpy_PyLib.c
```diff
@@ -2145,6 +2145,21 @@ JNIEXPORT jboolean JNICALL Java_org_jpy_PyLib_hasGil
     return result;
 }
 
+/*
+ * Class:     org_jpy_PyLib
+ * Method:    ensureGil
+ * Signature: (Ljava/util/function/Supplier;)Ljava/lang/Object;
+ */
+JNIEXPORT jobject JNICALL Java_org_jpy_PyLib_ensureGil
```
devinrsmith, line 2153: Add to `.h` file as well.

py/jpy/src/main/c/jni/org_jpy_PyLib.c
```diff
@@ -2145,6 +2145,21 @@ JNIEXPORT jboolean JNICALL Java_org_jpy_PyLib_hasGil
...
+  (JNIEnv* jenv, jclass jLibClass, jobject supplier)
+{
+    jobject result;
+    JPy_BEGIN_GIL_STATE
+    result = (*jenv)->CallObjectMethod(jenv, supplier, JPy_Supplier_get_MID);
+    JPy_END_GIL_STATE
+    return result;
```
devinrsmith, line 2160: Do we need exception checking / handling?

py/jpy/src/main/c/jpy_jmethod.c
```diff
@@ -326,48 +358,80 @@ PyObject* JMethod_InvokeMethod(JNIEnv* jenv, JPy_JMethod* method, PyObject* pyAr
...
+            Py_END_ALLOW_THREADS;
             JPy_ON_JAVA_EXCEPTION_GOTO(error);
             returnValue = JPy_FromJString(jenv, v);
             JPy_DELETE_LOCAL_REF(v);
         } else {
-            jobject v = (*jenv)->CallObjectMethodA(jenv, objectRef, method->mid, jArgs);
+            jobject v;
+            Py_BEGIN_ALLOW_THREADS;
+            v = (*jenv)->CallObjectMethodA(jenv, objectRef, method->mid, jArgs);
+            Py_END_ALLOW_THREADS;
```
devinrsmith, line 434: Do we think it's possible to wrap the BEGIN/END logic up in a macro?

```java
v = JPy_THREADED((*jenv)->CallObjectMethodA(jenv, objectRef, method->mid, jArgs));
```

py/jpy/src/main/c/jni/org_jpy_PyLib.c
niloc132: Probably. I'll look into it, see how to let it propagate. Off the top of my head, just plain java throwables need to be handled? No explicit allocations take place, so we don't need to trigger a java oom/etc, right?

py/jpy/src/main/c/jpy_jmethod.c
niloc132: I'm afraid my C is pretty weak, it would probably make more sense for me to punt to someone who knows it better to get it going more quickly.

py/jpy/src/main/c/jni/org_jpy_PyLib.c
niloc132: Is there a reason we don't package the generated .h files that the build emits?

py/jpy/src/main/c/jni/org_jpy_PyLib.c
niloc132: See the test added, it doesn't appear to be necessary. There is a JNI func to call from C to check if an exception is set in java, but it doesnt need to be checked if you don't want the exception addressed until you get back into java (at which time it is automatically handled).

py/jpy/src/main/java/org/jpy/PyObjectReferences.java
```diff
@@ -81,33 +81,30 @@ public int cleanupOnlyUseFromGIL() {
...
-            throw new IllegalStateException(
-                "We should only be calling PyObjectReferences.cleanupOnlyUseFromGIL if we have the GIL!");
-        }
-
-        int index = 0;
-        while (index < buffer.length) {
-            final Reference<? extends PyObject> reference = referenceQueue.poll();
-            if (reference == null) {
-                break;
+        return PyLib.ensureGil(() -> {
```
devinrsmith, line 84: Given this new construction, it likely makes sense to remove the Proxy-via-python than was previously enforcing the GIL; I'm happy to take lead on that part if you like.

niloc132: The proxy no-op may still serve a purpose, though it marginally more expensive than just writing more C, it is less complex to write - we're still not sure about the cost of making a new PyState per formula call. I have a followup patch I'm trying to benchmark using basically the same thing as the old proxy, but the runs I've done are very inconsistent., more than 25% differences in test runs. I'll sync up with you about that, it might well be a better project for you to consider.

py/jpy/src/main/c/jni/org_jpy_PyLib.c
devinrsmith: I don't know.

py/jpy/src/main/c/jpy_jmethod.c
devinrsmith: Same here. @cpwright thoughts? We could also create *methods* that wrap all JNI calls we make.

py/jpy-integration/src/pythonToJava/python/test/test_jpy.py
```diff
@@ -54,7 +54,7 @@ def test_has_jpy_classes(self):
 
   def test_has_gil(self):
     PyLib = jpy.get_type('org.jpy.PyLib')
-    self.assertTrue(PyLib.hasGil())
+    self.assertFalse(PyLib.hasGil())
```
devinrsmith, line 57: Change method name?

niloc132: Yes, we're now asserting that _in java_ when hasGil is called, java doesnt hold the GIL.

It's either that or delete the test, which seems to just test if a util method meant for java works from py (call from py, through c, into java, then turn around and go back into c, to ask the python C API something).

devinrsmith: My suggestion is we should change this to `test_java_doesnt_have_gil` or something like that.

devinrsmith (dismissed):

devinrsmith (approved):

PR #2321, opened 2022-04-22
Simplify the table listener interface

### LINKED ISSUE #2246: getColumn and getColumnSource for new python library tables
For the old python tables, equivalent functionality for `getColumn` and `getColumnSource` is not present in the new python tables. We should have equivalent methods for these in the new python table.

Example use cases for the old methods

```
timestamp = table.getColumnSource("Timestamp").get(0)
```

```
instrument_2nd = agg_by_instrument.getColumn("Instrument").get(1)
```

Current workaround involves using `table.j_object`

### LINKED ISSUE #2278: Python Listener wrappers talk about "ShiftObliviousListener" when they just mean "Listener"
We should be more accurate; shift-oblivious is a little confusing, anyway.

jmao-denver:
Fixes #2246
Fixes #2278

### DIFF
--- a/Integrations/src/main/java/io/deephaven/integrations/python/PythonListenerTableUpdateDataReader.java
+++ b/Integrations/src/main/java/io/deephaven/integrations/python/PythonListenerTableUpdateDataReader.java
@@ -1,0 +1,83 @@
+package io.deephaven.integrations.python;
+
+import io.deephaven.chunk.ChunkType;
+import io.deephaven.chunk.ResettableWritableChunk;
+import io.deephaven.chunk.attributes.Values;
+import io.deephaven.engine.rowset.RowSequence;
+import io.deephaven.engine.table.ChunkSource;
+import io.deephaven.engine.table.ColumnSource;
+import io.deephaven.engine.table.SharedContext;
+import io.deephaven.util.SafeCloseable;
+import org.jetbrains.annotations.NotNull;
+
+import java.lang.reflect.Array;
+import java.util.Arrays;
+
+public class PythonListenerTableUpdateDataReader {
+
+
+    public static io.deephaven.engine.table.Context makeContext(final int chunkCapacity,
+            @NotNull final ColumnSource<?>... columnSources) {
+        return new Context(chunkCapacity, columnSources);
+    }
+
+    private static class Context implements io.deephaven.engine.table.Context {
+
+        private final SharedContext sharedContext;
+        private final ChunkSource.FillContext[] fillContexts;
+        private final ResettableWritableChunk<Values>[] resettableChunks;
+
+        private Context(final int chunkCapacity, @NotNull final ColumnSource<?>... columnSources) {
+            sharedContext = SharedContext.makeSharedContext();
+            fillContexts = Arrays.stream(columnSources).map(cs -> cs.makeFillContext(chunkCapacity, sharedContext))
+                    .toArray(ChunkSource.FillContext[]::new);
+            // noinspection unchecked
+            resettableChunks =
+                    Arrays.stream(columnSources).map(cs -> cs.getChunkType().<Values>makeResettableWritableChunk())
+                            .toArray(ResettableWritableChunk[]::new);
+        }
+
+        @Override
+        public void close() {
+            sharedContext.close();
+            for (final SafeCloseable closeable : fillContexts) {
+                closeable.close();
+            }
+            for (final SafeCloseable closeable : resettableChunks) {
+                closeable.close();
+            }
+        }
+    }
+
+    public static Object[] readChunkColumnMajor(@NotNull final io.deephaven.engine.table.Context context,
+            final RowSequence rowSeq,
+            final ColumnSource<?>[] columnSources) {
+        final int nRows = rowSeq.intSize();
+        final int nCols = columnSources.length;
+        // final Object[] arrays = Arrays.stream(columnSources).map(cs ->
+        // cs.getChunkType().makeArray(rowSeq.intSize())).toArray(Object[]::new);
+        final Object[] arrays = Arrays.stream(columnSources).map(cs -> {
+            final ChunkType chunkType = cs.getChunkType();
+            if (chunkType == ChunkType.Object) {
+                return Array.newInstance(cs.getType(), rowSeq.intSize());
+            }
+            return cs.getChunkType().makeArray(rowSeq.intSize());
+        }).toArray(Object[]::new);
+        final Context typedContext = (Context) context;
+
+        for (int ci = 0; ci < nCols; ++ci) {
+            final ChunkSource.FillContext fillContext = typedContext.fillContexts[ci];
+            final ResettableWritableChunk<Values> chunk = typedContext.resettableChunks[ci];
+            final Object array = arrays[ci];
+            chunk.resetFromArray(array, 0, nRows);
+            final ColumnSource<?> colSrc = columnSources[ci];
+
+            // noinspection unchecked
+            colSrc.fillChunk(fillContext, chunk, rowSeq);
+            chunk.clear();
+        }
+        typedContext.sharedContext.reset();
+
+        return arrays;
+    }
+}
--- a/Integrations/src/main/java/io/deephaven/integrations/python/PythonListenerTableUpdateDataReader.java
+++ b/Integrations/src/main/java/io/deephaven/integrations/python/PythonListenerTableUpdateDataReader.java
@@ -1,28 +1,40 @@
 package io.deephaven.integrations.python;
 
-import io.deephaven.chunk.Chunk;
 import io.deephaven.chunk.ChunkType;
 import io.deephaven.chunk.ResettableWritableChunk;
-import io.deephaven.chunk.WritableChunk;
 import io.deephaven.chunk.attributes.Values;
 import io.deephaven.engine.rowset.RowSequence;
 import io.deephaven.engine.table.ChunkSource;
 import io.deephaven.engine.table.ColumnSource;
 import io.deephaven.engine.table.SharedContext;
+import io.deephaven.engine.table.TableUpdate;
 import io.deephaven.util.SafeCloseable;
 import org.jetbrains.annotations.NotNull;
 
 import java.lang.reflect.Array;
 import java.util.Arrays;
 
+/**
+ * An efficient reader for a Python table listener to extract columnar data based on the {@link RowSequence} in the
+ * {@link TableUpdate}
+ */
 public class PythonListenerTableUpdateDataReader {
 
-
+    /**
+     * Factory method for instance of {@link Context}
+     *
+     * @param chunkCapacity
+     * @param columnSources
+     * @return {@link Context}
+     */
     public static io.deephaven.engine.table.Context makeContext(final int chunkCapacity,
             @NotNull final ColumnSource<?>... columnSources) {
         return new Context(chunkCapacity, columnSources);
     }
 
+    /**
+     * Create a context that can be reused to read all the chunks of a row sequence
+     */
     private static class Context implements io.deephaven.engine.table.Context {
 
         private final SharedContext sharedContext;
@@ -51,6 +63,14 @@ public class PythonListenerTableUpdateDataReader {
         }
     }
 
+    /**
+     * Copy data from a table by chunks into a 2D array
+     *
+     * @param context the context used in filling the output array
+     * @param rowSeq indices of the rows of the table to put into the 2D array
+     * @param columnSources columns of data to put into the 2D array
+     * @return a 2D array
+     */
     public static Object[] readChunkColumnMajor(@NotNull final io.deephaven.engine.table.Context context,
             final RowSequence rowSeq,
             final ColumnSource<?>[] columnSources, final boolean prev) {
--- a/Integrations/src/main/java/io/deephaven/integrations/python/PythonListenerTableUpdateDataReader.java
+++ b/Integrations/src/main/java/io/deephaven/integrations/python/PythonListenerTableUpdateDataReader.java
@@ -1,7 +1,9 @@
 package io.deephaven.integrations.python;
 
+import io.deephaven.chunk.Chunk;
 import io.deephaven.chunk.ChunkType;
 import io.deephaven.chunk.ResettableWritableChunk;
+import io.deephaven.chunk.WritableChunk;
 import io.deephaven.chunk.attributes.Values;
 import io.deephaven.engine.rowset.RowSequence;
 import io.deephaven.engine.table.ChunkSource;
@@ -51,11 +53,9 @@ public class PythonListenerTableUpdateDataReader {
 
     public static Object[] readChunkColumnMajor(@NotNull final io.deephaven.engine.table.Context context,
             final RowSequence rowSeq,
-            final ColumnSource<?>[] columnSources) {
+            final ColumnSource<?>[] columnSources, final boolean prev) {
         final int nRows = rowSeq.intSize();
         final int nCols = columnSources.length;
-        // final Object[] arrays = Arrays.stream(columnSources).map(cs ->
-        // cs.getChunkType().makeArray(rowSeq.intSize())).toArray(Object[]::new);
         final Object[] arrays = Arrays.stream(columnSources).map(cs -> {
             final ChunkType chunkType = cs.getChunkType();
             if (chunkType == ChunkType.Object) {
@@ -73,7 +73,11 @@ public class PythonListenerTableUpdateDataReader {
             final ColumnSource<?> colSrc = columnSources[ci];
 
             // noinspection unchecked
-            colSrc.fillChunk(fillContext, chunk, rowSeq);
+            if (prev) {
+                colSrc.fillPrevChunk(fillContext, chunk, rowSeq);
+            } else {
+                colSrc.fillChunk(fillContext, chunk, rowSeq);
+            }
             chunk.clear();
         }
         typedContext.sharedContext.reset();
--- a/engine/api/src/main/java/io/deephaven/engine/table/ModifiedColumnSet.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/ModifiedColumnSet.java
@@ -8,6 +8,7 @@ import gnu.trove.map.hash.TObjectIntHashMap;
 
 import java.util.BitSet;
 import java.util.Map;
+import java.util.stream.IntStream;
 
 /**
  * Data structure to represent a subset of columns, used for propagating modifications via {@link TableUpdate}
@@ -430,6 +431,15 @@ public class ModifiedColumnSet {
      */
     public int numColumns() {
         return columns.size();
+    }
+
+    /**
+     * @return the names of the dirty columns in this set
+     */
+    public String[] dirtyColumnNames() {
+        return dirtyColumns.stream().mapToObj(ci -> columnNames[ci]).toArray(String[]::new);
+        // return IntStream.range(0, columnNames.length).filter(i -> dirtyColumns.get(i)).mapToObj(i ->
+        // columnNames[i]).toArray(String[]::new);
     }
 
     /**
--- a/engine/api/src/main/java/io/deephaven/engine/table/ModifiedColumnSet.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/ModifiedColumnSet.java
@@ -438,8 +438,6 @@ public class ModifiedColumnSet {
      */
     public String[] dirtyColumnNames() {
         return dirtyColumns.stream().mapToObj(ci -> columnNames[ci]).toArray(String[]::new);
-        // return IntStream.range(0, columnNames.length).filter(i -> dirtyColumns.get(i)).mapToObj(i ->
-        // columnNames[i]).toArray(String[]::new);
     }
 
     /**
--- a/engine/api/src/main/java/io/deephaven/engine/table/ModifiedColumnSet.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/ModifiedColumnSet.java
@@ -5,6 +5,7 @@ import gnu.trove.impl.Constants;
 import gnu.trove.map.hash.TObjectIntHashMap;
 import io.deephaven.base.verify.Assert;
 import io.deephaven.datastructures.util.CollectionUtil;
+import io.deephaven.util.type.ArrayTypeUtils;
 
 import java.util.BitSet;
 import java.util.Map;
@@ -437,7 +438,7 @@ public class ModifiedColumnSet {
      */
     public String[] dirtyColumnNames() {
         if (dirtyColumns == null)
-            return new String[0];
+            return ArrayTypeUtils.EMPTY_STRING_ARRAY;
         return dirtyColumns.stream().mapToObj(ci -> columnNames[ci]).toArray(String[]::new);
     }
 
--- a/engine/api/src/main/java/io/deephaven/engine/table/ModifiedColumnSet.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/ModifiedColumnSet.java
@@ -1,14 +1,13 @@
 package io.deephaven.engine.table;
 
-import io.deephaven.base.verify.Assert;
-import io.deephaven.datastructures.util.CollectionUtil;
 import com.google.common.collect.Iterators;
 import gnu.trove.impl.Constants;
 import gnu.trove.map.hash.TObjectIntHashMap;
+import io.deephaven.base.verify.Assert;
+import io.deephaven.datastructures.util.CollectionUtil;
 
 import java.util.BitSet;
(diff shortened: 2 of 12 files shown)


py/server/deephaven/table_listener.py
```diff
@@ -4,15 +4,113 @@
...
+    return col_defs
+
+
+class TableUpdate(JObjectWrapper):
+    j_object_type = _JTableUpdate
+
+    def __init__(self, table: Table, j_table_update):
+        self.table = table
+        self.j_table_update = j_table_update
+        self.chunk_size = DEFAULT_CHUNK_SIZE
```
chipkent, line 66: Does setting this as the chunk size make sense?  Does it ever make sense for the user to just say 100 or 1000 or whatever?  If so, then this would need to be an argument to the `added` and other methods.  I can imagine cases where the row set contains 1B entries, but the user code can only process 100 rows at a time without blowing up the memory usage.

py/server/deephaven/table_listener.py
```diff
@@ -4,15 +4,113 @@
...
+
+    @property
+    def j_object(self) -> jpy.JType:
+        return self.j_table_update
+
+    def added(self, cols: Union[str, List[str]] = None):
```
chipkent, line 72: Should there be a convenience argument to return everything as one numpy array, instead of a bunch of split up numpy arrays?

py/server/tests/test_table_listener.py
```diff
@@ -3,101 +3,162 @@
...
-                return (
-                    self.table_change_removed
-                    or self.table_change_modified
-                    or self.table_change_added
-                )
+                self.table_update = SimpleNamespace(added=None, removed=None, modified=None, modified_prev=None,
+                                                    Shifted=None)
+
+            def onUpdate(self, is_replay, update):
+                update.chunk_size = 4
```
chipkent, line 52: I see here that this is how you are intending to set the `chunk_size`.  If you are going this route, do you need to be carful that the chunk size isn't changed after data has started to be pulled?

chipkent: I'm also conflicted on this kind of API.  Ideally `update` is some sort of immutable that simply contains what has changed.  We are not that clean, because we are caching in the object.  Is there a way to get an API that maintains logical immutability while still providing a concise and clean user API?  The answer may be no, and we might have to accept some logical contamination to make it easy to use.

chipkent: I guess these API questions also come back to the "give me everything as a single array" question.  It raises other questions about if we always want the chunk sizes to be the same for different columns.

py/server/tests/test_table_listener.py
```diff
@@ -3,101 +3,162 @@
...
+                for chunk in update.removed():
+                    self.table_update.removed.append(chunk)
+
+                self.table_update.modified = []
+                for chunk in update.modified():
+                    self.table_update.modified.append(chunk)
+
+                self.table_update.modified_prev = []
+                for chunk in update.modified_prev():
+                    self.table_update.modified_prev.append(chunk)
```
chipkent, line 67: A user may want to look at differences between the modified and modified_prev values.  Since chunking and generators are being used, what would expect the design pattern to be like?  Something like below?

```
for m, mp in zip(update.modified(), update.modified_prev()):
    ... do stuff ...
```

py/server/deephaven/table_listener.py
```diff
@@ -87,48 +185,74 @@ def _do_locked(f: Callable, lock_type="shared") -> None:
...
 
-def _nargs_listener(listener) -> int:
-    """Returns the number of arguments the listener takes.
+class TableListener(ABC):
+    """An abstract table listener class that should be subclassed by any user Table listener class."""
 
-    Args:
-        listener: listener
+    @abstractmethod
+    def onUpdate(self, *args, **kwargs) -> None:
```
chipkent, line 192: Since we've thrown out most of the legacy listener types, I think the signature of the `onUpdate` method can probably be tightened up as well to something like:
```
def on_update(self, is_replay: bool, update: WhateverTheType) -> None:

**Review by chipkent — CHANGES_REQUESTED** [review]

File: py/server/deephaven/table_listener.py [hunk 6]
```diff
@@ -4,58 +4,243 @@
...
+            return next(
+                _changes_to_numpy(table=self.table, col_defs=col_defs, row_set=self.j_table_update.added.asRowSet(),
+                                  chunk_size=None))
+        except StopIteration:
+            return {}
+
+    def added_chunks(self, chunk_size: int, cols: Union[str, List[str]] = None) -> Generator[
+        Dict[str, numpy.ndarray], None, None]:
+        """Returns a generator that on each iteration, only returns a chunk of added rows in the form of a dict with
+        each key bing a column name and each value being a NumPy array of the rows in the chunk.
```
chipkent, line 97:
```suggestion
        """Returns a generator that, on each iteration, only returns a chunk of added rows in the form of a dict with
        each key being a column name and each value being a NumPy array of the rows in the chunk.
```

py/server/deephaven/table_listener.py
```diff
@@ -4,58 +4,243 @@
...
+            return next(
+                _changes_to_numpy(table=self.table, col_defs=col_defs, row_set=self.j_table_update.removed.asRowSet(),
+                                  chunk_size=None, prev=True))
+        except StopIteration:
+            return {}
+
+    def removed_chunks(self, chunk_size: int, cols: Union[str, List[str]] = None) -> Generator[
+        Dict[str, numpy.ndarray], None, None]:
+        """Returns a generator that on each iteration, only returns a chunk of removed rows in the form of a dict with
+        each key bing a column name and each value being a NumPy array of the rows in the chunk.
```
chipkent, line 137:
```suggestion
        """Returns a generator that, on each iteration, only returns a chunk of removed rows in the form of a dict with
        each key being a column name and each value being a NumPy array of the rows in the chunk.
```

py/server/deephaven/table_listener.py
```diff
@@ -4,58 +4,243 @@
...
-            raise RuntimeError("Attempting to deregister an unregistered listener..")
+        col_defs = _col_defs(table=self.table, cols=cols)
+        return _changes_to_numpy(table=self.table, col_defs=col_defs, row_set=self.j_table_update.removed.asRowSet(),
+                                 chunk_size=chunk_size, prev=True)
 
-        self.t.j_table.removeUpdateListener(self.listener)
-        self.isRegistered = False
+    def modified(self, cols: Union[str, List[str]] = None) -> Dict[str, numpy.ndarray]:
+        """Returns a dict with each key bing a column name and each value being a NumPy array of the current values of
+        all the modified rows in the columns.
```
chipkent, line 155:
```suggestion
        """Returns a dict with each key being a column name and each value being a NumPy array of the current values of
        all the modified rows in the columns.
```

py/server/deephaven/table_listener.py
```diff
@@ -4,58 +4,243 @@
...
+            return next(
+                _changes_to_numpy(self.table, col_defs=col_defs, row_set=self.j_table_update.modified.asRowSet(),
+                                  chunk_size=None))
+        except StopIteration:
+            return {}
+
+    def modified_chunks(self, chunk_size: int, cols: Union[str, List[str]] = None) -> Generator[
+        Dict[str, numpy.ndarray], None, None]:
+        """Returns a generator that on each iteration, only returns a chunk of modified rows in the form of a dict with
+        each key bing a column name and each value being a NumPy array of the current values of the rows in the chunk.
```
chipkent, line 177:
```suggestion
        """Returns a generator that, on each iteration, only returns a chunk of modified rows in the form of a dict with
        each key being a column name and each value being a NumPy array of the current values of the rows in the chunk.
```

py/server/deephaven/table_listener.py
```diff
@@ -4,58 +4,243 @@
...
+        if not self.j_table_update.modified:
+            return (_ for _ in ())
+
+        col_defs = _col_defs(table=self.table, cols=cols)
+        return _changes_to_numpy(self.table, col_defs=col_defs, row_set=self.j_table_update.modified.asRowSet(),
+                                 chunk_size=chunk_size)
+
+    def modified_prev(self, cols: Union[str, List[str]] = None) -> Dict[str, numpy.ndarray]:
+        """Returns a dict with each key bing a column name and each value being a NumPy array of the previous values of
+        all the modified rows in the columns.
```
chipkent, line 195:
```suggestion
        """Returns a dict with each key being a column name and each value being a NumPy array of the previous values of
        all the modified rows in the columns.
```

py/server/deephaven/table_listener.py
```diff
@@ -4,58 +4,243 @@
...
+        Args:
+             cols (Union[str, List[str]): the column(s) for which to return the added rows
+
+        Returns:
+            a dict
+        """
+        if not self.j_table_update.modified:
+            return {}
+
+        col_defs = _col_defs(table=self.table, cols=cols)
```
chipkent, line 206: This comment applies everywhere, but I'm just noticing this... should the `_col_defs` function get moved into `_changes_to_numpy`, since that would reduce scope and code complexity?

py/server/deephaven/table_listener.py
```diff
@@ -4,58 +4,243 @@
...
+            return next(
+                _changes_to_numpy(self.table, col_defs=col_defs, row_set=self.j_table_update.modified.asRowSet(),
+                                  chunk_size=None, prev=True))
+        except StopIteration:
+            return {}
+
+    def modified_prev_chunks(self, chunk_size: int, cols: Union[str, List[str]] = None) -> Generator[
+        Dict[str, numpy.ndarray], None, None]:
+        """Returns a generator that on each iteration, only returns a chunk of modified rows in the form of a dict with
+        each key bing a column name and each value being a NumPy array of the previous values of the rows in the chunk.
```
chipkent, line 217:
```suggestion
        """Returns a generator that, on each iteration, only returns a chunk of modified rows in the form of a dict with
        each key being a column name and each value being a NumPy array of the previous values of the rows in the chunk.
```

py/server/deephaven/table_listener.py
```diff
@@ -4,58 +4,243 @@
...
+        if not self.j_table_update.modified:
+            return (_ for _ in ())
+
+        col_defs = _col_defs(table=self.table, cols=cols)
+        return _changes_to_numpy(self.table, col_defs=col_defs, row_set=self.j_table_update.modified.asRowSet(),
+                                 chunk_size=chunk_size, prev=True)
+
+    @property
+    def shifted(self):
+        # return self.j_table_update.shifted
```
chipkent, line 235: rm?

py/server/deephaven/table_listener.py
```diff
@@ -4,58 +4,243 @@
...
+        raise NotImplemented("shifts are not supported yet.")
+
+    @property
+    def modified_columns(self) -> List[str]:
+        """The list of dirty columns in this update."""
```
chipkent, line 240:
```suggestion
        """The list of modified columns in this update."""
```

py/server/deephaven/table_listener.py
```diff
@@ -72,208 +257,164 @@ def _do_locked(f: Callable, lock_type="shared") -> None:
...
         ValueError
     """
-    throwing_runnable = jpy.get_type(
-        "io.deephaven.integrations.python.PythonThrowingRunnable"
-    )
-    update_graph_processor = jpy.get_type(
-        "io.deephaven.engine.updategraph.UpdateGraphProcessor"
-    )
+    throwing_runnable = jpy.get_type("io.deephaven.integrations.python.PythonThrowingRunnable")
+    update_graph_processor = jpy.get_type("io.deephaven.engine.updategraph.UpdateGraphProcessor")
```
chipkent, line 261: Do you want these here or with the other java types?
Do you want these names or the _J names?

I don't have a strong view.  Just a question of consistency.

py/server/deephaven/table_listener.py
```diff
@@ -72,208 +257,164 @@ def _do_locked(f: Callable, lock_type="shared") -> None:
...
 
     if lock_type == "exclusive":
         update_graph_processor.DEFAULT.exclusiveLock().doLocked(throwing_runnable(f))
     elif lock_type == "shared":
+        # slock = update_graph_processor.DEFAULT.sharedLock()
+        # try:
+        #     slock.lock()
+        #     f()
+        # finally:
+        #     slock.unlock()
```
chipkent, line 271: rm

py/server/deephaven/table_listener.py
```diff
@@ -72,208 +257,164 @@ def _do_locked(f: Callable, lock_type="shared") -> None:
...
-    Returns:
-        the number of arguments the listener takes
+class TableListener(ABC):
+    """An abstract table listener class that should be subclassed by any user table listener class."""
 
-    Raises:
-        ValueError, NotImplementedError
-    """
+    @abstractmethod
+    def on_update(self, update: TableUpdate, is_replay: bool) -> None:
```
chipkent, line 281: pydoc?

py/server/deephaven/table_listener.py
```diff
@@ -72,208 +257,164 @@ def _do_locked(f: Callable, lock_type="shared") -> None:
...
-        f = listener
-    elif hasattr(listener, "onUpdate"):
-        f = listener.onUpdate
-    else:
-        raise ValueError(
-            "ShiftObliviousListener is neither callable nor has an 'onUpdate' method"
-        )
 
-    return len(signature(f).parameters)
+def listener_wrapper(table: Table):
```
chipkent, line 285: Pydoc?
Public?

py/server/deephaven/table_listener.py
```diff
@@ -72,208 +257,164 @@ def _do_locked(f: Callable, lock_type="shared") -> None:
...
+        @wraps(listener)
+        def wrapper(update, *args):
+            sig = inspect.signature(listener)
+            n_params = len(sig.parameters)
+            t_update = TableUpdate(table=table, j_table_update=update)
 
+            if n_params == 2:
+                listener(t_update, args[0])
+            else:
+                listener(t_update)
```
chipkent, line 296: This looks problematic.
1) Is the 1 arg supported?
2) What if there are 3 or more args?  Shouldn't it raise an error?

py/server/deephaven/table_listener.py
```diff
@@ -72,208 +257,164 @@ def _do_locked(f: Callable, lock_type="shared") -> None:
...
-    listener: Union[Callable, Type[TableListener]],
-    description: str = None,
-    retain: bool = True,
-    listener_type: str = "auto",
-    start_listening: bool = True,
-    replay_initial: bool = False,
-    lock_type: str = "shared",
-) -> TableListenerHandle:
-    """Listen to table changes.
+def _wrap_listener_func(t: Table, description: str, listener: Callable):
```
chipkent, line 303: pydoc

py/server/deephaven/table_listener.py
chipkent: I see that functions using this are testing the 2 arg case.  The logic is spread out in multiple places.  If the logic was consolidated, it would be more simple.

py/server/deephaven/table_listener.py
```diff
@@ -72,208 +257,164 @@ def _do_locked(f: Callable, lock_type="shared") -> None:
...
-    Table change events are processed by 'listener', which can be either
-        (1) a callable (e.g. function) or
-        (2) an object that provides an "onUpdate" method.
-    In either case, the method must have one of the following signatures.
 
-    * (added, removed, modified): shift_oblivious
-    * (isReplay, added, removed, modified): shift_oblivious + replay
-    * (update): shift-aware
-    * (isReplay, update): shift-aware + replay
+def _wrap_listener_obj(t: Table, description: str, listener: TableListener):
```
chipkent, line 310: pydoc.
See comment above about simplification.

py/server/deephaven/table_listener.py
```diff
@@ -72,208 +257,164 @@ def _do_locked(f: Callable, lock_type="shared") -> None:
...
+        raise ValueError(f"The on_update method must have 2 (update, is_replay) parameters.")
+    listener.on_update = listener_wrapper(table=t)(listener.on_update)
+    return listener
 
-    For shift-oblivious listeners, 'added', 'removed', and 'modified' are the indices of the rows which changed.
-    For shift-aware listeners, 'update' is an object that describes the table update.
 
-    Listeners that support replaying the initial table snapshot have an additional parameter, 'isReplay', which is
-    true when replaying the initial snapshot and false during normal updates.
+def listen(t: Table, listener: Union[Callable, TableListener], description: str = None, do_replay: bool = False,
```
chipkent, line 318: See my earlier comment about the API confusion I had about if the listener was started or not.
... but you very clearly document the behavior, so I may be an idiot.  What is here may be ok.  I'm leaving my thoughts so that you can think about it.

py/server/deephaven/table_listener.py
```diff
@@ -72,208 +257,164 @@ def _do_locked(f: Callable, lock_type="shared") -> None:
...
+           replay_lock: str = "shared"):
+    """This is a convenience function that creates a TableListenerHandle object and immediately starts it to listen
+    for table updates.
 
-    See the Deephaven listener documentation for details on processing update events.  This documentation covers the
-    details of the added, removed, and modified row sets; row shift information; as well as the details of how to apply
-    the update object.  It also has examples on how to access current and previous-tick table values.
+    The function returns the created TableListenerHandle object whose 'stop' method can be called to stop listening.
+    If it goes out of scope and is garbage collected,
```
chipkent, line 324: incomplete fragment

py/server/deephaven/table_listener.py
```diff
@@ -72,208 +257,164 @@ def _do_locked(f: Callable, lock_type="shared") -> None:
...
-                raise ValueError(
-                    "Shift-aware listener must take 1 (update) or 2 (isReplay, update) arguments."
-                )
 
-        else:
-            raise ValueError(
-                "Unsupported listener type: ltype={}".format(listener_type)
-            )
+class TableListenerHandle:
+    """A handle for a table listener to manage its lifecycle."""
```
chipkent, line 346:
```suggestion
    """A handle to manage a table listener's lifecycle."""
```

py/server/deephaven/table_listener.py
```diff
@@ -72,208 +257,164 @@ def _do_locked(f: Callable, lock_type="shared") -> None:
...
+        (1) a callable (e.g. function) or
+        (2) an instance of TableListener type which provides an "on_update" method.
 
-            if start_listening:
-                handle.register()
+        The callable or the on_update method must have one of the following two signatures.
+        * (update: TableUpdate, is_replay: bool): support replaying the initial table snapshot and normal table updates
+        The 'update' parameter is an object that describes the table update;
+        The 'is_replay' parameter is used only by replay listeners, it is set to 'true' when replaying the initial
+            snapshot and 'false' during normal updates.
```
chipkent, line 359: I think this is inaccurate given the current code.  E.g. only one signature is supported.

py/server/deephaven/table_listener.py
```diff
@@ -72,208 +257,164 @@ def _do_locked(f: Callable, lock_type="shared") -> None:
...
+        """
+        self.t = t
+
+        if callable(listener):
+            listener_wrapped = _wrap_listener_func(t, description, listener)
+        elif isinstance(listener, TableListener):
+            listener_wrapped = _wrap_listener_obj(t, description, listener)
         else:
-            start()
+            raise ValueError("listener is neither callable nor TableListener nor TableReplayListener object")
```
chipkent, line 377: TableReplayListener no longer exists.

py/server/deephaven/table_listener.py
```diff
@@ -72,208 +257,164 @@ def _do_locked(f: Callable, lock_type="shared") -> None:
...
-    except Exception as e:
-        raise DHError(e, "failed to listen to the table.") from e
+        self.started = False
+
+    def start(self, do_replay: bool = False, replay_lock: str = "shared") -> None:
+        """Start the listener by registering it with the table and listening for updates.
+
+        lock_type (str): UGP lock type, used when replay_initial=True, valid values are "exclusive" and "shared".
+            Default is "shared".
+
```
chipkent, line 387:
```suggestion
```
needed?

chipkent (changes requested):

engine/api/src/main/java/io/deephaven/engine/table/ModifiedColumnSet.java
```diff
@@ -432,6 +433,13 @@ public int numColumns() {
         return columns.size();
     }
 
+    /**
+     * @return the names of the dirty columns in this set
+     */
+    public String[] dirtyColumnNames() {
+        return dirtyColumns.stream().mapToObj(ci -> columnNames[ci]).toArray(String[]::new);
+    }
+
```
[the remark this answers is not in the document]
jmao-denver: It is covered in the Python unit tests, so sure we need a Java case.

py/server/tests/test_table_listener.py
```diff
@@ -3,102 +3,228 @@
...
+            t_update.modified = []
+            t_update.modified_prev = []
+            t_update.added.append(update.added())
+            t_update.removed.append(update.removed())
+            t_update.modified.append(update.modified())
+            t_update.modified_prev.append(update.modified_prev())
+            replay_recorder.append(is_replay)
+
+        table_listener_handle = TableListenerHandle(self.test_table, listener_func)
+        table_listener_handle.start(do_replay=True)
```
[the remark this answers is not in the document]
jmao-denver: listen() is provided as a convenience function but I do see your point about confusing the users and if we want to promote the explicit life cycle management of the listeners. That said I also see the usefulness of listen(). Maybe put it up for a vote by the DevRel team?

py/server/tests/test_table_listener.py
```diff
@@ -3,102 +3,228 @@
...
+        time.sleep(1)
+        self.assertGreaterEqual(len(t_update.added), 1)
+        self.check_result(t_update.added)
+        self.check_result(t_update.removed)
+        self.check_result(t_update.modified)
+        self.check_result(t_update.modified_prev)
+        self.assertFalse(all(replay_recorder))
```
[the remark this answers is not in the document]
jmao-denver: Not sure about the DRY comment. We want to check all the changes captured.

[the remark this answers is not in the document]
chipkent: I agree that you want to check everything.  I put the DRY comment because the code could be structured in a way that the same code doesn't appear multiple times.  E.g. a class to keep track of values and assert everything, with a common API to update values.  As the code morphs, we currently have to hope that the next guy remembers to make changes in all of the relevant places.  As the code currently is, some values are still not being checked in all unit tests (e.g. modified columns).  If we at least tested all of the values for all of the cases, it would probably be ok.

chipkent (dismissed):

Integrations/src/main/java/io/deephaven/integrations/python/PythonListenerTableUpdateDataReader.java
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Copy data from a table by chunks into a 2D array
+     *
+     * @param context the context used in filling the output array
+     * @param rowSeq indices of the rows of the table to put into the 2D array
+     * @param columnSources columns of data to put into the 2D array
+     * @return a 2D array
+     */
+    public static Object[] readChunkColumnMajor(@NotNull final io.deephaven.engine.table.Context context,
```
jcferretti, line 74: Cosmetic comment: can we split each argument to its own line here?

engine/api/src/main/java/io/deephaven/engine/table/ModifiedColumnSet.java
```diff
@@ -432,6 +432,15 @@ public int numColumns() {
         return columns.size();
     }
 
+    /**
+     * @return the names of the dirty columns in this set
+     */
+    public String[] dirtyColumnNames() {
+        if (dirtyColumns == null)
+            return new String[0];
```
jcferretti, line 440: This could be `ArrayTypeUtils.EMPTY_STRING_ARRAY` to avoid creating a new empty array every time.

jcferretti (dismissed):
I had a short discussion with Ryan about the code in `TableUpdateDataReader.close` and whether caring about a single close subcall in there throwing and us not trying to close anything after that.

The short of the conversation is that this code follows a general pattern in our close implementations in the engine; there is an argument for simplicity versus an argument for correctness (strictly speaking we can leak, in practice we have a convention that means we don't care at this point).

If we wanted to do this differently moving on, that would require revisiting the implementations of close() in many places, not just this.  So doing anything in this PR is not appropriate, so leaving alone for now; if we ever decide otherwise we need to file a ticket (GitHub issue) and work towards it.

jcferretti (approved):

jcferretti (dismissed):

jcferretti (approved):

PR #2331, opened 2022-04-22
Static Select must flatten when using intermediate columns

### LINKED ISSUE #2308: Query engine: select on a cross join causes AIOOBE: "arraycopy: last source index 1000424 out of bounds for int[1000000]"
RuntimeError: java.lang.ArrayIndexOutOfBoundsException: arraycopy: last source index 1000424 out of bounds for int[1000000] 

Steps to reproduce:
In python, create the following tables - note that python may not be required
```
# First create a 1000 row table (100 does not work)
t = empty_table(1000).update("I=i")

# Rename columns and cross join with itself
t2 = t.view("A=I").join(table=t.view("B=I"), on=[])

# Select an original column, _then_ an expression
t3 = t2.select(["A", "C = A + B"])
```

Expected, a table with two columns, actual results, an error occurs.
```
copyFromTypedArray:128, WritableObjectChunk (io.deephaven.chunk)
lambda$fillChunkByRanges$0:107, ImmutableObjectArraySource (io.deephaven.engine.table.impl.sources.immutable)
accept:-1, 922147515 (io.deephaven.engine.table.impl.sources.immutable.ImmutableObjectArraySource$$Lambda$1128)
lambda$forAllRowKeyRanges$1:187, RowSequence (io.deephaven.engine.rowset)
accept:-1, 1990783618 (io.deephaven.engine.rowset.RowSequence$$Lambda$585)
forEachLongRangeInSpanWithOffsetAndMaxCardinality:3899, RspArray (io.deephaven.engine.rowset.impl.rsp)
forEachRowKeyRange:226, RspRowSequence (io.deephaven.engine.rowset.impl.rsp)
forAllRowKeyRanges:186, RowSequence (io.deephaven.engine.rowset)
fillChunkByRanges:105, ImmutableObjectArraySource (io.deephaven.engine.table.impl.sources.immutable)
fillChunk:96, ImmutableObjectArraySource (io.deephaven.engine.table.impl.sources.immutable)
getChunkByFilling:45, DefaultChunkSource (io.deephaven.engine.table.impl)
getChunk:32, DefaultChunkSource (io.deephaven.engine.table.impl)
getChunk:127, ImmutableObjectArraySource (io.deephaven.engine.table.impl.sources.immutable)
fillChunk:132, Formula (io.deephaven.temp.c1954111451987228486v55_0)
fillChunk:255, ViewColumnSource (io.deephaven.engine.table.impl.sources)
doApplyUpdate:287, SelectColumnLayer (io.deephaven.engine.table.impl.select.analyzers)
doSerialApplyUpdate:171, SelectColumnLayer (io.deephaven.engine.table.impl.select.analyzers)
lambda$onAllRequiredColumnsCompleted$1:134, SelectColumnLayer$1 (io.deephaven.engine.table.impl.select.analyzers)
run:-1, 771826563 (io.deephaven.engine.table.impl.select.analyzers.SelectColumnLayer$1$$Lambda$1125)
submit:573, SelectAndViewAnalyzer$ImmediateJobScheduler (io.deephaven.engine.table.impl.select.analyzers)
onAllRequiredColumnsCompleted:133, SelectColumnLayer$1 (io.deephaven.engine.table.impl.select.analyzers)
onLayerCompleted:433, SelectAndViewAnalyzer$SelectLayerCompletionHandler (io.deephaven.engine.table.impl.select.analyzers)
doSerialApplyUpdate:172, SelectColumnLayer (io.deephaven.engine.table.impl.select.analyzers)
lambda$onAllRequiredColumnsCompleted$1:134, SelectColumnLayer$1 (io.deephaven.engine.table.impl.select.analyzers)
run:-1, 771826563 (io.deephaven.engine.table.impl.select.analyzers.SelectColumnLayer$1$$Lambda$1125)
submit:573, SelectAndViewAnalyzer$ImmediateJobScheduler (io.deephaven.engine.table.impl.select.analyzers)
onAllRequiredColumnsCompleted:133, SelectColumnLayer$1 (io.deephaven.engine.table.impl.select.analyzers)
onLayerCompleted:433, SelectAndViewAnalyzer$SelectLayerCompletionHandler (io.deephaven.engine.table.impl.select.analyzers)
applyUpdate:70, BaseLayer (io.deephaven.engine.table.impl.select.analyzers)
applyUpdate:85, SelectColumnLayer (io.deephaven.engine.table.impl.select.analyzers)
applyUpdate:85, SelectColumnLayer (io.deephaven.engine.table.impl.select.analyzers)
lambda$selectOrUpdate$37:1281, QueryTable (io.deephaven.engine.table.impl)
get:-1, 168093566 (io.deephaven.engine.table.impl.QueryTable$$Lambda$1033)
withNugget:519, QueryPerformanceRecorder (io.deephaven.engine.table.impl.perf)
lambda$selectOrUpdate$38:1242, QueryTable (io.deephaven.engine.table.impl)
get:-1, 262983997 (io.deephaven.engine.table.impl.QueryTable$$Lambda$1032)
memoizeResult:3037, QueryTable (io.deephaven.engine.table.impl)
selectOrUpdate:1241, QueryTable (io.deephaven.engine.table.impl)
selectInternal:1186, QueryTable (io.deephaven.engine.table.impl)
select:1177, QueryTable (io.deephaven.engine.table.impl)
select:254, TableWithDefaults (io.deephaven.engine.table.impl)
```
screenshot781

Workarounds, change order of columns, or try update+dropColumns instead of select().

nbauernfeind:
Fixes #2308

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/SelectAndViewAnalyzer.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/SelectAndViewAnalyzer.java
@@ -3,12 +3,15 @@ package io.deephaven.engine.table.impl.select.analyzers;
 import io.deephaven.base.log.LogOutput;
 import io.deephaven.base.log.LogOutputAppendable;
 import io.deephaven.datastructures.util.CollectionUtil;
+import io.deephaven.engine.rowset.RowSetFactory;
 import io.deephaven.engine.table.ColumnSource;
 import io.deephaven.engine.table.ColumnDefinition;
 import io.deephaven.engine.table.TableUpdate;
 import io.deephaven.engine.table.WritableColumnSource;
 import io.deephaven.engine.table.impl.OperationInitializationThreadPool;
+import io.deephaven.engine.table.impl.QueryTable;
 import io.deephaven.engine.table.impl.perf.BasePerformanceEntry;
+import io.deephaven.engine.table.impl.util.WrappedRowSetWritableRowRedirection;
 import io.deephaven.engine.table.impl.util.WritableRowRedirection;
 import io.deephaven.engine.updategraph.AbstractNotification;
 import io.deephaven.engine.updategraph.UpdateGraphProcessor;
@@ -42,7 +45,7 @@ public abstract class SelectAndViewAnalyzer implements LogOutputAppendable {
     }
 
     public static SelectAndViewAnalyzer create(final Mode mode, final Map<String, ColumnSource<?>> columnSources,
-            final TrackingRowSet rowSet, final ModifiedColumnSet parentMcs, final boolean publishTheseSources,
+            TrackingRowSet rowSet, final ModifiedColumnSet parentMcs, final boolean publishTheseSources,
             final boolean allowInternalFlatten,
             final SelectColumn... selectColumns) {
         SelectAndViewAnalyzer analyzer = createBaseLayer(columnSources, publishTheseSources);
@@ -55,7 +58,8 @@ public abstract class SelectAndViewAnalyzer implements LogOutputAppendable {
             rowRedirection = null;
         }
 
-        final boolean flatResult = rowSet.isFlat();
+        final TrackingRowSet originalRowSet = rowSet;
+        boolean flatResult = rowSet.isFlat();
         // if we preserve a column, we set this to false
         boolean flattenedResult = !flatResult
                 && allowInternalFlatten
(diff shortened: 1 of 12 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/SelectAndViewAnalyzer.java
```diff
@@ -55,32 +58,50 @@ public static SelectAndViewAnalyzer create(final Mode mode, final Map<String, Co
...
+
             final Stream<String> allDependencies =
                     Stream.concat(sc.getColumns().stream(), sc.getColumnArrays().stream());
             final String[] distinctDeps = allDependencies.distinct().toArray(String[]::new);
             final ModifiedColumnSet mcsBuilder = new ModifiedColumnSet(parentMcs);
 
             if (shouldPreserve(sc)) {
-                analyzer =
-                        analyzer.createLayerForPreserve(sc.getName(), sc, sc.getDataView(), distinctDeps, mcsBuilder);
                 if (numberOfInternallyFlattenedColumns > 0) {
```
cpwright, line 97: How are we updating numberOfInternallyFlattenedColumns on line 81 so that we will bail out of the flattening to allow us to preserve other in-memory column sources?

engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/StaticFlattenLayer.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    void setBaseBits(BitSet bitset) {
+        inner.setBaseBits(bitset);
+    }
+
+    @Override
+    void populateModifiedColumnSetRecurse(ModifiedColumnSet mcsBuilder, Set<String> remainingDepsToSatisfy) {
+        inner.populateModifiedColumnSetRecurse(mcsBuilder, remainingDepsToSatisfy);
```
cpwright, line 57: Instead of depending just on the lower layer, it seems like we should be depending on this layer as well.  Maybe that is unnecessary, because once the inner is done there is no further work to do here ... but if this actually every did any work we would have incorrect dependencies.  It also seems "bad" for this notification to potentially fire after every other part of the select is done and we proceed with the operation.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/SelectAndViewAnalyzer.java
cpwright: I get it now.  We had to have at least one result column that was previously flattened to get into the case on line 81.  We'll flatten out the rest of our internal columns at that point.  If we end up hitting shouldPreserve later, then having one is enough to restart the analysis.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/StaticFlattenLayer.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @Override
+    Map<String, ColumnSource<?>> getColumnSourcesRecurse(GetMode mode) {
+        final Map<String, ColumnSource<?>> innerColumns = inner.getColumnSourcesRecurse(mode);
+
+        if (overriddenColumns.keySet().stream().noneMatch(innerColumns::containsKey)) {
```
cpwright, line 64: How would this be possible, why would the layer exist at all if you had nothing overridden?

nbauernfeind: This can happen when `mode == GetMode.NEW`.

cpwright (approved):

PR #2333, opened 2022-04-23
Add docstrings for predefined DType instances

### LINKED ISSUE #2325: Server Pydocs do not show dtypes
The server Pydocs do not show the `dtypes`, such as `dtypes.DateTime`.

See:  https://deephaven.io/core/pydoc/code/deephaven.dtypes.html

jmao-denver:
Fixes #2325

### DIFF
--- a/py/server/deephaven/dtypes.py
+++ b/py/server/deephaven/dtypes.py
@@ -105,15 +105,29 @@ JObject = DType(j_name="java.lang.Object")
 """Java Object type"""
 byte_array = DType(j_name='[B')
 """Byte array type"""
+int8_array = byte_array
+"""Byte array type"""
 short_array = DType(j_name='[S')
 """Short array type"""
-int_array = DType(j_name='[I')
+int16_array = short_array
+"""Short array type"""
+int32_array = DType(j_name='[I')
 """32bit integer array type"""
 long_array = DType(j_name='[J')
 """64bit integer array type"""
-float_array = DType(j_name='[S')
+int64_array = long_array
+"""64bit integer array type"""
+int_array = long_array
+"""64bit integer array type"""
+single_array = DType(j_name='[S')
+"""Single-precision floating-point array type"""
+float32_array = single_array
 """Single-precision floating-point array type"""
 double_array = DType(j_name='[D')
+"""Double-precision floating-point array type"""
+float64_array = double_array
+"""Double-precision floating-point array type"""
+float_array = double_array
 """Double-precision floating-point array type"""
 string_array = DType(j_name='[Ljava.lang.String;')
 """Java String array type"""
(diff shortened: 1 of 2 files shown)


py/server/deephaven/dtypes.py
```diff
@@ -60,34 +60,65 @@ def __call__(self, *args, **kwargs):
...
 BigDecimal = DType(j_name="java.math.BigDecimal")
+"""Java BigDecimal type"""
 StringSet = DType(j_name="io.deephaven.stringset.StringSet")
+"""Deephaven StringSet type"""
 DateTime = DType(j_name="io.deephaven.time.DateTime", np_type=np.dtype("datetime64[ns]"))
+"""Deephaven DateTime type"""
 Period = DType(j_name="io.deephaven.time.Period")
+"""Deephaven time period type"""
 PyObject = DType(j_name="org.jpy.PyObject")
+"""JPY PyObject type"""
```
chipkent, line 103: Should this just be:
```suggestion
"""Python object type"""
```

py/server/deephaven/dtypes.py
```diff
@@ -60,34 +60,65 @@ def __call__(self, *args, **kwargs):
...
 JObject = DType(j_name="java.lang.Object")
+"""JPY Java Object type"""
```
chipkent, line 105: Should this just be:
```suggestion
"""Java Object type"""
```

py/server/deephaven/dtypes.py
```diff
@@ -60,34 +60,65 @@ def __call__(self, *args, **kwargs):
...
 int_array = DType(j_name='[I')
+"""32bit integer array type"""
 long_array = DType(j_name='[J')
+"""64bit integer array type"""
 float_array = DType(j_name='[S')
+"""Single-precision floating-point array type"""
 double_array = DType(j_name='[D')
+"""Double-precision floating-point array type"""
 string_array = DType(j_name='[Ljava.lang.String;')
+"""Java String array type"""
```
chipkent, line 119: Does this need to be a "Java" string?

py/server/deephaven/dtypes.py
```diff
@@ -60,34 +60,65 @@ def __call__(self, *args, **kwargs):
...
+"""Deephaven time period type"""
 PyObject = DType(j_name="org.jpy.PyObject")
+"""JPY PyObject type"""
 JObject = DType(j_name="java.lang.Object")
+"""JPY Java Object type"""
 byte_array = DType(j_name='[B')
+"""Byte array type"""
 short_array = DType(j_name='[S')
+"""Short array type"""
 int_array = DType(j_name='[I')
```
chipkent, line 110: 1) Note that `int` here is inconsistent with `int_` above.  
2) Should the other type aliases be included here so that there is consistency? (e.g. `int8_array`)

`int_` above is 64b, but here, it is 32b.

chipkent (changes requested):

py/server/deephaven/dtypes.py
jmao-denver: yes, Java's array of strings is different from numpy's and Python natively doesn't have array of strings.

PR #2336, opened 2022-04-25
Wrap the Filter interface

### LINKED ISSUE #2335: wrap new Filter API

jmao-denver:
Fixes #2335

### DIFF
--- a/py/server/deephaven/filters.py
+++ b/py/server/deephaven/filters.py
@@ -2,22 +2,89 @@
 #   Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
 #
 """ This module implement various filters that can be used in deephaven table's filter operations."""
+from __future__ import annotations
+
+from typing import List, Union
 
 import jpy
 
 from deephaven import DHError
 from deephaven._wrapper import JObjectWrapper
+from deephaven.jcompat import to_sequence
 
 _JFilter = jpy.get_type("io.deephaven.api.filter.Filter")
 _JRegexFilter = jpy.get_type("io.deephaven.engine.table.impl.select.RegexFilter")
+_JFilterOr = jpy.get_type("io.deephaven.api.filter.FilterOr")
+_JFilterAnd = jpy.get_type("io.deephaven.api.filter.FilterAnd")
+_JFilterNot = jpy.get_type("io.deephaven.api.filter.FilterNot")
 
 
 class Filter(JObjectWrapper):
+    """A Filter object represents a filter that can be used in Table's filtering(where) operations."""
     j_object_type = _JFilter
 
     @property
     def j_object(self) -> jpy.JType:
         return self.j_filter
+
+    def __init__(self, j_filter):
+        self.j_filter = j_filter
+
+    @classmethod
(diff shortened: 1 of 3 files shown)


py/server/deephaven/filters.py
```diff
@@ -2,23 +2,90 @@
...
+    Args:
+        filters (List[filter]): the component filters
+
+    Returns:
+        a new Filter
+    """
+    return Filter(j_filter=_JFilterAnd.of(*[f.j_filter for f in filters]))
+
+
+def not_(filter_: Filter) -> Filter:
```
devinrsmith, line 78: Is this more pythonic, or is `my_filter.not()`? The java interface has `not()` as part of the interface.

py/server/deephaven/filters.py
```diff
@@ -2,23 +2,90 @@
...
+        """
+        conditions = to_sequence(conditions)
+        try:
+            filters = [cls(j_filter=j_filter) for j_filter in _JFilter.from_(conditions).toArray()]
+            return filters if len(filters) != 1 else filters[0]
+        except Exception as e:
+            raise DHError(e, "failed to create filters.") from e
+
+
+def or_(filters: List[Filter]) -> Filter:
```
devinrsmith, line 54: Looking at this, and unit tests, it looks like we have:

```python
my_table.where(or_(Filter.from_(["a > 100", "b < 1000", "c < 0"])))
```

I'm wondering if we don't want something like:

```python
my_table.where(or_(["a > 100", "b < 1000", "c < 0"]))
```

And, it might not just be strings, but filters themselves:

```python
f1 = and_(["a > 100", "b <1000"])
f2 = and_(["c > 1", "d < 1"])
f = or_([f1, f2, "z>42"])
my_table.where(f)
```

py/server/deephaven/filters.py
jmao-denver: 'not' is a Python keyword, can't be used as an identifier

py/server/deephaven/filters.py
jmao-denver: Not sure we want to allow mixing Filter objects with conditional strings. Arguably it might be more confusing than helpful.

py/server/deephaven/filters.py
devinrsmith: We *could* do `my_filter.not_()` (we wouldn't need to provide it in java since we can do it in the wrapper) if desired.

devinrsmith: Or, as property: `my_filter.not_`

py/server/deephaven/filters.py
devinrsmith: I'm fine w/ the conclusion as long as we are happy.
```python
f1 = and_(Filter.from_(["a > 100", "b <1000"]))
f2 = and_(Filter.from_(["c > 1", "d < 1"]))
f = or_([f1, f2, Filter.from_("z>42")])
my_table.where(f)
```

devinrsmith (approved):
I'm "ok" w/ as-is; unless we think there are usability concerns wrt creating filters.

PR #2337, opened 2022-04-25
Refactor ScriptSession injection

niloc132:
The current system of requiring that the consumer request the entire map and already know the session type (or read it from the singleton not-@Singleton configuration instance) is pretty onerous for downstream apps, and somewhat moot when we know that there is only one possible scriptsession at a time anyway.

Partial #2221


engine/table/src/main/java/io/deephaven/engine/util/PythonDeephavenSession.java
```diff
@@ -82,12 +73,20 @@ public PythonDeephavenSession(ObjectTypeLookup objectTypeLookup, boolean runInit
...
+     * @throws TimeoutException if jpy times out while obtaining configuration details
      */
     public PythonDeephavenSession(
             ObjectTypeLookup objectTypeLookup, @Nullable final Listener listener, boolean runInitScripts,
             boolean isDefaultScriptSession)
-            throws IOException {
+            throws IOException, InterruptedException, TimeoutException {
         super(objectTypeLookup, listener, isDefaultScriptSession);
+
+        //TODO in the py init code that would call in here, we should read the jpy system properties
```
niloc132, line 85: At the very least I need to remove this comment (doesnt belong in this branch), but this JpyInit.init(log) is probably the main objectionable thing here.

niloc132:
need to remove io.deephaven.server.console.ConsoleServiceGrpcImpl import in DeephavenApiServer

engine/table/src/main/java/io/deephaven/engine/util/PythonDeephavenSession.java
devinrsmith: Yeah, previously it was done in `initializeGlobalScriptSession`; DeephavenApiServer.run may be the "best" place for now?

devinrsmith (commented):
Looks good, save for your comment.

niloc132:
> Yeah, previously it was done in initializeGlobalScriptSession; DeephavenApiServer.run may be the "best" place for now?

The problem with doing it in DeephavenApiServer is that you don't know if jpy will be used at all - by starting it in here, we only start it if the python script session will be used, and lazily (after any init is needed).

I'm not entirely certain that it must be lazy any more, but at least at one point it was required to be lazy (and so the `Provider<ScriptSession>.get()` call does the script session and jpy init when ready) - would it be okay if we deferred investigating this until we land the embedded server, so we can be sure everything else works?

nbauernfeind (approved):

PR #2341, opened 2022-04-26
Add ISO8601 support for convertDateTime.

### LINKED ISSUE #2340: Support ISO8601 via convertDateTime
Currently `io.deephaven.time.DateTimeUtils#convertDateTime` does not parse ISO8601:

```
java.lang.RuntimeException: Cannot parse datetime : 2022-04-26T00:30:31.087360Z
	at io.deephaven.time.DateTimeUtils.convertDateTime(DateTimeUtils.java:1754)
```

Essentially, I think `convertDateTime` should be a superset of `java.time.Instant#parse`

devinrsmith:
Fixes #2340

### DIFF
--- a/engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
+++ b/engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
@@ -1744,7 +1744,8 @@ public class DateTimeUtils {
     /**
      * Converts a {@link DateTime} String from a few specific zoned formats to a {@link DateTime}.
      *
-     * <p>Supports {@link DateTimeFormatter#ISO_INSTANT} format and others.
+     * <p>
+     * Supports {@link DateTimeFormatter#ISO_INSTANT} format and others.
      *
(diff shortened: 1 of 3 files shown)


rcaudy (approved):
I'm happy with this change.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -1741,7 +1742,9 @@ public static LocalDate convertDate(String s) {
     }
 
     /**
-     * Converts a DateTime String from a few specific zoned formats to a DateTime
+     * Converts a {@link DateTime} String from a few specific zoned formats to a {@link DateTime}.
```
chipkent, line 1745: @jmao-denver I think the associated pydocs need to get updated to indicate that the new format is also supported.

chipkent (changes requested):

jmao-denver:
should we have a new Python test case for the new format?

devinrsmith:
> should we have a new Python test case for the new format?

We got a java side test. I don't see too much value in replicating in python - we've already validated that `to_datetime` passes through appropriately.

PR #2343, opened 2022-04-26
Disable CLA check for forks other than main deephaven one.

jcferretti:
This is to support internal deephaven team members working on an intermediate branch on one of their forks, like this:
https://github.com/kosak/deephaven-core/pull/3

Corey and myself are working together on the C++ client, and we intend to work on a branch to Corey's fork.  That branch should not be too long lived, the intention is to merge it to `deephaven/deephaven-core` on a fairly regular basis; the idea is that on our shared branch we will be less strict about invariants (ie, things may compile but not run or not have tests).  We will only merge upstream to `deephaven/deephaven-core` more consistent points.

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -17,6 +17,7 @@ on:
 jobs:
   CLAssistant:
     runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'deephaven' }}
     steps:
       - name: "CLA Assistant"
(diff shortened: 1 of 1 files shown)


devinrsmith (approved):

mofojed (commented):
Can we add some details in the description about why we're doing this? I figure because you have multiple co-workers working out of your fork or...?

Just wondering if we need to have something in place for if you're working out of your fork and have a 3rd-party contribute to it, and then you merge it into main - are you just taking ownership of that contribution then?

PR #2345, opened 2022-04-26
BarragePerformanceLog metrics for static requests

nbauernfeind:
Supports a few metrics for `DoGet` and `BarrageSnapshotRequest`s.

### DIFF
--- a/server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
+++ b/server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
@@ -22,6 +22,7 @@ import io.deephaven.engine.table.impl.BaseTable;
 import io.deephaven.engine.table.impl.QueryTable;
 import io.deephaven.engine.table.impl.remote.ConstructSnapshot;
 import io.deephaven.engine.table.impl.util.BarrageMessage;
+import io.deephaven.extensions.barrage.BarragePerformanceLog;
 import io.deephaven.extensions.barrage.BarrageSnapshotOptions;
 import io.deephaven.extensions.barrage.BarrageSubscriptionOptions;
 import io.deephaven.extensions.barrage.chunk.ChunkInputStreamGenerator;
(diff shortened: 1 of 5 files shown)


server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
```diff
@@ -551,6 +570,19 @@ public void handleMessage(BarrageProtoUtil.MessageInfo message) {
...
+
+            private void flushMetrics(long bytesWritten, long writeTm) {
+                if (tableKey == null) {
+                    // metrics for this request are not to be reported
+                    return;
+                }
+
+                try {
+                    BarragePerformanceLog.getInstance().getStaticLogger()
+                            .log(tableId, tableKey, requestTm, queueTm, snapshotTm, writeTm, bytesWritten);
```
rcaudy, line 582: You're reporting nanos, but your columns claim to be millis. Same for the other `flushMetrics` in `FlightServiceGrpcImpl`.

server/src/main/java/io/deephaven/server/arrow/FlightServiceGrpcImpl.java
```diff
@@ -147,17 +151,50 @@ public void getSchema(final Flight.FlightDescriptor request,
         });
     }
 
+    private static class GetCustomMetricsHelper {
```
rcaudy, line 154: Maybe we should re-use this helper class in the two places we record metrics?

server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
nbauernfeind: Per our discussion; nanos into the log method which converts to millis.

rcaudy (approved):
Merge away if you've tested enough. Do we need to update any queries or examples for the column name changes?

PR #2348, opened 2022-04-27
Add LayoutHint column grouping API

### LINKED ISSUE #2273: Improve layout hint front and end API naming

mattrunyon:
This API will be used to support column grouping in the UI. Users will be able to specify multiple column groups containing children. The children can be other groups or a source column.

Validation of the layout hints will occur client-side. This is how the other hints are validated currently (i.e. the frozen column name is actually in the table).

Also fixes #2273 by renaming `atEnd` to `atBack`.

Note that the UI for groups does not exist yet, but the JS API model should have values set on `model.layoutHints.columnGroups`

Python
```python
from deephaven.column import string_col, int_col

hints = {
   'front': ['D'],
   'back': ['A'],
   'groups': [
      { 'name': 'MyGroup', 'children': ['A', 'B'] },
      { 'name': 'MyGroup2', 'children': ['A', 'B'], 'color': '#123456' },
   ]
}

base = new_table([
   string_col("A", ["A"]),
   int_col("B", [2]),
   string_col("C", ["C"]),
   int_col("D", [4]),
   string_col("E", ["E"]),
   int_col("F", [6])
])


t1 = base.layout_hints(front='D', back=['A'], freeze=["C", 'E'], hide=['F'], groups=[{'name': 'Test', 'children': ['A', 'B']}])
t2 = base.layout_hints(**hints)

print(t1.j_table.getAttribute("LayoutHints"))
print(t2.j_table.getAttribute("LayoutHints"))
```

Groovy
```groovy

hints = LayoutHintBuilder.get()
   .freeze("NameOfIntCol")
   .group("MyGroup", ["NameOfIntCol", "NameOfStringCol"], "MEDIUMAQUAMARINE")
   .group("MyGroup2", ["COL", "OTHER_COL"]);


t = newTable(
   stringCol("NameOfStringCol", "Data String 1", 'Data String 2', "Data String 3"),
   intCol("NameOfIntCol", 4, 5, 6)
).setLayoutHints(hints.build());

println(hints.build())
println(LayoutHintBuilder.fromString(hints.build()).build()) // Test fromString
```

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
@@ -113,24 +113,21 @@ public class LayoutHintBuilder {
 
     private static class ColumnGroup {
 
-        String name;
-        final List<String> children;
-        Color color;
+        private final String name;
+        private final List<String> children;
+        private final Color color;
 
-        ColumnGroup(String name, List<String> children, Color color) {
+        public ColumnGroup(String name, List<String> children, Color color) {
             NameValidator.validateColumnName(name);
             children.forEach(c -> NameValidator.validateColumnName(c));
 
             this.name = name;
             this.children = children;
-
-            if (color != null) {
-                this.color = color;
-            }
+            this.color = color;
         }
 
         @NotNull
-        String serialize() {
+        public String serialize() {
             StringBuilder sb = new StringBuilder("name:" + name);
 
             sb.append("::children:" + StringUtils.joinStrings(children, ","));
@@ -147,16 +144,16 @@ public class LayoutHintBuilder {
          * @return a ColumnGroup instance
          */
         @NotNull
-        static ColumnGroup fromString(String string) {
+        public static ColumnGroup fromString(String string) {
             final Map<String, String> options = Arrays.stream(string.split("::"))
                     .map(option -> option.split(":"))
-                    .collect(Collectors.toMap(parts -> parts[0], parts -> parts.length == 2 ? parts[1] : ""));
+                    .collect(Collectors.toMap(parts -> parts[0], parts -> parts.length == 2 ? parts[1] : null));
 
             final String name = options.get("name");
             final List<String> children = Arrays.asList(options.get("children").split(","));
             final String color = options.get("color");
 
-            if (color == null || color.length() == 0) {
+            if (color == null) {
                 return new ColumnGroup(name, children, null);
             }
 
@@ -376,7 +373,7 @@ public class LayoutHintBuilder {
     @ScriptApi
     public LayoutHintBuilder group(String name, List<String> children, Color color) {
         if (columnGroups == null) {
-            columnGroups = new HashMap<>();
+            columnGroups = new LinkedHashMap<>();
         }
 
         if (children.size() == 0) {
@@ -390,7 +387,7 @@ public class LayoutHintBuilder {
 
     private void addGroupData(ColumnGroup group) {
         if (columnGroups == null) {
-            columnGroups = new HashMap<>();
+            columnGroups = new LinkedHashMap<>();
         }
 
         columnGroups.put(group.name, group);
--- a/engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
@@ -30,6 +30,8 @@ public class LayoutHintBuilder {
     private Set<String> alwaysSubscribedCols;
     private Set<String> groupableColumns;
 
+    private Map<String, ColumnGroup> columnGroups;
+
     /**
      * Helper class to maintain sub-properties for auto filter columns
      */
@@ -106,6 +108,31 @@ public class LayoutHintBuilder {
 
             return new AutoFilterData(column, localFetchSize);
         }
+    }
+
+    private static class ColumnGroup {
+
+        String name;
+        final List<String> children;
+        String color;
+
+        ColumnGroup(String name, List<String> children, String color) {
+            this.name = name;
+            this.children = children;
+            this.color = color;
+        }
+
+        @NotNull
+        String serialize() {
+            StringBuilder sb = new StringBuilder("name:" + name);
+
+            sb.append("::children:" + StringUtils.joinStrings(children, ","));
+            if (color != null) {
+                sb.append("::color:" + color);
+            }
+            return sb.toString();
+        }
+
     }
 
     private LayoutHintBuilder() {
@@ -279,6 +306,26 @@ public class LayoutHintBuilder {
         }
 
         hiddenCols.addAll(cols);
+
+        return this;
+    }
+
+    @ScriptApi
+    public LayoutHintBuilder group(String name, List<String> cols) {
+        return group(name, cols, null);
+    }
+
+    @ScriptApi
+    public LayoutHintBuilder group(String name, List<String> children, String style) {
+        if (columnGroups == null) {
+            columnGroups = new HashMap<>();
+        }
+
+        if (children.size() == 0) {
+            columnGroups.remove(name);
+        } else {
+            columnGroups.put(name, new ColumnGroup(name, children, style));
+        }
 
         return this;
     }
@@ -480,6 +527,10 @@ public class LayoutHintBuilder {
 
         if (groupableColumns != null && !groupableColumns.isEmpty()) {
             sb.append("groupable=").append(String.join(",", groupableColumns)).append(';');
+        }
+
+        if (columnGroups != null && !columnGroups.isEmpty()) {
+            columnGroups.values().forEach(g -> sb.append("group=").append(g.serialize()).append(';'));
         }
 
         return sb.toString();
--- a/engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
@@ -131,7 +131,7 @@ public class LayoutHintBuilder {
             StringBuilder sb = new StringBuilder("name:").append(name);
 
             sb.append("::children:");
-            boolean first = false;
+            boolean first = true;
             for (String child : children) {
                 if (!first) {
                     sb.append(",");
--- a/engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
@@ -161,7 +161,8 @@ public class LayoutHintBuilder {
         }
     }
 
-    private LayoutHintBuilder() {}
+    private LayoutHintBuilder() {
+    }
 
     // region Builder Methods
 
@@ -365,9 +366,9 @@ public class LayoutHintBuilder {
     /**
      * Create a named group of columns in the UI
      *
-     * @param name the column group name. Must be a valid Deephaven column name
+     * @param name     the column group name. Must be a valid Deephaven column name
      * @param children the columns and other groups belonging to this group
-     * @param color the background color for the group in the UI
+     * @param color    the background color for the group in the UI
      * @return this LayoutHintBuilder
      */
     @ScriptApi
@@ -437,7 +438,7 @@ public class LayoutHintBuilder {
      * Set the default initial number of rows to fetch for columns that have been marked as
      * {@link LayoutHintBuilder#autoFilter(Collection) AutoFilter} columns.
      *
-     * @param col the column to set the fetch size for
+     * @param col  the column to set the fetch size for
      * @param size the number of rows to fetch initially
      * @return this LayoutHintBuilder
      */
@@ -576,7 +577,7 @@ public class LayoutHintBuilder {
 
         if (autoFilterCols != null && !autoFilterCols.isEmpty()) {
             sb.append("autofilter=").append(
-                    StringUtils.joinStrings(autoFilterCols.values().stream().map(AutoFilterData::serialize), ","))
+                            StringUtils.joinStrings(autoFilterCols.values().stream().map(AutoFilterData::serialize), ","))
                     .append(';');
         }
 
@@ -594,9 +595,9 @@ public class LayoutHintBuilder {
 
         if (columnGroups != null && !columnGroups.isEmpty()) {
             sb.append("columnGroups=");
-            List<String> groupStrings =
-                    columnGroups.values().stream().map(ColumnGroup::serialize).collect(Collectors.toList());
-            sb.append(String.join("|", groupStrings)).append(';');
+            String groupStrings =
+                    columnGroups.values().stream().map(ColumnGroup::serialize).collect(Collectors.joining("|"));
+            sb.append(groupStrings).append(';');
         }
 
         return sb.toString();
--- a/engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
@@ -141,7 +141,7 @@ public class LayoutHintBuilder {
             }
             if (color != null) {
                 sb.append("::color:#")
-                    .append(Integer.toHexString(color.javaColor().getRGB()).substring(2));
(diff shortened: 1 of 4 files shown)


py/server/deephaven/table.py
```diff
@@ -1244,15 +1244,22 @@ def format_row_where(self, cond: str, formula: str) -> Table:
             raise DHError(e, "failed to color format rows conditionally.") from e
 
     def layout_hints(self, front: Union[str, List[str]] = None, back: Union[str, List[str]] = None,
-                     freeze: Union[str, List[str]] = None, hide: Union[str, List[str]] = None) -> Table:
+                     freeze: Union[str, List[str]] = None, hide: Union[str, List[str]] = None,
+                     groups: List[dict] = None) -> Table:
```
jmao-denver, line 1248: do we want to allow Union[dict, List[dict]] for groups?

py/server/deephaven/table.py
```diff
@@ -1244,15 +1244,22 @@ def format_row_where(self, cond: str, formula: str) -> Table:
...
             freeze (Union[str, List[str]]): the columns to freeze to the front.
                 These will not be affected by horizontal scrolling.
-            hide (Union[str, List[str]]): the columns to hide
+            hide (Union[str, List[str]]): the columns to hide.
+            groups (List[ColumnGroup]): A list of dicts specifying which columns should be grouped in the UI
+                The dicts can specify the following:
+
+                name (str): The group name
+                children (List[str]): The
+                color (Optional[Union[str, Color]]): The color name, hex string, or color object from deephaven.plot
```
jmao-denver, line 1263: Which definition does ```Color``` reference to? I don't see the deephaven.plot.color.Color being imported nor tested.

jmao-denver (commented):
Only reviewed the Python code changes.

py/server/deephaven/table.py
mattrunyon: I think just the `List[dict]` is fine since the prop name implies there are multiple, but if you think we should allow both then I can switch it

py/server/deephaven/table.py
mattrunyon: Removed `Color` and just used `str` for now. `Color` would have been used by `from python.plot import Colors` and then using `Colors.RED` as an example. The string param can just be the name and the Java method will get the matching color constant

jmao-denver (dismissed):

web/client-api/src/main/java/io/deephaven/web/client/api/JsLayoutHints.java
```diff
@@ -2,18 +2,55 @@
...
+        }
+    }
 
     private boolean savedLayoutsAllowed = true;
     private String[] frontColumns;
     private String[] backColumns;
     private String[] hiddenColumns;
     private String[] frozenColumns;
 
+    private Object[] columnGroups;
```
niloc132, line 52: ColumnGroup[]

web/client-api/src/main/java/io/deephaven/web/client/api/JsLayoutHints.java
```diff
@@ -48,6 +85,14 @@ public JsLayoutHints parse(String hints) {
             frozenColumns = JsObject.freeze(freezeStr.split(","));
         }
 
+        final String groupsStr = options.get("groups");
+        if (groupsStr != null && !groupsStr.isEmpty()) {
+            List<ColumnGroup> groups =
+                    Arrays.stream(groupsStr.split("\\|")).map(ColumnGroup::new).collect(Collectors.toList());
+
+            columnGroups = JsObject.freeze(groups.toArray());
```
niloc132, line 93: toArray(new ColumnGroup[0])

web/client-api/src/main/java/io/deephaven/web/client/api/JsLayoutHints.java
```diff
@@ -48,6 +85,14 @@ public JsLayoutHints parse(String hints) {
             frozenColumns = JsObject.freeze(freezeStr.split(","));
         }
 
+        final String groupsStr = options.get("groups");
+        if (groupsStr != null && !groupsStr.isEmpty()) {
+            List<ColumnGroup> groups =
+                    Arrays.stream(groupsStr.split("\\|")).map(ColumnGroup::new).collect(Collectors.toList());
```
niloc132, line 91: probably should freeze each column group instance as they are created, add another .map() in there for it?

engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
```diff
@@ -278,6 +343,56 @@ public LayoutHintBuilder hide(Collection<String> cols) {
...
+     *
+     * @param name the column group name. Must be a valid Deephaven column name
+     * @param children the columns and other groups belonging to this group
+     * @param color the background color for the group in the UI
+     * @return this LayoutHintBuilder
+     */
+    @ScriptApi
+    public LayoutHintBuilder group(String name, List<String> children, Color color) {
+        if (columnGroups == null) {
+            columnGroups = new HashMap<>();
```
niloc132, line 376: does order matter here, of the named groups? should this be a LinkedHashMap instead to keep that order?

engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
```diff
@@ -105,9 +108,63 @@ static AutoFilterData fromString(String string) {
         }
     }
 
+    private static class ColumnGroup {
+
+        String name;
```
niloc132, line 113: fields should be private, and it looks like final too

engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
```diff
@@ -105,9 +108,63 @@ static AutoFilterData fromString(String string) {
...
+
+        ColumnGroup(String name, List<String> children, Color color) {
+            NameValidator.validateColumnName(name);
+            children.forEach(c -> NameValidator.validateColumnName(c));
+
+            this.name = name;
+            this.children = children;
+
+            if (color != null) {
+                this.color = color;
```
niloc132, line 125: since we leave it null if the passed in color is null, why not just assign it without an if?

engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
```diff
@@ -105,9 +108,63 @@ static AutoFilterData fromString(String string) {
         }
     }
 
+    private static class ColumnGroup {
+
+        String name;
+        final List<String> children;
+        Color color;
+
+        ColumnGroup(String name, List<String> children, Color color) {
```
niloc132, line 117: constructor/methods should almost never be package-protected like this, but explicitly public/private/etc

engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
```diff
@@ -105,9 +108,63 @@ static AutoFilterData fromString(String string) {
...
+         * Convert a string of the format defined by {@link #serialize()} into a proper ColumnGroup object
+         *
+         * @param string the string to parse
+         * @return a ColumnGroup instance
+         */
+        @NotNull
+        static ColumnGroup fromString(String string) {
+            final Map<String, String> options = Arrays.stream(string.split("::"))
+                    .map(option -> option.split(":"))
+                    .collect(Collectors.toMap(parts -> parts[0], parts -> parts.length == 2 ? parts[1] : ""));
```
niloc132, line 150: why empty string instead of null? explicit null makes your `if` below simpler, and i dont think either empty string nor null are safe for name, or children?

engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
mattrunyon: I'm not sure if order will matter yet (haven't fully gotten into the UI side of determining how these will be ordered when a user specifies columns 1, 2, 3 in 1 group and 0, 4, 5 in another). We might need the definition order so I'll switch to `LinkedHashMap` to maintain it

engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
mattrunyon: I think I was just going off of the `JsLayoutHints.parse` call. Explicit null will work here

web/client-api/src/main/java/io/deephaven/web/client/api/JsLayoutHints.java
```diff
@@ -75,4 +121,9 @@ public String[] getHiddenColumns() {
     public String[] getFrozenColumns() {
         return frozenColumns;
     }
+
+    @JsProperty
+    public Object[] getColumnGroups() {
```
niloc132, line 126: return ColumnGroup[]

web/client-api/src/main/java/io/deephaven/web/client/api/JsLayoutHints.java
```diff
@@ -48,6 +85,15 @@ public JsLayoutHints parse(String hints) {
             frozenColumns = JsObject.freeze(freezeStr.split(","));
         }
 
+        final String groupsStr = options.get("groups");
+        if (groupsStr != null && !groupsStr.isEmpty()) {
```
niloc132, line 89: double checking: this means that the `columnGroups` will be null and not empty if groupsStr is either absent, provided as null, or an empty string. wouldn't it be nicer to work with if it was never null, but instead was empty?

mattrunyon: I'm just mimicking the logic from the other keys. It could be null b/c the layout hint builder does not add elements to the layout hints string if they aren't specified (i.e. no `columnGroups=""`)

py/server/deephaven/table.py
```diff
@@ -1333,15 +1333,22 @@ def format_row_where(self, cond: str, formula: str) -> Table:
...
-            front (Union[str, List[str]]): the columns to show at the front
-            back (Union[str, List[str]]): the columns to show at the back
+            front (Union[str, List[str]]): the columns to show at the front.
+            back (Union[str, List[str]]): the columns to show at the back.
             freeze (Union[str, List[str]]): the columns to freeze to the front.
                 These will not be affected by horizontal scrolling.
-            hide (Union[str, List[str]]): the columns to hide
+            hide (Union[str, List[str]]): the columns to hide.
+            column_groups (List[ColumnGroup]): A list of dicts specifying which columns should be grouped in the UI
+                The dicts can specify the following:
```
jmao-denver, line 1347: should be LIst[dict]?

jmao-denver (changes requested):
Found one inconsistency in table.py

engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
```diff
@@ -480,12 +592,19 @@ public String build() {
             sb.append("groupable=").append(String.join(",", groupableColumns)).append(';');
         }
 
+        if (columnGroups != null && !columnGroups.isEmpty()) {
+            sb.append("columnGroups=");
+            List<String> groupStrings =
+                    columnGroups.values().stream().map(ColumnGroup::serialize).collect(Collectors.toList());
```
niloc132, line 598: better: skip the list, and just use Collectors.joining(...) to emit a single string here.

or, iterate as you walk the stream and append to the stringbuilder:

```
...map(ColumnGroup::serialize).forEach(str -> sb.append("|").append(str));
```

web/client-api/src/main/java/io/deephaven/web/client/api/JsLayoutHints.java
```diff
@@ -51,6 +88,15 @@ public JsLayoutHints parse(String hints) {
             frozenColumns = JsObject.freeze(freezeStr.split(","));
         }
 
+        final String groupsStr = options.get("columnGroups");
+        if (groupsStr != null && !groupsStr.isEmpty()) {
+            List<ColumnGroup> groups =
+                    Arrays.stream(groupsStr.split("\\|")).map(ColumnGroup::new).map(JsObject::freeze)
+                            .collect(Collectors.toList());
+
+            columnGroups = JsObject.freeze(groups.toArray(new ColumnGroup[0]));
```
niloc132, line 97: skip the List and just call .toArray(ColumnGroup[]::new) on the stream itself, and freeze that

jmao-denver (dismissed):
LGTM

engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
```diff
@@ -108,9 +111,60 @@ static AutoFilterData fromString(String string) {
...
+
+        @NotNull
+        public String serialize() {
+            StringBuilder sb = new StringBuilder("name:" + name);
+
+            sb.append("::children:" + StringUtils.joinStrings(children, ","));
+            if (color != null) {
+                sb.append("::color:" + "#" + Integer.toHexString(color.javaColor().getRGB()).substring(2));
+            }
+            return sb.toString();
```
niloc132, line 137: Sorry, I should have called this out explicitly, but it is the same problem as below. This isn't a fatal issue, but it is "wrong" and to be avoided.

As with the code you changed below, here we have a collection of strings that need to be concatenated together, along with extra contents before and after. 

This is pretty minor in the grand scheme of things, but mixing building big strings in a helper method, then concating that long string with a small string, then appending that entire string "efficiently" into a string builder... If you're aiming for readability, just do away with the StringBuilder entirely and stick +s everywhere, but if you want the extra efficiency of the StringBuilder to handle that conditional below, try to avoid building little strings into big strings, and then big strings into bigger strings, but just accumulate the little strings into the bigger strings.

In JS, strings are usually implemented as "ropes", which can be thought of very loosely as a linked list or tree of actual character arrays, so that insertion at the beginning or end is cheap, and doesn't reallocate everything - but you, the developer, have very little control over what the engine will do when you actually want to say "ok I'm done building, make it all one byte array now".

In contrast, the Java string implementation is pretty nearly just an immutable wrapper over byte[] or char[], where you're expected to ... pretty much avoid expensive things, or they are expensive. A handful of cases are handled automatically and cheaply, but they basically boil down to "one long expression of string concatenation will be replaced by a StringBuilder with its append calls". Downside, you have to work a bit to get performance, upside, optimal performance is actually optimal.

Notes, sort of chronologically with this code, and a proposed set of changes to understand:
 * Don't `+` within append()s (and the constructor is sort of the first append). You can chain append calls if it improves readability, but split all `+`s up with appends instead.
 * Nested method calls that do lots of sub-stringbuilder or sub-`+` work is also not good
```suggestion
            StringBuilder sb = new StringBuilder("name:").append(name);

            sb.append("::children:");
            boolean first = false;
            for (String child : children) {
                if (!first) {
                    sb.append(",");
                }
                first = false;
                sb.append(child);
            }
            if (color != null) {
                sb.append("::color:#")
                    .append(Integer.toHexString(color.javaColor().getRGB()).substring(2));
            }
            return sb.toString();
```

engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
```diff
@@ -281,6 +352,56 @@ public LayoutHintBuilder hide(Collection<String> cols) {
...
+     * @param color the background color for the group in the UI
+     * @return this LayoutHintBuilder
+     */
+    @ScriptApi
+    public LayoutHintBuilder columnGroup(String name, List<String> children, Color color) {
+        if (columnGroups == null) {
+            columnGroups = new LinkedHashMap<>();
+        }
+
+        if (children.size() == 0) {
```
niloc132, line 388: generally prefer isEmpty()

niloc132 (approved):

niloc132 (approved):

PR #2350, opened 2022-04-27
Support providing Avro schemas as a JSON schema definition from python.  Fixes #2305.

### LINKED ISSUE #2305: Python integration for Kafka should support specifying schema as a string
As a user of our Kafka integration, I want to be able to use the Python wrappers when I have a schema as a string and no access to a schema registry server.

jcferretti:
Deephaven - Google Chrome_174

### DIFF
--- a/py/server/deephaven/stream/kafka/consumer.py
+++ b/py/server/deephaven/stream/kafka/consumer.py
@@ -230,7 +230,9 @@ def avro_spec(
     """Creates a spec for how to use an Avro schema when consuming a Kafka stream to a Deephaven table.
 
     Args:
-        schema (str): the name for a schema registered in a Confluent compatible Schema Server. The associated
+        schema (str): Either a JSON encoded Avro schema definition string, or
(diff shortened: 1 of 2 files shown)


rcaudy (approved):

py/server/deephaven/stream/kafka/consumer.py
```diff
@@ -250,14 +252,26 @@ def avro_spec(
         if mapping is not None:
             mapping = _dict_to_j_func(mapping, mapped_only)
 
-        if mapping:
-            return KeyValueSpec(
-                j_spec=_JKafkaTools_Consume.avroSpec(schema, schema_version, mapping)
-            )
+        if schema.strip().startswith("{"):
+            jschema = _JKafkaTools.getAvroSchema(schema);
```
jmao-denver, line 256: could use json.load() to check if the supplied string is valid JSON, but delegating the check to Java is OK here.

jmao-denver (approved):

jmao-denver (approved):

PR #2362, opened 2022-05-02
PythonDeephavenSession#getVariables should unwrap PyObjects

### LINKED ISSUE #2360: listFlights does not work against python session
It seems like listFlight against a python session does not work (empty list returned). listFields *does* work. At least since 0.11.0.

Tested with a python console session and the java client examples io.deephaven.client.examples.ListTables and io.deephaven.client.examples.SubscribeToFields.

nbauernfeind:
Fixes #2360

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/util/PythonDeephavenSession.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/PythonDeephavenSession.java
@@ -36,6 +36,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


engine/table/src/main/java/io/deephaven/engine/util/PythonDeephavenSession.java
```diff
@@ -193,7 +194,9 @@ protected void evaluate(String command, String scriptName) {
 
     @Override
     public Map<String, Object> getVariables() {
-        return Collections.unmodifiableMap(scope.getEntriesMap());
+        final Map<String, Object> outMap = new LinkedHashMap<>();
+        scope.getEntriesMap().forEach((key, value) -> outMap.put(key, maybeUnwrap(value)));
+        return outMap;
```
devinrsmith, line 199: I'm wondering if `io.deephaven.engine.util.PythonScopeJpyImpl#convertValue` or `io.deephaven.engine.util.PythonScopeJpyImpl#convert` is the more appropriate place to do the unwrapping.

devinrsmith, line 199: I'm wondering if `io.deephaven.engine.util.PythonScopeJpyImpl#convertValue` or `io.deephaven.engine.util.PythonScopeJpyImpl#convert` is the more appropriate place to do the unwrapping.

devinrsmith (approved):

PR #2365, opened 2022-05-04
Optimize snapshots for large tables

lbooker42:
Allow multiple snapshots for `DoGet` and `BarrageSnapshotRequest` when the table is static or append-only.  Add the capability to snapshot using subscribe-and-seal and make that the default for worker-to-worker and the java-client.

- break up DoGet into reasonable snapshots when static or append-only
- implement barrage repo changes (like client-provided `max_message_size` and removal of  `numBatches`)
- provide the mechanism for snapshot-through-subscribe for java-client and W2W


server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
```diff
@@ -66,6 +71,166 @@ public class ArrowFlightUtil {
...
+                });
+    }
+
+    private static void createAndSendSnapshot(BaseTable table, BitSet columns, RowSet viewport, boolean reverseViewport,
+            BarrageSnapshotOptions snapshotRequestOptions, StreamObserver<BarrageStreamGenerator.View> listener,
+            BarragePerformanceLog.SnapshotMetricsHelper metrics) {
+        // if the table is static or append-only and a full snapshot is requested, we can make and send multiple
+        // snapshots to save memory and operate more efficiently
+
+        if (viewport == null && (!table.isRefreshing() || table.isAddOnly())) {
```
lbooker42, line 128: Oops, AddOnly  != AppendOnly, can't verify this attribute

proto/proto-backplane-grpc/build.gradle
```diff
@@ -38,7 +38,7 @@ dependencies {
   // technically we should have a runtime dependency on barrage-core, but instead we are regenerating
   // the same output that it contains, and we have declared the correct dependencies as necessary
   //  compile 'io.deephaven.barrage:barrage-core:0.4.0'
```
nbauernfeind, line 40: Probably a silly request, but can you bump the barrage version in the comment as well?

server/jetty-app/build.gradle
```diff
@@ -28,6 +28,7 @@ def extraJvmArgs = [
         '-XX:MaxGCPauseMillis=100',
         '-XX:+UseStringDeduplication',
         '-XshowSettings:vm',
+        '-Xmx48G',
```
nbauernfeind, line 31: revert before committing

server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
nbauernfeind, line 128: As Ryan mentioned the other day, this is `isAddOnly` and not `isAppendOnly`. We'll have to drop to static tables only.

server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
```diff
@@ -66,6 +71,166 @@ public class ArrowFlightUtil {
...
+
+            boolean isComplete = false;
+
+            try (final WritableRowSet snapshotViewport = RowSetFactory.empty()) {
+                while (!isComplete) {
+                    // compute the next range to snapshot
+                    final long cellCount =
+                            Math.max(MIN_SNAPSHOT_CELL_COUNT,
+                                    Math.min(snapshotTargetCellCount, MAX_SNAPSHOT_CELL_COUNT));
+                    long endKey = startKey + (cellCount / columnCount) - 1;
```
nbauernfeind, line 145: I feel that we should try to batch even if there was a requested viewport.

server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
```diff
@@ -66,6 +71,166 @@ public class ArrowFlightUtil {
...
+                                        columns, snapshotPartialViewport, null);
+                        msg.modColumnData = ZERO_MOD_COLUMNS; // no mod column data for DoGet
+                        long elapsed = System.nanoTime() - start;
+                        // accumulate snapshot time in the metrics
+                        metrics.snapshotNanos += elapsed;
+
+                        // send out the data. Note that although a `BarrageUpdateMetaData` object will
+                        // be provided with each unique snapshot, vanilla Flight clients will ignore
+                        // these and see only an incoming stream of batches
+                        try (final BarrageStreamGenerator bsg = new BarrageStreamGenerator(msg, metrics)) {
```
nbauernfeind, line 170: Can any SizeExceptions be thrown here or from `getSnapshotView`? Should we attempt to further break this request up if things get too large?

server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
```diff
@@ -66,6 +71,166 @@ public class ArrowFlightUtil {
...
+                    final RowSet keySpaceViewport =
+                            viewport != null
+                                    ? msg.rowsAdded.subSetForPositions(viewport, reverseViewport)
+                                    : null) {
+                listener.onNext(
+                        bsg.getSnapshotView(snapshotRequestOptions, viewport,
+                                reverseViewport, keySpaceViewport, columns));
+            }
+        }
+    }
```
nbauernfeind, line 232: A minor nit on the method structure. The `if`/`else` feels like the majority of the method and therefore a less readable indentation. Let's move the iterative implementation into a helper, return, and drop the `else` block. Something like this:

```java
if (!table.isRefreshing()) {
    createAndSendStaticSnapshot(...);
    return;
}

// the original snapshot implementation that fetches a single BarrageMessage for the entire request
```

server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
```diff
@@ -716,6 +861,8 @@ public synchronized void close() {
                 if (preExportSubscriptions != null) {
                     preExportSubscriptions = null;
                 }
+
+                listener.onCompleted();
```
nbauernfeind, line 865: We probably want to do this only when `bmp == null`. The `bmp` will call `onCompleted` during the update propagation job after having invoked `removeSubscription`. 

Also, should we wrap this call in a `GrpcUtil.safelyExecute`?

Note that if we call `onCompleted` after having passed the listener to the bmp, then `onCompleted` must be called `GrpcUtil.safelyExecuteLocked(listener, listener::onCompleted)`. The reason is that the observers are not thread safe and you can end up deadlocking in gRPC land (the bmp might be in the middle of sending another update before actually removing the subscription).

server/src/main/java/io/deephaven/server/barrage/BarrageStreamGenerator.java
```diff
@@ -473,14 +481,13 @@ public SnapshotView(final BarrageStreamGenerator generator,
...
             // precompute add row offsets
             if (keyspaceViewport != null) {
-                try (WritableRowSet intersect = keyspaceViewport.intersect(generator.rowsIncluded.original)) {
-                    addRowOffsets = generator.rowsIncluded.original.invert(intersect);
-                }
+                addRowKeys = keyspaceViewport.intersect(generator.rowsIncluded.original);
+                addRowOffsets = generator.rowsIncluded.original.invert(addRowKeys);
             } else {
+                addRowKeys = generator.rowsAdded.original.copy();
                 addRowOffsets = RowSetFactory.flat(generator.rowsAdded.original.size());
```
nbauernfeind, line 488: I liked the above simplification, which can be done here, too.

```suggestion
                addRowOffsets = RowSetFactory.flat(addRowKeys.size());
```

web/client-backplane/src/main/java/io/deephaven/javascript/proto/dhinternal/io/deephaven/barrage/flatbuf/barrage_generated/io/deephaven/barrage/flatbuf/BarrageSnapshotOptions.java
```diff
@@ -15,13 +15,16 @@ public class BarrageSnapshotOptions {
     public static native void addColumnConversionMode(
             Builder builder, int columnConversionMode);
 
+    public static native void addMaxMessageSize(Builder builder, double maxMessageSize);
```
nbauernfeind, line 18: Were these flatbuffer changes made manually, or were they generated? (I don't know if you are aware these are generated, but require extra steps.)

props/test-configs/src/main/resources/dh-tests.prop
```diff
@@ -98,3 +98,12 @@ Liveness.cleanupLogEnabled=false
 
 # Force more parallelism
 QueryTable.minimumParallelSelectRows=1024
+
+BarrageMessageProducer.subscriptionGrowthEnabled=true
```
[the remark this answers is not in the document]
lbooker42: I'll get Ryan to comment on this.  This inclusion was accidental but it might be something we want tested

web/client-backplane/src/main/java/io/deephaven/javascript/proto/dhinternal/io/deephaven/barrage/flatbuf/barrage_generated/io/deephaven/barrage/flatbuf/BarrageSnapshotOptions.java
lbooker42: @niloc132 generated these since I was having trouble on my M1 dev machine

server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
lbooker42: I made this change, but please verify it looks correct.

server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
```diff
@@ -69,6 +71,185 @@ public class ArrowFlightUtil {
...
+                }
+            }
+        }
+    }
+
+    private static void createAndSendSnapshot(BaseTable table, BitSet columns, RowSet viewport, boolean reverseViewport,
+            BarrageSnapshotOptions snapshotRequestOptions, StreamObserver<BarrageStreamGenerator.View> listener,
+            BarragePerformanceLog.SnapshotMetricsHelper metrics) {
+
+        // if the table is static or append-only and a full snapshot is requested, we can make and send multiple
```
nbauernfeind, line 218: I believe the `append-only` portion of this comment is invalid - as in we don't have a way to detect this.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/table/BarrageTable.java
```diff
@@ -203,6 +209,10 @@ public BitSet getServerColumns() {
         return serverColumns;
     }
 
+    public void setInitialSnapshotViewportRowCount(long rowCount) {
+        initialSnapshotViewportRowCount = rowCount;
```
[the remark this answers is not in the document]
lbooker42: No, this is called before the initial request is ever made while serverViewport is already `null`. `serverViewport` will get updated when the first incoming barrage message is received.

nbauernfeind (approved):

PR #2367, opened 2022-05-05
Open Addressed Natural Join Hash Table

cpwright:


cpwright (commented):
Placeholder 3.

engine/table/src/main/java/io/deephaven/engine/table/impl/naturaljoin/typed/rightincopen/gen/RightIncrementalNaturalJoinHasherByte.java
```diff
@@ -0,0 +1,258 @@
...
+                        final long leftRowKeyForState = leftRowSet.getUnsafe(tableLocation).firstRowKey();
+                        throw new IllegalStateException("Natural Join found duplicate right key for " + extractKeyStringFromSourceTable(leftRowKeyForState));
+                    }
+                    found = true;
+                    break;
+                }
+                tableLocation = nextTableLocation(tableLocation);
+                Assert.neq(tableLocation, "tableLocation", firstTableLocation, "firstTableLocation");
+            }
+            if (!found) {
```
lbooker42, line 96: Is some action or assert expected here? (also in `removeRight()` / `addRightSide()` `modifyByRight()` / `applyRightShift()`?

engine/table/src/main/java/io/deephaven/engine/table/impl/naturaljoin/typed/staticopen/gen/StaticNaturalJoinHasherByte.java
```diff
@@ -0,0 +1,157 @@
...
+
+    protected void decorateWithRightSide(RowSequence rowSequence, Chunk[] sourceKeyChunks) {
+        final ByteChunk<Values> keyChunk0 = sourceKeyChunks[0].asByteChunk();
+        final LongChunk<OrderedRowKeys> rowKeyChunk = rowSequence.asRowKeyChunk();
+        final int chunkSize = keyChunk0.size();
+        for (int chunkPosition = 0; chunkPosition < chunkSize; ++chunkPosition) {
+            final byte k0 = keyChunk0.get(chunkPosition);
+            final int hash = hash(k0);
+            final int firstTableLocation = hashToTableLocation(hash);
+            boolean found = false;
```
lbooker42, line 132: The boolean `found` is written but never read in this function

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/BooleanSparseArraySource.java
```diff
@@ -834,13 +834,13 @@ private ReinterpretedAsByte(BooleanSparseArraySource wrapped) {
         }
 
         @Override
-        public byte getByte(long rowKey) {
-            return wrapped.getByte(rowKey);
+        public byte getByte(long rowSet) {
```
lbooker42, line 837: Curious renaming, `rowSet` seems incorrect. `index` or `rowKey` seems more appropriate

lbooker42 (dismissed):
Cosmetic items, mostly in the auto-gen code. LGTM, for what my opinion is worth.

rcaudy:
Closes #2380

engine/table/src/main/java/io/deephaven/engine/table/impl/naturaljoin/typed/rightincopen/gen/RightIncrementalNaturalJoinHasherByte.java
cpwright: No, we don't actually care.  If there is no state (because it was never added from the left), then a missing right hand side is fine by us, we drop it to avoid holding onto extra memory.

lbooker42 (approved):
Looks good, approved

PR #2374, opened 2022-05-10
Use CachedSchemaRegistryClient instead of manual REST.

### LINKED ISSUE #2373: Allow setting confluent standard properties for schema registry
Currently, `io.deephaven.kafka.KafkaTools#getAvroSchema(java.lang.String, java.lang.String, java.lang.String)` uses a manual REST request to fetch avro schemas. This works in the simple case, but won't work as-is with self-signed certificates or mTLS (or other configuration that confluent has available for schema registries, https://docs.confluent.io/platform/current/schema-registry/security/index.html#).

We should be able to use `io.confluent.kafka.schemaregistry.client.SchemaRegistryClient` to pick up all the necessary properties.

This is a useful feature for connecting to production schema registries.

devinrsmith:
Remove extra dependencies. Update avro and kafka dependencies.

Fixes #2373

### DIFF
--- a/extensions/kafka/build.gradle
+++ b/extensions/kafka/build.gradle
@@ -6,20 +6,17 @@ plugins {
 description 'Kafka: Integrating Engine tables with Kafka'
 
 dependencies {
-    api project(':engine-table'),
-            'org.apache.kafka:kafka-clients:2.4.0',
-            'org.apache.avro:avro:1.9.2',
-            'org.apache.httpcomponents:httpclient:4.5.13'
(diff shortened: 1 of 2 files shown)


devinrsmith:
I've looked at the python, and afaict, nobody is calling any of the methods that I have removed.

devinrsmith:
I've verified that I'm able to now set the property `schema.registry.ssl.keystore.location` and have it work wrt the schema registry and KafkaTools.

jcferretti (approved):

Issue #2375, opened 2022-05-10
Parquet support cleanup, consistency with DHE


niloc132:
This ticket covers assorted cleanup in the parquet implementation, mostly around code quality, as a pre-req for #2118. This will bring DHC in line with some DHE, which should make porting other features easier. The goal of this ticket is to track changes to make them consistent, other points listed here might be handled later.

Some fixes are obvious, so I'm just going to make them rather than list them here as I go over the diff, but others will be listed here for review:
 * ColumnWriterImpl.writePageV2 is unused
 * GenericArrayBuffer is unused
 * ParquetFIleWriter.addRowGroup is unused
 * RowGroupWriterImpl package-protected ctor is unused (blockWithPath too) except by addRowGroup above

Some followup we should do:
* Let ParquetInstructions offer guidance on precision and scale for BigDecimal columns
* Consider support for BigInteger, as decimals with scale 1(?), with metadata to indicate they are not actually decimals
* Look into DHE's ParquetTableWriter for grouping in tables, otherwise perhaps delete ParquetTableWriter.DestinationInfo

Other notes from Andy
* Check tests for bigdecimal vs serializable (that is, BigDecimal being encoded as a java serializable, rather than with the BigDecimal codec), ensure it always writes as bigdecimal - this might have been a bug introduced while working on adding BigInteger to begin with

niloc132:
cc @jcferretti @rcaudy @abaranec

rcaudy:
@abaranec I don't think we want to port changes to `ColumnChunkPageStore`. These were to accommodate compound regions (with binary search by offset to find rows) rather than the nested address-space division community uses, right?

rcaudy:
> @abaranec I don't think we want to port changes to `ColumnChunkPageStore`. These were to accommodate compound regions (with binary search by offset to find rows) rather than the nested address-space division community uses, right?

Let the record show that Andy answered yes in Slack.

Issue #2118, opened 2022-03-24
Investigate, remove platform-specific code 


niloc132:
We have several dependencies that add native code to the platform, with varying degrees of compatibility over our supported platforms.

## Parquet compression formats

The ParquetHadoop project brings in three runtime-only dependencies to add native support for the codecs we consider important, but does so with native libraries. Ostensibly this is to take advantage of full performance right out of the box, rather than using pure-java implementations that must wait for the JIT to warm up first.

### `com.hadoop.gplcompression:hadoop-lzo:0.4.20` for LZO codec support
We resolve this dependency from https://maven.twttr.com, Twitter's own maven repository, so any project that depends on ParquetHadoop must add this maven repo server to their project. The project appears to only add native support for decoding LZO on Linux on amd64, so no WIndows/Mac, nor Linux on arm64 (including Mac running docker).

Additionally this dependency appears to be GPL3, we should make sure this is safe for us to include.

### `org.lz4:lz4-java:1.8.0` for lz4 support
This project is in Maven Central, support mac/win/linux, and a variety of platforms all from the same jar (only windows is limited to amd64, the others support x86_64 and aarch64).

### `com.github.rdblue:brotli-codec:0.1.1` for brotli support
We must resolve this dependency from https://jitpack.io, as it isn't found in maven central, so any project that depends on ParquetHadoop must add this maven repo server to their project. This dependency includes a small amount of code, but mostly repackages three other dependencies into one jar:
  * 'org.meteogroup.jbrotli:jbrotli:0.5.0'
  * 'org.meteogroup.jbrotli:jbrotli-native-linux-x86-amd64:0.5.0'
  * 'org.meteogroup.jbrotli:jbrotli-native-darwin-x86-amd64:0.5.0'

A quick bit of searching suggests that these jars are not hosted any more in any typical maven repos (not in jitpack, the github repo https://github.com/MeteoGroup/jbrotli suggests that bintray should have it, but not present there either, and not present in central). It appears that when the MeteoGroup/jbrotli repo was made, it had support for more OS/arch pairs, but the distribution we are using only supports mac/linux on amd64, so brotli won't work on windows or any arm platform.

## `netty-transport-native-unix-common` and `netty-transport-native-kqueue`/`-epoll`
Unlike the above, this is native code that is included by arrow's flight, apparently for more efficient netty transport. I have not traced it to see how it is actually used, as deephaven-core itself doesn't ever use it, but any downstream maven projects will apparently need to.

[...]

PR #2376, opened 2022-05-10
Parquet cleanup to match DHE

niloc132:
Prerequisite for #2118, which will modify some of these classes.

Partial #2375


extensions/parquet/base/src/main/java/io/deephaven/parquet/base/PlainFixedLenChunkedWriter.java
```diff
@@ -17,12 +17,10 @@
 import java.nio.IntBuffer;
 import java.util.Arrays;
 
-
```
rcaudy, line 20: Same note as on PlainBinaryChunkedWriter. This is not replicated.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/PlainIntChunkedWriter.java
```diff
@@ -9,30 +9,30 @@
...
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
-import java.nio.*;
+import java.nio.IntBuffer;
 
 
 /**
- * Plain encoding except for booleans
+ * Plain encoding except for ints
+ * TODO see if we can replicate from this, dhe seems to
```
rcaudy, line 17: The first line of the comment is nonsense.
The second line is maybe unnecessary. We replicate this with `io.deephaven.replicators.ReplicateParquetChunkedWriters`, which is run from `:replication-static replicateParquetChunkedWriters`, apparently followed by manual fixups.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/PlainIntChunkedWriter.java
```diff
@@ -9,30 +9,30 @@
...
     private final int originalLimit;
 
-    private IntBuffer targetBuffer;
-    private ByteBuffer innerBuffer;
+    private final IntBuffer targetBuffer;
+    private final ByteBuffer innerBuffer;
 
     PlainIntChunkedWriter(int pageSize, ByteBufferAllocator allocator) {
-        innerBuffer = allocator.allocate(pageSize);
+        this.innerBuffer = allocator.allocate(pageSize);
```
rcaudy, line 27: Why are we making this code worse with unnecessary `this.` prefixes?
If you wanted, you could put the `this.allocator` assignment first, which would make the subsequent derived initializations look cleaner.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/BigDecimalParquetBytesCodec.java
```diff
@@ -27,7 +27,7 @@ public class BigDecimalParquetBytesCodec implements ObjectCodec<BigDecimal> {
      *        precision.
      */
     public BigDecimalParquetBytesCodec(final int precision, final int scale, final int encodedSizeInBytes,
-            final RoundingMode roundingMode) {
+                                       final RoundingMode roundingMode) {
```
rcaudy, line 30: I'm pretty sure spotless will put this back the way it was.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -92,11 +92,11 @@ public static final class DestinationInfo {
 
         private final String outputPath;
         private final Table inputTable;
-        private final Map<String, Map<?, long[]>> columnNameToGroupToRange;
+        private final Map<String, Map<?, RowSet>> columnNameToGroupToRange;
```
rcaudy, line 95: I understand why DHE changed this. This is unused code for us, it's tied closely to `MergeData`, and should probably just be deleted.

rcaudy (commented):
Some commentary.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileWriter.java
```diff
@@ -27,13 +28,14 @@
...
 public class ParquetFileWriter {
-    private static ParquetMetadataConverter metadataConverter = new ParquetMetadataConverter();
+    private static final ParquetMetadataConverter metadataConverter = new ParquetMetadataConverter();
     private static final int VERSION = 1;
 
     private final SeekableByteChannel writeChannel;
     private final MessageType type;
     private final int pageSize;
     private final ByteBufferAllocator allocator;
+    @ReferentialIntegrity
```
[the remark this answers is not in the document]
niloc132: Cloned from DHE - we will need it (...as much as dhe does) if we remove the unused addRowGroup method

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/PlainBinaryChunkedWriter.java
```diff
@@ -9,21 +9,17 @@
 import org.apache.parquet.column.Encoding;
 import org.apache.parquet.column.values.rle.RunLengthBitPackingHybridEncoder;
 import org.apache.parquet.io.api.Binary;
-import org.slf4j.Logger;
```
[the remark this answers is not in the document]
niloc132: We have no replicate for this file - interestingly, dhe _does_, but has no autogenerated header. I'm going to see if we can get the file to actually be replicated, but in the meantime since dhe has a replicate tool and had this change (unused logger, unused null count) it seemed to make sense to prune.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/PlainIntChunkedWriter.java
niloc132: This might be my attempt at keeping track what is happening in code that I don't understand, I'll remove these notes.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/PlainIntChunkedWriter.java
niloc132: Because the more similar the code is between dhc and dhe, the easier cherry-picking will be - that's the purpose of the commit.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/BigDecimalParquetBytesCodec.java
niloc132: See the second commit in this branch - the first was "transfer changes from dhe", the second was "run spotless"

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
niloc132: Yep, see the issue for this, there is a lot of dead code in here, but I haven't been as aggressive about removing things, esp stuff that I understood (i.e. "I interpreted, possibly wrongly, from discussions with andy and/or cristian") might be important later. Someone who knows the code at all, and has some idea of future plans/intentions should take a closer look at how deeply we can prune apparently unused code.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileWriter.java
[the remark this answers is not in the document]
rcaudy: I don't see any reason in the DHC codebase why we would keep that reference if we delete the unused `addRowGroup` method. It's not needed for referential integrity of anything I can find.
I slightly want to annotate that method `@SuppressWarnings("unused")`, rather than delete it. It's illustrating a pattern we don't use right now, but might consider some day.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/PlainBinaryChunkedWriter.java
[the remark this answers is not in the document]
rcaudy: Pruning is fine. Pruning the false header comment should happen, too. I suspect pretty strongly that they (DHE) have not gone in and written a replicator for this file. We have one that works (with supervision) for the numeric types, not the "special" types (see https://github.com/deephaven/deephaven-core/pull/2376#discussion_r869765838).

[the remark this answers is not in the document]
rcaudy: The version of DHE code I'm looking at does not have replication for this, Boolean, or FixedLen in `io.deephaven.replicate.parquet.Replicate`.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/PlainIntChunkedWriter.java
rcaudy: It's not worth it. Cherry-picking is beyond impossible for most of the code base already. We show our code to outsiders. Better to clean this up and suggest DHE make the same change.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
rcaudy: This is a case where my confidence level in our deletion is high. If this became important, we would be porting `MergeData` , and could bring it back over then. We'd be more likely to write our own merge tool that leverages parquet functionality better.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/PlainBinaryChunkedWriter.java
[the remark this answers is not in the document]
niloc132: Sorry, I misunderstood initially - you are commenting on this line, but you actually want the header to be removed for 3 files (plainbinary, boolean, and fixedlen), right?

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/PlainIntChunkedWriter.java
niloc132: ok, i'll include this in my followup patch for dhe

rcaudy (approved):

PR #2384, opened 2022-05-13
Wrap LivenessScope for advanced resource cleanup

### LINKED ISSUE #2250: Wrap livenessScope in deephaven V2
```
Deephaven's Liveness Scoping allows the "liveness" (whether or not they are active) of nodes in a refreshing query's update propagation graph to be updated proactively, rather than only via actions of the Java garbage collector. This does not replace garbage collection (GC), but it does allow cleanup to happen immediately when objects in the GUI are not needed. This is accomplished internally via reference counting, and works automatically for all users.
```

jmao-denver:
Fixes #2250 

1. Added a context manager that returns a scope to be used in the enclosed code block. This also matches the Java use pattern.
2. Cleaned up the Java test code and added Python tests.

### DIFF
--- a/py/server/tests/test_liveness.py
+++ b/py/server/tests/test_liveness.py
@@ -1,0 +1,60 @@
+#
+#     Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+#
+
+import unittest
+
+from deephaven.pandas import to_pandas
+
+from deephaven import time_table
+
+from deephaven._ugp import ugp_exclusive_lock
+from deephaven.liveness_scope import liveness_scope
+from tests.testbase import BaseTestCase
+
+
+def create_table():
+    with ugp_exclusive_lock():
+        return time_table("00:00:00.001").update(["X=i%11"]).sort("X").tail(16)
+
+
+class LivenessTestCase(BaseTestCase):
+    def test_liveness(self):
+        not_managed = create_table()
+        with liveness_scope() as l_scope:
+            to_discard = create_table()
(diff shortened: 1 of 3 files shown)


py/server/tests/test_liveness.py
```diff
@@ -0,0 +1,60 @@
...
+
+            with liveness_scope() as nested_l_scope:
+                nested_to_discard = create_table()
+                df = to_pandas(nested_to_discard)
+                nested_must_keep = create_table()
+                df = to_pandas(nested_must_keep)
+                nested_l_scope.preserve(nested_must_keep)
+            self.assertTrue(nested_must_keep.j_table.tryRetainReference())
+            self.assertFalse(nested_to_discard.j_table.tryRetainReference())
+            nested_must_keep.j_table.dropReference()
```
rcaudy, line 51: This should be unnecessary.

jmao-denver: I didn't have that at first. But the call to tryRetainReference() on line #49 actually increased the count which led to the assert after the outer scope exits to fail. Basically I want to verify that the 'preserve' is only valid for the enclosing parent scope.

rcaudy: You're right. Might be worth a comment to that effect (drop the extra reference acquired when testing within the block).

rcaudy (approved):

chipkent (changes requested):

chipkent (approved):

PR #2396, opened 2022-05-17
Implement Open Addressing to support AsOfJoin (Review Only)

### LINKED ISSUE #2381: Migrate Table.aj methods to use the same type of open-addressed hash table design as aggregations
As a user, I would appreciate it if natural join could be faster and simpler. The new open-addressed hash table design used for aggregations and natural join has helped the hashing phase for those operations significantly. We should pursue similar changes for as-of join, with benchmarking to show the results.

lbooker42:
Will close #2381


engine/table/src/main/java/io/deephaven/engine/table/impl/AsOfJoinHelper.java
```diff
@@ -157,8 +168,25 @@ private static Table rightStaticAj(JoinControl control,
...
+                            leftSources, originalLeftSources,
+                            control.tableSizeForLeftBuild(leftTable),
+                            control.getMaximumLoadFactor(), control.getTargetLoadFactor())
+                    : new StaticChunkedAsOfJoinStateManager(leftSources,
+                            size, originalLeftSources);
+        } else {
+            asOfJoinStateManager = USE_TYPED_STATE_MANAGER
+                    ? TypedHasherFactory.make(StaticAsOfJoinStateManagerTypedBase.class,
+                            leftSources, originalLeftSources,
+                            control.tableSizeForRightBuild(rightTable),
```
cpwright, line 184: Probably just set this to a variable,  so there is less manual diffing to do in order to understand the only difference is the size.

engine/table/src/main/java/io/deephaven/engine/table/impl/StaticAsOfJoinStateManager.java
```diff
@@ -0,0 +1,93 @@
...
+    public static final long NO_RIGHT_ENTRY_VALUE = RowSequence.NULL_ROW_KEY;
+
+    protected final ColumnSource<?>[] keySourcesForErrorMessages;
+
+    protected StaticAsOfJoinStateManager(ColumnSource<?>[] keySourcesForErrorMessages) {
+        this.keySourcesForErrorMessages = keySourcesForErrorMessages;
+    }
+
+    @SuppressWarnings("WeakerAccess")
+    public void checkExactMatch(boolean exactMatch, long leftKeyIndex, long rightSide) {
```
cpwright, line 30: I think this is unused for aj.

engine/table/src/main/java/io/deephaven/engine/table/impl/StaticAsOfJoinStateManager.java
```diff
@@ -0,0 +1,93 @@
...
+
+    // produce a pretty key for error messages
+    protected String extractKeyStringFromSourceTable(long leftKey) {
+        if (keySourcesForErrorMessages.length == 1) {
+            return Objects.toString(keySourcesForErrorMessages[0].get(leftKey));
+        }
+        return "[" + Arrays.stream(keySourcesForErrorMessages).map(ls -> Objects.toString(ls.get(leftKey))).collect(Collectors.joining(", ")) + "]";
+    }
+
+    public WritableRowRedirection buildRowRedirection(QueryTable leftTable, boolean exactMatch, LongUnaryOperator rightSideFromSlot, JoinControl.RedirectionType redirectionType) {
```
cpwright, line 44: Should this be static and shared?

engine/table/src/main/java/io/deephaven/engine/table/impl/asofjoin/StaticAsOfJoinStateManagerTypedBase.java
```diff
@@ -0,0 +1,390 @@
...
+
+    // the table will be rehashed to a load factor of targetLoadFactor if our loadFactor exceeds maximumLoadFactor
+    // or if it falls below minimum load factor we will instead contract the table
+    private final double maximumLoadFactor;
+
+    // the keys for our hash entries
+    protected final ChunkType[] chunkTypes;
+    protected final WritableColumnSource[] mainKeySources;
+
+    protected final ImmutableObjectArraySource<RowSetBuilderSequential> leftRowSetSource; // use immutable
```
cpwright, line 41: comment unnecessary.  If we want a comment it would be describing which is the "primary" thing that determines whether the state is valid and what the sentinel values (if any are)

engine/table/src/main/java/io/deephaven/engine/table/impl/asofjoin/StaticAsOfJoinStateManagerTypedBase.java
```diff
@@ -0,0 +1,390 @@
...
+    /**
+     * Returns true if this is the first left row key added to this slot.
+     */
+    protected boolean addLeftIndex(long tableLocation, long keyToAdd) {
+        return addIndex(leftRowSetSource, tableLocation, keyToAdd);
+    }
+
+    protected void addRightIndex(long tableLocation, long keyToAdd) {
+        // noinspection unchecked
+        addIndex((ImmutableObjectArraySource) rightRowSetSource, tableLocation, keyToAdd);
```
cpwright, line 120: why the cast?

engine/table/src/main/java/io/deephaven/engine/table/impl/asofjoin/StaticAsOfJoinStateManagerTypedBase.java
```diff
@@ -0,0 +1,390 @@
...
+            final TypedHasherUtil.BuildHandler buildHandler) {
+        try (final RowSequence.Iterator rsIt = buildRows.getRowSequenceIterator()) {
+            // noinspection unchecked
+            final Chunk<Values>[] sourceKeyChunks = new Chunk[buildSources.length];
+
+            while (rsIt.hasMore()) {
+                final RowSequence chunkOk = rsIt.getNextRowSequenceWithLength(bc.chunkSize);
+                final int nextChunkSize = chunkOk.intSize();
+
+                if (exceedsCapacity(nextChunkSize)) {
```
cpwright, line 136: I think this is going to be incorrect.  We want to allow rehashing for aj, because we have no idea how many buckets will actually exist.  we might have only a few dozen buckets in a 100MM row table.  naturalJoin has different behavior because it is constrained to be 1-1 on the right side

engine/table/src/main/java/io/deephaven/engine/table/impl/asofjoin/StaticAsOfJoinStateManagerTypedBase.java
```diff
@@ -0,0 +1,390 @@
...
+                }
+
+                handler.doProbe(chunkOk, sourceKeyChunks);
+
+                pc.resetSharedContexts();
+            }
+        }
+    }
+
+    private class StaticProbeHandler implements TypedHasherUtil.ProbeHandler {
```
cpwright, line 225: The two problems look different enough that you could have two classes pretty easily. One of the advantages of the javapoet approach is we can have as many probe/build functions as we want.  The other code had to rely on a single function with a bunch of flags, because we were just copying the file and replacing regions.  We can be cleaner in this version.

For the build, the parameters are identical so I can see why you woudl want to use the isLeft; but here I think we should change it.

engine/table/src/main/java/io/deephaven/engine/table/impl/asofjoin/StaticAsOfJoinStateManagerTypedBase.java
```diff
@@ -0,0 +1,390 @@
...
+    @Override
+    public RowSet getRightIndex(long slot) {
+        if (rightBuildersConverted) {
+            return (RowSet) rightRowSetSource.getUnsafe(slot);
+        }
+        RowSetBuilderSequential builder = (RowSetBuilderSequential) rightRowSetSource.getUnsafe(slot);
+        if (builder == null) {
+            return null;
+        }
+        return builder.build();
```
cpwright, line 321: I don't think a builder can give you an index more than once; so it might actually be the right move to blow this up; and make sure the caller has already done convertRightBuildersToIndex.

engine/table/src/main/java/io/deephaven/engine/table/impl/asofjoin/StaticHashedAsOfJoinStateManager.java
```diff
@@ -0,0 +1,109 @@
...
+                    }
+                }
+
+                return rowRedirection;
+            }
+        }
+        throw new IllegalStateException("Bad redirectionType: " + redirectionType);
+    }
+
+    public void errorOnDuplicates(LongArraySource leftHashSlots, long size, LongUnaryOperator groupPositionToRightSide, LongUnaryOperator firstLeftKey) {
```
cpwright, line 93: This is NJ cruft left over.

engine/table/src/main/java/io/deephaven/engine/table/impl/asofjoin/TypedAsOfJoinFactory.java
```diff
@@ -0,0 +1,409 @@
...
+    public static void staticProbeDecorateLeftFound(HasherConfig<?> hasherConfig, boolean alternate,
+            CodeBlock.Builder builder) {
+        builder.addStatement("final long indexKey = rowKeyChunk.get(chunkPosition)");
+        builder.beginControlFlow("if (addLeftIndex(tableLocation, indexKey) && hashSlots != null)");
+        builder.addStatement("hashSlots.set(hashSlotOffset.getAndIncrement(), tableLocation)");
+        builder.addStatement("foundBuilder.addKey(indexKey)");
+        builder.endControlFlow();
+    }
+
+    public static void staticProbeDecorateLeftMissing(CodeBlock.Builder builder) {
```
cpwright, line 48: We can specify null here (I think) to avoid the block being generated; or at least kill the unused comment.

engine/table/src/main/java/io/deephaven/engine/table/impl/asofjoin/TypedAsOfJoinFactory.java
```diff
@@ -0,0 +1,409 @@
...
+        // builder.addStatement("leftRedirections.set(hashSlotOffset++, $T.NULL_ROW_KEY)", RowSet.class);
+    }
+
+    public static void staticProbeDecorateRightFound(HasherConfig<?> hasherConfig, boolean alternate,
+            CodeBlock.Builder builder) {
+        builder.addStatement("addRightIndex(tableLocation, rowKeyChunk.get(chunkPosition))");
+    }
+
+    public static void rightIncrementalRehashSetup(CodeBlock.Builder builder) {
```
cpwright, line 57: The rest of this is cruft; but since you are going to do incremental maybe you wanted to leave it?

engine/table/src/main/java/io/deephaven/engine/table/impl/by/typed/TypedHasherFactory.java
```diff
@@ -255,6 +260,33 @@ public static <T> HasherConfig<T> hasherConfigForBase(Class<T> baseClass) {
                     TypedNaturalJoinFactory::incrementalShiftLeftMissing,
                     ParameterSpec.builder(long.class, "shiftDelta").build(),
                     probeContextParam));
+        } else if (baseClass.equals(StaticAsOfJoinStateManagerTypedBase.class)) {
+            builder.classPrefix("StaticAsOfJoinHasher").packageGroup("asofjoin").packageMiddle("staticopen")
+                    .openAddressedAlternate(false)
+                    .stateType(Object.class).mainStateName("rightRowSetSource")
+                    .emptyStateName("EMPTY_RIGHT_STATE")
+                    .includeOriginalSources(true)
+                    .supportRehash(false);
```
cpwright, line 269: We'll need rehash for aj().

replication/reflective/src/main/java/io/deephaven/replicators/ReplicateTypedHashers.java
```diff
@@ -24,13 +25,14 @@
...
 public class ReplicateTypedHashers {
     public static void main(String[] args) throws IOException {
-        generatePackage(StaticChunkedOperatorAggregationStateManagerTypedBase.class, true);
-        generatePackage(StaticChunkedOperatorAggregationStateManagerOpenAddressedBase.class, true);
-        generatePackage(IncrementalChunkedOperatorAggregationStateManagerTypedBase.class, true);
-        generatePackage(IncrementalChunkedOperatorAggregationStateManagerOpenAddressedBase.class, true);
-        generatePackage(StaticNaturalJoinStateManagerTypedBase.class, false);
-        generatePackage(RightIncrementalNaturalJoinStateManagerTypedBase.class, false);
-        generatePackage(IncrementalNaturalJoinStateManagerTypedBase.class, false);
+        // generatePackage(StaticChunkedOperatorAggregationStateManagerTypedBase.class, true);
```
cpwright, line 28: Will need to be put back before merging.

engine/table/src/test/java/io/deephaven/engine/table/impl/QueryTableAjTest.java
```diff
@@ -524,6 +524,11 @@ public void testAjRandomStaticOverflow() {
 
     @SuppressWarnings("SameParameterValue")
     private void testAjRandomStaticOverflow(int seed, int leftSize, int rightSize) {
+        if (AsOfJoinHelper.USE_TYPED_STATE_MANAGER) {
```
cpwright, line 527: aj() will need to rehash; though we should not use a load factor greater than 0.75.  It is probably the case that we will want this test to cause rehashing if we do not have another suitable one.

engine/table/src/main/java/io/deephaven/engine/table/impl/asofjoin/StaticAsOfJoinStateManagerTypedBase.java
lbooker42: It seems to be needed, chunked version has this:
`    private void addRightIndex(long tableLocation, long keyToAdd) {
        //noinspection unchecked
        addIndex((ObjectArraySource) rightRowSetSource, tableLocation, keyToAdd);
    }`

engine/table/src/main/java/io/deephaven/engine/table/impl/StaticAsOfJoinStateManager.java
lbooker42: Is not used and has been removed.

engine/table/src/main/java/io/deephaven/engine/table/impl/asofjoin/StaticHashedAsOfJoinStateManager.java
lbooker42: agreed, removed.

engine/table/src/test/java/io/deephaven/engine/table/impl/QueryTableAjTest.java
lbooker42: restored this as it properly tests the OA rehashing while building

engine/table/src/main/java/io/deephaven/engine/table/impl/by/typed/TypedHasherFactory.java
```diff
@@ -776,6 +849,95 @@ private static MethodSpec createRehashInternalFullMethod(HasherConfig<?> hasherC
                 .addAnnotation(Override.class).build();
     }
 
+    // leverage the slots array and size to more effi
```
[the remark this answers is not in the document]
cpwright: Where is the usage?  Why do you need this instead of a move function?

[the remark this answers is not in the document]
lbooker42: I was intending to write a rehashing version of the factory code that supported leftSlot rehashing.  This is abandoned code, pending cleanup

engine/table/src/main/java/io/deephaven/engine/table/impl/by/typed/TypedHasherFactory.java
```diff
@@ -296,70 +290,37 @@ public static <T> HasherConfig<T> hasherConfigForBase(Class<T> baseClass) {
                     true, TypedAsOfJoinFactory::staticProbeDecorateRightFound, null));
 
         } else if (baseClass.equals(RightIncrementalAsOfJoinStateManagerTypedBase.class)) {
-            final ParameterSpec modifiedSlotTrackerParam =
-                    ParameterSpec.builder(NaturalJoinModifiedSlotTracker.class, "modifiedSlotTracker").build();
+//            final ParameterSpec modifiedSlotTrackerParam =
```
cpwright, line 293: Dead code.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/typed/TypedHasherFactory.java
```diff
@@ -296,70 +290,37 @@ public static <T> HasherConfig<T> hasherConfigForBase(Class<T> baseClass) {
...
-                    .stateType(long.class).mainStateName("leftRowSetSource")
-                    .overflowOrAlternateStateName("alternateLeftRowSetSource")
-                    .emptyStateName("EMPTY_RIGHT_STATE")
+                    .stateType(byte.class).mainStateName("stateSource")
+                    .overflowOrAlternateStateName("alternateStateSource")
+                    .emptyStateName("ENTRY_EMPTY_STATE")
                     .includeOriginalSources(true)
                     .supportRehash(true)
-                    .addExtraPartialRehashParameter(modifiedSlotTrackerParam)
+//                    .addExtraPartialRehashParameter(modifiedSlotTrackerParam)
```
cpwright, line 308: dead code

cpwright (approved):

PR #2406, opened 2022-05-19
Rewrite JS/GWT to use row counts (vs. row batches)

### LINKED ISSUE #2405: Phantom barrage rows
It looks like #2287 has introduced a bug into barrage. At a minimum, it causes streaming time tables to produce phantom rows in the web UI.

```groovy
x = io.deephaven.engine.table.impl.TimeTable.newBuilder().period("00:00:01").streamTable(true).build()
```

Upon creation, the table occasionally ticks and produces a phantom row.

image

Upon re-viewporting, the data is initially "correct", but then phantom rows creep in.

lbooker42:
Optimize snapshots for large tables (#2365) was merged and it removed the guaranteed accuracy of the `num_add_batches` / `num_mod_batches` provided by the `BarrageUpdateMetadata` message.  These became an estimate (as a temporary measure since these metadata tags have been removed in barrage repo release 0.5.0).  This update rewrites GWT `BarrageUtils.java` to use row counts from `rowsIncluded` and `rowsModified` metadata instead of number of batches.

Expected to fix #2405


web/client-api/src/main/java/io/deephaven/web/client/api/barrage/BarrageUtils.java
```diff
@@ -171,7 +171,8 @@ public static class DeltaUpdatesBuilder {
         private final DeltaUpdates deltaUpdates = new DeltaUpdates();
         private final BarrageUpdateMetadata barrageUpdate;
         private final String[] columnTypes;
-        private int recordBatchesSeen = 0;
+        private long numAddRowsRemaining = 0;
+        private long numModRowsRemaining = 0;
```
devinrsmith, line 175: I might be missing something - but it looks like this value is set once in the constructor; but never decremented? But we have a condition later on it `== 0`? Seems like that will never get hit?

lbooker42: You're correct.  When porting, missed a critical line decrementing `numModRowsRemaining`. Please see the change below (line 264)

devinrsmith: I'm sad we don't have more thorough automated tests for producer -> barrage -> consumer; assuming it would be better at catching something like this. I'll let @nbauernfeind do deeper look into logic.

web/client-api/src/main/java/io/deephaven/web/client/api/barrage/BarrageUtils.java
```diff
@@ -195,23 +196,30 @@ public DeltaUpdatesBuilder(BarrageUpdateMetadata barrageUpdate, boolean isViewpo
...
+            if (numAddRowsRemaining > 0) {
                 handleAddBatch(recordBatch, body);
-            } else {
+            } else if (numModRowsRemaining > 0) {
                 handleModBatch(recordBatch, body);
             }
-            recordBatchesSeen++;
-            return recordBatchesSeen == barrageUpdate.numAddBatches() + barrageUpdate.numModBatches();
+            // return true when complete
+            return numAddRowsRemaining == 0 && numModRowsRemaining == 0;
```
devinrsmith, line 222: Is it even possible to have < 0? I wonder if we should prefer to throw IllegalStateException if we see that.

web/client-api/src/main/java/io/deephaven/web/client/api/barrage/BarrageUtils.java
```diff
@@ -229,6 +237,7 @@ private void handleAddBatch(RecordBatch recordBatch, ByteBuffer body) {
                 addedColumnData[columnIndex] = new DeltaUpdates.ColumnAdditions(columnIndex, columnData);
             }
             deltaUpdates.setSerializedAdditions(addedColumnData);
+            numAddRowsRemaining -= (long) recordBatch.length().toFloat64();
```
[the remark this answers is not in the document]
nbauernfeind: This is "the valid" thing to do, here.

nbauernfeind (approved):
I think this is fine. It's worth noting that the implementation does not technically support multiple record batches for deltas or snapshots. Deltas overwrite the earlier dataset. Snapshots aren't waiting for future fragments. 

So, the web client will still break with any amount of fragmentation.

PR #2414, opened 2022-05-19
Access Control hooks, incomplete

devinrsmith:
I think we need to approach access control in a generic way - allowing servers to completely take control when necessary.

This patch isn't complete (only sketched out table and console services), but essentially hooks up at the very front-end of every single gRPC call, and passes the request details on to an interface that the user is able to override via ServiceLoader mechanism. If service loader is not generic enough in the future, we can provide further means to modify via dagger modules.

As far as our default impl, there is a question of convention vs configuration. At the moment, I'm finding myself wanting the ability to essentially lock "unsafe" things down for eventually letting the public access tables.

### DIFF
--- a/Util/src/main/java/io/deephaven/util/auth/AuthContext.java
+++ b/Util/src/main/java/io/deephaven/util/auth/AuthContext.java
@@ -47,4 +47,5 @@ public interface AuthContext {
             return "default";
         }
     }
+
 }
--- a/Util/src/main/java/io/deephaven/util/auth/NormalUser.java
+++ b/Util/src/main/java/io/deephaven/util/auth/NormalUser.java
@@ -1,0 +1,16 @@
+package io.deephaven.util.auth;
+
+import org.jetbrains.annotations.NotNull;
+
+public class NormalUser implements AuthContext {
+
+    @NotNull
+    public final String getAuthRoleName() {
+        return "NormalUser";
+    }
+
+    @NotNull
+    public final String getAuthId() {
+        return "normal";
+    }
+}
--- a/docker-compose-common.yml
+++ b/docker-compose-common.yml
@@ -14,10 +14,10 @@ services:
       # with max memory.
       #
       # To turn on debug logging, add: -Dlogback.configurationFile=logback-debug.xml
-      - JAVA_TOOL_OPTIONS=-Xmx4g -Ddeephaven.console.type=${DEEPHAVEN_CONSOLE_TYPE} -Ddeephaven.application.dir=${DEEPHAVEN_APPLICATION_DIR}
+#      - JAVA_TOOL_OPTIONS=-Xmx4g -Ddeephaven.console.type=${DEEPHAVEN_CONSOLE_TYPE} -Ddeephaven.application.dir=${DEEPHAVEN_APPLICATION_DIR}
       #
       # For remote debugging switch the line above for the one below (and also change the ports below)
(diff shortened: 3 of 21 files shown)


niloc132 (changes requested):
In general, I'm not a fan, I think we need explicit design for this before trying it out. I've expressed some specific concerns with the approach (and a few doubts about particulars of the services).

I'm glad to see it being spitballed, but I think we need to plan it out a little more first.

Specific things I think we're missing here:
 * authentication - how do we know if a user is a superuser, aside from "their class name is SuperUser". It doesn't make sense to effectively build a runtime startup flag
 * Auditing will catch up eventually, need hooks for that, or the authn/authz is just informative for the user
 * Generalizing not just for initial call, but for subsequent streamed request. Not applicable in the methods handled in this patch, but still should have some idiom/approach proposed.
 * rate limiting and other measures - it _probably_ isn't sufficient to handle only flat deny to protect from misbehaving clients.
     * tooling for rate limiting streams - if you can do X and the client can stream requests, it isn't enough to rate limit the stream's creation

I don't want to use the code review to think up new ideas, but my first thought was simple codegen to wrap our `*GrpcImpl`, which checked conditions before we directly handled incoming calls, and autogenerated default-allow implementations for testing, local dev, trusted envs. Recall that there is a .proto for protobufs, which allows us _in java_ to read the ast of our own proto files, and can write out our own java to help with that sort of thing - that, combined with something like BrowserStreamInterceptor? Plus specific apis for authn, roles for authz, and some kind of rough attempt at hooks for logging auditing.

nbauernfeind:
As I mentioned offline, something about this feels verbose and cumbersome.

What I had in mind were calls to `AuthContext` hooks like this:
```java
session.getAuthContext().requirePrivilege(CreateConsole);
```

The interface changes to `AuthContext` would be something like this:
```java
interface AuthContext {
  boolean hasPrivilege(Privilege p);
  default void requirePrivilege(Privilege p) {
    if (!hasPrivilege(p)) {
      throw GrpcUtil.statusRuntimeException(Code.UNAUTHENTICATED, "User does not have privilege " + p);
    }
  }
}
```

The `Privilege` would be further developed like:
```java
interface Privilege {
  // likely this is just a token interface
}

enum TableServicePrivilege implements Privilege {
  EMPTY_TABLE,
  TIME_TABLE,
  DROP_COLUMNS,
  UPDATE,
  LAZY_UPDATE,
  VIEW,
  // etc
}
```

Note that using a token interface allows further extension by users without needing to modify a master enum type.

Restricting certain [obvious] categories of privileges is simple:
```java
class RestrictedAuthContext {
  boolean hasPrivilege(Privilege privilege) {
    if (privilege instanceof ConsoleServicePrivilege) {
      return false;
    }
    if (privilege instanceof CodeGenerationPrivilege) {
      // restricting on code generation could enable some subset 
      // of Table.{select, update, view, updateView} etc
      return false;
    }
    return true;
  }
}
```

Or they could be side-loaded from a database or file-based config:
```java
class WhitelistAuthContext {
  private final Set<Privilege> allowedPrivileges;
  boolean hasPrivilege(Privilege privilege) {
      return allowedPrivileges.contains(privilege);
  }
}
```

One could create a `RateLimitingAuthContext` to wrap around a delegate `AuthContext`.

To me, this feels flexible and cross-functional while minimizing maintenance and implementation overhead.

PR #2417, opened 2022-05-20
Add SSLConfig and use in java-client-session

devinrsmith:
Broken out from #2401


ssl/config/gradle.properties
```diff
@@ -0,0 +1,2 @@
+io.deephaven.project.ProjectType=JAVA_PUBLIC
+languageLevel=8
```
niloc132, line 2: why 8? would it make sense to document this - and should ssl/ (and possibly children) get a readme documenting what they are intended to be for etc?

niloc132 (approved):
Is there an issue for this, to continue to track the whole "replace config prop files" thing? https://github.com/deephaven/deephaven-core/issues/21 is kinda related, whatever approach we pick should also solve that, among other problems - but we should itemize the problems to solve, make sure yaml/json/xml/etc handles them all, while we still only have 3 config languages.

I've approved the review, but would like an issue for the above (linked from commit message), and maybe some explicit docs on why this is the way it is in the ssl/ dir, possibly sub dirs too as you think is appropriate.

ssl/config/gradle.properties
devinrsmith: Anything that the java-client depends on needs to be scoped to language level 8.

niloc132 (approved):

Issue #2306, opened 2022-04-20
Python Wrapper Properties include Java static method 


hythloda:
As a python user, I want no confusion about which method does what. 

We should attach a property to every wrapper function to hold the name of the wrapped Java static method.

This will allows for no confusion for the user with the difference between `lower_bin` and `lowerBin` and having to know when to use which.

Something akin to an alias so `theMethod` in Java can be `the_method` when used from python.

This issue was highlighted with @jcferretti from the [Debezium Python V2 rewrite]( https://github.com/deephaven/deephaven-core/pull/2299). As there are two ways for the same task
`['received_at_minute = lowerBin(received_at, minute_in_nanos)']`

`['received_at_minute = (DateTime)lower_bin(received_at, minute_in_nanos)'] `

Linked to #2303

jmao-denver:
A related but different issue is that when both Python and Java have symbols with the same name and both can be used in formula/filter evaluation.

niloc132:
See https://github.com/deephaven/deephaven-core/issues/3830#issuecomment-1593674478

jmao-denver:
made unnecessary by #4388

PR #2422, opened 2022-05-23
Make Python local and global scopes available for query formulas

### LINKED ISSUE #1072: Using Python globals() to add objects for use by the query engine doesn't work for sub modules
**Description**
Because Python globals() only provides access to the enclosing module's global symbols not the main module's, it doesn't work from within any sub modules imported.

**@alexpeters1208 ran into this problem with his learn module and had to revert back to using QueryScope in a hacky way. Once we find a permanent solution, we will need to fix his code.**

jmao-denver:
Fixes #1072

### DIFF
--- a/Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
+++ b/Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
@@ -178,6 +178,10 @@ public class PythonDeephavenSession extends AbstractScriptSession<PythonSnapshot
                 .orElse(defaultValue);
     }
 
+    public void modifyScope(PyObject object) {
+        scope.modifyScope(object);
+    }
+
     @Override
     protected void evaluate(String command, String scriptName) {
         log.info().append("Evaluating command: " + command).endl();
--- a/Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
+++ b/Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
@@ -179,6 +179,9 @@ public class PythonDeephavenSession extends AbstractScriptSession<PythonSnapshot
     }
 
     public void pushScope(PyObject pydict) {
+        if (!pydict.isDict()) {
+            throw new IllegalArgumentException("Expect a Python dict but got a" + pydict.repr());
+        }
         scope.pushScope(pydict);
     }
 
@@ -226,7 +229,7 @@ public class PythonDeephavenSession extends AbstractScriptSession<PythonSnapshot
 
     @Override
     protected PythonSnapshot takeSnapshot() {
-        return new PythonSnapshot(scope.globals().copy());
+        return new PythonSnapshot(scope.mainGlobals().copy());
     }
 
     @Override
@@ -281,7 +284,7 @@ public class PythonDeephavenSession extends AbstractScriptSession<PythonSnapshot
     @Override
     public synchronized void setVariable(String name, @Nullable Object newValue) {
         try (PythonSnapshot fromSnapshot = takeSnapshot()) {
-            final PyDictWrapper globals = scope.globals();
+            final PyDictWrapper globals = scope.mainGlobals();
             if (newValue == null) {
                 try {
                     globals.delItem(name);
--- a/Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
+++ b/Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
@@ -49,7 +49,7 @@ import java.util.stream.Collectors;
 
 /**
  * A ScriptSession that uses a JPy cpython interpreter internally.
- *
+ * <p>
  * This is used for applications or the console; Python code running remotely uses WorkerPythonEnvironment for it's
  * supporting structures.
  */
@@ -178,8 +178,12 @@ public class PythonDeephavenSession extends AbstractScriptSession<PythonSnapshot
                 .orElse(defaultValue);
     }
 
-    public void modifyScope(PyObject object) {
-        scope.modifyScope(object);
+    public void pushScope(PyObject pydict) {
+        scope.pushScope(pydict);
+    }
+
+    public void popScope() {
+        scope.popScope();
     }
 
     @Override
--- a/engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
@@ -327,7 +327,7 @@ public abstract class AbstractScriptSession<S extends AbstractScriptSession.Snap
         }
 
         public ScriptSession scriptSession() {
-            return this.scriptSession;
+            return scriptSession;
         }
     }
 
--- a/engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
@@ -325,6 +325,10 @@ public abstract class AbstractScriptSession<S extends AbstractScriptSession.Snap
         public void putObjectFields(Object object) {
             throw new UnsupportedOperationException();
         }
+
+        public ScriptSession scriptSession() {
+            return this.scriptSession;
+        }
     }
 
     public static class UnsynchronizedScriptSessionQueryScope extends ScriptSessionQueryScope {
--- a/engine/table/src/main/java/io/deephaven/engine/util/PythonScope.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/PythonScope.java
@@ -170,7 +170,7 @@ public interface PythonScope<PyObj> {
          */
     }
 
-    public PyDictWrapper globals();
+    public PyDictWrapper mainGlobals();
 
     /**
      * Push the provided Python scope into the thread scope stack for subsequent operations on Tables
--- a/engine/table/src/main/java/io/deephaven/engine/util/PythonScope.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/PythonScope.java
@@ -172,5 +172,16 @@ public interface PythonScope<PyObj> {
 
     public PyDictWrapper globals();
 
-    void modifyScope(PyObject object);
+    /**
+     * Push the provided Python scope into the thread scope stack for subsequent operations on Tables
+     *
+     * @param pydict a Python dictionary representing the current scope under which the Python code in running, it is
+     *        the combination of module globals and function locals
+     */
+    void pushScope(PyObject pydict);
+
+    /**
+     * Pop the last Python scope pushed by {@link #pushScope(PyObject pydict)} from the thread scope stack
+     */
+    void popScope();
 }
--- a/engine/table/src/main/java/io/deephaven/engine/util/PythonScope.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/PythonScope.java
@@ -170,6 +170,9 @@ public interface PythonScope<PyObj> {
          */
     }
 
+    /**
+     * @return the Python's __main__ module namespace
+     */
     public PyDictWrapper mainGlobals();
 
     /**
--- a/engine/table/src/main/java/io/deephaven/engine/util/PythonScope.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/PythonScope.java
@@ -4,6 +4,7 @@
 package io.deephaven.engine.util;
 
 import org.jpy.PyDictWrapper;
+import org.jpy.PyObject;
 
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.Collection;
@@ -170,4 +171,6 @@ public interface PythonScope<PyObj> {
     }
 
     public PyDictWrapper globals();
+
+    void modifyScope(PyObject object);
 }
--- a/engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
@@ -15,7 +15,7 @@ import java.util.stream.Stream;
 public class PythonScopeJpyImpl implements PythonScope<PyObject> {
     private final PyDictWrapper dict;
 
-    private final ThreadLocal<PyDictWrapper> moduleDict = new ThreadLocal<>();
+    private static final ThreadLocal<Deque<PyDictWrapper>> threadScopeStack = new ThreadLocal<>();
     private static final PyObject NUMBA_VECTORIZED_FUNC_TYPE = getNumbaVectorizedFuncType();
 
     // this assumes that the Python interpreter won't be re-initialized during a session, if this turns out to be a
@@ -37,11 +37,11 @@ public class PythonScopeJpyImpl implements PythonScope<PyObject> {
     }
 
     private PyDictWrapper currentScope() {
-        PyDictWrapper dict = moduleDict.get();
-        if (dict == null) {
+        Deque<PyDictWrapper> scopeStack = threadScopeStack.get();
+        if (scopeStack == null || scopeStack.isEmpty()) {
             return this.dict;
         } else {
-            return dict;
+            return scopeStack.peek();
         }
     }
 
@@ -213,10 +213,20 @@ public class PythonScopeJpyImpl implements PythonScope<PyObject> {
     }
 
     @Override
-    public void modifyScope(PyObject pydict) {
-        if (pydict.isNone())
-            moduleDict.set(null);
-        else
-            moduleDict.set(pydict.asDict());
+    public void pushScope(PyObject pydict) {
+        Deque<PyDictWrapper> scopeStack = threadScopeStack.get();
+        if (scopeStack == null) {
+            scopeStack = new ArrayDeque<>();
+            threadScopeStack.set(scopeStack);
+        }
+        scopeStack.push(pydict.asDict());
+    }
+
+    @Override
+    public void popScope() {
+        Deque<PyDictWrapper> scopeStack = threadScopeStack.get();
+        if (scopeStack != null) {
+            scopeStack.pop();
+        }
     }
 }
--- a/engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
@@ -14,6 +14,8 @@ import java.util.stream.Stream;
 
 public class PythonScopeJpyImpl implements PythonScope<PyObject> {
     private final PyDictWrapper dict;
+
+    private final ThreadLocal<PyDictWrapper> moduleDict = new ThreadLocal<>();
     private static final PyObject NUMBA_VECTORIZED_FUNC_TYPE = getNumbaVectorizedFuncType();
 
     // this assumes that the Python interpreter won't be re-initialized during a session, if this turns out to be a
@@ -34,26 +36,35 @@ public class PythonScopeJpyImpl implements PythonScope<PyObject> {
         this.dict = dict;
     }
 
+    private PyDictWrapper currentScope() {
+        PyDictWrapper dict = moduleDict.get();
+        if (dict == null) {
+            return this.dict;
+        } else {
+            return dict;
+        }
+    }
+
     @Override
     public Optional<PyObject> getValueRaw(String name) {
         // note: we *may* be returning Optional.of(None)
         // None is a valid PyObject, and can be in scope
-        return Optional.ofNullable(dict.get(name));
+        return Optional.ofNullable(currentScope().get(name));
     }
 
     @Override
     public Stream<PyObject> getKeysRaw() {
-        return dict.keySet().stream();
+        return currentScope().keySet().stream();
     }
 
     @Override
     public Stream<Entry<PyObject, PyObject>> getEntriesRaw() {
-        return dict.entrySet().stream();
+        return currentScope().entrySet().stream();
     }
 
     @Override
     public boolean containsKey(String name) {
-        return dict.containsKey(name);
+        return currentScope().containsKey(name);
     }
 
     @Override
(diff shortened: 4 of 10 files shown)


engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -325,6 +325,10 @@ private ScriptSessionQueryScope(ScriptSession scriptSession) {
         public void putObjectFields(Object object) {
             throw new UnsupportedOperationException();
         }
+
+        public ScriptSession scriptSession() {
+            return this.scriptSession;
```
rcaudy, line 330: No need for `this.` in this case.
```suggestion
            return scriptSession;
```

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
```diff
@@ -200,4 +211,12 @@ public static Object convert(PyObject pyObject) {
     public PyDictWrapper globals() {
         return dict;
     }
+
+    @Override
+    public void modifyScope(PyObject pydict) {
+        if (pydict.isNone())
+            moduleDict.set(null);
```
rcaudy, line 218: Could we just simplify this and `currentScope` by just installing `this.dict` as `moduleDict` when `null` is passed?

engine/table/src/main/java/io/deephaven/engine/util/PythonScope.java
```diff
@@ -170,4 +171,6 @@ default Map<String, Object> getEntriesMap() {
     }
 
     public PyDictWrapper globals();
+
+    void modifyScope(PyObject object);
```
rcaudy, line 175: Why not `setScope`? `setScopeDictionary`? `setModuleDict`? Need some JavaDoc.

py/server/deephaven/table.py
```diff
@@ -7,6 +7,8 @@
 
 from __future__ import annotations
 
+import inspect
+import sys
```
rcaudy, line 11: IntelliJ seems to think this is unnecessary.

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
```diff
@@ -200,4 +211,12 @@ public static Object convert(PyObject pyObject) {
     public PyDictWrapper globals() {
```
rcaudy, line 211: I'm a bit nervous about this. `globals()` is what is used to expose named objects to clients... I *think* that's appropriate going forward, but I want to make sure we discuss it. Module dictionaries can be used inside expressions, but putting tables in module dictionaries won't make them visible to clients as-is.

py/server/deephaven/table.py
```diff
@@ -330,6 +352,7 @@ def update(self, formulas: Union[str, Sequence[str]]) -> Table:
         """
         try:
             formulas = to_sequence(formulas)
+            _set_query_scope()
```
rcaudy, line 355: Don't we need this in a lot more places? All of `view`, `select`, `updateView`, `update`, `lazyUpdate`, `where`, `selectDistinct`, at a minimum.

py/server/deephaven/table.py
```diff
@@ -43,6 +45,26 @@
...
+_JQueryScope = jpy.get_type("io.deephaven.engine.table.lang.QueryScope")
+_JUnsynchronizedScriptSessionQueryScope = jpy.get_type(
+    "io.deephaven.engine.util.AbstractScriptSession$UnsynchronizedScriptSessionQueryScope")
+_JPythonScriptSession = jpy.get_type("io.deephaven.integrations.python.PythonDeephavenSession")
+_j_script_session = jpy.cast(_JQueryScope.getScope(), _JUnsynchronizedScriptSessionQueryScope).scriptSession()
+_j_py_script_session = jpy.cast(_j_script_session, _JPythonScriptSession)
+_deephaven_pkg_prefix = __file__[:__file__.index("deephaven") + len("deephaven")]
+
+
+def _set_query_scope():
```
rcaudy, line 58: Shouldn't this be a context manager? I think you'd want to install the module dict for the duration of an op, but not leave it installed after.

rcaudy (commented):
I haven't tried this out, but I have some commentary on the current state.

py/server/deephaven/table.py
```diff
@@ -352,8 +369,7 @@ def update(self, formulas: Union[str, Sequence[str]]) -> Table:
         """
         try:
             formulas = to_sequence(formulas)
-            _set_query_scope()
-            with auto_locking_ctx(self):
+            with auto_locking_ctx(self), _query_scope_ctx():
```
rcaudy, line 372: Is there an implied ordering here? I would prefer to hold the lock for the shortest possible window.

py/server/deephaven/table.py
```diff
@@ -52,18 +52,35 @@
...
-_deephaven_pkg_prefix = __file__[:__file__.index("deephaven") + len("deephaven")]
+_deephaven_pkg_prefix = __file__
 
 
-def _set_query_scope():
-    for frame, filename, line_num, func, source_code, source_index in inspect.getouterframes(inspect.currentframe()):
-        if filename and filename.startswith(_deephaven_pkg_prefix):
-            continue
+@contextlib.contextmanager
+def _query_scope_ctx():
```
rcaudy, line 59: Are we sure this should be internal-only?

py/server/tests/test_pandas.py
```diff
@@ -11,7 +11,7 @@
 from deephaven import dtypes, new_table, DHError
 from deephaven.column import byte_col, char_col, short_col, bool_col, int_col, long_col, float_col, double_col, \
     string_col, datetime_col, pyobj_col, jobj_col
-from deephaven.constants import NULL_LONG, NULL_SHORT, NULL_INT
+from deephaven.constants import NULL_LONG, NULL_SHORT, NULL_INT, NULL_BYTE as NB
```
rcaudy, line 14: Why the `as NB`?

py/server/deephaven/table.py
```diff
@@ -43,6 +45,43 @@
...
+
+@contextlib.contextmanager
+def _query_scope_ctx():
+    """A context manager to set/unset query scope based on the scope of the most immediate application code that invokes
+    Table operations."""
+
+    # locate the innermost Deephaven frame (i.e. any of the table operation methods that use this context manager)
+    outer_frames = inspect.getouterframes(inspect.currentframe())[1:]
+    for i, (frame, filename, *_) in enumerate(outer_frames):
+        if filename and filename == __file__:
```
rcaudy, line 66: Did you intend to use `_deephaven_pkg_prefix` here?

py/server/deephaven/table.py
```diff
@@ -43,6 +45,43 @@
...
+_j_py_script_session = jpy.cast(_j_script_session, _JPythonScriptSession)
+_deephaven_pkg_prefix = __file__
+
+
+@contextlib.contextmanager
+def _query_scope_ctx():
+    """A context manager to set/unset query scope based on the scope of the most immediate application code that invokes
+    Table operations."""
+
+    # locate the innermost Deephaven frame (i.e. any of the table operation methods that use this context manager)
```
rcaudy, line 63: Why do we want the innermost Deephaven frame, rather than the outermost Deephaven-only frame? I'm not sure I understand.

rcaudy (commented):
I'm happy enough with the current state, modulo my questions/comments. On to @chipkent or @devinrsmith

py/server/deephaven/table.py
```diff
@@ -349,7 +388,7 @@ def lazy_update(self, formulas: Union[str, Sequence[str]]) -> Table:
         """
         try:
             formulas = to_sequence(formulas)
-            with auto_locking_ctx(self):
+            with auto_locking_ctx(self), _query_scope_ctx():
```
devinrsmith, line 391: How does globals() work wrt defaults? `def my_method(formulas, context=globals())`? Does it capture `globals()` from the callers perspective, or from the currently defined module?

devinrsmith (commented):
Snazzy. I like the feel of this - would be nice to follow a similar pattern for groovy!

py/server/tests/test_table.py
```diff
@@ -560,6 +560,47 @@ def test_layout_hints(self):
         self.assertTrue(cm.exception.root_cause)
         self.assertIn("RuntimeError", cm.exception.compact_traceback)
 
+    def test_update_local(self):
```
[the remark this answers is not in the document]
chipkent: Should also test what happens when variables change.  e.g.
```python
x = 1
t = emptyTable(1).update("X=x")
assert X==1
x = 2
t = t.update("Y=x")
assert Y==2
```

py/server/deephaven/table.py
```diff
@@ -43,6 +45,43 @@
...
 _JTableOperations = jpy.get_type("io.deephaven.api.TableOperations")
 
+# Dynamic Query Scope
+_JQueryScope = jpy.get_type("io.deephaven.engine.table.lang.QueryScope")
+_JUnsynchronizedScriptSessionQueryScope = jpy.get_type(
+    "io.deephaven.engine.util.AbstractScriptSession$UnsynchronizedScriptSessionQueryScope")
+_JPythonScriptSession = jpy.get_type("io.deephaven.integrations.python.PythonDeephavenSession")
+_j_script_session = jpy.cast(_JQueryScope.getScope(), _JUnsynchronizedScriptSessionQueryScope).scriptSession()
+_j_py_script_session = jpy.cast(_j_script_session, _JPythonScriptSession)
+_deephaven_pkg_prefix = __file__
```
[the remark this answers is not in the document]
chipkent: after reading more and seeing the slick way you've made scope work in functions without users having to explicitly create a scope, do you want a private query scope module?

Are there any cases, other than operations on tables, where query scope can be used?  If not, then maybe it makes sense to have the functionality in this file with `__file__` hard coded.

py/server/deephaven/table.py
chipkent: After reading the tests, it might be totally reasonable for the context to be private, since it looks like @jmao-denver has managed to get function level scope automagically, without users needing to do anything.

py/server/deephaven/table.py
```diff
@@ -43,6 +45,43 @@
...
+
+
+@contextlib.contextmanager
+def _query_scope_ctx():
+    """A context manager to set/unset query scope based on the scope of the most immediate application code that invokes
+    Table operations."""
```
chipkent, line 61: Users aren't going to be able to understand this.  Consider making the description longer, so that it is understandable.

py/server/deephaven/table.py
```diff
@@ -43,6 +45,43 @@
...
+
+    # locate the innermost Deephaven frame (i.e. any of the table operation methods that use this context manager)
+    outer_frames = inspect.getouterframes(inspect.currentframe())[1:]
+    for i, (frame, filename, *_) in enumerate(outer_frames):
+        if filename and filename == __file__:
+            break
+
+    # we deem any module/function/method which calls a wrapped formula/filter-evaluating Table op as an application
+    # frame. This holds true and makes sense even when such module/function/method is part of the Deephaven package, and
+    # should not be a hurdle for developing such code as function/method parameters are now part of the query scope
```
chipkent, line 71: not totally clear.  I had to read it multiple times to get an idea of what it is trying to say.

py/server/deephaven/table.py
```diff
@@ -43,6 +45,43 @@
...
+        scope_dict = app_frame.f_globals.copy()
+        scope_dict.update(app_frame.f_locals)
+        try:
+            _j_py_script_session.pushScope(scope_dict)
+            yield
+        finally:
+            _j_py_script_session.popScope()
+    else:
+        # in the __main__ module, use the default main global scope
+        yield
```
chipkent, line 83: 1) Are function locals in the main global scope handled by the if-block?
2) Are we properly handling values in an `if __name__ == "__main__":` block?

I think we are ok on these, but I'm confirming, since I'm not 100% sure.  Are there any other weird cases to consider?

py/server/tests/test_table.py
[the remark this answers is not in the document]
chipkent: Similarly, what happens with `time_table`?  Are we freezing the values?

```python
x = 1
t = time_table(...).update("X=x")
for i in range(1e6):
    x = i
```

py/server/deephaven/table.py
chipkent, line 55: Is `_deephaven_pkg_prefix` needed?  I don't think it is used.

py/server/tests/test_table.py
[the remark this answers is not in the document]
chipkent: should also test other scoping cases that are unusual in python:
* lambdas
* list comprehensions
* dict comprehensions
* set comprehensions
* class methods
* static class methods

This link should give a number of weird cases to test against:
https://realpython.com/python-scope-legb-rule/

I'm just paranoid that we check edge cases.

py/server/deephaven/table.py
```diff
@@ -2123,7 +2167,7 @@ def select_distinct(self, formulas: Union[str, Sequence[str]] = None) -> Partiti
         """
         try:
             formulas = to_sequence(formulas)
-            with auto_locking_ctx(self):
+            with auto_locking_ctx(self), _query_scope_ctx():
```
chipkent, line 2170: Did we remove formulas from joins, aggs, etc., or do they need a scope?

py/server/tests/test_table.py
```diff
@@ -560,6 +560,47 @@ def test_layout_hints(self):
...
+
+        inner_func("param str")
+
+    def test_nested_scopes(self):
+        def inner_func(p) -> str:
+            t = empty_table(1).update("X = p * 10")
+            return t.to_string().split()[2]
+
+        t = empty_table(1).update("X = i").update("TableString = inner_func(X + 10)")
+        self.assertIn("100", t.to_string())
```
chipkent, line 596: Should test partitioned tables.

chipkent (changes requested):

py/server/deephaven/table.py
jmao-denver: it is written for the specific purpose of setting formula/filter evaluation scope, unless we expand these evaluation beyond table ops, it is appropriate to keep it internal for now

py/server/tests/test_pandas.py
jmao-denver: one of the test cases expects the Java constant NULL_BYTE in the formula, this is to work around the naming conflicts between Python and Java. A related ticket is #2306.

py/server/deephaven/table.py
jmao-denver: The next comment in the code tries to explain it. 
```
   # we deem any module/function/method which calls a wrapped formula/filter-evaluating Table op as an application
   # frame. This holds true and makes sense even when such module/function/method is part of the Deephaven package, and
   # should not be a hurdle for developing such code as function/method parameters are now part of the query scope
```
Basically we wan to find the wrapper function that directly uses this context manager and uses its caller's locals and the module globals as the query scope. I have changed the names from application/app to caller in the code to make it clear. This is less ambiguous than the notion of 'application scope', which in most cases is the same as the immediate caller's.

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
```diff
@@ -200,4 +211,22 @@ public static Object convert(PyObject pyObject) {
...
+        scopeStack.push(pydict.asDict());
+    }
+
+    @Override
+    public void popScope() {
+        Deque<PyDictWrapper> scopeStack = threadScopeStack.get();
+        if (scopeStack != null) {
+            scopeStack.pop();
+        }
+    }
```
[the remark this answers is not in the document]
jmao-denver: Throwing an illegal state exception is sound practice. Since push and pop are on the same thread, checking equality doesn't seem to be necessary.

engine/table/src/main/java/io/deephaven/engine/util/PythonScopeJpyImpl.java
```diff
@@ -200,4 +211,22 @@ public static Object convert(PyObject pyObject) {
...
+            threadScopeStack.set(scopeStack);
+        }
+        scopeStack.push(pydict.asDict());
+    }
+
+    @Override
+    public void popScope() {
+        Deque<PyDictWrapper> scopeStack = threadScopeStack.get();
+        if (scopeStack != null) {
+            scopeStack.pop();
```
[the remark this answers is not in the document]
jmao-denver: makes sense

py/server/deephaven/table.py
```diff
@@ -43,6 +45,43 @@
...
+
+    # we deem any module/function/method which calls a wrapped formula/filter-evaluating Table op as an application
+    # frame. This holds true and makes sense even when such module/function/method is part of the Deephaven package, and
+    # should not be a hurdle for developing such code as function/method parameters are now part of the query scope
+    if len(outer_frames) > i + 2:
+        app_frame, *_ = outer_frames[i + 1]
+        scope_dict = app_frame.f_globals.copy()
+        scope_dict.update(app_frame.f_locals)
+        try:
+            _j_py_script_session.pushScope(scope_dict)
```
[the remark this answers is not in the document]
jmao-denver: I don't think we need to protect the scope_dict by holding on to the GIL. It is a new dict. Although its values could be mutable, it is the nature of a multithreaded application and if some of the values are desired to remain unchanged during a table operation, it falls on the application to guard them with any of the available sync mechanisms.

That said, I believe we might encounter cases in the future where holding on to the GIL for the duration of a wrapped operation is necessary.

py/server/deephaven/table.py
jmao-denver: always the module where it is defined

py/server/deephaven/table.py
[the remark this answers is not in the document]
devinrsmith: It's not about protecting scope_dict - it's about ensuring `pushScope` is as fast as possible - as it stands right now, there's potential for somebody else to acquire the GIL and for the return from java pushScope -> python to block.

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -178,6 +178,14 @@ public <T> T getVariable(String name, T defaultValue) {
                 .orElse(defaultValue);
     }
 
+    public void pushScope(PyObject pydict) {
+        scope.pushScope(pydict);
```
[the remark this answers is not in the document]
jmao-denver: it doesn't hurt for future

engine/table/src/main/java/io/deephaven/engine/util/PythonScope.java
```diff
@@ -170,4 +171,17 @@ default Map<String, Object> getEntriesMap() {
     }
 
     public PyDictWrapper globals();
```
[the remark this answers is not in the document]
jmao-denver: will rename it to mainGlobals(). it is still meaningful for the gRPC server as table, plots etc in the main globals are automatically exported to the Web UI.

py/server/deephaven/table.py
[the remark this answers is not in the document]
jmao-denver: I can't find any case other than in the table module that query scope is applicable. If/when query scope is needed in other modules in the future, we should be able to easily generalize the context manager for broader use.

py/server/deephaven/table.py
jmao-denver: I think so. Support formulas in joins, etc. adds minimum convenience but would be outweighed by the added complexity in the UI and in the query engine.

py/server/deephaven/table.py
jmao-denver: 1. yes
2. yes
3. AFAIK, no

engine/table/src/main/java/io/deephaven/engine/util/PythonScope.java
```diff
@@ -170,7 +170,7 @@ default Map<String, Object> getEntriesMap() {
          */
     }
 
-    public PyDictWrapper globals();
+    public PyDictWrapper mainGlobals();
```
chipkent, line 173: javadoc.  It looks like you renamed it, but I'm still not certain what "it" is.  `globals` may have been a fine name.  It seems like these might be "query globals".

py/server/tests/test_pt_proxy.py
```diff
@@ -318,6 +318,40 @@ def test_agg_all_by(self):
             for gct in agg_pt_proxy.target.constituent_tables:
                 self.assertGreaterEqual(gct.size, 1)
 
+    def test_query_scope(self):
```
chipkent, line 321: Should the whole set of scope tests from table be here?

py/server/deephaven/table.py
chipkent: Is this needed?

py/server/tests/test_table.py
```diff
@@ -560,6 +562,109 @@ def test_layout_hints(self):
         self.assertTrue(cm.exception.root_cause)
         self.assertIn("RuntimeError", cm.exception.compact_traceback)
 
+    def verify_table_data(self, t: Table, expected: List[Any]):
+        t_data = to_pandas(t).values[0]
+        for s in expected:
+            self.assertIn(s, t_data)
+
+    def test_update_local(self):
```
chipkent, line 570: Should also test what happens when variables change. e.g.

```python
x = 1
t = emptyTable(1).update("X=x")
assert X==1
x = 2
t = t.update("Y=x")
assert Y==2
```

chipkent: Similarly, what happens with time_table? Are we freezing the values?
```python
x = 1
t = time_table(...).update("X=x")
for i in range(1e6):
    x = i
```

chipkent (changes requested):

py/server/tests/test_pt_proxy.py
jmao-denver: I don't think it is necessary. The new tests for table cover most if not all the use cases to verify that we build the correct scope for formula evaluation

engine/table/src/main/java/io/deephaven/engine/util/PythonScope.java
jmao-denver: it is the Python's main module namespace as opposed to any imported module's.

py/server/deephaven/table.py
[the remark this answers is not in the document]
jmao-denver: I don't believe the performance argument is as strong as the consistency one, but filed #2741

py/server/tests/test_pt_proxy.py
chipkent: I suggested this because @rcaudy implied that partitioned tables were somehow different than other cases.

chipkent (approved):
See questions in chat for @rcaudy

PR #2441, opened 2022-05-26
Remove cast to int when getting size from Range

### LINKED ISSUE #2436: very large timeTable does not render
```groovy
t = timeTable(new DateTime(0), "00:00:01")
```

works

```groovy
t = timeTable(new DateTime(0), "00:00:00.1")
```

does not. It renders as an empty table in the UI, and reports a negative number of rows. The underlying data structure appears to work `println(t.size())` appears correct.

mofojed:
I don't know why we'd cast to int and then cast right back to long.

Fixes #2436

### DIFF
--- a/web/shared-beans/src/test/java/io/deephaven/web/shared/data/RangeSetTest.java
+++ b/web/shared-beans/src/test/java/io/deephaven/web/shared/data/RangeSetTest.java
@@ -408,8 +408,48 @@ public class RangeSetTest {
         assertEquals(of(
                 new Range(5, 10),
                 new Range(15, 20)), rangeSet);
+    }
 
+    @Test
+    public void testLarge() {
+        long largeA = 16535734029L;
+        long largeB = 36535734029L;
 
+        // Add to empty range
+        RangeSet rangeSet = RangeSet.empty();
+        rangeSet.addRange(new Range(0, largeA));
+        assertEquals(of(new Range(0, largeA)), rangeSet);
+
+        rangeSet = RangeSet.empty();
+        rangeSet.addRange(new Range(largeA, largeA));
+        assertEquals(of(new Range(largeA, largeA)), rangeSet);
+
+        rangeSet = RangeSet.empty();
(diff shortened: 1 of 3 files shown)


niloc132 (approved):
Approved. I can't remember exactly why we did this once upon a time, but I want to say it was something to try to get the data flow better in compiled output, so that the compiler could trust that we had only ints, so that the extra generated long code didnt need to happen. I'll take a deeper look later, but IIRC the range/rangeset tests are pretty good - maybe you could add a unit test that confirms this, confirms adding multiple gigantic ranges to a rangeset, removing them, etc?

web/shared-beans/src/test/java/io/deephaven/web/shared/data/RangeTest.java
```diff
@@ -45,4 +45,13 @@ public void testAdjacentMerge() {
         assertEquals(new Range(0, 5), rangeB.overlap(rangeA));
     }
 
+    @Test
+    public void testLargeRange() {
+        long largeA = 16535734029L;
+        long largeB = 16535734030L;
```
niloc132, line 51: would you mind making these big numbers explicitly Integer.MAX_VALUE and then add some constant?

web/shared-beans/src/test/java/io/deephaven/web/shared/data/RangeSetTest.java
```diff
@@ -408,8 +408,48 @@ public void testRemove() {
         assertEquals(of(
                 new Range(5, 10),
                 new Range(15, 20)), rangeSet);
+    }
+
+    @Test
+    public void testLarge() {
+        long largeA = 16535734029L;
+        long largeB = 36535734029L;
```
niloc132, line 416: ditto these constants, just to be sure we know what they are.

Maybe consider a reallyLargeA/etc, for Long.MAX_VALUE - 100 or something, though that will break when we pass it to JS

niloc132 (dismissed):

niloc132 (approved):

PR #2445, opened 2022-05-26
Fix python plotting autogeneration method resolution order and type conversion for strings

### LINKED ISSUE #2444: `chart_title` does not resolve the proper methods in python
The following code:
1) fails with unresolved java function
2) chooses less specific methods over more specific methods.

```python
from deephaven.column import double_col

reg = new_table([
    double_col("R2", [0.2]),
    double_col("Beta", [2.0]),
    double_col("Intercept", [0.67]),
])

data = new_table([
    double_col("X", list(range(10))),
    double_col("Y", list(range(10))),
])

plt = Figure() \
    .plot_xy("Y Series", t=data, x="X", y="Y") \
    .chart_title(t=reg, format="My title: R2={0} Beta={1} Intercept={2}", columns=["R2", "Beta", "Intercept"]) \
    .show()

plt = Figure() \
    .plot_xy("Y Series", t=data, x="X", y="Y") \
    .chart_title(t=reg, columns=["R2", "Beta", "Intercept"]) \
    .show()
```

chipkent:
1) Fix the order that autogenerated python plotting functions are resolved, so that more specific methods are chosen before more general methods.
2) Fix a bug in how numpy types are resolved to DH types.  Specifically, string types are not handled correctly.

Resolves #2444

### DIFF
--- a/Generators/src/main/java/io/deephaven/plot/util/GeneratePyV2FigureAPI.java
+++ b/Generators/src/main/java/io/deephaven/plot/util/GeneratePyV2FigureAPI.java
@@ -1,5 +1,6 @@
 package io.deephaven.plot.util;
 
+import io.deephaven.base.Pair;
 import io.deephaven.libs.GroovyStaticImportGenerator.JavaFunction;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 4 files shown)


jmao-denver (approved):
LGTM, wish we had a better way for mapping the 'string' type to numpy ones

PR #2456, opened 2022-06-01
Python server sync - Fixes #1480

### LINKED ISSUE #1480: Pydeephaven Session() should display tables that currently exist in Deephaven
As a Python user, I want to see tables that already exist in Deephaven when I make a new connection via the `pydeephaven` package, so that I can interact with tables that have already been created (such as through app mode, another script, UI actions, etc).

For example, if I have a Deephaven instance and use the UI to create a table called `result`, I'd expect the following Python script to work

```
from pydeephaven import Session

session = Session()

print(session.tables) #Should display a list with ["result"]

result = session.open_table("result") #Should connect to the "result" table on the server
```

SuperTails:
This PR adds three things.
1. Sessions now request the list of tables when connecting to the server, so tables created before the session starts (from previous sessions, from the web UI, etc.) can be opened.
2. `session.subscribe_fields()` now starts up a thread that checks for new changes to tables, so tables created/deleted in other sessions/the web UI become visible automatically.
3. Tests for the above two features.

### DIFF
--- a/py/client/pydeephaven/_app_service.py
+++ b/py/client/pydeephaven/_app_service.py
@@ -1,0 +1,20 @@
+#
+#  Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
+#
+from pydeephaven.dherror import DHError
+from pydeephaven.proto import application_pb2_grpc, application_pb2
+
+class AppService:
+    def __init__(self, session):
+        self.session = session
+        self._grpc_app_stub = application_pb2_grpc.ApplicationServiceStub(session.grpc_channel)
+
+    def list_fields(self):
+        try:
+            fields = self._grpc_app_stub.ListFields(
+                application_pb2.ListFieldsRequest(),
+                metadata=self.session.grpc_metadata
+            )
+            return fields
+        except Exception as e:
+            raise DHError("failed to list fields.") from e
--- a/py/client/pydeephaven/session.py
+++ b/py/client/pydeephaven/session.py
@@ -69,6 +69,8 @@ class Session:
         self._never_timeout = never_timeout
         self._keep_alive_timer = None
         self._session_type = session_type
+        self._list_fields = None
+        self._field_update_thread = None
 
         self._connect()
 
@@ -127,12 +129,37 @@ class Session:
 
             return self._last_ticket
 
+    def subscribe_fields(self):
+        """
+        Allows tables created/removed on the server to be automatically seen on the client
+        """
+
+        with self._r_lock:
+            if self._field_update_thread is None:
+                self._field_update_thread = threading.Thread(target=self._update_fields)
+                self._field_update_thread.daemon = True
+                self._field_update_thread.start()
+
+    def _update_fields(self):
+        """
+        Constant loop that checks for any server-side table changes and adds them to the local list
+        """
+
+        while True:
+            fields_change = next(self._list_fields)
+            with self._r_lock:
+                self._parse_fields_change(fields_change)
+
     def _connect(self):
         with self._r_lock:
             self.grpc_channel, self.session_token, self._timeout = self.session_service.connect()
             self.is_connected = True
+
             if self._never_timeout:
                 self._keep_alive()
+            
+            self._list_fields = self.console_service.list_fields()
+            self._parse_fields_change(next(self._list_fields))
 
     def _keep_alive(self):
         if self._keep_alive_timer:
@@ -181,20 +208,23 @@ class Session:
     def release(self, ticket):
         self.session_service.release(ticket)
 
-    def _parse_script_response(self, response):
-        if response.changes.created:
-            for t in response.changes.created:
+    def _parse_fields_change(self, fields_change):
+        if fields_change.created:
+            for t in fields_change.created:
                 t_type = None if t.typed_ticket.type == '' else t.typed_ticket.type
                 self._tables[t.field_name] = (t_type, Table(session=self, ticket=t.typed_ticket.ticket))
 
-        if response.changes.updated:
-            for t in response.changes.updated:
+        if fields_change.updated:
+            for t in fields_change.updated:
                 t_type = None if t.typed_ticket.type == '' else t.typed_ticket.type
                 self._tables[t.field_name] = (t_type, Table(session=self, ticket=t.typed_ticket.ticket))
 
-        if response.changes.removed:
-            for t in response.changes.removed:
+        if fields_change.removed:
+            for t in fields_change.removed:
                 self._tables.pop(t.field_name, None)
+
+    def _parse_script_response(self, response):
+        self._parse_fields_change(response.changes)
 
     # convenience/factory methods
     def run_script(self, script: str) -> None:
--- a/py/client/pydeephaven/session.py
+++ b/py/client/pydeephaven/session.py
@@ -11,6 +11,7 @@ from bitstring import BitArray
 
 from pydeephaven._arrow_flight_service import ArrowFlightService
 from pydeephaven._console_service import ConsoleService
+from pydeephaven._app_service import AppService
 from pydeephaven._session_service import SessionService
 from pydeephaven._table_ops import TimeTableOp, EmptyTableOp, MergeTablesOp, FetchTableOp
 from pydeephaven._table_service import TableService
@@ -65,6 +66,7 @@ class Session:
         self._grpc_barrage_stub = None
         self._console_service = None
         self._flight_service = None
+        self._app_service = None
         self._tables = {}
         self._never_timeout = never_timeout
         self._keep_alive_timer = None
@@ -114,6 +116,13 @@ class Session:
             self._flight_service = ArrowFlightService(self)
 
         return self._flight_service
+    
+    @property
+    def app_service(self):
+        if not self._app_service:
+            self._app_service = AppService(self)
+        
+        return self._app_service
 
     def make_ticket(self, ticket_no=None):
         if not ticket_no:
@@ -158,7 +167,7 @@ class Session:
             if self._never_timeout:
                 self._keep_alive()
             
-            self._list_fields = self.console_service.list_fields()
+            self._list_fields = self.app_service.list_fields()
             self._parse_fields_change(next(self._list_fields))
 
     def _keep_alive(self):
--- a/py/client/tests/test_sync.py
+++ b/py/client/tests/test_sync.py
@@ -1,0 +1,38 @@
+
+
+import unittest
+from time import sleep
+
+import pyarrow as pa
+from pyarrow import csv
+
+from pydeephaven import DHError
+from pydeephaven import Session
+from tests.testbase import BaseTestCase
+
+
+class SyncTestCase(BaseTestCase):
+    def test_persistent_tables(self):
+        with Session() as session1:
+            session1 = Session()
+            session1.run_script('t = None')
+            t = session1.empty_table(10)
+            session1.bind_table('t', t)
+
+        sleep(1)
+
+        with Session() as session2:
+            self.assertIn('t', session2.tables)
+
+    def test_shared_tables(self):
+        session1 = Session()
+        session1.run_script('t = None')
+        session1.subscribe_fields()
+
+        session2 = Session()
+        t = session2.empty_table(10)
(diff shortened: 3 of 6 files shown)


SuperTails:

py/client/pydeephaven/session.py
```diff
@@ -127,12 +138,37 @@ def get_ticket(self):
 
             return self._last_ticket
 
+    def subscribe_fields(self):
+        """
+        Allows tables created/removed on the server to be automatically seen on the client
+        """
```
jmao-denver, line 144: 1. should we make it private? If so it could save us a lot of documentation on how it should be typically used it and the implication of the dynamically changing nature of the fields, until such need is requirement.
2. the docstring format should be """ Allows ... """ to be consistent with the rest of the code.

py/client/tests/test_sync.py
```diff
@@ -0,0 +1,38 @@
...
+import pyarrow as pa
+from pyarrow import csv
+
+from pydeephaven import DHError
+from pydeephaven import Session
+from tests.testbase import BaseTestCase
+
+
+class SyncTestCase(BaseTestCase):
+    def test_persistent_tables(self):
```
jmao-denver, line 15: 'sync' is a bit too generic. A more specific name is preferred.

jmao-denver (changes requested):
Very solid, needs only minor changes

py/client/pydeephaven/_app_service.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def list_fields(self):
+        try:
+            fields = self._grpc_app_stub.ListFields(
+                application_pb2.ListFieldsRequest(),
+                metadata=self.session.grpc_metadata
+            )
+            return fields
+        except Exception as e:
+            raise DHError("failed to list fields.") from e
```
devinrsmith, line 20: Not to distract from this PR - but I think we might want to look into the python grpc asyncio APIs - it might be a more natural fit for streaming responses like this. @jmao-denver

py/client/pydeephaven/session.py
devinrsmith: IMO, I think we do want a public method that is opt-in (or have a boolean on connect).

py/client/pydeephaven/session.py
```diff
@@ -127,12 +138,37 @@ def get_ticket(self):
...
     def _connect(self):
         with self._r_lock:
             self.grpc_channel, self.session_token, self._timeout = self.session_service.connect()
             self.is_connected = True
+
             if self._never_timeout:
                 self._keep_alive()
+            
+            self._list_fields = self.app_service.list_fields()
+            self._parse_fields_change(next(self._list_fields))
```
devinrsmith, line 167: Couple things:

1. I don't think we should be subscribing on connect.
2. As is, it's not correct - we've starting a list_fields subscription, but we haven't started the threading handler for it. I don't know exactly how python gRPC will handle this (either a memory leak in list_fields, or eventually an error b/c it's not being consumed).

I think we need this logic self-contained within the subscribe_fields method.

py/client/pydeephaven/session.py
```diff
@@ -181,21 +217,24 @@ def close(self) -> None:
...
                 self._tables[t.field_name] = (t_type, Table(session=self, ticket=t.typed_ticket.ticket))
 
-        if response.changes.removed:
-            for t in response.changes.removed:
+        if fields_change.removed:
+            for t in fields_change.removed:
                 self._tables.pop(t.field_name, None)
 
+    def _parse_script_response(self, response):
+        self._parse_fields_change(response.changes)
```
devinrsmith, line 232: I think we may be doubling up on our response handling for code we execute via scripts, in the case where we are also subscribing to list fields. We may want to ignore list_field changes that apply to the global script scope (FieldInfo.application_id == "scope"), or choose to only handle list_field changes (ignore script responses) when they are active.

The problem w/ the former is if we ignore the "scope" changes for list_field, that will miss updates executed in the web UI instead of via the python client.

The problem w/ the latter is that updates to the fields will be async with the response to script execution response.

py/client/tests/test_sync.py
```diff
@@ -0,0 +1,38 @@
...
+        with Session() as session1:
+            session1 = Session()
+            session1.run_script('t = None')
+            t = session1.empty_table(10)
+            session1.bind_table('t', t)
+
+        sleep(1)
```
devinrsmith: Do we need this sleep? I think we can probably formulate these tests w/o needing to rely on them.

devinrsmith (changes requested):
This is a good first pass, welcome to Deephaven!

py/client/tests/test_sync.py
devinrsmith: Anytime I see a sleep, it immediately triggers 'code smell'. I wonder if we should have a way for session1 to block on the successful notification that `t` has been propogated.

py/client/pydeephaven/session.py
SuperTails: This is now available as `Session(sync_fields=SYNC_ONCE)` or as `session.sync_fields(repeated=False)`

py/client/pydeephaven/session.py
SuperTails: I decided to go with the former option, because otherwise the user can't rely on any of the tables they create in a script being usable until after some arbitrary delay, which seems like a significant usability problem.
Plus, ignoring global-script-scope changes doesn't harm existing functionality.

py/client/tests/test_sync.py
SuperTails: Fixed by using a loop and a timeout instead (though the test is skipped for other reasons anyways...)

SuperTails:
There have been a number of changes made. The two new functions have been collapsed into one: `sync_fields(repeating=bool)`. This is to make it more clear that calling it multiple times is redundant.
`Session` now takes an optional argument that effectively just defers to a `sync_fields` call; this is solely so that this new feature can be found easier when looking through docs.
In order to avoid problems with duplicate handling of table changes, however, the repeating/background syncing doesn't register changes made at the global scope, so it's not super useful. This does not affect the one-time sync: all tables/plots/etc are picked up and are usable as request in the issue.

py/client/pydeephaven/session.py
```diff
@@ -181,20 +249,26 @@ def close(self) -> None:
     def release(self, ticket):
         self.session_service.release(ticket)
 
-    def _parse_script_response(self, response):
-        if response.changes.created:
-            for t in response.changes.created:
-                t_type = None if t.typed_ticket.type == '' else t.typed_ticket.type
-                self._tables[t.field_name] = (t_type, Table(session=self, ticket=t.typed_ticket.ticket))
```
devinrsmith, line 188: Looking at this old code, I actually think it was wrong. `field_name` by itself is *not* enough to uniquely distinguish fields - you need `(t.application_id, t.field_name)` or `t.ticket_ticket.ticket`. I would suggest `t.ticket_ticket.ticket` as it's more future-proof.

This also suggests that the documentation to `open_table(self, name: str)` should be updated to mention that it's opening a globally scoped table.

py/client/tests/test_multi_session.py
```diff
@@ -0,0 +1,43 @@
+
+
+import unittest
+from time import sleep
```
devinrsmith, line 4: No longer needed.

py/client/pydeephaven/session.py
SuperTails: Is it guaranteed that globally scoped tables use the `"s/tablename"` format for their tickets? If not, we probably need to use `(t.application_id, t.field_name)` so that we can look up tables in `open_table`.

py/client/pydeephaven/session.py
```diff
@@ -181,20 +249,26 @@ def close(self) -> None:
...
+        if fields_change.removed:
+            for t in fields_change.removed:
+                if allow_scope_changes or t.application_id != 'scope':
+                    self._fields.pop(t.field_name, None)
 
-        if response.changes.removed:
-            for t in response.changes.removed:
-                self._tables.pop(t.field_name, None)
+    def _parse_script_response(self, response):
+        self._parse_fields_change(response.changes, True)
```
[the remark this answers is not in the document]
SuperTails: I like this idea, but I think it should be flipped around. To me it seems most intuitive that the user's own operations should always be immediately accessible (and the most convenient); being able to get other sessions' tables should be an opt-in (this whole feature is an opt-in itself, after all). I think the "I need everyone's tables, but asynchronously" usecase is fairly niche.

```py
session1.sync_fields(True)
async_changes = session1.async_changes_handle()

session1.run_script('t = empty_table(1)')
x1 = session1.open_table('t') # works immediately
x2 = async_changes_handle.open_table('t') # eventual access if sync'd
```

Though, while we're throwing out ideas: What about a separate `AsyncSession` class with the same interface but different guarantees? No idea how it would work, but just throwing it out.

py/client/pydeephaven/session.py
devinrsmith: We can store the full field ` _fields`, and search for `f.application_id == "scope" and f.field_name == name`.

devinrsmith: We should not make any assumptions about the format of `t.typed_ticket.ticket`

SuperTails:
By canceling the ListFields request before a script is run, then restarting it afterwards, it is possible to sidestep the sync/async changes issue entirely at the cost of needing to request the full list of tables again. Since this is a significant usability improvement for a mode that may or may not be very popular, this seems reasonable.

py/client/pydeephaven/session.py
```diff
@@ -206,12 +287,23 @@ def run_script(self, script: str) -> None:
...
+                self._cancel_update_fields()
+            
             response = self.console_service.run_script(script)
-            self._parse_script_response(response)
+
+            if self._sync_fields == SYNC_REPEATED:
+                # We can ignore the script response because
+                # all the new tables are added by this call anyways
+                self._fields = {}
+                self.sync_fields(repeating=True)
```
devinrsmith, line 301: So this isn't actually true if there are other list field subscribers at the time run_script is called. If so, run_script changes get batched with a certain frequency (250ms by default), so they aren't necessarily visible on the first return from sync_fields. I think just applying both will "work".

py/client/pydeephaven/session.py
```diff
@@ -127,13 +154,63 @@ def get_ticket(self):
 
             return self._last_ticket
 
+    def sync_fields(self, repeating: bool):
+        """ Check for fields that have been added/deleted by other sessions and add them to the local list
+
+        This will start a new background thread when `repeating=True`.
```
devinrsmith, line 160: It should be noted - when repeating, only the *first* response will modify global "scope" - future responses against global "scope" will *not* be handled.

SuperTails: Since the issue with script response/listfields response conflicts is fixed, can the repeating sync_fields be switched back to handling all changes?

devinrsmith: Yes, I think that's correct. But my other comment is still relevant - we want run_script changes to be visible upon return.

devinrsmith (approved):

py/client/pydeephaven/session.py
```diff
@@ -29,18 +35,24 @@ class Session:
...
 
     Attributes:
-        tables (list[str]): names of the tables available in the server after running scripts
+        tables (list[str]): names of the global tables available in the server after running scripts
         is_alive (bool): check if the session is still alive (may refresh the session)
     """
 
-    def __init__(self, host: str = None, port: int = None, never_timeout: bool = True, session_type: str = 'python'):
+    def __init__(self, host: str = None, port: int = None, never_timeout: bool = True, session_type: str = 'python', sync_fields: int = NO_SYNC):
         """ Initialize a Session object that connects to the Deephaven server
```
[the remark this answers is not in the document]
jakemulf: I think `NO_SYNC` might actually be the most reasonable default. I generally find that no-ops are the best defaults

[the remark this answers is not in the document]
jmao-denver: Thanks @jakemulf

jmao-denver (approved):
Approve based on Jake's feedback

PR #2457, opened 2022-06-01
Update import_test_data.py

margaretkennedy:
Uses a working URL for a Parquet file, as the CSV URL was outdated.

### DIFF
--- a/py/client/examples/import_test_data.py
+++ b/py/client/examples/import_test_data.py
@@ -25,7 +25,7 @@ def import_taxi_records(dh_session: Session) -> Table:
                         "store_and_fwd_flag", "PULocationID", "DOLocationID"]
     pa_table = pa_table.drop(unwanted_columns)
 
-    # this is necessary because otherwise the data is too big to be sent
+    # this is necessary because otherwise, the data is too big to be sent as a gRPC message
     new_len = pa_table.num_rows // 2
     pa_table = pa_table.slice(length=new_len)
 


chipkent (approved):

py/client/examples/import_test_data.py
```diff
@@ -1,37 +1,39 @@
...
+    pa_table = parquet.read_table(file_name)
 
     # drop unwanted columns
     unwanted_columns = ["tpep_pickup_datetime", "tpep_dropoff_datetime", "RatecodeID",
                         "store_and_fwd_flag", "PULocationID", "DOLocationID"]
     pa_table = pa_table.drop(unwanted_columns)
 
+    # this is necessary because otherwise the data is too big to be sent
+    new_len = pa_table.num_rows // 2
+    pa_table = pa_table.slice(length=new_len)
```
chipkent, line 30: After looking, I'm not sure what this means.  What makes it too big?  Out of RAM on the server, or something else?

margaretkennedy: @SuperTails @jakemulf

SuperTails: The table is too big to be sent as a gRPC message. The deephaven server errors out with the following:
`gRPC message exceeds maximum size 104857600: 109183697`

py/client/examples/csv_downloader.py
```diff
@@ -5,12 +5,8 @@
 
 from requests import get
 
-
-def download_csv(url: str, file_name=None) -> str:
-    """ download a CSV located at the supplied URL and return the file name. """
-    if not file_name:
-        file_name = "sample.csv"
-
+def download_file(url: str, file_name: str) -> str:
```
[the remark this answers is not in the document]
SuperTails: This function worked okay, but since we're downloading parquet files instead the name/default didn't seem appropriate. (And the function works equally well for any filetype, really)

jmao-denver (dismissed):

py/client/examples/import_test_data.py
chipkent: @niloc132 making sure you've seen this.

niloc132: Short version, we want to have a max size of some kind - default was 5mb, we raised it to 100mb.

This is configurable, the property is `grpc.maxInboundMessageSize`, set to a value in bytes. This can be set from a prop file or additional args when starting the server.

Another resolution could be for the py client to break up its upload to smaller pieces, rather than send it all at once.

py/client/examples/import_test_data.py
```diff
@@ -1,37 +1,39 @@
...
+    # download the data and read it into a pyarrow table and prepare it for uploading into DH
+    file_name = download_file(url="https://nyc-tlc.s3.amazonaws.com/trip+data/yellow_tripdata_2020-12.parquet", file_name="sample.parquet")
+    pa_table = parquet.read_table(file_name)
 
     # drop unwanted columns
     unwanted_columns = ["tpep_pickup_datetime", "tpep_dropoff_datetime", "RatecodeID",
                         "store_and_fwd_flag", "PULocationID", "DOLocationID"]
     pa_table = pa_table.drop(unwanted_columns)
 
+    # this is necessary because otherwise, the data is too big to be sent as a gRPC message
```
chipkent, line 28:
```suggestion
    # Send only half of the table to avoid gRPC message size limit
```

chipkent (changes requested):

PR #2475, opened 2022-06-03
Better stream table support over Barrage and jsapi

### LINKED ISSUE #1492: stream table semantics via Barrage
If a stream table has a fast update cycle, and a remote subscriber has a slow update cycle; it would be nice if the update to the remote subscriber was the aggregate of the underlying stream changes.

### LINKED ISSUE #2377: Preserve stream semantics over barrage
It would be nice if stream table semantics were preserved over barrage.

This would allow nice data-mesh applications where workers exporting stream tables could be purpose-built as ingress/egress points.

Discussed with @nbauernfeind and @rcaudy, the implementation would likely be easy.

nbauernfeind:
Fixes #2377. Fixes #1492.

It also adds some stream table specific changes to jsTable that prevent flickering and improve overall GUI experience for stream tables.

### DIFF
--- a/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
+++ b/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
@@ -15,6 +15,8 @@ import io.deephaven.engine.table.Table;
 import io.deephaven.engine.table.TableDefinition;
 import io.deephaven.engine.table.MatchPair;
 import io.deephaven.extensions.barrage.chunk.vector.VectorExpansionKernel;
+import io.deephaven.internal.log.LoggerFactory;
+import io.deephaven.io.logger.Logger;
 import io.deephaven.proto.flight.util.MessageHelper;
 import io.deephaven.proto.flight.util.SchemaHelper;
 import io.deephaven.time.DateTime;
@@ -26,7 +28,6 @@ import io.deephaven.engine.table.impl.RollupInfo;
 import io.deephaven.chunk.ChunkType;
 import io.deephaven.proto.backplane.grpc.ExportedTableCreationResponse;
 import io.deephaven.util.type.TypeUtils;
-import io.deephaven.vector.ObjectVector;
 import io.deephaven.vector.Vector;
 import org.apache.arrow.flatbuf.KeyValue;
 import org.apache.arrow.util.Collections2;
@@ -58,8 +59,9 @@ import java.util.function.Function;
 import java.util.function.IntFunction;
 
 public class BarrageUtil {
-
     public static final long FLATBUFFER_MAGIC = 0x6E687064;
+
+    private static final Logger log = LoggerFactory.getLogger(BarrageUtil.class);
 
     // year is 4 bytes, month is 1 byte, day is 1 byte
     public static final ArrowType.FixedSizeBinary LOCAL_DATE_TYPE = new ArrowType.FixedSizeBinary(6);
@@ -77,6 +79,7 @@ public class BarrageUtil {
 
     private static final String ATTR_DH_PREFIX = "deephaven:";
     private static final String ATTR_ATTR_TAG = "attribute";
+    private static final String ATTR_ATTR_TYPE_TAG = "attribute_type";
     private static final String ATTR_TYPE_TAG = "type";
     private static final String ATTR_COMPONENT_TYPE_TAG = "componentType";
 
@@ -138,6 +141,7 @@ public class BarrageUtil {
                     val instanceof Character || val instanceof Boolean ||
                     (val instanceof String && ((String) val).length() < ATTR_STRING_LEN_CUTOFF)) {
                 putMetadata(schemaMetadata, ATTR_ATTR_TAG + "." + key, val.toString());
+                putMetadata(schemaMetadata, ATTR_ATTR_TYPE_TAG + "." + key, val.getClass().getCanonicalName());
             } else {
                 unsentAttributes.add(key);
             }
@@ -296,7 +300,7 @@ public class BarrageUtil {
         // a multiplicative factor to apply when reading; useful for eg converting arrow timestamp time units
         // to the expected nanos value for DateTime.
         public int[] conversionFactors;
-        public Map<String, String> attributes;
+        public Map<String, Object> attributes;
 
         public ConvertedArrowSchema(final int nCols) {
             this.nCols = nCols;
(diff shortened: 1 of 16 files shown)


server/src/main/java/io/deephaven/server/barrage/BarrageMessageProducer.java
```diff
@@ -1315,7 +1334,20 @@ private void updateSubscriptionsSnapshotAndPropagate() {
...
+                if (!isStreamTable) {
+                    snapshot = getSnapshot(growingSubscriptions, snapshotColumns, snapshotRowSet,
+                            reverseSnapshotRowSet);
+                } else {
+                    // acquire an empty snapshot to properly align column subscription changes to a UGP step
+                    snapshot = getSnapshot(growingSubscriptions, snapshotColumns, RowSetFactory.empty(),
+                            RowSetFactory.empty());
+
+                    // in the event that the stream table was not empty; pretend it was
+                    if (!snapshot.rowsAdded.isEmpty()) {
```
lbooker42, line 1346: Is this possible after passing `RowSetFactory.empty()` into `getSnapshot()`?

lbooker42 (commented):
These changes look good and are easy to follow, would like to see unit tests exercising this code in `BarrageMessageProducer` and the associated `BarrageTable`s . The `web` changes should be reviewed by @niloc132.

web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
```diff
@@ -490,6 +499,11 @@ public void setInternalViewport(double firstRow, double lastRow, Column[] column
         if (firstRow < 0) {
             throw new IllegalArgumentException(firstRow + " < " + 0);
         }
+        final ActiveTableBinding sub = state().getActiveBinding(this);
+        if (isStreamTable() && sub != null && sub.getSubscription() != null) {
```
niloc132, line 503: this feels very racy to me, having gotten bitten by this in the past. i think instead you want to test for `subscriptions.get(getHandle())`, and if non-null, check the status of that subscription...

...that said, isn't this still a ui issue, where the UI is calling setViewport unnecessarily?

--

i'd also try to push this into the TableViewportSubscription, which should manage the "is this right one?" question, and just let you say "in this case, never forward the viewport change". If i'm reading this right, you are going to end up passing through there anyway.

web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
```diff
@@ -212,6 +217,10 @@ public boolean hasInputTable() {
         return hasInputTable;
     }
 
+    public boolean isStreamTable() {
```
niloc132, line 220: if js should know about this, mark it as @JsProperty

server/src/main/java/io/deephaven/server/barrage/BarrageMessageProducer.java
```diff
@@ -2133,17 +2219,23 @@ BarrageMessage getSnapshot(
             final BitSet columnsToSnapshot,
             final RowSet positionsToSnapshot,
             final RowSet reversePositionsToSnapshot) {
-        if (onGetSnapshot != null) {
+        if (onGetSnapshot != null && onGetSnapshotIsPreSnap) {
             onGetSnapshot.run();
         }
 
         // TODO: Use *this* as snapshot tick source for fail fast.
         // TODO: Let notification-indifferent use cases skip notification test
```
niloc132, line 2227: nit, these are missing issues

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
```diff
@@ -388,12 +392,65 @@ private static ConvertedArrowSchema convertArrowSchema(
...
+        final HashMap<String, String> attributeTypeMap = new HashMap<>();
         tableMetadataVisitor.accept((key, value) -> {
-            final String prefix = ATTR_DH_PREFIX + ATTR_ATTR_TAG + ".";
-            if (!key.startsWith(prefix)) {
+            final String isAttributePrefix = ATTR_DH_PREFIX + ATTR_ATTR_TAG + ".";
+            final String isAttributeTypePrefix = ATTR_DH_PREFIX + ATTR_ATTR_TYPE_TAG + ".";
+            if (key.startsWith(isAttributePrefix)) {
+                result.attributes.put(key.substring(isAttributePrefix.length()), value);
+            } else if (key.startsWith(isAttributeTypePrefix)) {
+                attributeTypeMap.put(key.substring(isAttributeTypePrefix.length()), value);
```
niloc132, line 403: looks unrelated, would you consider breaking this out and/or also including this in the js impl, so that it can have support for knowing what the types of attrs should be?

web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
```diff
@@ -1220,6 +1234,18 @@ public void handleDelta(ClientTableState current, DeltaUpdates updates) {
                 nonViewportSub.handleDelta(updates);
                 return;
             }
+            if (isStreamTable) {
+                // stream tables remove all rows from the previous step, if there are no adds this step then defer
+                // removal until new data arrives -- this makes stream tables GUI friendly
+                if (updates.getAdded().isEmpty()) {
+                    deferredDeltaRemoves = updates.getRemoved();
```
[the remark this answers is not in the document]
nbauernfeind: added an assert

server/src/main/java/io/deephaven/server/barrage/BarrageMessageProducer.java
nbauernfeind: Looks like Ryan added them; I'll keep up with him to figure something out.

web/client-api/src/main/java/io/deephaven/web/client/api/barrage/def/TableAttributesDefinition.java
```diff
@@ -41,6 +45,10 @@ public boolean isInputTable() {
         return remainingAttributeKeys.contains(INPUT_TABLE_ATTRIBUTE);
     }
 
+    public boolean isStreamTable() {
+        return map.getOrDefault(STREAM_TABLE_ATTRIBUTE, "").equals("true");
```
niloc132, line 49: consider `"true".equals(map.get(STREAM_TABLE_ATTRIBUTE)` so you don't have to care about nulls

web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
```diff
@@ -1220,6 +1230,20 @@ public void handleDelta(ClientTableState current, DeltaUpdates updates) {
                 nonViewportSub.handleDelta(updates);
                 return;
             }
+            if (isStreamTable) {
```
niloc132, line 1233: This whole block should be able to move upstream too, so that if there is more than one Table that is attached to a single ticket, they all get their removes deferred at once, instead of each tracking its own deferrals?

ClientTableState.handleDelta is before it fans out to each downstream sub (admittedly it is weird to hold state about a subscription there), or WorkerConnection.incrementalUpdates is one step upstream of that. I could be pretty easily convinced that neither is appropriate.

Upstream another step is the anonymous class that handles FlightData messages arriving - this _does_ track state of messages, presently to be able to see how many more messages are needed before the current delta is done. This use case smells a bit like that?

niloc132 (approved):
Marked approved, but with one "maybe change this, it seems potentially better to me" and one "this could be simpler if you wanted", but both optional.

Issue #2469, opened 2022-06-02
Python PartitionedTable.constituentFor wrapper issues


rcaudy:
# Bug Description:

There is a bug in `test_partitioned_table.py` for `test_get_constituent`. Excerpted from results of `./gradlew :Integrations:test-py-deephaven`:
```
======================================================================
FAIL [0.013s]: test_get_constituent (tests.test_partitioned_table.PartitionedTableTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/python/tests/test_partitioned_table.py", line 73, in test_get_constituent
    self.assertIsNotNone(self.partitioned_table.get_constituent(keys))
AssertionError: unexpectedly None

----------------------------------------------------------------------
```

Adding some code to the Java side highlights that this is an issue with type conversions (Python types not matching the expected Java types). There is an additional mystery of why it fails for me, but not CI or @jmao-denver .

# Version Information:

Git SHA:
`ef8560504bd9b650a8e0717aed8aad93a74261d3`
(current head of `deephaven/deephaven-core:main`)

```
python3 -V -V
Python 3.8.9 (default, Apr 13 2022, 08:48:06)
[Clang 13.1.6 (clang-1316.0.21.2.5)]
```
```
java -version
openjdk version "11.0.14" 2022-01-18 LTS
OpenJDK Runtime Environment Zulu11.54+23-CA (build 11.0.14+9-LTS)
OpenJDK 64-Bit Server VM Zulu11.54+23-CA (build 11.0.14+9-LTS, mixed mode)
```
```
uname -a
Darwin megalodon.local 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:37 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T6000 arm64
```

# Java Patch:

The following Java patch makes the issue evident:
```
Index: engine/api/src/main/java/io/deephaven/engine/table/PartitionedTable.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/engine/api/src/main/java/io/deephaven/engine/table/PartitionedTable.java b/engine/api/src/main/java/io/deephaven/engine/table/PartitionedTable.java
--- a/engine/api/src/main/java/io/deephaven/engine/table/PartitionedTable.java	(revision ef8560504bd9b650a8e0717aed8aad93a74261d3)
[...]

PR #2478, opened 2022-06-06
Fix a JPY byte conv issue on arm64

jmao-denver:
C char type on arm64 is unsigned, resulting in negative Java byte values being created when converting a value in the range of 127 - 255

Fixes: #2469

### DIFF
--- a/py/jpy/src/main/c/jpy_jtype.c
+++ b/py/jpy/src/main/c/jpy_jtype.c
@@ -455,7 +455,7 @@ int JType_CreateJavaNumberFromPythonInt(JNIEnv* jenv, JPy_JType* type, PyObject*
     j = JPy_AS_JLONG(pyArg);
     i = (int) j;
     s = (short) j;
-    b = (char) j;
+    b = (signed char) j;
 
     if (i != j) {
         value.j = j;


devinrsmith (commented):
Brilliant! Assuming our M1 users can verify this fixes it, easy merge. I wonder if we should audit the codebase for other potentially unsafe casts?

py/jpy/src/main/c/jpy_jtype.c
```diff
@@ -455,7 +455,7 @@ int JType_CreateJavaNumberFromPythonInt(JNIEnv* jenv, JPy_JType* type, PyObject*
     j = JPy_AS_JLONG(pyArg);
     i = (int) j;
     s = (short) j;
-    b = (char) j;
+    b = (signed char) j;
```
rcaudy, line 458: Shouldn't we declare `b` with the same type? Do we need a similar fix in anything else?

jmao-denver: don't think that's needed, but it is good practice anyways. btw, char is the only integer type in C that can be either signed or unsigned by default, all the other types: short, int, long long are signed by default.

jmao-denver: I have already done a quick scan/audit of the code base, this appears the only place where we use C char type to convert to Java byte.

devinrsmith (approved):

rcaudy (approved):

Issue #2404, opened 2022-05-18
combined-javadoc fundamentally flawed


devinrsmith:
There's a fundamental flaw in how we do combined-javadoc, smashing all the compile classpaths together.

web-client-ide -> gwt -> jetty 9

our jetty server -> jetty 11

We've been lucky so far not to hit this wrt our jetty server since all of the classes we are using exist in jetty 9.

Demonstration in https://github.com/deephaven/deephaven-core/pull/2403

PR #2479, opened 2022-06-06
Make the hotspot-impl module JAVA_PUBLIC.

jcferretti:


devinrsmith:
Looks like the combined javadoc task is failing. This may be something we can work around, or it may be more fundamentally related to this https://github.com/deephaven/deephaven-core/issues/2404. I can help diagnose tomorrow.

PR #2480, opened 2022-06-07
DhFormulaColumn allow rowSet column name

nbauernfeind:
Although `rowSet` would be a rare / arbitrary column name, this PR enables `rowSet` and fails `__rowSet` earlier than it would have been caught.

Here's a failing query (fails on the `other` column in formula generation):
```groovy
t = emptyTable(1).update("rowSet = 0", "other = rowSet")
```

### DIFF
--- a/table-api/src/main/java/io/deephaven/api/util/NameValidator.java
+++ b/table-api/src/main/java/io/deephaven/api/util/NameValidator.java
@@ -89,7 +89,7 @@ public class NameValidator {
     }
 
     private static final Set<String> QUERY_LANG_RESERVED_VARIABLE_NAMES =
-            Stream.of("in", "not", "i", "ii", "k").collect(
+            Stream.of("in", "not", "i", "ii", "k", "__rowSet").collect(
                     Collectors.collectingAndThen(Collectors.toSet(), Collections::unmodifiableSet));
 
     public static String validateTableName(String name) {
@@ -340,7 +340,7 @@ public class NameValidator {
      * @param customReplace a function that is applied to the name before processing legality
      * @param takenNames the list of names that are already taken
      * @return whether the name is valid for a new table
-     * 
+     *
      */
     public static boolean isLegalTableName(String name, Function<String, String> customReplace,
             Set<String> takenNames) {
--- a/table-api/src/main/java/io/deephaven/api/util/NameValidator.java
+++ b/table-api/src/main/java/io/deephaven/api/util/NameValidator.java
@@ -89,7 +89,7 @@ public class NameValidator {
     }
 
     private static final Set<String> QUERY_LANG_RESERVED_VARIABLE_NAMES =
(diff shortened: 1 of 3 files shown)


table-api/src/main/java/io/deephaven/api/util/NameValidator.java
```diff
@@ -89,7 +89,7 @@ private String validate(String name) {
     }
 
     private static final Set<String> QUERY_LANG_RESERVED_VARIABLE_NAMES =
-            Stream.of("in", "not", "i", "ii", "k").collect(
+            Stream.of("in", "not", "i", "ii", "k", "__rowSet").collect(
```
rcaudy, line 92: >I checked Formula.java and rowSet appeared to be the only non-mangled item. Though it looks like __lazy __columnsToData and __params are also column names that would fail in formula generation. Those are from the constructor. The column name __temp0 fails which implies __temp[n] is a failing name. Should we add all of these as bad column names, or should we drop __rowSet from the list and just rely on the prefix being an obscure user choice?

I am inclined to take the "`__` is obscure" attitude, but I could see it both ways. I would either add the other items you noted, or take `__rowSet` out of this validation.

*If* we're going to maintain the list, It might be best to add:
1. A static list of internally-used parameters that will conflict with generated formulas (in `DhFormulaColumn` or similar)
2. A unit test of the validator (necessarily in `engine-table`) that verifies that those names aren't allowed.

Maybe there's a better way, involving banning all `__` names, but that feels draconian. Maybe we could use `__DEEPHAVEN_FORMULA_INTERNAL__` as a prefix and ban that?

table-api/src/main/java/io/deephaven/api/util/NameValidator.java
```diff
@@ -340,7 +340,7 @@ public static boolean isLegalTableName(String name, Function<String, String> rep
      * @param customReplace a function that is applied to the name before processing legality
      * @param takenNames the list of names that are already taken
      * @return whether the name is valid for a new table
-     * 
+     *
```
rcaudy, line 343: Might as well just delete this line.

table-api/src/main/java/io/deephaven/api/util/NameValidator.java
nbauernfeind: Yeah, I think `__DEEPHAVEN_FORMULA_INTERNAL__` as a prefix is the way to go if we end up needing to ban a prefix.

I think, for now, it makes sense to assume `__` is a sufficiently obscure prefix.

rcaudy (approved):
Should we audit filters, too?
```
io.deephaven.engine.table.impl.select.FormulaSample // you changed
io.deephaven.engine.table.impl.select.FormulaKernelSample // looks fine
io.deephaven.engine.table.impl.select.FilterKernelSample // does not
io.deephaven.engine.table.impl.select.FilterKernelArraySample // does not
```

nbauernfeind:
Sounds good, preparing another PR for those.

Issue #2214, opened 2022-04-06
Non-deterministic JS API behavior on plots with multiple series


mattrunyon:
**Description**

On a plot with multiple series, if the source tables are defined in the plot params, there is sometimes an error where the plot does not map the series to source table correctly.

This only seems to happen on multi-series plots, and it happens maybe 10% of the time. Happens with the legacy plot Python API and new plot API.

**Steps to reproduce**
1. Open a Python Web UI
2. Run this snippet (uses recent Python plot API)
```python
from deephaven import read_csv
from deephaven.plot.figure import Figure

source = read_csv("https://media.githubusercontent.com/media/deephaven/examples/main/MetricCentury/csv/metriccentury.csv")

figure = Figure()
plotSharedTwinX = figure.plot_xy(series_name = "Altitude", t = source.where(filters = ["SpeedKPH > 0"]), x = "Time", y = "AltitudeMeters")\
    .plot_xy(series_name = "Speed", t = source.where(filters = ["SpeedKPH > 0"]),  x = "Time",  y = "SpeedKPH")\
    .show()
```
3. Make sure the plot tab is visible
4. Re-run the snippet. If the plot stays visible, there is a chance it errors instead of showing the plot

**Expected results**

No errors on the plot

**Actual results**

Sometimes an error occurs when loading the plot. It can fail to find either column at times (sometimes fails to find `SpeedKPH` and sometimes `AltitudeMeters`

**Additional details and attachments**

Does not occur with only a single series
Does not seem to occur when `table.where` is moved outside the series params to its own variable

Stack trace that occurs when `session.getObject(plot)` is called. Colin and I set a breakpoint on line 28665 in `dh-core.js` to inspect the error (on edge containers)

```
 Error: java.util.NoSuchElementException: AltitudeMeters
        at NoSuchElementException_0.createError (<anonymous>:1190:10)
        at NoSuchElementException_0.initializeBackingError (<anonymous>:1210:46)
        at NoSuchElementException_0.Throwable_0 (<anonymous>:1155:8)
        at NoSuchElementException_0.Exception_1 (<anonymous>:1234:18)
        at NoSuchElementException_0.RuntimeException_1 (<anonymous>:1247:18)
        at new NoSuchElementException_0 (<anonymous>:3949:25)
        at Object.$findColumn_1 (<anonymous>:28665:16)
        at Object.$findColumn (<anonymous>:9920:16)
        at Object.$initColumnType (<anonymous>:27517:38)
[...]

PR #2485, opened 2022-06-07
Implement gRPC and JS API for PartitionedTable

### LINKED ISSUE #62: Implement PartitionedTables in the gRPC API
See #37 for approach to TableMaps (and the replacement ideas for SmartKeys).

The exact API has not yet been decided on, but details of ser/deser approach are documented on.

niloc132:
This also should include support for "plotby" and "one click" features. The JS Table method byExternal() still exists, but now returns a JsPartitionedTable, but the new method partitionBy() is preferred. The old TableMap type should be eventually removed, but still exists to make it easier to reference events consistently.

Will probably rewrite history at least one more on this, for easier review, breaking up the generated code changes, server changes, and js api changes.

Fixes #62


server/src/main/java/io/deephaven/server/partitionedtable/PartitionedTableServiceGrpcImpl.java
```diff
@@ -0,0 +1,126 @@
...
+    @Override
+    public void merge(MergeRequest request, StreamObserver<ExportedTableCreationResponse> responseObserver) {
+        GrpcUtil.rpcWrapper(log, responseObserver, () -> {
+            final SessionState session = sessionService.getCurrentSession();
+
+            SessionState.ExportObject<PartitionedTable> partitionedTable =
+                    ticketRouter.resolve(session, request.getPartitionedTable(), "partitionedTable");
+
+            session.newExport(request.getResultId(), "resultId")
+                    .require(partitionedTable)
```
rcaudy, line 70: We need to run this merge under the UGP shared lock if `PartitionedTable.table().isRefreshing()`.
Can we *conditionally* add `                    .requiresSerialQueue()` ?

server/src/main/java/io/deephaven/server/partitionedtable/PartitionedTableServiceGrpcImpl.java
```diff
@@ -0,0 +1,126 @@
...
+
+            SessionState.ExportObject<PartitionedTable> partitionedTable =
+                    ticketRouter.resolve(session, request.getPartitionedTable(), "partitionedTable");
+            SessionState.ExportObject<Table> keys =
+                    ticketRouter.resolve(session, request.getKeyTableTicket(), "keyTableTicket");
+
+            session.newExport(request.getResultId(), "resultId")
+                    // Due to the multiple operations running and the explicit size check, explicit contents read, we
+                    // run under the lock
+                    .requiresSerialQueue()
```
rcaudy, line 98: I sort of hate that we need to use `whereIn` for single get, thus requiring the serial queue and the UGP lock. Can we differentiate this a little bit? If the key table is static and only has one row, we could definitely use `constituentFor` which is concurrent-safe.

rcaudy (commented):
This review pass is confined to the server side and GRPC impl.

server/src/main/java/io/deephaven/server/partitionedtable/PartitionedTableServiceGrpcImpl.java
niloc132: Can't conditionally add it, but we can conditionally wrap the contents of the submit

server/src/main/java/io/deephaven/server/partitionedtable/PartitionedTableServiceGrpcImpl.java
niloc132: will drop this line - if the key table is static, we can validate if it is one row, then read the values out into Object[] and use that to read the constituent table. If not, we need the ugp and do the whereIn, as below

server/src/main/java/io/deephaven/server/partitionedtable/PartitionedTableServiceGrpcImpl.java
rcaudy: After our discussion, I think we agree that what we want to do is conditionally acquire the shared lock, but continue running this on the concurrent queue.

mofojed (changes requested):

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/partitionedtable.proto
```diff
@@ -0,0 +1,91 @@
...
+   * Transforms a table into a partitioned table, consisting of many separate table, each individually
+   * addressable. The result will be a FetchObjectResponse populated with a PartitionedTable.
+   */
+  rpc PartitionBy(PartitionByRequest) returns (PartitionByResponse);
+  /*
+   * Given a partitioned table, returns a table with the contents of all of the constituent tables.
+   */
+  rpc Merge(MergeRequest) returns (ExportedTableCreationResponse);
+  /*
+   * Given a partitioned table and a row indicated by position or by the values of a 1-row table,
```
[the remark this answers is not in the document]
niloc132: Yeah I removed that ability - do you think it is worth restoring, so that a sufficiently smart client could query by row key/position at all? Or just assume we want safer operations?

web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
```diff
@@ -0,0 +1,204 @@
...
+            requestMessage.setPartitionedTable(widget.getTicket());
+            requestMessage.setResultId(cts.getHandle().makeTicket());
+            connection.partitionedTableServiceClient().merge(requestMessage, connection.metadata(), c::apply);
+        }, "tablemap merged table")
+                .refetch(this, connection.metadata())
+                .then(cts -> Promise.resolve(new JsTable(cts.getConnection(), cts)));
+    }
+
+    public JsSet<Object> getKeys() {
+        if (keys.getColumns().length == 1) {
```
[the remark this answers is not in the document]
niloc132: We wrap them internally, but the TableMap api has historically allowed passing in a single string as distinct from a string in an array. For example, in DHE, these two statements do different things:

```
const t = await tablemap.getTable("i'm a string")
```
```
const t = await tablemap.getTable(["i'm an array of strings"])
```
and there are cases where one vs the other is correct and appropriate.

For this method, I was seeking to return a plain string if the partitioned table was only partitioned on a single (string) column, so that .getKeys() returns an set of strings in that case, rather than an set of array of strings. Note that in this version of the code, the API can tell the difference and decide what to do - in DHE we cannot tell the difference, since it is possible to have heterogeneous key types.

web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
```diff
@@ -0,0 +1,204 @@
...
+    @JsIgnore
+    public JsPartitionedTable(WorkerConnection connection, JsWidget widget) {
+        this.connection = connection;
+        this.widget = widget;
+    }
+
+    @JsIgnore
+    public Promise<JsPartitionedTable> refetch() {
+        if (keys != null) {
+            keys.close();
```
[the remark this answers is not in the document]
niloc132: refetch() isn't exported, and should only be called internally - we don't offer canceling refetch in other cases. The purpose of refetch is to be able to gracefully handle being reconnected, rather than getting the data again or something.

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -219,6 +222,7 @@ public WorkerConnection(QueryConnectable<?> info, Supplier<Promise<ConnectToken>
         inputTableServiceClient =
                 new InputTableServiceClient(info.getServerUrl(), JsPropertyMap.of("debug", debugGrpc));
         objectServiceClient = new ObjectServiceClient(info.getServerUrl(), JsPropertyMap.of("debug", debugGrpc));
+        partitionedTableServiceClient = new PartitionedTableServiceClient(info.getServerUrl(), JsPropertyMap.of("debug", debugGrpc));
```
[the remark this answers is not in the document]
niloc132: Probably, but the logging can be pretty noisy - the auth update logs would be every few seconds, and make it hard to tell what others were doing, so i usually enable only one at a time.

If you have a strong preference, I'll make these all have an empty map, and a specific "debug=true" object, rather than just default to false so any one boolean can be flipped.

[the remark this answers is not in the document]
mofojed: I don't have a strong preference.
I guess in JS I'd do something like:
```
const options = { debug: debugGrpc };
const inputTableClient = new ITSC(url, options);
const partitionedTableServiceClient = new PTSC(url, options);
... etc
```
Then if I wanted to flip just one:
```
const options = { debug: debugGrpc };
const inputTableClient = new ITSC(url, options);
const partitionedTableServiceClient = new PTSC(url, { ...options, debug: true });
... etc
```
Though I don't really know how best to translate that to Java land (`JsPropertyMap.of(options, "debug", debugGrpc)` or something?). And it's not that big a deal anyway.

rcaudy (dismissed):

Plot/src/main/java/io/deephaven/plot/datasets/multiseries/AbstractMultiSeries.java
```diff
@@ -48,17 +48,22 @@
     private transient Set<Object> seriesKeys;
     private transient Map<String, Object> seriesNames;
 
-    private final transient java.util.function.Function<Object[], String> DEFAULT_NAMING_FUNCTION = key -> {
+    private final transient java.util.function.Function<Object, String> DEFAULT_NAMING_FUNCTION = key -> {
```
rcaudy, line 51: There are a lot more changes to this than I would have expected, but maybe I'm failing to understand the nature of the issue.
I do see one error in the previous version, which is that I was using `Objects.toString(key)` instead of `Objects.toString(key[0])` in the unary case.
I considered including a `null` check, and rejected it because I didn't see an obvious code path to need it. If it's necessary, so be it.
I'm not sure why you've reverted the signature change.

plugin/figure/src/main/java/io/deephaven/figure/FigureWidgetTranslator.java
```diff
@@ -722,6 +724,7 @@ private SourceDescriptor makeSourceDescriptor(IndexableNumericData data, SourceT
 
             source.setColumnName(columnHandler.getColumnName());
             source.setTableId(tablePositionMap.get(columnHandler.getTableHandle()));
+            source.setPartitionedTableId(-1);
```
rcaudy, line 727: Is `-1` a canonical way to express "null"?

niloc132: It is what the API expects for "unset, don't try to index into the list of tickets with this" - PB returns zero for unset values, which clearly is a valid index, so -1 is "this is explicitly unset".

web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
```diff
@@ -920,6 +920,11 @@ public Promise<JsTable> naturalJoin(JsTable rightTable, JsArray<String> columnsT
                 .then(state -> Promise.resolve(new JsTable(workerConnection, state)));
     }
 
+    @JsMethod
+    public Promise<TableMap> byExternal(Object keys, @JsOptional Boolean dropKeys) {
```
mofojed, line 924: This doesn't match your PR description:
> The JS Table method byExternal() still exists, but now returns a JsPartitionedTable, but the new method partitionBy() is preferred.
It's still returning a TableMap, and the partitionBy looks like it just throws at this point. More changes not pushed?

mofojed (changes requested):

Plot/src/main/java/io/deephaven/plot/datasets/multiseries/AbstractMultiSeries.java
niloc132: Resolved by cherry-picking your commit that made this change.

niloc132:
Testing process:
### setup
```
from deephaven import *
from deephaven.plot.selectable_dataset import one_click

figure = Figure()
t = empty_table(100).update_view(formulas = ["X=i", "Z=i%5", "Y=X*X + Z * 10000", "Z = `` + Z"])
t2=t.update_view(formulas=["Y=Y+1000"])
```
### Verify a simple plot
 ```
 p = figure.plot_xy(series_name="z1", t=t.where("Z=`1`"), x="X", y="Y").show()
 ```
### Verify a simple ploy by
```
p = figure.plot_xy(series_name="partitioned plotting", t=t, by=["Z"], x="X", y="Y").show()

```
### Verify a chart with plain plot and plotby
```
p = (figure.plot_xy(series_name="z1", t=t2.where("Z=`1`"), x="X", y="Y")
.plot_xy(series_name="partitioned ", t=t, by=["Z"],x="X", y="Y")
).show()
```
### Verify a chart with multiple tables, multiple plotbys, repeatedly (ensure #2214 remains fixed)
```
p = (figure.plot_xy(series_name="z1", t=t2.where("Z=`1`"), x="X", y="Y")
.plot_xy(series_name="z2", t=t2.where("Z=`2`"), x="X", y="Y")
.plot_xy(series_name="z3", t=t2.where("Z=`3`"), x="X", y="Y")
.plot_xy(series_name="partitioned1 ", t=t, by=["Z"],x="X", y="Y")
.plot_xy(series_name="partitioned2 ", t=t.update_view("X=X+50"), by=["Z"],x="X", y="Y")
).show()
```

### Verify a oneclick works
```
p = figure.plot_xy(series_name="name", t=one_click(t=t, by=["Z"]), x="X", y="Y").show()
# after this appears, create an Input Filter from the Controls menu, and set the value to 1, 2, or 3
```

web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
niloc132: Right, not yet implemented, I'll get this in too, shouldnt take long to deal with these remaining bits. That will reduce TableMap.java to just a few event names, so that the web UI can still point to that class for names, but use the new (same api, different class) JsPartitionedTable.java.

web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
```diff
@@ -0,0 +1,202 @@
...
+            return w.getExportedObjects()[0].fetch();
+        }).then(result -> {
+            keys = (JsTable) result;
+            subscription = keys.subscribe(
+                    JsArray.asJsArray(keys.findColumns(descriptor.getKeyColumnNamesList().asArray(new String[0]))));
+            subscription.addEventListener(TableSubscription.EVENT_UPDATED, this::handleKeys);
+
+            LazyPromise<JsPartitionedTable> promise = new LazyPromise<>();
+            subscription.addEventListenerOneShot(TableSubscription.EVENT_UPDATED, data -> promise.succeed(this));
+            keys.addEventListener(JsTable.EVENT_DISCONNECT, promise::fail);
```
mofojed, line 94: Maybe fail with an appropriate error message here... and do you need to close the `subscription` here?

mofojed (changes requested):

py/jpy/build.gradle
```diff
@@ -1,3 +1,5 @@
+import nl.javadude.gradle.plugins.license.License
```
niloc132, line 1: this file should be removed from this review, will rebase it out

py/client/build.gradle
```diff
@@ -16,6 +16,10 @@ wheel {
     src 'tests'
 }
 
+license {
```
niloc132, line 19: this file should be removed from this review

web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
```diff
@@ -938,14 +949,40 @@ public Promise<TableMap> partitionBy(Object keys, @JsOptional Boolean dropKeys)
...
+            }
+            workerConnection.partitionedTableServiceClient().partitionBy(partitionBy, workerConnection.metadata(),
+                    c::apply);
+        });
+        // construct the partitioned table around the ticket created above
+        Promise<JsPartitionedTable> fetchPromise =
+                new JsPartitionedTable(workerConnection, new JsWidget(workerConnection, c -> {
+                    FetchObjectRequest partitionedTableRequest = new FetchObjectRequest();
+                    partitionedTableRequest.setSourceId(new TypedTicket());
+                    partitionedTableRequest.getSourceId().setType("PartitionedTable");
```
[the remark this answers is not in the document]
niloc132: JsVariableChanges.PARTITIONEDTABLE works great, i'm going to use that until we have a more coherent strategy.

web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
niloc132: Disconnect implies that we lost contact with the server in a way we couldnt fix, so the subscription is dead too.

Note that disconnect/reconnect is not implemented at all in DHC.

plugin/figure/src/main/java/io/deephaven/figure/FigureWidgetTranslator.java
```diff
@@ -669,12 +670,13 @@ private BoolMapWithDefault boolMapWithDefault(PlotUtils.HashMapWithDefault<? ext
...
 
-    private MultiSeriesSourceDescriptor makePartitionedTableSourceDescriptor(int plotHandleId, String columnName,
+    private MultiSeriesSourceDescriptor makePartitionedTableSourceDescriptor(PartitionedTableHandle plotHandle,
+            String columnName,
             SourceType sourceType, AxisDescriptor axis) {
         MultiSeriesSourceDescriptor.Builder source = MultiSeriesSourceDescriptor.newBuilder();
         source.setAxisId(axis.getId());
         source.setType(sourceType);
-        // TODO (deephaven-core#62): source.setPartitionedTableId(plotHandleId);
+        source.setPartitionedTableId(partitionedTablePositionMap.get(plotHandle));
```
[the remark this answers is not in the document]
niloc132: protobuf defaults define what the value will be if not explicitly set, and the default for an integer is 0, which is a valid value for a table/partitiontable reference, so we do need an explicit -1 here. I missed this one, thanks for pointing it out.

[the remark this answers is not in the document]
niloc132: Nope, nm, not missing after all - MultiSeriesSourceDescriptor doesnt have a tableId field.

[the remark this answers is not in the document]
niloc132: No reason not to use builder pattern, just was translating old beans to new pb builder objects as simply as possible. There is no perf impact on one vs the other approach.

mofojed (approved):

rcaudy (approved):

PR #2493, opened 2022-06-07
Disable intrinsic for Thread::currentThread

devinrsmith:
Allows us to remove all of the compiler excludes we had to previously maintain.

Investigated from https://github.com/openjdk/jdk/pull/9060

### DIFF
--- a/buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
+++ b/buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
@@ -90,45 +90,10 @@ tasks.withType(JavaCompile).configureEach {
   }
 }
 
+// Opting to leave dummy excludes in place.
+// A useful structuring that will get make it easier to re-introduce later if necessary.
 def tier4CompilationExcludes = [
-  // All ChunkedSumOperators
-  'io.deephaven.engine.table.impl.by.BigDecimalChunkedSumOperator':['addChunk'],
-  'io.deephaven.engine.table.impl.by.BigIntegerChunkedSumOperator':['addChunk'],
-  'io.deephaven.engine.table.impl.by.BooleanChunkedSumOperator':['addChunk'],
-  'io.deephaven.engine.table.impl.by.ByteChunkedSumOperator':['addChunk'],
-  'io.deephaven.engine.table.impl.by.CharChunkedSumOperator':['addChunk'],
-  'io.deephaven.engine.table.impl.by.DoubleChunkedSumOperator':['addChunk'],
-  'io.deephaven.engine.table.impl.by.FloatChunkedSumOperator':['addChunk'],
-  'io.deephaven.engine.table.impl.by.IntChunkedSumOperator':['addChunk'],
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
```diff
@@ -90,45 +90,10 @@ tasks.withType(JavaCompile).configureEach {
...
-  'io.deephaven.engine.table.impl.RightIncrementalChunkedAsOfJoinStateManager':['addToIndex', 'buildTable', 'decorationProbe'],
-  'io.deephaven.engine.table.impl.SymbolTableCombiner':['addToIndex', 'buildTable', 'decorationProbe'],
-  'io.deephaven.engine.table.impl.LeftOnlyIncrementalChunkedCrossJoinStateManager':['addToIndex', 'buildTable', 'decorationProbe'],
-  'io.deephaven.engine.table.impl.RightIncrementalChunkedCrossJoinStateManager':['addToIndex', 'buildTable', 'decorationProbe'],
-  'io.deephaven.engine.table.impl.StaticChunkedCrossJoinStateManager':['addToIndex', 'buildTable', 'decorationProbe'],
-  'io.deephaven.engine.table.impl.by.StaticChunkedOperatorAggregationStateManager':['addToIndex', 'buildTable', 'decorationProbe'],
-  'io.deephaven.engine.table.impl.by.IncrementalChunkedOperatorAggregationStateManager':['addToIndex', 'buildTable', 'decorationProbe'],
-
-  'io.deephaven.engine.table.impl.by.AddOnlyFirstOrLastChunkedOperator':['addChunk', 'updateRedirections']
+  'io.deephaven.ClassToExclude':['methodToExclude'],
```
devinrsmith, line 96: Note: some of the downstream tasks fail if the excludes list is empty, so I figured leaving a dummy class / method was OK.

rcaudy:
For posterity, if we re-add the excludes we should also do:
```
C2 CompilerThread0   329067 20413       4       io.deephaven.engine.table.impl.ssa.NullAwareCharSsaSsaStamp::processEntry (259 bytes)
C2 CompilerThread0   530747 22459 % !   4       io.deephaven.engine.table.impl.ssa.IntSsaSsaStamp::processRemovals @ 42 (271 bytes)
```

rcaudy (dismissed):

buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
niloc132: Probably a dumb oversight on my part when I first set it up, might make more sense to just remove this? We'll have it in history if we need it back again later.

Maybe we continue to create an empty json file, so we can add more to it, but not worry about saying "turn off c2 for all of these things: <empty list>"?

niloc132 (commented):
Please update py/embedded-server/deephaven_server/start_jvm.py too - we can leave the directives file as you've done here, but to add the DisableIntrinsic arg.

niloc132 (approved):

Issue #2471, opened 2022-06-02
Object based configuration


devinrsmith:
It would be nice to have structured objects serve as the basis for configurable properties. This makes them easier to organize, document, reference, (de)serialize, re-use, and use with dependency injection.

As it is today, configuration is coded on a case-by-case basis against a flat keyspace with an implied hierarchy. A variety of constructs have been created to "make it easy": `get<Type>ForClassWithDefault`, `get<Type>WithDefault`, `get<Type>`, `getIntegerArray`, `getStringArrayFromProperty`; but these are incomplete and leave a lot to be desired. Often times code that needs configuration will have their own prefix and looping logic, essentially making them responsible for deserialization logic.

As something concrete, I think we should move towards using the Immutables library to get code generation for immutable configuration objects. This also allows us to get built-in support for Jackson or GSON to get deseralization for "free". (Jackson can use JSON, YAML, TOML, XML, or likely other configuration formats if desired.)

This isn't a proposal to remove Configuration - that would be a much larger issue. Most immediately, we'd do away with .prop files, and rely on .json (or other hierarchical format); Configuration would still be able to resolve from the new format. The jetty / netty main entry points would be the first to get new configuration objects, but hopefully it would also give us the opportunity to consolidate application mode configuration into the fold (instead of continuing to have that be a different thing to configure).

One of the challenges might be "realizing" our configuration dependencies - up until now, we've been able to rely on a "bag" that anything in the code base could access. If we moved to a typed configuration system, it forces us to design the inter-relationship between our components in a cleaner manner. But also, some components may be optional, so from a compile-time standpoint, we can't have a hard-dependency on it - how will those pieces of the system get configured?

Relates to #21

Partial sketches / ideas in #2401, #2417

devinrsmith:
There might also be benefits when configuring python if there are java objects that can be wrapped, or represented as dataclasses in python.

PR #2494, opened 2022-06-07
Add SSL Configuration for Netty, Jetty, and Embedded

devinrsmith:
The minimum-viable configuration for enabling SSL.

Can run a server w/ self-signed certs:

```
./gradlew server-jetty-app:run -Pgroovy -PdevCerts
```
or
```
./gradlew server-netty-app:run -Pgroovy -PdevCerts
```

Can verify the the server via:
```
./java-client/session-examples/build/install/java-client-session-examples/bin/connect-check --target dh://localhost:8443 --ssl server/dev-certs/client.json
```

SSL configuration can be applied manually with the properties "ssl.identity.type", "ssl.identity.certChainPath", and
"ssl.identity.privateKeyPath".


server/netty/src/main/java/io/deephaven/server/netty/NettyConfig.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * The netty server configuration.
+ */
+@Immutable
+@BuildableStyle
+public abstract class NettyConfig implements ServerConfig {
+
+    public static final int DEFAULT_SSL_PORT = 443;
+    public static final int DEFAULT_PLAINTEXT_PORT = 8080;
```
niloc132, line 16: why force 8080, why not the actual default plaintext port of 80?

if we allow port numbers <1024 (i.e. 443), we already require that we can run as root, so 80 should be fine?

server/jetty/src/main/java/io/deephaven/server/jetty/JettyServerComponent.java
```diff
@@ -12,14 +14,18 @@
 @Singleton
 @Component(modules = {
         DeephavenApiServerModule.class,
+        DeephavenApiConfigModule.class,
         PythonGlobalScopeCopyModule.class,
         HealthCheckModule.class,
         PythonPluginsRegistration.Module.class,
-        JettyServerModule.class
+        JettyServerModule.class,
```
niloc132, line 21: like this comma ;)

server/jetty/src/main/java/io/deephaven/server/jetty/JettyMain.java
```diff
@@ -2,28 +2,50 @@
...
 
-        // defaults to 5 minutes
-        int httpSessionExpireMs = config.getIntegerWithDefault("http.session.durationMs", 300000);
-        int httpPort = config.getIntegerWithDefault("http.port", 10000);
-        int schedulerPoolSize = config.getIntegerWithDefault("scheduler.poolSize", 4);
-        int maxInboundMessageSize = config.getIntegerWithDefault("grpc.maxInboundMessageSize", 100 * 1024 * 1024);
+        // Defaults defined in JettyConfig
+        int httpSessionExpireMs = config.getIntegerWithDefault("http.session.durationMs", -1);
+        String httpHost = config.getStringWithDefault("http.host", null);
+        int httpPort = config.getIntegerWithDefault("http.port", -1);
```
niloc132, line 20: I am leery of divorcing the default configs from either the .prop file or the actual config.getFoo() invocation, at least as a matter of changing conventions.

py/embedded-server/java-runtime/src/main/java/io/deephaven/python/server/EmbeddedServer.java
```diff
@@ -48,16 +57,34 @@ interface Builder extends DeephavenApiServerComponent.Builder<PythonServerCompon
...
+            builder.host(httpHost);
+        }
+        if (schedulerPoolSize > -1) {
+            builder.schedulerPoolSize(schedulerPoolSize);
+        }
+        if (maxInboundMessageSize > -1) {
+            builder.maxInboundMessageSize(maxInboundMessageSize);
+        }
+        if (httpWebsockets != null) {
+            builder.websockets(Boolean.parseBoolean(httpWebsockets));
```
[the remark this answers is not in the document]
devinrsmith: The defaults defined by the server object - in this example, `websockets` defaults to *true*, so `Boolean.parseBoolean(null)` would be the wrong behavior.

server/jetty-app/build.gradle
```diff
@@ -27,7 +27,7 @@ def extraJvmArgs = [
         '-XX:+UseG1GC',
         '-XX:MaxGCPauseMillis=100',
         '-XX:+UseStringDeduplication',
-        '-XshowSettings:vm',
+        '-XshowSettings:vm'
```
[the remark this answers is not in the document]
devinrsmith: Didn't mean to remove, will re-add.

server/netty/src/main/java/io/deephaven/server/netty/NettyConfig.java
devinrsmith: I chose it this way to leave the current default behavior the same and working out-of-the-box for localhost - no config -> plaintext + 8080. ssl without an explicit port -> ssl + 443.

I think when you are going through the act of ssl configuration, you'll be aware that you need privs to run at 443 (or, you'll know soon enough after startup), or you can do the small additional configuration for a port >= 1024.

server/jetty/src/main/java/io/deephaven/server/jetty/JettyMain.java
devinrsmith: I don't think there is much of a coherent convention in how we currently use config. At least, in a lot of cases we tell the "objects" that they are responsible themselves for decoded how to parse their own properties...

I'm hoping this PR will at least present some code that we can agree is "better than before", and I think the code can stand alone (but is a nod towards https://github.com/deephaven/deephaven-core/issues/2471). For example, nowhere do we document that you can configure the property `scheduler.poolSize`, nor that the default is 4.

I will concede though, due to this prompting, I think the code can be consolidated and documented better, which I will do... hoping to motivate more that it is "better than before".

server/jetty/build.gradle
```diff
@@ -11,15 +11,29 @@ dependencies {
...
 
-    implementation('org.eclipse.jetty:jetty-servlet:11.0.8')
-    implementation 'org.eclipse.jetty:jetty-webapp:11.0.8'
-    implementation 'org.eclipse.jetty.http2:http2-server:11.0.8'
+    implementation 'org.eclipse.jetty:jetty-servlet'
+    implementation 'org.eclipse.jetty:jetty-webapp'
+    implementation 'org.eclipse.jetty.http2:http2-server'
+    implementation 'org.eclipse.jetty:jetty-alpn-server'
+
+    // Todo: is this still true w/ Java 11+, jetty 11.0.x?
```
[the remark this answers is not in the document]
devinrsmith: I think we've got some good runway until we are past saturating 1Gib. I've collected details in https://github.com/deephaven/deephaven-core/issues/2506.

server/src/main/java/io/deephaven/server/config/ServerConfig.java
```diff
@@ -11,22 +13,91 @@
...
+        if (maxInboundMessageSize != -1) {
+            builder.maxInboundMessageSize(maxInboundMessageSize);
+        }
+        Main.parseSSLConfig(config).ifPresent(builder::ssl);
+        return builder;
     }
 
+    /**
+     * The network interface this server binds to as an IP address or a hostname. If not set, then bind to all
+     * interfaces.
```
niloc132, line 97: stupid/pedantic question: will this bind to ipv6 interfaces as part of "all of the interfaces"? or do we need to be more explicit about that. I seem to recall that in java >=9 you can specify `*` in some spots, but this might not be one of them...

niloc132 (approved):

server/src/main/java/io/deephaven/server/config/ServerConfig.java
devinrsmith: I've tested both netty and jetty - they bind ipv4 and ipv6 as part of "all interfaces". Didn't look into wildcard `*` at all - if we want some alternative syntaxes we should be able to adapt them.

PR #2499, opened 2022-06-08
Missing licenses

### LINKED ISSUE #404: automate / check copyright header
We've got the current gradle task "licenseFormat" that applies license-header. There may be issues with this plugin. We may be able to replace w/ spotless if necessary.

Do we need to exclude certain paths?

niloc132:
This is the "readable" version of this patch, and the first to land to lock us in with correct license headers going forward.

The next PR will include all of the source files (~7700) to change to correctly reflect these changes, the result of calling `./gradlew licenseFormat`.

After that, the third commit will make the build fail, by flipping `ignoreFailures` to false after all other changes have been made.

This update addresses python, cpp, java, and proto files (which are in a java project).

Fixes #404?

### DIFF
--- a/Generators/src/main/java/io/deephaven/libs/GroovyStaticImportGenerator.java
+++ b/Generators/src/main/java/io/deephaven/libs/GroovyStaticImportGenerator.java
@@ -264,9 +264,9 @@ public class GroovyStaticImportGenerator {
 
     private String generateCode() {
 
-        String code = "/*\n" +
-                " * Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending\n" +
-                " */\n\n" +
+        String code = "/**\n" +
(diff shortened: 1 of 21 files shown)


devinrsmith (commented):
Applying `./gradlew licenseFormat` to your branch, I see:

```
 189 files changed, 435 insertions(+), 127 deletions(-)
```

Looks like it's not formatting .java files w/ this patch (ie, not 7k files)?

I'm a little concerned about changing generated files:

py/client/pydeephaven/proto/*
cpp-client/deephaven/client/proto/*
cpp-client/deephaven/client/flatbuf/*

niloc132:
Whoops, good catch, I made a typo before I understood how the license plugin(s) interacted with sourcesets. Will have the fix up in a minute...

devinrsmith (approved):

devinrsmith:
We should probably also look into updating to the latest version of the plugin.

PR #2502, opened 2022-06-08
Port MovingAverage as-is from legacy to v2

### LINKED ISSUE #2386: Wrap the moving averages in deephaven v2
The MovingAverages is supported in legacy and documented. It needs to be migrated to the new API.

```
from deephaven_legacy.MovingAverages import ByEmaSimple
import random

ema1sec = ByEmaSimple('BD_SKIP','BD_SKIP','TIME',10,'SECONDS', type='LEVEL')

result = time_table('00:00:01').update(formulas=["Sym = (char)(int)random.randint(65, 70)", "Numbers = (int)random.randint(0, 100)"])

ema_data = result.view(formulas=["EMA_data = ema1sec.update(Timestamp, Numbers)"]).last_by()
ema_data_grouped = result.view(formulas=["Sym", "EMA_data = ema1sec.update(Timestamp, Numbers, Sym)"]).last_by(by=["Sym"]).sort(order_by=["Sym"])
```
It seems to me that during formula evaluation this will lead to a round-trip from Java to Python then back just to call 'ema1sec.update' if we wrap the result of 'ByEmaSimple' call in Python. What we want is to unwrap  'ema1sec' during formula parsing and generate evaluation code with the wrapped Java object.

The more generalized solution for minimizing Python/Java boundary crossing is to be addressed by #2363

jmao-denver:
Skipped creating a unit test for it but tested with the example in the docs.
```
from deephaven.experimental.ema import ByEmaSimple
import random

def random_char():
    return random.choice(['A', 'B', 'C', 'D', 'E', 'F'])

ema1sec = ByEmaSimple('BD_SKIP','BD_SKIP','TIME',10,'SECONDS', type='LEVEL')

result = time_table('00:00:01').update(formulas=["Sym = (String)random_char()", "Numbers = (int)random.randint(0, 100)"])

ema_data = result.view(formulas=["EMA_data = ema1sec.update(Timestamp, Numbers)"]).last_by()
ema_data_grouped = result.view(formulas=["Sym", "EMA_data = ema1sec.update(Timestamp, Numbers, Sym)"]).last_by(by=["Sym"]).sort(order_by=["Sym"])
```

Fixes #2386

### DIFF
--- a/py/server/deephaven/experimental/ema.py
+++ b/py/server/deephaven/experimental/ema.py
@@ -1,0 +1,106 @@
+#
+#     Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+#
+"""This modules allows users to define exponential moving averages that can be used in the Deephaven query language."""
+import numpy
+import jpy
+
+_JAbstractMAType = jpy.get_type('io.deephaven.numerics.movingaverages.AbstractMa$Type')
+_JAbstractMATMode = jpy.get_type('io.deephaven.numerics.movingaverages.AbstractMa$Mode')
+_JBadDataBehavior = jpy.get_type('io.deephaven.numerics.movingaverages.ByEma$BadDataBehavior')
+_JTimeUnit = jpy.get_type('java.util.concurrent.TimeUnit')
+_JByEmaSimple = jpy.get_type('io.deephaven.numerics.movingaverages.ByEmaSimple')
+_JEma = jpy.get_type('io.deephaven.numerics.movingaverages.Ema')
+_JEmaArray = jpy.get_type('io.deephaven.numerics.movingaverages.EmaArray')
+_JExponentiallyDecayedSum = jpy.get_type('io.deephaven.numerics.movingaverages.ExponentiallyDecayedSum')
+
+
+def _convertEnumBehavior(value, enumType):
+    if value is None or isinstance(value, enumType):
+        return value
+    elif hasattr(enumType, value):
+        return getattr(enumType, value, None)
+    return value
+
+
+def ByEmaSimple(nullBehavior, nanBehavior, mode, timeScale, timeUnit, type=None):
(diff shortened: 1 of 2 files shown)


py/server/deephaven/experimental/ema.py
```diff
@@ -0,0 +1,106 @@
...
+
+def _convertEnumBehavior(value, enumType):
+    if value is None or isinstance(value, enumType):
+        return value
+    elif hasattr(enumType, value):
+        return getattr(enumType, value, None)
+    return value
+
+
+def ByEmaSimple(nullBehavior, nanBehavior, mode, timeScale, timeUnit, type=None):
```
chipkent, line 26: There are a lot of non-pep8 names.  Should these be changed for this case, or not?  I can agree to not, since this is in experimental and it is just a stop gap, but making it standards compliant is also a thing.  Just a note.  I don't have a view in this case.

py/server/deephaven/experimental/ema.py
```diff
@@ -0,0 +1,106 @@
...
+    """Constructor for an engine aware Exponential Moving Average (EMA) which performs a `groupBy` ema calculation without
```
chipkent, line 27:
```suggestion
    """Constructor for an engine aware Exponential Moving Average (EMA) which performs a `group_by` ema calculation without
```

chipkent (approved):

py/server/deephaven/experimental/ema.py
jmao-denver: Yeah, it was a deliberate decision to make it obvious that these are not  fully wrapped final version.

py/server/deephaven/experimental/ema.py
jmao-denver: I think groupBy is OK because we are talking about the operation on the server.

PR #2503, opened 2022-06-09
Remove deephaven-legacy and related code - 1st attempt, no reorg

jmao-denver:
Fixes: 2498


Integrations/README.md
```diff
@@ -1,11 +0,0 @@
-### Building wheel with docker
```
rcaudy, line 1: I'm concerned that we should just delete this file, rather than leaving it empty.

rcaudy (approved):
I would love to merge this, FWIW.

niloc132 (commented):
Please update .github/workflows/docs-ci.yml to no longer copy these legacy docs over - check with @dsmmcken how we want to leave those legacy docs up, or if we just want to take them all down right away, how that rsync should be tweaked.

niloc132:
I vote don't worry about the testing cleanup in Integrations/, we can do a followup for that.

devinrsmith (approved):
Happy to merge, see Colin's responses first.

jakemulf (dismissed):

rcaudy (dismissed):

rcaudy (approved):

PR #2508, opened 2022-06-10
Update license headers

niloc132:


devinrsmith (commented):
I think doing a best-effort on removing changes to generated files. Anything that has generated in the path, or has "generated" in the source file.

niloc132:
I didn't remove all "generated" headers, but most of them, and also excluded Container, jpy and grpc-java as well, as those are external projects.

I am not certain that we want to exclude generated across the board, as we do claim copyright to those files.

devinrsmith (approved):

PR #2516, opened 2022-06-10
Add `updateBy` functionality to Deephaven Community

lbooker42:


engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -3048,6 +3021,14 @@ public <R> R memoizeResult(MemoizedOperationKey memoKey, Supplier<R> operation)
         }
     }
 
+    @Override
+    public Table updateBy(@NotNull final UpdateByControl control,
+            @NotNull final Collection<UpdateByClause> ops,
+            @NotNull final MatchPair... byColumns) {
+        return QueryPerformanceRecorder.withNugget("copy()", sizeForInstrumentation(),
```
devinrsmith, line 3028: Bad nugget string.

engine/api/src/main/java/io/deephaven/engine/table/ColumnUpdateClause.java
```diff
@@ -0,0 +1,30 @@
+package io.deephaven.engine.table;
+
+import io.deephaven.engine.table.updateBySpec.UpdateBySpec;
+import org.jetbrains.annotations.NotNull;
+
+public class ColumnUpdateClause implements UpdateByClause {
+    private final UpdateBySpec spec;
+    @NotNull
+    private final String[] columns;
```
devinrsmith, line 9: Are these pure column names? Or, can we expect expressions here?

Might be tighter to use `io.deephaven.api.ColumnName`. Also, if making immutable, should be `List` instead of `[]`.

engine/table/src/main/java/io/deephaven/engine/table/impl/UpdateByOperatorFactory.java
```diff
@@ -0,0 +1,356 @@
...
+        public void visit(ColumnUpdateClause clause) {
+            final MatchPair[] pairs = parseMatchPairs(clause.getColumns());
+            if (pairs.length == 0) {
+                columnStr = "[All]";
+            } else {
+                columnStr = MatchPair.matchString(pairs);
+            }
+
+            descriptionBuilder.append(clause.getSpec().describe()).append("(").append(columnStr).append("), ");
+            columnStr = null;
```
devinrsmith, line 122: This is an odd usage. IMO, `columnStr` should be local to the method.

engine/table/src/main/java/io/deephaven/engine/table/impl/SimpleSourceTable.java
```diff
@@ -60,4 +65,12 @@ protected final Table redefine(TableDefinition newDefinitionExternal, TableDefin
         deferredViewTable.setRefreshing(isRefreshing());
         return deferredViewTable;
     }
+
+    @Override
+    public Table updateBy(@NotNull final UpdateByControl control,
+            @NotNull final Collection<UpdateByClause> ops,
+            @NotNull final MatchPair... byColumns) {
+        return QueryPerformanceRecorder.withNugget("copy()", sizeForInstrumentation(),
```
devinrsmith, line 73: bad nugget name.

engine/table/src/main/java/io/deephaven/engine/table/impl/TableWithDefaults.java
```diff
@@ -1113,6 +1113,110 @@ default Table rollup(Collection<? extends Aggregation> aggregations, boolean inc
...
+    // -----------------------------------------------------------------------------------------------------------------
+
+    /**
+     * <p>
+     * Create a table with the same {@link #getRowSet() rowSet} as it's parent that will perform the specified set of
+     * row based operations to it. As opposed to {@link #update(String...)} these operations are more restricted but are
+     * capable of processing state between rows. This operation will group the table by the specified set of keys if
+     * provided before applying the operation.
+     * </p>
+     *
```
devinrsmith, line 1128: missing control param.

engine/table/src/main/java/io/deephaven/engine/table/impl/TableWithDefaults.java
```diff
@@ -1113,6 +1113,110 @@ default Table rollup(Collection<? extends Aggregation> aggregations, boolean inc
...
+     * @param operations the operations to apply to the table.
+     * @param byColumns the columns to group by before applying.
+     * @return a table with the same index, with the specified operations applied to each group defined by the
+     *         {@code byColumns}
+     */
+    @ConcurrentMethod
+    Table updateBy(@NotNull final UpdateByControl control, @NotNull final Collection<UpdateByClause> operations,
+            final MatchPair... byColumns);
```
devinrsmith, line 1136: If these are true MatchPair, we might prefer to use the type `Collection<? extends Selectable> groupByColumns`?

engine/table/src/main/java/io/deephaven/engine/table/impl/UpdateByOperatorFactory.java
```diff
@@ -0,0 +1,356 @@
...
+
+            if (isTimeBased) {
+                timeStampRecorder = makeLongRecordingOperator(source, timestampCol);
+                ops.add(timeStampRecorder);
+            } else {
+                timeStampRecorder = null;
+            }
+
+            Arrays.stream(pairs)
+                    .filter(p -> !isTimeBased || !p.right().equals(timestampCol))
```
devinrsmith, line 163: This equals is between incompatible types - makes me think that ema.getTimestampCol() should be ColumnName.

engine/table/src/main/java/io/deephaven/engine/table/impl/UpdateByOperatorFactory.java
```diff
@@ -0,0 +1,356 @@
...
+            pairs = createColumnsToAddIfMissing(source, parseMatchPairs(clause.getColumns()), spec, groupByColumns);
+            spec.walk(this);
+            pairs = null;
+        }
+
+        @Override
+        public void visit(@NotNull final EmaSpec ema) {
+            final String timestampCol = ema.getTimestampCol();
+            final LongRecordingUpdateByOperator timeStampRecorder;
+            final boolean isTimeBased = !StringUtils.isNullOrEmpty(timestampCol);
```
devinrsmith, line 153: Why even allow empty column name?

devinrsmith (commented):
Skipped a lot of the implementation. I think we should consider chatting about Immutables when you have time.

engine/table/src/main/java/io/deephaven/engine/table/impl/UpdateByOperatorFactory.java
lbooker42: `null` timestampCol forces tick-based calculation, empty string is undefined behavior but rather than throw an error just treat it as a tick-based EMA.

engine/api/src/main/java/io/deephaven/engine/table/ColumnUpdateClause.java
lbooker42: I think we could use `ColumnName` but I can't manage getting the namespaces sorted to use them

engine/api/src/main/java/io/deephaven/engine/table/EmaControl.java
```diff
@@ -21,75 +24,64 @@
...
-    public final BadDataBehavior onNegativeDeltaTime;
-    public final MathContext bigValueContext;
-
+@Immutable
+@BuildableStyle
+public abstract class EmaControl {
     @NotNull
-    public static Builder get() {
-        return new Builder();
+    public static ImmutableEmaControl.Builder builder() {
```
devinrsmith, line 31: You should return `Builder` here not the package private version.

engine/api/src/main/java/io/deephaven/engine/table/EmaControl.java
```diff
@@ -21,75 +24,64 @@
...
-            BadDataBehavior onZeroDeltaTime, BadDataBehavior onNegativeDeltaTime,
-            @NotNull final MathContext bigValueContext) {
-        this.onNullValue = onNullValue;
-        this.onNanValue = onNanValue;
-        this.onNullTime = onNullTime;
-        this.onZeroDeltaTime = onZeroDeltaTime;
-        this.onNegativeDeltaTime = onNegativeDeltaTime;
-        this.bigValueContext = bigValueContext;
-    }
+    public static EmaControl DEFAULT = ImmutableEmaControl.builder().build();
```
devinrsmith, line 35: final

engine/api/src/main/java/io/deephaven/engine/table/updateBySpec/CumProdSpec.java
```diff
@@ -1,20 +1,18 @@
...
 import io.deephaven.util.type.TypeUtils;
+import org.immutables.value.Value.Immutable;
 import org.jetbrains.annotations.NotNull;
 
 /**
  * A {@link UpdateBySpec} for performing a Cumulative Product of the specified columns.
  */
+@Immutable
+@SimpleStyle
 public class CumProdSpec implements UpdateBySpec {
```
devinrsmith, line 13: I'm a little surprised this worked - but CumProdSpec should be abstract as well.

engine/api/src/main/java/io/deephaven/engine/table/updateBySpec/CumSumSpec.java
```diff
@@ -1,20 +1,18 @@
...
 import io.deephaven.util.type.TypeUtils;
+import org.immutables.value.Value.Immutable;
 import org.jetbrains.annotations.NotNull;
 
 /**
  * A {@link UpdateBySpec} for performing a Cumulative Sum of the specified columns.
  */
+@Immutable
+@SimpleStyle
 public class CumSumSpec implements UpdateBySpec {
```
devinrsmith, line 13: Abstract.

engine/api/src/main/java/io/deephaven/engine/table/updateBySpec/EmaSpec.java
```diff
@@ -1,61 +1,46 @@
...
             @NotNull final String timestampCol,
             long timeScaleNanos) {
-        return new EmaSpec(control, timestampCol, timeScaleNanos);
+        return ImmutableEmaSpec.of(control, timestampCol, timeScaleNanos);
     }
 
     public static EmaSpec ofTicks(@NotNull EmaControl control, long tickWindow) {
-        return new EmaSpec(control, null, tickWindow);
+        return ImmutableEmaSpec.of(control, null, tickWindow);
     }
```
devinrsmith, line 25: I see 3 different variable names: `timeScaleNanos`, `tickWindow` and `timeScaleUnits`. Are these all the same thing?

engine/api/src/main/java/io/deephaven/engine/table/updateBySpec/EmaSpec.java
```diff
@@ -1,61 +1,46 @@
...
-    public String getTimestampCol() {
-        return timestampCol;
-    }
-
-    public long getTimeScaleUnits() {
-        return timeScaleUnits;
-    }
+    @Parameter
+    @Nullable
+    public abstract String timestampCol();
```
devinrsmith, line 32: IMO, this would be better as an Optional, and a BuildableStyle instead of SimpleStyle.

engine/api/src/main/java/io/deephaven/engine/table/updateBySpec/EmaSpec.java
```diff
@@ -1,61 +1,46 @@
...
-    @Override
-    public String describe() {
-        return "Ema(" + (timestampCol == null ? "TickBased" : timestampCol) + ", " + timeScaleUnits + ", " + control
-                + ")";
-    }
+    @Parameter
+    public abstract long timeScaleUnits();
 
     @Override
     public boolean applicableTo(@NotNull Class<?> inputType) {
```
devinrsmith, line 38: I find it useful to mark all methods implemented as final - not because I would ever expect Immutables to code-generate/overwrite them, but there may be some bytecode benefits. (I see you've taken that pattern that we've taken before w/ `walk`)

engine/api/src/main/java/io/deephaven/engine/table/updateBySpec/FillBySpec.java
```diff
@@ -1,19 +1,18 @@
...
+import org.immutables.value.Value.Immutable;
 import org.jetbrains.annotations.NotNull;
 
 /**
  * A {@link UpdateBySpec} for performing a forward fill of the specified columns.
  */
+
+@Immutable
+@SimpleStyle
 public class FillBySpec implements UpdateBySpec {
```
devinrsmith, line 13: abstract

engine/table/src/main/java/io/deephaven/engine/table/impl/UpdateByControl.java
```diff
@@ -19,13 +24,18 @@ public interface UpdateByControl {
     double DEFAULT_MAXIMUM_STATIC_SPARSE_MEMORY_OVERHEAD =
             Configuration.getInstance().getDoubleWithDefault("UpdateByControl.maximumStaticMemoryOverhead", 1.1);
 
-    UpdateByControl DEFAULT = new UpdateByControl() {};
+    static ImmutableUpdateByControl.Builder builder() {
```
devinrsmith, line 27: We should be creating a public version of this interface and returning it instead.

engine/table/src/main/java/io/deephaven/engine/table/impl/UpdateByOperatorFactory.java
devinrsmith: You could add a check and disallow the creation of empty timestamp cols. I find doing validation as early as possible is better than trying to have every single consumer responsible for these special empty checks.

devinrsmith (commented):
I think we still need to consider `Selectable` instead of `MatchPair`... given that any thought?

engine/api/src/main/java/io/deephaven/engine/table/updateBySpec/EmaSpec.java
```diff
@@ -1,46 +1,105 @@
...
+    }
 
     @Parameter
-    @Nullable
-    public abstract String timestampCol();
+    public abstract EmaControl control();
 
     @Parameter
-    public abstract long timeScaleUnits();
+    public abstract TimeScale timeScale();
```
devinrsmith, line 87: Everything in the hierarchy of an immutable should (ideally) be immutable itself. I see that TimeScale is not - and does not implement hashcode, equals, etc. 

I think Timescale should be an interface w/ visitor pattern, and two impls TimescaleTicks, TimescaleDuration.

lbooker42: This is used internally, never exposed to the user.  Is there a need for the complexity of a class hierarchy?

engine/api/src/main/java/io/deephaven/engine/table/UpdateByClause.java
```diff
@@ -0,0 +1,194 @@
...
+    static UpdateByClause ema(@NotNull final String timestampColumn, final long timeScaleNanos,
+            @NotNull final EmaControl control, final String... columns) {
+        return of(EmaSpec.ofTime(control, timestampColumn, timeScaleNanos), columns);
+    }
+    // endregion
+
+    <V extends Visitor> V walk(final @NotNull V v);
+
+    interface Visitor {
+        void visit(ColumnUpdateClause clause);
```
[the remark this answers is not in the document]
lbooker42: A major change coming is windowing support.  Not sure yet if that will be as new clauses or modifications to the current ones.

table-api/src/main/java/io/deephaven/api/updateBy/spec/CumProdSpec.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    @Override
+    public final boolean applicableTo(Class<?> inputType) {
+        boolean isPrimitiveNumeric = inputType.equals(double.class) || inputType.equals(float.class)
+                || inputType.equals(int.class) || inputType.equals(long.class) || inputType.equals(short.class)
+                || inputType.equals(byte.class);
+        boolean isBoxedNumeric = Number.class.isAssignableFrom(inputType);
+
+        return isPrimitiveNumeric || isBoxedNumeric;
```
devinrsmith, line 23: By separating as two booleans, we don't get lazy evaluation (at least, not without JVM optimization) - so we always call isAssignableFrom. Similar elsewhere.

devinrsmith (changes requested):

table-api/src/main/java/io/deephaven/api/updateBy/spec/CumMinMaxSpec.java
```diff
@@ -0,0 +1,42 @@
...
+    @Override
+    public final boolean applicableTo(Class<?> inputType) {
+        return
+        // is primitive numeric?
+        inputType.equals(double.class) || inputType.equals(float.class)
+                || inputType.equals(int.class) || inputType.equals(long.class) || inputType.equals(short.class)
+                || inputType.equals(byte.class)
+
+                // is boxed numeric?
+                || Number.class.isAssignableFrom(inputType)
```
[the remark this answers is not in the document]
lbooker42: yes, any object that implements `compareTo()` (extends `Comparable`) will work. See `ComparableCumMinMaxOperator.java` for impl.

engine/table/src/main/java/io/deephaven/engine/table/impl/TableWithDefaults.java
```diff
@@ -1112,6 +1113,145 @@ default Table rollup(Collection<? extends Aggregation> aggregations, boolean inc
...
+     * </p>
+     *
+     * @param control the {@Link UpdateByControl control} to use when updating the table.
+     * @param operations the operations to apply to the table.
+     * @param byColumns the columns to group by before applying.
+     * @return a table with the same index, with the specified operations applied to each group defined by the
+     *         {@code byColumns}
+     */
+    @ConcurrentMethod
+    Table updateBy(@NotNull final UpdateByControl control, @NotNull final Collection<UpdateByClause> operations,
```
devinrsmith, line 1136: I notice that TableWithDefaults has *more* methods for updateBy than Table does. I think we want the definitions to be on the Table interface.

engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -1719,6 +1720,96 @@ Table rollup(Collection<? extends Aggregation> aggregations, boolean includeCons
...
+     * provided
+     * </p>
+     *
+     * @param operations the operations to apply to the table.
+     * @param byColumns the columns to group by before applying.
+     * @return a table with the same index, with the specified operations applied to each group defined by the
+     *         {@code byColumns}
+     */
+    @ConcurrentMethod
+    Table updateBy(@NotNull final Collection<UpdateByClause> operations, final String... byColumns);
```
devinrsmith, line 1741: Collection of polymorphic types  in this context should ideally be `Collections<? extends Type>`, ie `Collection<? extends UpdateByClause>`.

engine/table/src/main/java/io/deephaven/engine/table/impl/UpdateByControl.java
```diff
@@ -0,0 +1,112 @@
+package io.deephaven.engine.table.impl;
```
devinrsmith, line 1: Is UpdateByControl something we want to be able to have clients be able to set via gRPC eventually? If so, this should move out of the engine impl and into the table-api (like other arguments to updateBy).

table-api/src/main/java/io/deephaven/api/updateby/UpdateByControl.java
```diff
@@ -107,6 +117,38 @@ default MathContext getDefaultMathContext() {
         return MathContext.DECIMAL64;
     }
 
+    @Check
+    default void checkChunkCapacity() {
```
devinrsmith, line 121: `@Check` should be `final`, not `default`.

devinrsmith (changes requested):
So close!

devinrsmith (dismissed):

devinrsmith (approved):

PR #2518, opened 2022-06-13
Add mapping Pandas's nullable numeric types

### LINKED ISSUE #2509: Support Python Pandas NA value
Add support for the experimental Python Pandas NA missing value.

https://pandas.pydata.org/docs/user_guide/missing_data.html

### LINKED ISSUE #2512: Mapping pandas wrapped array types (such as FloatArray) led to PyObject being found.
We should check and retrieve the real type (numpy.dtype) and use it to map to DType.

jmao-denver:
This change would recognize the Pandas ExtenstionDtype types and handle the conversion in to_table() according
Fixes #2512
Fixes #2509

### DIFF
--- a/py/server/deephaven/dtypes.py
+++ b/py/server/deephaven/dtypes.py
@@ -7,10 +7,11 @@ Each data type is represented by a DType class which supports creating arrays of
 """
 from __future__ import annotations
 
-from typing import Any, Sequence, Callable, Dict, Type
+from typing import Any, Sequence, Callable, Dict, Type, Union
 
 import jpy
 import numpy as np
+import pandas as pd
 
 from deephaven import DHError
 
@@ -180,14 +181,16 @@ def from_jtype(j_class: Any) -> DType:
         return dtype
 
 
-def from_np_dtype(np_dtype) -> DType:
-    """ Looks up a DType that matches the provided numpy dtype or Pandas's nullable equivalent, if not found,
+def from_np_dtype(np_dtype: Union[np.dtype, pd.api.extensions.ExtensionDtype]) -> DType:
+    """ Looks up a DType that matches the provided numpy dtype or Pandas's nullable equivalent; if not found,
     returns PyObject. """
 
-    if not isinstance(np_dtype, np.dtype):
-        # check if it is a Pandas nullable numeric types such as pd.core.arrays.floating.Float64Dtype/Int32Dtype
+    if isinstance(np_dtype, pd.api.extensions.ExtensionDtype):
+        # check if it is a Pandas nullable numeric types such as pd.Float64Dtype/Int32Dtype/BooleanDtype etc.
         if hasattr(np_dtype, "numpy_dtype"):
             np_dtype = np_dtype.numpy_dtype
+        elif isinstance(np_dtype, pd.StringDtype):
+            return string
         else:
             return PyObject
 
--- a/py/server/deephaven/dtypes.py
+++ b/py/server/deephaven/dtypes.py
@@ -168,7 +168,7 @@ def array(dtype: DType, seq: Sequence, remap: Callable[[Any], Any] = None) -> jp
 
 
 def from_jtype(j_class: Any) -> DType:
-    """ look up a DType that matches the java type, if not found, create a DType for it. """
+    """ looks up a DType that matches the java type, if not found, creates a DType for it. """
     if not j_class:
         return None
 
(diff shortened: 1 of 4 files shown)


chipkent (changes requested):

jjbrosnan (commented):
Second Chip's review

py/server/tests/test_pandas.py
```diff
@@ -165,6 +165,23 @@ def test_to_table_category(self):
...
+        df = pd.DataFrame({"NullableInt64": int_array, "NullableDouble": double_array})
+
+        with self.assertRaises(DHError) as cm:
+            table = to_table(df)
+        self.assertRegex(str(cm.exception), r"TypeError.* NAType")
+
+        int_array = int_array.fillna(NULL_LONG)
+        double_array = double_array.fillna(NULL_DOUBLE)
+        df = pd.DataFrame({"NullableInt64": int_array, "NullableDouble": double_array})
+        table = to_table(df)
```
chipkent, line 180: It seems like `to_table` should be transparently handling the NA conversions without the user having to mess with it.

chipkent (changes requested):

py/server/tests/test_pandas.py
jmao-denver: I don't oppose to the idea and we have a ticket #2509 for that. To solve would most likely require providing a way for the users to specify replacement values which will be quite a bit more work than what the PR intends to solve.

chipkent: I'm ok if it is another PR.

chipkent (dismissed):
Approved with the last work to be done in another PR.

py/server/tests/test_pandas.py
jmao-denver: After discussing it with @rbasralian I went ahead and implemented the transparent handling of pd.NA values. No interface change and it should not preclude us from adding user-supplied custom mapping of pd.NA in the future.

py/server/deephaven/constants.py
```diff
@@ -9,7 +9,7 @@
 _JQueryConstants = jpy.get_type("io.deephaven.util.QueryConstants")
 
 # Deephaven Special Null values for primitive types
-NULL_BOOLEAN = _JQueryConstants.NULL_BOOLEAN
+NULL_BOOLEAN = _JQueryConstants.NULL_BYTE
```
chipkent, line 12: Is this change intentional?  If so, it seems to deserve a comment.  This makes me nervous without justification.

py/server/deephaven/dtypes.py
```diff
@@ -180,8 +181,18 @@ def from_jtype(j_class: Any) -> DType:
         return dtype
 
 
-def from_np_dtype(np_dtype: np.dtype) -> DType:
-    """ Look up a DType that matches the numpy.dtype, if not found, return PyObject. """
+def from_np_dtype(np_dtype) -> DType:
```
chipkent, line 184: annotation?

py/server/deephaven/numpy.py
```diff
@@ -164,11 +164,12 @@ def to_table(np_array: np.ndarray, cols: List[str]) -> Table:
                             f"the number of column names {len(cols)}")
 
         input_cols = []
+        dtype = dtypes.from_np_dtype(np_array.dtype)
         if len(cols) == 1:
```
chipkent, line 168:
```suggestion
        dtype = dtypes.from_np_dtype(np_array.dtype)
        
        if len(cols) == 1:
```
Easier to read

py/server/deephaven/pandas.py
```diff
@@ -88,6 +89,33 @@ def to_pandas(table: Table, cols: List[str] = None) -> pandas.DataFrame:
         raise DHError(e, "failed to create a Pandas DataFrame from table.") from e
 
 
+EX_DTYPE_NULL_MAP = {
+    pd.BooleanDtype: NULL_BYTE,
```
chipkent, line 93: NULL_BOOLEAN?

py/server/deephaven/pandas.py
```diff
@@ -112,7 +140,10 @@ def to_table(df: pandas.DataFrame, cols: List[str] = None) -> Table:
 
         input_cols = []
         for col in cols:
-            input_cols.append(_make_input_column(col, df.get(col).values))
+            np_array = df.get(col).values
+            dtype = dtypes.from_np_dtype(np_array.dtype)
+            np_array = _map_na(df.get(col).values)
```
chipkent, line 145:
```suggestion
            np_array = _map_na(np_array)
```

py/server/tests/test_pandas.py
```diff
@@ -165,6 +165,41 @@ def test_to_table_category(self):
...
+        self.assertIs(table.columns[2].data_type, dtypes.int16)
+        self.assertIs(table.columns[3].data_type, dtypes.int32)
+        self.assertIs(table.columns[4].data_type, dtypes.int64)
+        self.assertIs(table.columns[5].data_type, dtypes.float32)
+        self.assertIs(table.columns[6].data_type, dtypes.double)
+        self.assertIs(table.columns[7].data_type, dtypes.string)
+        self.assertEqual(table.size, 3)
+        table_string = table.to_string()
+        self.assertEqual(8, table_string.count("null"))
+
```
chipkent, line 202: What happens if a user has a table like we saw, where a column is `object` and contains strings plus NaNs for missing values?  Are we going to force users to convert to `string` columns?  Yes is probably ok.  I'm just trying to understand the thinking.

chipkent (changes requested):

py/server/deephaven/constants.py
jmao-denver: yes, left a comment

py/server/tests/test_pandas.py
jmao-denver: Yes, Attempting to eagerly infer data type probably has more drawbacks than benefit. Also it is reasonable to ask the users to provide specific data types if they know what they are and don't want to see PyObject columns.

chipkent (changes requested):

chipkent (dismissed):

jjbrosnan (approved):

Issue #2515, opened 2022-06-10
Port Parquet PageCache creation deferrment


rcaudy:
Port https://gitlab.deephaven.io/illumon/iris/-/merge_requests/6597/diffs to lower memory usage when possible.

PR #2519, opened 2022-06-13
Defer PageCache creation, start at smaller initial size and grow intead

abaranec:


Util/src/main/java/io/deephaven/util/datastructures/intrusive/IntrusiveSoftLRU.java
```diff
@@ -84,18 +150,76 @@ public synchronized void touch(@NotNull T t) {
...
+            this.size = 0;
+            for (final SoftReference<T> reference : oldRefs) {
+                final T maybeItem = reference.get();
+                if (maybeItem != null) {
+                    final int slotForItem = size++;
+                    adapter.setSlot(maybeItem, slotForItem);
+                    softReferences[slotForItem] = reference;
+                }
+            }
+        }
```
rcaudy, line 209: This entire block could be safely reduced to an array copy and a conditional initializeLinks. *Maybe* with an additional step to replace cleared references that aren't the null reference with the null reference. That would more accurately reflect the assumption that nothing in the `softReferences` array is ever `null`.
```suggestion
        final int oldCapacity = softReferences.length;
        final int newCapacity = Math.min(oldCapacity << 1, maxCapacity);
        softReferences = Arrays.copyOf(softReferences, newCapacity);
        Arrays.fill(softReferences, oldCapacity, newCapacity, getNull());
        if (newCapacity == maxCapacity) {
            initializeLinks();
        }
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/location/ParquetColumnLocation.java
```diff
@@ -68,9 +68,10 @@
 
     private static final int CHUNK_SIZE = Configuration.getInstance()
             .getIntegerForClassWithDefault(ParquetColumnLocation.class, "chunkSize", 4096);
-    private static final int PAGE_CACHE_SIZE =
-            Configuration.getInstance().getIntegerForClassWithDefault(ParquetColumnLocation.class, "pageCacheSize",
-                    1 << 13);
+    private static final int INITIAL_PAGE_CACHE_SIZE = Configuration.getInstance()
+            .getIntegerForClassWithDefault(ParquetColumnLocation.class, "initialPageCacheSize", 128);
+    private static final int MAX_PAGE_CACHE_SIZE = Configuration.getInstance()
+            .getIntegerForClassWithDefault(ParquetColumnLocation.class, "maxPageCacheSize", 1 << 13);
```
rcaudy, line 74:
```suggestion
            .getIntegerForClassWithDefault(ParquetColumnLocation.class, "maxPageCacheSize", 8192);
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/location/ParquetColumnLocation.java
```diff
@@ -83,7 +84,7 @@
     private final boolean hasGroupingTable;
 
     // We should consider moving this to column level if needed. Column-location level likely allows more parallelism.
-    private final PageCache<ATTR> pageCache = new PageCache<>(PAGE_CACHE_SIZE);
+    private PageCache<ATTR> pageCache;
```
rcaudy, line 87:
```suggestion
    private volatile PageCache<ATTR> pageCache;
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/location/ParquetColumnLocation.java
```diff
@@ -108,6 +109,21 @@
...
+        }
+
+        synchronized (this) {
+            if (pageCache != null) {
+                return pageCache;
+            }
+
+            pageCache = new PageCache<>(INITIAL_PAGE_CACHE_SIZE, MAX_PAGE_CACHE_SIZE);
+        }
+        return pageCache;
```
rcaudy, line 124:
```suggestion
        PageCache<ATTR> localPageCache;
        if ((localPageCache = pageCache) != null) {
            return localPageCache;
        }
        synchronized (this) {
            if ((localPageCache = pageCache) != null) {
                return localPageCache;
            }
            return pageCache = new PageCache<>(INITIAL_PAGE_CACHE_SIZE, MAX_PAGE_CACHE_SIZE);
        }
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/PageCache.java
```diff
@@ -7,7 +7,9 @@
 import java.lang.ref.WeakReference;
 
 /**
- * Page cache data structure.
+ * A Cache for {@link IntrusivePage IntrusivePages}. This datastructure stores pages as soft references and maintains
+ * them as an LRU cache. Externally references can be held as WeakReferences so that as memory pressure builds they can
+ * be evicted from the cache.
```
rcaudy, line 12:
```suggestion
 * A cache for {@link IntrusivePage IntrusivePages}. This data structure stores pages as {@link SoftReference soft references} and maintains
 * them as an LRU cache. External references to cached pages should be held via {@link WeakReference weak references} so that as memory pressure builds the pages can
 * be evicted from the cache.
```

Util/src/main/java/io/deephaven/util/datastructures/intrusive/IntrusiveSoftLRU.java
```diff
@@ -119,15 +243,44 @@ public synchronized boolean verify(int numStored) {
...
-        SoftReference<T> getOwner(T t);
-
-        int getSlot(T t);
-
-        void setSlot(T t, int slot);
+        /**
+         * Get the {@link SoftReference} object which owns the object being cached.
+         *
+         * @param cachedObject the object being cached.
+         * @return the {@link SoftReference} that owns the item.
```
rcaudy, line 258:
```suggestion
         * Get a {@link SoftReference} object which refers to the object being cached and will act as its "owner" in the cache.
         *
         * @param cachedObject the object being cached.
         * @return a {@link SoftReference} that refers to the item and will act as its "owner" in the cache.
```

Util/src/main/java/io/deephaven/util/datastructures/intrusive/IntrusiveSoftLRU.java
abaranec: Accepted -- I added a step before so that we can still do compaction early and maybe get lucky and not expand

Util/src/main/java/io/deephaven/util/datastructures/intrusive/IntrusiveSoftLRU.java
```diff
@@ -176,36 +179,34 @@ private void touchSimple(@NotNull final T itemToTouch) {
...
+                softReferences[refIdx - nullsFound] = softReferences[refIdx];
+                softReferences[refIdx] = getNull();
+                adapter.setSlot(object, refIdx - nullsFound);
+            }
+        }
 
-        // We have grown too large, switch to LRU mode.
+        // We free'd up space! Cool!
+        size -= nullsFound;
+        if (size < softReferences.length - 1) {
```
rcaudy, line 199: I think we want `(size < softReferences.length)`

Util/src/main/java/io/deephaven/util/datastructures/intrusive/IntrusiveSoftLRU.java
```diff
@@ -176,36 +179,34 @@ private void touchSimple(@NotNull final T itemToTouch) {
...
+            return;
+        }
```
rcaudy, line 201: If you agree, delete line 192 and accept this suggestion, too.
```suggestion
        if (nullsFound > 0) {
            // We free'd up space! Cool!
            size -= nullsFound;
            Assert.lt(size, "size", softReferences.length, "softReferences.length");
            Arrays.fill(softReferences, size, softReferences.length, getNull());
            return;
        }
```

rcaudy:
Closes #2515

PR #2521, opened 2022-06-13
jsapi: fix null check for optional color param

nbauernfeind:
This just fixes some whining when an optional parameter is not specified.


mattrunyon:
I get the same error on the docs example with this fix

I think the issue Colin and I found originally was the tables get bound in a different order or 1 table gets bound as both.

We set a breakpoint and inspected and saw that table 0 was table 1 and table 1 was table 0, or table 0 was both table 0 and table 1 (I don't remember which, but I think it was swapped). This caused the issue with `getColumn` because the column name is from table 1, but looking up on table 0.

mattrunyon (changes requested):
Still getting the error. I think it is more in depth and has to do with using a HashSet or HashMap and losing order somewhere

mattrunyon (approved):

mattrunyon (approved):

Issue #734, opened 2021-06-09
Fix DB-test python tests


niloc132:
Follow-up to #430

In enterprise as well as in this repo, all junit tests that use python in DB-test are marked `@Ignore`, so they are not run - presumably something is wrong with the test or with the code under test, and should be investigated.

pete-petey:
Back to @niloc132  per 6/16 update meeting.

niloc132:
Back to @jmao-denver https://github.com/deephaven/deephaven-core/pull/881

niloc132:
DB-test is renamed to python-engine-test as of 5c1b52c85ab66a3fb77a0127095453276d024fe2

PR #2538, opened 2022-06-14
Cleanup python deephaven package dependencies

devinrsmith:
I don't think we have a real dependency on dill nor wrapt. Since we only support >= 3.7, numba is always required. And enum34 is never required. The Deprecated classes should eventually be removed, nothing uses them except tests.


jmao-denver (dismissed):
LGTM

devinrsmith:
I've gone ahead with the further cleanup here.

python-engine-test/src/test/java/io/deephaven/engine/table/impl/select/PythonMatchFilterTest.java
```diff
@@ -1,73 +0,0 @@
...
-import java.util.List;
-
-import org.junit.Before;
-import org.junit.Ignore;
-import org.junit.Test;
-
-/**
- * Test MatchFilters that reference Python lists.
- */
-@Ignore // TODO (deephaven-core#734)
```
chipkent, line 27: https://github.com/deephaven/deephaven-core/issues/734 Says that someone needs to look into why these are failing before deciding to delete.  Was this analysis done?

python-engine-test/src/test/java/io/deephaven/engine/table/impl/select/TestConditionFilter.java
```diff
@@ -1,440 +0,0 @@
...
-import java.util.function.Predicate;
-import org.junit.Before;
-import org.junit.Ignore;
-import org.junit.Test;
-
-import static io.deephaven.engine.table.impl.select.FormulaTestUtil.*;
-import static org.jpy.PyLib.getMainGlobals;
-import static org.junit.Assert.fail;
-
-@Ignore // TODO (deephaven-core#734)
```
chipkent, line 41: https://github.com/deephaven/deephaven-core/issues/734 Says that someone needs to look into why these are failing before deciding to delete.  Was this analysis done?

python-engine-test/src/test/java/io/deephaven/engine/util/TestWorkerPythonEnvironment.java
```diff
@@ -1,108 +0,0 @@
...
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.util.concurrent.TimeoutException;
-
-import org.junit.Ignore;
-
-/**
- * Test various Jpy related overloading methods.
- */
-@Ignore // TODO (deephaven-core#734)
```
chipkent, line 25: https://github.com/deephaven/deephaven-core/issues/734 Says that someone needs to look into why these are failing before deciding to delete.  Was this analysis done?

engine/table/src/main/java/io/deephaven/engine/util/PickledResult.java
```diff
@@ -1,27 +0,0 @@
-/**
```
chipkent, line 1: There is a feature request to make python object columns persistable.  I can't find the ticket in DHC, so it must be in DHE.  Is this functionality we need for that feature, or would that feature be a complete redo on pickled serialization?

python-engine-test/src/test/java/io/deephaven/engine/table/impl/select/PythonMatchFilterTest.java
chipkent: If this resolves all of #734, we should close that issue with this PR.

chipkent (changes requested):

devinrsmith:
I've removed the file deletions, per @niloc132:

>moving it to the python-engine-test stuff seems cheaper long term, since at least we want to get something like those tests running at some point, rather than deleting and restoring imho

engine/table/src/main/java/io/deephaven/engine/util/PickledResult.java
```diff
@@ -8,6 +8,7 @@
 /**
  * Pickled result for a Python fetch.
  */
+@Deprecated
```
jmao-denver, line 11: Could you leave the change out? I moved it in #2582?

engine/table/src/test/java/io/deephaven/engine/util/WorkerPythonEnvironment.java
```diff
@@ -24,6 +24,7 @@
  * remote Python session should execute queries which interact wtih this class. The script sessions that run for
  * PersistentQueries or consoles are handled separately by the {@link PythonDeephavenSession}.
  */
+@Deprecated
```
jmao-denver, line 27: same as above

engine/table/src/main/java/io/deephaven/engine/util/PickledResult.java
devinrsmith: Removed - that said, it would have resolved cleanly with you move.

jmao-denver (approved):
LGTM

engine/table/src/test/java/io/deephaven/engine/util/WorkerPythonEnvironment.java
devinrsmith: Removed - that said, it would have resolved cleanly with you move.

PR #2539, opened 2022-06-15
ExecutionContext: record state of globals before deferring work

### LINKED ISSUE #2496: Instantiation Context Recording for Deferred Operations
As a user, when I invoke an operation that may be deferred, I want the `QueryScope`, `QueryLibrary`, and `CompilerTools.Context` that were active at invocation time to apply when the operation is ultimately applied. Otherwise, it can be difficult to ensure correct operation if this context is mutated between my invocation and eventual instantiation.

This has a few aspects:
1. Deferred views and filters applied to `UncoalescedTable`
2. Update propagation for `Table.update()` and `Table.select()` with a `Table` as input
3. Update propagation for `PartitionedTable.transform()` and `PartitionedTable.partitionedTransform()`, as well as operations invoked via `PartitionedTable.Proxy` _(These are refinements of case 2, but we may be able to automatically narrow the scope of the context we record to avoid unnecessarily preserving `Object`s past their intended lifetime.)_
4. User listeners that may apply operations to `Table`s on notification

This is important to deliver for multiple reasons:
1. Users may choose to mutate the `QueryScope` or other global context after invoking deferred operations, incorrectly assuming that this is already guaranteed to be safe
2. Users may choose to install a different `QueryScope`, `QueryLibrary`, or `CompilerTools.Context` with the intention of isolating invocations from the global context, and we should fully honor this intention
3. Python modules should be able to correctly expose their own globals to the query engine without polluting the enclosing modules globals
4. Isolation is required if we ever want to support multiple script sessions within the same worker

nbauernfeind:
Fixes #2496.

The QueryScope, QueryLibrary, CompilerTools were merged into a new package `io.deephaven.engine.context` and are bundled as a distinct artifact `engine-context`.

Tools that perform deferred work may record the execution context, if explicitly set by the user, to reinstall in deferred contexts. WhereFilters and SelectColumns will be initialized immediately and copies will retain details instead of needing to reevaluate in a deferred context.

This commit introduces `TableOperationsDefaults` which lifts many default `TableOperations` implementations from `TableDefaults`. This helps reinforce consistency in other implementations of TableOperations.

The QueryLanguageParser can now deduce the return type of a generic method if it is fixed due to inheritance rules (both via interfaces and parent classes).


Util/src/main/java/io/deephaven/util/ExecutionContext.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+ */
+package io.deephaven.util;
```
rcaudy, line 4: Maybe this should be somewhere higher up? Maybe introduce a new module `engine-context` lower than `engine-updategraph`, in a package called `io.deephaven.engine.context`? Maybe move compiler tools up to that package, too?

CompilerTools/src/main/java/io/deephaven/compilertools/CompilerTools.java
```diff
@@ -146,12 +147,32 @@ public static void writeClass(final File destinationDirectory, final String clas
     public static final String FORMULA_PREFIX = "io.deephaven.temp";
     public static final String DYNAMIC_GROOVY_CLASS_PREFIX = "io.deephaven.dynamic";
 
-    public static class Context {
+    public interface Context {
```
rcaudy, line 150: Maybe `CompilerTools` need not be all the way down in its own module.

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -147,30 +137,16 @@ protected Changes applyDiff(S from, S to, RuntimeException e) {
...
-                QueryLibrary.setLibrary(queryLibrary);
+        // retain any objects which are created in the executed code, we'll release them when the script session
+        // closes
+        try (S fromSnapshot = takeSnapshot();
+                final SafeCloseable ignored = LivenessScopeStack.open(this, false)) {
 
+            try {
                 // actually evaluate the script
-                evaluate(script, scriptName);
+                executionContext.apply(() -> evaluate(script, scriptName));
```
rcaudy, line 147: I think this needs to be applied for the snapshots, since they are capturing QueryScope.

rcaudy (commented):
First review pass on the latest version. Have you run the nightlies?

py/server/test_helper/__init__.py
```diff
@@ -94,7 +94,7 @@ def start_jvm(jvm_props: Dict[str, str] = None):
         # Set up a Deephaven Python session
         py_scope_jpy = jpy.get_type("io.deephaven.engine.util.PythonScopeJpyImpl").ofMainGlobals()
         py_dh_session = jpy.get_type("io.deephaven.integrations.python.PythonDeephavenSession")(py_scope_jpy)
-        jpy.get_type("io.deephaven.engine.table.lang.QueryScope").setScope(py_dh_session.newQueryScope())
+        py_dh_session.getExecutionContext().open()
```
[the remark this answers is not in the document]
nbauernfeind: It looks like the test framework creates the jvm, instantiates a python session, and then wants to specify which query scope to work against.

The original code was never resetting the scope. I was aiming for the same intent but using the execution context.

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
nbauernfeind: I think Snapshots are already OK. They are capturing via the `GroovyShell` and `PythonScope` directly without asking the `QueryScope` thread local value.

engine/table/src/test/java/io/deephaven/engine/table/impl/lang/TestQueryLanguageParser.java
```diff
@@ -2187,6 +2189,51 @@ public void testInvalidExpr() throws Exception {
...
+        default T getFromInterfaceSub() {
+            return null;
+        }
+    }
+    public static abstract class TestGenericSuper<T> implements TestGenericInterfaceForSuper<TestGenericResult2> {
+        public T getFromSuper() {
+            return null;
+        }
+    }
+    public static class TestGenericSub extends TestGenericSuper<TestGenericResult3>
```
rcaudy, line 2220: Should we test with multiple declared interfaces on the same level?

engine/table/src/main/java/io/deephaven/engine/table/impl/select/TimeSeriesFilter.java
```diff
@@ -107,7 +107,12 @@ public void setRecomputeListener(RecomputeListener listener) {
 
     @Override
     public TimeSeriesFilter copy() {
-        return new TimeSeriesFilter(columnName, nanos);
+        final TimeSeriesFilter copy = new TimeSeriesFilter(columnName, nanos);
+        if (initialized) {
+            copy.initialized = true;
+            UpdateGraphProcessor.DEFAULT.addSource(copy);
```
rcaudy, line 113: This code as it currently exists is pretty buggy. We should not be in the UGP unless we have `setRecomputeListener` has been called. I think you should eliminate the `initialized` boolean, make `setRecomputeListener` check that the `listener` has not already been set, and do the UGP add there.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/BaseIncrementalReleaseFilter.java
```diff
@@ -212,6 +212,14 @@ public void setRecomputeListener(RecomputeListener listener) {
     @Override
     abstract public BaseIncrementalReleaseFilter copy();
 
+    protected void onCopy(BaseIncrementalReleaseFilter copy) {
+        copy.started = started;
+        copy.initialized = initialized;
+        if (started && initialized) {
+            copy.addToUpdateGraphProcessor();
+        }
```
rcaudy, line 220: I don't think we should copy this state. "started" seems instance-specific.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/RollingReleaseFilter.java
```diff
@@ -88,7 +92,9 @@ public void setRecomputeListener(RecomputeListener listener) {
 
     @Override
     public RollingReleaseFilter copy() {
-        return new RollingReleaseFilter(workingSize, rollingSize);
+        final RollingReleaseFilter copy = new RollingReleaseFilter(workingSize, rollingSize);
+        copy.addToUpdateGraphProcessor();
```
rcaudy, line 96: Same comment as on some other code. You can't add to the UGP without a set listener.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/BaseIncrementalReleaseFilter.java
nbauernfeind: As discussed offline I changed this to throw an UnsupportedOperation.

rcaudy (approved):

rcaudy (approved):

PR #2532, opened 2022-06-14
Tensorboard support

anastasia-si:
added tensorboard to the TensorFlow/PyTorch images

PR #2545, opened 2022-06-15
Add myself as codeowner for docker directory

devinrsmith:
I should have reviewer status for docker/ changes. https://github.com/deephaven/deephaven-core/pull/2532 unintentionally broke part of the build (that isn't tested via CI).

Changes to the requirements files is managed via https://github.com/deephaven/deephaven-base-images, will follow up with a fix there and here.

```shell
./gradlew docker-server:buildDocker-server-pytorch
```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -18,4 +18,4 @@
 /README.md @chipkent @rcaudy @margaretkennedy
 /TRIAGE.md @chipkent
 /licenses @chipkent
-/docker @devinrsmith
+/docker @devinrsmith @jcferretti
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -18,3 +18,4 @@
 /README.md @chipkent @rcaudy @margaretkennedy
 /TRIAGE.md @chipkent
 /licenses @chipkent
+/docker @devinrsmith
```
chipkent, line 21: Since there is a non-zero chance that Docker creates a time-critical problem and a non-zero chance you will go on vacation, can we add a backup?  e.g. me, Colin, etc.

chipkent (changes requested):

chipkent (approved):

PR #2548, opened 2022-06-15
Add server configuration for mTLS authentication

devinrsmith:
Fixes https://github.com/deephaven/deephaven-core/issues/2551


server/src/main/java/io/deephaven/server/runner/Main.java
```diff
@@ -25,7 +26,11 @@ public class Main {
     public static final String SSL_IDENTITY_TYPE = "ssl.identity.type";
     public static final String SSL_IDENTITY_CERT_CHAIN_PATH = "ssl.identity.certChainPath";
     public static final String SSL_IDENTITY_PRIVATE_KEY_PATH = "ssl.identity.privateKeyPath";
+    public static final String SSL_TRUST_TYPE = "ssl.trust.type";
```
niloc132, line 29: consider making these keys much more specific, or is it intended that these will be usable for all cases of ssl trust - for example, the starting process connecting as a ssl client to another worker, and needing a trust store/etc for that tls (whether or not mtls is used).

as names go, this wouldnt be terrible for an eventual fallback, but i would be concerned that surprises could emerge - such as telling a given worker "this is the path to the trust store to use for ssl" as a general statement (i.e. if the cert on the server you connect to doesn't match this, fail it), and finding that your ssl clients using $XYZ authn can no longer connect because that forced mtls to be enabled, etc.

devinrsmith: This is meant to be more general, and can be useful w/o mTLS. That said, it's also necessary for mTLS.

Yeah, I think there's a lot of potential configuration patterns; and unfortunately, I think using property based configuration is just not the best way to do it. (Note: I've seen how Java Kafka is configured in these regards using properties, and it's not pretty.) The big thing is the lack of distinction between "who do I trust as a client" vs "who do I trust as a server"; but that can be extended larger as soon as you want to start distinguishing based on hostname or another criteria.

As far as "who do I trust as a client", I've been opinionated with the configuration with:
> If trust material is present, {@link TrustJdk} will also be added
So, if you connect out to a server you'll trust public CAs (or, however your JDK is configured), as well as the trust certificates specified.

But this does let us down on the flip side of "who do I trust as a server". Anybody could new connect to this server with a publicly signed identity. Which I guess defeats one of the major goals of this PR - allow a locked-down mTLS intranet-of-sorts that wouldn't need application level authorization.

My preference here is to solve this by being more opinionated as opposed to expanding how nuanced we can make the property based configuration.

I would propose creating two ssl configuration based on the properties already defined (when ssl.trust.type is defined):

client trust = trusts ssl.trust.path + TrustJDK

server trust = trusts ssl.trust.path

Otherwise (ssl.trust.type not defined):

client trust = server trust = TrustJDK

devinrsmith: And make ssl.clientAuthentication mandatory for mTLS.

niloc132: I still don't understand how this is sufficient, but it also isn't clear to me how the latest changes have implemented the proposed tweaks above. 

>  Which I guess defeats one of the major goals of this PR - allow a locked-down mTLS intranet-of-sorts that wouldn't need application level authorization.
This is exactly my concern. With this flaw in place, the patch serves no purpose, since TLS already works - the `m` has no value if any random off the internet can use it, and reduced value if any server a DH instance can authenticate to is also allowed to authenticate back to it.

Let me outline the clear problem that is keeping me from being willing to approve this.

Givens:
* The purpose of mTLS and this patch is to go above transport layer security and enable the server to be configured to trust that a connecting user has a recognized identity for a given dh core instance. Otherwise its just "TLS, but the client has to have a cert"
* Within a single instance which will connect both to upstream servers (other mTLS-guarded, or just plain TLS and assume other authn) and allow downstream clients to connect to it (only considering mTLS-authenticated clients in this case)
* That instance has only one user-specified trust store, and optionally (according to your last comment) the client mTLS connects may use the JDK trust store (note that I might be flipping the meaning here of "connect as a client" for "permit a client to connect". That same trust store will be used for connecting to other servers (or, if I flipped the above, being a server that others can connect _to_).

Flaws:
 * If the "clients who can connect to me" trust store includes TrustJDK, then any random off the internet can get an ACME client cert and "prove their identity" to the server.
 * If the "I can connect as a client" trust store is a non-standard CA (for example, run by another external company, so that they tightly control client/server certs) but TrustJDK could be part of that set, then that same external company, desired or not, could have signed any number of other client certs which will be trusted by the current instance, with no way to exclude them without outright disabling mTLS.

In short, I don't understand the purpose of the ticket outside a very narrow case where a dh core instance is only a server, only a client, or only exists in a 100% homogeneous trust environment. Accepting a client cert _must_ be able to come from a _distinct_ truststore (the CA for "users that are permitted to connect"), to allow the flexibility of connecting to a servers that use their own private CA.

devinrsmith: > no value if any random off the internet can use it

That's not what's happening here. It *can* be configured that way, but it's not that way be default. Here an example from the gradle `-PdevMTLS`:

```properties
ssl.identity.type=privatekey
ssl.identity.certChainPath=server.chain.crt
ssl.identity.privateKeyPath=server.key
ssl.trust.type=certs
ssl.trust.path=ca.crt
ssl.clientAuthentication=NEEDED
```

The configuration manifests as:

1. We present ourselves server.chain.crt
1. Only clients that are trusted by ca.crt can connect to us
1. For outbound connections, we trust ca.crt + JDK

If we'd prefer to be more explicit w/ our outbound configuration, we can use the `outbound.` prefix instead:

```
ssl.identity.type=privatekey
ssl.identity.certChainPath=server.chain.crt
ssl.identity.privateKeyPath=server.key
ssl.trust.type=certs
ssl.trust.path=ca.crt
ssl.clientAuthentication=NEEDED

outbound.ssl.identity.type=privatekey
outbound.ssl.identity.certChainPath=outbound-identity.chain.crt
outbound.ssl.identity.privateKeyPath=outbound-identity.key
outbound.ssl.trust.type=certs
outbound.ssl.trust.path=ca-or-some-other-outbound.crt
```

niloc132: If `outbound.ssl...` is "i'm behaving as a client, connecting to an external server" and unprefixed `ssl...` is "i'm a server, checking an incoming client", then yes, that's what i was originally asking for, ensuring that more than disjoint CAs can be used in each direction. _Very likely_ the default will need to either be "outbound uses JDK Trust store, unless otherwise specified" and "inbound uses a specific internal CA" or they both share the same CA, but I believe that the option (and examples and docs!) must exist to configure inbound and outbound separately to make the feature worth more than vanilla TLS.

devinrsmith: Yep - that's the distinction between `ssl.` and `outbound.ssl.` If we prefer to be more explicit on both sides, we can `server.ssl.`, `client.ssl.`; but I think the way it is now makes the most sense to convey the notion that `outbound.ssl` is more specific than `ssl` (and inherits the setup when `outbound.ssl` is not present). I think the defaults I've shown w/ `devMTLS` is a very reasonable and common setup where both inbound and outbound will share common CA, and the outbound will additionally trust the JDK.

devinrsmith: I'll add some documentation w/ these points.

server/src/main/java/io/deephaven/server/runner/DeephavenApiConfigModule.java
```diff
@@ -6,8 +6,12 @@
 import dagger.Module;
 import dagger.Provides;
 import io.deephaven.server.config.ServerConfig;
+import io.deephaven.ssl.config.SSLConfig;
+import io.deephaven.ssl.config.Trust;
+import io.deephaven.ssl.config.TrustJdk;
 
 import javax.inject.Named;
+import java.util.Optional;
```
niloc132, line 14: unused

server/src/main/java/io/deephaven/server/auth/TrivialAuthContextProvider.java
```diff
@@ -19,6 +27,11 @@ public boolean supportsProtocol(final long authProtocol) {
 
     @Override
     public AuthContext authenticate(final long protocolVersion, final ByteString payload) {
+        List<X509Certificate> x509Certificates = MTlsCertificate.CLIENT_CERTIFICATES.get();
+        if (x509Certificates != null && !x509Certificates.isEmpty()) {
+            log.info().append(x509Certificates.get(0).getSubjectDN().toString()).endl();
```
niloc132, line 32: i put this in for debugging, do we want to land it as-is?

server/src/main/java/io/deephaven/server/uri/BarrageTableResolver.java
```diff
@@ -24,7 +26,9 @@
 import io.grpc.ManagedChannel;
 import org.apache.arrow.memory.BufferAllocator;
 
+import javax.annotation.Nullable;
```
niloc132, line 29: unused i think?

server/src/main/java/io/deephaven/server/runner/Main.java
```diff
@@ -88,29 +93,84 @@ public static Configuration init(String[] args, Class<?> mainClass) throws IOExc
...
+        final String trustPath = config.getStringWithDefault(prefix(prefix, SSL_TRUST_PATH), null);
+        if (trustPath == null) {
+            throw new IllegalArgumentException(String.format("Must specify `%s`", prefix(prefix, SSL_TRUST_PATH)));
+        }
+        return Optional.of(TrustCertificates.of(trustPath));
+    }
+
+    private static Optional<ClientAuth> parseClientAuth(String prefix, Configuration config) {
+        final String clientAuth = config.getStringWithDefault(prefix(prefix, SSL_CLIENT_AUTH), null);
+        if (clientAuth == null) {
```
niloc132, line 167: opinionated, but why not `Optional.ofNullable(clientAuth).map(ClientAuth::valueOf)`.

server/src/main/java/io/deephaven/server/runner/Main.java
```diff
@@ -88,29 +93,84 @@ public static Configuration init(String[] args, Class<?> mainClass) throws IOExc
...
+        final String trustType = config.getStringWithDefault(prefix(prefix, SSL_TRUST_TYPE), null);
+        if (trustType == null) {
+            return Optional.empty();
+        }
+        if (!CERTS.equals(trustType)) {
+            throw new IllegalArgumentException(
+                    String.format("Only support `%s` trust type through Configuration", CERTS));
+        }
+        final String trustPath = config.getStringWithDefault(prefix(prefix, SSL_TRUST_PATH), null);
+        if (trustPath == null) {
```
niloc132, line 159: opinionated, but why not `Optional.ofNullable(trustPath).map(TrustCertificates::of).orElseThrow(...)`? It seems silly to use optional and then do manual null checks?

server/src/main/java/io/deephaven/server/auth/TrivialAuthContextProvider.java
devinrsmith: I imagine we can take it out - we can create a more appropriate `AuthContext` for mTLS as the authorization gets fleshed out.

devinrsmith: Or maybe, keep a specific AuthContextProvider w/ mTLS example - that may be better?...

server/src/main/java/io/deephaven/server/runner/Main.java
devinrsmith: `orElseThrow` returns `T` - I *could* do:

`Optional.of(Optional.ofNullable(trustPath).map(TrustCertificates::of).orElseThrow(...))`, but that seems worse?

niloc132: or just or(() -> throw...)?

server/src/main/java/io/deephaven/server/auth/TrivialAuthContextProvider.java
niloc132: Yep, seems reasonable either way - maybe kick the log level down to debug/trace in the meantime if you leave it in?

server/src/main/java/io/deephaven/server/runner/Main.java
devinrsmith: Done.

server/src/main/java/io/deephaven/server/auth/TrivialAuthContextProvider.java
devinrsmith: I've removed it for now.

niloc132 (approved):

PR #2555, opened 2022-06-17
Write Encoding stats with data.

abaranec:


extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -411,147 +539,199 @@ public Integer get() {
...
+            RowSet rowSet,
+            ColumnSource<DATA_TYPE> columnSource,
+            ColumnSource<?> lengthSource,
+            RowSet lengthRowSet,
+            LongSupplier rowStepGetter,
+            LongSupplier valuesStepGetter,
+            int targetRowsPerPage,
+            int pageCount,
+            Class<DATA_TYPE> columnType,
+            ColumnWriter columnWriter) throws IOException {
```
rcaudy, line 591: Since we're adding `final`s and `@NotNull`s, maybe the two new methods should get the same treatment?

engine/table/src/test/java/io/deephaven/engine/table/impl/sources/regioned/TestSymbolTableSource.java
```diff
@@ -0,0 +1,58 @@
...
+    public void testWriteAndReadSymbols() {
+        final Table t = TableTools.emptyTable(10).update("TheBestColumn=`S`+k");
+        final File toWrite = new File(dataDirectory, "table.parquet");
+        ParquetTools.writeTable(t, toWrite, t.getDefinition());
+
+        // Make sure we have the expected symbol table (or not)
+        final Table readBack = ParquetTools.readTable(toWrite);
+        // noinspection unchecked
+        SymbolTableSource<String> source = (SymbolTableSource) readBack.getColumnSource("TheBestColumn");
+        assertTrue(source.hasSymbolTable(readBack.getRowSet()));
```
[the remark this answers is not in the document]
abaranec: In DHE there are tests elsewhere for that -- I assumed you still had those

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -411,147 +539,199 @@ public Integer get() {
...
+            final RowSet lengthRowSet,
+            final int targetRowsPerPage,
+            final LongSupplier rowStepGetter,
+            final LongSupplier valuesStepGetter,
+            final int pageCount,
+            final ColumnWriter columnWriter) throws IOException {
+        // Note: We only support strings as dictionary pages. Knowing that, we can make some assumptions about chunk
+        // types and avoid a bunch of lambda and virtual method invocations. If we decide to support more, than
+        // these assumptions will need to be revisited.
+        Assert.eqTrue(supportsDictionary(columnSource.getType()), "ColumnSource supports dictionary");
```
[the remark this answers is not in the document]
abaranec: This is what it did before.  If it offends you,  I can just remove all that crap.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -411,147 +539,199 @@ public Integer get() {
...
+                    final ObjectChunk<String, Values> chunk =
+                            (ObjectChunk<String, Values>) columnSource.getChunk(context, rs);
+                    final IntBuffer posInDictionary = IntBuffer.allocate((int) rs.size());
+                    for (int vi = 0; vi < chunk.size(); ++vi) {
+                        final String key = chunk.get(vi);
+                        int dictionaryPos = keyToPos.get(key);
+                        if (dictionaryPos == keyToPos.getNoEntryValue()) {
+                            if (key == null) {
+                                hasNulls = true;
+                                dictionaryPos = QueryConstants.NULL_INT;
```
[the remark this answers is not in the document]
abaranec: I thought it was more 'correct' to use the thing from the class.  Not that it makes any difference.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -307,100 +401,134 @@ private static ParquetFileWriter getParquetFileWriter(
...
+
+                                // Reset the counts to compute these values for the next page.
+                                originalRowsInPage = 0;
+                                totalItemsInPage = 0;
                             }
-                            runningSize += chunk.get(i);
+                            totalItemsInPage += curLength;
                         }
-                        originalRowsCount++;
+                        originalRowsInPage++;
```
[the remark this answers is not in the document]
abaranec: Null stuff is done in the maybeEncodeDictionary and encodePlain methods.  There is further null stuff buried deeply in the code that actually stuffs the data into the parquet writers

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -307,100 +401,134 @@ private static ParquetFileWriter getParquetFileWriter(
...
+            valuesStepGetter = new LongSupplier() {
                 int step;
 
                 @Override
-                public Integer get() {
-                    return valueChunkSize.get(step++);
+                public long getAsLong() {
+                    return rawItemCountPerPage.get(step++);
                 }
             };
```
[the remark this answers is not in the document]
abaranec: I decided to keep the LongSupplier, because switching to TIntIterator requires adding another silly inner class for the non-array type.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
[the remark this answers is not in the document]
rcaudy: Before, it hid that using the two methods you removed:
```
    private static Function<Object, Object> getToParquetConversion(Class<?> type) {
        if (type == String.class) {
            // noinspection unchecked,rawtypes
            return (Function) (Function<String, Binary>) Binary::fromString;
        }
        throw new UnsupportedOperationException("Dictionary storage not supported for " + type);
    }

    private static Function<Integer, Object[]> getKeyArrayBuilder(Class<?> type) {
        if (type == String.class) {
            return Binary[]::new;
        }
        throw new UnsupportedOperationException("Dictionary storage not supported for " + type);
    }
```

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/TestSymbolTableSource.java
```diff
@@ -40,19 +41,24 @@ public final void tearDown() {
...
     public void testWriteAndReadSymbols() {
-        final Table t = TableTools.emptyTable(10).update("TheBestColumn=`S`+k");
+        final Table t = TableTools.emptyTable(100).update("TheBestColumn=`S`+ (k % 10)", "Sentinel=k");
         final File toWrite = new File(dataDirectory, "table.parquet");
         ParquetTools.writeTable(t, toWrite, t.getDefinition());
 
         // Make sure we have the expected symbol table (or not)
         final Table readBack = ParquetTools.readTable(toWrite);
         // noinspection unchecked
         SymbolTableSource<String> source = (SymbolTableSource) readBack.getColumnSource("TheBestColumn");
```
rcaudy, line 56:
```suggestion
        final SymbolTableSource<String> source =
                (SymbolTableSource<String>) readBack.getColumnSource("TheBestColumn", String.class);
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -543,55 +517,53 @@ public long getAsLong() {
...
+                        columnWriter,
                         columnSource,
                         lengthSource,
-                        lengthRowSet,
-                        targetRowsPerPage,
                         rowStepGetter,
                         valuesStepGetter,
-                        pageCount,
-                        columnWriter);
+                        targetRowsPerPage,
```
rcaudy, line 529: I'm decently sure you need to compute the max rows per page, and that is what you should be using to size the data `GetContext` in the `encodeX` methods. I thought that was the intent behind your "escape hatch", which you removed entirely instead of just using it as a place to record max data page size.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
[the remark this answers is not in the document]
rcaudy: That's totally correct. I'm just saying you can skip the `dictionaryPos = QueryConstants.NULL_INT;` assignment if you use `QueryConstants.NULL_INT` instead of `-1` as the no-entry value.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -307,100 +399,114 @@ private static ParquetFileWriter getParquetFileWriter(
...
+
+        int maxValuesPerPage;
+        int maxOriginalRowsPerPage;
+        int pageCount;
         if (columnSource.getComponentType() != null
                 && !CodecLookup.explicitCodecPresent(writeInstructions.getCodecName(columnDefinition.getName()))
                 && !CodecLookup.codecRequired(columnDefinition)) {
-            targetSize = getTargetSize(columnSource.getComponentType());
-            HashMap<String, ColumnSource<?>> columns = new HashMap<>();
+            int targetRowsPerPage =
```
rcaudy, line 436:
```suggestion
            final int targetRowsPerPage =
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -411,147 +517,198 @@ public Integer get() {
...
-    }
+            List<IntBuffer> arraySizeBuffers = null;
+            if (lengthSource != null) {
+                arraySizeBuffers = new ArrayList<>();
+                try (final ChunkSource.GetContext context = lengthSource.makeGetContext(maxOriginalRowsPerPage);
+                        final RowSequence.Iterator it = originalRowSet.getRowSequenceIterator()) {
+                    while (it.hasMore()) {
+                        final RowSequence rs = it.getNextRowSequenceWithLength(rowStepGetter.getAsLong());
+                        // noinspection unchecked
+                        final IntChunk<Values> chunk = (IntChunk<Values>) lengthSource.getChunk(context, rs);
```
rcaudy, line 683:
```suggestion
                        final IntChunk<? extends Values> chunk = lengthSource.getChunk(context, rs).asIntChunk();
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -411,147 +517,198 @@ public Integer get() {
...
+                final IntBuffer repeatCount = lengthSource != null ? IntBuffer.allocate(maxValuesPerPage) : null;
+                for (int step = 0; step < pageCount; ++step) {
+                    final RowSequence rs = it.getNextRowSequenceWithLength(valuesStepGetter.getAsLong());
+                    transferObject.fetchData(rs);
+                    transferObject.propagateChunkData();
+                    if (lengthIndexIt != null) {
+                        // noinspection unchecked
+                        final IntChunk<Values> lenChunk = (IntChunk<Values>) lengthSource.getChunk(
+                                lengthSourceContext,
+                                lengthIndexIt.getNextRowSequenceWithLength(rowStepGetter.getAsLong()));
```
rcaudy, line 596:
```suggestion
                        final IntChunk<? extends Values> lenChunk = lengthSource.getChunk(
                                lengthSourceContext,
                                lengthIndexIt.getNextRowSequenceWithLength(rowStepGetter.getAsLong())).asIntChunk();
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -411,147 +517,198 @@ public Integer get() {
...
+            final boolean supportNulls = supportNulls(columnType);
+            final Object bufferToWrite = transferObject.getBuffer();
+            final Object nullValue = getNullValue(columnType);
+            try (final RowSequence.Iterator lengthIndexIt =
+                    lengthSource != null ? originalRowSet.getRowSequenceIterator() : null;
+                    final ChunkSource.GetContext lengthSourceContext =
+                            lengthSource != null ? lengthSource.makeGetContext(maxOriginalRowsPerPage) : null;
+                    final RowSequence.Iterator it = dataRowSet.getRowSequenceIterator()) {
+
+                final IntBuffer repeatCount = lengthSource != null ? IntBuffer.allocate(maxValuesPerPage) : null;
```
rcaudy, line 587:
```suggestion
                final IntBuffer repeatCount = lengthSource != null ? IntBuffer.allocate(maxOriginalRowsPerPage) : null;
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -411,147 +517,198 @@ public Integer get() {
...
+                            QueryConstants.NULL_INT);
+            int keyCount = 0;
+            boolean hasNulls = false;
+            try (final ChunkSource.GetContext context = dataSource.makeGetContext(maxRowsPerPage);
+                    final RowSequence.Iterator it = dataRowSet.getRowSequenceIterator()) {
+                for (int curPage = 0; curPage < pageCount; curPage++) {
+                    final RowSequence rs = it.getNextRowSequenceWithLength(valuesStepGetter.getAsLong());
+                    final ObjectChunk<String, ? extends Values> chunk =
+                            dataSource.getChunk(context, rs).asObjectChunk();
+                    final IntBuffer posInDictionary = IntBuffer.allocate((int) rs.size());
```
rcaudy, line 647:
```suggestion
                    final IntBuffer posInDictionary = IntBuffer.allocate(rs.intSize());
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -411,147 +517,198 @@ public Integer get() {
...
+            @NotNull final ColumnSource<DATA_TYPE> dataSource,
+            @Nullable final ColumnSource<?> lengthSource,
+            @NotNull final LongSupplier rowStepGetter,
+            @NotNull final LongSupplier valuesStepGetter,
+            @NotNull final Map<String, Map<CacheTags, Object>> computedCache,
+            final int maxValuesPerPage,
+            final int maxOriginalRowsPerPage,
+            final int pageCount) throws IOException {
+        try (final TransferObject<?> transferObject = getDestinationBuffer(computedCache,
+                originalRowSet,
```
rcaudy, line 572: You should notice that this is the wrong row set for columns of arrays or vectors, which will break for columns of arrays or vectors of `BigDecimal`. Not relevant to this PR, but relevant to the other one you're working on.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -307,100 +399,114 @@ private static ParquetFileWriter getParquetFileWriter(
...
-        RowSet lengthRowSet = null;
-        int targetSize = getTargetSize(columnSource.getType());
-        Supplier<Integer> rowStepGetter;
-        Supplier<Integer> valuesStepGetter;
-        int stepsCount;
+        LongSupplier rowStepGetter;
+        LongSupplier valuesStepGetter;
+
+        int maxValuesPerPage;
+        int maxOriginalRowsPerPage;
```
rcaudy, line 431: Maybe we need a unit test for max > target.

rcaudy (commented):
Marker review. I'm happy with the changes at this stage, pending unit test resolution that Andy previewed may be more complex.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetInstructions.java
```diff
@@ -62,6 +62,32 @@ public static int getDefaultMaximumDictionaryKeys() {
...
+     *
+     * @param newDefaultSizeBytes the new default target page size.
+     */
+    public static void setDefaultTargetPageSize(final int newDefaultSizeBytes) {
+        if (newDefaultSizeBytes < MIN_DEFAULT_PAGE_SIZE) {
+            throw new IllegalArgumentException(
+                    "default page sizes should be larger than " + MIN_DEFAULT_PAGE_SIZE + " bytes");
+        }
+        defaultTargetPageSize = newDefaultSizeBytes;
+    }
```
rcaudy, line 80:
```suggestion
    private static final int MIN_DEFAULT_PAGE_SIZE = 64 << 10;
    private static volatile int defaultTargetPageSize = 1 << 20;

    /**
     * Set the default target page size (in bytes) used to section rows of data into pages during column writing. This
     * number should be no smaller than {@value #MIN_DEFAULT_PAGE_SIZE}.
     *
     * @param newDefaultSizeBytes the new default target page size.
     */
    public static void setDefaultTargetPageSize(final int newDefaultSizeBytes) {
        if (newDefaultSizeBytes < MIN_DEFAULT_PAGE_SIZE) {
            throw new IllegalArgumentException(
                    "Default target page size should be larger than " + MIN_DEFAULT_PAGE_SIZE + " bytes");
        }
        defaultTargetPageSize = newDefaultSizeBytes;
    }
```

rcaudy (approved):

PR #2567, opened 2022-06-22
Follow-up fix for tables loading non-deterministically

niloc132:
Instead of re-ordering the loaded tables, store them in the expected
order in the first place.

Related to #2214

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/widget/plot/JsFigure.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/widget/plot/JsFigure.java
@@ -638,18 +638,18 @@ public class JsFigure extends HasEventHandling {
 
             Promise<?>[] promises = new Promise[response.getTypedExportIdList().length];
 
-            response.getTypedExportIdList().forEach((p0, p1, p2) -> {
-                if (!p0.getType().equals("Table")) {
+            response.getTypedExportIdList().forEach((ticket, index, array) -> {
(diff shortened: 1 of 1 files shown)


web/client-api/src/main/java/io/deephaven/web/client/api/widget/plot/JsFigure.java
```diff
@@ -638,37 +638,31 @@ public Promise fetch(JsFigure figure, FetchObjectResponse response) {
...
+                promises[index] = Callbacks.<ExportedTableCreationResponse, Object>grpcUnaryPromise(c -> {
+                    connection.tableServiceClient().getExportedTableCreationResponse(ticket.getTicket(),
                             connection.metadata(),
                             c::apply);
                 }).then(etcr -> {
                     ClientTableState cts = connection.newStateFromUnsolicitedTable(etcr, "table for figure");
                     JsTable table = new JsTable(connection, cts);
                     // never attempt a reconnect, since we might have a different figure schema entirely
                     table.addEventListener(JsTable.EVENT_DISCONNECT, ignore -> table.close());
+                    tables[index] = table;
```
niloc132, line 660: This was previously `tables[tables.length] = table` when it should have been using `p1` for the subscript as line 651 does. I changed the names of the lambda params to make this more obvious.

nbauernfeind (approved):

mattrunyon (approved):
Working for me locally

PR #2582, opened 2022-06-24
Wrap bound variables created by clients in corresponding Python wrappers 

### LINKED ISSUE #2543: Wrap python script session BindTableToVariableRequest
There's a usability issue when a client binds a table to a variable (`BindTableToVariableRequest`) against a python script session, and later tries to interact with that variable with script code. The table is bound "raw" - ie, without the `deephaven.table.Table` wrapper.

From an implementation standpoint, we need to figure out whether we want to wrap all `io.deephaven.engine.util.ScriptSession#setVariable` in the language-appropriate wrappers, or we only want to do this for `BindTableToVariableRequest`.

jmao-denver:
In order to reuse the PythonObjectWrapper defined in Integrations and not cause circular references, PythonDeephavenSession has be moved to Integrations, which is the first step in the direction to decouple the engine from concrete language script sessions.

Fixes #2543


chipkent (changes requested):

server/jetty/build.gradle
```diff
@@ -5,6 +5,7 @@ plugins {
 
 dependencies {
     api project(':server')
+    compile project(':Integrations')
```
niloc132, line 8: use `implementation` instead of `compile`

server/netty/build.gradle
```diff
@@ -5,6 +5,7 @@ plugins {
 
 dependencies {
     api project(':server')
+    compile project(':Integrations')
```
niloc132, line 8: implementation

server/src/main/java/io/deephaven/server/appmode/ApplicationFactory.java
```diff
@@ -4,17 +4,11 @@
...
 
 import com.google.rpc.Code;
-import io.deephaven.appmode.ApplicationConfig;
-import io.deephaven.appmode.ApplicationContext;
-import io.deephaven.appmode.ApplicationState;
-import io.deephaven.appmode.DynamicApplication;
-import io.deephaven.appmode.QSTApplication;
-import io.deephaven.appmode.ScriptApplication;
-import io.deephaven.appmode.StaticClassApplication;
+import io.deephaven.appmode.*;
```
niloc132, line 7: generally avoid * imports

python-engine-test/build.gradle
```diff
@@ -15,6 +15,7 @@ test {
 }
 
 dependencies {
+  compile project(':Integrations')
```
niloc132, line 18: probably `testCompile` in keeping with the other deps in this project, since there is no non-test code

niloc132: (and technically we'd prefer testImplementation, but this project hasnt been updated to avoid that yet)

py/server/deephaven/_wrapper.py
```diff
@@ -8,14 +8,40 @@
 """
 from __future__ import annotations
 
+import importlib
 import inspect
+# a set of all the directly initializable wrapper classes
```
niloc132, line 13: looks like this comment doesnt belong here?

chipkent (approved):

PR #2583, opened 2022-06-24
Depend on public jpy

devinrsmith:


py/embedded-server/setup.py
```diff
@@ -25,26 +25,35 @@ def normalize_version(version):
...
-      install_requires=['deephaven-jpy=={}'.format(__normalized_version__),
-                        'deephaven=={}'.format(__normalized_version__),
-                        ],
-      )
+setup(
+    name='deephaven-server',
+    version=__normalized_version__,
+    description='Deephaven Embedded Server Python Package',
+    long_description=README,
+    long_description_content_type='text/markdown',
```
devinrsmith, line 33: `long_description_content_type` is necessary for publication.

py/embedded-server/setup.py
```diff
@@ -25,26 +25,35 @@ def normalize_version(version):
...
-      packages=find_namespace_packages(exclude=("tests")),
-      package_data={'deephaven_server': ['jars/*']},
-      url='https://deephaven.io/',
-      author='Deephaven Data Labs',
-      author_email='python@deephaven.io',
-      license='Deephaven Community License',
-      test_loader='unittest:TestLoader',
-      classifiers=[
-          'Development Status :: 2 - Pre-Alpha',
-          'Intended Audience :: Developers, Data Scientists',
```
devinrsmith, line 41: Data Scientists is improper classifier.

py/server/setup.py
```diff
@@ -25,34 +25,40 @@ def normalize_version(version):
...
-      entry_points={
-          'deephaven.plugin': ['registration_cls = deephaven.pandasplugin:PandasPluginRegistration']
-      },
-      )
+setup(
+    name='deephaven-core',
+    version=__normalized_version__,
+    description='Deephaven Engine Python Package',
+    long_description=README,
+    long_description_content_type='text/markdown',
```
devinrsmith, line 33: `long_description_content_type` is necessary for publication

py/server/setup.py
```diff
@@ -25,34 +25,40 @@ def normalize_version(version):
...
-      long_description=README,
-      packages=find_namespace_packages(exclude=("tests", "tests.*", "integration-tests", "test_helper")),
-      url='https://deephaven.io/',
-      author='Deephaven Data Labs',
-      author_email='python@deephaven.io',
-      license='Deephaven Community License',
-      test_loader='unittest:TestLoader',
-      classifiers=[
-          'Development Status :: 2 - Pre-Alpha',
-          'Intended Audience :: Developers, Data Scientists',
```
devinrsmith, line 40: Data Scientists is invalid classifier.

devinrsmith (commented):
Some notes.

devinrsmith:
Nightly run: https://github.com/devinrsmith/deephaven-core/actions/runs/2558392461

buildSrc/src/main/groovy/io.deephaven.java-repository-conventions.gradle
```diff
@@ -1,6 +1,12 @@
 repositories {
     mavenCentral()
-
+    // Uncomment the following if doing local development work that depends on SNAPSHOTs
```
niloc132, line 3: probably should remove this? or if helpful generally, guard it behind a -P flag, and add mavenLocal for any locally installed builds?

personally i'd prefer it not be there, since it could have surprising other effects, but i can see either way

docker/server/src/main/server-tensorflow/requirements.txt
```diff
@@ -10,10 +10,11 @@ gast==0.4.0
 google-auth==2.8.0
 google-auth-oauthlib==0.4.6
 google-pasta==0.2.0
-grpcio==1.46.3
+grpcio==1.47.0
```
niloc132, line 13: why advance this past the 1.46 that the server (and java client) presently uses?

niloc132: that said, make sure all py modules get updated accordingly - at a glance i see sphinx (for our docs) is still using 1.46.0, not changed in this diff

jmao-denver (approved):

py/embedded-server/deephaven_server/start_jvm.py
```diff
@@ -88,66 +95,3 @@ def start_jvm(
         config=config)
     import jpy
     jpy.VerboseExceptions.enabled = True
-
-
-def _expandLinks(dirname):
```
niloc132, line 93: these specifically existed in case a caller was specifying wildcard paths to their own extra jars

py/embedded-server/java-runtime/build.gradle
```diff
@@ -21,6 +21,10 @@ dependencies {
     runtimeOnly project(':logback-print-stream-globals')
     runtimeOnly project(':logback-logbuffer')
     Classpaths.inheritLogbackClassic(project)
+
+    if (!hasProperty('excludeHotspotImpl')) {
```
niloc132, line 25: this might not be intuitive, but probably worth having

docker/server/src/main/server-tensorflow/requirements.txt
devinrsmith: This is a requirement inherited from tensorflow - the dependencies regularly get updated. As part of update adding `jpy` as a proper base image dependency, some of the transitive dependencies got updated as well. https://github.com/deephaven/deephaven-base-images/pull/29

py/embedded-server/deephaven_server/start_jvm.py
```diff
@@ -39,7 +51,6 @@ def start_jvm(
         jvm_args = DEFAULT_JVM_ARGS,
         jvm_properties = DEFAULT_JVM_PROPERTIES,
         java_home: str = os.environ.get('JAVA_HOME', None),
-        classpath = [],
```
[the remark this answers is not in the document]
devinrsmith: I removed it in the sake of simplifying the code. I might even suggest we should limit the parameters even more (ie, not give the user ability to change args, properties, etc). Users who really need it can do all of this logic themselves, I think? I'm not against adding it back in (in a modified form), but I do think it may be functionality that initial users won't necessarily need.

Users can still call start_jvm, yeah?

py/embedded-server/deephaven_server/start_jvm.py
devinrsmith: Note: java has the mechanisms accepting wildcards natively - I don't think we should try to do the expanding ourselves if we re-introduce custom classpaths.

py/jpy-integration/src/pythonToJava/python/test/test_jpy.py
```diff
@@ -1,6 +1,5 @@
-import unittest
 import jpyutil
-import jpy
```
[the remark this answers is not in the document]
devinrsmith: We were getting lucky before; our integration environment matched the environment where jpy was built - so it was able to resolve libjvm.so via jpy. I'll attach error shortly.

py/embedded-server/java-runtime/src/main/java/io/deephaven/python/server/EmbeddedServer.java
```diff
@@ -77,7 +78,8 @@ public EmbeddedServer(String host, Integer port, PyObject dict) throws IOExcepti
     public void start() throws Exception {
         server.run();
         checkGlobals(scriptSession.get(), null);
-        System.out.println("Server started on port " + server.server().getPort());
+        System.out.println("Server started on: "
```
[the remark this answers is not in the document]
devinrsmith: Yes - I'll pull out this for separate consideration.

py/embedded-server/deephaven_server/start_jvm.py
```diff
@@ -59,24 +72,24 @@ def start_jvm(
     # Append user-created args, allowing them to override these values
     system_properties.update(jvm_properties)
 
-    # Build the classpath and expand wildcards, if any
-    jvm_classpath = [os.path.join(os.path.dirname(__file__), 'jars', '*.jar')]
-    jvm_classpath.extend(classpath)
-    jvm_classpath = _expandWildcardsInList(jvm_classpath)
+    # Expand the classpath, so a user can resolve wildcards
+    expanded_classpath = list(itertools.chain.from_iterable(iglob(e, recursive=True) for e in extra_classpath))
```
devinrsmith, line 76: @jmao-denver - can you give this a lookover - I think it replaces the previous functionality. I've tested it out, and it seems to work.

py/embedded-server/deephaven_server/start_jvm.py
niloc132: Note that while java does have this mechanism from the command line, it does not work when starting the jvm as we do (at least not in my testing, which IIRC was java 11).

buildSrc/src/main/groovy/io.deephaven.java-repository-conventions.gradle
devinrsmith: I don't like the idea of `-P` guards in buildSrc. Maybe we've already done that elsewhere? That said, I think there are some development workflows we may want to make easier now that jpy is external (and deephaven-csv). I'll remove for now.

py/embedded-server/deephaven_server/start_jvm.py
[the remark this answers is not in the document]
devinrsmith: Also, have successfully tested it out w/ `extra_classpath = ["/tmp/my-libs/*.jar"]`

py/embedded-server/java-runtime/build.gradle
devinrsmith: This is how the server apps build up their classpath... hotspot-impl was previously missing from the embedded classpath, even though we are adding the jvm flags appropriately.

niloc132: Right, I think they "conflicted" when they were added, so it didn't get put here. My remark is mostly about the fact that no user will have the chance to remove this dependency, the flag is effectively only good for local debugging, since you would need to produce and install a custom wheel.

niloc132 (dismissed):
I think you might still be working on the py 3.9 bug - but one final request either in this ticket or in a post-release followup, that we broaden the testing a bit, and get CI to confirm on multiple py versions that we can at least start up the server, run something simple, and shut it down safely.

py/server/build.gradle
```diff
@@ -3,7 +3,7 @@ plugins {
 }
 
 dependencies {
-    pythonWheel project(path: ':deephaven-jpy', targetConfiguration: 'pythonWheel')
+
```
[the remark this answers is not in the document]
devinrsmith: Done

devinrsmith:
Follow-up: https://github.com/deephaven/deephaven-core/issues/2593

devinrsmith:
Follow-up: https://github.com/deephaven/deephaven-core/issues/2592

niloc132 (approved):

jcferretti (approved):

Issue #2596, opened 2022-06-28
Ticking table join operations should raise an exception when invoked without holding a UGP lock


jmao-denver:
These join operations are considered to be unsafe, but an exception isn't guaranteed when they are performed without holding the UGP locks. This behavior isn't consistent with other unsafe Table operations.

When resolved, skipped test cases in py/server/test/test_ugp.py needs to be un-skipped.

rcaudy:
This should generally be checked and enforced for methods that are not annotated `@ConcurrentMethod`.

Issue #2522, opened 2022-06-13
Python embedded server needs type annotations and better documentation


chipkent:
The python embedded server needs type annotations on all methods.  

Also, many methods are incompletely documented.  For example the `Server` constructor does not document the inputs.  For example, if no host and no port are provided, is the web IDE spun up?  If it is spun up, what is the port?  etc.

jmao-denver:
Spoke with @devinrsmith about some of the changes he made that aren't easy to document clearly and he pointed out that there are more work that needs to be done WRT to custom configuration of the server such as SSL etc. This means we might need to change how the server is created and configured quite a bit, and any additional doc effort should probably come after those changes.

PR #2597, opened 2022-06-28
Expose UGP locking for use in py server

### LINKED ISSUE #2492: Improved Python UGP locking for Deephaven-as-a-Library and Jupyter
Our current DHaaL implementation requires explicit UGP locking for many operations.  This is very explicit but can be inconvenient. 
 Additionally, these locking methods are currently in the private `deephaven._ugp` module. 

I propose the following changes:
1) Rename the `deephaven._ugp` module to `deephaven.ugp` to make it public.
2) Add annotations to `deephaven.ugp` that allow users to annotate a function so that it executes under a UGP lock
3) Add an "eager" execution mode that automatically obtains a UGP lock, when one is not already held.  This trades explicit locking for ease of use for use cases like Jupyter.  It could be easily implemented via an annotation on relevant methods.

Eagar execution would be enabled via something like:
```
from deephaven import ugp
ugp.ugp_eagar_mode(True)
```

This proposal is based off of design changes made between TensorFlow v1 and v2.

Here is a summary of the relevant TF 2 design changes.
https://github.com/tensorflow/community/blob/master/rfcs/20180918-functions-not-sessions-20.md
Here is a simple example.
https://www.tensorflow.org/guide/basics

jmao-denver:
1. rename _ugp.py to ugp.py
2. decorators for running user functions under UGP locks
3. auto_locking flag to run unsafe table ops under UGP shared lock
4. auto_locking decorator (internal use only) to annotate unsafe Table operation methods

Fixes #2492 

Needs #2596 to be resolved to un-skip tests on joins


devinrsmith:
I don't think you are referencing the correct ticket w/ #2522

PR #2609, opened 2022-06-30
Add UpdateByTable QST

### LINKED ISSUE #2608: UpdateByTable QST impl

devinrsmith:
Fixes #2608

### DIFF
--- a/engine/api/src/main/java/io/deephaven/engine/table/Table.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/Table.java
@@ -1726,33 +1726,33 @@ public interface Table extends
     // -----------------------------------------------------------------------------------------------------------------
 
     /**
-     * <p>
      * Create a table with the same {@link #getRowSet() rowSet} as its parent that will perform the specified set of row
-     * based operations to it. As opposed to {@link #update(String...)} these operations are more restricted but are
-     * capable of processing state between rows. This operation will group the table by the specified set of keys if
-     * provided before applying the operation.
-     * </p>
+     * based operation to it.
      *
-     * @param control the {@Link UpdateByControl control} to use when updating the table.
-     * @param operations the operations to apply to the table.
-     * @param byColumns the columns to group by before applying.
-     * @return a table with the same index, with the specified operations applied to each group defined by the
-     *         {@code byColumns}
+     * @param operation the operation to apply to the table.
+     * @return a table with the same index, with the specified operations applied to the entire table
      */
     @ConcurrentMethod
-    Table updateBy(@NotNull final UpdateByControl control,
-            @NotNull final Collection<? extends UpdateByClause> operations,
-            final Collection<? extends Selectable> byColumns);
+    Table updateBy(@NotNull final UpdateByClause operation);
 
     /**
-     * <p>
+     * Create a table with the same {@link #getRowSet() rowSet} as its parent that will perform the specified set of row
+     * based operations to it. As opposed to {@link #update(String...)} these operations are more restricted but are
+     * capable of processing state between rows.
+     *
+     * @param operations the operations to apply to the table.
+     * @return a table with the same index, with the specified operations applied to the entire table.
+     */
+    @ConcurrentMethod
+    Table updateBy(@NotNull final Collection<? extends UpdateByClause> operations);
+
+    /**
      * Create a table with the same {@link #getRowSet() rowSet} as its parent that will perform the specified set of row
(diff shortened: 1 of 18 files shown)


engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -1761,6 +1761,23 @@ Table updateBy(@NotNull final UpdateByControl control,
...
+     * </p>
+     *
+     * @param operations the operations to apply to the table.
+     * @param byColumns the columns to group by before applying.
+     * @return a table with the same index, with the specified operations applied to each group defined by the
+     *         {@code byColumns}
+     */
+    @ConcurrentMethod
+    Table updateBy(@NotNull Collection<? extends UpdateByClause> operations,
+            @NotNull Collection<? extends Selectable> byColumns);
```
devinrsmith, line 1779: Note: I left this out of your original review (b/c it was easy for me to add in this PR); but from a QST point of view, I need to be able to call updateBy w/ these exact parameters. Usage in TableAdapterImpl.

table-api/src/main/java/io/deephaven/api/TableOperations.java
```diff
@@ -517,6 +519,23 @@ TOPS raj(TABLE rightTable, Collection<? extends JoinMatch> columnsToMatch,
 
     // -------------------------------------------------------------------------------------------
 
+    TOPS updateBy(UpdateByClause operation);
```
rcaudy, line 522: Should the JavaDoc be lifted up to here, rather than on `Table`?

table-api/src/main/java/io/deephaven/api/TableOperations.java
```diff
@@ -517,6 +519,23 @@ TOPS raj(TABLE rightTable, Collection<? extends JoinMatch> columnsToMatch,
...
+
+    TOPS updateBy(UpdateByClause operation, final String... byColumns);
```
rcaudy, line 524: I've been changing the order (in another branch) for aggs to keep the no-key variants separate from the variants that have `byColumns`. I think the organization makes more logical sense that way.

qst/src/main/java/io/deephaven/qst/table/UpdateByTable.java
```diff
@@ -0,0 +1,49 @@
...
+import io.deephaven.api.updateby.UpdateByControl;
+import org.immutables.value.Value.Check;
+import org.immutables.value.Value.Immutable;
+
+import java.util.List;
+import java.util.Optional;
+
+@Immutable
+@NodeStyle
+public abstract class UpdateByTable extends ByTableBase {
```
rcaudy, line 17: This class hierarchy surprised me a little bit, but I think it makes sense.

engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -1761,6 +1761,23 @@ Table updateBy(@NotNull final UpdateByControl control,
     Table updateBy(@NotNull final UpdateByControl control,
             @NotNull final Collection<? extends UpdateByClause> operations);
 
+    /**
+     * <p>
+     * Create a table with the same {@link #getRowSet() rowSet} as its parent that will perform the specified set of row
+     * based operations to it. As opposed to {@link #update(String...)} these operations are more restricted but are
+     * capable of processing state between rows.This operation will group the table by the specified set of keys if
```
rcaudy, line 1768:
```suggestion
     * capable of processing state between rows. This operation will group the table by the specified set of keys if
```

engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -1761,6 +1761,23 @@ Table updateBy(@NotNull final UpdateByControl control,
...
+     * provided
```
rcaudy, line 1769: Truncated. I think you mean 
```suggestion
     * provided before applying the operations.
```

table-api/src/main/java/io/deephaven/api/TableOperations.java
devinrsmith: Maybe? At least right now, the javadoc refers to `#getRowSet` which isn't a concept at the TableOperations level. We've been a bit inconsistent in these regards - some of the TableOperations have javadoc, others just on Table.

table-api/src/main/java/io/deephaven/api/TableOperations.java
devinrsmith: Done

rcaudy (approved):

PR #2615, opened 2022-07-01
Add PartitionedTable keys() helper method

### LINKED ISSUE #2520: PartitionedTable in Python needs a better way to get key values
The current documentation for getting key values from partitioned tables is nasty.  We need a helper method.

See:
https://deephaven.io/core/docs/how-to-guides/partition-by/#get-keys-in-a-partitionedtable

jmao-denver:
It is equivalent to performing the select_distinct(key_columns) op on the underlying partitioned table.

Fixes #2520

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -1455,6 +1455,10 @@ class PartitionedTable(JObjectWrapper):
             self._key_columns = list(self.j_partitioned_table.keyColumnNames().toArray())
         return self._key_columns
 
+    def keys(self) -> Table:
+        """Returns a Table containing all the keys of the underlying partitioned table."""
+        return self.table.select_distinct(self.key_columns)
+
     @property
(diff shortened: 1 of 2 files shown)


py/server/deephaven/table.py
```diff
@@ -1455,6 +1455,10 @@ def key_columns(self) -> List[str]:
             self._key_columns = list(self.j_partitioned_table.keyColumnNames().toArray())
         return self._key_columns
 
+    def keys(self) -> Table:
+        """Returns a Table containing all the keys of the underlying partitioned table."""
+        return self.table.select_distinct(self.key_columns)
```
rcaudy, line 1460: If `uniqueKeys` this should just be `view` (or maybe `select`), not `select_distinct`.

rcaudy (changes requested):

rcaudy (approved):

PR #2632, opened 2022-07-08
Wrap PartitionedTableProxy in Python

### LINKED ISSUE #2447: Wrap PartitionedTable.Proxy / TableOperations
`PartitionedTable.Proxy` extends a sub-interface of `Table`,  called `TableOperations`, in addition to a method `target()` for getting the enclosing `PartitionedTable`.

`TableOperations` currently has no analog in the server-side Python API. Note that it is the *same* interface implemented by client-side proxy tables in Java.

We would like to be able to enable this functionality for Python users.

jmao-denver:
Not including updateBy (#2625) and absSumBy, wsumBy, ... (#2631)

Fixes #2447

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -1673,8 +1673,9 @@ class PartitionedTable(JObjectWrapper):
             require_matching_keys (bool): whether to ensure that both partitioned tables have all the same keys
                 present when an operation uses this PartitionedTable and another PartitionedTable as inputs for a
                 :meth:`~PartitionedTable.partitioned_transform`, default is True
-            sanity_check_joins (bool): whether to check that for join operations involving two proxies, a given join
-                key can only be found in one constituent table of both proxied partitioned tables, default is True
+            sanity_check_joins (bool): whether to check that for proxied join operations, a given join key only occurs 
+            in exactly one constituent table of the underlying partitioned table. If the other table argument is also a 
+            PartitionedTableProxy, its constituents will also be subjected to this constraint. 
         """
         return PartitionedTableProxy(
             j_pt_proxy=self.j_partitioned_table.proxy(require_matching_keys, sanity_check_joins))
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -2,8 +2,8 @@
 # Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
 #
 
-""" This module implements the Table and PartitionedTable classes which are the main instruments for working with
-Deephaven refreshing and static data."""
+""" This module implements the Table, PartitionedTable and PartitionedTableProxy classes which are the main
+instruments for working with Deephaven refreshing and static data. """
 
 from __future__ import annotations
 
@@ -1672,7 +1672,7 @@ class PartitionedTable(JObjectWrapper):
         Args:
             require_matching_keys (bool): whether to ensure that both partitioned tables have all the same keys
                 present when an operation uses this PartitionedTable and another PartitionedTable as inputs for a
-                partitioned_transform, default is True
+                :meth:`~PartitionedTable.partitioned_transform`, default is True
             sanity_check_joins (bool): whether to check that join operations will only find a given join key in
                 one constituent table and the right table argument if it is also a proxy, default is True
         """
@@ -1689,7 +1689,7 @@ class PartitionedTableProxy(JObjectWrapper):
         target (PartitionedTable): the underlying partitioned table of the proxy
         require_matching_keys (bool): whether to ensure that both partitioned tables have all the same keys
             present when an operation uses this PartitionedTable and another PartitionedTable as inputs for a
-            partitioned_transform, default is True
+            :meth:`~PartitionedTable.partitioned_transform`, default is True
         sanity_check_joins (bool): whether to check that join operations will only find a given join key in
             one constituent table and the right table argument if it is also a proxy, default is True
     """
@@ -1760,16 +1760,19 @@ class PartitionedTableProxy(JObjectWrapper):
             raise DHError(e, "reverse operation on the PartitionedTableProxy failed.") from e
 
     @auto_locking_op
-    def snapshot(self, source_table: Table, do_init: bool = False,
+    def snapshot(self, source_table: Union[Table, PartitionedTableProxy], do_init: bool = False,
                  cols: Union[str, List[str]] = None) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.snapshot` table operation to all constituent tables of the underlying
-        partitioned table with the provided source table, and produces a new PartitionedTableProxy with the result
-        tables as the constituents of its underlying partitioned table.
+        partitioned table with the provided source table or PartitionedTableProxy, and produces a new
+        PartitionedTableProxy with the result tables as the constituents of its underlying partitioned table.
+
+        In the case of source table being another PartitionedTableProxy, the :meth:`~Table.snapshot` table operation
+        is applied to the matching pairs of the constituent tables from both underlying partitioned tables.
 
         Note, the constituent tables are often time tables that add new rows at a regular, user-defined interval.
 
         Args:
-            source_table (Table): the table to be snapshot
+            source_table (Union[Table, PartitionedTableProxy]): the table or PartitionedTableProxy to be snapshot
             do_init (bool): whether to snapshot when this method is initially called, default is False
             cols (Union[str, List[str]]): names of the columns of the constituent table to be included in the snapshot,
                 default is None, meaning all the columns
@@ -1782,7 +1785,8 @@ class PartitionedTableProxy(JObjectWrapper):
         """
         try:
             cols = to_sequence(cols)
-            return PartitionedTableProxy(j_pt_proxy=self.j_pt_proxy.snapshot(source_table.j_table, do_init, *cols))
+            table_op = jpy.cast(source_table.j_object, _JTableOperations)
+            return PartitionedTableProxy(j_pt_proxy=self.j_pt_proxy.snapshot(table_op, do_init, *cols))
         except Exception as e:
             raise DHError(e, "snapshot operation on the PartitionedTableProxy failed.") from e
 
@@ -2008,11 +2012,14 @@ class PartitionedTableProxy(JObjectWrapper):
     def natural_join(self, table: Union[Table, PartitionedTableProxy], on: Union[str, Sequence[str]],
                      joins: Union[str, Sequence[str]] = None) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.natural_join` table operation to all constituent tables of the underlying
-        partitioned table with the provided right table, and produces a new PartitionedTableProxy with the result
-        tables as the constituents of its underlying partitioned table.
+        partitioned table with the provided right table or PartitionedTableProxy, and produces a new
+        PartitionedTableProxy with the result tables as the constituents of its underlying partitioned table.
+
+        In the case of the right table being another PartitionedTableProxy, the :meth:`~Table.natural_join` table
+        operation is applied to the matching pairs of the constituent tables from both underlying partitioned tables.
 
         Args:
-            table (Union[Table, PartitionedTableProxy]): the right-table of the join
+            table (Union[Table, PartitionedTableProxy]): the right table or PartitionedTableProxy of the join
             on (Union[str, Sequence[str]]): the column(s) to match, can be a common name or an equal expression,
                 i.e. "col_a = col_b" for different column names
             joins (Union[str, Sequence[str]], optional): the column(s) to be added from the right table to the result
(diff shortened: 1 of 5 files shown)


py/server/deephaven/partitioned_table.py
```diff
@@ -0,0 +1,1175 @@
...
+        """
+        try:
+            cols = to_sequence(cols)
+            return PartitionedTableProxy(j_pt_proxy=self.j_pt_proxy.whereIn(filter_table.j_table, *cols))
+        except Exception as e:
+            raise DHError(e, "where_in operation on the PartitionedTableProxy failed.") from e
+
+    @auto_locking_op
+    def where_not_in(self, filter_table: Table, cols: Union[str, Sequence[str]]) -> PartitionedTableProxy:
+        """Applies the table 'where_not_in' operation to all constituent tables of the underlying partitioned table
```
chipkent, line 575: Another comment on this repeated docs string.  Can we make the docs link to the table method?

py/server/deephaven/partitioned_table.py
```diff
@@ -0,0 +1,1175 @@
...
+        Raises:
+            DHError
+        """
+        try:
+            formulas = to_sequence(formulas)
+            return PartitionedTableProxy(j_pt_proxy=self.j_pt_proxy.update(*formulas))
+        except Exception as e:
+            raise DHError(e, "update operation on the PartitionedTableProxy failed.") from e
+
+    @auto_locking_op
```
chipkent, line 656: @rcaudy should check all auto locking

py/server/deephaven/partitioned_table.py
```diff
@@ -0,0 +1,1175 @@
...
+            if joins:
+                joins = [_JJoinAddition.parse(p) for p in joins]
+
+            on = j_array_list(on)
+            joins = j_array_list(joins)
+
+            if match_rule is AsOfMatchRule.LESS_THAN_EQUAL:
+                match_rule = _JAsOfJoinRule.LESS_THAN_EQUAL
+            else:
+                match_rule = _JAsOfJoinRule.LESS_THAN
```
chipkent, line 828: I would make this an explicit check and then raise an error for the default case to protect us in case someone changes the java.

py/server/deephaven/partitioned_table.py
```diff
@@ -0,0 +1,1175 @@
...
+            if joins:
+                joins = [_JJoinAddition.parse(p) for p in joins]
+
+            on = j_array_list(on)
+            joins = j_array_list(joins)
+
+            if match_rule is AsOfMatchRule.GREATER_THAN_EQUAL:
+                match_rule = _JReverseAsOfJoinRule.GREATER_THAN_EQUAL
+            else:
+                match_rule = _JReverseAsOfJoinRule.GREATER_THAN
```
chipkent, line 869: I would make this an explicit check and then raise an error for the default case to protect us in case someone changes the java.

py/server/deephaven/table.py
```diff
@@ -1316,343 +1313,5 @@ def partition_by(self, by: Union[str, Sequence[str]], drop_keys: bool = False) -
...
-        Raises:
-            DHError
-        """
-        try:
-            j_operator = j_binary_operator(func, dtypes.from_jtype(Table.j_object_type.jclass))
-            j_pt = self.j_partitioned_table.partitionedTransform(other.j_partitioned_table, j_operator)
-            return PartitionedTable(j_partitioned_table=j_pt)
-        except Exception as e:
-            raise DHError(e, "failed to transform the PartitionedTable with another PartitionedTable.") from e
+# break circular imports
```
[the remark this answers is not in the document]
chipkent: I would be more ok with `table.partitioned` than with `partitioned_table`.  If remodueling is being done, it needs thought from the user side to avoid extra cognitive load.

py/server/tests/test_pt_proxy.py
```diff
@@ -0,0 +1,235 @@
...
+                self.assertLessEqual(filtered_ct.size, ct.size)
+
+        with self.subTest("where-not-in filter"):
+            filtered_pt_proxy2 = self.pt_proxy.where_not_in(unique_table, cols=["c"])
+            for ct, filtered_ct, filtered_ct2 in zip(self.pt_proxy.target.constituent_tables,
+                                                     filtered_pt_proxy.target.constituent_tables,
+                                                     filtered_pt_proxy2.target.constituent_tables):
+                self.assertEqual(ct.size, filtered_ct.size + filtered_ct2.size)
+
+    def test_usv(self):
```
chipkent, line 91: maybe rename `usv` to `update_select_view` since I thought this was a typo for `use`.

chipkent (changes requested):

py/server/deephaven/partitioned_table.py
```diff
@@ -0,0 +1,1175 @@
...
+        self.j_partitioned_table = j_partitioned_table
+        self._schema = None
+        self._table = None
+        self._key_columns = None
+        self._unique_keys = None
+        self._constituent_column = None
+        self._constituent_changes_permitted = None
+        self._is_refreshing = None
+
+    @classmethod
```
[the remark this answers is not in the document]
jmao-denver: @classmethod is the go-to mechanism in Python for factory methods

py/server/deephaven/table.py
```diff
@@ -1656,3 +1664,849 @@ def partitioned_transform(self, other: PartitionedTable, func: Callable[[Table,
...
+    def proxy(self, require_matching_keys: bool = True, sanity_check_joins: bool = True) -> PartitionedTableProxy:
+        """Makes a proxy that allows table operations to be applied to the constituent tables of this
+        PartitionedTable.
+
+        Args:
+            require_matching_keys (bool): whether to ensure that both partitioned tables have all the same keys
+                present when an operation uses this PartitionedTable and another PartitionedTable as inputs for a
+                :meth:`~PartitionedTable.partitioned_transform`, default is True
+            sanity_check_joins (bool): whether to check that join operations will only find a given join key in
+                one constituent table and the right table argument if it is also a proxy, default is True
```
chipkent, line 1677: Not 100% clear after 2 reads.

py/server/deephaven/table.py
```diff
@@ -1656,3 +1664,849 @@ def partitioned_transform(self, other: PartitionedTable, func: Callable[[Table,
...
+        """
+        return PartitionedTableProxy(
+            j_pt_proxy=self.j_partitioned_table.proxy(require_matching_keys, sanity_check_joins))
+
+
+class PartitionedTableProxy(JObjectWrapper):
+    """A PartitionedTableProxy is a table operation proxy for the underlying partitioned table. It provides methods
+    that apply table operations on the constituent tables of the underlying partitioned table and produce a new
+    partitioned table from the result constituent tables.
```
chipkent, line 1686: Does this produce a new partitioned table or partitioned table proxy?

result -> resulting?

py/server/deephaven/table.py
```diff
@@ -1656,3 +1664,849 @@ def partitioned_transform(self, other: PartitionedTable, func: Callable[[Table,
...
+
+    Attributes:
+        target (PartitionedTable): the underlying partitioned table of the proxy
+        require_matching_keys (bool): whether to ensure that both partitioned tables have all the same keys
+            present when an operation uses this PartitionedTable and another PartitionedTable as inputs for a
+            :meth:`~PartitionedTable.partitioned_transform`, default is True
+        sanity_check_joins (bool): whether to check that join operations will only find a given join key in
+            one constituent table and the right table argument if it is also a proxy, default is True
```
chipkent, line 1694: See comment above on this.

py/server/deephaven/table.py
```diff
@@ -1656,3 +1664,849 @@ def partitioned_transform(self, other: PartitionedTable, func: Callable[[Table,
...
+        Raises:
+            DHError
+        """
+        try:
+            order_by = to_sequence(order_by)
+            if not order:
+                order = (SortDirection.ASCENDING,) * len(order_by)
+            order = to_sequence(order)
+            if len(order_by) != len(order):
+                raise DHError(message="The number of sort columns must be the same as the number of sort directions.")
```
chipkent, line 1816: Making sure that our sort operations cover something like:
```
t.sort(order_by=["A","B",C"], order=SortDirection.DESCENDING)
```
where order is a single item that gets duplicated over the 3 columns.  ... or did we decide not to do this because someone could just do:
```
t.sort(order_by=["A","B",C"], order=[SortDirection.DESCENDING]*3)
```

chipkent (dismissed):

py/server/deephaven/table.py
rcaudy: JavaDoc version for reference:
```
    /**
     * <p>
     * Make a proxy that allows {@link TableOperations table operations} to be applied to the constituent tables of this
     * PartitionedTable.
     * <p>
     * Each operation thus applied will produce a new PartitionedTable with the results as in
     * {@link #transform(UnaryOperator)} or {@link #partitionedTransform(PartitionedTable, BinaryOperator)}, and return
     * a new proxy to that PartitionedTable.
     *
     * @param requireMatchingKeys Whether to ensure that both partitioned tables have all the same keys present when a
     *        proxied operation uses {@code this} and another {@link PartitionedTable} as inputs for a
     *        {@link #partitionedTransform(PartitionedTable, BinaryOperator) partitioned transform}
     * @param sanityCheckJoinOperations Whether to check that proxied join operations will only find a given join key in
     *        one constituent table for {@code this} and the {@link Table table} argument if it is also a
     *        {@link PartitionedTable.Proxy proxy}
     * @return A proxy that allows {@link TableOperations table operations} to be applied to the constituent tables of
     *         this PartitionedTable
     */
```

py/server/deephaven/table.py
```diff
@@ -1656,3 +1664,849 @@ def partitioned_transform(self, other: PartitionedTable, func: Callable[[Table,
...
+            sanity_check_joins (bool): whether to check that join operations will only find a given join key in
+                one constituent table and the right table argument if it is also a proxy, default is True
+        """
+        return PartitionedTableProxy(
+            j_pt_proxy=self.j_partitioned_table.proxy(require_matching_keys, sanity_check_joins))
+
+
+class PartitionedTableProxy(JObjectWrapper):
+    """A PartitionedTableProxy is a table operation proxy for the underlying partitioned table. It provides methods
+    that apply table operations on the constituent tables of the underlying partitioned table and produce a new
```
rcaudy, line 1685:
```suggestion
    that apply table operations to the constituent tables of the underlying partitioned table and produce a new
```

py/server/deephaven/table.py
```diff
@@ -1656,3 +1664,849 @@ def partitioned_transform(self, other: PartitionedTable, func: Callable[[Table,
...
+
+        Raises:
+            DHError
+        """
+        try:
+            return PartitionedTableProxy(j_pt_proxy=self.j_pt_proxy.reverse())
+        except Exception as e:
+            raise DHError(e, "reverse operation on the PartitionedTableProxy failed.") from e
+
+    @auto_locking_op
```
rcaudy, line 1762: I don't see an `is_refreshing` attribute on `PartitionedTableProxy`. I assume auto-locking does not work right now, except based on `Table` RHSs?

py/server/tests/test_pt_proxy.py
```diff
@@ -0,0 +1,300 @@
...
+import unittest
+
+from deephaven import read_csv, empty_table, SortDirection, AsOfMatchRule, DHError
+from deephaven.agg import sum_, avg, pct, weighted_avg, formula, group, first, last, max_, median, min_, std, abs_sum, \
+    var
+from deephaven.table import PartitionedTableProxy
+from tests.testbase import BaseTestCase
+
+
+class PartitionedTableProxyTestCase(BaseTestCase):
```
rcaudy, line 14: If I'm correct about the `is_refreshing` attribute, you may need more testing to ensure that the lock is being acquired when it should be.

py/server/deephaven/table.py
```diff
@@ -1673,32 +1673,37 @@ def proxy(self, require_matching_keys: bool = True, sanity_check_joins: bool = T
             require_matching_keys (bool): whether to ensure that both partitioned tables have all the same keys
                 present when an operation uses this PartitionedTable and another PartitionedTable as inputs for a
                 :meth:`~PartitionedTable.partitioned_transform`, default is True
-            sanity_check_joins (bool): whether to check that join operations will only find a given join key in
-                one constituent table and the right table argument if it is also a proxy, default is True
+            sanity_check_joins (bool): whether to check that for join operations involving two proxies, a given join
+                key can only be found in one constituent table of both proxied partitioned tables, default is True
```
rcaudy, line 1677:
```suggestion
            sanity_check_joins (bool): whether to check that for proxied join operations, a given join key only occurs 
            in exactly one constituent table of the underlying partitioned table. If the other table argument is also a 
            PartitionedTableProxy, its constituents will also be subjected to this constraint. 
```

rcaudy (approved):

rcaudy (approved):

PR #2633, opened 2022-07-08
UpdateBy config changes in support of gRPC

devinrsmith:
These changes allow the consumer to determine if they caller left the values at default, or if they were explicitly set. This may become important when we are sending configuration bits over the wire, where the caller might want to override some values, but not others.

### DIFF
--- a/table-api/src/main/java/io/deephaven/api/updateby/EmaControl.java
+++ b/table-api/src/main/java/io/deephaven/api/updateby/EmaControl.java
@@ -1,10 +1,11 @@
 package io.deephaven.api.updateby;
 
+import org.immutables.value.Value;
 import org.immutables.value.Value.Immutable;
-import org.immutables.value.Value.Default;
 import io.deephaven.annotations.BuildableStyle;
 
 import java.math.MathContext;
+import java.util.Optional;
 
 /**
  * <p>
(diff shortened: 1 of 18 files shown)


table-api/src/main/java/io/deephaven/api/updateby/EmaControl.java
```diff
@@ -29,67 +30,95 @@ public static Builder builder() {
...
      */
-    @Default
-    public MathContext bigValueContext() {
-        return MathContext.DECIMAL128;
+    @Value.Derived
+    public MathContext bigValueContextOrDefault() {
+        return bigValueContext().orElse(MathContext.DECIMAL128);
+    }
+
+    public final EmaControl materialize() {
```
lbooker42, line 113: A comment on this function describing how might be used would be helpful.

table-api/src/main/java/io/deephaven/api/updateby/UpdateByControl.java
```diff
@@ -26,32 +31,88 @@ public static UpdateByControl defaultInstance() {
...
+    public abstract OptionalDouble targetLoadFactor();
+
+    public abstract Optional<MathContext> mathContext();
+
     /**
-     * if redirections should be used for output sources instead of sparse array sources.
+     * If redirections should be used for output sources instead of sparse array sources.
      *
      * <p>
      * Default is `false`. Can be changed with system property {@code UpdateByControl.useRedirection} or by providing a
```
lbooker42, line 93: Should we move these comments (esp re: default values) to the functions where the system properties are read (vs. the `OrDefault()` versions? Probably OK either way but might get overlooked when/if defaults are changed.

Issue #830, opened 2021-07-08
Add more table api Expression structuring


devinrsmith:
`io.deephaven.api.expression.Expression` could use more structuring. Mathematical operators, function calls, etc.

PR #2635, opened 2022-07-11
UpdateBy gRPC

### LINKED ISSUE #2607: UpdateByTable gRPC impl

devinrsmith:
Fixes #2607


proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -267,6 +272,150 @@ message SelectOrUpdateRequest {
...
+  // Throw an exception and abort processing when bad data is encountered.
+  THROW = 2;
+
+  // Allow the bad data to poison the result. This is only valid for use with NaN.
+  POISON = 3;
+}
+
+message UpdateByRequest {
+
+  message Options {
```
niloc132, line 324: IIRC not all languages qualify these with namespacing, it probably would be a good idea to give more descriptive names to avoid collisions.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -267,6 +272,150 @@ message SelectOrUpdateRequest {
...
+    string raw = 3;
+  }
+}
+
+message Selectable {
+  string column_name = 1;
+  Expression expression = 2;
+}
+
+message Pair {
```
niloc132, line 288: convention elsewhere has been to just encode these as `foo=bar`/`foo` strings - ComboAggregateRequest.Aggregate.match_pairs for example.

i see the point that we can be more descriptive by describing it this way, but for `a=a` we have to write the string twice anyways, and a plain string for each of the two fields is also not accurate, since it can only be java identifiers.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -267,6 +272,150 @@ message SelectOrUpdateRequest {
...
+}
+
+message UpdateByRequest {
+
+  message Options {
+    optional bool use_redirection = 1;
+
+    optional int32 chunk_capacity = 2;
+
+    optional double maxStaticSparseMemoryOverhead = 3;
```
niloc132, line 329: snake case for message fields (maximumLoadFactor, targetLoadFactor too)

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -267,6 +272,150 @@ message SelectOrUpdateRequest {
...
+          CumulativeProduct product = 4;
+          Fill fill = 5;
+          Ema ema = 6;
+        }
+      }
+
+      Spec spec = 1;
+      repeated Pair pair = 2;
+    }
+    oneof type {
```
niloc132, line 407: why is this a oneof?

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -267,6 +272,150 @@ message SelectOrUpdateRequest {
...
+      Column column = 1;
+    }
+  }
+
+  Ticket result_id = 1;
+  TableReference source_id = 2;
+  Options options = 3;
+  repeated Operation operations = 4;
+  repeated Selectable group_by = 5;
```
niloc132, line 416: generally `repeated` fields should be plural

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -267,6 +272,150 @@ message SelectOrUpdateRequest {
...
+    HALF_EVEN = 6;
+    UNNECESSARY = 7;
+  }
+  sint32 precision = 1;
+  RoundingMode rounding_mode = 2;
+}
+
+enum BadDataBehavior {
+  // Reset the state for the bucket to {@code null} when invalid data is encountered.
+  RESET = 0;
```
niloc132, line 310: Since 0 is the "This is not set" default, consider making it fail-safe with THROW, so as to not surprise users who fail to set this?

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -267,6 +272,150 @@ message SelectOrUpdateRequest {
   repeated string column_specs = 3;
 }
 
+message Expression {
+  oneof type {
+    string column_name = 1;
+    int64 long_value = 2;
+    string raw = 3;
```
niloc132, line 279: why only raw and long_value? if various other numbers/etc can be raw, why can't long?

Also it seems odd to refer to the "type" as being "column_name" or "raw" - perhaps expression could have a "value" which could be "column reference" or "long literal", "raw X value", etc?

niloc132: Just found the Expression type hierarchy in existing code, and I'm only more confused - why is RawString not a Value, but ColumnName is? Shouldn't ColumnName be a "Reference" or something rather than a Value?

server/src/main/java/io/deephaven/server/table/ops/UpdateByGrpcImpl.java
```diff
@@ -0,0 +1,231 @@
...
+    public UpdateByGrpcImpl() {
+        super(BatchTableRequest.Operation::getUpdateBy, UpdateByRequest::getResultId, UpdateByRequest::getSourceId);
+    }
+
+    @Override
+    TableSpec createTableSpec(UpdateByRequest request) {
+        return adapt(request);
+    }
+
+    private static UpdateByTable adapt(UpdateByRequest request) {
```
niloc132, line 49: Requesting a method name other than "adapt" for the the next 100+ lines of this class

niloc132 (changes requested):
:-1: as it stands - this is introducing a lot (all? not sure) of QST via gRPC in this one new feature, and that means there is a lot of code that isn't obvious from the pull request which is now reachable from any client. Easiest example: a Expression.type.raw is an easy way to smuggle arbitrary java for execution, but there are no safeguards introduced to prevent this (the updateby group_by field may hold those raw expression strings, and engine-table's SelectColumn.ExpressionAdapter will invoke `SelectColumnFactory.getExpression(..)` on them without a second look).

At the very least, if this feature can only be implemented by using QST's Expression types, I would a) like to see the qst protos moved to a new .proto file, and b) a safe-by-default config option preventing the use of the UpdateByGrpcImpl type.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
devinrsmith: That's unfortunate :/ but good to know.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
devinrsmith: I'll add some documentation - but the idea is that `input_column_name` can be empty for the `a=a` case:

```java
    public static Pair adapt(io.deephaven.proto.backplane.grpc.Pair pair) {
        final ColumnName output = ColumnName.of(pair.getOutputColumnName());
        return pair.getInputColumnName().isEmpty() ? output : Pair.of(ColumnName.of(pair.getInputColumnName()), output);
    }
```

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
devinrsmith: Good catch.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
devinrsmith: In context, at least right now, this is always being prefaced w/ `optional`. The idea is that if the client does not provide a value for it, it will inherit the appropriate server defaults (the defaults *vary* depending on the specific field):

```grpc
message UpdateByEmaOptions {
  optional BadDataBehavior on_null_value = 1;
  optional BadDataBehavior on_nan_value = 2;
  ...
```

I'll add documentation at this layer to explain that.

That said, I'm not *against* making `THROW = 0`, but there may be reasons to want the order to be the same as the java enum it is mapping? (We can update the java enum as well...)

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
devinrsmith: column name, long value, and raw are the only plumbed through parts at the table api layer right now, but I agree we should flesh these out more (https://github.com/deephaven/deephaven-core/issues/830). I can complete the loop a bit here in these regards.

As it is right now,

Expression = Value | RawString
Value = ColumnName | long (literal)

Maybe we should add more hierarchy here, with "Reference" and "Literal" instead of "Value"

devinrsmith:
At first I was a bit confused by your comment on "easy way to smuggle arbitrary java for execution" - is there anything preventing the equivalent via a view or select? Looking at the code now though, I see `io.deephaven.server.table.validation.ColumnExpressionValidator#validateColumnExpressions`... it looks like this is the protection you are referring to?

I'll adapt these checks in.

I'm hoping that we can (eventually) migrate from `grpc -> table api` into `grpc -> qst -> table api`, as I think it reduces duplication and provides other benefits IMO.

devinrsmith:
Looking deeper, I'm having trouble finding out if we disable arbitrary code execution via gRPC table operations api today. Afaict, we only do "this looks like a valid operation" validation. I would also argue, this *should* be the responsibility of the engine, not the grpc layer. Trying to get clarification from @rcaudy

java-client/session/src/main/java/io/deephaven/client/impl/PairBuilder.java
```diff
@@ -0,0 +1,23 @@
...
+        }
+        return io.deephaven.proto.backplane.grpc.Pair.newBuilder()
+                .setOutputColumnName(pair.output().name())
+                .setInputColumnName(pair.input().name())
+                .build();
+    }
+
+    private static io.deephaven.proto.backplane.grpc.Pair adapt(ColumnName pair) {
+        return io.deephaven.proto.backplane.grpc.Pair.newBuilder()
+                .setOutputColumnName(pair.name())
```
rcaudy, line 20: It surprised me that you were opting to not make input column name explicit.

server/src/main/java/io/deephaven/server/table/ops/GrpcQstTableOperation.java
```diff
@@ -0,0 +1,70 @@
...
+        } catch (IllegalArgumentException e) {
+            throw GrpcUtil.statusRuntimeException(Code.INVALID_ARGUMENT, e.getMessage());
+        }
+    }
+
+    @Override
+    public final Table create(T request, List<ExportObject<Table>> sourceTables) {
+        final TableSpec tableSpec = createTableSpec(request);
+        final TableCreatorGrpcImpl creator = new TableCreatorGrpcImpl(sourceTables);
+        // todo: UpdateGraphProcessor for certain ops
```
rcaudy, line 47: Needs to be a ticket.

rcaudy (commented):
Minor comments. It's a lot of code for not a lot of functionality.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -267,6 +272,150 @@ message SelectOrUpdateRequest {
   repeated string column_specs = 3;
 }
 
+message Expression {
```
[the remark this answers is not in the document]
devinrsmith: Deleted.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -267,6 +272,150 @@ message SelectOrUpdateRequest {
...
+    UP = 0;
+    DOWN = 1;
+    CEILING = 2;
+    FLOOR = 3;
+    HALF_UP = 4;
+    HALF_DOWN = 5;
+    HALF_EVEN = 6;
+    UNNECESSARY = 7;
+  }
+  sint32 precision = 1;
```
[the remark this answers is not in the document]
devinrsmith: We don't use uint32 or uint64 in our protos. Pinging #ddl-api-dev

server/src/main/java/io/deephaven/server/table/ops/GrpcQstTableOperation.java
devinrsmith: No ticket, is fixed.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
devinrsmith: I've removed this proto.

rcaudy:
I don't hate this. I do sort of feel like the layering is getting excessive. I think the evidence of this can be inferred from how long it took you (@devinrsmith ) and I to trace through some of the table creation logic for parent resolution.

devinrsmith:
Note: this PR also brings selectDistinct in-line with the security of view, update, etc (previously, it was very locked down):

image

rcaudy (commented):
Marker, I've reviewed up to here.

devinrsmith:
I've force pushed removing any changes to TableSpec and gRPC impl via TableSpec. I haven't resolved any concerns wrt the .proto structure yet.

java-client/session-dagger/src/test/java/io/deephaven/client/UpdateOrSelectSessionTest.java
```diff
@@ -57,12 +57,17 @@ public TableSpec apply(TableSpec spec, String[] formulas) {
...
                 return spec.lazyUpdate(formulas);
             }
-        }
+        },
+        // SELECT_DISTINCT {
+        // @Override
+        // public TableSpec apply(TableSpec spec, String[] formulas) {
+        // return spec.selectDistinct(formulas);
+        // }
+        // }
```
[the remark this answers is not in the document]
devinrsmith: I'll remove it, and looks like I need to also fix some merge conflicts.

niloc132 (approved):

PR #2638, opened 2022-07-15
Implemented Open Addressing hashing in `UpdateBy` operations

lbooker42:
Closes #2653


engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/hashing/TypedUpdateByFactory.java
```diff
@@ -0,0 +1,88 @@
...
+    public static void incrementalMoveMainFull(CodeBlock.Builder builder) {}
+
+    public static void incrementalMoveMainAlternate(CodeBlock.Builder builder) {}
+
+    public static void incrementalProbeFound(HasherConfig<?> hasherConfig, boolean alternate,
+            CodeBlock.Builder builder) {
+        if (!alternate) {
+            builder.addStatement("// map the existing bucket to this chunk position");
+            builder.addStatement("outputPositions.set(chunkPosition, rowState)");
+            // builder.beginControlFlow("if (sequentialBuilders != null)");
```
cpwright, line 59: dead code

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/hashing/UpdateByStateManager.java
```diff
@@ -8,33 +8,53 @@
...
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
 /**
  * The basis for operators that participate in an updateBy operation.
  */
-public interface ChunkedUpdateByStateManager {
-    void add(final SafeCloseable bc,
+public abstract class UpdateByStateManager {
```
cpwright, line 18: why class and not interface?

Looking at the actual hash tables there is a fair bit of duplication in the TypedBases.  I wonder if we should be collapsing them together; or if that is more trouble than it is worth.  The duplication of just the key sources seems like not really worth having a class instead of something more swappable.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/hashing/typed/addonlyopen/gen/TypedHashDispatcher.java
```diff
@@ -0,0 +1,64 @@
...
+                        maximumLoadFactor, targetLoadFactor);
+            case Float:
+                return new AddOnlyUpdateByHasherFloat(tableKeySources, originalTableKeySources, tableSize,
+                        maximumLoadFactor, targetLoadFactor);
+            case Double:
+                return new AddOnlyUpdateByHasherDouble(tableKeySources, originalTableKeySources, tableSize,
+                        maximumLoadFactor, targetLoadFactor);
+            case Object:
+                return new AddOnlyUpdateByHasherObject(tableKeySources, originalTableKeySources, tableSize,
+                        maximumLoadFactor, targetLoadFactor);
```
cpwright, line 61: For the final version, we might want to do the double type dispatching so that those versions don't need to wait for compilation.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/hashing/AddOnlyUpdateByStateManagerTypedBase.java
```diff
@@ -0,0 +1,269 @@
...
+        Require.gtZero(tableSize, "tableSize");
+        Require.eq(Integer.bitCount(tableSize), "Integer.bitCount(tableSize)", 1);
+        Require.inRange(maximumLoadFactor, 0.0, 0.95, "maximumLoadFactor");
+
+        mainKeySources = new WritableColumnSource[tableKeySources.length];
+        alternateKeySources = new WritableColumnSource[tableKeySources.length];
+        chunkTypes = new ChunkType[tableKeySources.length];
+
+        for (int ii = 0; ii < tableKeySources.length; ++ii) {
+            chunkTypes[ii] = keySourcesForErrorMessages[ii].getChunkType();
```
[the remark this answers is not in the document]
lbooker42: Agreed, changed for other classes where this was propagated.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/hashing/UpdateByStateManager.java
lbooker42: Re: why class and not interface, having a constructor is useful.

I think the `TypedBase`s should be collapsed.  `Incremental` contains all of `AddOnly` plus probing so should be easy to collapse these into one.

engine/table/src/main/java/io/deephaven/engine/table/impl/BucketedUpdateBy.java
```diff
@@ -1080,8 +1082,8 @@ private BucketedUpdateBy(@NotNull final UpdateByOperator[] operators,
             slotTracker = new UpdateBySlotTracker(control.chunkCapacityOrDefault());
 
             this.hashTable = USE_TYPED_STATE_MANAGER
-                    ? TypedHasherFactory.make(IncrementalUpdateByStateManagerTypedBase.class,
-                            keySources, keySources,
+                    ? TypedHasherFactory.make(UpdateByStateManagerTypedBase.class,
```
cpwright, line 1085: Looks like we are making the same thing; so if we ditch the USE_TYPED_STATE_MANAGER parameter, we can collapse these if blocks a bit.

engine/table/src/main/java/io/deephaven/engine/table/impl/BucketedUpdateBy.java
```diff
@@ -34,11 +36,16 @@
 import static io.deephaven.engine.rowset.RowSequence.NULL_ROW_KEY;
 
 public class BucketedUpdateBy extends UpdateBy {
+    static boolean USE_TYPED_STATE_MANAGER =
```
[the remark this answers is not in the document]
lbooker42: Removed

cpwright (approved):

PR #2644, opened 2022-07-15
Ensure that the server has been properly initialized before import

devinrsmith:
From the context of the python embedded server, this will cause `import deephaven` to fail with an appropriate error message, unless `deephaven_server.start_jvm(...)` has been called.


py/server/deephaven/__init__.py
```diff
@@ -9,6 +9,12 @@
 
 __version__ = "0.15.0"
 
+import deephaven_internal
+try:
+    deephaven_internal.check_server()
+finally:
+    del deephaven_internal
+
```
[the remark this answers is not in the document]
devinrsmith: No - this is exactly what we *don't* want to do. `import jpy` will try (and may succeed) in starting the JVM if it hasn't already been started, and that's one of the conditions we want to guard against.

And even if this method was okay, we still want to guard against the user starting jpy manually. Essentially, `deephaven_internal` will be our implementation entrypoint  that lets us know "yep, Deephaven JVM and Python were properly initialized via Deephaven code".

py/embedded-server/deephaven_server/start_jvm.py
```diff
@@ -98,5 +100,6 @@ def start_jvm(
         jvm_options=jvm_args,
         # config_file=config_file,
         config=config)
+    deephaven_internal.server_ready()
```
niloc132, line 103: this is too early - we should run this only after the Server's constructor has completed (so we know the logging, config is all finished, not just able to start)

niloc132: It might even be worth not doing this at all, and instead letting the java JpyInit set this, since it knows everything is done on the java side.

niloc132: The more I think about it, the more I like that - we can be certain that way that everything ran and was happy on the java side, rather than just assuming it will work out, plus, two fewer lines of code to worry about here.

There is no downside from the perspective of helping a user understand what is wrong - the check will still fail if they try to import deephaven and this flag hasn't been set. If the flag failed to be set (from importing deephaven_server.Server or from instantiating a Server), then the user has that log to look at.

devinrsmith: I think that's a good idea, but unfortunately that will require a bit more work (I'll forge ahead to see how much actual work...).

During startup we do:

```java
public PythonDeephavenSession( ... ) {
...
this.module = (PythonScriptSessionModule) PyModule.importModule("deephaven.server.script_session")
                .createProxy(CallableKind.FUNCTION, PythonScriptSessionModule.class);
 ```
 
 And this causes `deephaven` module to be imported.

As the code exists right now in the PR, variable state I'm really setting is "was/has Deephaven initialized the JVM & Python?". It may be worth having hooks both for that, and for "is the server ready" - and this might make the sphinx and test parts a little more explicit.

Comment (devinrsmith) [reply py/embedded-server/deephaven_server/start_jvm.py:103]: Looks like even if `module` is made to be lazy, it's still called as part of `AbstractScriptSession.publishInitial`

```
at org.jpy.PyLib.importModule(Native Method)
2022-07-20T15:09:04.120006480Z 	at org.jpy.PyModule.importModule(PyModule.java:94)
2022-07-20T15:09:04.120008334Z 	at io.deephaven.integrations.python.PythonDeephavenSession.module(PythonDeephavenSession.java:336)
2022-07-20T15:09:04.120018182Z 	at io.deephaven.integrations.python.PythonDeephavenSession.createDiff(PythonDeephavenSession.java:228)
2022-07-20T15:09:04.120042338Z 	at io.deephaven.integrations.python.PythonDeephavenSession.createDiff(PythonDeephavenSession.java:56)
2022-07-20T15:09:04.120062015Z 	at io.deephaven.engine.util.AbstractScriptSession.applyDiff(AbstractScriptSession.java:139)
2022-07-20T15:09:04.120072144Z 	at io.deephaven.engine.util.AbstractScriptSession.publishInitial(AbstractScriptSession.java:107)
2022-07-20T15:09:04.120093665Z 	at io.deephaven.integrations.python.PythonDeephavenSession.<init>(PythonDeephavenSession.java:97)
2022-07-20T15:09:04.120107261Z 	at io.deephaven.server.console.python.PythonConsoleSessionModule.bindPythonSession(PythonConsoleSessionModule.java:31)
```

File: py/server/deephaven_internal/__init__.py [hunk 3]
```diff
@@ -0,0 +1,15 @@
...
+    _server_is_ready = True
+
+def check_server():
+    """Checks if the server is ready. Raises an exception if the server is not ready."""
+    global _server_is_ready
+    if not _server_is_ready:
+        raise Exception("The Deephaven Server has not been initialized. "
+                        "Please ensure that deephaven_server.Server has been constructed, "
+                        "or that deephaven_server.start_jvm() has been invoked, "
+                        "before importing deephaven.")
```
chipkent, line 15:
```suggestion
        raise Exception("The Deephaven Server has not been initialized. "
                        "Please ensure that deephaven_server.Server has been constructed "
                        "before importing deephaven.")
```
I think in >99% of cases, the user is going to be starting the server.  I don't see any obvious user use cases where they would directly start the JVM.  As a result, I suggest not confusing the most common user.

sphinx/source/conf.py
```diff
@@ -78,12 +78,15 @@
 from glob import glob
 import os
 import jpyutil
+import deephaven_internal
```
chipkent, line 81: @jmao-denver We should use DHaaL to simplify sphinx config.  I created a ticket.  https://github.com/deephaven/deephaven-core/issues/2654

chipkent (changes requested):

py/embedded-server/deephaven_server/start_jvm.py
devinrsmith: While I think you are correct that actually checking that the "server is running" is "best", I think given the lifecycle of how things are initialized, that will be very tricky. I've renamed the state checking and added docs to make clear this is JVM tracking, and where it fits into the lifecycle of java-first and python-first processes.

I'm hoping it's "good enough" for now

py/server/deephaven_internal/__init__.py
devinrsmith: Done

jmao-denver (approved):
LGTM

niloc132 (approved):
Please make sure we have an open followup ticket for this to get it right in the next release - this technically protects the easy path to brokenness, but

PR #2648, opened 2022-07-18
Add heap size API to community

mofojed:
- Port over getWorkerHeapInfo from [Enterprise](https://gitlab.deephaven.io/illumon/iris/-/merge_requests/6626)
- Added new GetHeapInfoRequest, GetHeapInfoResponse, HeapInfo to console.proto
- Added `getHeapInfo()` methods on QueryConnectable

**Testing Done:**
Tested in browser on http://localhost:10000/jsapi/table_basic.html. Executed following snippet:
```
var result = await connection.getWorkerHeapInfo()
console.log('Heap info', result.freeMemory, result.totalHeapSize, result.maximumHeapSize)
```

Resulted in output:
```
Heap info 918892184 1054867456 4294967296
```


server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -200,6 +180,26 @@ public void executeCommand(ExecuteCommandRequest request, StreamObserver<Execute
...
+    @Override
+    public void getHeapInfo(GetHeapInfoRequest request, StreamObserver<GetHeapInfoResponse> responseObserver) {
+        GrpcUtil.rpcWrapper(log, responseObserver, () -> {
+            final SessionState session = sessionService.getCurrentSession();
+            session.nonExport()
+                    .onError(responseObserver)
+                    .submit(() -> {
+                        GetHeapInfoResponse.Builder infoResponse = GetHeapInfoResponse.newBuilder();
+                        HeapInfo.Builder heapInfo = HeapInfo.newBuilder();
+                        heapInfo.setTotalHeapSize(RuntimeMemory.getInstance().totalMemory());
```
niloc132, line 192: just call getInstance() once?

And maybe chain the builder calls, but that's up to you

web/client-api/src/main/java/io/deephaven/web/client/api/JsWorkerHeapInfo.java
```diff
@@ -0,0 +1,31 @@
...
+    private long totalHeapSize;
+
+    public JsWorkerHeapInfo(HeapInfo heapInfo) {
+        this.maximumHeapSize = Long.parseLong(heapInfo.getMaximumHeapSize());
+        this.freeMemory = Long.parseLong(heapInfo.getFreeMemory());
+        this.totalHeapSize = Long.parseLong(heapInfo.getTotalHeapSize());
+    }
+
+    @JsProperty
+    public long getMaximumHeapSize() {
```
niloc132, line 18: don't directly expose long to JS, you'll at best get compiler warnings and at worst sometimes see objects instead of numbers.

instead return double if you're okay with truncating the value, or clean it up with LongWrapper or String

mofojed: Yea truncating the value is probably fine. I'll need to fix this on Enterprise as well.

In testing it was just returning a number,

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/console.proto
```diff
@@ -61,6 +62,19 @@ message StartConsoleResponse {
     io.deephaven.proto.backplane.grpc.Ticket result_id = 1;
 }
 
+message GetHeapInfoRequest {
+    // left empty for future compatibility
+}
+message GetHeapInfoResponse {
+    string error_message = 1;
```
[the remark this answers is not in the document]
mofojed: True, I was copying the ExecuteCommandResponse, really I should just be doing similar to GetConsoleTypes response. I don't think I even need to export the session. I'll clean up.

niloc132 (dismissed):

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/console.proto
```diff
@@ -61,6 +62,15 @@ message StartConsoleResponse {
     io.deephaven.proto.backplane.grpc.Ticket result_id = 1;
 }
 
+message GetHeapInfoRequest {
+    // left empty for future compatibility
+}
+message GetHeapInfoResponse {
+    int64 maximum_heap_size = 1 [jstype=JS_STRING];
+    int64 total_heap_size = 2 [jstype=JS_STRING];
```
devinrsmith, line 70: Does it make sense to copy-paste the relevant javadoc from java.lang.Runtime here?

Also, maybe better to use the same term that Runtime uses: `maximum_memory`, `total_memory`?

engine/table/src/main/java/io/deephaven/engine/table/impl/util/RuntimeMemory.java
```diff
@@ -188,10 +188,24 @@ public void read(final Sample buf) {
...
+     */
+    public long totalMemory() {
+        return currSnapshot.lastTotalMemory;
+    }
+
     /**
      * See {@link Runtime#maxMemory()}.
      */
-    long getMaxMemory() {
+    public long getMaxMemory() {
```
devinrsmith, line 208: Change method name to `maxMemory()`, or others to `getX()`?

mofojed: I'm copying over what was in Enterprise. Not sure if @cpwright has an opinion on this.

cpwright: No opinion.

web/WebDevelopersGuide.md
```diff
@@ -1188,6 +1190,11 @@ as needed.
  * `VariableDefinition created`
  * `VariableDefinition updated`
  * `VariableDefinition removed`
+
+##### Class `HeapInfo`
+ * `long maximumHeapSize` - Maximum heap size of this worker.
```
niloc132, line 1195: these three should be number - that said, this document is not the canonical docs any more, should be in deephaven.io

devinrsmith:
You'll need to run some new steps when update .proto files:

```
./gradlew go:updateProtobuf py-client:updateProto
```

devinrsmith (approved):
LGTM, didn't look at any JS stuff.

niloc132:
> You'll need to run some new steps when update .proto files:
> 
> ```
> ./gradlew go:updateProtobuf py-client:updateProto
> ```

Note that you can also just run `./gradlew updateProto` to hit both of this (and the build should log one or both of those to tell you what it needs).

niloc132 (approved):

PR #2650, opened 2022-07-19
Simplify ColumnExpressionValidator

devinrsmith:


engine/table/src/main/java/io/deephaven/engine/table/impl/select/SwitchColumn.java
```diff
@@ -129,7 +129,7 @@ public String toString() {
     }
 
     public SelectColumn getRealColumn() {
-        return realColumn;
+        return Require.neqNull(realColumn, "realColumn");
```
rcaudy, line 132: There's a bit of inconsistency in this. Some methods check it, others don't, and none are going to give an appropriate error message. Maybe the other accesses should thread through `getRealColumn()`, and maybe it should have a better validation, like:
```
        if (realColumn == null) {
            throw new IllegalStateException(
                    "Real column is not available until this SwitchColumn is initialized; ensure that initInputs or initDef has been called first"
            );
        }
```

server/src/main/java/io/deephaven/server/table/validation/ColumnExpressionValidator.java
```diff
@@ -85,7 +76,7 @@ public class ColumnExpressionValidator extends GenericVisitorAdapter<Void, Void>
                 .collect(Collectors.collectingAndThen(Collectors.toSet(), Collections::unmodifiableSet));
     }
 
-    public static WhereFilter[] validateSelectFilters(final String[] conditionalExpressions, final Table table) {
+    public static WhereFilter[] validateSelectFilters(final String[] conditionalExpressions, Table table) {
```
rcaudy, line 79: Why this change? `table` isn't reassigned that I can see.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/SwitchColumn.java
devinrsmith: Good idea.

rcaudy (approved):

PR #2651, opened 2022-07-19
Update redpanda docker repo

mofojed:
Was getting an unauthorized error from docker.vectorized.io/redpanda:
```
➜  main git:(heap-size-api) PATH=/usr/lib/jvm/java-11-openjdk-amd64/bin/:$PATH ./gradlew :Integrations:composeUp

> Task :Integrations:composeUp
redpanda uses an image, skipping
Building with native build. Learn about native build in Compose here: https://docs.docker.com/go/compose-native-build/
Building with native build. Learn about native build in Compose here: https://docs.docker.com/go/compose-native-build/
Pulling redpanda (docker.vectorized.io/vectorized/redpanda:v21.9.5)...
unauthorized: authentication required
Removing network 7068f1a4e250e19015aa8e4c8be5964a_integrations__default

> Task :Integrations:composeUp FAILED

```

Check to use docker.redpanda.com/vectorized instead, and now it works again.

### DIFF
--- a/containers/python-examples-redpanda/docker-compose.yml
+++ b/containers/python-examples-redpanda/docker-compose.yml
@@ -59,7 +59,7 @@ services:
     - PLAINTEXT://0.0.0.0:29092,OUTSIDE://0.0.0.0:9092
     - --advertise-kafka-addr
     - PLAINTEXT://redpanda:29092,OUTSIDE://localhost:9092
-    image: docker.vectorized.io/vectorized/redpanda:v21.4.13
+    image: docker.redpanda.com/vectorized:v21.4.13
     ports:
     - 9092:9092
     - 29092:29092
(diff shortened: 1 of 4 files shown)


containers/python-examples-redpanda/docker-compose.yml
```diff
@@ -59,7 +59,7 @@ services:
     - PLAINTEXT://0.0.0.0:29092,OUTSIDE://0.0.0.0:9092
     - --advertise-kafka-addr
     - PLAINTEXT://redpanda:29092,OUTSIDE://localhost:9092
-    image: docker.vectorized.io/vectorized/redpanda:v21.4.13
+    image: docker.redpanda.com/vectorized:v21.4.13
```
devinrsmith, line 62: I don't think you want to remove the Redpanda name...

mofojed: Whoops, I'm a dummie

PR #2652, opened 2022-07-19
Create java client tests for view, update, updateView, select, where, and aggBy

devinrsmith:


rcaudy (commented):
A lot of this testing code feels repetitive/redundant, but it seems _fine_.

rcaudy (approved):

PR #2653, opened 2022-07-20
Wrap updateBy operation

### LINKED ISSUE #2625: wrap the new updateBy operation on both Table and PartitionedTableProxy in Python

jmao-denver:
1. added update_by in both Table and PartitionedTableProxy classes
2. add @auto_locking_op to all the operations in PartitionedTableProxy

Fixes #2625

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -19,6 +19,7 @@ from deephaven.column import Column, ColumnType
 from deephaven.filters import Filter
 from deephaven.jcompat import j_array_list, to_sequence, j_unary_operator, j_binary_operator
 from deephaven.ugp import auto_locking_op
+from deephaven.updateby_clause import UpdateByClause
 
 # Table
 _JTableTools = jpy.get_type("io.deephaven.engine.util.TableTools")
@@ -1389,6 +1390,31 @@ class Table(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "failed to create a partitioned table.") from e
 
+    def update_by(self, clauses: Union[UpdateByClause, List[UpdateByClause]],
+                  by: Union[str, List[str]] = None) -> Table:
+        """This method applies the specified row based operations to this table on a cumulative basis within the
+        groups and creates a new table that has the same set of row as this table. As opposed to
+        :meth:`~Table.update` which is strictly individual row based, these operations are capable of processing
+        state between rows within the same row groups which are identified by the provided key columns. If no key
+        columns are provided, all the rows of the table forms a single group.
+
+        Args:
+            clauses (Union[UpdateByClause, List[UpdateByClause]]): the update-by operation definition(s)
+            by (Union[str, List[str]]): the key column name(s) to group the rows of the table
+
+        Returns:
+            a new Table
+
+        Raises:
+            DHError
+        """
+        try:
+            clauses = to_sequence(clauses)
+            by = to_sequence(by)
+            return Table(j_table=self.j_table.updateBy(j_array_list(clauses), *by))
+        except Exception as e:
+            raise DHError(e, "table update-by operation failed.") from e
+
 
 class PartitionedTable(JObjectWrapper):
     """A partitioned table is a table containing tables, known as constituent tables.
@@ -1777,6 +1803,7 @@ class PartitionedTableProxy(JObjectWrapper):
         self.sanity_check_joins = self.j_pt_proxy.sanityChecksJoins()
         self.target = PartitionedTable(j_partitioned_table=self.j_pt_proxy.target())
 
+    @auto_locking_op
     def head(self, num_rows: int) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.head` table operation to all constituent tables of the underlying partitioned
         table, and produces a new PartitionedTableProxy with the result tables as the constituents of its underlying
@@ -1796,6 +1823,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "head operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def tail(self, num_rows: int) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.tail` table operation to all constituent tables of the underlying partitioned
         table, and produces a new PartitionedTableProxy with the result tables as the constituents of its underlying
@@ -1815,6 +1843,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "tail operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def reverse(self) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.reverse` table operation to all constituent tables of the underlying partitioned
         table, and produces a new PartitionedTableProxy with the result tables as the constituents of its underlying
@@ -1862,6 +1891,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "snapshot operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def sort(self, order_by: Union[str, Sequence[str]],
              order: Union[SortDirection, Sequence[SortDirection]] = None) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.sort` table operation to all constituent tables of the underlying partitioned
@@ -1893,6 +1923,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "sort operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def sort_descending(self, order_by: Union[str, Sequence[str]]) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.sort_descending` table operation to all constituent tables of the underlying
         partitioned table, and produces a new PartitionedTableProxy with the result tables as the constituents of its
@@ -1913,6 +1944,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "sort_descending operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def where(self, filters: Union[str, Filter, Sequence[str], Sequence[Filter]] = None) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.where` table operation to all constituent tables of the underlying partitioned
         table, and produces a new PartitionedTableProxy with the result tables as the constituents of its underlying
@@ -1978,6 +2010,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "where_not_in operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def view(self, formulas: Union[str, Sequence[str]]) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.view` table operation to all constituent tables of the underlying partitioned
         table, and produces a new PartitionedTableProxy with the result tables as the constituents of its underlying
@@ -1998,6 +2031,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "view operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def update_view(self, formulas: Union[str, Sequence[str]]) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.update_view` table operation to all constituent tables of the underlying
         partitioned table, and produces a new PartitionedTableProxy with the result tables as the constituents of its
@@ -2060,6 +2094,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "select operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def select_distinct(self, formulas: Union[str, Sequence[str]] = None) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.select_distinct` table operation to all constituent tables of the underlying
         partitioned table, and produces a new PartitionedTableProxy with the result tables as the constituents of its
@@ -2281,6 +2316,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "reverse as-of join operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def group_by(self, by: Union[str, Sequence[str]] = None) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.group_by` table operation to all constituent tables of the underlying
         partitioned table, and produces a new PartitionedTableProxy with the result tables as the constituents of its
@@ -2304,6 +2340,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "group-by operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def agg_by(self, aggs: Union[Aggregation, Sequence[Aggregation]],
                by: Union[str, Sequence[str]] = None) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.agg_by` table operation to all constituent tables of the underlying partitioned
@@ -2328,6 +2365,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "agg_by operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def agg_all_by(self, agg: Aggregation, by: Union[str, Sequence[str]] = None) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.agg_all_by` table operation to all constituent tables of the underlying
         partitioned table, and produces a new PartitionedTableProxy with the result tables as the constituents of its
@@ -2352,6 +2390,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "agg_all_by operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def count_by(self, col: str, by: Union[str, Sequence[str]] = None) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.count_by` table operation to all constituent tables of the underlying partitioned
         table with the provided source table, and produces a new PartitionedTableProxy with the result tables as the
@@ -2376,6 +2415,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "count_by operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def first_by(self, by: Union[str, Sequence[str]] = None) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.first_by` table operation to all constituent tables of the underlying
         partitioned table, and produces a new PartitionedTableProxy with the result tables as the constituents of its
@@ -2399,6 +2439,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "first_by operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def last_by(self, by: Union[str, Sequence[str]] = None) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.last_by` table operation to all constituent tables of the underlying partitioned
         table, and produces a new PartitionedTableProxy with the result tables as the constituents of its underlying
@@ -2422,6 +2463,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "last_by operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def min_by(self, by: Union[str, Sequence[str]] = None) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.min_by` table operation to all constituent tables of the underlying partitioned
         table, and produces a new PartitionedTableProxy with the result tables as the constituents of its underlying
@@ -2445,6 +2487,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "min_by operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def max_by(self, by: Union[str, Sequence[str]] = None) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.max_by` table operation to all constituent tables of the underlying partitioned
         table, and produces a new PartitionedTableProxy with the result tables as the constituents of its underlying
@@ -2468,6 +2511,7 @@ class PartitionedTableProxy(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "max_by operation on the PartitionedTableProxy failed.") from e
 
+    @auto_locking_op
     def sum_by(self, by: Union[str, Sequence[str]] = None) -> PartitionedTableProxy:
         """Applies the :meth:`~Table.sum_by` table operation to all constituent tables of the underlying partitioned
         table, and produces a new PartitionedTableProxy with the result tables as the constituents of its underlying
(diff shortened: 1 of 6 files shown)


rcaudy (dismissed):

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
+#
```
chipkent, line 1: Overall, I have the feeling that the updateby nomenclature is not as well hashed out as something like `agg`.  For example:
1) the `updateby_clause` module name feels much worse than `agg`.  would we be better off with `updateby`, `update_by`, or something else?
2) Nomencluature is inconsistent with other modules.  e.g. `pairs` seems to be roughly the same thing as `cols` in `agg`.
3) Is "clause" the right word?  In the docs, they are called "operators".  What would convey the concept best to users.  Same comment applies to the `update_by` argument name.

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
...
+    """Reset the state for the bucket to {@code null} when invalid data is encountered"""
+
+    SKIP = _JBadDataBehavior.SKIP
+    """Skip and do not process the invalid data without changing state"""
+
+    THROW = _JBadDataBehavior.THROW
+    """Throw an exception and abort processing when bad data is encountered"""
+
+    POISON = _JBadDataBehavior.POISON
+    """Allow the bad data to poison the result. This is only valid for use with NaN"""
```
chipkent, line 47: These look like they are potentially generally useful.  Do they belong specifically in this module?

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
...
+                 on_nan: BadDataBehavior = BadDataBehavior.SKIP,
+                 on_null_time: BadDataBehavior = BadDataBehavior.SKIP,
+                 on_negative_deltatime: BadDataBehavior = BadDataBehavior.THROW,
+                 on_zero_deltatime: BadDataBehavior = BadDataBehavior.SKIP,
+                 big_value_context: MathContext = MathContext.DECIMAL128):
+        """Initializes a EmaControl for evaluating EMAs (exponential moving average) with table update-by operation.
+
+        Args:
+            on_null (BadDataBehavior): the behavior for when null values are encountered, default is SKIP
+            on_nan (BadDataBehavior): the behavior for when Nan values are encountered, default is SKIP
```
chipkent, line 69:
```suggestion
            on_nan (BadDataBehavior): the behavior for when NaN values are encountered, default is SKIP
```

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
...
+from deephaven._wrapper import JObjectWrapper
+from deephaven.jcompat import to_sequence
+
+_JUpdateByClause = jpy.get_type("io.deephaven.api.updateby.UpdateByClause")
+_JBadDataBehavior = jpy.get_type("io.deephaven.api.updateby.BadDataBehavior")
+_JEmaControl = jpy.get_type("io.deephaven.api.updateby.EmaControl")
+_JMathContext = jpy.get_type("java.math.MathContext")
+
+
+class MathContext(Enum):
```
chipkent, line 20: 1) No docstring
2) I see that this is mirroring a Java class named "MathContext".  The word "Context" is heavily overloaded in DH.  Another name would likely be more clear for a user.

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
...
+    @property
+    def j_object(self) -> jpy.JType:
+        return self.j_ema_control
+
+    def __init__(self, on_null: BadDataBehavior = BadDataBehavior.SKIP,
+                 on_nan: BadDataBehavior = BadDataBehavior.SKIP,
+                 on_null_time: BadDataBehavior = BadDataBehavior.SKIP,
+                 on_negative_deltatime: BadDataBehavior = BadDataBehavior.THROW,
+                 on_zero_deltatime: BadDataBehavior = BadDataBehavior.SKIP,
+                 big_value_context: MathContext = MathContext.DECIMAL128):
```
chipkent, line 64: See note above about "context'

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
...
+        Args:
+            on_null (BadDataBehavior): the behavior for when null values are encountered, default is SKIP
+            on_nan (BadDataBehavior): the behavior for when Nan values are encountered, default is SKIP
+            on_null_time (BadDataBehavior): the behavior for when null timestamps are encountered, default is SKIP
+            on_negative_deltatime (BadDataBehavior): the behavior for when negative sample-to-sample time differences are
+                encountered, default is THROW
+            on_zero_deltatime (BadDataBehavior): the behavior for when zero sample-to-sample-time differences are
+                encountered, default is SKIP
+            big_value_context (MathContext): the context to use when processing big numeric values (BigDecimal/BigInteger).
+                default is DECIMAL128.
```
chipkent, line 76: 1) See note above about "context".
2) Is it best to call these "big numeric values" or "arbitrary precision numeric values".  I might lean to the latter, since that sounds like a type, while the current may be interpreted as the magnitude of a number (e.g. 1e6 is big).
3) Maybe a better name is `arb_precision_math`?

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
...
+                   ema_control: EmaControl = None) -> UpdateByClause:
+    """Creates an EMA (exponential moving average) UpdateByClause for the supplied column name pairs, using ticks as
+    the decay unit.
+
+    The formula used is
+        a = e^(-1 / timeScaleTicks)
+        ema_next = a * ema_last + (1 - a) * value
+
+    Args:
+        time_scale_ticks (int): the decay rate (tau) in ticks
```
chipkent, line 116: 1) There is no reason `time_scale_ticks` should be constrained to integers.  It should be a `float` or an `int`.  Maybe a java change is needed as well?
2) `tau` is not used anywhere in the docs.

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
...
+        pairs (Union[str, List[str]]): the input/output column name pairs
```
chipkent, line 117: Seemingly inconsistently named compared to `agg`.  Comment throughout.

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
...
+        return self.j_updateby_clause
+
+
+def ema_tick_decay(time_scale_ticks: int, pairs: Union[str, List[str]],
+                   ema_control: EmaControl = None) -> UpdateByClause:
+    """Creates an EMA (exponential moving average) UpdateByClause for the supplied column name pairs, using ticks as
+    the decay unit.
+
+    The formula used is
+        a = e^(-1 / timeScaleTicks)
```
chipkent, line 112:
```suggestion
        a = e^(-1 / time_scale_ticks)
```
Need to use the pythonicarg names.  Check other cases for this.

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
...
+        ema_next = a * ema_last + (1 - a) * value
+
+    Args:
+        time_scale_ticks (int): the decay rate (tau) in ticks
+        pairs (Union[str, List[str]]): the input/output column name pairs
+        ema_control (EmaControl): defines how special cases should behave,  when None, the default EmaControl settings
+            as specified in :meth:`~EmaControl.__init__` will be used
```
chipkent, line 119:
```suggestion
        ema_control (EmaControl): defines how special cases should behave, when None, the default EmaControl settings
            as specified in :meth:`~EmaControl.__init__` will be used
```

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
...
+        raise DHError(e, "failed to create a tick-decay EMA UpdateByClause.") from e
+
+
+def ema_time_decay(ts_col: str, time_scale_nanos: int, pairs: Union[str, List[str]],
+                   ema_control: EmaControl = None) -> UpdateByClause:
+    """Creates an EMA(exponential moving average) UpdateByClause for the supplied column name pairs, using time as
+    the decay unit.
+
+    The formula used is
+        a = e^(-dt / timeScaleNanos)
```
chipkent, line 144: pythonic variables

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
...
+        ema_next = a * ema_last + (1 - a) * value
+
+     Args:
+        ts_col (str): the column in the source table to use for timestamps
+        time_scale_nanos (int): the decay rate (tau) in nanoseconds
```
chipkent, line 149: 1) `tau` is not listed in the docs.
2) See `time_table`.  The most user friendly use case would be to specify time like '00:00:10', instead of working in nanos.  The methods could support both `int` nanos and `str` values, but we really need string.  If `int` is also supported, the change should also be made to `time_table`.

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
...
+    except Exception as e:
+        raise DHError(e, "failed to create a time-decay EMA UpdateByClause.") from e
+
+
+def cum_sum(pairs: Union[str, List[str]]) -> UpdateByClause:
+    """Creates a CumSum (cumulative sum) update-by clause for the supplied column name pairs.
+
+    Args:
+        pairs (Union[str, List[str]]): the input/output column name pairs; when empty, update_by perform the
+            cumulative sum operation on all the applicable columns.
```
chipkent, line 176:
```suggestion
        pairs (Union[str, List[str]]): the input/output column name pairs; when empty, update_by performs the
            cumulative sum operation on all the applicable columns.
```
Problem is repeated in many places.

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+    try:
+        pairs = to_sequence(pairs)
+        return UpdateByClause(j_updateby_clause=_JUpdateByClause.CumSum(pairs))
+    except Exception as e:
+        raise DHError(e, "failed to create a cumulative sum UpdateByClause.") from e
+
+
+def cum_prod(pairs: Union[str, List[str]]) -> UpdateByClause:
+    """Creates a CumProd (cumulative produce) update-by clause for the supplied column name pairs.
```
chipkent, line 192:
```suggestion
    """Creates a CumProd (cumulative product) update-by clause for the supplied column name pairs.
```

chipkent (changes requested):

py/server/deephaven/updateby_clause.py
jmao-denver: For now these are only used by update-by

py/server/deephaven/updateby_clause.py
jmao-denver: core#2660 was filed to relax the types of the decay parameter to include float.

py/server/deephaven/updateby_clause.py
```diff
@@ -0,0 +1,269 @@
...
+
+    Returns:
+        a UpdateByClause
+
+    Raises:
+        DHError
+    """
+    try:
+        pairs = to_sequence(pairs)
+        return UpdateByClause(j_updateby_clause=_JUpdateByClause.Fill(pairs))
```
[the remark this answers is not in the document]
jmao-denver: #2661 was filed and assigned to @lbooker42

py/server/deephaven/table.py
```diff
@@ -1324,6 +1325,31 @@ def partition_by(self, by: Union[str, Sequence[str]], drop_keys: bool = False) -
...
+    def update_by(self, clauses: Union[UpdateByClause, List[UpdateByClause]],
+                  by: Union[str, List[str]] = None) -> Table:
+        """This method applies the specified row based operations to this table on a cumulative basis within the
+        groups and creates a new table that has the same set of row as this table. As opposed to
+        :meth:`~Table.update` which is strictly individual row based, these operations are capable of processing
+        state between rows within the same row groups which are identified by the provided key columns. If no key
+        columns are provided, all the rows of the table forms a single group.
+
+        Args:
+            clauses (Union[UpdateByClause, List[UpdateByClause]]): the update-by operation definition(s)
```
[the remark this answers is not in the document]
jmao-denver: I agree that the term 'operation' is better and I think we better change 'updateByClause' to 'updateByOperation'.  Since we don't diverge too far from the Java code, which we have been able to do so far, it is best that we do the renaming in Java as well. Thoughts?

py/server/deephaven/updateby_clause.py
chipkent: I understand.  I'm just trying to see if we expect it to stay that way, or if we think there are reasonable odds that they might expand in the future.

py/server/deephaven/table.py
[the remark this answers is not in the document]
chipkent: Ok with me.

py/server/deephaven/table.py
```diff
@@ -1389,6 +1390,30 @@ def partition_by(self, by: Union[str, Sequence[str]], drop_keys: bool = False) -
         except Exception as e:
             raise DHError(e, "failed to create a partitioned table.") from e
 
+    def update_by(self, ops: Union[UpdateByClause, List[UpdateByClause]],
+                  by: Union[str, List[str]] = None) -> Table:
+        """Creates a table with additional columns calculated from window-based aggregations of columns in this table.
+        The aggregations are defined by the provided operations, which support incremental aggregations over the
+        corresponding rows in the this table. The aggregations will apply position or time-based windowing and
+        compute the results over the entire table or each row group as identified by the provided kye columns.
```
chipkent, line 1398:
```suggestion
        compute the results over the entire table or each row group as identified by the provided key columns.
```

py/server/deephaven/updateby.py
```diff
@@ -0,0 +1,276 @@
...
+    """unlimited precision arithmetic, rounding is half-up"""
+
+    DECIMAL32 = _JMathContext.DECIMAL32
+    """a precision setting matching the IEEE 754R Decimal32 format, 7 digits, rounding is half-even"""
+
+    DECIMAL64 = _JMathContext.DECIMAL64
+    """a precision setting matching the IEEE 754R Decimal64 format, 16 digits, rounding is half-even"""
+
+    DECIMAL128 = _JMathContext.DECIMAL128
+    """a precision setting matching the IEEE 754R Decimal128 format, 34 digits, rounding is half-evne"""
```
chipkent, line 33:
```suggestion
    """a precision setting matching the IEEE 754R Decimal128 format, 34 digits, rounding is half-even"""
```

py/server/deephaven/updateby.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+class BadDataBehavior(Enum):
+    """An Enum defining ways to handle invalid data during EMA processing."""
+
+    RESET = _JBadDataBehavior.RESET
+    """Reset the state for the bucket to {@code null} when invalid data is encountered"""
```
chipkent, line 40: Should we use `null` or `None` for python docs?

py/server/deephaven/updateby.py
```diff
@@ -0,0 +1,276 @@
...
+    """Throw an exception and abort processing when bad data is encountered"""
+
+    POISON = _JBadDataBehavior.POISON
+    """Allow the bad data to poison the result. This is only valid for use with NaN"""
+
+
+class EmaControl(JObjectWrapper):
+    """A EmaControl represents control parameters for performing EMAs (exponential moving average) with the table
+    update-by operation. """
+    j_object_type = _JEmaControl
```
chipkent, line 55: I'm trying to think a little down the line here.  Might `EmaControl` be too narrow of a name, especially if we are planning to add simple moving averages?  Do we expect SMA's to use the same control struct (which would need renaming to `MAControl` or such), or do we expect there to be a `SmaControl`?

You could imagine the current control structure being used for other kinds of averaging or aggregations.  Do we want to reuse it or have these other update-by-avgs implement their own control?

This may be a question for @lbooker42

py/server/deephaven/updateby.py
```diff
@@ -0,0 +1,276 @@
...
+    def j_object(self) -> jpy.JType:
+        return self.j_ema_control
+
+    def __init__(self, on_null: BadDataBehavior = BadDataBehavior.SKIP,
+                 on_nan: BadDataBehavior = BadDataBehavior.SKIP,
+                 on_null_time: BadDataBehavior = BadDataBehavior.SKIP,
+                 on_negative_deltatime: BadDataBehavior = BadDataBehavior.THROW,
+                 on_zero_deltatime: BadDataBehavior = BadDataBehavior.SKIP,
+                 big_value_context: MathContext = MathContext.DECIMAL128):
+        """Initializes a EmaControl for evaluating EMAs (exponential moving average) with table update-by operation.
```
chipkent, line 67:
```suggestion
        """Initializes an EmaControl for evaluating EMAs (exponential moving average) with table update-by operation.
```

py/server/deephaven/updateby.py
```diff
@@ -0,0 +1,276 @@
...
+                                  .onNanValue(on_nan.value)
+                                  .onNullTime(on_null_time.value)
+                                  .onNegativeDeltaTime(on_negative_deltatime.value)
+                                  .onZeroDeltaTime(on_zero_deltatime.value)
+                                  .bigValueContext(big_value_context.value).build())
+        except Exception as e:
+            raise DHError(e, "failed to build a EmaControl object.") from e
+
+
+class UpdateByClause(JObjectWrapper):
```
chipkent, line 95: See rename comment.

py/server/deephaven/updateby.py
```diff
@@ -0,0 +1,276 @@
...
+    except Exception as e:
+        raise DHError(e, "failed to create a cumulative sum UpdateByClause.") from e
+
+
+def cum_prod(cols: Union[str, List[str]]) -> UpdateByClause:
+    """Creates a CumProd (cumulative product) update-by clause for the supplied column names.
+
+    Args:
+        cols (Union[str, List[str]]): the column(s) to be operated on, can be renaming expressions, i.e. "new_col =
+            col"; when empty, update_by perform the cumulative produce operation on all the applicable columns.
```
chipkent, line 203:
```suggestion
            col"; when empty, update_by performing the cumulative product operation on all the applicable columns.
```
Check other cases for this.

chipkent: Looks like this is copied in many places

py/server/deephaven/updateby.py
```diff
@@ -0,0 +1,276 @@
...
+            return UpdateByClause(j_updateby_clause=_JUpdateByClause.Ema(ts_col, time_scale, *cols))
+        else:
+            return UpdateByClause(
+                j_updateby_clause=_JUpdateByClause.Ema(ema_control.j_ema_control, ts_col, time_scale, *cols))
+    except Exception as e:
+        raise DHError(e, "failed to create a time-decay EMA UpdateByClause.") from e
+
+
+def cum_sum(cols: Union[str, List[str]]) -> UpdateByClause:
+    """Creates a CumSum (cumulative sum) update-by clause for the supplied column names.
```
chipkent, line 178: Is there a reason that we want `CumSum` and such in these docs, instead of just having:
```suggestion
    """Creates a cumulative sum update-by clause for the supplied column names.
```
Is `CumSum` supposed to have some significant meaning here?  It feels like it is just a reference to a Java type.  If so, are we keeping this to make it easier to search the Javadocs?  Did we do the same in `agg`?

py/server/deephaven/updateby.py
```diff
@@ -0,0 +1,276 @@
...
+    except Exception as e:
+        raise DHError(e, "failed to create a cumulative minimum UpdateByClause.") from e
+
+
+def cum_max(cols: Union[str, List[str]]) -> UpdateByClause:
+    """Creates a CumMax (cumulative maximum) update-by clause for the supplied column names.
+
+    Args:
+        cols (Union[str, List[str]]): the column(s) to be operated on, can be renaming expressions, i.e. "new_col = 
+            col"; when empty, update_by perform the cumulative maximum operation on all the applicable columns.
```
chipkent, line 243:
```suggestion
        cols (Union[str, List[str]]): the column(s) to be operated on, can include expressions to rename the output, i.e. "new_col = 
            col"; when empty, update_by performs the cumulative maximum operation on all the applicable columns.
```
Duplicated in many places.

py/server/deephaven/updateby.py
```diff
@@ -0,0 +1,276 @@
...
+    try:
+        cols = to_sequence(cols)
+        return UpdateByClause(j_updateby_clause=_JUpdateByClause.CumMax(cols))
+    except Exception as e:
+        raise DHError(e, "failed to create a cumulative maximum UpdateByClause.") from e
+
+
+def forward_fill(cols: Union[str, List[str]]) -> UpdateByClause:
+    """Creates a forward fill update-by clause for the supplied column names. Null values in the columns are
+    replaced by the last known non-Null values. This operation is forward only.
```
chipkent, line 260: Should this be:
```suggestion
    replaced by the last known non-null values. This operation is forward only.
```

chipkent (changes requested):

py/server/deephaven/updateby.py
```diff
@@ -175,7 +175,7 @@ def ema_time_decay(ts_col: str, time_scale: Union[int, str], cols: Union[str, Li
 
 
 def cum_sum(cols: Union[str, List[str]]) -> UpdateByClause:
-    """Creates a CumSum (cumulative sum) update-by clause for the supplied column names.
+    """Creates a cumulative sum update-by clause for the supplied column names.
```
chipkent, line 178: If we are doing this change, it needs to happen in many functions.

chipkent (changes requested):

py/server/deephaven/table.py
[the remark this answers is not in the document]
lbooker42: Created PR #2680 that renamed this object to `UpdateByOperations`

py/server/deephaven/updateby.py
lbooker42: Created PR #2680 that renamed this object to `OperationControl`

py/server/deephaven/updateby.py
lbooker42: Created PR #2680 that renamed this object to `OperationControl`

py/server/deephaven/updateby.py
```diff
@@ -0,0 +1,280 @@
...
+
+    DECIMAL64 = _JMathContext.DECIMAL64
+    """a precision setting matching the IEEE 754R Decimal64 format, 16 digits, rounding is half-even"""
+
+    DECIMAL128 = _JMathContext.DECIMAL128
+    """a precision setting matching the IEEE 754R Decimal128 format, 34 digits, rounding is half-even"""
+
+
+class BadDataBehavior(Enum):
+    """An Enum defining ways to handle invalid data during EMA processing."""
```
chipkent, line 37:
```suggestion
    """An Enum defining ways to handle invalid data during update-by operations."""
```

py/server/deephaven/updateby.py
```diff
@@ -0,0 +1,280 @@
...
+            return UpdateByOperation(j_updateby_op=_JUpdateByOperation.Ema(ts_col, time_scale, *cols))
+        else:
+            return UpdateByOperation(
+                j_updateby_op=_JUpdateByOperation.Ema(op_control.j_op_control, ts_col, time_scale, *cols))
+    except Exception as e:
+        raise DHError(e, "failed to create a time-decay EMA UpdateByOperation.") from e
+
+
+def cum_sum(cols: Union[str, List[str]]) -> UpdateByOperation:
+    """Creates a cumulative sum UpdateByOperation for the supplied column names.
```
chipkent, line 178: Here we removed "CumSum".  We did not do that in the other doc strings.  We need to be consistent.

py/server/deephaven/updateby.py
```diff
@@ -0,0 +1,280 @@
...
+    def j_object(self) -> jpy.JType:
+        return self.j_op_control
+
+    def __init__(self, on_null: BadDataBehavior = BadDataBehavior.SKIP,
+                 on_nan: BadDataBehavior = BadDataBehavior.SKIP,
+                 on_null_time: BadDataBehavior = BadDataBehavior.SKIP,
+                 on_negative_deltatime: BadDataBehavior = BadDataBehavior.THROW,
+                 on_zero_deltatime: BadDataBehavior = BadDataBehavior.SKIP,
+                 big_value_context: MathContext = MathContext.DECIMAL128):
+        """Initializes an OperationControl for use with certain UpdateByOperation such as EMAs.
```
chipkent, line 67:
```suggestion
        """Initializes an OperationControl for use with certain UpdateByOperation, such as EMAs.
```

chipkent (changes requested):

chipkent (approved):

PR #2663, opened 2022-07-21
Add wavgBy, wsumBy, absSumBy in both Table and PartitionedTableProxy

### LINKED ISSUE #2631: Add absSumBy, wsumBy, wavgBy operations to both Table and PartitonedTableProxy in Python

jmao-denver:
Fixes #2631

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -1013,6 +1013,49 @@ class Table(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "table sum_by operation failed.") from e
 
+    def abs_sum_by(self, by: Union[str, Sequence[str]] = None) -> Table:
+        """The abs_sum_by method creates a new table containing the absolute sum for each group.
+
+        Args:
+            by (Union[str, Sequence[str]], optional): the group-by column name(s), default is None
+
+        Returns:
+            a new table
+
+        Raises:
+            DHError
+        """
+        try:
+            by = to_sequence(by)
+            if by:
+                return Table(j_table=self.j_table.absSumBy(*by))
+            else:
+                return Table(j_table=self.j_table.absSumBy())
+        except Exception as e:
+            raise DHError(e, "table asb_sum_by operation failed.") from e
+
+    def weighted_sum_by(self, wcol: str, by: Union[str, Sequence[str]] = None) -> Table:
+        """The sum_by method creates a new table containing the weighted sum for each group.
+
+        Args:
+            wcol (str): the name of the weight column
+            by (Union[str, Sequence[str]], optional): the group-by column name(s), default is None
+
+        Returns:
+            a new table
+
+        Raises:
+            DHError
+        """
+        try:
+            by = to_sequence(by)
+            if by:
+                return Table(j_table=self.j_table.wsumBy(wcol, *by))
+            else:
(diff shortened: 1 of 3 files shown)


py/server/deephaven/table.py
```diff
@@ -1034,6 +1077,28 @@ def avg_by(self, by: Union[str, Sequence[str]] = None) -> Table:
         except Exception as e:
             raise DHError(e, "table avg_by operation failed.") from e
 
+    def weighted_avg_by(self, wcol: str, by: Union[str, Sequence[str]] = None) -> Table:
+        """The avg_by method creates a new table containing the weighted average for each group.
```
rcaudy, line 1081:
```suggestion
        """The weighted_avg_by method creates a new table containing the weighted average for each group.
```

py/server/deephaven/table.py
```diff
@@ -1013,6 +1013,49 @@ def sum_by(self, by: Union[str, Sequence[str]] = None) -> Table:
...
+            by = to_sequence(by)
+            if by:
+                return Table(j_table=self.j_table.absSumBy(*by))
+            else:
+                return Table(j_table=self.j_table.absSumBy())
+        except Exception as e:
+            raise DHError(e, "table asb_sum_by operation failed.") from e
+
+    def weighted_sum_by(self, wcol: str, by: Union[str, Sequence[str]] = None) -> Table:
+        """The sum_by method creates a new table containing the weighted sum for each group.
```
rcaudy, line 1038:
```suggestion
        """The weighted_sum_by method creates a new table containing the weighted sum for each group.
```

py/server/deephaven/table.py
```diff
@@ -1034,6 +1077,28 @@ def avg_by(self, by: Union[str, Sequence[str]] = None) -> Table:
         except Exception as e:
             raise DHError(e, "table avg_by operation failed.") from e
 
+    def weighted_avg_by(self, wcol: str, by: Union[str, Sequence[str]] = None) -> Table:
```
chipkent, line 1080: I think there is a question about how best to make things pythonic in the API.  Here we have `weighted_avg_by` which mirrors the java API and is quite verbose.  I looked into what numpy is doing for weighted averages, and it just adds a named argument (https://numpy.org/doc/stable/reference/generated/numpy.average.html).  Should we consider this path?  Similar question for `abs`.

chipkent (changes requested):

py/server/deephaven/table.py
jmao-denver: This is not a very clear-cut case. While being concise is definitely a trait of being Pythonic, being explicit is also one. I have considered to use wavg_by, wsum_by, asum_by for the method names but not happy with 'asum_by' and gave up. Another factor influencing the picking of these names is that we already have 'weighted_avg/sum' in agg.py.

chipkent: I'm not sure how `avg_by(wcol="X", by=["A","B"])` is not explicit.  Having said that, I'm not opposed to what is here.  I'm just pushing the question to confirm that what is here is the best way to do the API for a user.

jmao-denver: I am not saying it is not very explicit but maybe less so than weighted_avg_by(). I think I can live with the names especially since they are sorta consistent with what's in agg already.

chipkent (approved):

Issue #2671, opened 2022-07-22
Upgrade to Gradle >= 7


devinrsmith:

devinrsmith:
Completed by https://github.com/deephaven/deephaven-core/pull/2672

PR #2672, opened 2022-07-23
Upgrade gradle

devinrsmith:


devinrsmith:
Fixes #2671

devinrsmith:
I've verified the netty and jetty application classpaths are the same before and after this PR.

devinrsmith:
Nightly running: https://github.com/devinrsmith/deephaven-core/actions/runs/2722086016

devinrsmith:
Another nightly running: https://github.com/devinrsmith/deephaven-core/actions/runs/2733909434

devinrsmith:
Final nightly? https://github.com/devinrsmith/deephaven-core/actions/runs/2740517179

rcaudy:
When I try to build with this PR squashed into main, I get:
```
Build file '/Users/rcaudy/dev/deephaven/community/review/core/extensions/parquet/table/build.gradle' line: 51

A problem occurred evaluating project ':extensions-parquet-table'.
> Unable to determine proper docker architecture for aarch64
```

Numerics/Numerics.gradle
```diff
@@ -8,8 +8,8 @@ configurations {
 }
 
 dependencies {
-    compile project(':engine-table'),
-            project(':engine-function')
+    implementation project(':engine-table')
+    implementation project(':engine-function')
```
rcaudy, line 12: I think this is wrong, and we just need an `api` dependency on `engine-time`.

open-api/lang-tools/lang-tools.gradle
```diff
@@ -1,18 +1,18 @@
...
 
 Classpaths.inheritCommonsText(project)
 dependencies {
 
-    compile project(':open-api-lang-parser')
-
-    compile project(':engine-table'),
-            project(':engine-function'),
-            project(':IO')
+    api project(':open-api-lang-parser')
```
rcaudy, line 12: Why?

Plot/Plot.gradle
```diff
@@ -10,9 +10,9 @@ configurations {
 
 dependencies {
 
-    compile project(':engine-table'),
-            project(':engine-function'),
-            project(':engine-tuple')
+    implementation project(':engine-table')
+    implementation project(':engine-function')
+    implementation project(':engine-tuple')
```
rcaudy, line 15: I suspect we want an `api` dep on `engine-api`.

web/client-api/client-api.gradle
```diff
@@ -1,14 +1,15 @@
...
     id 'io.deephaven.project.register'
+    id 'java-library'
 }
 
 apply from: "$rootDir/gradle/web-client.gradle"
 
 dependencies {
-    compile project(':web-shared-beans')
-    compile project(':web-client-backplane')
+    api project(':web-shared-beans')
```
rcaudy, line 9: Why?

web/shared-beans/shared-beans.gradle
```diff
@@ -11,7 +12,7 @@ configurations {
 }
 
 dependencies {
-    compile project(':open-api-shared-fu')
+    api project(':open-api-shared-fu')
```
rcaudy, line 15: Why?

buildSrc/src/main/groovy/Classpaths.groovy
```diff
@@ -152,32 +148,32 @@ class Classpaths {
         }
     }
 
-    static void inheritJavaParser(Project p, String name = JAVA_PARSER_NAME) {
-        Configuration compile = compile p
-        addDependency compile, JAVA_PARSER_GROUP, name, JAVA_PARSER_VERSION
+    static void inheritJavaParser(Project p, String configName = JavaPlugin.IMPLEMENTATION_CONFIGURATION_NAME) {
```
rcaudy, line 151: Is having a default `configName` on these methods actually beneficial? I would think it's better to be explicit in the caller, unless we never use anything but the default.

engine/table/build.gradle
```diff
@@ -42,7 +42,7 @@ dependencies {
 
     Classpaths.inheritGroovy(project, 'groovy', 'implementation')
     Classpaths.inheritGroovy(project, 'groovy-json', 'implementation')
-    Classpaths.inheritJavaParser(project)
+    Classpaths.inheritJavaParser(project, 'compile')
```
rcaudy: I am confused by this change (and the other ones in this commit). Shouldn't we be using `compileOnly` or `implementation`? I think you intend this commit to not change anything, and that you're correcting in subsequent commits, in which case this can just be resolved.

engine/table/build.gradle
```diff
@@ -42,7 +42,7 @@ dependencies {
 
     Classpaths.inheritGroovy(project, 'groovy', 'implementation')
     Classpaths.inheritGroovy(project, 'groovy-json', 'implementation')
-    Classpaths.inheritJavaParser(project, 'compile')
+    Classpaths.inheritJavaParser(project, 'api')
```
rcaudy, line 54: Why not `implementation`?

web/shared-beans/shared-beans.gradle
```diff
@@ -21,6 +21,6 @@ dependencies {
 Classpaths.inheritGwt project
 
 // upgrade inherited dependencies to become transitive
-Classpaths.inheritJsInterop(project, 'jsinterop-annotations', 'compile')
+Classpaths.inheritJsInterop(project, 'jsinterop-annotations', 'compileOnly')
 Classpaths.inheritJavaxAnnotations(project)
-Classpaths.inheritElemental(project, 'elemental2-core', 'compile')
+Classpaths.inheritElemental(project, 'elemental2-core')
```
rcaudy, line 26: This illustrates why I think having a default config name is sub-optimal for intelligibility.

gradle/web-gwt-test.gradle
```diff
@@ -21,7 +21,7 @@ task 'gwtTest', type: Test, {
                                'dhTestServer': testServer
         ]
         t.include '**/*TestSuite.class'
-        t.logging.level = 'INFO'
+//        t.logging.level = 'INFO'
```
rcaudy, line 24: Is this intended for merge? Probably needs to be uncommented or deleted.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.9.1-all.zip
-distributionSha256Sum=b13f5d97f08000996bf12d9dd70af3f2c6b694c2c663ab1b545e9695562ad1ee
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.5-all.zip
```
rcaudy, line 3: Wooh!

py/jpy-ext/build.gradle
```diff
@@ -16,10 +16,11 @@ plugins {
...
 
 configurations {
-  compile.extendsFrom fishBase
+  implementation.extendsFrom fishBase
 }
 
 dependencies {
   api 'org.jpyconsortium:jpy:0.11.0'
   api project(':deephaven-jpy-config')
+  api project(':Base')
```
rcaudy, line 25: `api`?

engine/table/build.gradle
devinrsmith: Correct - this commit was meant to keep everything the same and it was fixed one-by-one later.

gradle/web-gwt-test.gradle
devinrsmith: Ah yes, I forgot, this doesn't gradle-ism doesn't work anymore. Will need to see how to fix it...

engine/table/build.gradle
devinrsmith:
```
io/deephaven/server/table/validation/ColumnExpressionValidator.java:6: error: package com.github.javap
arser does not exist                                                                                                                                                
import com.github.javaparser.JavaParser;                        
```

ModelFarm/ModelFarm.gradle
```diff
@@ -11,7 +11,7 @@ configurations {
 }
 
 dependencies {
-    compile project(':engine-table')
+    implementation project(':engine-table')
```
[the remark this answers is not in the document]
devinrsmith: Yep, and should have api on engine-time after digging in.

FishUtil/build.gradle
```diff
@@ -1,10 +1,16 @@
...
 }
 
 dependencies {
-    compile project(':Base'), project(':DataStructures'), project(':IO'), project(':Configuration'), project(':log-factory')
-    compile depCommonsEmail, depCommonsNet
+    implementation project(':Base')
+    implementation project(':DataStructures')
+    implementation project(':IO')
+    implementation project(':Configuration')
+    api project(':log-factory')
```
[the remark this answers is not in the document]
devinrsmith: Yep. Looks like a few projects were passively picking this up. Changed to implementation.

DataStructures/build.gradle
```diff
@@ -1,7 +1,8 @@
 plugins {
     id 'io.deephaven.project.register'
+    id 'java-library'
 }
 
 dependencies {
-    compile project(':Base')
+    api project(':Base')
```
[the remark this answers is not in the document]
devinrsmith: Given that RequirementFailure is never actually declared as part of the methods, I don't *think* we should consider it part of the public API. I was able to make this impl dependency (and added another impl elsewhere).

engine/table/build.gradle
devinrsmith: Even though server uses it as part of "public" api, we really should eventually refactor that - it's really an implementation detail, so I have it explicitly as implementation now for server.

py/jpy-ext/build.gradle
devinrsmith: `JpyConfigExt implements LogOutputAppendable`

Base/build.gradle
```diff
@@ -1,14 +1,16 @@
...
     id 'io.deephaven.project.register'
+    id 'java-library'
 }
 
 dependencies {
 
-    compile depTrove3, depAnnotations, depCommonsCompress
+    implementation depTrove3
+    compileOnlyApi depAnnotations
+    api depCommonsCompress
```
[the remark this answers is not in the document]
devinrsmith: removed

Numerics/Numerics.gradle
devinrsmith: It does need impl on function, but not table. Added engine-time b/c it does have public DateTime

open-api/lang-tools/lang-tools.gradle
devinrsmith: Changed to implementation.

Plot/Plot.gradle
devinrsmith: Done

buildSrc/src/main/groovy/Classpaths.groovy
devinrsmith: done

web/client-api/client-api.gradle
devinrsmith: `public class TableHandle extends ServerReplyHandle`

web/shared-beans/shared-beans.gradle
devinrsmith: `public class TableHandle extends ServerReplyHandle`

web/client-api/client-api.gradle
devinrsmith: Part of web-client-api public api.

web/shared-beans/shared-beans.gradle
devinrsmith: fixed

ModelFarm/ModelFarm.gradle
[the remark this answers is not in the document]
rcaudy: `engine-api` declares an `api` dependency on `engine-time`, so we don't *need* to add it directly, but it's not harmful.

buildSrc/src/main/groovy/Classpaths.groovy
rcaudy: `inheritCommonsText` still has a default.

ModelFarm/ModelFarm.gradle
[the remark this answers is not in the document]
devinrsmith: It's (usually) best-practice to explicitly declare everything you are re-exposing, even if you would otherwise inherit it IMO. It's more important with external dependencies (b/c a version bump may change what they expose).

gradle/web-gwt-test.gradle
devinrsmith: I removed this line from James' advice. I think it was used for manually inspection when things went wrong. If things go wrong gwtTest, we can look into `t.testLogging { ... }` to re-enable (I wasn't able to figure it out).

buildSrc/src/main/groovy/Classpaths.groovy
rcaudy: And `inhertJavaxAnnotatiions`.

rcaudy (approved):

web/client-api/client-api.gradle
niloc132: This is wrong, but only in theory - i'll continue to clean this mess up on my side.

niloc132 (approved):

PR #2681, opened 2022-07-25
removed j_object references from listener notebook

jakemulf:


jjbrosnan (commented):
I'm all for removing `j_object` from any notebook, but using Pandas to perform the same operation seems like an inefficient way to do it.  I'm going to check with the dev team to see if there's any better way to do this using the Python API

PR #2683, opened 2022-07-26
Use better Dockerfile layer ordering to maximize cacheability

devinrsmith:


py/client/build.gradle
```diff
@@ -95,11 +95,10 @@ tasks.getByName('check').dependsOn(Docker.registerDockerTask(project, 'testPyCli
         runCommand '''set -eux; \\
                       pip3 install unittest-xml-reporting==3.0.4 pyarrow==7.0.0 protobuf==3.20.1 grpcio==1.46.0 bitstring==3.1.9 pandas==1.2.5 timeout-decorator==0.5.0;\\
                       mkdir -p /out/report'''
-        environmentVariable 'DH_HOST', deephavenDocker.containerName.get()
-        environmentVariable 'DH_PORT', '8080'
-
         copyFile('project', '/project')
         workingDir('/project')
+        environmentVariable 'DH_HOST', deephavenDocker.containerName.get()
+        environmentVariable 'DH_PORT', '8080'
```
niloc132, line 101: wouldnt it be better to set these first, to avoid adding two more layers each time any project contents change?

devinrsmith: `containerName` is randomized, so each time it changes it busts all the layers after it.

niloc132: Ah fair point. The port is still fair game, but approved.

niloc132 (approved):

PR #2687, opened 2022-07-26
Update Figure Generator to allow nullable parameters

mofojed:
- Allow parameters to be specified as nullable
- For nullable parameters, don't check that they're not null when verifying
- Tested with my treemap-cleanup branch (not yet posted for review)


jmao-denver (approved):
Lgtm

PR #2689, opened 2022-07-27
Use autolocking ctx manager instead of decorator

### LINKED ISSUE #2639: The '@auto_locking_op' annotation breaks autocomplete
The `@auto_locking_op` annotation interferes with InetlliJ's ability to understand and autocomplete Python code.

You can see from this screenshot that the arguments for many methods are missing (and when writing code using these methods, you're unable to see the parameters:

image

image

The results of these methods are not identified as Tables, either, so you lose autocomplete on subsequent method calls:

image

Rebuilding a `deephaven-core` wheel with the annotations removed from `table.py` fixes the issue:

image

jmao-denver:
Fixes #2639

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -18,7 +18,7 @@ from deephaven.agg import Aggregation
 from deephaven.column import Column, ColumnType
 from deephaven.filters import Filter
 from deephaven.jcompat import j_array_list, to_sequence, j_unary_operator, j_binary_operator
-from deephaven.ugp import auto_locking_op
+from deephaven.ugp import auto_locking_ctx
 from deephaven.updateby import UpdateByOperation
 
 # Table
@@ -166,7 +166,6 @@ class Table(JObjectWrapper):
         """Returns a coalesced child table."""
         return Table(j_table=self.j_table.coalesce())
 
-    @auto_locking_op
     def snapshot(self, source_table: Table, do_init: bool = False, cols: Union[str, List[str]] = None) -> Table:
         """Produces an in-memory copy of a source table that refreshes when this table changes.
 
@@ -186,11 +185,11 @@ class Table(JObjectWrapper):
         """
         try:
             cols = to_sequence(cols)
-            return Table(j_table=self.j_table.snapshot(source_table.j_table, do_init, *cols))
+            with auto_locking_ctx(self, source_table):
+                return Table(j_table=self.j_table.snapshot(source_table.j_table, do_init, *cols))
         except Exception as e:
             raise DHError(message="failed to create a snapshot table.") from e
 
-    @auto_locking_op
     def snapshot_history(self, source_table: Table) -> Table:
         """Produces an in-memory history of a source table that adds a new snapshot when this table (trigger table)
         changes.
@@ -213,7 +212,8 @@ class Table(JObjectWrapper):
             DHError
         """
         try:
-            return Table(j_table=self.j_table.snapshotHistory(source_table.j_table))
+            with auto_locking_ctx(self, source_table):
+                return Table(j_table=self.j_table.snapshotHistory(source_table.j_table))
         except Exception as e:
             raise DHError(message="failed to create a snapshot history table.") from e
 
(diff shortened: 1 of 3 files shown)


py/server/deephaven/table.py
```diff
@@ -2339,7 +2341,6 @@ def group_by(self, by: Union[str, Sequence[str]] = None) -> PartitionedTableProx
         except Exception as e:
             raise DHError(e, "group-by operation on the PartitionedTableProxy failed.") from e
 
-    @auto_locking_op
```
rcaudy, line 2342: This doesn't have a matching context.

py/server/deephaven/table.py
```diff
@@ -2311,11 +2313,11 @@ def raj(self, table: Union[Table, PartitionedTableProxy], on: Union[str, Sequenc
             else:
                 raise ValueError("invalid match_rule value")
 
-            return PartitionedTableProxy(j_pt_proxy=self.j_pt_proxy.raj(table_op, on, joins, match_rule))
+            with auto_locking_ctx(self, table):
+                return PartitionedTableProxy(j_pt_proxy=self.j_pt_proxy.raj(table_op, on, joins, match_rule))
         except Exception as e:
             raise DHError(e, "reverse as-of join operation on the PartitionedTableProxy failed.") from e
 
-    @auto_locking_op
```
rcaudy, line 2318: This doesn't have a matching context.

py/server/deephaven/ugp.py
```diff
@@ -122,3 +122,18 @@ def do_locked(*args, **kwargs):
             return f(*args, **kwargs)
 
     return do_locked
+
+
+@contextlib.contextmanager
+def auto_locking_ctx(*args, **kwargs):
+    """A auto-locking aware context manager. It ensures that the enclosed code block runs under
+    the UGP shared lock when ugp.auto_locking is True, the target table-like object is refreshing, and the current
+    thread doesn't own any UGP locks."""
```
rcaudy, line 131: I think this is missing some information. Same applies to the decorator above, except for the first sentence.
Maybe this is a better wording:
An auto-locking aware context manager. Ensures that the enclosed code block runs under the UGP shared lock if ugp.auto_locking is True, the target table-like object or any table-like arguments are refreshing, and the current thread doesn't own any UGP locks.

rcaudy (changes requested):

py/server/deephaven/ugp.py
```diff
@@ -106,9 +106,9 @@ def _has_refreshing_tables(*args, **kwargs):
 
 
 def auto_locking_op(f: Callable) -> Callable:
-    """A decorator for annotating unsafe Table operations. It ensures that the decorated Table operation runs under
-    the UGP shared lock when ugp.auto_locking is True, the target table is refreshing, and the current thread doesn't
-    own any UGP locks."""
+    """A decorator for annotating unsafe Table operations. It ensures that the enclosed code block runs under the UGP
+    shared lock if  ugp.auto_locking is True, the target table-like object or any table-like arguments are
+    refreshing, and the current thread doesn't own any UGP locks."""
```
rcaudy, line 111:
```suggestion
    """A decorator for annotating unsafe Table operations. It ensures that the decorated function runs under the UGP
    shared lock if ugp.auto_locking is True, the target table-like object or any table-like arguments are refreshing,
    and the current thread doesn't own any UGP locks."""
```

py/server/deephaven/ugp.py
```diff
@@ -126,9 +126,9 @@ def do_locked(*args, **kwargs):
 
 @contextlib.contextmanager
 def auto_locking_ctx(*args, **kwargs):
-    """A auto-locking aware context manager. It ensures that the enclosed code block runs under
-    the UGP shared lock when ugp.auto_locking is True, the target table-like object is refreshing, and the current
-    thread doesn't own any UGP locks."""
+    """An auto-locking aware context manager. It ensures that the enclosed code block runs under the UGP shared lock
+    if  ugp.auto_locking is True, the target table-like object or any table-like arguments are refreshing,
+    and the current thread doesn't own any UGP locks."""
```
rcaudy, line 131:
```suggestion
    """An auto-locking aware context manager. It ensures that the enclosed code block runs under the UGP shared lock if
    ugp.auto_locking is True, the target table-like object or any table-like arguments are refreshing, and the current
    thread doesn't own any UGP locks."""
```

rcaudy (approved):

rcaudy (approved):

PR #2697, opened 2022-07-28
Add `go mod verify` to go test

devinrsmith:
As suggested via https://github.com/docker-library/docs/tree/master/golang#start-a-go-instance-in-your-app


SuperTails (approved):
Don't know anything about Gradle, but assuming that `runCommand` checks the return code of each of the commands I don't see any problem here.

rcaudy (approved):

Issue #711, opened 2021-06-04
Java API Flush Remaining Operations for QST


nbauernfeind:
After we have buy-in on issue #710, then we need to go back and add hooks for all of the operations that are part of the TableOperation API. (i.e. avoid time waste if approach isn't ideal.)

Epic: Java API #257

PR #2699, opened 2022-07-29
LazyUpdateTable

devinrsmith:
Partial #711

### DIFF
--- a/java-client/session-dagger/src/test/java/io/deephaven/client/UpdateOrSelectSessionTest.java
+++ b/java-client/session-dagger/src/test/java/io/deephaven/client/UpdateOrSelectSessionTest.java
@@ -51,8 +51,13 @@ public class UpdateOrSelectSessionTest extends DeephavenSessionTestBase {
             public TableSpec apply(TableSpec spec, String[] formulas) {
                 return spec.select(formulas);
             }
+        },
+        LAZY_UPDATE {
+            @Override
+            public TableSpec apply(TableSpec spec, String[] formulas) {
+                return spec.lazyUpdate(formulas);
+            }
         }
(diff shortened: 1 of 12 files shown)


java-client/session-dagger/src/test/java/io/deephaven/client/UpdateOrSelectSessionTest.java
```diff
@@ -51,8 +51,13 @@ public TableSpec apply(TableSpec spec, String[] formulas) {
             public TableSpec apply(TableSpec spec, String[] formulas) {
                 return spec.select(formulas);
             }
+        },
+        LAZY_UPDATE {
```
rcaudy, line 55: This creates a conflict with your other PR, but you probably know that.

table-api/src/main/java/io/deephaven/api/TableOperations.java
```diff
@@ -166,6 +166,12 @@
 
     // -------------------------------------------------------------------------------------------
 
+    TOPS lazyUpdate(String... columns);
```
rcaudy, line 169: This reminds me that JavaDoc should really make its way up to `TableOperations`, but not necessarily in this PR. I think `table-api` will never become a tool for new users without that.

rcaudy (approved):

java-client/session-dagger/src/test/java/io/deephaven/client/UpdateOrSelectSessionTest.java
devinrsmith: Yep. I'm hoping to get #2635 merged in first, will fix conflicts here.

PR #2706, opened 2022-07-30
Fixed a race condition in one of the examples

SuperTails:
In the input table example code, I added some rows to an input table and then immediately snapshotted a table derived from it. Sometimes the server doesn't have time to propogate the new rows, though, so the snapshot returns the wrong data.

This fix just snapshots in a loop until we find the correct data. I don't use an explicit timeout here to keep the example simple. `go test` has a default timeout of 1 minute so we don't have to worry about CI hanging.

### DIFF
--- a/go/pkg/client/example_input_table_test.go
+++ b/go/pkg/client/example_input_table_test.go
@@ -81,17 +81,28 @@ func Example_inputTable() {
 		return
 	}
 
-	// Now, we take a snapshot of the outputTable to see what data it currently contains.
-	// We should see the new rows we added, filtered by the condition we specified when creating outputTable.
-	outputRec, err := outputTable.Snapshot(ctx)
-	if err != nil {
-		fmt.Println("error when snapshotting table", err.Error())
-		return
-	}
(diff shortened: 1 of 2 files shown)


go/pkg/client/example_input_table_test.go
```diff
@@ -81,17 +81,28 @@ func Example_inputTable() {
...
 
-	// Now, we take a snapshot of the outputTable to see what data it currently contains.
-	// We should see the new rows we added, filtered by the condition we specified when creating outputTable.
-	outputRec, err := outputTable.Snapshot(ctx)
-	if err != nil {
-		fmt.Println("error when snapshotting table", err.Error())
-		return
-	}
-	defer outputRec.Release()
+	for {
```
devinrsmith, line 84: I'm worried about an infinite loop. At the very least, some sort of limiter. @rcaudy, I'm surprised that upon completion of `AddTable` that the data might not be available for snapshotting?

devinrsmith (approved):

PR #2716, opened 2022-08-04
Remove explicit field syncing from Python client

SuperTails:
Listing the available tables on the Python client now just performs a request to the server instead of trying to manage an entire cache of tables.
The `open_table` function now just constructs the ticket based on the table name and tries to open it directly.
This means that all of the field-fetching and syncing stuff is gone.


py/client/pydeephaven/session.py
```diff
@@ -317,10 +235,19 @@ def open_table(self, name: str) -> Table:
             DHError
         """
         with self._r_lock:
-            if name not in self.tables:
-                raise DHError(f"no table by the name {name}")
-            table_op = FetchTableOp()
-            return self.table_service.grpc_table_op(self._fields[('scope', name)][1], table_op)
+            ticket = ticket_pb2.Ticket(ticket=f's/{name}'.encode(encoding='ascii'))
+
+            faketable = Table(session=self, ticket=ticket)
```
devinrsmith, line 240: So - I wonder if we want to actually add abstractions, or safety barriers, such that returning `faketable` here makes sense. (For example, we need to make sure that we don't try to `close()` w/ a `s/...` ticket.) 

Or may introduce `ref_table(...)` and `fetch()`, such that `open_table(...) == ref_table(...).fetch()`? (I think naming could be better.)

Of particular interest is the interest in obtaining `ref_table(...).snapshot()` without needing a FetchTableOp against the server.

devinrsmith: Or maybe, still have only `open_table`, but only execute the `FetchTableOp` when the user calls into a method that the data that provides?

jmao-denver: I think having open_table() still makes more sense if we consider the typical use case of a client wants to get a reference to a table in the server for the only reason to perform some operations on it. So lazy fetching isn't too meaningful here.

SuperTails: I also think it would be more intuitive that the "this table doesn't exist" error shows up when you try to open the table, rather than the first time you try to use it. If we wanted lazy fetching, I think a separate `open_lazy_table` method would make more sense (and that can go into a separate PR if we thought it was useful).

devinrsmith (commented):
Ok, all fine by me. Will leave up to @jmao-denver for final approval.

jmao-denver (approved):
LGTM

PR #2725, opened 2022-08-09
Fail incoming grpc connections that don't use http2

### LINKED ISSUE #2724: Ensure jetty server does not proceed with plain HTTP 1.1
It's possible that proxies will try and communicate with the jetty server using HTTP 1.1 without upgrading to websockets or using h2c. This is an unsupported configuration, and we should be proactive in erroring out as soon as we determine this is the case.

Note: at least one proxy, https://doc.traefik.io/traefik/user-guides/grpc/, suggests that specifying h2c isn't necessary when the backend is using TLS - so maybe it's a non-issue in general w/ TLS?

niloc132:
This is overzealous in general, as grpc-web can technically operate on
http/*, but we aren't interested in supporting that use case, since
streaming data is important to us.

Fixes #2724

### DIFF
--- a/server/jetty/src/main/java/io/deephaven/server/jetty/GrpcFilter.java
+++ b/server/jetty/src/main/java/io/deephaven/server/jetty/GrpcFilter.java
@@ -3,19 +3,31 @@
  */
 package io.deephaven.server.jetty;
 
+import io.deephaven.configuration.Configuration;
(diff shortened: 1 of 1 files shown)


devinrsmith (approved):
I've been able to test this out with my setup in the browser, as well as from the java client. Both provide the helpful error message when the proxy is not properly configured w/ h2c.

PR #2727, opened 2022-08-10
Remove Jetty SNI check

devinrsmith:


server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
```diff
@@ -139,7 +139,14 @@ private static ServerConnector createConnector(Server server, ServerConfig confi
...
         if (config.ssl().isPresent()) {
-            httpConfig.addCustomizer(new SecureRequestCustomizer());
+            // Note: explicitly setting SNI host check to false here to better match the configuration that gRPC Netty
+            // has out-of-the-box. gRPC Netty does not even have an easy way to enable SNI host checking,
+            // https://github.com/grpc/grpc-java/issues/7397. Essentially, the client retains responsibility for
+            // hostname verification based on the hostname it's connecting to, the certificate presented, and the
+            // configured trust stores. An SNI-less Jetty server setup potentially affords proxied deployments a simpler
+            // configuration model (essentially, the server does not need to know it is being proxied).
+            final boolean sniHostCheck = false;
+            httpConfig.addCustomizer(new SecureRequestCustomizer(sniHostCheck));
```
niloc132, line 149: I believe this will conflict with #2548, which explicitly set it the opposite way.

I would still prefer that we fail "safe" instead of "unsafe" as this is doing, though would be happy to see it fail during the SNI handshake instead of during the HTTP conv as it is doing. If we want to support "unsafe" like this, my opinion is that it should default to safe and require explicit configuration to make it unsafe.

devinrsmith: It's probably worthwhile to follow the same pattern here #2731 for configuring SNI - I'll wait until that has merged until pursuing this further.

devinrsmith:
Superseded by #3843

PR #2731, opened 2022-08-11
Add explicit configuration for websockets, http1, and a proxy hint.

devinrsmith:


devinrsmith:
Related to https://github.com/deephaven/deephaven-core/pull/2725

server/jetty/src/main/java/io/deephaven/server/jetty/JettyConfig.java
```diff
@@ -68,15 +77,60 @@ public int port() {
...
+        if (Boolean.TRUE.equals(proxyHint())) {
+            return false;
+        }
+        return ssl().isEmpty();
+    }
+
+    @Value.Check
+    final void checkWebsocketsNeedsHttp1() {
+        if (websocketsOrDefault() && !http1OrDefault()) {
+            throw new IllegalArgumentException("Jetty configuration includes websockets but not HTTP1.1");
```
niloc132, line 126: It is my understanding that there is no reason that websockets _can't_ work on h2 - for example (as we have discussed elsewhere) if the server is behind multiple proxies, and one doesn't support h2, we would want to encourage websockets.

Example:
```mermaid
graph TD
    C[deephaven-server-jetty-app]
    B[Proxy #1]
    A[Proxy #2]
    D[plain h2 grpc client]
    E["public internet"]
    A-->|http/1.1|B
    D-->|h2|C
    B-->|h2|C
    E-->A
```

server/jetty/src/main/java/io/deephaven/server/jetty/JettyConfig.java
```diff
@@ -68,15 +77,60 @@ public int port() {
...
+     */
+    public final boolean http1OrDefault() {
+        final Boolean http1 = http1();
+        if (http1 != null) {
+            return http1;
+        }
+        if (Boolean.TRUE.equals(proxyHint())) {
+            return false;
+        }
+        return ssl().isEmpty();
```
niloc132, line 120: if ssl is enabled without a proxy, have you confirmed that a browser can connect to it without needing to first try http/1.1? It has been a while since I tested that, and recall some hiccup there.

server/jetty/src/main/java/io/deephaven/server/jetty/JettyConfig.java
devinrsmith: Ah yes - I think this fact was established after I created this PR. I'll remove this improper check.

server/jetty/src/main/java/io/deephaven/server/jetty/JettyConfig.java
devinrsmith: At least with my version of Firefox on Linux, yes. I can't speak for Chrome, older versions, and other operating systems at this point in time. It would surprise me a bit if anything relatively modern couldn't handle it - but might be worth some further testing?

niloc132: I'd vote do the testing before turning it off, yes - bad form to ship it and then find out we broke it. 

Alternatively, just be less aggressive about disabling http/1.1, and merge the rest.

niloc132 (commented):
(i need a comment here to kick it back to you, either test in more cases to ensure we arent breaking browsers, or drop the "ssl implies no http/1.1")

server/jetty/src/main/java/io/deephaven/server/jetty/JettyConfig.java
devinrsmith: I've set it to be more conservative, and filed a follow-up issue.

niloc132 (approved):

PR #2733, opened 2022-08-12
Update DeepHavenJsApiLinker to allow exporting GWT JsTypes as ES6 module

vegegoku:


vegegoku:

web/client-api/src/main/java/io/deephaven/web/DeephavenJsApiLinkerTemplate.js
```diff
@@ -0,0 +1,31 @@
...
+
+var dh = {}
+$wnd.dh = dh;
+
+var $gwt_version = "__GWT_VERSION__";
+
+__JAVASCRIPT_RESULT__
+
+gwtOnLoad(null, '__MODULE_NAME__', null);
+export {dh};
```
mofojed, line 31: Would prefer to export as `default` here:
```suggestion
export default dh;
```

web/client-api/src/main/java/io/deephaven/web/DeephavenJsApiLinker.java
```diff
@@ -54,23 +54,41 @@ public ArtifactSet link(TreeLogger logger, LinkerContext context, ArtifactSet ar
...
+
+    protected String getSelectionScriptTemplate() {
+        return "io/deephaven/web/DeephavenJsApiLinkerTemplate.js";
+    }
+
+    protected static void replaceAll(StringBuffer buf, String search,
+            String replace) {
+        int len = search.length();
+        for (int pos = buf.indexOf(search); pos >= 0; pos = buf.indexOf(search,
+                pos + 1)) {
```
mofojed, line 90: I think this should be `pos + replace.length()` for correctness... if for example you are passing in a buf with `HELLOHELLO`, search `HELLO` and replace with empty string ``, you will only get rid of the first `HELLO`
By doing `+ replace.length()` you're handling that case and also skipping some chars that can't match.

web/client-api/src/main/java/io/deephaven/web/public/linked/child.html
```diff
@@ -26,7 +25,10 @@ <h3>Outgoing link summary</h3>
 
 </ul>
 
-<script>
+<script type="module">
+    import {dh} from './dh-core.js';
```
mofojed, line 29: I think this needs `../`
```suggestion
    import {dh} from '../dh-core.js';
```

web/client-api/src/main/java/io/deephaven/web/public/rollup.html
```diff
@@ -72,7 +71,8 @@ <h3>Rollup Table Configuration</h3>
 <input id="includeConstituents" type="checkbox"><label for="includeConstituents">Include Constituent values</label><br/>
 <button id="open">Display Rollup Table</button>
 <div id="treetable"></div>
-<script>
+<script type="module">
+    import {dh} from './dh-core.js'
```
mofojed, line 75: Whitespace issue
```suggestion
  import {dh} from './dh-core.js'
```

web/client-api/src/main/java/io/deephaven/web/public/linked/child.html
mofojed: Also fix the whitespace to be consistent

web/client-api/src/main/java/io/deephaven/web/public/freeze.html
```diff
@@ -40,16 +39,17 @@ <h3>Selected table data</h3>
 <button id="next_frozen">&gt;</button>
 <button id="end_frozen">&gt;&gt;</button>
 
-<script>
-
+<script type="module">
+    import {dh} from './dh-core.js';
```
mofojed, line 43: Ideally we'd be able to destructure at import time, i.e. `import { Client } from './dh-core.js';`, but I think as @niloc132 explained that would be difficult/impossible to do with GWT, so this should be fine - just doing `import dh from './dh-core.js';`

web/client-api/src/main/java/io/deephaven/web/public/linked/index.html
```diff
@@ -68,16 +67,18 @@ <h3>Persistent Query Configurations</h3>
 <ul id="queries">
 
 </ul>
-<script>
+<script type="module">
+    import {dh} from './dh-core.js';
+    const {Client} = dh;
```
mofojed, line 72: Whitespace
```suggestion
  import {dh} from './dh-core.js';
  const {Client} = dh;
```

mofojed (changes requested):
Going to create tickets for some additional things, such as publishing these as a package in npmjs (e.g. `@deephaven/jsapi` and/or `@deephaven/jsapi-internal`).

Whitespacing is inconsistent in many of the new lines in these examples.

mofojed:
Spoke with @niloc132 about this, and the change looks good as a halfway point, _however_ it makes consuming the library more difficult. Ultimately suggest we keep this as a fork or a separate branch until we have the full thing exported as a module (dh-internal as well) and published to npmjs.

The reason why this is difficult is because the Web UI still needs to load it remotely from the server. Currently, it's just doing a regular `script` tag with type `text/javascript`, which add `dh` to the global namespace: https://github.com/deephaven/web-client-ui/blob/1d10956499cd451145e704690bc8f2d5cfb6a786/packages/code-studio/public/index.html#L41
There is then the `jsapi-shim` which pulls `dh` from the global namespace, which we use from our modules to be able to import the jsapi: https://github.com/deephaven/web-client-ui/blob/1d10956499cd451145e704690bc8f2d5cfb6a786/packages/jsapi-shim/src/dh.ts#L11

That shim only works if `dh` is already on the global namespace. However, when importing a script using `type="module"`, it defers loading until after the page is already loaded. So we'd need to defer the web UIs startup entirely to wait for the dh module to import before we load the rest of our code. It's simply a pain in the butt, and doesn't really have any benefits yet at this point. We also can't just do `import dh from './dh-core.js'`, since it's loading from a remote address (e.g. we need to do `<script type="module" src="http://localhost:10000/jsapi/dh-core.js">`). As a consumer of the library, this halfway point is actually more difficult to use/consume (even though it is a step in the right direction).

Once dh-internal is also a module, and they're both published to npmjs, theoretically should be able to change jsapi-shim to just:
```
import dh from '@deephaven/jsapi';
export default dh;
```
And everything should work. (We could also just change all our imports from `import dh from '@deephaven/jsapi-shim'` to `import dh from '@deephaven/jsapi'`, but by keeping the shim we can actually have the same code work on both Enterprise and Community until we get Enterprise to export as a module as well).

So, I approve these changes, but please don't merge them to main, and keep a separate long running branch for this development :).

mofojed (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2180
Conceptual: https://github.com/deephaven/deephaven.io/issues/2181
Reference: https://github.com/deephaven/deephaven.io/issues/2179
Blog: https://github.com/deephaven/deephaven.io/issues/2178

PR #2736, opened 2022-08-16
Make jpy related jars java 8 compatible.

devinrsmith:
Removes jpy-ext dependency on Base.


JamesXNelson (approved):
looks great! exactly what we need for enterprise to be able to consume this library. tyvm.

rcaudy (approved):

PR #2739, opened 2022-08-17
Update py/embedded-server setup for PyPi

jjbrosnan:
This PR updates the setup.py file to use a new README for PyPi, which contains the basic installation instructions above the dev environment instructions.

### DIFF
--- a/py/embedded-server/README_PyPi.md
+++ b/py/embedded-server/README_PyPi.md
@@ -1,14 +1,16 @@
 # deephaven-server
 
-The Dockerless version of [deephaven-core](https://www.github.com/deephaven/deephaven-core).
+A Python-native way of launching [deephaven-core](https://www.github.com/deephaven/deephaven-core).
 
 Deephaven Community Core is a real-time, time-series, column-oriented analytics engine with relational database features. It is an open version of [Deephaven Enterprise](https://deephaven.io/enterprise/), which serves as the data backbone for prominent hedge funds, banks, and financial exchanges.
 
-`deephaven-server` enables users to run Deephaven Community Core without Docker.
+## Requirements
+
+Java 11+ is required for this module, and the `JAVA_HOME` environment variable must be set appropriately.
+
+This module also requires Python version 3.7 or newer.
 
 ## Setup
-
-Java 11+ is required for this module, and the `JAVA_HOME` environment variable must be set appropriately.  Once that's done, you can use pip:
 
 ```shell
 pip3 install --upgrade pip setuptools wheel
--- a/py/embedded-server/README_PyPi.md
+++ b/py/embedded-server/README_PyPi.md
@@ -1,0 +1,41 @@
+# Deephaven Python Embedded Server
+
+Embeds the Deephaven Core server into a python module, so that it can be started from python, with tables produced
+directly as part of startup.
+
+## Setup
+
+Java 11+ is required for this module, and the `JAVA_HOME` environment variable must be set appropriately.  Once that's done, you can use pip:
+
+```shell
+pip3 install --upgrade pip setuptools wheel
+pip3 install deephaven-server
+```
+
+## Dev environment setup
+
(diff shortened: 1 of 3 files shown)


py/embedded-server/README_PyPi.md
```diff
@@ -0,0 +1,41 @@
+# Deephaven Python Embedded Server
```
devinrsmith, line 1: I would have the header be `deephaven-server` to match the PyPi package name.

py/embedded-server/README_PyPi.md
```diff
@@ -0,0 +1,41 @@
...
+### Build
+From the root directory of this repository
+```shell
+$ ./gradlew :py-embedded-server:assemble
+```
+
+### Install
+```shell
+$ pip install py/embedded-server/build/wheel/deephaven_server-0.12.0-py3-none-any.whl
+```
```
devinrsmith, line 30: IMO, we should ax all of this - these are development notes and can go elsewhere, but they don't belong on PyPi.

py/embedded-server/README_PyPi.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Quick start
+
```
devinrsmith, line 33: We should have a note that Java 11+ is required, and `JAVA_HOME` should be set.

devinrsmith (changes requested):

py/embedded-server/README_PyPi.md
```diff
@@ -0,0 +1,41 @@
+# Deephaven Python Embedded Server
+
+Embeds the Deephaven Core server into a python module, so that it can be started from python, with tables produced
+directly as part of startup.
```
devinrsmith, line 4: I think this should be a more python-end-user introduction - we shouldn't be focused on the "embeds" part of it. A link to deephaven-core is fine.

py/embedded-server/README_PyPi.md
jjbrosnan: Is the first line in the setup section not sufficient?

margaretkennedy (approved):

py/embedded-server/README_PyPi.md
margaretkennedy: The line seems enough to me.

py/embedded-server/README_PyPi.md
```diff
@@ -0,0 +1,27 @@
+# deephaven-server
+
+The Dockerless version of [deephaven-core](https://www.github.com/deephaven/deephaven-core).
+
+Deephaven Community Core is a real-time, time-series, column-oriented analytics engine with relational database features. It is an open version of [Deephaven Enterprise](https://deephaven.io/enterprise/), which serves as the data backbone for prominent hedge funds, banks, and financial exchanges.
+
+`deephaven-server` enables users to run Deephaven Community Core without Docker.
```
devinrsmith, line 7: I'm not a fan of explicitly calling our main product a "docker" product; and thus, `deephaven-server` is "Deephaven minus docker". I think highlighting that this is a python-native way of starting Deephaven is what we should strive for conveying.

py/embedded-server/README_PyPi.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Setup
+
+Java 11+ is required for this module, and the `JAVA_HOME` environment variable must be set appropriately.  Once that's done, you can use pip:
```
devinrsmith, line 11: I don't like attaching this as a *setup* requirement - it's not necessary for pip installs. That said, it is important info, and having it set during pip install won't hurt.

I'd suggest creating a "## Requirements" header before Setup; where we can also mention that it requires Python 3.7+.

devinrsmith (changes requested):

devinrsmith (approved):

devinrsmith (approved):

Issue #2734, opened 2022-08-15
Logs in DHaaL are confusing


niloc132:
Split off from https://github.com/deephaven/deephaven-core/issues/2453

From the original issue, linked above:
>  Revisit how logs are handled, redirected, displayed, to be sure that if the web console is used at all, it sees what it would expect to see.

Roughly, there are two kinds of logs in Deephaven - stdout/stderr which will be written mostly by script code or handle some basic app logging, and Deephaven logs, which is configurable at startup or at runtime but almost exclusively runs from inside the server or engine. Stdout/err should almost always write to the python console, whether it is interactive or not, and both should write to the web console. Additionally, Deephaven logs should generally be written to stderr (not stdout) so a long-running batch script or server can see its output.

Likely will be fixed along with https://github.com/deephaven/deephaven-core/issues/2723.

niloc132:
Current flow of logs in DHaaL:
 * python starts, creates its own sys.stdout/err objects, which point to the process's own stdout/stderr
 * jvm starts, creates its own System.out/err objects, which point to the process's own stdout/stderr
 * Deephaven logging rewires sys.stdout/err objects with ones that _only_ forward to java's System.out/err, which has no net effect (but enables other steps)

Simple change to enable stdout/err to make it to the web UI:
 * Deephaven java code should replace System.out/err with implementations that "tee" output to DH logs, so they are buffered to the UI

Simple change to enable DH logs to write to stderr (preferred over stdout in case there is a process watching output)
 * Deephaven logging that replaces System.out/err can test if python's _old_ sys.stderr is a tty, if so skip. Note that this introduces a dependency on no other changes be made along the way. We cannot test accurately if Java's System.err is going to the tty or not, so we have to lean on python for this.
 * Either write logs directly to python's _old_ sys.stderr, or to Java's _old_ System.err. Writing to Java's old System.err seems preferable in terms of staying in Java and avoiding the GIL, but we do want to take care to defer to python's expectations in this case.

Potential changes to avoid future surprises:
 * If python code replaces stdout/stderr before DH Java starts, it might surprise the user that their redirection no longer works. Instead, consider also changing our current implementation to replace Java's System.out/err with redirections to whatever the current Python sys.stdout/stderr values are. This must take place _before_ Deephaven replaces Java System.out/err with a "tee" output to DH logs. Example: jupyter.

Fix for #2723 
[...]

PR #2743, opened 2022-08-18
Change Bootstrap.log to System.out instead of System.err

mofojed:
- There are just info messages, they are not errors so they should not be output to System.err

See screenshot below for example of how it currently looks:
image

### DIFF
--- a/log-factory/src/main/java/io/deephaven/internal/log/Bootstrap.java
+++ b/log-factory/src/main/java/io/deephaven/internal/log/Bootstrap.java
@@ -11,7 +11,7 @@ class Bootstrap {
 
     public static void log(Class<?> source, String message) {
         if (isEnabled()) {
-            System.err.printf("# %s: %s%n", source.getName(), message);
+            System.out.printf("# %s: %s%n", source.getName(), message);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


devinrsmith (approved):
I'm probably OK with this change... but in general, just because logging goes to stderr does *not* mean it's an error.  A lot of times the convention is that diagnostic information goes to stderr as well. I might claim that the bootstrapping system properties should more appropriately go to stderr instead.

I could also claim "it should be configurable", but I'd prefer to just make the switch to what we think the best default is, and add in configuration later if the need arises.

@niloc132 has been deep into Python logs (https://github.com/deephaven/deephaven-core/issues/2734) - assuming he'll feel the same way I do.

mofojed:
Where's the convention that diagnostic information outputs to stderr? I've only seen that in debug environments. Having errors output visible to the user makes it look like errors are occurring, or become blind to other errors because they've been trained to ignore them...

devinrsmith:
It's a fundamental POSIX convention: https://pubs.opengroup.org/onlinepubs/9699919799/functions/stdin.html

> standard error (for writing diagnostic output)

command line utilities are essentially plumbed together and enabled by this convention.

That said, DH is typically run as an application, and we concern ourselves with logging framework lifecycles; and we could make the case that all of our diagnostic or bootstrap logging should go to stdout instead.

mofojed:
> standard error (for writing diagnostic output)
Wouldn't the definition of "diagnostic output" mean there is something to diagnose? E.g. an error has occurred, so we log information to help diagnose the error.

In any case - anything logged to stderr shows up as red/error lines in jupyter (and our own IDE) console output, which as a user, looks like an error has occurred.

PR #2748, opened 2022-08-19
Improve `SortedRanges` iterator performance through caching

lbooker42:
While examining profiler data for `partitionBy`/`groupBy` found that a lot of time was spent in `nextRange()`, specifically the call to `sar.unpackedGet()`.

The current code computes the call `sar.unpackedGet(nextRangeIdx)` then discards the results if the range is a single value.  Rowsets that are sparse (consist of few consecutive values) will benefit by storing this value for the next call to `nextRange()` and this change will not penalize the non-sparse rowsets because the lookup must be performed anyway.

Performance metrics (average of 5 consecutive runs):
```
Rows            Buckets     partitionBy (ms)    partitionBy_opt (ms)    speedup
100,000,000     100         1,384               1,382                   0.1%
100,000,000     25,000      124,946             93,637                  25.1%
100,000,000     1,000,000   30,775              30,373                  1.3%
                
Rows            Buckets     groupBy (ms)        groupBy_opt (ms)        speedup
100,000,000     100         1,445               1,406                   2.7%
100,000,000     25,000      124,256             101,950                 18.0%
100,000,000     1,000,000   30,402              28,202                  7.2%
                
Rows            Buckets     partitionBy (ms)    partitionBy_opt (ms)    speedup
100,000,000     100         1,855               1,492                   19.6%
100,000,000     25,000      110,660             87,067                  21.3%
100,000,000     1,000,000   30,857              30,997                  -0.5%
                 
Rows            Buckets     groupBy (ms)        groupBy_opt (ms)        speedup
100,000,000     100         1,884               1,493                   20.8%
100,000,000     25,000      107,324             85,797                  20.1%
100,000,000     1,000,000   30,524              29,773                  2.5%
```

Note: this only is useful because `unpackedGet()` is more expensive in practice than expected given the simplicity of the  implementation.

### DIFF
--- a/engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/sortedranges/SortedRanges.java
+++ b/engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/sortedranges/SortedRanges.java
@@ -808,7 +808,9 @@ public abstract class SortedRanges extends RefCountedCow<SortedRanges> implement
             if (neg) {
                 currRangeStart = currRangeEnd = -data;
                 nextRangeIdx = i + 1;
-                cachedNextValue = sar.unpackedGet(nextRangeIdx);
+                if (nextRangeIdx < sar.count) {
+                    cachedNextValue = sar.unpackedGet(nextRangeIdx);
+                }
                 return currRangeStart;
             }
             final int next = i + 1;
@@ -825,7 +827,9 @@ public abstract class SortedRanges extends RefCountedCow<SortedRanges> implement
                         (final long v) -> comp.compareTargetTo(v, dir));
                 currRangeEnd = -nextData;
                 nextRangeIdx = next + 1;
-                cachedNextValue = sar.unpackedGet(nextRangeIdx);
+                if (nextRangeIdx < sar.count) {
+                    cachedNextValue = sar.unpackedGet(nextRangeIdx);
+                }
                 return currRangeStart;
             }
             currRangeStart = currRangeEnd = data;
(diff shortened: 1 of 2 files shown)


engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/sortedranges/SortedRanges.java
```diff
@@ -590,14 +601,18 @@ public final boolean advanceImpl(final long v) {
...
+                currRangeEnd = -cachedNextValue;
                 nextRangeIdx = p + 1;
+                if (nextRangeIdx < sar.count) {
+                    cachedNextValue = sar.unpackedGet(nextRangeIdx);
+                }
                 return true;
             }
             currRangeEnd = currRangeStart;
             nextRangeIdx = p;
+            cachedNextValue = sar.unpackedGet(nextRangeIdx);
```
cpwright, line 615: I think this change is good overall.  What I am concerned about, but did not carefully run down is if there are cases like this where we will end up overflowing sar count or if we got all the potential off-by-one errors.  Presumably if the full test suite has passed, so I should get unnervous.

lbooker42:
Added performance data, the maximum improvement is found at 25k buckets, where each sorted range is ~4k (near the limit for conversion to `RspBitmap`)

engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/sortedranges/SortedRanges.java
```diff
@@ -460,6 +460,7 @@ private static final class Iterator implements RowSet.Iterator {
         private int nextRangeIdx = 0;
         private long rangeCurr = -1;
         private long rangeEnd = -1;
+        private long cachedNextValue;
```
jcferretti, line 463: I would call this field `nextValue` instead of `cachedNextValue`, because I think is more in line with the rest of the naming of fields we do in Iterators around index.

cpwright (dismissed):

cpwright (approved):

PR #2752, opened 2022-08-23
Support reading Kafka JSON composite fields as string. Fixes #2751

### LINKED ISSUE #2751: In Kafka JSON reading, support getting composite JSON fields as string.
We support reading individual sub-fields of composite JSON fields today; however, some users have requested reading the composite field as a JSON string, and today attempting to do that just yields an empty string.

jcferretti:


devinrsmith (approved):
Looks great, thanks!

PR #2755, opened 2022-08-23
Support reading composite Kafka JSON fields as a JsonNode. Fixes #2754.

### LINKED ISSUE #2754: Support reading composite Kafka JSON fields as a JsonNode (jackson) object.
Related to #2751.
For efficiency it makes sense to allow composite Json fields to be read as Jackson's JsonNode objects (in a DH column of either Object or that type).
Having done that change, it also would make sense to default any non-string object type to be read as that.

jcferretti:
```
cfs@guanaco 14:58:30 ~/dh/oss1/deephaven-core/redpanda/examples/python
$ python3 kafka-produce.py orders 0 '' 'str:{ "Symbol" : "MSFT", "Side" : "BUY", "Price" : "278.85", "Qty" : "200", "User" : { "Name" : "Pepo", "Id" : 1234 } }'
Message key=|b''|, value=|b'{ "Symbol" : "MSFT", "Side" : "BUY", "Price" : "278.85", "Qty" : "200", "User" : { "Name" : "Pepo", "Id" : 1234 } }'| delivered to topic orders partition 0
(confluent-kafka)
```

Deephaven - Google Chrome_005

### DIFF
--- a/extensions/kafka/src/main/java/io/deephaven/kafka/ingest/JsonNodeChunkAdapter.java
+++ b/extensions/kafka/src/main/java/io/deephaven/kafka/ingest/JsonNodeChunkAdapter.java
@@ -63,9 +63,13 @@ public class JsonNodeChunkAdapter extends MultiFieldChunkAdapter {
             case Double:
                 return new JsonNodeDoubleFieldCopier(fieldName);
             case Object:
-                return (dataType == String.class)
-                        ? new JsonNodeStringFieldCopier(fieldName)
-                        : new JsonNodeObjectFieldCopier(fieldName);
+                if (dataType == String.class) {
+                    return new JsonNodeStringFieldCopier(fieldName);
+                }
+                if (dataType.isAssignableFrom(com.fasterxml.jackson.databind.JsonNode.class)) {
+                    return new JsonNodeJsonNodeFieldCopier(fieldName);
+                }
+                throw new UncheckedDeephavenException("Type " + dataType.getSimpleName() + " not supported for JSON");
         }
         throw new IllegalArgumentException("Can not convert field of type " + dataType);
     }
--- a/extensions/kafka/src/main/java/io/deephaven/kafka/ingest/JsonNodeChunkAdapter.java
+++ b/extensions/kafka/src/main/java/io/deephaven/kafka/ingest/JsonNodeChunkAdapter.java
@@ -63,11 +63,10 @@ public class JsonNodeChunkAdapter extends MultiFieldChunkAdapter {
             case Double:
                 return new JsonNodeDoubleFieldCopier(fieldName);
             case Object:
-                if (dataType == String.class) {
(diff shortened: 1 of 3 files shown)


extensions/kafka/src/main/java/io/deephaven/kafka/ingest/JsonNodeChunkAdapter.java
```diff
@@ -63,11 +63,10 @@ private static FieldCopier makeFieldCopier(
...
                 return new JsonNodeDoubleFieldCopier(fieldName);
             case Object:
-                if (dataType == String.class) {
-                    return new JsonNodeStringFieldCopier(fieldName);
-                } else {
-                    throw new UncheckedDeephavenException("Raw objects not supported for JSON");
-                }
+                return (dataType == String.class)
+                        ? new JsonNodeStringFieldCopier(fieldName)
+                        : new JsonNodeObjectFieldCopier(fieldName)
```
devinrsmith: It probably makes sense to check `dataType == com.fasterxml.jackson.databind.JsonNode`?

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/JsonNodeObjectFieldCopier.java
```diff
@@ -0,0 +1,34 @@
...
+package io.deephaven.kafka.ingest;
+
+import com.fasterxml.jackson.core.JsonPointer;
+import io.deephaven.chunk.ObjectChunk;
+import io.deephaven.chunk.WritableChunk;
+import io.deephaven.chunk.WritableObjectChunk;
+import com.fasterxml.jackson.databind.JsonNode;
+import io.deephaven.chunk.attributes.Values;
+
+public class JsonNodeObjectFieldCopier implements FieldCopier {
```
devinrsmith, line 13: Does it make sense to call this a `JsonNodeNodeFieldCopier` instead of "Object"?

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/JsonNodeChunkAdapter.java
jcferretti: Either that or a base class of that I guess (eg, Object).

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/JsonNodeObjectFieldCopier.java
jcferretti: Done.

devinrsmith (approved):

devinrsmith:
We decided to not bake in `JsonNode` type as primitive - for a user to reference it in python, they can create:

```python
JsonNode = deephaven.dtypes.DType('com.fasterxml.jackson.databind.JsonNode')
```

PR #2756, opened 2022-08-24
Fix ExecutionContext and PartitionTable bugs caught by enabling parallel updates

nbauernfeind:


py/server/tests/test_partitioned_table.py
```diff
@@ -116,6 +117,12 @@ def test_constituents(self):
         self.assertGreater(len(constituent_tables), 0)
 
     def test_transform(self):
+        _JExecutionContext = jpy.get_type("io.deephaven.engine.context.ExecutionContext")
+        context = _JExecutionContext.newBuilder() \
+                .captureCompilerContext()         \
+                .captureQueryLibrary()            \
+                .emptyQueryScope()                \
+                .build().open()
```
jmao-denver, line 125: This puts too much burden on the users. The ExecutionContext should be transparent to the application code. We might need to redesign it or find a way in Python to wrap it.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/CharacterSparseArraySource.java
```diff
@@ -414,38 +414,39 @@ public void ensurePrevious(RowSet changedRows) {
...
+            prevFlusher.maybeActivate();
         }
-        prevFlusher.maybeActivate();
 
-        try (final RowSet.Iterator it = changedRows.iterator()) {
-            long key = it.nextLong();
-            while (true) {
-                final long firstKey = key;
+        try (final RowSequence.Iterator it = changedRows.getRowSequenceIterator()) {
+            while (it.hasMore()) {
```
rcaudy, line 424: This could be a `do-while` loop, since we check `changedRows.isEmpty()` above.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/CharacterSparseArraySource.java
```diff
@@ -414,38 +414,39 @@ public void ensurePrevious(RowSet changedRows) {
             return;
         }
 
-        if (prevFlusher == null) {
-            return;
+        // Note that this method is used to precondition the column source before processing an update in parallel. So,
```
rcaudy, line 417: Maybe we should rename the method. `ensurePreviousAndPreallocateBlocks` or some other such mouthful.

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -1266,7 +1267,10 @@ private Table selectOrUpdate(Flavor flavor, final SelectColumn... selectColumns)
 
                     final QueryTable resultTable;
                     final LivenessScope liveResultCapture = isRefreshing() ? new LivenessScope() : null;
-                    try (final SafeCloseable ignored = liveResultCapture != null ? liveResultCapture::release : null) {
+                    try (final SafeCloseable ignored1 = liveResultCapture != null ? liveResultCapture::release : null;
+                            final SafeCloseable ignored2 = ExecutionContext.getDefaultContext().open()) {
+                        // we open the default context here to ensure that the update processing happens in the default
```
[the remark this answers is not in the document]
nbauernfeind: This is to save the user from himself. If the formula column needs access to the ExecutionContext it should be installed by the user from within his formula. By opening the default poisoned context, we guarantee that the `fakeUpdate` pass through the formula is invoked without any user installed context nor any systemic context.

Note that the context installed at this point in time is not the context that will be installed from a view column (maybe we should install the default poisoned execution context inside of ConstructSnapshot, too). As discussed in chat, it might be appropriate to back-out the change for update graph notifications to specify their execution context; the user is supposed to install the execution context from within their formula.

engine/table/src/main/java/io/deephaven/engine/table/impl/SelectOrUpdateListener.java
```diff
@@ -103,6 +103,7 @@ protected void onError(Exception error) {
 
     private void handleException(Exception e) {
         dependent.notifyListenersOnError(e, getEntry());
+        getParent().removeUpdateListener(this);
```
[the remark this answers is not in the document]
nbauernfeind: One side effect of this approach is that each failing listener pass this same error to the `AsyncClientErrorNotifier`. This simple case prints out four such errors:
```groovy

err = { _ -> if (true) throw new RuntimeException() else 0 }
tt = TableTools.timeTable("00:00:01").update("I = ii", "J = new Object()")
m = TableTools.merge(tt, tt.update("I = ii", "J = I > 5 ? err() : 0"))
```

engine/table/src/main/java/io/deephaven/engine/table/impl/InstrumentedTableListenerBase.java
```diff
@@ -152,9 +152,12 @@ public boolean satisfied(final long step) {
 
     @Override
     public void onFailure(Throwable originalException, Entry sourceEntry) {
+        deregisterOnFailure();
         onFailureInternal(originalException, sourceEntry == null ? entry : sourceEntry);
     }
 
+    protected abstract void deregisterOnFailure();
```
rcaudy, line 159: I think this current implementation is a big improvement over the status quo. I'm trying to think if we could go one better and drop the reference count to zero, thus forcing a destroy. Maybe this whole change set belongs in a separate PR.

engine/table/src/main/java/io/deephaven/engine/table/impl/partitioned/PartitionedTableProxyImpl.java
```diff
@@ -128,20 +128,29 @@ public boolean sanityChecksJoins() {
...
-            context = ExecutionContext.newBuilder()
+            final ExecutionContext.Builder builder = ExecutionContext.newBuilder()
                     .captureCompilerContext()
-                    .markSystemic()
-                    .build();
+                    .markSystemic();
+            if (requiresFullContext) {
+                builder.newQueryLibrary();
+                builder.emptyQueryScope();
+            }
```
rcaudy, line 140: This is for `AggFormula`?

engine/table/src/main/java/io/deephaven/engine/table/impl/SelectOrUpdateListener.java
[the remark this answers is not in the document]
nbauernfeind: This is now pulled out into #2789 and evolving separately.

rcaudy (approved):

PR #2757, opened 2022-08-24
fix deephaven learn issue with query param scoping

nbauernfeind:
The docs nightly CI found that the use deephaven learn documentation failed.

The run: https://github.com/deephaven/deephaven.io/runs/7966897017?check_suite_focus=true#step:4:2304
The post: https://deephaven.io/core/docs/how-to-guides/use-deephaven-learn/


jmao-denver (approved):
LGTM

PR #2759, opened 2022-08-24
Use Sample for getHeapInfo

devinrsmith:
Quick fix that saves a volatile read and ensures we grab consistent values from currSnapshot. This also preserves the (potentially unintentional) semantics that getHeapInfo does *not* trigger a new snapshot.

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/util/RuntimeMemory.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/util/RuntimeMemory.java
@@ -123,11 +123,13 @@ public class RuntimeMemory {
          */
         public long totalMemory;
         /**
-         * The total number of GC collections since program start. The sum of {@link GarbageCollectorMXBean#getCollectionCount()}.
+         * The total number of GC collections since program start. The sum of
+         * {@link GarbageCollectorMXBean#getCollectionCount()}.
          */
         public long totalCollections;
         /**
-         * The approximated total time of GC collections since program start, in milliseconds. The sum of the {@link GarbageCollectorMXBean#getCollectionTime()}.
+         * The approximated total time of GC collections since program start, in milliseconds. The sum of the
+         * {@link GarbageCollectorMXBean#getCollectionTime()}.
          */
         public long totalCollectionTimeMs;
 
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/util/RuntimeMemory.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/util/RuntimeMemory.java
@@ -68,6 +68,13 @@ public class RuntimeMemory {
          * The approximated total time of GC collections since program start, in milliseconds.
          */
         long totalCollectionTimeMs;
+
+        private void readInto(Sample buf) {
+            buf.freeMemory = lastFreeMemory;
+            buf.totalMemory = lastTotalMemory;
+            buf.totalCollections = totalCollections;
+            buf.totalCollectionTimeMs = totalCollectionTimeMs;
+        }
     }
 
     private volatile Snapshot currSnapshot;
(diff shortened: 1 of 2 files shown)


jcferretti (approved):

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -185,12 +186,13 @@ public void executeCommand(ExecuteCommandRequest request, StreamObserver<Execute
     public void getHeapInfo(GetHeapInfoRequest request, StreamObserver<GetHeapInfoResponse> responseObserver) {
         GrpcUtil.rpcWrapper(log, responseObserver, () -> {
             final RuntimeMemory runtimeMemory = RuntimeMemory.getInstance();
+            final Sample sample = new Sample();
+            runtimeMemory.readOnly(sample);
```
mofojed, line 190: I think you're right - this was unintentional, and the getHeapInfo should call `read`, triggering an update (`if now >= snapshot.nextCheck`).

On Enterprise this class is different, but checking `totalMemory()` calls a `maybeUpdateValues()` (and Enterprise side could also benefit from reading it all into one sample, as it's susceptible to the volatile read as well - @cpwright for comment).

Side note I don't know why we need to pass in our own buff - why not just get the current snapshot? The entire snapshot gets replaced on each read. The `reset()` and `copy()` function are never called on `Sample`.

engine/table/src/main/java/io/deephaven/engine/table/impl/util/RuntimeMemory.java
```diff
@@ -172,10 +181,7 @@ public void read(final Sample buf) {
...
             }
         }
-        buf.freeMemory = snapshot.lastFreeMemory;
-        buf.totalMemory = snapshot.lastTotalMemory;
-        buf.totalCollections = snapshot.totalCollections;
-        buf.totalCollectionTimeMs = snapshot.totalCollectionTimeMs;
+        snapshot.readInto(buf);
         if (logInterval > 0 && now >= snapshot.nextLog) {
             synchronized (this) {
                 if (now >= currSnapshot.nextLog) {
```
mofojed, line 187: This part is kind of weird, I think to clean it up:
- The JSDoc for `logInterval` should specify that a value of `0` means "Don't log at all" (as opposed to logging every time)
- Checking against snapshot.nextCheck twice... I think it would be clearer inside the synchronized blocks if the if check was just `if (snapshot == currSnapshot)`... right now it's possible for one thread to replace the snapshot, but the logging to be logged by the next thread that comes in (which could replace the snapshot again even). Not that it _really_ matters.

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
cpwright: The runtime memory doesn't give you data atomically for free/total anyway so I'm not very concerned about the millisecond caching maybe giving you a cached value for one but fetching for the next.

engine/table/src/main/java/io/deephaven/engine/table/impl/util/RuntimeMemory.java
cpwright: The problem is you might not replace the snapshot, because another thread won.  An else of the now >= currSnapshot.nextCheck likely should set snapshot to currSnapshot.

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
devinrsmith: I'm not *too* concerned about getting values from inconsistent Snapshots (since the underlying calls to Runtime are independent anyways); I would like to use the "best API possible" though, and `Sample` is the way to do that IMO. (I'm going to motivate in a future patch to add timestamp b/c I think that's an obvious omission.)

devinrsmith: Snapshot is a private class, that's why we can't just return it. I think we could definitely improve RuntimeMemory in this regards, but this was meant to be a simple PR - I'd prefer follow-ups PRs if necessary.

devinrsmith: I've updated the PR to trigger snapshot when necessary from getHeapInfo

engine/table/src/main/java/io/deephaven/engine/table/impl/util/RuntimeMemory.java
devinrsmith: Happy to improve, but in future PRs?

mofojed (approved):

Issue #2723, opened 2022-08-08
Dh as a python library: h2o package running problem with deephaven_server.


Jeremiahcheng1206:
**Description**
I was trying to running h2o (using GPU to do computation) package with deephaven as a python library.  It gave me this error 'io.deephaven.integrations.python.PythonLogAdapter' object has no attribute 'isatty'. If I run it separately, meaning without starting the deephaven server. It works. Below are the code and output. 

**Steps to reproduce**
### With deephaven server:
```python
from deephaven_server import Server
s = Server(port=10000, jvm_args=["-Xmx4g"])
s.start()
from deephaven import ugp
ugp.auto_locking = True
import h2o
from h2o.automl import H2OAutoML
h2o.init()
data_frame=h2o.import_file("data07292022112829.parquet")
```
### The error:
'io.deephaven.integrations.python.PythonLogAdapter' object has no attribute 'isatty'

### Without deephaven_server:
```python
import h2o
from h2o.automl import H2OAutoML
h2o.init()
data_frame=h2o.import_file("data07292022112829.parquet")
print(data_frame)
```
### The output:
![image](https://user-images.githubusercontent.com/99699304/183514809-63221cf3-6f78-4b16-92a8-dc173deae25e.png)

**Versions**
 - Deephaven: 0.15.0 Dh as a python library
 - OS:  WSL
 - Browser: Google chrome
 - Docker: no docker

niloc132:
Arguably, h2o is misbehaving here, but its a reasonable thing for us to work with. From `h2o/utils/progressbar.py`:
```
        # Fix for PUBDEV-5048. H2O depends on isatty attribute, but some Python Notebooks override stdout and doesn't
        # specify it. The same holds for the encoding attribute bellow
        if not hasattr(sys.stdout, "isatty"):
            sys.stdout.isatty = lambda: False

        if not hasattr(sys.stdout, "encoding"):
            sys.stdout.encoding = sys.getdefaultencoding()

        if file_mode is None: file_mode = not sys.stdout.isatty()
```
Our Java PythonLogAdapter doesn't like being monkey patched like this.

This is broken in both directions: 
 * When starting from python, our PythonLogAdapter needs to support any possible attr that might be set on it - rather than actually supporting everything, it probably makes more sense to handle arbitrary attributes, and thus permit our logger to be monkey-patched like this, rather than imagine we can implement all possible use cases directly.
[...]

PR #2763, opened 2022-08-26
Rewire python<->java logging to be generally friendlier to python

### LINKED ISSUE #2734: Logs in DHaaL are confusing

niloc132:
Fixes #2734, #2723


niloc132:
To fully fix the `h2o` case, https://github.com/jpy-consortium/jpy/pull/85 will be required as well, as this apparently tries to log zero-byte strings.

py/server/deephaven_internal/stream.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    TextIOBase subclass that splits output between a delegate instance and a set of lambdas. Can delegate some calls to
+    the provided stream but actually write only to the lambdas.
+
+    Useful to adapt any existing file-like object (such as sys.stdout or sys.stderr), but actually let some other
+    processing take place before writing.
+    """
+
+    encoding = 'UTF-8'
+    closed = False
```
devinrsmith, line 14: These are variables attached to the *class* not the object instance. I'm assuming that's not what we want?

Should be set in constructor.

py/server/deephaven_internal/stream.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @staticmethod
+    def redirect(py_stream, java_stream):
+        if hasattr(py_stream, "encoding"):
+            encoding = py_stream.encoding
+        else:
+            encoding = 'UTF-8'
+        return TeeStream(
+            py_stream,
+            True,
```
devinrsmith, line 34: Should this be `False`? - I don't see `should_write_to_orig_stream` being used differently anywhere.

devinrsmith (commented):
Quick first pass. I'm going to run and introspect this locally later. Generally looks good.

py/server/deephaven_internal/stream.py
niloc132: Because these are declared (in C, as read-only), these must be re-declared at class level, or else they cannot be assigned in the constructor (or other methods).

There might be a better way to work around a C superclass, but this is what @jmao-denver directed.

py/server/deephaven_internal/stream.py
niloc132: It is False for `redirect()`, and True for `split`, and TeeStream has a field for it, so that write/flush/close can either pass through to the original stream, or ignore the original stream (and only delegate methods that query for characteristics, like isatty(), fileno(), and the encoding/closed properties).

There may well be a better way to handle this, but I think it is readable enough? Maybe use named params here instead of positional ones?

devinrsmith: Am I missing something? This is `redirect`, and the value is `True` here; and in `split`.

py/server/deephaven_internal/stream.py
devinrsmith: I'm confused, b/c `closed` should be a property of each `TeeStream`, not a global value. I'll take a look at TextIOBase later.

py/server/deephaven_internal/stream.py
niloc132: Whoops - it should definitely be False for redirect. Probably got switch in my "am I going crazy" debugging around jpy.

jmao-denver (changes requested):
Minor Pythonic/coding style issues

py/server/deephaven_internal/stream.py
niloc132: Looks like this isn't needed after all - my best guess is that this was an artifact of the other bugs we were running into with jpy.

jmao-denver (approved):
The changes look good to me. It might be nice if we could add some type hints in the Python code but as it stands, the built-in super class TextIOBase doesn't provide those either and the methods defined are so universally understood and expected, so skipping the type hints here doesn't really break any convention.

Integrations/src/main/java/io/deephaven/integrations/python/PyLogOutputStream.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    public void write(int i) throws IOException {
+        write(new byte[] {(byte) i});
+    }
+
+    @Override
+    public void write(@NotNull byte[] b) throws IOException {
+        // noinspection PrimitiveArrayArgumentToVarargsMethod
+        rawIoBaseSupplier.get().callMethod("write", new String(b));
```
devinrsmith, line 28: Do we need / want to account for Charset here?

devinrsmith: IMO, it would be better if we could call `org.jpy.PyObject#call(java.lang.Class<T>, java.lang.String, java.lang.Class<?>[], java.lang.Object[])` - this allows us to be more explicit w/ the parameter and return types.

As it stands now, all of the calls to `callMethod` return `PyObject`, which I think is extra garbage we shouldn't need to create.

Integrations/src/main/java/io/deephaven/integrations/python/PyLogOutputStream.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    public void flush() throws IOException {
+        rawIoBaseSupplier.get().callMethod("flush");
+    }
+
+    @Override
+    public void close() throws IOException {
+        rawIoBaseSupplier.get().callMethod("close");
+    }
```
devinrsmith, line 46: Ditto w/ `callMethod`s.

py/server/deephaven_internal/stream.py
```diff
@@ -0,0 +1,87 @@
...
+            flush_func=lambda: java_stream.flush(),
+            close_func=lambda: java_stream.close()
+        )
+
+    @classmethod
+    def redirect(cls, py_stream, java_stream):
+        if hasattr(py_stream, "encoding"):
+            encoding = py_stream.encoding
+        else:
+            encoding = 'UTF-8'
```
devinrsmith, line 27: I'm curious why we need to check `hasattr` here, but not for `split`?

py/server/deephaven_internal/stream.py
```diff
@@ -0,0 +1,87 @@
...
+        return TeeStream(
+            orig_stream=py_stream,
+            should_write_to_orig_stream=False,
+            write_func=lambda t: java_stream.write(bytes(t, encoding)),
+            flush_func=lambda: java_stream.flush(),
+            close_func=lambda: java_stream.close()
```
devinrsmith, line 33: I'm wondering - would it make sense to "simplify" and remove the lambdas?

```python
TeeStream(
  orig_stream=py_stream,
  other_stream=java_stream,
  should_write_to_orig_stream=...)
```

The write implementation could check it's own `encoding` property as you have it written.

py/server/deephaven_internal/stream.py
niloc132: Should be for both, thanks, was missing, since jpy doesn't break "split" case (nor have other writers i've encountered broken it).

Integrations/src/main/java/io/deephaven/integrations/python/PyLogOutputStream.java
niloc132: Yeah, that raises a question - why can jpy convert cleanly from py bytes to java byte[], but not vice versa? That's the real answer here, since it is already encoded (and we can't cleanly ask for the underlying encoding... since jpy might break it).

py/server/deephaven_internal/stream.py
niloc132: Possibly, I was imagining that while my current use case was limited to redirecting into or splitting into the JVM, we might case about generally teeing or redirecting output at a later date, and the io.IOBase type is not fun to work with - once I had the specifics worked out, I have a constructor that is easy to generalize, and two factory methods that are specific to their java use cases.

Integrations/src/main/java/io/deephaven/integrations/python/PyLogOutputStream.java
niloc132: That said, why be more explicit with the param types? Both call into `PyLib.callAndReturnValue`, but the variant that takes lots of `Class<?>`s does java-side type checks, rather than just passing the string instance - and there is no return type to be concerned with.

(Also, `.call()` vs `.callMethod()` has semantics that _technically_ we don't want here, we actually want `callMethod`. However, as far as I can tell, jpy just throws away that extra flag with a "todo why isn't this used?" sort of note)

niloc132: To be clear, these are not the same - `call` sets methodCall=false and its docs clarify that this is invoking a python callable object, not a method. All overrides do this.

In contrast, `callMethod` sets methodCall=true, and is intended for calling methods. We don't have (but arguably should?) an overload of `callMethod` like `call` does to support each arg type. I'd support going so far as to add overloads that return primitives or void, to avoid that extra cost you describe.

In this case, we want to call a method, we are not interacting with a callable object.

devinrsmith: the jpy implementation does not actually use that boolean internally - some sort of historic wart before my time.

devinrsmith (changes requested):
I'd like changes to the `PyObject#call` conventions.

Integrations/src/main/java/io/deephaven/integrations/python/PyLogOutputStream.java
niloc132: Yes - that's what I said above - in which case we should either figure out what issue might be there or historically have been there, or change the API - I'm just going off of the current stated contract of JPy, which is that this isn't to be used for method calls. But at the end of the day, we're using the wrong method that happens to do the right thing, for now, probably, to save on creating one object (that will either be called approx once per jvm, in the case of close(), or for write and maybe flush, are likely to be jitted to be inlined and have escap analysis handle anyway)?

I've you insist, I'll write the wrong code that currently does what we want but not what we're asking for, but this is an improvement to make to jpy and then improve our usage of it, so I'm objecting to deliberately doing the wrong thing.

devinrsmith: I've filed https://github.com/jpy-consortium/jpy/issues/87.

`org.jpy.PyObject#call(java.lang.Class<T>, java.lang.String, java.lang.Class<A0>, A0)` doesn't expose the notion of "methodCall", or have any javadoc (my fault). I think your argument falls down a little bit though, b/c if the implementation of that method instead used `true` instead of `false`, the outcome would be exactly the same, and is arguably an implementation detail.

I do like to consider myself against unnecessary pre-optimization. With respect to PyObject though, it's not the allocation itself I'm usually worried about, but the implications that extraneous PyObjects may cause (PyObjectReferences growth, unnecessarily keeping python references alive). And specifically as a glue layer between streams, there's potential for this to be heavily used. In general, if there is a way to accomplish the same outcome but without creating a PyObject, that's the direction I'll lean.

All that said - I've argued my case, and can let the code stand as-is if that's still your preference.

devinrsmith (dismissed):
Make sure to read latest comment.

devinrsmith (approved):

PR #2776, opened 2022-08-29
Add io.deephaven.web.WebLayoutType ObjectType plugin

### LINKED ISSUE #2774: Programmatic web layout
It should be simple to register an `io.deephaven.plugin.type.ObjectType` plugin so that users can point to filesystem locations for different layouts for different contexts.

For example:

```groovy
my_layout_1 = new WebLayout(Path.of("/tmp/layout1.json"))
my_layout_2 = new WebLayout(Path.of("/tmp/layout2.json"))
```

```
$ subscribe-fields
Created: typed_ticket {
  ticket {
    ticket: "s/my_layout_1"
  }
  type: "io.deephaven.web.WebLayout"
}
field_name: "my_layout_1"
field_description: "query scope variable"
application_id: "scope"

Created: typed_ticket {
  ticket {
    ticket: "s/my_layout_2"
  }
  type: "io.deephaven.web.WebLayout"
}
field_name: "my_layout_2"
field_description: "query scope variable"
application_id: "scope"
```

The web client can eventually recognize the type `io.deephaven.web.WebLayout` to have a UX around listing and changing layouts.

devinrsmith:
Fixes #2774

### DIFF
--- a/plugin/web/src/main/java/io/deephaven/web/WebLayoutType.java
+++ b/plugin/web/src/main/java/io/deephaven/web/WebLayoutType.java
@@ -1,0 +1,22 @@
+package io.deephaven.web;
+
+import com.google.auto.service.AutoService;
+import io.deephaven.plugin.type.ObjectType;
+import io.deephaven.plugin.type.ObjectTypeClassBase;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.file.Files;
+
+
+@AutoService(ObjectType.class)
+public final class WebLayoutType extends ObjectTypeClassBase<WebLayout> {
+    public WebLayoutType() {
+        super(WebLayout.class.getName(), WebLayout.class);
+    }
+
+    @Override
+    public void writeToImpl(Exporter exporter, WebLayout layout, OutputStream out) throws IOException {
+        Files.copy(layout.path(), out);
+    }
+}
--- a/plugin/web/build.gradle
+++ b/plugin/web/build.gradle
@@ -1,0 +1,10 @@
+plugins {
+    id 'io.deephaven.project.register'
+    id 'java-library'
+}
+
+dependencies {
+    api project(':plugin')
+
+    Classpaths.inheritAutoService(project)
+}
--- a/plugin/web/gradle.properties
+++ b/plugin/web/gradle.properties
@@ -1,0 +1,1 @@
+io.deephaven.project.ProjectType=JAVA_PUBLIC
--- a/plugin/web/src/main/java/io/deephaven/web/WebLayout.java
+++ b/plugin/web/src/main/java/io/deephaven/web/WebLayout.java
@@ -1,0 +1,16 @@
+package io.deephaven.web;
+
+import java.nio.file.Path;
+import java.util.Objects;
+
+public final class WebLayout {
+    private final Path path;
+
+    public WebLayout(Path path) {
+        this.path = Objects.requireNonNull(path);
+    }
+
+    Path path() {
+        return path;
+    }
+}
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -69,6 +69,7 @@ dependencies {
 
     runtimeOnly project(':plugin-figure')
     runtimeOnly project(':plugin-partitionedtable')
+    runtimeOnly project(':plugin-web')
 
     Classpaths.inheritGrpcPlatform(project, 'api')
 
--- a/settings.gradle
+++ b/settings.gradle
@@ -233,6 +233,9 @@ project(':plugin-figure').projectDir = file('plugin/figure')
 include(':plugin-partitionedtable')
 project(':plugin-partitionedtable').projectDir = file('plugin/partitionedtable')
 
+include(':plugin-web')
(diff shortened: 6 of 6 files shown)


plugin/web/src/main/java/io/deephaven/web/WebLayoutType.java
```diff
@@ -0,0 +1,22 @@
...
+
+@AutoService(ObjectType.class)
+public final class WebLayoutType extends ObjectTypeClassBase<WebLayout> {
+    public WebLayoutType() {
+        super(WebLayout.class.getName(), WebLayout.class);
+    }
+
+    @Override
+    public void writeToImpl(Exporter exporter, WebLayout layout, OutputStream out) throws IOException {
+        Files.copy(layout.path(), out);
```
mofojed, line 20: This paradigm seems to be valid for reading _any_ type of file from the server, are we limiting ourselves by calling it WebLayout?
Could be used for notebooks too?

devinrsmith: For sure - the reason why we are calling it `WebLayout` is because the client needs some context about *what* this file is. We could just as easily create `WebNotebook`, etc. I think using the plugin typing system is better than trying to have a generic "File" type with a "type" field in it.

devinrsmith: I'm happy to expand the scope of the PR / issue to have `WebNotebook` - note: there will be no inherent ability to save the file, you'll only be able to fetch/read.

mofojed: Devil's advocate - Why is using the plugin system better? If I want to read a new type of file, I'd need to add another plugin on the server in addition to updating web plugin.
A `File` object could just have metadata about the type of file loaded.

devinrsmith: In that case, we would need an "untyped" subtyping system; and your serialization would necessarily be more complex:

```groovy
my_layout = new FileType(Path.of("/tmp/my_layout.json"), "layout");
my_notebook = new FileType(Path.of("/tmp/my_notebook.groovy"), "notebook");
```

```
$ fetch-object --variable my_layout --type "io.deephaven.FileType"  
{
  "subtype": "layout",
  "data": "<hex-encoding-of-data>"
}

$ fetch-object --variable my_notebook --type "io.deephaven.FileType"  
{
  "subtype": "notebook",
  "data": "<hex-encoding-of-data>"
}
```

And also, you wouldn't know the subtype *before* you fetched it.

devinrsmith: The server side is dead-simple. We could make it even easier w/ an abstract base class if we want to support WebLayout, WebNotebook, etc.

niloc132: I think I'm with Bender - it is too bland and unopinionated to be an actual "web layout" feature, we should either spec it out further and support that use case rigorously, or back it off and be "i'm just a file, `io.deephaven.File`, here's my bytes", let the client work it out with mime type, extension (instead of subtype?), or be explicit about the kind of data it holds, "I'm a json blob, `io.deephaven.Json`, read my data to see what i hold".

niloc132: If we just need a way to offer files, the webdav or fileservice should suffice - offering layouts as objects, doesn't solve the chicken/egg discovery problem

"> i need to know what to start with"
"< here, use this layout"
"> i see 3 layouts, which one is the _start_ one"

if we just need a naming convention for "startup layout", we can also make that a filesystem convention? Since these objects can only contain bytes anyways (and not embedded tickets), do they do things that files don't do?

devinrsmith: I'm not trying to spec out the web UX here - I'm happy to have Don / Bender go at that perspective more. The server side implementation itself though is very simple on purpose. If we feel like we need to have an end-to-end working layout feature PoC first, I'm happy to hold this until we have the web-side at an equivalent state.

I don't think webdav or fileservice in-and-of-itself solves the "programmatic" part of this ticket.

Again, the big reason why I don't like a generic File type is that the client has to fetch the object before it can learn what the specific type is. IMO, the client should be able to create a UX list of layouts without needing to fetch every single generic File.

mofojed: Are we expecting to serve up layouts both from files on the file system and through this programmatic method as well then? Wouldn't it just be as easy to just have the layouts in a directory and use the file system we're already developing (webdav or whatever)?
Or put another way - what use case is this solving? To create the layout programmatically, you're going to need to read it from disk already, which means it'd need to be in the data directory (right?) so why not just put it in the `layouts` dir?

devinrsmith: It's trying to solve the ability to source programmatic layouts. Users may choose to integrate with external APIs that they need to source layouts from.

PR #2792, opened 2022-09-01
Try to Pool / re use compressors where possible.  Fixes #2788

### LINKED ISSUE #2788: ColumnWriterImpl exacerbates bad behavior of SnappyCompressor
SnappyCompressor has some internal bad allocation behavior that causes it to be O(n^2) in the number of blocks compressed.  A separate bug will be files with apache -- however we can mitigate it by ensuring that we pool and re-use internal compressors where possible

abaranec:


extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReaderImpl.java
```diff
@@ -38,7 +38,7 @@ public class ColumnChunkReaderImpl implements ColumnChunkReader {
     private final ColumnChunk columnChunk;
     private final SeekableChannelsProvider channelsProvider;
     private final Path rootPath;
-    private final Compressor decompressor;
+    private final CompressorAdapter decompressor;
```
niloc132, line 41: this can never be closed, so never is returned to the pool, as modified in this patch - this block in CodecWrappingCompressorImpl.close() will never execute:
```
            if (innerDecompressor != null && innerDecompressorPooled) {
                CodecPool.returnDecompressor(innerDecompressor);
            }
```

extensions/parquet/compression/src/main/java/io/deephaven/parquet/compress/DeephavenCompressorAdapterFactory.java
```diff
@@ -45,73 +41,109 @@ public class DeephavenCodecFactory {
...
-        private CodecWrappingCompressor(CompressionCodec compressionCodec) {
+        private boolean innerCompressorPooled;
+        private Compressor innerCompressor;
+
+        public boolean innerDecompressorPooled;
+        private Decompressor innerDecompressor;
+
+        private CodecWrappingCompressorAdapter(CompressionCodec compressionCodec) {
             this.compressionCodec = compressionCodec;
+            this.innerCompressor = compressionCodec.createCompressor();
```
niloc132, line 75: I dont think this line makes any sense. Either a) it returns non-null, and so compress(OutputStream) below can never borrow from the pool, or b) it returns null, and CodecPoo.getCompressor will also return null, as will the same call to compressionCodec.createCompressor().

With this line removed, ParquetTableReadWriteTest will correctly stop in close()'s first if block to return the codec to the pool, otherwise it never stops there.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReaderImpl.java
niloc132: We still aren't returning decompressors, right?

niloc132 (approved):
Looks good. Filed a follow-up issue #2920 for myself, unless you want to handle any of these points in this patch.

- [ ] Disable/fix meaningless "Got brand-new [de]compressor" messages. `zstd`, `lzo`, and `gzip` end up logging multiple times when decompressing each table, and `zstd` and `gzip` also multiple times when compressing. 
- [ ] Once we find that we can't ask the pool for a compressor, quit asking. This will somewhat mitigate the above, but also keeps us from doing the extra work of asking the pool for an instance we know we can't find (which itself requires taking the lock, etc). Specifically, if `innerCompressor` is null and the compress method ends up calling `compressionCodec.createOutputStream(os)`, we should keep track of that fact. This will specifically protect the zstd and gzip codecs from having to rediscover this for each page within each column, as well as for the dictionary for each column.
- [ ] For x86, investigate providing the native codec, which will enable the gzip codec not worry about the above case and pool its de/compressors.

PR #2794, opened 2022-09-01
When an ExportObject runner throws StatusRuntimeException forward to client

### LINKED ISSUE #1871: StatusRuntimeException from inside submit not propogated
Discovered as part of https://github.com/deephaven/deephaven-core/pull/1787, `StatusRuntimeException`s that are thrown during execution of `io.deephaven.server.session.SessionState.ExportBuilder#submit(java.util.concurrent.Callable<T>)` don't have their message propogated to the client.

nbauernfeind:
Fixes #1871

### DIFF
--- a/server/src/main/java/io/deephaven/server/session/SessionState.java
+++ b/server/src/main/java/io/deephaven/server/session/SessionState.java
@@ -524,6 +524,7 @@ public class SessionState {
         /** used to identify and propagate error details */
         private String errorId;
         private String dependentHandle;
+        private Exception caughtException;
 
         /**
          * @param exportId the export id for this export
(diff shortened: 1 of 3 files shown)


devinrsmith (dismissed):
LGTM - make sure to get Colin signoff, I think he had insight wrt #1871

niloc132 (commented):
Two cases that need to be tested, that don't go through batch or have an explicit comment about not wanting to call onError on the export builder:

 * ScopeTicketResolver.publish calls submit() on the `publish` instance with no `onError` being set
 * ExportTicketResolver.flightInfoFor explicitly throws grpc status ex, but has no `onError`

niloc132 (approved):

PR #2808, opened 2022-09-06
Refactor CompilerTools into QueryCompiler

### LINKED ISSUE #2781: CompilerTools.Context should be refactored to be an instance of CompilerTools
The last piece of cleaning up QueryScope/QueryLibrary/CompilerTools looks like simplifying the design of CompilerTools.

Currently the entry points for CompilerTools are static methods that all fetch the ExecutionContext instance of CompilerTools.Context. This pattern (static methods that use an instance of context) is almost the same as using member methods on the installed context.

A small, but generally good, improvement would be to rename CompilerTools to QueryCompiler and embed the context as instance variables.

nbauernfeind:
Fixes #2781.


engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -53,24 +53,85 @@ public class CompilerTools {
...
+        additionalClassLocations = null;
+    }
+
+    private QueryCompiler(File classDestination) {
+        this(classDestination, null, false);
+    }
+
+    private QueryCompiler(File classDestination, ClassLoader parentClassLoader, boolean isCacheDirectory) {
+        if (parentClassLoader == null) {
+            parentClassLoader = this.getClassLoader();
```
rcaudy, line 101: I think this is incorrect. `this.getClassLoader()` returns an as-yet-uninitialized instance member, `ucl`. Maybe you want `QueryCompiler.class.getClassLoader()`?

Also, I slightly prefer the following pattern:
1. Declare `parentClassLoader` `final`.
2. Create `final ClassLoader parentClassLoaderToUse = parentClassLoader != null ? parentClassLoader : QueryCompiler.class.getClassLoader();`

engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -144,327 +193,255 @@ public static void writeClass(final File destinationDirectory, final String clas
...
-                    // if it is a groovy class, always try to use the instance in the shell
-                    if (name.startsWith(DYNAMIC_GROOVY_CLASS_PREFIX)) {
-                        try {
-                            return ucl.getParent().loadClass(name);
-                        } catch (final ClassNotFoundException ignored) {
-                            // we'll try to load it otherwise
-                        }
-                    }
+    private static void ensureDirectories(final File file, final Supplier<String> runtimeErrMsg) {
+        // File.mkdirs() checks for existrance on entry, in which case it returns false.
```
rcaudy, line 275: Can you correct this old typo?
```suggestion
        // File.mkdirs() checks for existence on entry, in which case it returns false.
```

rcaudy (approved):

PR #2811, opened 2022-09-06
Make DeephavenSessionTestBase use NoLanguageDeephavenSession

nbauernfeind:
`AnnotatedTableTest` began to take ~20m to execute after the ExecutionContext merge. Groovy sessions were running their init scripts and many tests required a query compilation as no results were being cached. It now runs all 968 tests within 10-20s (back to where it originally was).

### DIFF
--- a/server/jetty/src/test/java/io/deephaven/server/jetty/JettyFlightRoundTripTest.java
+++ b/server/jetty/src/test/java/io/deephaven/server/jetty/JettyFlightRoundTripTest.java
@@ -7,7 +7,7 @@ import dagger.Component;
 import dagger.Module;
 import dagger.Provides;
 import io.deephaven.server.arrow.ArrowModule;
-import io.deephaven.server.log.LogModule;
+import io.deephaven.server.runner.ExecutionContextUnitTestModule;
 import io.deephaven.server.session.SessionModule;
 import io.deephaven.server.test.AuthTestModule;
 import io.deephaven.server.test.FlightMessageRoundTripTest;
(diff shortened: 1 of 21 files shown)


server/src/test/java/io/deephaven/server/runner/DeephavenApiServerTestBase.java
```diff
@@ -40,8 +82,8 @@ public void setUp() throws Exception {
         logBuffer = new LogBuffer(128);
         LogBufferGlobal.setInstance(logBuffer);
 
-        serverComponent = DaggerDeephavenApiServerInProcessGroovyComponent
-                .builder()
+        serverComponent = DaggerDeephavenApiServerTestBase_TestComponent.builder()
+                .withExecutionContext(ExecutionContext.createForUnitTests())
```
devinrsmith, line 86: I see that `FlightTestModule` has a provider for this - should we have a `ExecutionContextUnitTestModule` that provides this instead?

server/jetty/src/test/java/io/deephaven/server/jetty/JettyFlightRoundTripTest.java
```diff
@@ -7,6 +7,7 @@
 import dagger.Module;
 import dagger.Provides;
 import io.deephaven.server.arrow.ArrowModule;
+import io.deephaven.server.log.LogModule;
```
devinrsmith, line 10: extraneous

devinrsmith (commented):
Very nice - I've tested this locally and it fixes the long test times.

devinrsmith (approved):

PR #2813, opened 2022-09-07
Remove Python specific checks for parallel evaluation of formulas

### LINKED ISSUE #2742: Enable parallel formula evaluation in the Python script session.
As a DH Python user, I want to take advantage of the query engine's parallel processing capability as much as possible by utilizing the modified behavior of JPY releasing and reacquiring of the GIL around calling a Java method from Python.

jmao-denver:
The change in JPY to release the GIL whenever a Java method is called has made the check to prevent potential deadlock unnecessary. 

Fixes #2742

### DIFF
--- a/engine/api/src/main/java/io/deephaven/engine/table/ColumnSource.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/ColumnSource.java
@@ -174,19 +174,6 @@ public interface ColumnSource<T>
     }
 
     /**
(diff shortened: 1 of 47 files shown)


rcaudy (commented):
I've reviewed the code, and I think it does what it purports to. We have a wider discussion about the resulting performance to account for before we merge this.

rcaudy (approved):

PR #2824, opened 2022-09-08
Parse date times correctly when importing CSV

### LINKED ISSUE #2814: DH-13126 Web CSV Upload Does not Parse DBDateTime correctly
**Description**

This Deephaven produced file is read by readCsv w/o problems, the Web refuses to upload it because it does not like “NY”.

[tde20.csv](https://github.com/deephaven/web-client-ui/files/9507322/tde20.csv)

mofojed:
- Wasn't taking into account subsecond patterns and time zones
- Copied logic over from CsvTypeParser in Enterprise
- Optimize by only changing the pattern when it fails to parse
- Added additional short code time zones to match what is on Enterprise
- Fixes #2814
- Tested with a file with multiple formats for dates:
[multiple-formats.csv](https://github.com/deephaven/deephaven-core/files/9538411/multiple-formats.csv)

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/parse/JsDataHandler.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/parse/JsDataHandler.java
@@ -94,32 +94,32 @@ public enum JsDataHandler {
     },
     DATE_TIME(Type.Int, "io.deephaven.time.DateTime", "datetime") {
         // Ensures that the 'T' separator character is in the date time
-        private String ensureSeparator(String s, ParseContext context) {
-            if (s.charAt(context.separatorIndex) == ' ') {
+        private String ensureSeparator(String s) {
+            if (s.charAt(SEPARATOR_INDEX) == ' ') {
                 StringBuilder stringBuilder = new StringBuilder(s);
-                stringBuilder.setCharAt(context.separatorIndex, 'T');
+                stringBuilder.setCharAt(SEPARATOR_INDEX, 'T');
                 return stringBuilder.toString();
             }
             return s;
         }
 
-        // Updates the pattern for the correct number of subsecond digits 'S'
-        private String getSubsecondPattern(String s, ParseContext context) {
+        // Guess the pattern for the correct number of subsecond digits 'S'
+        private String getSubsecondPattern(String s) {
             final int decimalIndex = s.indexOf('.');
             if (decimalIndex == -1) {
                 // No subsecond digits
-                return context.dateTimePattern;
+                return DEFAULT_DATE_TIME_PATTERN;
             }
             final int numDigits = s.length() - decimalIndex - 1;
             final StringBuilder stringBuilder = new StringBuilder(numDigits);
             for (int i = 0; i < numDigits; i++) {
                 stringBuilder.append('S');
             }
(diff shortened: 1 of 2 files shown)


web/client-api/src/main/java/io/deephaven/web/client/api/parse/JsDataHandler.java
```diff
@@ -94,32 +94,32 @@ public void write(Object[] data, ParseContext context, JsConsumer<Node> addNode,
...
 
-        // Updates the pattern for the correct number of subsecond digits 'S'
-        private String getSubsecondPattern(String s, ParseContext context) {
+        // Guess the pattern for the correct number of subsecond digits 'S'
+        private String getSubsecondPattern(String s) {
             final int decimalIndex = s.indexOf('.');
             if (decimalIndex == -1) {
                 // No subsecond digits
-                return context.dateTimePattern;
+                return DEFAULT_DATE_TIME_PATTERN;
```
niloc132, line 111: why use this constant instead of the specified pattern in the context (which will default to the same constant)?

web/client-api/src/main/java/io/deephaven/web/client/api/parse/JsDataHandler.java
```diff
@@ -136,12 +136,23 @@ private long parseDateString(String str, ParseContext context) {
...
+            boolean needsAdjustment = JsTimeZone.needsDstAdjustment(timeZoneString);
+
+            try {
+                // First try with the pattern we already have
+                return JsDateTimeFormat.getFormat(context.dateTimePattern).parseWithTimezoneAsLong(dateTimeString,
+                        timeZone, needsAdjustment);
+            } catch (IllegalArgumentException e) {
+                // We failed to parse with the existing context pattern, try and update the pattern from the string of
+                // text and do it again
+                context.dateTimePattern = getSubsecondPattern(dateTimeString);
```
niloc132, line 152: if the format was already something custom but missing subseconds, we're reverting to ISO and attempting to add subseconds to that?

web/client-api/src/main/java/io/deephaven/web/client/api/parse/JsDataHandler.java
mofojed: The pattern in the context can change

web/client-api/src/main/java/io/deephaven/web/client/api/parse/JsDataHandler.java
mofojed: The `getSubsecondPattern` basically just determines what pattern is appropriate given the string, adding the amount of subseconds necessary

web/client-api/src/main/java/io/deephaven/web/client/api/parse/JsDataHandler.java
niloc132: Yes, isn't that the point? if the user's pattern (which could be the default), and the data changes partway (as we seem to allow differences in data, with additional subsecond precision), wouldnt it make sense to base it on the provided format? Or do we think the user will switch formats entirely when their data changes?

I guess I don't understand the use case where the data changes, and we want to _replace_ the specified format, but not use the original format later?

niloc132 (approved):

PR #2828, opened 2022-09-08
Upgrade to jetty

jjbrosnan:
This updates every docker-compose file (and other aspects of the folders) in the containers directory.

There should be discussion about what belongs in data and notebooks.

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -57,7 +57,7 @@ jobs:
           tags: |
             type=edge,branch=main
             type=raw,priority=950,enable=${{ startsWith(github.ref, 'refs/heads/release/v') }},value=${{ steps.semver.outputs.semver }}
-            type=raw,enable=${{ startsWith(github.ref, 'refs/heads/release/v0.17.') }},value=latest
+            type=raw,enable=${{ startsWith(github.ref, 'refs/heads/release/v0.18.') }},value=latest
             type=ref,event=branch
 
       - name: Docker meta server nltk
@@ -71,7 +71,7 @@ jobs:
           tags: |
             type=edge,branch=main
             type=raw,priority=950,enable=${{ startsWith(github.ref, 'refs/heads/release/v') }},value=${{ steps.semver.outputs.semver }}
-            type=raw,enable=${{ startsWith(github.ref, 'refs/heads/release/v0.17.') }},value=latest
+            type=raw,enable=${{ startsWith(github.ref, 'refs/heads/release/v0.18.') }},value=latest
             type=ref,event=branch
 
       - name: Docker meta server pytorch
@@ -85,7 +85,7 @@ jobs:
           tags: |
             type=edge,branch=main
             type=raw,priority=950,enable=${{ startsWith(github.ref, 'refs/heads/release/v') }},value=${{ steps.semver.outputs.semver }}
-            type=raw,enable=${{ startsWith(github.ref, 'refs/heads/release/v0.17.') }},value=latest
+            type=raw,enable=${{ startsWith(github.ref, 'refs/heads/release/v0.18.') }},value=latest
             type=ref,event=branch
 
       - name: Docker meta server sklearn
@@ -99,7 +99,7 @@ jobs:
           tags: |
             type=edge,branch=main
             type=raw,priority=950,enable=${{ startsWith(github.ref, 'refs/heads/release/v') }},value=${{ steps.semver.outputs.semver }}
-            type=raw,enable=${{ startsWith(github.ref, 'refs/heads/release/v0.17.') }},value=latest
+            type=raw,enable=${{ startsWith(github.ref, 'refs/heads/release/v0.18.') }},value=latest
             type=ref,event=branch
 
       - name: Docker meta server tensorflow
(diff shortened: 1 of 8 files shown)


containers/groovy-examples/.gitignore
```diff
@@ -0,0 +1,5 @@
+notebooks/notebooks/*
+!notebooks/notebooks/01-example.py
+
+data/*
+!data/.ignore
```
devinrsmith, line 5: We may want to employ a strategy at our root gitignore to *keep* any files named `.gitkeep`; I think I've seen this used elsewhere. That way we only need the logic in one .gitignore file.

The one problem we found with that is then there is a `.gitkeep` file in notebooks UI. There are various ways we could solve this problem, ranging from the client knowing to ignore that file, or from updated the webdav server to ignore it.

containers/groovy-examples/docker-compose.yml
```diff
@@ -1,49 +1,35 @@
 version: "3.4"
 
 services:
-  server:
-    image: ghcr.io/deephaven/server:${VERSION:-latest}
-    expose:
-      - '8080'
+  deephaven:
+    image: ghcr.io/deephaven/server:0.16.0-groovy
```
devinrsmith, line 5: I probably owe a "latest" style tag for the jetty images. Ie,

```
ghcr.io/deephaven/server:latest-groovy
ghcr.io/deephaven/server:latest-python
```

containers/groovy-examples/docker-compose.yml
```diff
@@ -1,49 +1,35 @@
...
     environment:
-      - JAVA_TOOL_OPTIONS=-Xmx4g -Ddeephaven.console.type=groovy
+      - JAVA_OPTS=-Xmx4g
 
-  web:
-    image: ghcr.io/deephaven/web:${VERSION:-latest}
-    expose:
-      - '80'
+  notebooks:
+    image: ghcr.io/devinrsmith/webdav:latest
```
devinrsmith, line 14: We probably want to transition this to the deephaven namespace from https://github.com/devinrsmith/deephaven-webdav

devinrsmith (commented):
Thanks for this - I think this is largely the direction we should move in. There are a few items I've specifically noted that I think I need to follow-up with. That should not stop us IMO from testing this in the meantime.

I haven't run most of these examples, but I'm assuming you've run them all successfully.

jjbrosnan:
When I created this PR, I'd tested the base setups and everything works as expected. Our Python + AI containers are all failing with the same error messages:

```
scikit-learn-reverse-proxy-1  | time="2022-09-12T17:29:08Z" level=error msg="service \"deephaven-scikit-learn\" error: port is missing" container=deephaven-scikit-learn-4867bdc29662ee63cb8efb55e038c15e325f077fe1a939a5a61c3fc467c675ad providerName=docker
```

I'm not sure what the cause of this is, but I'm guessing our Python + AI images are out of date, and that they need to be updated to use v0.16, or whatever is latest in future releases.

containers/groovy-examples/docker-compose.yml
chipkent: Yep

containers/groovy-examples/docker-compose.yml
chipkent: I assume we want env controls so that we can easily change versions.
```
ghcr.io/deephaven/server:${VERSION:-latest-groovy}
```

containers/groovy-examples/docker-compose.yml
chipkent: env controls

containers/groovy-examples/.gitignore
```diff
@@ -0,0 +1,5 @@
+notebooks/notebooks/*
+!notebooks/notebooks/01-example.py
```
chipkent, line 2: Is there a reason this one is singled out?

containers/groovy/.gitignore
```diff
@@ -0,0 +1,5 @@
+notebooks/notebooks/*
+!notebooks/notebooks/01-example.py
```
chipkent, line 2: Is there a reason this one is singled out?

containers/groovy/docker-compose.yml
```diff
@@ -1,43 +1,29 @@
 version: "3.4"
 
 services:
```
chipkent, line 3: same comments from above

containers/python-examples/NLTK/docker-compose.yml
```diff
@@ -1,48 +1,35 @@
 version: "3.4"
 
 services:
```
chipkent, line 3: same comments

containers/python-examples/PyTorch/.gitignore
```diff
@@ -0,0 +1,5 @@
+notebooks/notebooks/*
+!notebooks/notebooks/01-example.py
```
chipkent, line 2: why exclude one?

containers/python-examples/NLTK/notebooks/notebooks/01-example.py
```diff
@@ -0,0 +1 @@
+print("Hello world!")
```
chipkent, line 1: Do we want a more DH specific example?

chipkent (changes requested):
Changes need to be applied in many places.

containers/python-examples/NLTK/notebooks/notebooks/01-example.py
jjbrosnan: Yes. The example will likely vary depending on which deployment is picked.

Everything should come with at least one base DH example. Containers w/ examples should come w/ scripts to work on local data, and AI containers should come w/ a script that uses that particular package. Probably the iris stuff.

Issue #2563, opened 2022-06-22
EPIC: Authorization support


niloc132:
Part of a series of work to implement AAA in Deephaven Core.

Authorization is a mechanism to describe access to a system to users, typically a set of rules that define whether a given operation is allowed or rejected. Sometimes abbreviated as authz (as opposed to authn for authentication), authorization does not require authentication as a prerequisite, as anonymous users could still be granted access to certain operations. Typically however most authz rules will be based on the identity of the user, as handled by the authn mechanism.

Identifying operations and granting access can be handled in several ways - popular ones would include defining roles to assign to users (or to groups, and in turn let users be assigned to groups). The most generic approach would likely be to offer fine-grained API hooks to allow/deny each operation based on coarse grained roles, and to have a simple implementation that only checks if a user is granted that role. As the vast majority of our API surface area is in gRPC calls, it seems simplest to generate those api hooks directly, using the proto objects that describe the underlying services, taking care not just to intercept the original request, but later streamed requests too.

Rate limiting is another form of access control in this way, allowing even non-authenticated users to be limited in what actions they can perform based on what they have done so far. Limiting this by session is probably not sufficient, but by IP address might be necessary too. An http2 proxy can probably help with some of this, but might not be sufficient for bidi grpc streams.

rcaudy:
There's been a bit of related discussion in #2414, which we should incorporate if appropriate,

PR #2843, opened 2022-09-13
Implement gRPC storage API for notebooks and layouts

### LINKED ISSUE #1768: Notebook and layout support with authentication
WebDAV support, for notebooks. Arguably, this feature should be rewritten to be more secure, at least to require authentication before allowing write access. There are servlet implementations for webdav, but they aren't compatible with the latest servlet apis - they can probably be mechanically updated as a build step, but it may be easier to just switch to a gRPC api instead.

Let's do this right with gRPC services that let you save and request layouts and notebooks.
~1 week ish.

@devinrsmith  and @mofojed  discussions will consume a lot of this time.

niloc132:
This patch aims to replace existing WebDAV use cases with something that can be rolled into the main server process, and leaves room to establish additional authorization rules and use case conventions around this feature. The present implementation is built to allow the server implementation to be replaced by something else that would use deephaven-server dependencies and is backed by something other than the file system.

The expectation is that this isn't specific to notebooks, but serves as a general storage API for clients, so they can read/write their own data/code, and possibly share some code. Likely, this would be read-only or largely unavailable by non-privileged users. It could be that different users see different contents to the same top-level directory in some future implementation, but for this version a single directory is made available to all users.

Also introduces a JavaScript API for this and other upcoming features, in the form of `dh.CoreClient`, a starting point to connect to a deephaven-core instance or cluster. At this time authentication is not implemented, but will be soon - only the storage service is exposed. Example:

```javascript
const client = new dh.CoreClient("http://localhost:10000");
const store = client.getStorageService();
await store.listItems("", "*.txt").then(console.dir);
```

Fixes #1768


proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
```diff
@@ -0,0 +1,89 @@
...
+option java_multiple_files = true;
+option optimize_for = SPEED;
+option go_package = "github.com/deephaven/deephaven-core/go/internal/proto/storage";
+
+/*
+ * Shared storage management service.
+*/
+service StorageService {
+  rpc ListItems(ListItemsRequest) returns (ListItemsResponse) {}
+  rpc FetchFile(FetchFileRequest) returns (FetchFileResponse) {}
```
niloc132, line 17: We might want this to be a streaming endpoint for very large files, see also the ByteRange message (commented out) by which a client could request regions of a file.

If we go down this road we would likely want a way to update ranges in a file too, appending new ranges, truncating the file, etc.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
```diff
@@ -0,0 +1,89 @@
...
+  // Can create new files or modify existing
+  rpc SaveFile(SaveFileRequest) returns (SaveFileResponse) {}
+  rpc MoveItem(MoveItemRequest) returns (MoveItemResponse) {}
+  rpc CreateDirectory(CreateDirectoryRequest) returns (CreateDirectoryResponse) {}
+  rpc DeleteItem(DeleteItemRequest) returns (DeleteItemResponse) {}
+}
+
+message ListItemsRequest {
+  string path = 1;// the path to the directory to list. empty to list top level
+  optional string filter_glob = 2;// A pattern to filter for, with "*" as a wildcard
```
niloc132, line 27: More docs would be good here, this is currently implemented using https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystem.html#getPathMatcher-java.lang.String- with the following restrictions:
 - no path separators allowed, so other directories can't be queried
 - no `**` so sub-directories can't be queried

web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
```diff
@@ -0,0 +1,85 @@
+package io.deephaven.web.client.api;
+
+import elemental2.promise.Promise;
+import io.deephaven.web.client.api.storage.JsStorageService;
+import jsinterop.annotations.JsType;
+
+@JsType(namespace = "dh")
+public class CoreClient extends QueryConnectable<CoreClient> {
```
niloc132, line 8: Most of this file is not ready for prime time, but exists to grant access to `getStorageService()`. I'd be happy to remove or comment out most of it if preferred, or just expect changes in the near future with auth/config tickets.

web/client-api/src/main/java/io/deephaven/web/client/api/LoginCredentials.java
```diff
@@ -0,0 +1,47 @@
+package io.deephaven.web.client.api;
+
+import jsinterop.annotations.JsOverlay;
+import jsinterop.annotations.JsPackage;
+import jsinterop.annotations.JsType;
+
+@JsType(namespace = JsPackage.GLOBAL, name = "Object", isNative = true)
+public class LoginCredentials {
```
niloc132, line 8: Not necessary yet, except as part of the larger picture for dh.CoreClient.

server/src/main/java/io/deephaven/server/notebook/FilesystemStorageServiceGrpcImpl.java
```diff
@@ -0,0 +1,188 @@
...
+            Path dir = resolveOrThrow(request.getPath());
+            try (Stream<Path> list = Files.list(dir)) {
+                for (Path p : (Iterable<Path>) list::iterator) {
+                    if (!matcher.matches(dir.relativize(p))) {
+                        continue;
+                    }
+                    builder.addItems(FileInfo.newBuilder()
+                            .setPath(p.getFileName().toString())
+                            .setSize(Files.isDirectory(p) ? 0 : Files.size(p))
+                            .setKind(Files.isDirectory(p) ? FileKind.DIRECTORY : FileKind.FILE)
```
nbauernfeind, line 84: You could call `Files.isDirectory` once and re-use the value.

server/src/main/java/io/deephaven/server/notebook/FilesystemStorageServiceGrpcImpl.java
```diff
@@ -0,0 +1,188 @@
...
+                }
+            } catch (NoSuchFileException noSuchFileException) {
+                throw GrpcUtil.statusRuntimeException(Code.FAILED_PRECONDITION, "Directory does not exist");
+            }
+            responseObserver.onNext(builder.build());
+            responseObserver.onCompleted();
+        });
+    }
+
+    private PathMatcher makeMatcher(String filterGlob) {
```
nbauernfeind, line 95: `private static`?

server/src/main/java/io/deephaven/server/notebook/FilesystemStorageServiceGrpcImpl.java
```diff
@@ -0,0 +1,188 @@
...
+
+    @Override
+    public void fetchFile(FetchFileRequest request, StreamObserver<FetchFileResponse> responseObserver) {
+        GrpcUtil.rpcWrapper(log, responseObserver, () -> {
+            final byte[] bytes;
+            try {
+                bytes = Files.readAllBytes(resolveOrThrow(request.getPath()));
+            } catch (NoSuchFileException noSuchFileException) {
+                throw GrpcUtil.statusRuntimeException(Code.FAILED_PRECONDITION, "File does not exist");
+            }
```
nbauernfeind, line 117: Should we treat this as a non-export to prevent too many concurrent callers? The same comment applies to all of the gRPCs that interact with the file system.

server/src/main/java/io/deephaven/server/notebook/FilesystemStorageServiceGrpcImpl.java
```diff
@@ -0,0 +1,188 @@
...
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.util.Optional;
+import java.util.regex.PatternSyntaxException;
+import java.util.stream.Stream;
+
+@Singleton
+public class FilesystemStorageServiceGrpcImpl extends StorageServiceGrpc.StorageServiceImplBase {
```
nbauernfeind, line 43: I feel that this impl should at least ensure that there exists a session for each of these calls. The `getSession` method already throws a StatusRuntimeException; it's sufficient to fetch the session and not use it.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
niloc132: Yeah, I'm thinking about something like a user uploading/downloading a 10gb parquet file to the server

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
```diff
@@ -0,0 +1,89 @@
...
+  string path = 1;
+  FileKind kind = 2;
+  sint64 size = 3 [jstype=JS_STRING];
+}
+
+message ListItemsResponse {
+  repeated FileInfo items = 1;
+}
+
+//message ByteRange {
```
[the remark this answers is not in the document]
niloc132: Yes, just want it here for thought/discussion, will remove if we're confident we don't want it yet.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
```diff
@@ -0,0 +1,89 @@
...
+message SaveFileRequest {
+  // True to force new file creation and prevent overwriting an existing file
+  bool new_file = 1;
+  string path = 2;
+  bytes contents = 3;
+}
+message SaveFileResponse {
+}
+
+message MoveItemRequest {
```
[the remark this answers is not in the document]
niloc132: To what end? To allow overwriting a file for example?

web/client-api/src/main/java/io/deephaven/web/client/api/storage/JsStorageService.java
```diff
@@ -0,0 +1,97 @@
...
+        return Callbacks.<DeleteItemResponse, Object>grpcUnaryPromise(c -> client().deleteItem(req, metadata(), c::apply))
+                .then(response -> Promise.resolve((Void)null));
+    }
+
+    @JsMethod
+    public Promise<Void> saveFile(String path, FileContents contents) {
+        return contents.arrayBuffer().then(ab -> {
+            SaveFileRequest req = new SaveFileRequest();
+            req.setContents(new Uint8Array(ab));
+            req.setPath(path);
```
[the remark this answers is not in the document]
niloc132: Probably, but the current JS API for webdav doesnt have that as far as I can tell, so I didn't plumb it here yet.

server/src/main/java/io/deephaven/server/notebook/FilesystemStorageServiceGrpcImpl.java
niloc132: Not following - are you referring to gRPC's guarantees of a single stream always having messages handled serially? If so, these are all unary calls, so should be safe to handle concurrently.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
devinrsmith: I do wonder - are we trying to bite off too general of use case by trying to imagine this as a service for large binary blobs? At least in the case of notebooks, the use case is more specifically targeted towards "support script/text files and hierarchical management of said files".

I think there is potential for confusion where this feature gets overly pigeon-holed as a disk-backed local-storage available to the server process. IE, your example - a user uploads a 10GB parquet file to the *service* and now is confused why it's not available on the local disk.

In which case, we may need to have some sort support for resolving _StorageService_ URIs - something that may be local, or may be remote. But then we may run into a lot of NFS-like issues, where we ask "why is opening a parquet file so slow?"

Note: I'm not against the binary blob use case, but I think we need to think off all that it might entail.

devinrsmith: Depending on the layers of abstraction, this sort of issue might already be present - for example, remote docker or kubernetes volumes masquerading as disk-backed storage.

There *may* be room for us to establish a convention where werecommend setting up some (or all) of the docker server volumes as `StorageService`. This might also mean we should have ways of starting *multiple* `StorageService` for the same deployment.

devinrsmith (commented):
Initial comments - posting now b/c I've context switched, and want to make sure my pending comments don't get lost.

I'll have another review where I go over more code (instead of just .proto), as well as checking out and running.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
niloc132: Biting off too much: Agreed, I don't want to solve every possible problem, but I do want to at least have a plan on how we might solve other issues, to avoid changing the api later (for example, making FetchFile server-streaming or SaveFile client-streaming). 

Multi-service: this is an explicit design goal for this and the config service, though the discovery mechanism isn't defined yet. gRPC itself disallows (well, technically "strongly discourages" prefixing URLs with additional path segments, but we are effectively free to add other request metadata or fields to the message to clarify "which" service a client is talking to. The `path` field may itself an effective way to communicate which sets of storage are available (by convention) in a single server, but with any kind of load balancing or discovery (even as simple as just different hosts/ports, etc) we can easily build on this simple service to offer customized setups.

My initial suggestion is to keep this and the config service light and flexible, and encourage requesting configuration first to establish specific path prefixes that are interesting - such as "where are notebooks", "where are layouts".

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
```diff
@@ -0,0 +1,89 @@
...
+//message ByteRange {
+//  sint64 start = 1 [jstype=JS_STRING];
+//  sint64 end = 2 [jstype=JS_STRING];
+//}
+
+
+message FetchFileRequest {
+  string path = 1;
+//  ByteRange range = 2;//helpful? do we need this to save too, or have a truncate/append?
+}
```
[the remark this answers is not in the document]
niloc132: Some kind of a hash or last-modified could be interesting, both here and in the FileInfo message? Any thoughts on being able to use that right away @mofojed, or does this seem like a feature we could reasonably defer?

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
```diff
@@ -0,0 +1,89 @@
...
+message FetchFileResponse {
+  bytes contents = 1;
+}
+
+message SaveFileRequest {
+  // True to force new file creation and prevent overwriting an existing file
+  bool new_file = 1;
+  string path = 2;
+  bytes contents = 3;
+}
```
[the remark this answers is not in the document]
niloc132: Right, we would need "append"/"truncate"/"replace" to cleanly be able to apply arbitrary patches to files - likely combined with some assertion on contents (via etag?) and or maybe batch changes?

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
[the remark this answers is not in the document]
niloc132: @devinrsmith (tagging because your comment isn't reply-able and doesnt appear in the "thread") it does already throw, see the implementation's catch for FileAlreadyExistsException.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
```diff
@@ -0,0 +1,89 @@
...
+  string old_path = 1;
+  string new_path = 2;
+}
+
+message MoveItemResponse {
+}
+
+message CreateDirectoryRequest {
+  string path = 1;
+}
```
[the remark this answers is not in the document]
niloc132: Also already fails if something exists with that name (and also if the base path doesn't exist). Does not create directories recursively (that is harder to handle atomic "fail if it already exists" in a uniform way).

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
```diff
@@ -0,0 +1,89 @@
...
+
+enum FileKind {
+  UNKNOWN = 0;// Should not be used, exists only to indicate that this was unset
+  DIRECTORY = 1;
+  FILE = 2;
+}
+
+message FileInfo {
+  string path = 1;
```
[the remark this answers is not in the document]
niloc132: Can you elaborate? Are you suggesting a filesystem that we would interact with using non-unicode paths, i.e. either "ascii-only" or "there exist valid bytes that can be in a path that are not valid utf8"?

I don't think either case makes sense, at least from the perspective of "the server is written in Java, and Java's filesystem APIs all use Strings, which implies unicode". A quick read of sun.nio.fs.UnixPath shows that while paths are encoded as bytes, null bytes are disallowed, and a specific charset (provided by system property `sun.jnu.encoding`) will be used to do the actual encoding/decoding. https://stackoverflow.com/a/63688076/860630 suggests that this is "the default charset" on linux at least, and as of recent java versions, that means utf-8 (as of java 18).

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
```diff
@@ -0,0 +1,89 @@
...
+  rpc ListItems(ListItemsRequest) returns (ListItemsResponse) {}
+  rpc FetchFile(FetchFileRequest) returns (FetchFileResponse) {}
+  // Can create new files or modify existing
+  rpc SaveFile(SaveFileRequest) returns (SaveFileResponse) {}
+  rpc MoveItem(MoveItemRequest) returns (MoveItemResponse) {}
+  rpc CreateDirectory(CreateDirectoryRequest) returns (CreateDirectoryResponse) {}
+  rpc DeleteItem(DeleteItemRequest) returns (DeleteItemResponse) {}
+}
+
+message ListItemsRequest {
```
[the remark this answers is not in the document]
niloc132: @mofojed will defer to you

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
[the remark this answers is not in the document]
devinrsmith: The "there exist valid bytes that can be in a path that are not valid utf8" part - technically, Unix-like OSes, paths are *bytes*. It's a fact that I think a lot of APIs overlook. https://jod.al/2019/12/10/pathlib-and-paths-with-arbitrary-bytes/

margaretkennedy:
https://github.com/deephaven/deephaven.io/issues/1718

mofojed:
A login method was added, so can run this as one snippet and it works:
```
const client = new dh.CoreClient("http://localhost:10000");
await client.login(null)
const store = client.getStorageService();
await store.listItems("", "*.txt").then(console.dir);
```

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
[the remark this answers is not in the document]
mofojed: I think it would be nice to have, but yea can be deferred.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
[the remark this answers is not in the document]
mofojed: Shouldn't need it, this also isn't in the WebDAV spec.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
```diff
@@ -0,0 +1,89 @@
...
+
+option java_multiple_files = true;
+option optimize_for = SPEED;
+option go_package = "github.com/deephaven/deephaven-core/go/internal/proto/storage";
+
+/*
+ * Shared storage management service.
+*/
+service StorageService {
+  rpc ListItems(ListItemsRequest) returns (ListItemsResponse) {}
```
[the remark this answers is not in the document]
mofojed: I don't think it matters too much - we're not paginating, so can just sort client if we need.

web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
```diff
@@ -0,0 +1,85 @@
...
+    public static final String EVENT_CONNECT = "connect",
+            EVENT_DISCONNECT = "disconnect",
+            EVENT_RECONNECT = "reconnect",
+            EVENT_RECONNECT_AUTH_FAILED = "reconnectauthfailed",
+            EVENT_CONFIG_ADDED = "configadded",
+            EVENT_CONFIG_REMOVED = "configremoved",
+            EVENT_CONFIG_UPDATED = "configupdated",
```
mofojed, line 15: I don't think we need these?

web/client-api/src/main/java/io/deephaven/web/client/api/storage/ItemDetails.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @JsProperty
+    public String getFilename() {
+        return path.substring(path.lastIndexOf('/'));
+    }
+
+    @JsProperty
+    public String getBasename() {
+        return path.substring(0, path.lastIndexOf('/'));
```
mofojed, line 41: This is incorrect - basename should be the just the last part of the path: https://en.wikipedia.org/wiki/Basename

I've included an example from the webdav impl below of what it returns. Some things to note that have come up elsewhere in this review:
- Filename is the full path
- Basename is just the last part of the path after the last slash (as noted on the wikipedia page linked above)
- No separate `path` property
- `type` for directory vs file
- It is not sorted, the sorting happens client side
- There's a spot for etag, but it's obviously `null` so not doing anything
- lastmod, mime, are interesting bits of detail, the Web UI does not use them though

```
[
  {
    "filename": "/bar",
    "basename": "bar",
    "lastmod": "Mon, 04 Oct 2021 22:43:08 GMT",
    "size": 0,
    "type": "directory",
    "etag": null
  },
  {
    "filename": "/images",
    "basename": "images",
    "lastmod": "Wed, 06 Oct 2021 19:33:51 GMT",
    "size": 0,
    "type": "directory",
    "etag": null
  },
  {
    "filename": "/linktest.md",
    "basename": "linktest.md",
    "lastmod": "Fri, 08 Apr 2022 13:01:59 GMT",
    "size": 192,
    "type": "file",
    "etag": null,
    "mime": ""
  },
  {
    "filename": "/ticking-size.py",
    "basename": "ticking-size.py",
    "lastmod": "Mon, 27 Jun 2022 14:19:22 GMT",
    "size": 1844,
    "type": "file",
    "etag": null,
    "mime": ""
  },
  {
    "filename": "/pete-examples",
    "basename": "pete-examples",
    "lastmod": "Mon, 04 Oct 2021 23:02:17 GMT",
    "size": 0,
    "type": "directory",
    "etag": null
  },
  {
    "filename": "/01 Basic Streams.md",
    "basename": "01 Basic Streams.md",
    "lastmod": "Thu, 07 Oct 2021 15:24:38 GMT",
    "size": 7135,
    "type": "file",
    "etag": null,
    "mime": ""
  },
  {
    "filename": "/introduction.md",
    "basename": "introduction.md",
    "lastmod": "Wed, 06 Oct 2021 21:04:08 GMT",
    "size": 3254,
    "type": "file",
    "etag": null,
    "mime": ""
  },
  {
    "filename": "/treemap",
    "basename": "treemap",
    "lastmod": "Fri, 09 Sep 2022 14:41:51 GMT",
    "size": 0,
    "type": "directory",
    "etag": null
  },
  {
    "filename": "/testtty.py",
    "basename": "testtty.py",
    "lastmod": "Wed, 13 Oct 2021 14:30:44 GMT",
    "size": 14,
    "type": "file",
    "etag": null,
    "mime": ""
  },
  {
    "filename": "/02 Stream and Batch Together.md",
    "basename": "02 Stream and Batch Together.md",
    "lastmod": "Fri, 29 Oct 2021 15:01:42 GMT",
    "size": 4981,
    "type": "file",
    "etag": null,
    "mime": ""
  },
  {
    "filename": "/03 Kafka Stream vs Append.md",
    "basename": "03 Kafka Stream vs Append.md",
    "lastmod": "Fri, 08 Oct 2021 19:23:48 GMT",
    "size": 6886,
    "type": "file",
    "etag": null,
    "mime": ""
  },
  {
    "filename": "/examples",
    "basename": "examples",
    "lastmod": "Fri, 10 Sep 2021 15:30:18 GMT",
    "size": 0,
    "type": "directory",
    "etag": null
  },
  {
    "filename": "/RandomValues.py",
    "basename": "RandomValues.py",
    "lastmod": "Fri, 02 Sep 2022 19:06:59 GMT",
    "size": 2212,
    "type": "file",
    "etag": null,
    "mime": ""
  }
]
```

web/client-api/src/main/java/io/deephaven/web/client/api/storage/ItemDetails.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @JsProperty
+    public String getKind() {
+        return kind == Storage_pb.FileKind.getDIRECTORY() ? "directory" : "file";
+    }
+
+    @JsProperty
+    public Double getSize() {
+        return size == null ? null : Double.parseDouble(size);
```
mofojed, line 56: I think we discussed this before, but why would size be `null`?

web/client-api/src/main/java/io/deephaven/web/client/api/storage/ItemDetails.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @JsProperty
+    public String getPath() {
+        return path;
+    }
+
+    @JsProperty
+    public String getKind() {
+        return kind == Storage_pb.FileKind.getDIRECTORY() ? "directory" : "file";
+    }
```
mofojed, line 52: Would prefer `type` to `kind`, but not a big deal.

web/client-api/src/main/java/io/deephaven/web/client/api/storage/ItemDetails.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public static ItemDetails fromProto(FileInfo item) {
+        return new ItemDetails(item.getPath(), item.getKind(), item.getSize());
+    }
+
+    @JsProperty
+    public String getFilename() {
+        return path.substring(path.lastIndexOf('/'));
+    }
```
mofojed, line 37: This is actually the basename

mofojed (changes requested):
I didn't play around with it too extensively/didn't try to wire it up to the UI at all yet. Main thing to change is `getBasename` is incorrect, and we should just be consistent about path vs. basename.

I still wonder if this is worth doing over keeping WebDAV. It's another thing to maintain, and it loses any interoperability with other UIs (e.g. macOS finder). As mentioned on [my comment](https://github.com/deephaven/deephaven-core/issues/1768#issuecomment-1253899083) on the initial ticket, there does still appear to be active WebDAV servlet implementations, and it seems like it'd be easier to use those rather than create our own protocol.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
devinrsmith: With a forward looking, I'd be in favor of a streaming response here - I don't think we need any additional metadata on the response. The contract being for now that you append all the bytes together to get the full file.

server/src/main/java/io/deephaven/server/notebook/FilesystemStorageServiceGrpcImpl.java
```diff
@@ -0,0 +1,210 @@
...
+
+    @Override
+    public void listItems(ListItemsRequest request, StreamObserver<ListItemsResponse> responseObserver) {
+        GrpcUtil.rpcWrapper(log, responseObserver, () -> {
+            sessionService.getCurrentSession();
+
+            ListItemsResponse.Builder builder = ListItemsResponse.newBuilder();
+            PathMatcher matcher = request.hasFilterGlob() ? makeMatcher(request.getFilterGlob()) : ignore -> true;
+            Path dir = resolveOrThrow(request.getPath());
+            try (Stream<Path> list = Files.list(dir)) {
```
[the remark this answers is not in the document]
devinrsmith: Also, we may not need to make sure the glob is safe w/ newDirectoryStream - I think it only resolves in that directory (but not 100% sure).

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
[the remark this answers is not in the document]
niloc132: Technically correct, but is it a useful distinction? Are there appropriate java apis to interact with the file system in this way?

server/src/main/java/io/deephaven/server/notebook/FilesystemStorageServiceGrpcImpl.java
```diff
@@ -0,0 +1,210 @@
...
+                for (Path p : (Iterable<Path>) list::iterator) {
+                    if (!matcher.matches(dir.relativize(p))) {
+                        continue;
+                    }
+                    boolean isDirectory = Files.isDirectory(p);
+                    builder.addItems(FileInfo.newBuilder()
+                            .setPath(p.getFileName().toString())
+                            .setSize(isDirectory ? 0 : Files.size(p))
+                            .setKind(isDirectory ? FileKind.DIRECTORY : FileKind.FILE)
```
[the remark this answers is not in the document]
niloc132: My initial thinking here was that the clients should (at least to start with) be agnostic of the implementation - if the admin decides to move a directory and link it with a symlink, the client shouldn't have to work out how to ask "okay, this is a symlink, what is the type of the target, resolve the target", but should be able to just treat it like a "file" or "directory".

That said, I left this as an enum so that we could add other file types like that, but wanted to be sure we had a nice way to convey all of their metadata... without attempting to design a general purpose filesystem api.

server/src/main/java/io/deephaven/server/notebook/FilesystemStorageServiceGrpcImpl.java
```diff
@@ -0,0 +1,210 @@
...
+    @Override
+    public void moveItem(MoveItemRequest request, StreamObserver<MoveItemResponse> responseObserver) {
+        GrpcUtil.rpcWrapper(log, responseObserver, () -> {
+            sessionService.getCurrentSession();
+
+            Path source = resolveOrThrow(request.getOldPath());
+            Path target = resolveOrThrow(request.getNewPath());
+
+            try {
+                Files.move(source, target);
```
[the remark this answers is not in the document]
niloc132: Will test to confirm, but the goal here is to _not_ support that.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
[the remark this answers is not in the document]
devinrsmith: > but a useful distinction?

"maybe"?

I am being pedantic here. I'm not too concerned with the java APIs - if we move path to `bytes`, we'll assert UTF-8. I'm thinking about implementations beyond java and potential breaking changes we would need otherwise in the future. I know some languages expose byte-oriented path apis.

This was actually from somebody I know from college: https://beets.io/blog/paths.html

server/src/main/java/io/deephaven/server/notebook/FilesystemStorageServiceGrpcImpl.java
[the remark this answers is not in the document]
niloc132: Yes, that saves a line here, but it still has to be closed.

We want the glob to only resolve in that directory, so that would be good. I wouldnt mind sending an error back still if the user attempted to do a glob that doesnt actually make sense.

server/src/main/java/io/deephaven/server/notebook/FilesystemStorageServiceGrpcImpl.java
```diff
@@ -0,0 +1,210 @@
...
+        if (resolved.startsWith(root)) {
+            return resolved;
+        }
+        throw GrpcUtil.statusRuntimeException(Code.INVALID_ARGUMENT, "Invalid path: " + relativePath);
+    }
+
+    @Override
+    public void listItems(ListItemsRequest request, StreamObserver<ListItemsResponse> responseObserver) {
+        GrpcUtil.rpcWrapper(log, responseObserver, () -> {
+            sessionService.getCurrentSession();
```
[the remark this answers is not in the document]
niloc132: At this time, that's all we can do to assert that the user has a session (i.e. the method will throw an appropriate error if there is no session). With the  authorization #2563 work, we'll have more tools, but those havent been finalized yet.

niloc132:
> I still wonder if this is worth doing over keeping WebDAV. It's another thing to maintain, and it loses any interoperability with other UIs (e.g. macOS finder). As mentioned on https://github.com/deephaven/deephaven-core/issues/1768#issuecomment-1253899083 on the initial ticket, there does still appear to be active WebDAV servlet implementations, and it seems like it'd be easier to use those rather than create our own protocol.

The linked impl from jcr/jackrabbit is for javax.servlet, which we do not support (only the oldest presently supported version of jetty (10.x) still has support for it, and it isn't clear to me how long they will keep that (12.x is alpha now). We could fork and modernize it, but I suspect that will take substantially more work to maintain than what we have here.

The impl in tomcat at least supports `jakarta.servlet` (the replacement for `javax.servlet` as of 3 years ago), but depends on other tomcat internals (for a start, `DefaultServlet`, the supertype of the linked class). We instead use Jetty, another implementation of the servlet spec, and tomcat's DefaultServlet (and other classes, like WebResource that this depends on) won't be compatible with jetty - and tomcat has worse support for HTTP2, which our gRPC wiring depends on.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
niloc132: A streaming "fetch" requires a matching multi-call "save", or you can only download and not upload. Make save client-upload, and wire up our fake streaming upload feature for gRPC-web?

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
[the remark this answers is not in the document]
niloc132: For the moment, I propose a) requiring a delete before a move, and b) if it becomes an issue we can add a new boolean flag to the message to enable overwriting?

Or does the UI already expect overwriting to work @mofojed ?

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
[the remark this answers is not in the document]
niloc132: To be clear, you're suggesting lists of paths with a total byte size of greater than our default of 100mb? or even grpc's much more pessimistic default of 4mb?

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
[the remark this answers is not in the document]
mofojed: Overwriting is allowed with the WebDAV impl. A flag would be good or we can make the client do a delete/write if the initial write fails or something.

web/client-api/src/main/java/io/deephaven/web/client/api/storage/ItemDetails.java
niloc132: Changed to zero based on the above summary

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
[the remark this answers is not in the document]
devinrsmith: It's not just about total size - it's potentially a model that would work well for "slow" or async backing storage where they may be already producing batched lists of some sort.

For example, an S3 backed filesystem; your list might produce a lot of paths in some sort of paginated fashion, even though I doubt the total amount breaks 100MB.

server/src/main/java/io/deephaven/server/notebook/FilesystemStorageServiceGrpcImpl.java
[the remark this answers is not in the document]
niloc132: On second thought, to get good parse errors, i'm going to split the difference a bit, and keep parsing the pattern in our own code, but pass it to newDirectoryStream. It does still save us a matcher check and a method reference.

niloc132:
After implementing etags, I've come to the conclusion that streaming isn't going to make sense in the general case - extra work for the server and the client to get along. Since the primary use case is smallish files (notebooks and layouts), we're going to punt on streaming, and intend to introduce specific support for large files and streaming their contents when we have use cases to support it.

The basic issue is that any etag send with a file's contents must match that file's contents. That implies either some kind of immutable filesystem (files can be "edited" by replacing them), or reading the entire contents from disk before computing the etag, so that they can be sent consistently. In that case, streaming doesn't help a lot, since the server is holding the entire file in memory (for a typical filesystem) for at least a moment while hashing it and sending it in pieces.

Given the choice, I think etags are more important than large files. This next patch uses 123bit murmur3 from guava as a "fast, non-cryptographic hash", to provide hints as to when the file changes, but not to defend specifically against any form of collisions. The patch is somewhat naive, hoping for group feedback on how to tune it - for example rather than hashing every file in a directory, perhaps it should be limited to only hash the first N bytes, possibly first sorting files by size, ascending.

Then, in a later revision we might add support for reading and writing large files, as needed.

web/client-api/src/main/java/io/deephaven/web/client/api/storage/ItemDetails.java
niloc132: I think this is all resolved, but double check?

web/client-api/src/main/java/io/deephaven/web/client/api/LoginCredentials.java
niloc132: Actually slightly necessary now, so login() can be used to wait for a connection.

server/src/main/java/io/deephaven/server/notebook/FilesystemStorageServiceGrpcImpl.java
```diff
@@ -0,0 +1,251 @@
...
+            try (DirectoryStream<Path> list = Files.newDirectoryStream(dir, matcher)) {
+                for (Path p : list) {
+                    BasicFileAttributes attrs = Files.readAttributes(p, BasicFileAttributes.class);
+                    boolean isDirectory = attrs.isDirectory();
+                    ItemInfo.Builder info = ItemInfo.newBuilder().setPath(p.getFileName().toString());
+                    if (isDirectory) {
+                        info.setType(ItemType.DIRECTORY);
+                    } else {
+                        info.setSize(attrs.size())
+                                .setEtag(hash(p))// Note, there is a potential race here between the size and the hash
```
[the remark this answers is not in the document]
niloc132: My hunch is no, better to give the server a decent heuristic for deciding to send them in a given list call. HTTP servers tend to send them only for the file you're requesting.

server/src/main/java/io/deephaven/server/notebook/FilesystemStorageServiceGrpcImpl.java
```diff
@@ -0,0 +1,251 @@
...
+    private static final String STORAGE_PATH =
+            Configuration.getInstance().getStringWithDefault("storage.path", "<workspace>/storage")
+                    .replace("<workspace>", Configuration.getInstance().getWorkspacePath());
+
+    /**
+     * Non-cryptographic hash, not resistant to adversarial collisions, but should suffice for quickly checking for
+     * edits to files. By design, hashes might not be stable from one startup to the next, but will at least efficiently
+     * guard against client's needing to load the same file again rapidly.
+     */
+    private static final HashFunction HASH_FUNCTION = Hashing.goodFastHash(128);
```
devinrsmith, line 68: Discussed on phone, I'd be in favor of `com.google.common.hash.Hashing#murmur3_128(int)` with a fixed seed. I think we should still have the comment that "the client should not *rely* on this implementation and it _may_ change from run to run"

mofojed (dismissed):

server/src/main/java/io/deephaven/server/notebook/FilesystemStorageServiceGrpcImpl.java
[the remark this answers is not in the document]
niloc132: Nope, this doesn't work - newDirectoryStream will not pass the actual item being streamed, but first resolved it against the parent. Technically we could probably prefix all globs with the parent path (must be the same absolute vs relative as the stream itself will use) or prefix all globs with `**`, but it seems cleaner to just match patterns as they were written, at the cost of having to make an iterable.

web/client-api/src/main/java/io/deephaven/web/client/api/storage/JsStorageService.java
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * Moves (and/or renames) an item from its old path to its new path. The optional newFile parameter can be passed to
+     * enforce that an existing item must not be overwritten.
+     *
+     * Note that directories must be empty to be overwritten.
+     *
+     * @param oldPath the path of the existing item
+     * @param newPath the new path to move the item to
+     * @param newFile true to forbid an existing file be created, false to allow an existing file to be overwritten
```
mofojed, line 143: This comment wording doesn't make sense. This param/comment should match the `saveFile` method as well, though I don't think `newFile` is the best name. I'd probably change it to `allowOverwrite`, `disallowOverwrite`, or something like that instead.

devinrsmith (approved):
LGTM - one comment about javadoc, but not worth another patch unless you are already making other changes.

mofojed (changes requested):

web/client-api/src/main/java/io/deephaven/web/client/api/storage/JsItemDetails.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @JsProperty
+    public String getBasename() {
+        return path.substring(path.lastIndexOf('/'));
+    }
+
+    @JsProperty
+    public String getDirname() {
+        return path.substring(0, path.lastIndexOf('/'));
+    }
```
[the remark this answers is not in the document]
mofojed: Looks like the path for subdirectories isn't getting set properly either:
image

web/client-api/src/main/java/io/deephaven/web/client/api/storage/FileContents.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private Promise<Blob> contents() {
+        if (data != null) {
+            return Promise.resolve(data);
+        }
+        return Promise.reject("No contents available, please use provided etag");
+    }
+
+    @JsMethod
+    public Promise<String> text() {
```
mofojed, line 59: Any reason these need to be Promises? We've already got the data...

mofojed: I guess somewhat following the `File` interface?

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/storage.proto
```diff
@@ -0,0 +1,124 @@
...
+  // an existing file or creating a new one.
+  bool new_file = 1;
+  // The path to the file to write contents to
+  string path = 2;
+
+  // The contents to use when creating then file, or to use to replace the file.
+  bytes contents = 3;
+}
+message SaveFileResponse {
+}
```
devinrsmith, line 93: We should probably add etag to save file response.

web/client-api/src/main/java/io/deephaven/web/client/api/storage/FileContents.java
niloc132: I was aiming at Blob's api (FIle doesnt expose practically anything afaict), but yeah. It also lets us be consistent creating a FileContents and keeping it locally, if you happen to do that.

It might not be ideal if we end up needing to stream results later.

web/client-api/src/main/java/io/deephaven/web/client/api/storage/JsItemDetails.java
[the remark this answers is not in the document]
niloc132: Updated the server (and proto docs) to always use+validate a leading slash.

web/client-api/src/main/java/io/deephaven/web/client/api/storage/JsStorageService.java
niloc132: To avoid confusion from double negatives (disallowOverwrite=false), I would need to make this "default to safe" (since proto defaults to false, so allowOverwrite=false would mean that callers would need to specify that they want to overwrite files in both contexts.

server/jetty-app/storage/notebooks/sample.txt
```diff
@@ -0,0 +1 @@
+asdf
```
devinrsmith, line 1: oops

web/client-api/src/main/java/io/deephaven/web/client/api/storage/JsItemDetails.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @JsProperty
+    public String getFilename() {
+        return path;
+    }
+
+    @JsProperty
+    public String getBasename() {
+        return path.substring(path.lastIndexOf('/'));
```
mofojed, line 39: Don't want to include the `/` in the basename, should add +1
```suggestion
        return path.substring(path.lastIndexOf('/') + 1);
```

mofojed (changes requested):

devinrsmith (approved):

mofojed (approved):

PR #2845, opened 2022-09-13
TicketResolver: return failed export object instead of throwing

nbauernfeind:
The scope ticket resolver no longer throws an exception but instead returns an export object that is in a failed state. This allows commands downstream of a failed dependency to 1) be created, and 2) assume a dependency_failed state.

### DIFF
--- a/server/src/test/java/io/deephaven/server/session/SessionStateTest.java
+++ b/server/src/test/java/io/deephaven/server/session/SessionStateTest.java
@@ -6,9 +6,7 @@ package io.deephaven.server.session;
 import io.deephaven.base.verify.Assert;
 import io.deephaven.base.verify.AssertionFailure;
 import io.deephaven.engine.context.ExecutionContext;
-import io.deephaven.engine.util.NoLanguageDeephavenSession;
 import io.deephaven.proto.util.ExportTicketHelper;
-import io.deephaven.server.table.ExportTableUpdateListenerTest;
 import io.deephaven.time.DateTimeUtils;
 import io.deephaven.engine.liveness.LivenessArtifact;
 import io.deephaven.engine.liveness.LivenessReferent;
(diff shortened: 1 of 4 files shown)


server/src/test/java/io/deephaven/server/session/SessionStateTest.java
```diff
@@ -81,7 +79,7 @@ public void teardown() {
     public void testDestroyOnExportRelease() {
         final CountingLivenessReferent export = new CountingLivenessReferent();
         final SessionState.ExportObject<Object> exportObj;
-        try (final SafeCloseable scope = LivenessScopeStack.open()) {
+        try (final SafeCloseable ignored = LivenessScopeStack.open()) {
```
kosak, line 82: BTW, FYI, no action required, some PL people use the variable _ (that is, single underscore) to mean "dummy" or "ignored"

kosak (approved):
This looks good to me. (given my fairly limited knowledge of the server)
Thanks for diagnosing and working on this!!!!

server/src/test/java/io/deephaven/server/session/SessionStateTest.java
nbauernfeind: That's a neat fact. I'm not sure if that existed before Scala, but `_` is a very heavy scala ism. You can even use it as an unlabeled single parameter into a lambda (i.e. `_.convertToInt(radix)`). `_` is an invalid identifier in java, so we can't do the same. IntelliJ auto suppresses unused warnings when the name is `ignored`.

kosak: Ah, didn't realize _ was illegal in Java, thanks. I think I picked up the convention from Go and Rust (where _ is varying degrees of special).

niloc132: The reason that `_` is illegal is that this is coming in a future release, they wanted to reserve it to not break existing code.

https://mail.openjdk.org/pipermail/amber-spec-experts/2022-September/003485.html

As with other languages, you will be able to mark more than one local as `_`, and it looks like even type params will be able to be `_` (as opposed to `?`).

niloc132 (approved):

PR #2847, opened 2022-09-14
De-root nginx (and grpc-proxy)

JamesXNelson:
Also adds /ide routing in nginx, so we always serve index.html, but preserve full url so js can read the /ide/path.


docker/web/build.gradle
```diff
@@ -12,7 +12,7 @@ dependencies {
     js project(path: ':web', targetConfiguration: 'js')
 }
 
-evaluationDependsOn Docker.registryProject('nginx-base')
+evaluationDependsOn Docker.registryProject('nginx-noroot-base')
```
niloc132, line 15: do we have any references left to nginx-base, or is it safe to either remove nginx-base from the registry, or reuse nginx-base (and change its imageName etc) instead of adding a new proj to the registry?

docker/web/src/main/docker/Dockerfile
```diff
@@ -1,11 +1,29 @@
-FROM deephaven/nginx-base:local-build
+FROM deephaven/nginx-noroot-base:local-build
+
+# In order to run without root, we start as root, install sudo w/ sudo rules, then switch to nginx user (id 101)
+USER 0
+
+# We want to install sudo, and start the sudo service (update policy-rc.d to exit 0, instead of 101)
+RUN ls -la /usr/sbin/policy-rc.d ; \
```
niloc132, line 7: convention is to prefix any multi-line RUN commands with `set -eux` to avoid surprises when one command fails, then the `&&`s aren't required. Alternatively, move contents into a `.sh` file, and start that with `set -eux`, invoke that directly (though that involves copying it first, etc)

docker/web/build.gradle
JamesXNelson: If there are references left, I can remove them, however I thought "we probably have a bunch of documentation which includes the old version, we should leave removing it until after users have had a migration period of at least one release".

I can open a ticket for that?

docker/web/src/main/docker/Dockerfile
JamesXNelson: fixed

JamesXNelson:
For documentation, I'm thinking to just update anywhere we might reference `nginx` w/ `nginx-noroot`.... however, I'm not sure if that's even needed at all, since users are likely consuming the completed web image and won't care much about base images. Changing the port from 80 to 8080 tho... that is likely to break examples and consumers of docker images, so... not 100% sure how much docs will be needed.

grpc-proxy/src/main/common/docker-entrypoint.sh
```diff
@@ -2,10 +2,38 @@
...
+if [ -s "$TLS_CRT" ]; then
+
+    exec /app/dist/grpcwebproxy \
+      --run_http_server=false \
+      --server_http_debug_port="${PROXY_DEBUG_PORT}" \
+      --backend_addr="${BACKEND_ADDR}" \
+      --server_http_tls_port="${PROXY_TLS_PORT}" \
+      --server_tls_cert_file="$TLS_CRT" \
+      --server_tls_key_file="$TLS_KEY" \
+      --server_tls_client_ca_files="$TLS_CA,/etc/ssl/certs/ca-certificates.crt,/var/run/secrets/kubernetes.io/serviceaccount/ca.crt" \
```
devinrsmith, line 25: Is this kubernetes path going to break anybody who may want to do TLS on non-kube?

JamesXNelson: Nope; I don't _think_ so at least. It's just a list of places where there are CA files.

If you want, I can build the list and check if the files are readable first, that way it will work in kube but won't risk breaking someone w/o that ca cert.

docker/web/build.gradle
niloc132: I don't _think_ we consider those internal images as being part of our public api/artifacts, so removing all traces of them should be safe.

@devinrsmith can you confirm?

margaretkennedy:
https://github.com/deephaven/deephaven.io/issues/1717

docker/web/src/main/docker/nginx/default.conf
```diff
@@ -18,6 +18,13 @@ server {
         add_header Cache-Control 'must-revalidate, max-age=0';
     }
 
+    # Route /ide/whatever to serve ide/index.html but preserve user-supplied URL for js to process
+    location ~ ^/ide/ {
+        try_files $uri $uri/ /ide/index.html =404;
+        #expires -1;
+        #add_header Cache-Control 'must-revalidate, max-age=0';
```
mofojed, line 25: Don't think these should be commented out?

docker/web/build.gradle
mofojed: I'll let @devinrsmith but I'm in agreement, should trash nginx-base

JamesXNelson: The port number we use needs to be changed everywhere.  Code, documentation, any published works, etc. It's a lot of extra work :-/

docker/web/src/main/docker/nginx/default.conf
JamesXNelson: Seems to work w/ these uncommented out.

grpc-proxy/src/main/common/docker-entrypoint.sh
JamesXNelson: I checked if they exist before adding to the flag, so there's no way it can break non-kube users now.

docker/web/build.gradle
JamesXNelson: I looked in deephaven.io and can't see anywhere obvious that has docker-compose yaml to edit for the updated port...

@margaretkennedy any ideas?

JamesXNelson: I looked for `docker-compose` and `port` and found nothing to update

docker/web/src/main/docker/nginx/default.conf
JamesXNelson: Even w/ these uncommented tho, a page using this feature like https://jxn-test-changes.demo.deephaven.app/ide/notebook/03%20Kafka%20Stream%20vs%20Append.md still gives me `304` cached status...

docker/web/build.gradle
margaretkennedy: > I looked in deephaven.io and can't see anywhere obvious that has docker-compose yaml to edit for the updated port...
> 
> @margaretkennedy any ideas?

We don't include the actual yaml anywhere in the user guide. Sounds like any documentation updates would be in the file itself or in deephaven-core's README, if even necessary?

niloc132: Nothing references that port, it isn't exposed to users at all. Instead, only the envoy proxy talks to it, the only place to update is the envoy yaml (which I think you are doing anyway). Envoy acts as a reverse proxy in front of nginx, and there is no other way to connect to it.

JamesXNelson:
Nm... I see now the examples for the images are in the deephaven-core repo, and will need updates.  One more commit coming.

JamesXNelson:
ok, all pushed, conflicts resolved, should be g2g

niloc132 (approved):

niloc132:
Be sure to hit the "Re-request review" button for each reviewer when it is ready to go.

devinrsmith (approved):
Everything seems to run fine for me. That said, it's tough for me to say whether there will be unintended fallout wrt the web image. Good news is we are moving away from the web image as a construct.

PR #2848, opened 2022-09-14
Add HealthStatusManager and correctly enterTerminalState

JamesXNelson:


proto/proto-backplane-grpc/Dockerfile
```diff
@@ -24,7 +24,8 @@ RUN set -eux; \
     /includes/deephaven/proto/table.proto \
     /includes/deephaven/proto/application.proto \
     /includes/deephaven/proto/inputtable.proto \
-    /includes/deephaven/proto/partitionedtable.proto; \
+    /includes/deephaven/proto/partitionedtable.proto \
+    /includes/grpc/health/v1/health.proto; \
```
niloc132, line 28: why do we need to generate java/js/go/py for this file? If we include grpc-services, then the generated java at least should already be present?

proto/proto-backplane-grpc/src/main/proto/grpc/health/v1/health.proto
```diff
@@ -0,0 +1,63 @@
...
+  enum ServingStatus {
+    UNKNOWN = 0;
+    SERVING = 1;
+    NOT_SERVING = 2;
+    SERVICE_UNKNOWN = 3;  // Used only by the Watch method.
+  }
+  ServingStatus status = 1;
+}
+
+service Health {
```
niloc132, line 42: this appears to be the same contents as the file in grpc-services.jar, why not read the file out of there if we do need to regenerate it?

server/src/main/java/io/deephaven/server/runner/DeephavenApiServerInProcessGroovyComponent.java
```diff
@@ -21,6 +22,7 @@
         GroovyConsoleSessionModule.class,
         ServerBuilderInProcessModule.class,
         SessionToExecutionStateModule.class,
+        HealthCheckModule.class,
```
niloc132, line 25: it seems irritating to make this a dependency of tests in this way - is it because the server now has a dependency on the health check module instead of vice versa?

JamesXNelson: yes. The server needs to manipulate the HealthStatusManager in it's shutdown callback, so it now has a dependency on the module.

Maybe there's some way to make this slightly cleaner, but this is what I did which works.

proto/proto-backplane-grpc/src/main/proto/grpc/health/v1/health.proto
JamesXNelson: um, I'll see if I can just use the jar and skip the extra protos.

It was this proto in particular that I had wanted to pull out for grpcurl to work, but I already have some "extract all proto from all jars" code, so it shouldn't matter where it comes from.

proto/proto-backplane-grpc/Dockerfile
JamesXNelson: I'll try it again w/o the extra proto...

proto/proto-backplane-grpc/src/main/proto/grpc/health/v1/health.proto
JamesXNelson: confirmed, I can safely remove the proto and the dockerfile changes.

grpc-java/grpc-mtls/build.gradle
```diff
@@ -7,4 +7,5 @@ description = 'Utility project to make x509 client cert available in grpc calls
 dependencies {
     Classpaths.inheritGrpcPlatform(project)
     implementation 'io.grpc:grpc-core'
+    implementation 'io.grpc:grpc-services'
```
devinrsmith, line 10: None of the grpc-services dependencies added seem necessary. I've removed all of them and things still compile and run. server has an api dependency on them w/ the comment:

```
    api('io.grpc:grpc-services') {
        because 'downstream dagger compile, see deephaven-core#1722'
    }
```

What's the reason you've added them as deps?

nbauernfeind:
I've +1'd Devin's question, but otherwise LGTM.

grpc-java/grpc-mtls/build.gradle
JamesXNelson: Huh, IDE had been complaining about not being able to find HealthStatusManager, but if you removed them and it's happy, I'll remove them too.

JamesXNelson: Removed them, and it still works.

JamesXNelson:
All issues are addressed, I believe.

I removed the extra proto and dependencies and it still works fine.

Ready to merge w/ approval.

devinrsmith (changes requested):
I don't think you've pushed your latest? Also, merge conflicts.

JamesXNelson:
resolved conflicts (twice now)

devinrsmith (changes requested):
Looks like DeephavenApiServerInProcessGroovyComponent was added as an empty file.

niloc132 (approved):

devinrsmith (approved):

PR #2850, opened 2022-09-14
Make version runtime configurable

JamesXNelson:


buildSrc/src/main/groovy/Docker.groovy
```diff
@@ -164,6 +164,11 @@ class Docker {
          */
         String platform;
 
+        /**
+         * Optional set of environment variables to set on the container.
+         */
+        Map<String, String> envVars;
```
niloc132, line 170: this seems unused/unrelated?

server/netty/src/main/java/io/deephaven/server/netty/NettyServerModule.java
```diff
@@ -49,7 +52,11 @@ static GrpcServer serverBuilder(
         interceptors.forEach(serverBuilder::intercept);
         serverBuilder.intercept(MTlsCertificate.DEFAULT_INTERCEPTOR);
         serverBuilder.maxInboundMessageSize(serverConfig.maxInboundMessageSize());
+        Logger log = LoggerFactory.getLogger(NettyServerModule.class);
         if (serverConfig.ssl().isPresent()) {
+            if (log.isInfoEnabled()) {
```
niloc132, line 57: while not a bad thing, this seems unused/unrelated to version being runtime configurable?

py/server/deephaven/__init__.py
```diff
@@ -7,7 +7,9 @@
 
 """
 
-__version__ = "0.17.0"
+import os
+
+__version__ = os.environ.get('DEEPHAVEN_VERSION') or "0.17.0"
```
niloc132, line 12: doesn't this run the risk of installing version X, but having DEEPHAVEN_VERSION=Y set, and so the library reports the wrong version when asked?

niloc132 (commented):
Unless I misunderstand, only the

buildSrc/src/main/groovy/Docker.groovy
JamesXNelson: It was used for passing ENV vars into Dockerfiles, but I no longer needed it b/c I'm building the demo images in a different repo. 

I can remove it if you guys don't want to leave the plumbing there.

py/server/deephaven/__init__.py
JamesXNelson: That's actually kinda the point. When the deploy tool pushes a deployment, it should report the version number used to deploy the system.  For the main demo controller, this lets us bump the versions w/o touching main versions. For single machine deployments used to test, we also stick `_branch-name` into the version as a suffix for we can tell where each machine came from, and so `print(deephaven.__version__)` will give the more useful value.

server/netty/src/main/java/io/deephaven/server/netty/NettyServerModule.java
JamesXNelson: Yes, this was "make version runtime configurable + all other extras from demo branch".

I'd be happy to remove any part of this you don't like, it's just the catch all "here's little things I saw along the way".

py/server/deephaven/__init__.py
JamesXNelson: I set it to NoDocumentationNeeded... we want it to be possible, but not really advertised. imho.

server/netty/src/main/java/io/deephaven/server/netty/NettyServerModule.java
devinrsmith: We *can* do this, but I think we'll want to update the SSLConfig hierarchy to annotate some fields as `org.immutables.value.Value.Redacted`.

devinrsmith (changes requested):
Happy to have a small catch all PR for extra stuff. Nothing offensive.

py/server/deephaven/__init__.py
niloc132: It seems like we should move the logic to the `print()` statement if that's the purpose:

```
print(os.environ.get('DEEPHAVEN_VERSION') or deephaven.__version__)
```

jmao-denver: I am with Colin,  would prefer not to have the built-in package version number contingent on an env var.

JamesXNelson: I can't tell the user running code in console to check the env var. That defeats the purpose.

The idea is that if I build some stuff on top of a deephaven base version, I want to tell user they have built a custom version and not the packaged version the code was based on.

JamesXNelson: When I deploy a machine / cluster, I _DO_ set a custom version for all the other artifacts and those versions are respected, but I can't touch this one b/c it's hardcoded.  Unless we want to template / find replace the hardcoded version when building the code, I have no other way to indicate "this is not the version you actually built".

w/o this admittedly hacky method of injecting a version number, then I _can't_ update the version the console reports without manually editing this file.  Everything else I can route through a gradle property, but there's nothing else I can think of here (b/c editing the file before building wheel is super-duper hacky).

server/netty/src/main/java/io/deephaven/server/netty/NettyServerModule.java
JamesXNelson: what needs to be redacted?
I found three password fields, I think I redacted correctly...

py/server/deephaven/__init__.py
niloc132: So, the goal is to be able to detect both a) the version of the `deephaven-core` wheel and b) the version of the demo installation? As there is no actual `deephaven-core` wheel with version x.y.z_BRANCH_NAME, wouldn't it make more sense to actually want them to ask specifically for the branch name, rather than have the version reflect a version that doesn't exist?

If this isn't something we have scripted, but something we ask the user to do, can't we introduce a non-standard property/def instead? something like 
```
deephaven.printVersion()
```
which could then wrap up the custom logic you're after, and potentially also log details like the jvm version, etc?

server/netty/src/main/java/io/deephaven/server/netty/NettyServerModule.java
devinrsmith: lgtm

py/server/deephaven/__init__.py
JamesXNelson: I added `deephaven.print_version()` since I assume that's "more pythonic"

buildSrc/src/main/groovy/Docker.groovy
JamesXNelson: Do you want me to remove this or not? If someone ever does want to use it, the wiring is there now.  I don't actually care, just want a decision on stay-or-go.

PR #2852, opened 2022-09-14
Update `PartitionTable` to use `RowSetBuilder` instead of `RowSet.insert()`

lbooker42:
Overview:
Use a sequential builder during the creation stage of the PartitionTable since we can rely on ordered keys, transition to using random builder on following updates.

Initial tests show a 30-90% speedup in some cases (data to follow)
```
Rows	        Buckets	    partitionBy (ms)	partitionBy_opt (ms)	speedup
100,000,000	100	        1,454	            2,253	                -54.9%
100,000,000	25,000	    95,338	            9,083	                90.5%
100,000,000	1,000,000	30,935	            19,952	                35.5%
				
				
Rows	        Buckets	    partitionBy (ms)	partitionBy_opt (ms)	speedup
100,000,000	100	        1,500	            2,346	                -56.4%
100,000,000	25,000	    87,767	            8,952	                89.8%
100,000,000	1,000,000	30,993	            19,811	                36.1%
```

The only regression is in the very large bucket (1M items per bucket) where direct `RowSet.insert()` is measurably faster.  The other cases greatly benefit from these changes however and the overall change is beneficial.

NOTE: It's likely that `groupBy()` can also benefit from these changes.


engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -471,23 +554,25 @@ public void propagateInitialState(@NotNull final QueryTable resultTable) {
             final boolean setCallSite = QueryPerformanceRecorder.setCallsite(callSite);
             try (final ResettableWritableObjectChunk<QueryTable, Values> tablesResettableChunk =
                     ResettableWritableObjectChunk.makeResettableChunk();
-                    final ResettableWritableObjectChunk<WritableRowSet, Values> addedRowSetsResettableChunk =
-                            ResettableWritableObjectChunk.makeResettableChunk();
+                    final ResettableWritableObjectChunk<RowSetBuilderSequential, Values> addedRowSetsResettableChunk =
```
lbooker42, line 557: This test for `initialized` is useless (always false)

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -58,11 +53,19 @@ public final class PartitionByChunkedOperator implements IterativeChunkedAggrega
...
     /**
-     * Sentinel value for the row set belonging to a table that was never created because either the result table was no
-     * longer live or the aggregation update listener was no longer live. Should be used for assignment and reference
-     * equality tests, only.
+     * Sentinel value for the row set builders belonging to a table that was never created because either the result
+     * table was no longer live or the aggregation update listener was no longer live. Should be used for assignment and
+     * reference equality tests, only.
+     */
+    private static final RowSetBuilderRandom NONEXISTENT_TABLE_ROW_SET_BUILDER_RANDOM = RowSetFactory.builderRandom();
+    private static final RowSetBuilderSequential NONEXISTENT_TABLE_ROW_SET_BUILDER_SEQUENTIAL =
```
lbooker42, line 61: `NONEXISTENT_TABLE_ROW_SET_BUILDER_SEQUENTIAL` is never assigned and not needed.  Can remove it and all tests for it.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -367,39 +393,86 @@ public boolean modifyRowKeys(final SingletonContext context, @NotNull final Long
...
     }
 
-    private static void accumulateToRowSet(@NotNull final ObjectArraySource<WritableRowSet> rowSetColumn,
+    private static void accumulateToBuilderSequential(
+            @NotNull final ObjectArraySource<RowSetBuilderSequential> rowSetColumn,
             @NotNull final RowSet rowSetToAdd, final long destination) {
-        final WritableRowSet rowSet = rowSetColumn.getUnsafe(destination);
-        if (rowSet == NONEXISTENT_TABLE_ROW_SET) {
+        final RowSetBuilderSequential builder = rowSetColumn.getUnsafe(destination);
+        if (builder == NONEXISTENT_TABLE_ROW_SET_BUILDER_SEQUENTIAL) {
```
lbooker42, line 426: test unnecessary

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -871,15 +971,32 @@ private void propagateUpdatesToModifiedDestinations(@NotNull final RowSequence m
...
+    }
+
+    private static WritableRowSet extractAndClearBuilderSequential(
+            @NotNull final WritableObjectChunk<RowSetBuilderSequential, Values> rowSetChunk,
             final int offset) {
-        final WritableRowSet rowSet = rowSetChunk.get(offset);
-        Assert.neq(rowSet, "rowSet", NONEXISTENT_TABLE_ROW_SET, "NONEXISTENT_TABLE_ROW_SET");
-        if (rowSet != null) {
+        final RowSetBuilderSequential builder = rowSetChunk.get(offset);
+        Assert.neq(builder, "builder", NONEXISTENT_TABLE_ROW_SET_BUILDER_SEQUENTIAL,
```
lbooker42, line 992: test unnecessary

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -367,39 +392,80 @@ public boolean modifyRowKeys(final SingletonContext context, @NotNull final Long
...
-    private static void accumulateToRowSet(@NotNull final ObjectArraySource<WritableRowSet> rowSetColumn,
+    private static void accumulateToBuilderSequential(
+            @NotNull final ObjectArraySource<RowSetBuilderSequential> rowSetColumn,
             @NotNull final LongChunk<OrderedRowKeys> rowKeysToAdd,
             final int start, final int length, final long destination) {
-        final WritableRowSet rowSet = rowSetColumn.getUnsafe(destination);
-        if (rowSet == NONEXISTENT_TABLE_ROW_SET) {
+        final RowSetBuilderSequential builder = rowSetColumn.getUnsafe(destination);
+        // slice the chunk to the start and length
+        LongChunk<OrderedRowKeys> sliced = rowKeysToAdd.slice(start, length);
```
rcaudy, line 406: I don't particularly like allocating a sliced view for every add. I would have one `ResettableLongChunk` stored as an instance member and re-use it here and in the other method that slices. Be sure to allocate using `io.deephaven.chunk.ResettableLongChunk#makeResettableChunkForPool` to avoid complaints about chunk leaks in the unit test instrumentation.

Irrelevant style note: this code is pedantic about using `final` wherever possible. It scratches some itch in my brain.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -90,13 +91,17 @@ public interface AttributeCopier {
     private final String callSite;
 
     private final ObjectArraySource<QueryTable> tables;
-    private final ObjectArraySource<WritableRowSet> addedRowSets;
-    private final ObjectArraySource<WritableRowSet> removedRowSets;
-    private final ObjectArraySource<WritableRowSet> modifiedRowSets;
+    private final ObjectArraySource<RowSetBuilderSequential> addedBuildersSequential;
+    private final ObjectArraySource<RowSetBuilderRandom> addedBuilders;
```
rcaudy, line 95: We could get away with just one source, here, if we aren't offended by casting on retrieval. It wouldn't be the only time we ever did such a thing. That would avoid some unnecessary re-allocation of arrays, etc.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -411,12 +477,19 @@ private boolean appendShifts(@NotNull final LongChunk<? extends RowKeys> preShif
...
             final RowSet tableRowSet = tables.getUnsafe(destination).getRowSet();
-            final RowSet removedRowSet = removedRowSets.getUnsafe(destination);
+            final RowSetBuilderRandom removedRowSetBuilder = removedBuilders.getUnsafe(destination);
             final RowSet preShiftKeys;
-            if (removedRowSet == null) {
+            if (removedRowSetBuilder == null) {
                 preShiftKeys = tableRowSet.copy();
             } else {
-                preShiftKeys = tableRowSet.minus(removedRowSet);
+                // we want to keep the builder in `removedRowSetBuilders` ready for a call to `extractAndClearRowSet` so
```
rcaudy, line 485: Rebuilding the `RowSet` this way is unfortunate, and we should avoid it. We could instead use a dummy builder that "holds" a single `RowSet` to return on `build()` and throws `UnsupportedOperationException` for all other methods.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -446,10 +519,13 @@ private boolean appendShifts(@NotNull final LongChunk<? extends RowKeys> preShif
     @Override
     public void ensureCapacity(final long tableSize) {
         tables.ensureCapacity(tableSize);
-        addedRowSets.ensureCapacity(tableSize);
+        addedBuilders.ensureCapacity(tableSize);
```
rcaudy, line 522: The way you are currently doing things, `addedBuilders` is only ever needed for refreshing parents, and should be moved into the appropriate `if` block. That said, I am advocating some consolidation.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -471,23 +547,25 @@ public void propagateInitialState(@NotNull final QueryTable resultTable) {
             final boolean setCallSite = QueryPerformanceRecorder.setCallsite(callSite);
             try (final ResettableWritableObjectChunk<QueryTable, Values> tablesResettableChunk =
                     ResettableWritableObjectChunk.makeResettableChunk();
-                    final ResettableWritableObjectChunk<WritableRowSet, Values> addedRowSetsResettableChunk =
-                            ResettableWritableObjectChunk.makeResettableChunk();
+                    final ResettableWritableObjectChunk<RowSetBuilderSequential, Values> addedBuildersResettableChunk =
```
rcaudy, line 550: This method is called exactly once, and it switches us over to `initialized`. I would add an `Assert.neqTrue(initialized, "initialized")` at the top of the method. You can and should remove code here that allows for `initialized == true`.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -508,6 +586,7 @@ public void propagateInitialState(@NotNull final QueryTable resultTable) {
                 }
             }
         }
+        initialized = true;
```
rcaudy, line 589: We should be throwing away `addedBuildersSequential` here if you don't accept my consolidation suggestion. We should not permanently waste that heap.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -789,11 +872,11 @@ private void propagateUpdatesToModifiedDestinations(@NotNull final RowSequence m
         }
         try (final ResettableWritableObjectChunk<QueryTable, Values> tablesResettableChunk =
                 ResettableWritableObjectChunk.makeResettableChunk();
-                final ResettableWritableObjectChunk<WritableRowSet, Values> addedRowSetsResettableChunk =
-                        ResettableWritableObjectChunk.makeResettableChunk();
-                final ResettableWritableObjectChunk<WritableRowSet, Values> removedRowSetsResettableChunk =
+                final ResettableWritableObjectChunk<RowSetBuilderRandom, Values> addedBuildersResettableChunk =
```
rcaudy, line 875: The `initialized` check here seems unnecessary. This can only be called after `propagateInitialState`.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -820,9 +904,13 @@ private void propagateUpdatesToModifiedDestinations(@NotNull final RowSequence m
                 final long firstBackingChunkDestination =
                         tables.resetWritableChunkToBackingStore(tablesResettableChunk, firstSliceDestination);
                 // The (valid) assumption is that the other write-through resets will address the same range.
-                addedRowSets.resetWritableChunkToBackingStore(addedRowSetsResettableChunk, firstSliceDestination);
-                removedRowSets.resetWritableChunkToBackingStore(removedRowSetsResettableChunk, firstSliceDestination);
-                modifiedRowSets.resetWritableChunkToBackingStore(modifiedRowSetsResettableChunk, firstSliceDestination);
+                addedBuilders.resetWritableChunkToBackingStore(addedBuildersResettableChunk,
+                        firstSliceDestination);
+
```
rcaudy, line 909: Weird extra new blank line is distracting, here and in subsequent blocks. Is there a reason to separate added from the rest?

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -90,13 +91,17 @@ public interface AttributeCopier {
...
+    private final ObjectArraySource<RowSetBuilderSequential> addedBuildersSequential;
+    private final ObjectArraySource<RowSetBuilderRandom> addedBuilders;
+    private final ObjectArraySource<RowSetBuilderRandom> removedBuilders;
+    private final ObjectArraySource<RowSetBuilderRandom> modifiedBuilders;
+
     private final ObjectArraySource<RowSetShiftData.SmartCoalescingBuilder> shiftDataBuilders;
     private final ModifiedColumnSet resultModifiedColumnSet;
     private final ModifiedColumnSet.Transformer upstreamToResultTransformer;
 
+    private volatile boolean initialized;
```
rcaudy, line 103: Need not be `volatile`; arguably same applies to the other two `volatile` fields. I'm fairly convinced of the happens-before relationship between initialization and subsequent update processing without the additional protection.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -139,11 +144,13 @@ public interface AttributeCopier {
             @NotNull final String... keyColumnNames) {
         this.parentTable = parentTable;
         this.resultName = resultName;
+        this.initialized = false;
```
rcaudy, line 147: Strictly personal opinion: `this.` is misleading when used unnecessarily (that is, in cases where the assignment unambiguous).
```suggestion
        initialized = false;
```

rcaudy (commented):
The changes are generally sound. We can be a little cleaner with allocation.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -477,19 +470,46 @@ private boolean appendShifts(@NotNull final LongChunk<? extends RowKeys> preShif
...
                 preShiftKeys = tableRowSet.copy();
             } else {
-                // we want to keep the builder in `removedRowSetBuilders` ready for a call to `extractAndClearRowSet` so
-                // we remake the builder after this call
-                try (final RowSet rs = removedRowSetBuilder.build()) {
-                    preShiftKeys = tableRowSet.minus(rs);
-                    RowSetBuilderRandom newBuilder = RowSetFactory.builderRandom();
-                    newBuilder.addRowSet(rs);
-                    removedBuilders.set(destination, newBuilder);
+                // dummy builder to allow a subsequent call to build()`
```
rcaudy, line 479: I think `io.deephaven.engine.table.impl.util` would be a good package to put this in as a top-level class. We have another "dummy" builder class in the chunked agg helper, and will soon have yet another when I next merge, but maybe I'll follow your lead.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/PartitionByChunkedOperator.java
```diff
@@ -477,19 +470,46 @@ private boolean appendShifts(@NotNull final LongChunk<? extends RowKeys> preShif
...
+                    public WritableRowSet build() {
+                        // make sure to release our reference to the rowset after `build()` is called
+                        WritableRowSet tmp = rowSet;
+                        rowSet = null;
+                        return tmp;
+                    }
+
+                    @Override
+                    public void addKey(long rowKey) {
+                        throw (new UnsupportedOperationException("PrecomputedRowSetBuilder cannot be modified)"));
```
rcaudy, line 497: Why the weird parentheses here?

lbooker42: TIL (Today I Learned) that `throw` is a keyword and not a function call

rcaudy (approved):

PR #2855, opened 2022-09-14
Add Python builtins to the query scope

### LINKED ISSUE #2854: make Python enclosing and built-in namespaces available in the query scope
Right now, only a function's locals and its host module's globals are included in the query scope, which means we don't fully support the Python's LEGB scope rules.

jmao-denver:
Added Python builtins namespace, **BUT**  to use symbols in the outer scopes (excluding module), you would still need to declare them or access them first in the inner function 

Fixes #2854

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -70,7 +70,9 @@ def _query_scope_ctx():
     caller_frame = outer_frames[i + 1].frame
     function = outer_frames[i + 1].function
     if len(outer_frames) > i + 2 or function != "<module>":
-        scope_dict = caller_frame.f_globals.copy()
+        # create the query scope in the reverse order of Python scope rules - LEGB
+        scope_dict = caller_frame.f_builtins.copy()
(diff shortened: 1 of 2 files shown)


jjbrosnan (commented):
Do we know what kind of conflicts occur if you use a Python built-in with the same name as a query string built-in? Something like `sum`, `all`, etc?

jmao-denver:
> Do we know what kind of conflicts occur if you use a Python built-in with the same name as a query string built-in? Something like `sum`, `all`, etc?

int, float, bool, these Java keywords will break the formula parser.
max, sum, abs, etc. these ones that we have our own implementations will not have lower precedence than the Python builtin ones and if a user wants to use them, the full qualified Java name must be provided, e.g. ```io.deephaven.function.Numeric.max```.

chipkent:
After many discussions, we have decided that merging this PR will break the DH query language in many ways.  For example, the Python built-in `int` function will clash with the Java type cast `(int)`.  We are better off not supporting the Python built-in scope.

PR #2856, opened 2022-09-14
Improve `RowSetBuilderSequential` performance when adding keys from a chunk

lbooker42:
very rough and only subjective perf testing.  would rather not make these changes myself...


engine/rowset/src/main/java/io/deephaven/engine/rowset/RowSetBuilderSequential.java
```diff
@@ -43,7 +43,11 @@ default void appendKeys(PrimitiveIterator.OfLong it) {
     }
 
     default void appendOrderedRowKeysChunk(final LongChunk<OrderedRowKeys> chunk) {
-        appendKeys(new LongChunkIterator(chunk));
+        appendOrderedRowKeysChunk(chunk, 0, chunk.size());
+    }
+
+    default void appendOrderedRowKeysChunk(final LongChunk<OrderedRowKeys> chunk, int offset, int length) {
```
rcaudy, line 49: Are we missing an overload for `addOrderedRowKeysChunk` with `OrderedRowKeys` to map to append?

rcaudy (approved):

PR #2861, opened 2022-09-15
Add __version__ in deephaven_server

### LINKED ISSUE #2858: Python `deephaven_server` module does not contain `__version__`
The `deephaven_server` module does not contain `__version__`, so users can not get the version programmatically.

jmao-denver:
Fixes #2858


jmao-denver:
@devinrsmith you probably need to add this file to the list where the version numbers are updated during the release process, if that's the hard-coded list.

chipkent (approved):

Issue #2564, opened 2022-06-22
EPIC: Authentication support


niloc132:
Part of a series of work to implement AAA in Deephaven Core.

Authentication is a mechanism to confirm an assertion of identity. Within deephaven-core, a user's identity could be used by a separate [authentication feature](https://github.com/deephaven/deephaven-core/issues/2563) to authorize that they are allowed to connect, perform operations, read/write data, etc. Sometimes abbreviated as authn (as opposed to authz for authorization), this does not cover granting access based on identity, only identity itself.

Implementations to consider:
 * Pre-shared key (PSK) support: a single key exists in a configuration parameter, and knowledge of that key identifies the user.
 * mTLS support: By virtue of connecting with a client certificate, the server can confirm who the user is.
    * The trivial implementation of this, where the server doesn't check the client's certificate as an identity assertion is technically only sort of authorization, as they can only connect after undergoing this, but have no actual identity asserted.
 * Username/Password with some lookup: Like PSK, but allow the user to claim an identity with a "token" of some kind that can be used to assert that identity when tested against some repository. The password token could be a traditional password, tested against a store, or could be a one-time-key, etc.
 * Private key signed nonce: The server can challenge the client to sign a nonce with their private key. The server then verifies the signature, knowing the public key. Valid signatures indicate that the user is in possession of the correct private key.
 * SAML/OAuth2/OIDC/etc: Leverage some generic third-party identity provider, which asserts the identity of the user, but requires a more complex flow than just sending a token to the server

Within deephaven-core, the gRPG SessionService's NewSession and RefreshSessionToken may be deprecated as part of this work, and FlightService's Handshake used to replace it, so that any Flight client is able to authenticate with deephaven. Note that this will not be necessary for the trivial mTLS implementation, as that is not asserting any identity.

Given that change, we should probably abstract out the implementations above so that FlightService.Handshake can delegate to them in a consistent way. The mTLS implementations might not make sense for this directly, but gRPC and Flight both suggest mTLS as a way to authenticate, so there is likely some prior art to check out in this area.

PR #2865, opened 2022-09-16
Refine Authentication apis, clarify deprecation and migration

niloc132:
Partial #2564


proto/proto-backplane-grpc/src/main/proto/deephaven/proto/session.proto
```diff
@@ -99,12 +107,12 @@ message HandshakeRequest {
    * - protocol = 0: most recent HandshakeResponse payload
    * - protocol = 1: payload is BasicAuth
    */
-  sint32 auth_protocol = 1;
+  sint32 auth_protocol = 1 [deprecated=true];
```
nbauernfeind, line 110: Is it possible to deprecate the whole HandshakeRequest object, or no?

niloc132: It appears not. I used the answer at https://stackoverflow.com/a/43380742/860630 to guide this change, and spam the "this is deprecated, plan on us removing it!" notices.

server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
```diff
@@ -65,10 +65,14 @@ public SessionServiceGrpcImpl(final SessionService service, final TicketRouter t
     @Override
     public void newSession(final HandshakeRequest request, final StreamObserver<HandshakeResponse> responseObserver) {
         GrpcUtil.rpcWrapper(log, responseObserver, () -> {
-            // TODO: once jsapi is updated to use flight auth, then newSession can be deprecated or removed
-            final AuthContext authContext = new AuthContext.SuperUser();
+            // TODO: once client APIs are updated to use flight auth, then newSession can be deprecated or removed
```
niloc132, line 68: going to revert changes in this file, the old impl "works" and any change here will break existing clients.

server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
```diff
@@ -295,8 +295,8 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(final ServerCall<Re
                 try {
                     session = service.getSessionForAuthToken(token);
                 } catch (AuthenticationException e) {
-                    log.error().append("Failed to authenticate: ").append(e).endl();
-                    throw Status.UNAUTHENTICATED.asRuntimeException();
+                    call.close(Status.UNAUTHENTICATED, new Metadata());
```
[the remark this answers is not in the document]
niloc132: It looks like "probably", as it can throw if something already closed it.

nbauernfeind (approved):

PR #2867, opened 2022-09-19
Wrap ExecutionContext in Python

### LINKED ISSUE #2769: Wrap ExecutionContext for Python usages
ExecutionContext exists java-side and is a bit cumbersome for python users to use as it requires using jpy and loading a specific class.

One such usage (though not a recommended as it creates tables from within a formula column) looks like:
```python
   def test_nested_scopes(self):
        _ExecutionContext = jpy.get_type("io.deephaven.engine.context.ExecutionContext")
        _context = _ExecutionContext.newBuilder() \
          .captureCompilerContext() \
          .captureQueryLibrary() \
          .captureQueryScope() \
          .build()
          
        def inner_func(p) -> str:
            openContext = _context.open()
            t = empty_table(1).update("X = p * 10")
            openContext.close()
            return t.to_string().split()[2]

        t = empty_table(1).update("X = i").update("TableString = inner_func(X + 10)")
        self.assertIn("100", t.to_string())
```

jmao-denver:
Fixes #2769

### DIFF
--- a/engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryCompiler.java
+++ b/engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryCompiler.java
@@ -20,7 +20,7 @@ public class PoisonedQueryCompiler extends QueryCompiler {
 
     private <T> T fail() {
         logger.error().append(
-                "No ExecutionContext provided; cannot use QueryCompiler; please refer to documentation on ExecutionContext for how to use it.")
+                "No ExecutionContext provided; cannot use QueryCompiler; If this is being run in a thread, did you specify an ExecutionContext for the thread?  Please refer to the documentation on ExecutionContext for details.")
                 .endl();
         throw new NoExecutionContextRegisteredException();
     }
--- a/engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryCompiler.java
+++ b/engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryCompiler.java
@@ -19,7 +19,9 @@ public class PoisonedQueryCompiler extends QueryCompiler {
     private PoisonedQueryCompiler() {}
 
     private <T> T fail() {
-        logger.error().append("No ExecutionContext provided; cannot use QueryCompiler").endl();
+        logger.error().append(
+                "No ExecutionContext provided; cannot use QueryCompiler; please refer to documentation on ExecutionContext for how to use it.")
+                .endl();
         throw new NoExecutionContextRegisteredException();
     }
 
--- a/engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryCompiler.java
+++ b/engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryCompiler.java
@@ -20,7 +20,7 @@ public class PoisonedQueryCompiler extends QueryCompiler {
 
     private <T> T fail() {
         logger.error().append(
-                "No ExecutionContext provided; cannot use QueryCompiler; If this is being run in a thread, did you specify an ExecutionContext for the thread?  Please refer to the documentation on ExecutionContext for details.")
+                "No ExecutionContext provided, cannot use QueryCompiler. If this is being run in a thread, did you specify an ExecutionContext for the thread? Please refer to the documentation on ExecutionContext for details.")
                 .endl();
         throw new NoExecutionContextRegisteredException();
     }
--- a/engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryLibrary.java
+++ b/engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryLibrary.java
@@ -14,7 +14,7 @@ public class PoisonedQueryLibrary extends QueryLibrary {
 
     private <T> T fail() {
         logger.error().append(
-                "No ExecutionContext provided; cannot use QueryLibrary; please refer to documentation on ExecutionContext for how to use it.")
+                "No ExecutionContext provided; cannot use QueryCompiler; If this is being run in a thread, did you specify an ExecutionContext for the thread?  Please refer to the documentation on ExecutionContext for details.")
                 .endl();
         throw new NoExecutionContextRegisteredException();
     }
--- a/engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryLibrary.java
+++ b/engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryLibrary.java
@@ -14,7 +14,7 @@ public class PoisonedQueryLibrary extends QueryLibrary {
 
     private <T> T fail() {
         logger.error().append(
-                "No ExecutionContext provided; cannot use QueryCompiler; If this is being run in a thread, did you specify an ExecutionContext for the thread?  Please refer to the documentation on ExecutionContext for details.")
+                "No ExecutionContext provided, cannot use QueryLibrary. If this is being run in a thread, did you specify an ExecutionContext for the thread? Please refer to the documentation on ExecutionContext for details.")
                 .endl();
         throw new NoExecutionContextRegisteredException();
     }
--- a/engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryLibrary.java
+++ b/engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryLibrary.java
@@ -13,7 +13,9 @@ public class PoisonedQueryLibrary extends QueryLibrary {
     private PoisonedQueryLibrary() {}
 
     private <T> T fail() {
-        logger.error().append("No ExecutionContext provided; cannot use QueryLibrary").endl();
+        logger.error().append(
+                "No ExecutionContext provided; cannot use QueryLibrary; please refer to documentation on ExecutionContext for how to use it.")
+                .endl();
         throw new NoExecutionContextRegisteredException();
     }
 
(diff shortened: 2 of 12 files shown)


py/server/deephaven/exec_ctx.py
```diff
@@ -0,0 +1,87 @@
...
+        Question: without wrapping these raw Java types and without useful alternative options in the server for
+        these types, this method doesn't make much sense. leave out for now?
+        """
+        j_exec_ctx = (_JExecutionContext.newBuilder()
+                      .setQueryCompiler(j_query_compiler)
+                      .setQueryLibrary(j_query_library)
+                      .setQueryScope(j_query_scope)
+                      .build())
+        return ExecutionContext(j_exec_ctx=j_exec_ctx)
+
```
jmao-denver, line 85: @nbauernfeind  Please let me know if my thinking in the two questions makes sense?

py/server/tests/test_exec_ctx.py
```diff
@@ -0,0 +1,67 @@
...
+            self.assertIn(str(i * i), rlt)
+
+    def test_thread_pool_executor(self):
+        thread_results = {}
+
+        main_exec_ctx = ExecutionContext.get_exec_ctx()
+
+        def thread_update(p: int):
+            t = empty_table(1)
+            with main_exec_ctx:
```
chipkent, line 51: Throughout the tests, there is a pattern that looks like:
```
main_exec_ctx = ExecutionContext.get_exec_ctx()
...
def thread_func(...):
    ...
    with main_exec_ctx:
        query stuff
    ....
```
or
```
main_exec_ctx = ExecutionContext.get_exec_ctx()
...
def thread_func(...):
    ExecutionContext.set_exec_ctx(main_exec_ctx)
    ...
```
Is there a clean way to condense the same logic into a decorator, so that the user has the option to just do one of the following?
```
@exec_context
def thread_func(...):
    ...
```
or 
```
@exec_context(specific_context)
def thread_func(...):
    ...
```
It may be tricky to do this because the thread func is probably called on some other thread, making the context difficult to obtain from the point the function is called.  I suspect that this will result in many use cases that look like the following since the context needs to be handed around.  I guess it will end up looking like Go code where `context.Context` is everywhere in user code.
```
def thread_func(exec_ctx, ...):
    ...
threading.Thread(target=thread_func, args=(exec_ctx,...,))
```

py/server/tests/test_exec_ctx.py
```diff
@@ -0,0 +1,67 @@
...
+                i = future_table_updates[future]
+                thread_results[i] = future.result()
+
+        self.assertTrue(thread_results)
+        for i, rlt in thread_results.items():
+            self.assertIn(str(i * i), rlt)
+
+
+if __name__ == '__main__':
+    unittest.main()
```
chipkent, line 67: One more thing I'm thinking about.  When I got the original error that triggered this PR, it was very cryptic to us.  Is it possible to make the error user-friendly, telling the user that they need to use an execution context or something?  I'm concerned about UX if we still report nasty errors.

chipkent (changes requested):

py/server/deephaven/exec_ctx.py
```diff
@@ -0,0 +1,63 @@
...
+    A default, systemic ExecutionContext is created when a Deephaven script session is initialized, and it is used to
+    evaluate queries submitted through this script session. However, to be able to evaluate a query in a deferred
+    manner, such as in a different thread from the script session thread, an application must explicitly obtain or
+    build a ExecutionContext and either install it on the thread or use it as a context manager to enclose the query
+    in the body of the with statement.
+
+    Note that, a ExecutionContext can be shared among many threads. The most typical use pattern would be to obtain
+    the script session's systemic ExecutionContext and set it on a thread explicitly created by the user or simply
+    use it to wrap a query run in that thread.
```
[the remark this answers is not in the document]
jmao-denver: 1. Python scoping rules apply to threads as well
2. Since query scope and compiler tools are not something we allow users to access, the only possible issue is with sharing the the query library among threads, but even that is a very unlikely case
3. EC is thread-local, so locking should interfere with using it

engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryCompiler.java
```diff
@@ -19,7 +19,9 @@ public class PoisonedQueryCompiler extends QueryCompiler {
     private PoisonedQueryCompiler() {}
 
     private <T> T fail() {
-        logger.error().append("No ExecutionContext provided; cannot use QueryCompiler").endl();
+        logger.error().append(
+                "No ExecutionContext provided; cannot use QueryCompiler; please refer to documentation on ExecutionContext for how to use it.")
```
chipkent, line 23:
```suggestion
                "No ExecutionContext provided; cannot use QueryCompiler; If this is being run in a thread, did you specify an ExecutionContext for the thread?  Please refer to the documentation on ExecutionContext for details.")
```

engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryLibrary.java
```diff
@@ -13,7 +13,9 @@ public class PoisonedQueryLibrary extends QueryLibrary {
     private PoisonedQueryLibrary() {}
 
     private <T> T fail() {
-        logger.error().append("No ExecutionContext provided; cannot use QueryLibrary").endl();
+        logger.error().append(
+                "No ExecutionContext provided; cannot use QueryLibrary; please refer to documentation on ExecutionContext for how to use it.")
```
chipkent, line 17:
```suggestion
                "No ExecutionContext provided; cannot use QueryCompiler; If this is being run in a thread, did you specify an ExecutionContext for the thread?  Please refer to the documentation on ExecutionContext for details.")
```

engine/context/src/main/java/io/deephaven/engine/context/PoisonedQueryScope.java
```diff
@@ -16,7 +16,9 @@ public class PoisonedQueryScope extends QueryScope {
     private PoisonedQueryScope() {}
 
     private <T> T fail() {
-        logger.error().append("No ExecutionContext provided; cannot use QueryScope").endl();
+        logger.error().append(
+                "No ExecutionContext provided; cannot use QueryScope; please refer to documentation on ExecutionContext for how to use it.")
```
chipkent, line 20:
```suggestion
                "No ExecutionContext provided; cannot use QueryCompiler; If this is being run in a thread, did you specify an ExecutionContext for the thread?  Please refer to the documentation on ExecutionContext for details.")
```

chipkent (changes requested):

chipkent (dismissed):

rcaudy (changes requested):
Please review error messages for copy-paste issues and english punctuation.

chipkent (approved):

rcaudy (dismissed):

py/server/deephaven/table.py
```diff
@@ -55,8 +55,11 @@
 def _j_py_script_session() -> _JPythonScriptSession:
     j_execution_context = _JExecutionContext.getContext()
     j_query_scope = j_execution_context.getQueryScope()
-    j_script_session_query_scope = strict_cast(j_query_scope, _JScriptSessionQueryScope)
-    return strict_cast(j_script_session_query_scope.scriptSession(), _JPythonScriptSession)
+    try:
+        j_script_session_query_scope = strict_cast(j_query_scope, _JScriptSessionQueryScope)
+        return strict_cast(j_script_session_query_scope.scriptSession(), _JPythonScriptSession)
+    except Exception as e:
+        return None
```
chipkent, line 62: I don't totally understand the exception case here:
1) Do we need to log anything?
2) Is `Exception` so wide that we might swallow problem cases?

chipkent (approved):

py/server/deephaven/execution_context.py
```diff
@@ -0,0 +1,98 @@
...
+    return ExecutionContext(j_exec_ctx=_JExecutionContext.getContext())
+
+
+def set_exec_ctx(exec_ctx: ExecutionContext) -> None:
+    """Sets the current thread's ExecutionContext.
+
+    Args:
+        exec_ctx (ExecutionContext): the ExecutionContext to be installed on the current thread
+    """
+    _JExecutionContext.setContext(exec_ctx.j_exec_ctx)
```
nbauernfeind, line 98: I don't think we should really support this. We should enforce the usage of `open` and `close`.

nbauernfeind (dismissed):

nbauernfeind (approved):

chipkent:
I have been exploring potential ways to keep the `ExecutionContext` explicit while reducing typing and burden on the user.  I think this can easily be done by making the `ExecutionContext` into a `ContextDecorator`.  This should allow both the current syntax plus:
1) decorators to perform context management
2) A way to add context management in lambdas

This would allow a user to write a function that works both in the script and the thread context.  For the thread, they would just need to use the execution context and a lambda to get correct behavior.

```python
from contextlib import ContextDecorator

class MyCm(ContextDecorator):
    def __enter__(self):
        print("CM: OPEN")

    def __exit__(self, *exc):
        print("CM: CLOSE")

def f1(x):
    print(x)

# func without context
f1(0)

# use context with func
with MyCm():
    f1(1)

# wrap func with context
@MyCm()
def f2(x):
    print(x)

f2(2)

# wrap lambda with context
f3 = MyCm()(lambda x: f1(x))

f3(3)

###

cm = MyCm()

f1(10)

with cm:
    f1(11)

@cm
def f2(x):
    print(x)

f2(12)

f3 = cm(lambda x: f1(x))
f3(13)
```

chipkent (approved):

nbauernfeind (approved):

PR #2872, opened 2022-09-19
Remove java execution logic during table.py module init

### LINKED ISSUE #2853: Need to assert that `deephaven_server.Server` is started before importing `deephaven`
When using Deephaven as a Python library, the `deephaven` module can only be safely imported _after_ the `deephaven_server.Server` has been started.  A check needs to be added to `deephaven/__init__.py` to ensure the server has been started.  

The bug was introduced in `v0.16`.  In versions before `v0.16`, the server needed to be created (to create the JVM), but the server did not need to be started.  Changes related to Python variable scoping in the query language introduced the new requirement that the server be started.

devinrsmith:
Fixes #2853

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -55,8 +55,6 @@ _JPythonScriptSession = jpy.get_type("io.deephaven.integrations.python.PythonDee
 def _j_py_script_session() -> _JPythonScriptSession:
     j_execution_context = _JExecutionContext.getContext()
     j_query_scope = j_execution_context.getQueryScope()
-    if not j_query_scope:
-        raise DHError("ExecutionContext does not have QueryScope")
     j_script_session_query_scope = strict_cast(j_query_scope, _JScriptSessionQueryScope)
     return strict_cast(j_script_session_query_scope.scriptSession(), _JPythonScriptSession)
 
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -8,6 +8,7 @@ instruments for working with Deephaven refreshing and static data. """
 from __future__ import annotations
 
 import contextlib
+import functools
 import inspect
 from enum import Enum, auto
 from typing import Union, Sequence, List, Any, Optional, Callable
@@ -47,12 +48,16 @@ _JTableOperations = jpy.get_type("io.deephaven.api.TableOperations")
 
 # Dynamic Query Scope
 _JExecutionContext = jpy.get_type("io.deephaven.engine.context.ExecutionContext")
-_JQueryScope = jpy.get_type("io.deephaven.engine.context.QueryScope")
 _JUnsynchronizedScriptSessionQueryScope = jpy.get_type(
     "io.deephaven.engine.util.AbstractScriptSession$UnsynchronizedScriptSessionQueryScope")
 _JPythonScriptSession = jpy.get_type("io.deephaven.integrations.python.PythonDeephavenSession")
-_j_script_session = jpy.cast(_JExecutionContext.getContext().getQueryScope(), _JUnsynchronizedScriptSessionQueryScope).scriptSession()
-_j_py_script_session = jpy.cast(_j_script_session, _JPythonScriptSession)
+
+
+@functools.lru_cache(maxsize=None)
+def _j_py_script_session() -> _JPythonScriptSession:
+    _j_script_session = jpy.cast(
+        _JExecutionContext.getContext().getQueryScope(), _JUnsynchronizedScriptSessionQueryScope).scriptSession()
+    return jpy.cast(_j_script_session, _JPythonScriptSession)
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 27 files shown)


py/server/deephaven/table.py
```diff
@@ -47,12 +48,16 @@
...
 _JExecutionContext = jpy.get_type("io.deephaven.engine.context.ExecutionContext")
-_JQueryScope = jpy.get_type("io.deephaven.engine.context.QueryScope")
 _JUnsynchronizedScriptSessionQueryScope = jpy.get_type(
     "io.deephaven.engine.util.AbstractScriptSession$UnsynchronizedScriptSessionQueryScope")
 _JPythonScriptSession = jpy.get_type("io.deephaven.integrations.python.PythonDeephavenSession")
-_j_script_session = jpy.cast(_JExecutionContext.getContext().getQueryScope(), _JUnsynchronizedScriptSessionQueryScope).scriptSession()
-_j_py_script_session = jpy.cast(_j_script_session, _JPythonScriptSession)
+
+
+@functools.lru_cache(maxsize=None)
```
devinrsmith, line 56: Note: I've chosen to cache this method call to preserve the "speed" of `_query_scope_ctx()`. We may decide that it's reasonable to do this invocation every time instead of caching the value - in which case, I'm happy to remove the cache altogether. Thoughts?

jmao-denver (dismissed):
LGTM

py/server/deephaven/table.py
rcaudy: I think the execution context work enabled us to have multiple script sessions if we want to. Does this effectively break that?

jmao-denver: there is another issue with this lazy eval approach, if the very first table operation happens in non-main thread, then we will be hitting the poisoned execution context.

rcaudy (approved):

chipkent (dismissed):

devinrsmith:
Related #2867

py/server/tests/test_filters.py
```diff
@@ -11,10 +11,12 @@
 
 class FilterTestCase(BaseTestCase):
     def setUp(self):
+        BaseTestCase.setUp(self)
```
jmao-denver, line 14: super().setUp()?

py/server/deephaven/table.py
```diff
@@ -47,12 +48,17 @@
...
 _JPythonScriptSession = jpy.get_type("io.deephaven.integrations.python.PythonDeephavenSession")
-_j_script_session = jpy.cast(_JExecutionContext.getContext().getQueryScope(), _JUnsynchronizedScriptSessionQueryScope).scriptSession()
-_j_py_script_session = jpy.cast(_j_script_session, _JPythonScriptSession)
+
+
+def _j_py_script_session() -> _JPythonScriptSession:
+    j_execution_context = _JExecutionContext.getContext()
+    j_query_scope = j_execution_context.getQueryScope()
+    if not j_query_scope:
+        raise DHError("ExecutionContext does not have QueryScope")
```
jmao-denver, line 59: check for PoisonedQueryScope?

devinrsmith: The following cast will throw an appropriate exception, essentially saying that a poisoned query scope can't be cast into the type we are wanting to.

devinrsmith: That said; this may have been me fumbling around trying to figure out why my DHError wasn't working. Looking at the code, I don't think any of the values in ExecutionContext can/should be null. I'll add a not null check to ExecutionContext, and I think we can get rid of this extraneous check.

py/server/tests/test_filters.py
devinrsmith: Will do.

jmao-denver (dismissed):

py/server/tests/test_table.py
```diff
@@ -624,41 +626,24 @@ def closure_fn() -> str:
...
-
-        def inner_func(p) -> str:
-            openContext = context.open()
-            t = empty_table(1).update("X = p * 10")
-            openContext.close()
-            return t.to_string().split()[2]
-
-        t = empty_table(1).update("X = i").update("TableString = inner_func(X + 10)")
+        with make_user_exec_ctx():
+            def inner_func(p) -> str:
```
nbauernfeind, line 630: I feel that it is more clear if the method is not defined within the open context. Only the `update` calls need it.

py/server/tests/test_table.py
```diff
@@ -624,41 +626,24 @@ def closure_fn() -> str:
...
-                     .captureQueryScope()
-                     .build())
-
-        def inner_func(p) -> str:
-            open_ctx = j_context.open()
-            t = empty_table(1).update("X = p * 10")
-            open_ctx.close()
-            return t.to_string().split()[2]
+        with make_user_exec_ctx():
+            def inner_func(p) -> str:
```
nbauernfeind, line 639: same on this inner function

py/server/tests/test_table.py
devinrsmith: Good idea.

nbauernfeind:
LGTM, too (yah, I know it's already merged)

Issue #2868, opened 2022-09-19
EnsureCapacity validation safety error in static natural join / exact join state manager when used with ticking LHS


rcaudy:
User image for the error encountered:
![e2](https://user-images.githubusercontent.com/5342661/190949068-441bb23c-dbcd-4211-bec1-071cefdb92b8.png)

Further details:
LHS is a ticking table read from Kafka
RHS is a static table

rcaudy:
Implementation notes:
1. Natural join and exact join are the same operation, exact join just has extra validation.
2. We use the “static” join hash table implementation any time the RHS is static, unless we hit particular optimization paths.

In this case, LHS is refreshing, RHS is static. We can safely claim that if there is no RHS value for a given LHS value after initialization, there never will be. We thus only need to keep track of which LHS values match actual RHS values, not RHS values that might come into existence later.

I've asked the user for a more complete stack trace, if you have one available. My working theory is that  we’re allocating a tiny hash table due to a tiny RHS, and then overzealously assuming the worst case outcome when hashing the LHS in. I’ll attempt to reproduce this in the morning.

rcaudy:
Super straightforward Groovy reproducer:
```
rhs = emptyTable(100000).update("II=ii")
lhs = emptyTable(2).update("II=ii", "A=(char)ii")
lhs.setRefreshing(true)
j = lhs.naturalJoin(rhs, "II")
```

Exception:
```
java.lang.IllegalStateException: Static naturalJoin does not permit rehashing, table must be allocated with sufficient size at the beginning of initialization.
        at io.deephaven.engine.table.impl.naturaljoin.StaticNaturalJoinStateManagerTypedBase.buildTable(StaticNaturalJoinStateManagerTypedBase.java:187)
        at io.deephaven.engine.table.impl.naturaljoin.StaticNaturalJoinStateManagerTypedBase.buildFromRightSide(StaticNaturalJoinStateManagerTypedBase.java:141)
        at io.deephaven.engine.table.impl.NaturalJoinHelper.naturalJoinInternal(NaturalJoinHelper.java:265)
        at io.deephaven.engine.table.impl.NaturalJoinHelper.naturalJoin(NaturalJoinHelper.java:44)
        at io.deephaven.engine.table.impl.NaturalJoinHelper.naturalJoin(NaturalJoinHelper.java:37)
        at io.deephaven.engine.table.impl.QueryTable.naturalJoinInternal(QueryTable.java:1767)
[...]

PR #2877, opened 2022-09-19
Fix the table used for state manager sizing in ticking left + static right NJ when building from right

rcaudy:
Biggest PR ever.


nbauernfeind (approved):

rcaudy:
Fixes #2868

cpwright (approved):

PR #2885, opened 2022-09-20
Add query language functions to get data structure sizes.

### LINKED ISSUE #2883: Add a method to the query library to get data structure sizes
When using the query library and the query language, it is very common to need to get the size of a `Vector` or of a Java array.  Unfortunately, `Vector` uses `.size()`, and Java arrays use `.length`.  This makes code nastier than needed.  Consider adding methods to the query library to make this easier on users.  Also consider using the python `len` naming convention.

chipkent:
Add query language functions to get data structure (`Vector` and java array) sizes.

Fixes #2883.

### DIFF
--- a/engine/function/src/templates/Basic.ftl
+++ b/engine/function/src/templates/Basic.ftl
@@ -154,7 +154,7 @@ public class Basic {
      */
     @SuppressWarnings("rawtypes")
     static public long len(LongSizedDataStructure values) {
-        if (values == null){
+        if (values == null) {
             return NULL_LONG;
         }
 
--- a/engine/function/src/templates/Basic.ftl
+++ b/engine/function/src/templates/Basic.ftl
@@ -5,6 +5,7 @@
 package io.deephaven.function;
 
 import io.deephaven.vector.*;
+import io.deephaven.util.datastructures.LongSizedDataStructure;
 import io.deephaven.util.QueryConstants;
 import gnu.trove.list.array.*;
 import gnu.trove.set.*;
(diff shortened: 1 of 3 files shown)


engine/function/src/templates/Basic.ftl
```diff
@@ -131,6 +132,35 @@ public class Basic {
...
+
+    /**
+     * Returns the length of the input.
+     *
+     * @param values values.
+     * @return length of the input or zero for null inputs.
+     */
+    @SuppressWarnings("rawtypes")
+    static public long len(LongSizedDataStructure values) {
+        if(values == null){
```
kosak, line 157: space after "if" and also space before opening brace :-)

kosak (approved):
Two whitespace nits, but approved.

engine/function/src/templates/Basic.ftl
```diff
@@ -131,6 +132,35 @@ public class Basic {
...
+
+    /**
+     * Returns the length of the input.
+     *
+     * @param values values.
+     * @return length of the input or the Deephaven null constant for null inputs.
+     */
+    @SuppressWarnings("rawtypes")
+    static public long len(LongSizedDataStructure values) {
+        if (values == null){
```
kosak, line 157: one more space (between closing parenthesis and open brace) and I'll be happy :-)

kosak (approved):
Approved with one whitespace nit

kosak (approved):

niloc132:
Ping @nbauernfeind who suggested instead (or "also") supporting .size() on arrays, so that the code is not just consistent with python, but consistent with java (the syntax used in these kinds of expressions).

PR #2886, opened 2022-09-20
Use builders for `GroupBy` initialization and updates

### LINKED ISSUE #2873: Improve initialization performance of `GroupBy` operator
As a user, I want groupBy operations as efficient and performant as possible so that large queries complete faster and require less computing power.

Follow the actions of #2852, which introduced performance improvements for `PartitionTable` creation that are applicable for the `groupBy()` operations as well.

lbooker42:
Similar to #2852, change `GroupByChunkedOperator` to use `RowSetSequentialBuilder` during the initial creation of the `groupBy` table. 

Closes #2873

```
Rows        Buckets    groupBy (ms)    groupBy_opt (ms)    speedup
100,000,000 100        1,754           1,485               15.3%
100,000,000 25,000     106,402         11,268              89.4%
100,000,000 1,000,000  32,800          27,687              15.6%
                
                
Rows        Buckets    groupBy (ms)    groupBy_opt (ms)    speedup
100,000,000 100        1,518           1,485               2.1%
100,000,000 25,000     93,826          10,995              88.3%
100,000,000 1,000,000  3,313           22,397              48.3%
```


engine/table/src/main/java/io/deephaven/engine/table/impl/by/GroupByChunkedOperator.java
```diff
@@ -41,20 +39,23 @@
     private final boolean registeredWithHelper;
     private final boolean live;
     private final ObjectArraySource<WritableRowSet> rowSets;
+    private ObjectArraySource<RowSetBuilderSequential> addedBuilders;
```
rcaudy, line 42: If we need another commit, can you put this down with the other non-`final` vars, just so I feel better about it?

rcaudy (commented):
As-written, this only speeds up the instantiation pass. For partition-by, since we already needed to keep track of adds, removes, and modifies each cycle, changing update processing was a no-brainer. Do you think it would be worthwhile to start using random builders in updating processing to track adds and removes in group-by? I suspect the trade-off to be worth it.

rcaudy:
A thing I've just thought of for group-by and partition-by. If we're using initial groups and !preserveEmpty, we might be orphaning empty builders in the addedBuilders sources indefinitely, since we only do any work with the ones that make it into the initial row set. I'm not sure it's a big deal, but it's not optimal.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/GroupByChunkedOperator.java
```diff
@@ -331,15 +439,123 @@ public ModifiedColumnSet apply(@NotNull final ModifiedColumnSet upstreamModified
...
+                            firstBackingChunkDestination + rowSetBackingChunk.size() - 1;
+                    final RowSequence initialDestinationsSlice =
+                            destinationsIterator.getNextRowSequenceThrough(lastBackingChunkDestination);
+
+                    initialDestinationsSlice.forAllRowKeys((final long destination) -> {
+                        final int backingChunkOffset =
+                                Math.toIntExact(destination - firstBackingChunkDestination);
+                        final WritableRowSet addRowSet = nullToEmpty(extractAndClearBuilderSequential(addedBuildersBackingChunk, backingChunkOffset));
+                        rowSets.set(destination, live ? addRowSet.toTracking() : addRowSet);
+                        onReincarnated(destination);
```
rcaudy, line 470: There's no reincarnation in initial state.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/GroupByChunkedOperator.java
```diff
@@ -331,15 +439,123 @@ public ModifiedColumnSet apply(@NotNull final ModifiedColumnSet upstreamModified
...
+                                nullToEmpty(extractAndClearBuilderRandom(addedBuildersBackingChunk, backingChunkOffset));
+                             final WritableRowSet removeRowSet =
+                                nullToEmpty(extractAndClearBuilderRandom(removedBuildersBackingChunk, backingChunkOffset))) {
+                            final WritableRowSet workingRowSet = rowSetForSlot(backingChunkOffset);
+
+                            // process removes
+                            final boolean wasNonEmpty = removeRowSet.isNonempty();
+                            workingRowSet.remove(removeRowSet);
+                            if (wasNonEmpty && workingRowSet.isEmpty()) {
+                                onEmptied(destination);
```
rcaudy: This is too late in update processing to meaningfully report emptied to reincarnated. We use that to determine `downstream` before this call is made. If we can't determine emptied or reincarnated correctly in the `addChunk` and `removeChunk` methods (and I don't think we can), we should just remove `extends BasicStateChangeRecorder` from the class signature.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/GroupByChunkedOperator.java
```diff
@@ -331,15 +439,123 @@ public ModifiedColumnSet apply(@NotNull final ModifiedColumnSet upstreamModified
...
+                            destinationsIterator.getNextRowSequenceThrough(lastBackingChunkDestination);
+
+                    initialDestinationsSlice.forAllRowKeys((final long destination) -> {
+                        final int backingChunkOffset =
+                                Math.toIntExact(destination - firstBackingChunkDestination);
+                        try (final WritableRowSet addRowSet =
+                                nullToEmpty(extractAndClearBuilderRandom(addedBuildersBackingChunk, backingChunkOffset));
+                             final WritableRowSet removeRowSet =
+                                nullToEmpty(extractAndClearBuilderRandom(removedBuildersBackingChunk, backingChunkOffset))) {
+                            final WritableRowSet workingRowSet = rowSetForSlot(backingChunkOffset);
```
rcaudy: We should consider a change here. If `rowSets.getUnsafe` returns `null`, we can safely just take the `addedRowSet` (and convert it to tracking if needed), rather than having to go through the rigamarole of dealing with reference counting on the `OrderedLongSet`, etc. `removedRowSet` had better be empty in this situation.
Note that a `null` `AutoCloseable` is cool, the try-with-resources will just ignore it.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/GroupByChunkedOperator.java
```diff
@@ -331,15 +439,123 @@ public ModifiedColumnSet apply(@NotNull final ModifiedColumnSet upstreamModified
...
+
+                            // process removes
+                            final boolean wasNonEmpty = removeRowSet.isNonempty();
+                            workingRowSet.remove(removeRowSet);
+                            if (wasNonEmpty && workingRowSet.isEmpty()) {
+                                onEmptied(destination);
+                            }
+
+                            // process adds
+                            final boolean wasEmpty = workingRowSet.isEmpty();
```
rcaudy: Moot since I think this is coming out, but I wanted to note that it's *correct* that a state can be emptied and reincarnated on the same cycle, and you are handling that correctly. The helper notes that such rows are "added back" and removes them from adds/removes and instead adds them to modifies.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/GroupByChunkedOperator.java
```diff
@@ -494,67 +478,69 @@ public void propagateInitialState(@NotNull final QueryTable resultTable) {
...
+
+                        final long lastBackingChunkDestination =
+                                firstBackingChunkDestination + rowSetBackingChunk.size() - 1;
+                        final RowSequence initialDestinationsSlice =
+                                destinationsIterator.getNextRowSequenceThrough(lastBackingChunkDestination);
+
+                        initialDestinationsSlice.forAllRowKeys((final long destination) -> {
+                            final int backingChunkOffset =
+                                    Math.toIntExact(destination - firstBackingChunkDestination);
+                            WritableRowSet workingRowSet = rowSetBackingChunk.get(backingChunkOffset);
```
rcaudy, line 518: `getUnsafe`

engine/table/src/main/java/io/deephaven/engine/table/impl/by/GroupByChunkedOperator.java
```diff
@@ -494,67 +478,69 @@ public void propagateInitialState(@NotNull final QueryTable resultTable) {
...
+                            if (workingRowSet == null) {
+                                // use the addRowSet as the new rowset
+                                final WritableRowSet addRowSet = nullToEmpty(extractAndClearBuilderRandom(addedBuildersBackingChunk, backingChunkOffset));
+                                rowSetBackingChunk.set(backingChunkOffset, live ? addRowSet.toTracking() : addRowSet);
                                 onReincarnated(destination);
```
rcaudy, line 523: Can't be reincarnated, has to be brand new. New slots are handled by the helper, anyway.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/GroupByChunkedOperator.java
```diff
@@ -494,67 +478,69 @@ public void propagateInitialState(@NotNull final QueryTable resultTable) {
...
+                            } else {
+                                try (final WritableRowSet addRowSet =
+                                             nullToEmpty(extractAndClearBuilderRandom(addedBuildersBackingChunk, backingChunkOffset));
+                                     final WritableRowSet removeRowSet =
+                                             nullToEmpty(extractAndClearBuilderRandom(removedBuildersBackingChunk, backingChunkOffset))) {
+                                    final boolean wasEmpty = workingRowSet.isEmpty();
+                                    workingRowSet.remove(removeRowSet);
+                                    workingRowSet.insert(addRowSet);
+                                    if (!wasEmpty && workingRowSet.isEmpty()) {
```
rcaudy, line 532: See note about the previous version of this, and "added back" slots.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/GroupByChunkedOperator.java
lbooker42: I didn't think `getUnsafe()` was available on a `WritableObjectChunk`

rcaudy: It's not, I was mistakenly thinking this was an array source. Thank you for your generosity in pretending I might be right.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/GroupByChunkedOperator.java
```diff
@@ -41,30 +41,43 @@
     private final boolean registeredWithHelper;
     private final boolean live;
     private final ObjectArraySource<WritableRowSet> rowSets;
+    private final ObjectArraySource<Object> addedBuilders;
+    private final ObjectArraySource<Object> removedBuilders;
+    private RowSetBuilderRandom stepUpdatedDestinations;
```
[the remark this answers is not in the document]
lbooker42: I went with `BitmapRandomBuilder` after failing tests even with including all modified rows. There are some things going on I don't quite understand...

rcaudy (commented):
I think this is basically ready to merge, if nightlies look good. Minor nitpicks only.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/GroupByChunkedOperator.java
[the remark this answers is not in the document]
rcaudy: Let me know if you want to discuss the problems you saw. It would be good to check both of our understandings.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/AggregationContext.java
```diff
@@ -234,10 +234,11 @@ ModifiedColumnSet[] getInputModifiedColumnSets(QueryTable input) {
      * initialization.
      *
      * @param resultTable The result {@link QueryTable} after initialization
+     * @param maxDestination The maximum destination slot created during initialization
```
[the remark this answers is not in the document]
lbooker42: Agree, was overlooked

engine/table/src/main/java/io/deephaven/engine/table/impl/by/ChunkedOperatorAggregationHelper.java
```diff
@@ -2412,7 +2413,7 @@ public void addRange(final long firstRowKey, final long lastRowKey) {
      * We also know that we will only modify the rows that existed when we start, so that we can clamp the maximum key
      * for the builder to the maximum output position without loss of fidelity.
      */
-    private static class BitmapRandomBuilder implements RowSetBuilderRandom {
+    public static class BitmapRandomBuilder implements RowSetBuilderRandom {
```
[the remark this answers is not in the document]
lbooker42: Extracted to `io/deephaven/engine/rowset/impl`

engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/BitmapRandomBuilder.java
```diff
@@ -0,0 +1,96 @@
...
+ * <p>
+ * This version is O(1) for updating a modified slot, then linear in the number of output positions (not the number of
+ * result values) to build the RowSet. The memory usage is 1 bit per output position, vs. the standard builder is 128
+ * bits per used value (though with the possibility of collapsing adjacent ranges when they are modified back-to-back).
+ * For random access patterns, this version will be more efficient; for friendly patterns the default random builder is
+ * likely more efficient.
+ * <p>
+ * We also know that we will only modify the rows that existed when we start, so that we can clamp the maximum key for
+ * the builder to the maximum output position without loss of fidelity.
+ */
```
rcaudy, line 24: We should either re-write the top comment and add an implementation for addRange, or put this class back in `io.deephaven.engine.table.impl.by` instead.

rcaudy (approved):

PR #2904, opened 2022-09-23
Wrap consumeToPartitionedTable in Python

### LINKED ISSUE #2901: Python Kafka partitioned table support
During #2418 we lost the ability to create a kafka partitioned table from python; ala, `io.deephaven.kafka.KafkaTools#consumeToPartitionedTable`.

From a python api perspective, it probably makes sense to have it as a simple flag

```python
consume(..., partitioned=True)
```

jmao-denver:
Fixes #2901

### DIFF
--- a/py/server/deephaven/stream/kafka/consumer.py
+++ b/py/server/deephaven/stream/kafka/consumer.py
@@ -203,7 +203,9 @@ def consume_to_partitioned_table(
             default is TableType.stream()
 
     Returns:
-        a Deephaven live partitioned table that will update based on Kafka messages consumed for the given topic
+        a Deephaven live partitioned table that will update based on Kafka messages consumed for the given topic,
+        the keys of this partitioned table are the partition numbers of the topic, and its constituents are tables per
+        topic partition.
 
     Raises:
         DHError
--- a/py/server/deephaven/stream/kafka/consumer.py
+++ b/py/server/deephaven/stream/kafka/consumer.py
@@ -199,7 +199,8 @@ def consume_to_partitioned_table(
             works the same as KeyValueSpec.FROM_PROPERTIES, in which case, the kafka_config param should include values
             for dictionary keys 'deephaven.key.column.name' and 'deephaven.key.column.type', for the single resulting
             column name and type
-        table_type (TableType): a TableType enum, default is TableType.stream()
+        table_type (TableType): a TableType enum, specifying the type of the expected result's constituent tables,
+            default is TableType.stream()
 
     Returns:
         a Deephaven live partitioned table that will update based on Kafka messages consumed for the given topic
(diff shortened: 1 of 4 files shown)


py/server/deephaven/stream/kafka/consumer.py
```diff
@@ -161,6 +163,65 @@ def consume(
...
+            It can be the result of calling one of the functions: simple_spec(),avro_spec() or json_spec() in this
+            module, or the predefined KeyValueSpec.IGNORE or KeyValueSpec.FROM_PROPERTIES. The default is None which
+            works the same as KeyValueSpec.FROM_PROPERTIES, in which case, the kafka_config param should include values
+            for dictionary keys 'deephaven.key.column.name' and 'deephaven.key.column.type', for the single resulting
+            column name and type
+        table_type (TableType): a TableType enum, specifying the type of the expected result's constituent tables,
+            default is TableType.stream()
+
+    Returns:
+        a Deephaven live partitioned table that will update based on Kafka messages consumed for the given topic
```
rcaudy, line 206: I think we should clarify that the keys are ints from the topic partition, and the constituents are tables per topic partition.

rcaudy (commented):
Minor comment regarding documentation.

rcaudy (approved):

py/server/deephaven/stream/kafka/consumer.py
```diff
@@ -184,17 +247,28 @@ def consume(
             raise ValueError("at least one argument for 'key' or 'value' must be different from KeyValueSpec.IGNORE")
 
         kafka_config = j_properties(kafka_config)
-        return Table(
-            j_table=_JKafkaTools.consumeToTable(
+        if not to_partitioned:
```
devinrsmith, line 250: Thoughts on:

```python
java_func = _JKafkaTools.consumeToTable if not to_partitioned else _JKafkaTools.consumeToPartitionedTable
return java_func(kafka_config, topic, ...)
```
and then having the calling methods wrap w/ Table/PartitionedTable?

jmao-denver: That would make the code a bit cleaner and more concise, but OTH we won't be able property type hint the return type, just more a generic jpy.JType. I slightly prefer to keep it as it.

devinrsmith: I'll approve as-is; I personally am not too concerned about type hinting for *internal* methods in this case. An alternative is also:

```python
java_func = _JKafkaTools.consumeToTable if not to_partitioned else _JKafkaTools.consumeToPartitionedTable
val = java_func(kafka_config, topic, ...)
return Table(val) if not to_partitioned else PartitionTable(val)
```

Feel free to take the approach your most prefer.

devinrsmith (approved):
See earlier comment.

PR #2907, opened 2022-09-25
Wrap the partitionedAggBy convenience mtd in py

### LINKED ISSUE #2786: The new convenience method - partition_agg_by needs to be wrapped for Python

### LINKED ISSUE #2874: Additional TableOperations.aggBy() overloads need to be wrapped in Python
Specifically, there are two of them
```
    /**
     * Produce an aggregated result by grouping {@code this} according to the {@code groupByColumns} and applying
     * {@code aggregations} to each resulting group of rows. The result table will have one row per group, ordered by
     * the <em>encounter order</em> within {@code this}, thereby ensuring that the row key for a given group never
     * changes.
     *
     * @param aggregations The {@link Aggregation aggregations} to apply
     * @param preserveEmpty Whether to keep result rows for groups that are initially empty or become empty as a result
     *        of updates. Each aggregation operator defines its own value for empty groups.
     * @param initialGroups A table whose distinct combinations of values for the {@code groupByColumns} should be used
     *        to create an initial set of aggregation groups. All other columns are ignored. This is useful in
     *        combination with {@code preserveEmpty == true} to ensure that particular groups appear in the result
     *        table, or with {@code preserveEmpty == false} to control the encounter order for a collection of groups
     *        and thus their relative order in the result. Changes to {@code initialGroups} are not expected or handled;
     *        if {@code initialGroups} is a refreshing table, only its contents at instantiation time will be used. If
     *        {@code initialGroups == null}, the result will be the same as if a table with no rows was supplied.
     * @param groupByColumns The {@link ColumnName columns} to group by
     * @return A new table aggregating the rows of {@code this}
     */
    @ConcurrentMethod
    TOPS aggBy(Collection<? extends Aggregation> aggregations, boolean preserveEmpty, TABLE initialGroups,
            Collection<? extends ColumnName> groupByColumns);
```
and
```
    /**
     * Produce an aggregated result by grouping all rows from {@code this} into a single group of rows and applying
     * {@code aggregations} to the result.
     *
     * @param aggregations The {@link Aggregation aggregations} to apply
     * @param preserveEmpty If {@code preserveEmpty == true}, the result table will always have one row. Otherwise, the
     *        result table will have one row if {@code this} has one or more rows, or else zero rows.
     * @return A new table aggregating the rows of {@code this}
     */
    @ConcurrentMethod
    TOPS aggBy(Collection<? extends Aggregation> aggregations, boolean preserveEmpty);
```

jmao-denver:
Fixes #2786
Fixes #2874

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -16,7 +16,7 @@ import jpy
 
 from deephaven import DHError, dtypes
 from deephaven._jpy import strict_cast
-from deephaven._wrapper import JObjectWrapper
+from deephaven._wrapper import JObjectWrapper, unwrap
 from deephaven.agg import Aggregation
 from deephaven.column import Column, ColumnType
 from deephaven.filters import Filter
@@ -1303,6 +1303,43 @@ class Table(JObjectWrapper):
             return Table(j_table=self.j_table.aggBy(j_agg_list, *by))
         except Exception as e:
             raise DHError(e, "table agg_by operation failed.") from e
+
+    def partitioned_agg_by(self, aggs: Union[Aggregation, Sequence[Aggregation]],
+                           by: Union[str, Sequence[str]] = None, preserve_empty: bool = True,
+                           initial_groups: Table = None) -> PartitionedTable:
+        """The partitioned_agg_by method is a convenience method that performs an agg_by operation on this table and
+        wraps the result in a PartitionedTable. If the argument 'aggs' does not include a partition aggregation
+        created by calling :py:func:`agg.partition`, one will be added automatically with the default constituent column
+        name __CONSTITUENT__.
+
+        Args:
+            aggs (Union[Aggregation, Sequence[Aggregation]]): the aggregation(s)
+            by (Union[str, Sequence[str]]): the group-by column name(s), default is None
+            preserve_empty (bool): whether to keep result rows for groups that are initially empty or become empty as
+                a result of updates. Each aggregation operator defines its own value for empty groups. Default is True.
+            initial_groups (Table): a table whose distinct combinations of values for the group-by column name(s)
+                should be used to create an initial set of aggregation groups. All other columns are ignored. This is
+                useful in combination with preserve_empty == True to ensure that particular groups appear in the result
+                table, or with preserve_empty == False to control the encounter order for a collection of groups and
+                thus their relative order in the result. Changes to this table are not expected or handled; if this
+                table is a refreshing table, only its contents at instantiation time will be used. Default is None,
+                the result will be the same as if a table is provided but no rows was supplied.
+
+        Returns:
+            a PartitionedTable
+
+        Raises:
+            DHError
+        """
+        try:
+            aggs = to_sequence(aggs)
+            by = to_sequence(by)
+            j_agg_list = j_array_list([agg.j_aggregation for agg in aggs])
+            initial_groups = unwrap(initial_groups)
+            return PartitionedTable(
+                j_partitioned_table=self.j_table.partitionedAggBy(j_agg_list, preserve_empty, initial_groups, *by))
+        except Exception as e:
+            raise DHError(e, "table partitioned_agg_by operation failed.") from e
 
     def agg_all_by(self, agg: Aggregation, by: Union[str, Sequence[str]] = None) -> Table:
         """The agg_all_by method creates a new table containing grouping columns and grouped data. The resulting
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -1282,14 +1282,25 @@ class Table(JObjectWrapper):
         except Exception as e:
             raise DHError(e, "table count_by operation failed.") from e
 
-    def agg_by(self, aggs: Union[Aggregation, Sequence[Aggregation]], by: Union[str, Sequence[str]] = None) -> Table:
+    def agg_by(self, aggs: Union[Aggregation, Sequence[Aggregation]], by: Union[str, Sequence[str]] = None,
+               preserve_empty: bool = False, initial_groups: Table = None) -> Table:
         """The agg_by method creates a new table containing grouping columns and grouped data. The resulting
         grouped data is defined by the aggregations specified.
 
         Args:
             aggs (Union[Aggregation, Sequence[Aggregation]]): the aggregation(s)
-            by (Union[str, Sequence[str]]): the group-by column name(s), default is None
-
+            by (Union[str, Sequence[str]]): the group-by column name(s), if not provided, all rows from this table are
+                grouped into a single group of rows before the aggregations are applied to the result, default is None.
+            preserve_empty (bool): whether to keep result rows for groups that are initially empty or become empty as
+                a result of updates. Each aggregation operator defines its own value for empty groups. Default is False.
+            initial_groups (Table): a table whose distinct combinations of values for the group-by column name(s)
+                should be used to create an initial set of aggregation groups. All other columns are ignored. This is
+                useful in combination with preserve_empty == True to ensure that particular groups appear in the result
+                table, or with preserve_empty == False to control the encounter order for a collection of groups and
+                thus their relative order in the result. Changes to this table are not expected or handled; if this
+                table is a refreshing table, only its contents at instantiation time will be used. Default is None,
+                the result will be the same as if a table is provided but no rows were supplied. When it is provided,
+                the 'by' argument must be provided to explicitly specify the grouping columns.
         Returns:
             a new table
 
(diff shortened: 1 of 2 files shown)


py/server/deephaven/table.py
```diff
@@ -1304,6 +1304,43 @@ def agg_by(self, aggs: Union[Aggregation, Sequence[Aggregation]], by: Union[str,
...
+                           initial_groups: Table = None) -> PartitionedTable:
+        """The partitioned_agg_by method is a convenience method that performs an agg_by operation on this table and
+        wraps the result in a PartitionedTable. If the argument 'aggs' does not include a partition aggregation
+        created by calling :py:func:`agg.partition`, one will be added automatically with the default constituent column
+        name __CONSTITUENT__.
+
+        Args:
+            aggs (Union[Aggregation, Sequence[Aggregation]]): the aggregation(s)
+            by (Union[str, Sequence[str]]): the group-by column name(s), default is None
+            preserve_empty (bool): whether to keep result rows for groups that are initially empty or become empty as
```
rcaudy, line 1318: Default should be `false`.

py/server/deephaven/table.py
```diff
@@ -1304,6 +1304,43 @@ def agg_by(self, aggs: Union[Aggregation, Sequence[Aggregation]], by: Union[str,
...
+                a result of updates. Each aggregation operator defines its own value for empty groups. Default is True.
+            initial_groups (Table): a table whose distinct combinations of values for the group-by column name(s)
+                should be used to create an initial set of aggregation groups. All other columns are ignored. This is
+                useful in combination with preserve_empty == True to ensure that particular groups appear in the result
+                table, or with preserve_empty == False to control the encounter order for a collection of groups and
+                thus their relative order in the result. Changes to this table are not expected or handled; if this
+                table is a refreshing table, only its contents at instantiation time will be used. Default is None,
+                the result will be the same as if a table is provided but no rows was supplied.
```
rcaudy, line 1326:
```suggestion
                the result will be the same as if a table is provided but no rows were supplied.
```

py/server/deephaven/table.py
```diff
@@ -1304,6 +1304,43 @@ def agg_by(self, aggs: Union[Aggregation, Sequence[Aggregation]], by: Union[str,
         except Exception as e:
             raise DHError(e, "table agg_by operation failed.") from e
 
+    def partitioned_agg_by(self, aggs: Union[Aggregation, Sequence[Aggregation]],
```
rcaudy, line 1307: I think this PR should also include the new args to `agg_by`. `partitioned_agg_by` might make more sense when coupled with those.

jmao-denver: will include #2874 in this PR

rcaudy (approved):

py/server/deephaven/table.py
```diff
@@ -1299,11 +1310,56 @@ def agg_by(self, aggs: Union[Aggregation, Sequence[Aggregation]], by: Union[str,
...
+                           by: Union[str, Sequence[str]] = None, preserve_empty: bool = False,
+                           initial_groups: Table = None) -> PartitionedTable:
+        """The partitioned_agg_by method is a convenience method that performs an agg_by operation on this table and
+        wraps the result in a PartitionedTable. If the argument 'aggs' does not include a partition aggregation
+        created by calling :py:func:`agg.partition`, one will be added automatically with the default constituent column
+        name __CONSTITUENT__.
```
chipkent, line 1332: I'm unclear here.

Does `t.partitioned_agg_by(aggs=[agg.sum("X")],by=["Y"])` create a partitioned table containing 3 columns ("X", "Y", and "`__CONSTITUENT__`", while `t.partitioned_agg_by(aggs=[agg.sum("X"), agg.partition("Z")],by=["Y"])` creates a table containing 3 columns ("X", "Y", "Z")?

What is the use case for this?

py/server/tests/test_table.py
```diff
@@ -406,30 +423,66 @@ def test_agg_by_2(self):
...
+            result_table = t.agg_by(self.aggs, preserve_empty=False)
+            self.assertEqual(result_table.size, 0)
+            result_table = t.agg_by(self.aggs, preserve_empty=True)
+            self.assertEqual(result_table.size, 1)
+            print(result_table.to_string())
+
+        with self.subTest("with initial-groups, no-by, and preserve_empty"):
+            init_groups = test_table.update("grp_id=i")
+            with self.assertRaises(DHError):
+                result_table = test_table.agg_by(self.aggs, initial_groups=init_groups)
```
chipkent, line 450: Is this raising an exception because not all groups are present?  If so, I don't think this concept is clear in the doc string above.

chipkent (changes requested):

py/server/deephaven/table.py
rcaudy: I don't think you're too unclear, since you got it exactly right. The use case is to avoid having to have users construct their own `PartitionedTable` when they do other aggregations in addition to the partition.

py/server/deephaven/table.py
```diff
@@ -1282,14 +1282,25 @@ def count_by(self, col: str, by: Union[str, Sequence[str]] = None) -> Table:
...
+            preserve_empty (bool): whether to keep result rows for groups that are initially empty or become empty as
+                a result of updates. Each aggregation operator defines its own value for empty groups. Default is False.
+            initial_groups (Table): a table whose distinct combinations of values for the group-by column name(s)
+                should be used to create an initial set of aggregation groups. All other columns are ignored. This is
+                useful in combination with preserve_empty == True to ensure that particular groups appear in the result
+                table, or with preserve_empty == False to control the encounter order for a collection of groups and
+                thus their relative order in the result. Changes to this table are not expected or handled; if this
+                table is a refreshing table, only its contents at instantiation time will be used. Default is None,
+                the result will be the same as if a table is provided but no rows were supplied. When it is provided,
+                the 'by' argument must be provided to explicitly specify the grouping columns.
```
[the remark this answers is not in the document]
rcaudy: `preserve_empty` is not tri-state. `None` had better be treated as `False` or an error. The Java Table API only supports a primitive `boolean` for this argument.

Isn't the stuff you're reacting to in the section for `initial_groups`? Is it really that confusing?

- Encounter order is the output order for all aggregations; output rows are ordered relative to the order in which the operation first observed them, so if a group "appears" earlier in upstream row key space on a subsequent update it still shows up after existing groups in the downstream output.
- It's *slightly* different than the norm, because it's the only case in the whole query engine where you might get something other than strict, repeatable order preservation; we do this for practical reasons, because of the impact on downstream queries.

py/server/tests/test_table.py
chipkent: I would comment this test.  I did read that in the docs and code.

rcaudy (approved):

py/server/deephaven/table.py
```diff
@@ -1293,10 +1293,10 @@ def agg_by(self, aggs: Union[Aggregation, Sequence[Aggregation]], by: Union[str,
                 grouped into a single group of rows before the aggregations are applied to the result, default is None.
             preserve_empty (bool): whether to keep result rows for groups that are initially empty or become empty as
                 a result of updates. Each aggregation operator defines its own value for empty groups. Default is False.
-            initial_groups (Table): a table whose distinct combinations of values for the group-by column name(s)
+            initial_groups (Table): a table whose distinct combinations of values for the group-by column(s)
                 should be used to create an initial set of aggregation groups. All other columns are ignored. This is
```
chipkent, line 1297: all other columns in `initial_groups`?

py/server/deephaven/table.py
```diff
@@ -1293,10 +1293,10 @@ def agg_by(self, aggs: Union[Aggregation, Sequence[Aggregation]], by: Union[str,
...
-                useful in combination with preserve_empty == True to ensure that particular groups appear in the result
-                table, or with preserve_empty == False to control the encounter order for a collection of groups and
+                useful in combination with preserve_empty=True to ensure that particular groups appear in the result
+                table, or with preserve_empty=False to control the encounter order for a collection of groups and
                 thus their relative order in the result. Changes to this table are not expected or handled; if this
                 table is a refreshing table, only its contents at instantiation time will be used. Default is None,
```
chipkent, line 1301: Does instantiation mean when `initial_groups` is created or when this method is called?  I'm guessing the latter, but this is not clear.

py/server/tests/test_table.py
```diff
@@ -446,6 +446,7 @@ def test_agg_by_initial_groups_preserve_empty(self):
 
         with self.subTest("with initial-groups, no-by, and preserve_empty"):
             init_groups = test_table.update("grp_id=i")
+            # can't specify 'initial-groups' without also specifying    'by'
```
chipkent, line 449:
```suggestion
            # can't specify 'initial-groups' without also specifying 'by'
```

chipkent (changes requested):

py/server/deephaven/table.py
rcaudy: Yes.

chipkent (approved):

PR #2931, opened 2022-09-29
Authorization

nbauernfeind:
This introduces authorization into the gRPC API, and ExecutionContext.


authentication/src/main/java/io/deephaven/auth/MultiAuthorizationSource.java
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+ */
+package io.deephaven.auth;
+
+/**
+ * An AuthorizationSource that composes multiple AuthorizationSources together.
+ * <p>
+ * If an AuthorizationSource returns UNDECIDED, the next AuthorizationSource is consulted. If all AuthorizationSources
+ * return UNDECIDED, the result is UNDECIDED.
+ */
```
devinrsmith, line 11: It looks like it should be documented that _order_ matters - the first non-undecided result will be returned.

authentication/src/main/java/io/deephaven/auth/RateLimitingAuthorizationSource.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public AuthorizationResult hasPrivilege(Privilege privilege) {
+        final Optional<RateLimiter> rateLimiter;
+        synchronized (rateLimiterProvider) {
+            rateLimiter = rateLimiterProvider.getRateLimiter(privilege);
+        }
+        if (!rateLimiter.map(RateLimiter::tryAcquire).orElse(true)) {
+            return AuthorizationResult.BLOCK;
+        }
```
devinrsmith, line 35: Should we attach reasons to AuthorizationResult? I've definitely seen APIs that will tell you specifically that you've been "rate limited" and that's the reason for the deny.

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -153,6 +152,7 @@ public void subscribeToLogs(LogSubscriptionRequest request, StreamObserver<LogSu
     public void executeCommand(ExecuteCommandRequest request, StreamObserver<ExecuteCommandResponse> responseObserver) {
         GrpcUtil.rpcWrapper(log, responseObserver, () -> {
             final SessionState session = sessionService.getCurrentSession();
+            session.getAuthContext().requirePrivilege(ConsoleServicePrivilege.CAN_SUBSCRIBE_TO_LOGS);
```
devinrsmith, line 155: Wrong priv?

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -185,6 +185,8 @@ public void executeCommand(ExecuteCommandRequest request, StreamObserver<Execute
     @Override
     public void getHeapInfo(GetHeapInfoRequest request, StreamObserver<GetHeapInfoResponse> responseObserver) {
         GrpcUtil.rpcWrapper(log, responseObserver, () -> {
+            final SessionState session = sessionService.getCurrentSession();
+            session.getAuthContext().requirePrivilege(ConsoleServicePrivilege.CAN_SUBSCRIBE_TO_LOGS);
```
devinrsmith, line 189: Wrong priv?

server/src/main/java/io/deephaven/server/table/TableServiceGrpcImpl.java
```diff
@@ -280,6 +280,7 @@ public void batch(final BatchTableRequest request,
             final StreamObserver<ExportedTableCreationResponse> responseObserver) {
         GrpcUtil.rpcWrapper(log, responseObserver, () -> {
             final SessionState session = sessionService.getCurrentSession();
+            session.getAuthContext().requirePrivilege(TableServicePrivilege.CAN_BATCH);
```
devinrsmith, line 283: I'm assuming that each of the individual operations in a batch still need privs, correct?

server/src/main/java/io/deephaven/server/table/inputtables/InputTableServiceGrpcImpl.java
```diff
@@ -40,6 +40,7 @@ public InputTableServiceGrpcImpl(TicketRouter ticketRouter, SessionService sessi
     public void addTableToInputTable(AddTableRequest request, StreamObserver<AddTableResponse> responseObserver) {
         GrpcUtil.rpcWrapper(log, responseObserver, () -> {
             final SessionState session = sessionService.getCurrentSession();
+            session.getAuthContext().requirePrivilege(InputTableServicePrivilege.CAN_ADD_TO_INPUT_TABLE);
```
devinrsmith, line 43: I'm mentioning it here - but it's a much more widely applicable comment.

How do we enable authorization based on the contents of the request?

For example, I want "devin" to be able to add to the input table `devins_table`, but not `nates_table`.

authentication/src/main/java/io/deephaven/auth/Privilege.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+ */
+package io.deephaven.auth;
+
+/**
+ * A privilege is a permission to perform a specific action. This is a token interface.
+ */
+public interface Privilege {
```
devinrsmith, line 9: It looks like everything that extends this must be an enum. If so, should define type:

`interface Privilege<T extends Privilege<T>> extends Enum<T>`

engine/context/src/main/java/io/deephaven/engine/context/ExecutionContext.java
```diff
@@ -112,7 +116,22 @@ public ExecutionContext withSystemic(boolean isSystemic) {
...
+
+    /**
+     * Returns, or creates, an execution context with the given value for {@code authContext} and existing values for
+     * the other members. This is not intended to be used by user code.
+     *
+     * @param authContext the auth context to use instead
+     * @return the execution context
+     */
+    public ExecutionContext withAuthContext(final AuthContext authContext) {
+        // TODO: should the result be a composition of existing auth context and the provided?
```
devinrsmith, line 130: I don't think that decision belongs at this level. If we need composition, the caller should be responsible IMO:

```java
context.withAuthContext(combined(existing, newAuthContext))
```

devinrsmith (commented):
So it looks like we have content-agnostic / method-level filtering. I think this is great. That said, I wonder if we'll find ourselves re-implementing a bunch of hooks (and potentially doing away w/ the enums) as soon as we need more fine grained control based on the requests' content?

Another question; does this structure currently support the ability to disable arbitrary code execution via formulas?

authentication/src/main/java/io/deephaven/auth/Privilege.java
devinrsmith: Alternatively, you can update the builder method where it assumes they are enums.

engine/context/src/main/java/io/deephaven/engine/context/ExecutionContext.java
```diff
@@ -25,7 +24,8 @@ public static ExecutionContext makeExecutionContext(boolean isSystemic) {
 
     @VisibleForTesting
     public static ExecutionContext createForUnitTests() {
-        return newBuilder()
+        // TODO: should I move this into a test-artifact that uses package private access?
```
rcaudy, line 27: 👍

engine/context/src/main/java/io/deephaven/engine/context/ExecutionContext.java
rcaudy, line 130: I think it makes sense to use a multi-context, and only allow this to narrow things, at least when used by users.
We can do a little bit of accidental mis-use checking by examining the stack trace.

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -81,6 +81,8 @@ public ConsoleServiceGrpcImpl(final TicketRouter ticketRouter,
     public void getConsoleTypes(final GetConsoleTypesRequest request,
             final StreamObserver<GetConsoleTypesResponse> responseObserver) {
         GrpcUtil.rpcWrapper(log, responseObserver, () -> {
+            // for legacy reasons, this method is used before the session is created
```
rcaudy, line 84: I think in the brave-new exec context world, we can un-do this pattern and allow multiple sessions or language selection from the GUI.

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
rcaudy, line 155: Wrong privilege.

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
rcaudy, line 189: Wrong priv.

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -217,15 +219,22 @@ private static FieldInfo makeVariableDefinition(String title, String type) {
 
     @Override
     public void cancelCommand(CancelCommandRequest request, StreamObserver<CancelCommandResponse> responseObserver) {
-        // TODO not yet implemented, need a way to handle stopping a command in a consistent way
-        super.cancelCommand(request, responseObserver);
+        GrpcUtil.rpcWrapper(log, responseObserver, () -> {
+            final SessionState session = sessionService.getCurrentSession();
+            session.getAuthContext().requirePrivilege(ConsoleServicePrivilege.CAN_CANCEL_COMMAND);
+
+            // TODO not yet implemented, need a way to handle stopping a command in a consistent way
```
rcaudy, line 226: Can we add a ticket?

server/src/main/java/io/deephaven/server/console/ConsoleServicePrivilege.java
```diff
@@ -0,0 +1,7 @@
+package io.deephaven.server.console;
+
+import io.deephaven.auth.Privilege;
+
+public enum ConsoleServicePrivilege implements Privilege {
+    CAN_GET_CONSOLE_TYPES, CAN_START_CONSOLE, CAN_GET_HEAP_INFO, CAN_SUBSCRIBE_TO_LOGS, CAN_EXECUTE_COMMAND, CAN_CANCEL_COMMAND, CAN_BIND_TABLE_TO_VARIABLE, CAN_AUTO_COMPLETE,
```
rcaudy, line 6: I prefer one per line, with JavaDoc for each constant.

server/src/main/java/io/deephaven/server/console/SessionToExecutionStateModule.java
```diff
@@ -2,13 +2,15 @@
...
+import io.deephaven.auth.AuthContext;
 import io.deephaven.engine.context.ExecutionContext;
 import io.deephaven.engine.util.ScriptSession;
 
 @Module
 public class SessionToExecutionStateModule {
     @Provides
     ExecutionContext bindExecutionContext(ScriptSession session) {
-        return session.getExecutionContext();
+        // TODO: users should be able to provide an auth context that is used for start up scripts, etc
```
rcaudy, line 13: Ticket or implement.

server/src/main/java/io/deephaven/server/table/ops/ApplyPreviewColumnsGrpcImpl.java
```diff
@@ -23,7 +25,10 @@ protected ApplyPreviewColumnsGrpcImpl() {
     }
 
     @Override
-    public Table create(ApplyPreviewColumnsRequest request, List<SessionState.ExportObject<Table>> sourceTables) {
+    public Table create(final AuthContext authContext,
+            final ApplyPreviewColumnsRequest request,
+            final List<SessionState.ExportObject<Table>> sourceTables) {
+        authContext.requirePrivilege(TableServicePrivilege.CAN_APPLY_PREVIEW_COLUMNS);
```
rcaudy, line 31: I wonder what a user is supposed to do if this fails. Should we drop the columns that would be previewed?

rcaudy (commented):
Good start. I think we should consider a bit of a refactoring to allow differentiation on user, and I'm not totally clear what the entry point is for initial auth context installation.

niloc132 (commented):
I'd like to suggest that instead of authoring an enum value per service rpc call, we instead generate the enums and their values. I think this could happen at build time (of :server itself, so that it can compile against :authentication and its types, plus keep references to the session), with the same pattern of one enum type per service, and one value per rpc call. Naming seems to be almost something we can automate (with some very slight concessions, could totally automate it), and making an interceptor means that trivially we would automatically call a matching method (and never have the mismatches that the current impl has).

This could be accomplished most easily without reflection, but by referencing the service descriptors that were already generated. Those give us access to service calls (and info about them), so we can emit the appropriate java type.

In theory we could also skip the enum entirely, and let Privilege be an object with properties that describe it (plus probably some handy methods to make one from descriptors or just strings, which could be validated to ensure they belong to a service). That would entirely remove the need for code generation, as the descriptor for the service call is available in an interceptor.

Going the other direction, a more complex "can the user do this" check could be implemented by making something generic on the request type - presumably a lambda could be a Privilege, and expand the contains check to actually invoke it with the current request object. Even if we don't need this out of the box, I definitely believe we'll need it sooner than later, so planning now for it would be a good idea. (Achieving this without codegen is possible, just requires that we reflect to find out the request type at runtime, and the lambdas or whatever declare the type they expect). This use case was addressed in my initial summary of "things authorization should probably be able to do", and I see Devin has also mentioned it in this review.


--

Two other general notes:
 * I'd like to see the authorization stuff split out from the `:authentication` project (and the `auth` package split too, so that we don't break jpms rules), offering APIs for downstream projects to implement as they desire.
 * I'd also like to see either a general RBAC tool built in, that maps roles to privileges, or a "this is one possible way to implement privileges" project that shows how to get to RBAC from what we have here. My experience suggests that while privileges should be our "unit of what the user can do", roles should be the mapping to identity (potentially with "groups" as an intermediate piece) so that we can configure _in deephave-core_ the set or privileges that a set of users might have, and allow whatever "auth module" to merely specify roles that the current usercontext holds. I don't have an off-the-shelf tool in mind for this, but https://en.wikipedia.org/wiki/Role-based_access_control documents this pretty well (search for "the following conventions").

server/src/main/java/io/deephaven/server/table/ops/ComboAggregateGrpcImpl.java
```diff
@@ -109,8 +111,9 @@ private boolean isSimpleAggregation(ComboAggregateRequest request) {
     }
 
     @Override
-    public Table create(final ComboAggregateRequest request,
+    public Table create(final AuthContext authContext, final ComboAggregateRequest request,
             final List<SessionState.ExportObject<Table>> sourceTables) {
```
cpwright, line 115: Presumably, if I can reference something that is in sourceTable's I've already gotten an export for it, and any mutations (row/column filtering) that authorization would apply have been applied?

authentication/build.gradle
```diff
@@ -6,6 +6,7 @@ plugins {
 description 'authentication: Deephaven authentication and identity'
 
 dependencies {
+    api project(':Util')
```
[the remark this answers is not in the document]
niloc132: fwiw, it used to be that AuthContext was in `:Util`, so it was necessary to export this. Now we need to have `:Base` instead, so that LogOutputAppendable can be exposed.

nbauernfeind:
Closing in favor of #3079, which addresses the majority of comments via a pretty different approach.

PR #2933, opened 2022-09-29
Always create an empty array, in case the provided columns are null

### LINKED ISSUE #2932: JSAPI Stream tables setViewport without columns throws
**Description**

When setting a viewport in the JS API for a stream table, if the table columns are omitted, the followup `getViewportData` call will fail.

```js
table.setViewport(0, 1); // 3rd param columns omitted
const data = await table.getViewportData();
```

**Versions**
 - Deephaven: 0.17

niloc132:
Fixes #2932


mattrunyon (approved):
Works w/ docs snapshot tool and works with web UI

Issue #2890, opened 2022-09-21
`PartitionedTableProxy` is missing table methods, leading to a poor UX


chipkent:
`PartitionedTableProxy` is missing table methods, which results in a very poor user experience.

Consider this case.  `merge` and `drop_columns` are missing from `PartitionedTableProxy`.  As a result, I had to:
1. Explicitly add `.target` 
2. Move `merge` below `.target`
3. Move `drop_columns` below `.target`

```python
from deephaven import time_table
from deephaven.updateby import ema_time_decay

t1 = time_table("00:00:01").update(["Label = ii%2", "X = ii%2 == 0 ? cos(0.1*ii) : sin(0.01*ii)"])

t2 = t1.partition_by(["Label"]).proxy() \
    .update(["Data=X_.subVector(ii-14,ii)"]) \
    .update(["H=max(Data)"]) \
    .update(["L=min(Data)"]) \
    .update(["ATR=max(H-L, abs(H-X), abs(L-X))"]) \
    .target \
    .merge() \
    .drop_columns("Data")
```

These rough edges break the user view that partitioned tables are "just tables broken into smaller parts that can be operated on separately and then recombined".

rcaudy:
- Proxies behave as `TableOperations`. It's a more limited interface than `Table`, because you *cannot* interrogate for schema, or try to pull data out. If you'd like to add more methods to `TableOperations` (like `dropColumns`), write an issue for that and we can debate/prioritize. Note that this is the same interface used for client-side proxies. We intentionally constrain the operations to those that make sense when proxied.
- `merge` isn't a `Table` method, it's a `PartitionedTable` method. I strongly dislike the idea of muddying the interface layers by making `merge()` work on a proxy directly. I spent years cleaning up after similar "convenience" functions in other places.

chipkent:
Here is an organized list of what is not currently supported:

**Should support**
- where_one_of
- lazy_update
- tail_by
- head_by
- head_pct
- tail_pct
- ungroup
- drop_columns
- rename_columns
- move_columns
- move_columns_up
- move_columns_down

**Should not support:**
- size
- columns
- coalesce
- restrict_sort_to
- to_string
- meta_table

**Probably not support**
- partition_by
- snapshot_history

**Not sure if we should support (probably not)**
- format_columns
- format_column_where
- format_row_where
- layout_hints

rcaudy:
I think we can mostly agree on this categorization. A few notes:
[...]

PR #2936, opened 2022-09-29
Plumb ungroup through TableOperations. 

devinrsmith:
Ensure empty array uses all array/vector columns.

### DIFF
--- a/table-api/src/main/java/io/deephaven/api/TableOperations.java
+++ b/table-api/src/main/java/io/deephaven/api/TableOperations.java
@@ -1246,16 +1246,53 @@ public interface TableOperations<TOPS extends TableOperations<TOPS, TABLE>, TABL
 
     // -------------------------------------------------------------------------------------------
 
+    /**
+     * Ungroups a table by expanding all columns of arrays or vectors into columns of singular values, creating one row
+     * in the output table for each value in the columns to be ungrouped. Columns that are not ungrouped have their
+     * values duplicated in each output row corresponding to a given input row. All arrays and vectors must be the same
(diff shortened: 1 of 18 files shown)


devinrsmith:
Related to https://github.com/deephaven/deephaven-core/issues/2890

table-api/src/main/java/io/deephaven/api/TableOperations.java
```diff
@@ -1245,4 +1245,24 @@ TOPS updateBy(UpdateByControl control, Collection<? extends UpdateByOperation> o
...
+    TOPS ungroup(boolean nullFill, String... columnsToUngroup);
+
+    /**
+     * Ungroups a table by converting arrays into columns.
+     *
+     * @param nullFill indicates if the ungrouped table should allow disparate sized arrays filling shorter columns with
+     *        null values. If set to false, then all arrays should be the same length.
+     * @param columnsToUngroup the columns to ungroup
+     * @return the ungrouped table
+     */
```
rcaudy, line 1264: This is an excuse to update the JavaDoc and copy it to the other overloads. For the basic case:
```
Ungroups a table by expanding columns of arrays or vectors into columns of singular values, creating one row in the output table for each value in the columns to be ungrouped. Columns that are not ungrouped have their values duplicated in each output row corresponding to a given input row.
```
We want to be clear that it's arrays *and* vectors.

devinrsmith: Done

rcaudy (approved):

PR #2941, opened 2022-10-03
Switch over developer prepareCompose workflow to jetty

devinrsmith:
Tested w/ `./gradlew prepareCompose; docker compose up -d`


niloc132 (commented):
Main feedback is that i think we might want to also rename/alias server to server-netty, and consider making server instead point to server-jetty (or leave it for one more version to avoid breaking anyone using deephaven/server and expecting netty). In theory there should be no difference, since the current server-jetty-app only does _more_ than server-netty-app, but perhaps worth doing this work to be clear about the transition period?

devinrsmith:
The big noticable different is port 8080 vs port 10000.

devinrsmith:
Instead of disabling some and leaving others, it was easiest to rename all of the relevant server images w/ the `-netty` suffix.

docker/server-jetty/build.gradle
```diff
@@ -0,0 +1,76 @@
...
+    from(configurations.serverApplicationDist) {
+        into 'serverApplicationDist'
+    }
+    from(dockerLicenses.outputs) {
+        into 'licenses'
+    }
+    into context
+}
+
+def buildDockerClosure = { String base, String server ->
```
rcaudy, line 65: The netty build has:
```
// TODO(deephaven-core#1596): Remove extra dependencies for build-ci.yml
// Currently, GH build-ci.yml calls prepareDockerAll, and relies on buildx; so we need to
// make sure that dockerCreateDockerfile has the proper dependencies.
// If we change how our build works in the future (so that gradle is responsible for pushing the
// image), we can remove this extra dependency.
project.tasks.register('prepareDockerAll') {
    dependsOn prepareDocker, baseMap.keySet().collect { base -> Docker.registryTask(project, base) }
}
```

Why don't we need this, just because we're still using netty for CI?

rcaudy (commented):
At a cargo-cult review quality level, this seems fine. If I understand correctly, for now we're using jetty for local development and the most simple images we'll publish, and using netty for build/CI/complex images.

docker/server-jetty/build.gradle
devinrsmith: This was mostly a copy-paste from the netty logic; I'm not sure if this specific block is still necessary since we don't have an CI artifacts for docker + jetty in deephaven-core.

There are non-trivial dependencies beyond for other build related things that depend on server-netty. For example, sphinx. I'm betting these will be "easy" to fix over to server-jetty; but it doesn't serve the purpose of the release IMO so I tried to limit the scope as much as possible.

Technically, there are a bunch of folder renaming that I didn't need to do - but I did anyways, just to keep things clearer w/ the universal -netty suffix.

The build logic in server-jetty is meant to serve local development purposes for `./gradlew prepareCompose; docker compose up -d` workflows.

build.gradle
```diff
@@ -59,11 +59,7 @@ tasks.register('nightly') {
 tasks.register('prepareCompose') {
     it.group 'Deephaven lifecycle'
     it.description 'A lifecycle task that prepares prerequisites for local docker-compose builds'
-    it.dependsOn project(':docker-server-slim').tasks.findByName('buildDocker-server-slim'),
-            project(':docker-server').tasks.findByName('buildDocker-server'),
-            project(':docker-web').tasks.findByName('buildDocker'),
-            project(':envoy').tasks.findByName('buildDocker'),
```
niloc132, line 65: to confirm, CI will still build and publish these, right?

devinrsmith: Correct.

buildSrc/src/main/groovy/io/deephaven/tools/docker/DeephavenInDockerExtension.groovy
```diff
@@ -45,7 +45,7 @@ public abstract class DeephavenInDockerExtension {
         // irritating configuration order of operations to work out here, so just leaving
         // these as constants until we decide they aren't any more
         deephavenServerProject = ':docker-server'
-        serverTask = 'buildDocker-server'
```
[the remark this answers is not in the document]
devinrsmith: Agreed.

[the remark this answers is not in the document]
niloc132: https://github.com/deephaven/deephaven-core/issues/2943

niloc132 (approved):

mofojed (approved):

PR #2950, opened 2022-10-05
Bump guava

devinrsmith:
Previous dependency resolution inherited from:

com.google.guava:guava:31.0.1-android
+--- io.grpc:grpc-api:1.46.0


devinrsmith:
Question: should we depend on the `-jre` or `-android` flavor?

rcaudy (approved):

niloc132:
There may come a time when we want android, but I don't think it will be terrible soon.

PR #2961, opened 2022-10-05
Upload error logs on pydoc failure

devinrsmith:
This is meant to catch JVM error logs of the format `**/*_pid*.log`. This is something we've seen from the pydoc task / sphinx / jpy.

### DIFF
--- a/.github/workflows/docs-ci.yml
+++ b/.github/workflows/docs-ci.yml
@@ -140,6 +140,14 @@ jobs:
           remote_user: ${{ secrets.DOCS_USER }}
           remote_key: ${{ secrets.DEEPHAVEN_CORE_SSH_KEY }}
 
+      - name: Upload JVM Error Logs
+        uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: docs-ci-pydoc-jvm-err
+          path: '**/*_pid*.log'
+          if-no-files-found: ignore
+
   cppdoc:
     runs-on: ubuntu-22.04
     steps:


.github/workflows/docs-ci.yml
```diff
@@ -140,6 +140,14 @@ jobs:
           remote_user: ${{ secrets.DOCS_USER }}
           remote_key: ${{ secrets.DEEPHAVEN_CORE_SSH_KEY }}
 
+      - name: Upload JVM Error Logs
+        uses: actions/upload-artifact@v3
+        if: failure()
```
mofojed, line 145: Does this need to be in double `${{ }}` ? Like https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-status-check-functions

mofojed (changes requested):
Update the PR description with more details, such as "What logs are uploaded?" "Where do these logs go/how do I check them in event of a failure?"

.github/workflows/docs-ci.yml
devinrsmith: This example was essentially copied from other github examples. Maybe they recommend always using `${{ ... }}` now, but it _does_ work without. If we want to update the syntax, we should probably do it everywhere.

https://docs.github.com/en/actions/learn-github-actions/expressions#about-expressions
> When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression.

mofojed (approved):

PR #2962, opened 2022-10-05
QueryCompiler: Use SHA-256 hash instead of String#hashCode on class body

### LINKED ISSUE #2949: QueryCompiler collision issues?
See:
https://gitlab.deephaven.io/illumon/iris/-/merge_requests/6879/diffs

We should reproduce and port the fix if appropriate. It's apparently easy to repro by changing the `packageNameSuffix` calculation, as in:
```
final String packageNameSuffix = "c" + pi + "v" + JAVA_CLASS_VERSION;
```

Note that the improved hash in the code to port is not an essential part of the fix, it just reduces the probability of collisions. The key is not allowing a duplicate result.

nbauernfeind:
Fixes #2949.

### DIFF
--- a/Util/src/main/java/io/deephaven/util/ByteUtils.java
+++ b/Util/src/main/java/io/deephaven/util/ByteUtils.java
@@ -12,7 +12,7 @@ public class ByteUtils {
 
         for (int ii = 0; ii < bytes.length; ii++) {
             // extract the upper 4 bits
-            buffer[ii << 1] = HEX_LOOKUP[bytes[ii] >>> 4];
+            buffer[ii << 1] = HEX_LOOKUP[(bytes[ii] >> 4) & 0xF];
             // extract the lower 4 bits
             buffer[(ii << 1) + 1] = HEX_LOOKUP[bytes[ii] & 0xF];
         }
--- a/Util/src/main/java/io/deephaven/util/ByteUtils.java
+++ b/Util/src/main/java/io/deephaven/util/ByteUtils.java
@@ -1,0 +1,23 @@
+/**
+ * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+ */
+package io.deephaven.util;
+
+public class ByteUtils {
+    private static final char[] HEX_LOOKUP = new char[] {
+            '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'};
+
+    public static String byteArrToHex(byte[] bytes) {
+        // our output size will be exactly 2x byte-array length
+        final char[] buffer = new char[bytes.length * 2];
+
+        for (int ii = 0; ii < bytes.length; ii++) {
+            // extract the upper 4 bits
+            buffer[ii << 1] = HEX_LOOKUP[(bytes[ii] >> 4) & 0xF];
+            // extract the lower 4 bits
+            buffer[(ii << 1) + 1] = HEX_LOOKUP[bytes[ii] & 0xF];
+        }
+
+        return new String(buffer);
+    }
+}
--- a/Util/src/main/java/io/deephaven/util/ByteUtils.java
+++ b/Util/src/main/java/io/deephaven/util/ByteUtils.java
@@ -4,8 +4,7 @@
 package io.deephaven.util;
 
 public class ByteUtils {
-    private static final char[] HEX_LOOKUP = new char[] {
-            '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'};
+    private static final char[] HEX_LOOKUP = "0123456789abcdef".toCharArray();
 
     public static String byteArrToHex(byte[] bytes) {
         // our output size will be exactly 2x byte-array length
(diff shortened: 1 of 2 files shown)


engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -488,7 +495,12 @@ private Class<?> compileHelper(@NotNull final String className,
                 knownClasses.put(identifyingFieldValue, p);
                 // It's also possible that some other code has already fulfilled this promise with exactly the same
                 // class. That's ok though: the promise code does not reject multiple sets to the identical value.
-                p.complete(result);
+                // But if the result is already set, that means that we've already been through this logic, determined
+                // the class, and set it (each run through this logic can be using a different ClassLoader, so it may
+                // be a different instance of the same Class).
+                if (!p.complete(result)) {
+                    log.warn("Found pre-existing class result for " + fqClassName);
```
devinrsmith, line 502: need `endl()`

engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -488,7 +495,12 @@ private Class<?> compileHelper(@NotNull final String className,
                 knownClasses.put(identifyingFieldValue, p);
                 // It's also possible that some other code has already fulfilled this promise with exactly the same
                 // class. That's ok though: the promise code does not reject multiple sets to the identical value.
-                p.complete(result);
+                // But if the result is already set, that means that we've already been through this logic, determined
+                // the class, and set it (each run through this logic can be using a different ClassLoader, so it may
+                // be a different instance of the same Class).
+                if (!p.complete(result)) {
```
rcaudy, line 501: We need to have a bit of a think about something, here. We "accidentally" fixed the issue Al reported when we switched to `CompletableFuture`, since it won't let us override the result. That said, that invalidates Corey's original commentary about `identifyingFieldValue`. If we will never overwrite the class in the future, we should also not be changing the key in `knownClasses`.

We should switch from `remove`+`put` to `get` followed by `remove` + `put` if and only if we succeed in `complete`.
That is, I think the code should be changed to this:
```
                CompletableFuture<Class<?>> future = knownClasses.get(identifyingFieldValue);
                final boolean found = future != null;
                if (!found) {
                    future = new CompletableFuture<>();
                }
                final boolean completed = future.complete(result);
                if (completed) {
                    if (found) {
                        knownClasses.remove(identifyingFieldValue);
                    }
                    knownClasses.put(identifyingFieldValue, future);
                }
```

Util/src/main/java/io/deephaven/util/ByteUtils.java
```diff
@@ -0,0 +1,23 @@
+/**
+ * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+ */
+package io.deephaven.util;
+
+public class ByteUtils {
+    private static final char[] HEX_LOOKUP = new char[] {
+            '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'};
+
+    public static String byteArrToHex(byte[] bytes) {
```
rcaudy, line 10: I feel like this *must* exist in some library we use already.

Util/src/main/java/io/deephaven/util/ByteUtils.java
```diff
@@ -0,0 +1,23 @@
...
+        // our output size will be exactly 2x byte-array length
+        final char[] buffer = new char[bytes.length * 2];
+
+        for (int ii = 0; ii < bytes.length; ii++) {
+            // extract the upper 4 bits
+            buffer[ii << 1] = HEX_LOOKUP[(bytes[ii] >> 4) & 0xF];
```
rcaudy, line 16: Is this (using shift right logical) better?
```suggestion
            buffer[ii << 1] = HEX_LOOKUP[bytes[ii] >>> 4];
```

rcaudy (changes requested):

Util/src/main/java/io/deephaven/util/ByteUtils.java
nbauernfeind: I was using this for reference: https://stackoverflow.com/a/58118078/1028765

The ExecutionContext module is pretty lean. We could add a dependency on guava's impl, but the post says this one is about 4x faster than guavas. DHE is using commons-codec, and that post says (more or less) to avoid that impl (although it is 2-3x faster than the guava impl).

devinrsmith: We recently moved away from commons-codec elsewhere, and leaned on `io.deephaven.proto.util.ByteHelper#byteBufToHex`. That said, I don't want proto-backplane-grpc depending on Util, nor does it make sense the other way. I think it's okay to have this duplication. (If we really want, we can start a new "byte util" module that has zero dependencies, but that's overkill right now IMO.)

devinrsmith:
I tried to remove myself as a reviewer, but I'm not sure I can? Anyways, leaving for Ryan.

rcaudy (approved):

PR #2968, opened 2022-10-06
Bump dagger to 2.44

devinrsmith:
The api dependency is a good find by dagger - GrpcFilter extends HttpFilter, so HttpFilter should be part of public api.

For reference, this is the error without the api dependency:

  Dependency trace:
      => element (CLASS): io.deephaven.server.jetty.GrpcFilter
      => type (ERROR superclass): jakarta.servlet.http.HttpFilter

  If type 'jakarta.servlet.http.HttpFilter' is a generated type, check above for compilation errors that may have prevented the type from being generated. Otherwise, ensure that type 'jakarta.servlet.http.HttpFilter' is on your classpath.


nbauernfeind (approved):
Neat feature!

PR #2970, opened 2022-10-06
Bump kafka clients and avro

devinrsmith:


jcferretti (approved):

devinrsmith:
Leaving this here for searchability.

`org.apache.kafka.common.errors.UnsupportedVersionException: The broker does not support INIT_PRODUCER_ID`

If you see this error message, you may need to upgrade the server as the latest clients seem to take it as a given. If this becomes a strong blocker to usability, we can look into our code connection paths to see if we can change this behavior, or revert to an older client.

PR #2971, opened 2022-10-06
Auto create docs issues from closed PRs

mattrunyon:
This adds a workflow which will automatically open corresponding issues in the Deephaven docs repo for PRs closed with the `DocumentationNeeded` label. Since we're creating issues to a private repo, we need to use a PAT with repo access in the github secrets

I tested this in a dummy org to ensure the workflow works as expected
Test org: https://github.com/orgs/mattrunyon-test-org/repositories

I used `test-public` as the core repo and `test-private` as docs. I just made `test-private` public so you can see it


devinrsmith (commented):
I've added the appropriate secret. I'll let @mofojed take a pass for final approval.

.github/workflows/create-docs-issues.yml
```diff
@@ -0,0 +1,53 @@
...
+
+            const promises = issueTypes.map(type => github.rest.issues.create({
+              owner: context.repo.owner,
+              repo: ISSUE_REPO_NAME,
+              title: `${type.toUpperCase()}: ${context.payload.pull_request.title}`,
+              body,
+              labels: [type, 'documentation', 'triage', 'autogenerated']
+            }));
+
+            const issues = await Promise.allSettled(promises);
```
mofojed, line 39: What happens if there's an error creating the tickets - action fails and we see the error there?

.github/workflows/create-docs-issues.yml
```diff
@@ -0,0 +1,53 @@
...
+              const { status, value, reason } = promise;
+              const link = status === 'fulfilled' ? value.data.html_url : `Failed to create issue: ${reason}`;
+              return `${type[0].toUpperCase() + type.slice(1)}: ${link}`
+            });
+
+            await github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'Labels indicate documentation is required. Issues for documentation have been opened\n\n' +
```
mofojed, line 51: Feel like this should have a `:` at the end before the two new line chars

.github/workflows/create-docs-issues.yml
mofojed: Nevermind - it'll appear in the comment. Cool.

mofojed (dismissed):

mofojed (approved):

Issue #2732, opened 2022-08-12
TestTableTools.testMergeIndexShiftingPerformance test failure


devinrsmith:
Failure in the task `engine-table:testOutOfBand`

https://github.com/deephaven/deephaven-core/runs/7800870316?check_suite_focus=true

> junit.framework.AssertionFailedError: This test is expected to take around 5 seconds on a Mac with the new shift behavior, something is not right.

Looks like things get things get slow at step 125.

```
Step=0, duration=7ms, stepDuration=7ms	
Step=1, duration=17ms, stepDuration=10ms	
Step=2, duration=20ms, stepDuration=3ms	
Step=3, duration=23ms, stepDuration=3ms	
Step=4, duration=27ms, stepDuration=4ms	
Step=5, duration=32ms, stepDuration=5ms	
Step=6, duration=37ms, stepDuration=5ms	
Step=7, duration=43ms, stepDuration=6ms	
Step=8, duration=49ms, stepDuration=6ms	
Step=9, duration=57ms, stepDuration=8ms	
Step=10, duration=67ms, stepDuration=10ms	
Step=11, duration=75ms, stepDuration=8ms	
Step=12, duration=88ms, stepDuration=13ms	
Step=13, duration=104ms, stepDuration=16ms	
Step=14, duration=119ms, stepDuration=15ms	
Step=15, duration=137ms, stepDuration=18ms	
Step=16, duration=158ms, stepDuration=21ms	
Step=17, duration=181ms, stepDuration=23ms	
Step=18, duration=194ms, stepDuration=13ms	
Step=19, duration=224ms, stepDuration=30ms	
Step=20, duration=256ms, stepDuration=32ms	
Step=21, duration=269ms, stepDuration=13ms	
Step=22, duration=310ms, stepDuration=41ms	
Step=23, duration=352ms, stepDuration=42ms	
Step=24, duration=365ms, stepDuration=13ms	
Step=25, duration=408ms, stepDuration=43ms	
Step=26, duration=424ms, stepDuration=16ms	
Step=27, duration=439ms, stepDuration=15ms	
Step=28, duration=456ms, stepDuration=17ms	
Step=29, duration=499ms, stepDuration=43ms	
Step=30, duration=517ms, stepDuration=18ms	
Step=31, duration=535ms, stepDuration=18ms	
Step=32, duration=582ms, stepDuration=47ms	
Step=33, duration=610ms, stepDuration=28ms	
[...]

PR #2974, opened 2022-10-06
Devroot cleanup

devinrsmith:
This removes the necessity for the `devroot` property to be set - it doesn't make sense for end-users. In code-generation or test cases where access to the source is still necessary, the appropriate system properties can be set as necessary.


Generators/src/main/java/io/deephaven/libs/GroovyStaticImportGenerator.java
```diff
@@ -373,15 +372,13 @@ public static void main(String[] args) throws ClassNotFoundException, IOExceptio
...
-            devroot = Configuration.getInstance().getDevRootPath();
-        } else if (args.length == 1) {
+        if (args.length == 1) {
             devroot = args[0];
         } else if (args.length == 2) {
             devroot = args[0];
             assertNoChange = Boolean.parseBoolean(args[1]);
         } else {
-            System.out.println("Usage: [<devroot> [assertNoChange]]");
+            System.out.println("Usage: <devroot> [assertNoChange]");
```
rcaudy, line 381: Have you tested all of the generators?

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -56,9 +56,7 @@ public class PythonDeephavenSession extends AbstractScriptSession<PythonSnapshot
     private static final Logger log = LoggerFactory.getLogger(PythonDeephavenSession.class);
 
     private static final String DEFAULT_SCRIPT_PATH = Configuration.getInstance()
-            .getProperty("PythonDeephavenSession.defaultScriptPath")
-            .replace("<devroot>", Configuration.getInstance().getDevRootPath())
-            .replace("<workspace>", Configuration.getInstance().getWorkspacePath());
+            .getStringWithDefault("PythonDeephavenSession.defaultScriptPath", ".");
```
rcaudy, line 59: Are we confident about making `.` the new default for all the script sessions? What testing have you done?

rcaudy, line 59: Related to this change:
`io.deephaven.engine.util.ScriptFinder#findScriptEx(java.lang.String, java.lang.String)` has some comments that probably need cleanup after devroot is retired.

devinrsmith: This previous defaults were `<X>.defaultScriptPath=<devroot>`; and by default, `devroot=.`. So in theory, it's the same defaults as before, just a layer of indirection removed.

I think the defaultScriptPath logic could use serious reworking - I'm not sure it makes sense anymore if we have a better way w/ application mode. There's a comment to this effect in `io.deephaven.engine.util.ScriptFinder#findScriptEx`

I've run locally natively in python and groovy. I'll do testing in docker as well. I'll start a nightly as well.

devinrsmith: Ha! Just mentioned this in the previous comment before I noticed this.

Generators/src/main/java/io/deephaven/libs/GroovyStaticImportGenerator.java
devinrsmith: I've run `./gradlew Generators:generateAll` and I think testing should also assert all the generators are correct.

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
devinrsmith: Everything I've run seems to work. nightly is passing except I hit this https://github.com/deephaven/deephaven-core/issues/2732

devinrsmith: Also run python embedded server, work fine.

devinrsmith: https://github.com/deephaven/deephaven-core/issues/2978

rcaudy (approved):

jcferretti (approved):

PR #2977, opened 2022-10-07
Add jetty all-ai image

JamesXNelson:


JamesXNelson:
Note, this is probably NOT the best way to do this, I'm mostly just opening this to share code w/ JJ and track this changeset.

I'm not really supposed to be working today either, so feel free to punt on this until next week.

devinrsmith:
For reference, our published jetty images are built over at https://github.com/deephaven/deephaven-server-docker

JamesXNelson:
@devinrsmith this definitely won't work the way it is. It built for me earlier NP, but now I had to remove a `--no-index` to get a certain dependency, so we'll want to "fix this properly" when I'm back in on Tuesday.

JamesXNelson:
ok, we def can't merge this as-is. Had to remove `--no-index` flag to be able to download some dependency change. 

Need to do this properly with new base images.  Hopin' we can lean on you @devinrsmith to help make that happen easily.

JJ and I are now both able to deploy but we should NOT merge this `--no-index` change w/ this branch.

PR #2979, opened 2022-10-07
ConfigService gRPC and JS API

### LINKED ISSUE #2668: Feature suggestion: Configuration gRPC service
Subtask for #2564 

As a GUI client author/consumer, I would like to know what authentication methods are supported/suggested by the server, and any values to have suggested when trying to use them.

As stated, this will come in the form of "before I manage to authenticate, what should I know", but it might also make sense to provide a "after I have successfully authenticated, what should I know" to provide other details about the server and how it is set up. Script session language, runtime version of various tools are some examples that might be gated by authentication settings, DHE also includes features like enabled/deployed plugins, custom logo/docs/help info.

Example use cases for pre-auth config:
 * Server only has SAML enabled, user shouldn't first attempt anon access, then username/password, etc
 * Server has multiple SAML providers, user should pick from the list rather than guessing what IDP they need to visit for their own account
 * Server configuration is set to allow and prefer anon access, but permit some other authentication, to encourage quick conversion of users, but support richer experience for authenticated users
 * Server configuration is set to permit anon access, but always require them to click past the login prompt, so that they are first given shown that the experience will be better with an account

--

Proposed API would be a pair of unary gRPC calls, each of which returns a repeated pair of strings, for key and value. In time we could make value more generic, but we should probably map as closely as possible to DHE (for compat in deephaven/web-client-api), and as closely to the existing config system (to allow administrators so add their own configuration settings that will be accessible by clients. Could also offer a single method, and have it return different values pre- and post- authentication, but it might be clearer to explicitly provide two (as DHE does internally, so that the JS API can expose them differently to the API consumer).
```
service ConfigService {
  rpc GetAuthenticationConstants(AuthenticationConstantsRequest) returns (AuthenticationConstantsResponse) {}
  rpc GetConfigurationConstants(ConfigurationConstantsRequest) returns (ConfigurationConstantsResponse) {}
}

message AuthenticationConstantsRequest {}
message ConfigurationConstantsRequest {}
message AuthenticationConstantsResponse {
  repeated ConfigPair config_values = 1;
}
message ConfigurationConstantsResponse {
  repeated ConfigPair config_values = 1;
}
message ConfigPair {
  string key = 1;
  oneof value {//leave room for more types
    string string_value = 2;
  }
}
```

Server configuration then will have two new properties, one to list pre-auth config properties to provide to clients, and one for post-auth config properties.

niloc132:
Adds a new gRPC service that provides access to a limited set of configuration properties to unauthenticated users, and a different set to authenticated users.

Includes two drafts of a mechanism to read versions of various jars on the classpath. The goal in either case is to read the runtime version, rather than picking up the version that dhc was compiled against, in case of some version management/conflicts in downstream projects.

There are three "kinds" of versions we collect at this time:
 * `java.version` is special, and is set by the jvm itself
 * `barrage.version` or other external dependencies, where we need to read a version value from the jar. Barring the jar containing a specific file to indicate version, there are roughly two standard ways to do this, both supported by barrage (and an example of each is included):
    * Maven publishes a file at `/META-INF/maven/$groupId/$artifactId/pom.properties` from which the `version` property can be read. This doesn't work for jars built outside of maven (such as gradle, or via `mvn deploy:deploy-file`). Upside is that you can attempt to read it by GAV coords across the classpath.
    * Java jars include a manifest where the `Implementation-Version` can optionally be added, allowing it to be read from `ClassInThatJar.class.getPackage().getImplementationVersion()`. This is also optional, and many of our jars don't include it (very quick non-scientific survey showed it was missing from flatbuffers, guava, gwt, deephaven-hash). I'm leaning towards this.
 * `deephaven.version` is the version of the build itself, so could be baked in to a properties file instead of relying on the above. 

In order to keep versions available as a deephaven configuration value, they must be set as system properties at runtime, either iterated and set before Configuration is first init'd, or explicitly set on the configuration instance. The commented out code in Main.java is an example of the former, while the ConfigServiceGrpcImpl.java demonstrates the latter. I'm leaning towards the latter, as it makes it possible to read from configuration... to know what versions should be published _to_ configuration. 

The current implementation of reading versions from configuration works by expecting a list property be populated with key=fully.qualified.ClassName pairs. That will result in `$key.version` being published, by finding the Class instance based on the name (_without_ initializing the class), then getting the Implementation-Version key from there. (This could be adjusted to read the pom.properties instead, given a GAV instead of a fqcn).

Fixes #2668


proto/proto-backplane-grpc/src/main/proto/deephaven/proto/config.proto
```diff
@@ -0,0 +1,35 @@
...
+option go_package = "github.com/deephaven/deephaven-core/go/internal/proto/config";
+
+/**
+ * Provides simple configuration data to users. Unauthenticated users may call GetAuthenticationConstants
+ * to discover hints on how they should proceed with providing their identity, while already-authenticated
+ * clients may call GetConfigurationConstants for details on using the platform.
+ */
+service ConfigService {
+  rpc GetAuthenticationConstants(AuthenticationConstantsRequest) returns (AuthenticationConstantsResponse) {}
+  rpc GetConfigurationConstants(ConfigurationConstantsRequest) returns (ConfigurationConstantsResponse) {}
```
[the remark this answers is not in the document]
niloc132: Yep, naming is here to be critiqued. I don't _think_ we want the server to push changes to these configuration values, so from a given client's perspective, we should assume that config values _received_ are functionally constants, but at the same time, there is nothing preventing the server from changing them at runtime without a restart.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/VersionLookup.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Resolves the runtime version of Barrage, in case it has been updated on the classpath beyond what this project was
+ * built with. Should be invoked before it will first be requested to ensure it is set properly.
+ */
+public class VersionLookup {
+    public static void lookup() {
+        String version;
+
+        try (InputStream is = BarrageMessageWrapper.class
+                .getResourceAsStream("/META-INF/maven/io.deephaven.barrage/barrage-format/pom.properties")) {
```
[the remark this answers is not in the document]
niloc132: Agreed, I just wanted more than one option to discuss, see description. I think this is the better option, it is _slightly_ more generally adopted, but you still can't assume it is present.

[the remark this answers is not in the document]
devinrsmith: I wonder if there is a way we can have gradle output a "manifest" file of all the GAV coordinates it actually uses at runtime (in the context of jetty-server-app, etc).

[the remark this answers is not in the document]
niloc132: (also discussed in slack)
Certainly possible, not even difficult - but it does but a burden on downstream consumers of the io.deephaven:server-jetty artifact to produce their own such artifact if they happen to manage any dependency differently that DHC itself specifies it. For example, if a bugfix release of barrage were to emerge that io.dh:server-jetty is compatible with, a project could specify that newer one in maven/ivy/sbt/gradle, but the version file wouldn't update unless we had scripts for their given build system.

In contrast, while the Package.getImplementationVersion() is not uniformly applied, it is part of Java itself, so a reasonable deployment should include it. In the event that a dependency doesn't include it but a project needs it, a "dummy" jar could be added to the project that exposes the dependency version, or set a regular configuration property and expose it through this service.

server/src/main/java/io/deephaven/server/runner/Main.java
```diff
@@ -78,6 +79,9 @@ public static Configuration init(String[] args, Class<?> mainClass) throws IOExc
 
         log.info().append("Starting up ").append(mainClass.getName()).append("...").endl();
 
+        // Version properties to bootstrap before we load configuration for the first time
+        // VersionLookup.lookup();
+        // System.setProperty("deephaven.version", mainClass.getPackage().getImplementationVersion());
```
mofojed, line 84: Unused, remove?

web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
```diff
@@ -54,7 +81,12 @@ public Promise<Void> relogin(String token) {
     }
 
     public Promise<String[][]> getServerConfigValues() {
-        return Promise.resolve(new String[0][]);
+        return getConfigs(
```
mofojed, line 84: Just making sure - these are general server information, and we don't expect the user to be authenticated before retrieving them.

mofojed (changes requested):
Just some unused code to remove, tested it out in console and looks good:
```
var coreClient = new dh.CoreClient('http://localhost:10000')
console.log('serverConfigValues:', await coreClient.getServerConfigValues())
console.log('authConfigValues:', await coreClient.getAuthConfigValues())
```

web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
niloc132: Ah good catch - the server will have to auth this and fail the user in this case. i'll double check what dhe does here, but I think this is the same?

server/src/main/java/io/deephaven/server/runner/Main.java
niloc132: Yep, only left in for discussion's sake. I'll remove the commented-out implementation since we seem to be on the same page.

web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
mofojed: Yes, on DHE it checks if you're authenticated for server config values: https://gitlab.deephaven.io/illumon/iris/-/blob/rc/silverheels/web/server-frontend/src/main/java/com/illumon/iris/web/server/WebApiServerImpl.java#L844
You can get the authConfigValues before being authenticated though (as you need them to know what you can login with): https://gitlab.deephaven.io/illumon/iris/-/blob/rc/silverheels/web/server-frontend/src/main/java/com/illumon/iris/web/server/WebApiServerImpl.java#L384
Also looks like there's a minor bug in DHE - if you try `client.getServerConfigValues()` before logged in, you'll get a `Not logged in` error (which makes sense), however if you then login with that same `client` object and call `client.getServerConfigValues()` afterward, you still get a `Not logged in` error. This is because the serverConfigValues is just a lazy that initializes the first time the get is called, and does not reset even if there was an error (or you login): https://gitlab.deephaven.io/illumon/iris/-/blob/rc/silverheels/web/client-api/src/main/java/com/illumon/iris/web/client/api/IrisClient.java#L80
Such a minor issue.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/config.proto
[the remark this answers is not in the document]
mofojed: On DHE the config values are certainly treated as constants. I'm alright if they change at runtime as long as there's a notification there's been a change, such that UI can re-load appropriately. It's definitely less complex if we just treat them as constants.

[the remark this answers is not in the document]
mofojed: If we do have changes, sending a disconnect also would force the client to reconnect/fetch the server config values again as well. I don't think we're expecting any use cases where config values change on the fly, I'd think it's easier to just do disconnect/reload everything if the user does want to change config values (which shouldn't happen often?)

web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
niloc132: I'm leaving this as is, re-querying the server each time, to avoid the bug you describe. The tradeoff is that if the client asks for the config again, they will get an extra round trip, and in theory could get different values, going by the current implementation. That might be a feature too, letting the client poll for values.

mofojed (changes requested):
I can still get the serverConfigValues without logging in first, as long as I get the auth constants first. I don't think that behaviour is consistent.

If I run:
```
var coreClient = new dh.CoreClient('http://localhost:10000')
```
then:
```
console.log('serverConfigValues:', await coreClient.getServerConfigValues())
```
I get an error. However, if I do:
```
var coreClient = new dh.CoreClient('http://localhost:10000')
console.log('authConfigValues:', await coreClient.getAuthConfigValues())
```
then:
```
console.log('serverConfigValues:', await coreClient.getServerConfigValues())
```

Then it works

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/config.proto
```diff
@@ -0,0 +1,35 @@
...
+}
+message ConfigurationConstantsResponse {
+  repeated ConfigPair config_values = 1;
+}
+message ConfigPair {
+  string key = 1;
+  oneof value {//leave room for more types
+    string string_value = 2;
+  }
+}
```
devinrsmith, line 35: So, I do want to talk about pros and cons of our choices here wrt https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/struct.proto. Is there benefit to relying on Struct/Value instead? Or, we could potentially mirror it with our own type and leave out the parts we don't need right now:

```proto
message Value {
  // The kind of value.
  oneof kind {
    // Represents a string value.
    string string_value = 3;
  }
}
```
This might leave us room in the future if we _did_ want to transition to upstream types.
Thoughts on `repeated` vs `map<string, `? We get a bit stronger of typing w/ a map.

niloc132: Agreed, in theory, will make the change.

I'm not sure what you mean by stronger typing though, the key is still a string the value is still whatever fits in the one-of/etc, which is still checked by tag. Indeed, the `map<k,v>` syntax is just sugar over a message with a `1` and `2` tag of the specified type, so key collisions can happen (though presumably the receiving side's runtime will prune the duplicates, like it would for non-repeated fields). Note especially the warning in the docs at https://developers.google.com/protocol-buffers/docs/proto3#backwards_compatibility:

> Any protocol buffers implementation that supports maps must both produce and accept data that can be accepted by the above definition.

Functionally, the only difference here is that the oneof is moved out to its own message type, so we just pay for an extra tag "go decode a value now" "here's the type of the data" instead of just "here's the type of the data".

devinrsmith: Ah, great link. I didn't realize the semantics of map were equivalent to repeated.

niloc132:
> I can still get the serverConfigValues without logging in first, as long as I get the auth constants first. I don't think that behaviour is consistent.
> 
> If I run:
> 
> ```
> var coreClient = new dh.CoreClient('http://localhost:10000')
> ```
> 
> then:
> 
> ```
> console.log('serverConfigValues:', await coreClient.getServerConfigValues())
> ```
> 
> I get an error. However, if I do:
> 
> ```
> var coreClient = new dh.CoreClient('http://localhost:10000')
> console.log('authConfigValues:', await coreClient.getAuthConfigValues())
> ```
> 
> then:
> 
> ```
> console.log('serverConfigValues:', await coreClient.getServerConfigValues())
> ```
> 
> Then it works

Good catch. This is a bug in the new authentication changes - apparently in the absence of an attempt to authenticate, the server assumes you meant to be anonymous and sends back the appropriate headers, and the client "always try to keep auth headers up to date" accepts those. Ill follow-up with that separately.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/config.proto
niloc132: Updated, PTAL.

niloc132:
Nope, I'm wrong - the client is incorrectly authenticating as anonymous automatically, nothing to do with the server. Once we make CoreClient the only way to get an IdeSession etc, this will go away.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/config.proto
```diff
@@ -0,0 +1,35 @@
...
+message AuthenticationConstantsResponse {
+  map<string, ConfigValue> config_values = 1;
+}
+message ConfigurationConstantsResponse {
+  map<string, ConfigValue> config_values = 1;
+}
+
+message ConfigValue {
+  oneof kind {//leave room for more types
+    string string_value = 2;
```
devinrsmith, line 33: Please set as value 3 to match struct.proto.

devinrsmith (changes requested):

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/config.proto
niloc132: Ill make the change for easy compat, but I wouldn't consider struct.proto to be a goal or even good code. It "works" as a "well I guess most things can interchange with this if they can at least interop with JSON cleanly", but you effectively have to convert to JSON to be able to query it cleanly, instead of knowing what types/keys to expect, at least not without some irritating boilerplate ("Key `xyz` is a list of strings, let's look for it. Is it present? Okay good, read the Value. Is it a null? No, okay good, we don't return a null value. Is it a ListValue? Okay, data isn't mangled. Read each Value..."). Additionally, despite JSON specifying no limits to math precision, but allowing that implementations might support some limits, Value allows `double`, so int32 values have to be narrowed when read from a strongly typed language, and int64 values may lose precision (or get encoded as strings...).

I'd nearly rather have CSV.

devinrsmith (approved):

mofojed (approved):

PR #2982, opened 2022-10-10
Remove logroot and logDir. Fixup HeapDump usage.

devinrsmith:


Util/src/main/java/io/deephaven/util/HeapDump.java
```diff
@@ -3,74 +3,21 @@
...
-            heapDump(heapDumpPath);
-        } catch (Exception e) {
-            log.info()
-                    .append("Exception while trying to dump heap on assertion failure: " + e.getMessage() + ":\n")
-                    .append(e)
-                    .endl();
-        }
-    }
-
-    public static void setupHeapDumpWithDefaults(final Configuration configuration,
```
rcaudy, line 61: Why would you remove this? We use it pretty heavily for error capture.

devinrsmith:
Superceded by #3035

PR #2998, opened 2022-10-13
Update Clock interface, remove TimeProvider

### LINKED ISSUE #2992: MicroTimer/DateTimeUtils.currentTime()/DateTime.now() can return wrong time
The `io.deephaven.util.clock.MicroTimer` computes a fixed offset from System.nanoTime() so that it can return unix time that ticks based on nanoTime(), but this causes it to return inaccurate results when the clock jumps(/slews, probably), because there's no way to 

```
println(currentTime())
println(millisToTime(System.currentTimeMillis()))
2022-10-12T14:50:15.692889000 NY
2022-10-12T14:50:15.652000000 NY
```

After manually rolling my clock back a few minutes:
```
println(currentTime())
println(millisToTime(System.currentTimeMillis()))
2022-10-12T14:50:43.948189000 NY
2022-10-12T14:47:40.934000000 NY
```

As @devinrsmith noted there appear to be three reasonable ways to obtain the time, depending somewhat on the environment/application:

1. Use `System.currentTimeMillis()` (and multiply by 1000 to fake a currentTimeMicros)
2. Use `Instant.now()` — be aware of potential for extra GC
3. Use `jdk.internal.misc.VM` — extra setup involved for external consumption

devinrsmith:
Updates the Clock interface providing a much richer selection of methods, allowing the caller to best specify how they actually want to use the results. Removes TimeProvider interface and replaces with Clock. Allows the removal of a lot of extraneous allocation. An allocation-free Clock implementation is provided, but gracefully falls back to the Java UTC clock / Instant implementation.

Fixes #2992


devinrsmith:
This is _somewhat_ orthogonal to https://github.com/deephaven/deephaven-core/pull/2999, but there will be some conflicts in either case.

This PR is more about the clock interface and plumbing the callers intentions as deep as possible; while the other PR is more about making an allocation free system clock.

devinrsmith:
This is _somewhat_ orthogonal to https://github.com/deephaven/deephaven-core/pull/2999, but there will be some conflicts in either case.

This PR is more about the clock interface and plumbing the callers intentions as deep as possible; while the other PR is more about making an allocation free system clock.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -2336,4 +2376,38 @@ public static String getPartitionFromTimestampSeconds(@NotNull final DateTimeFor
         }
         return dateTimeFormatter.format(Instant.ofEpochMilli(timestampSeconds * 1_000));
     }
+
+    private enum TimeProviderImpl implements TimeProvider {
```
rbasralian, line 2380: why enum?

I think it'd be more straightforward to just add this INSTANCE as a field in TimeProvider

FishUtil/src/main/java/io/deephaven/util/clock/DeltaLocalClockImpl.java
```diff
@@ -1,34 +0,0 @@
...
-package io.deephaven.util.clock;
-
-import io.deephaven.base.clock.Clock;
-
-import java.io.Serializable;
-
-/**
- * Real-time local clock impl with a delta adjustment for all time values.
- */
-public class DeltaLocalClockImpl implements Clock, Serializable {
```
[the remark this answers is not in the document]
devinrsmith: I'll bring it back.

Base/src/main/java/io/deephaven/base/clock/Clock.java
```diff
@@ -3,87 +3,58 @@
...
+     * 
+     * <pre>
+     * long startNanoTime = clock.nanoTime();
+     * // ... the code being measured ...
+     * long elapsedNanos = clock.nanoTime() - startNanoTime;
+     * </pre>
+     * 
+     * @return the nano time
+     */
+    long nanoTime();
```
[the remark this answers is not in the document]
devinrsmith: No - this is explicitly required IMO when you want to have a replayer, but want to be able to finely replay "timings". For example, to use AutoTuningIncrementalReleaseFilter with a replayer, we need to rely on `Clock.nanoTime()` instead of `System.nanoTime()`.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
devinrsmith: IMO, when singleton structures can be represented as enum, you gain niceties - equality, hashcode, etc for free.

engine/time/src/main/java/io/deephaven/time/TimeProviderNanoBase.java
```diff
@@ -0,0 +1,14 @@
...
+import io.deephaven.base.clock.ClockNanoBase;
+
+/**
+ * A base implementation of TimeProvider, with all methods being sourced from {@link #currentTimeNanos()}.
+ */
+public abstract class TimeProviderNanoBase extends ClockNanoBase implements TimeProvider {
+
+    @Override
+    public final DateTime currentTime() {
+        return new DateTime(currentTimeNanos());
```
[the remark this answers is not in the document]
devinrsmith: I'm strongly anti-default methods - I think it muddies the water, and potentially provides places where calls are virtualized where they otherwise wouldn't need to be.

For example, `io.deephaven.engine.table.impl.replay.DataDrivenReplayer` is explicitly based off of `DateTime`; if we accidentally forget to override `currentTime()`, then we'd still get correct, but inefficient, behavior.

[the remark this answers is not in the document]
devinrsmith: Also, it's completely acceptable to create a `TimeProvider` wholly from `System.currentTimeMillis()` if you wanted.

Base/src/main/java/io/deephaven/base/clock/DeltaClock.java
```diff
@@ -0,0 +1,78 @@
...
+     * @param deltaNanos the delta nanos
+     */
+    public DeltaClock(Clock delegate, long deltaNanos) {
+        this.delegate = Objects.requireNonNull(delegate);
+        this.deltaNanos = deltaNanos;
+    }
+
+    @Override
+    public long currentTimeMillis() {
+        return delegate.currentTimeMillis() + deltaNanos / 1_000_000;
```
rbasralian, line 50: think we should use `Math.addExact()` for this and `currentTimeMicros()`

Base/src/main/java/io/deephaven/base/clock/Clock.java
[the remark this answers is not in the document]
rbasralian: Ok, I'm still a bit skeptical of the legitimacy of case 1, but it does make the interface more complete

rbasralian (dismissed):

py/embedded-server/deephaven_server/start_jvm.py
```diff
@@ -81,8 +81,10 @@ def start_jvm(
     required_jvm_args = [
         # Allow netty to access java.nio.Buffer fields
         '--add-opens=java.base/java.nio=ALL-UNNAMED',
-        # Allow our hotspotImpl project to access internals
+        # Allow our hotspot-impl project to access internals
         '--add-opens=java.management/sun.management=ALL-UNNAMED',
+        # Allow our clock-impl project to access internals
+        '--add-opens=java.base/jdk.internal.misc=ALL-UNNAMED',
```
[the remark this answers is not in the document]
devinrsmith: `--add-opens` is necessary from the runtime perspective. `--add-exports` is a compile-time flag.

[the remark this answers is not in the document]
devinrsmith: Great find - `--add-opens` is more aggressive than `--add-exports` in allowing reflective access; something we  don't need for `hotspot-impl` nor `clock-impl`. Updated to reflect this.

rbasralian (dismissed):

clock/src/main/java/io/deephaven/base/clock/DeltaClock.java
```diff
@@ -0,0 +1,78 @@
...
+import java.time.Instant;
+import java.util.Objects;
+
+/**
+ * A clock impl with a delta adjustment.
+ */
+public final class DeltaClock implements Clock {
+
+    /**
+     * Constructs a clock delta with a {@link Clock#systemUTC()} clock.
```
rcaudy, line 16: "clock delta" -> "DeltaClock"?

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/DataDrivenReplayer.java
```diff
@@ -20,8 +22,33 @@ public DataDrivenReplayer(DateTime startTime, DateTime endTime) {
     }
 
     @Override
-    public DateTime currentTime() {
-        return currentTime;
+    public long currentTimeMillis() {
```
rcaudy, line 25: Maybe we can move this up a level to avoid repetition, delegating to an abstract accessor for `currentTime`.

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/Replayer.java
```diff
@@ -13,37 +14,31 @@
...
 
-import static io.deephaven.time.DateTimeUtils.millisToNanos;
-import static io.deephaven.time.DateTimeUtils.nanosToTime;
-
 /**
  * Replay historical data as simulated real-time data.
  */
 public class Replayer implements ReplayerInterface, Runnable {
     private static final Logger log = LoggerFactory.getLogger(ShiftObliviousInstrumentedListener.class);
+    private static final int MILLIS_TO_NANOS_FACTOR = 1_000_000;
```
rcaudy, line 33: Maybe we want to prefer using `TimeUnit`?

engine/table/src/test/java/io/deephaven/engine/table/impl/QueryTableAggregationTest.java
```diff
@@ -149,7 +149,7 @@ public void testStaticNoKeyByWithChunks() {
     @Test
     public void testStaticReinterpretableKeyByWithChunks() {
         final String nowName = "__now_" + Thread.currentThread().hashCode() + "__";
-        QueryScope.addParam(nowName, DateTime.now());
+        QueryScope.addParam(nowName, DateTime.nowMillis());
```
rcaudy, line 152: Let's change these two back. Other cases in other files, too.

engine/table/src/test/java/io/deephaven/engine/util/TimeTableTest.java
```diff
@@ -43,7 +44,12 @@ protected void setUp() throws Exception {
         super.setUp();
 
         now = new MutableLong(0);
-        timeProvider = () -> new DateTime(now.longValue());
+        clock = new ClockNanoBase() {
```
rcaudy, line 47: We do this a lot. Maybe we want one based on a `MutableLong`?

clock/src/main/java/io/deephaven/base/clock/DeltaClock.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    public long currentTimeNanos() {
+        return Math.addExact(delegate.currentTimeNanos(), deltaNanos);
+    }
+
+    @Override
+    public Instant instantNanos() {
+        return delegate.instantNanos().plusNanos(deltaNanos);
```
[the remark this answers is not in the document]
devinrsmith: So, I'm not as concerned w/ the efficiency of non-standard clock implementations - regardless, I think the current implementation is the "correct" implementation, given that we should always delegate the same method to `delegate` given it knows more about its implementation than DeltaClock does.

If we really needed to optimize for applying deltas to clocks, that would suggest we should instead have an interface:

```java
interface Clock {
...
public Instant instantNanos(long deltaNanos);
...
}
```

But I _think_ we can agree that would be overkill.

engine/table/src/test/java/io/deephaven/engine/util/TimeTableTest.java
devinrsmith: Elevated TestClock.

engine/table/src/test/java/io/deephaven/engine/table/impl/QueryTableAggregationTest.java
devinrsmith: Done

clock/src/main/java/io/deephaven/base/clock/SystemClockInstance.java
```diff
@@ -0,0 +1,52 @@
+/**
+ * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+ */
+package io.deephaven.base.clock;
+
+import java.lang.reflect.InvocationTargetException;
+import java.time.Instant;
+
+enum SystemClockInstance implements SystemClock {
```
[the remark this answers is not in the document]
devinrsmith: Done

clock/src/main/java/io/deephaven/base/clock/ClockNanoBase.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    public final Instant instantMillis() {
+        return Instant.ofEpochSecond(0, currentTimeNanos());
+    }
+
+    @Override
+    public final long nanoTime() {
+        return currentTimeNanos();
```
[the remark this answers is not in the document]
devinrsmith: This is a larger question of "Should we have unit-testability for nanoTime logic; and should that interface be on Clock?"

Confirmed w/ Charles, we might want fine replayability of nanoTime for unit tests, even though we don't have them right now.

The next part is "should this be part of clock?"

It _may_ be reasonable to have a separate interface since it is more bespoke than Clock; but what I *don't* want to encourage is users just deciding to use `Clock#currentTimeNanos` because it's more convenient than plumbing through an additional nanoTime interface. That said, there aren't many users of `Clock#nanoTime` (only AutoTuningIncrementalReleaseFilter and BaseIncrementalReleaseFilter), so it would be a small change against this PR.

Options:
1. Leave `Clock#nanoTime` as-is
2. Create new interface `MeasureNanos`.
3. Remove `Clock#nanoTime`, and explicitly use `System.nanoTime` in BaseIncrementalReleaseFilter/AutoTuningIncrementalReleaseFilter

I'm actually leaning 3 > 2 > 1; at least until we get to the point where we _need_ nanoTime replayability for unit tests.

clock/src/main/java/io/deephaven/base/clock/SystemClockInstance.java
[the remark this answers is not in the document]
devinrsmith: I think `SystemClockUtc` is better?

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/DataDrivenReplayer.java
devinrsmith: So, the hierarchy right now is a bit confusing, with DataDrivenReplayer and FixedStepReplayer (extends io.deephaven.engine.table.impl.replay.Replayer) having completely separate overrides than the parent class.

I _think_ it could be cleaned up, but it would be a pretty major gutting that would be better done in isolation IMO.

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/Replayer.java
devinrsmith: I'll delegate back to DateTimeUtils.millisToNanos etc. And remove ENABLE_MICROTIME_HACK in the process.

clock/src/main/java/io/deephaven/base/clock/ClockNanoBase.java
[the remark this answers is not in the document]
rcaudy: I kind of like #3 for now. It feels better not to conflate sources of wall-clock time with sources of monotonically-increasing measurement time. I'm not going to complain too much if we do #2 or #1, though, at this point.

clock/src/main/java/io/deephaven/base/clock/DeltaClock.java
```diff
@@ -13,7 +13,7 @@
 public final class DeltaClock implements Clock {
 
     /**
-     * Constructs a clock delta with a {@link Clock#systemUTC()} clock.
+     * Constructs a delta clock with a {@link Clock#systemUTC()} clock.
```
rcaudy, line 16: Maybe do the other places, too?

clock/src/main/java/io/deephaven/base/clock/ClockNanoBase.java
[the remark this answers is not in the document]
devinrsmith: Proceeded w/ 3

rcaudy (dismissed):

clock/src/main/java/io/deephaven/base/clock/Clock.java
```diff
@@ -0,0 +1,79 @@
...
+public interface Clock {
+
+    /**
+     * The {@link SystemClock}. Provides singleton semantics around {@link SystemClock#of()}.
+     *
+     * @return the system clock
+     */
+    static SystemClock systemUTC() {
+        return SystemClockInstance.INSTANCE;
+    }
```
chipkent, line 23: I don't think this method belongs.  It is specific to a single clock, and that clock can be obtained via other methods.

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/ReplayerInterface.java
```diff
@@ -3,16 +3,16 @@
...
-import io.deephaven.time.DateTime;
 
 import java.io.IOException;
 import java.util.TimerTask;
 
 /**
  * An interface for replaying historical data as simulated real-time data.
  */
-public interface ReplayerInterface {
+public interface ReplayerInterface extends Clock {
```
chipkent, line 15: Thoughts on making the replayer instance a `Clock` vs returning a `Clock`?  I might lean to the second, since it seems funny to say that the replayer is a clock.  Also, "is a `Clock`" is causing a bunch of methods to be duplicated in many files.

engine/time/src/main/java/io/deephaven/time/DateTime.java
```diff
@@ -154,9 +202,7 @@ public org.joda.time.DateTime getJodaDateTime(TimeZone timeZone) {
      * @return a Java Instant representing this DateTime
      */
     public Instant getInstant() {
-        long epochSecond = nanos / 1000000000;
-        long nanoAdjustment = nanos % 1000000000;
-        return Instant.ofEpochSecond(epochSecond, nanoAdjustment);
+        return Instant.ofEpochSecond(0, nanos);
```
chipkent, line 205: This looked correct before.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -171,10 +171,10 @@ private enum DateGroupId {
     public static String lastBusinessDayNyOverride;
 
     /**
-     * Allows setting a custom time provider instead of actual current time. This is mainly used when setting up for a
-     * replay simulation.
+     * Allows setting a custom clock instead of actual current time. This is mainly used when setting up for a replay
+     * simulation.
      */
-    public static TimeProvider timeProvider;
+    public static Clock clock;
```
chipkent, line 177: Having a static clock here is funky, since we can support multiple replayers, where we probably want the clock to be set based upon the context of the replayer.  We are certainly no worse than we were, but we should have a ticket to address this at some point.

chipkent (changes requested):

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/ReplayerInterface.java
devinrsmith: I'm happy to change this to "replayer returns a clock" but it's not going to solve any of the duplication - the replayer and classes that extend replayer seem to be built fundamentally different, so I don't think there's an easy path to de-duplication.

engine/time/src/main/java/io/deephaven/time/DateTime.java
devinrsmith: This is a slight improvement, we don't have to do the math. ourselves. The internals of ofEpochSecond can deal directly w/ `long nanos`.

clock/src/main/java/io/deephaven/base/clock/Clock.java
devinrsmith: I think the method may actually be better named `system()`...

That said, it does serve an important service not fulfilled by other methods - it's essentially providing a singleton cache around `SystemClock.of()`. (`SystemClockInstance` itself is not public.)

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
devinrsmith: I agree - it would be much better to inject and pass along a Clock instead of having a static. I'll create an issue and link it from DateTimeUtils.

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/ReplayerInterface.java
devinrsmith: I did find a slight way to simplify this by doing "returns a Clock" - and I agree it's a better interface.

chipkent (approved):

PR #2999, opened 2022-10-13
Allocation-free precise and accurate clocks

rbasralian:
Support for using `jdk.internal.misc.VM#getNanoTimeAdjustment`, `Instant.now()`, or `System.currentTimeMillis()` as the source of "now" in the engine


clock/src/main/java/io/deephaven/clock/RealTimeClockLoader.java
```diff
@@ -0,0 +1,53 @@
...
+    private static final String REAL_TIME_CLOCK_SOURCE =
+            System.getProperty(REAL_TIME_CLOCK_SOURCE_PROPERTY, "instant").toLowerCase();
+
+    private static final String JDK_INTERNALS_CLOCK_CLASS_NAME = "io.deephaven.clock.impl.JdkInternalsRealTimeClock";
+
+    static RealTimeClock loadImpl() {
+        final RealTimeClock impl;
+        switch (REAL_TIME_CLOCK_SOURCE) {
+            case "jdk-internals":
+                final Iterator<RealTimeClock> it = ServiceLoader.load(RealTimeClock.class).iterator();
```
cpwright, line 23: Can you safely try jdk-internals if you have something like "default"; then degrade to instant?

engine/time/build.gradle
```diff
@@ -27,3 +27,9 @@ dependencies {
             project(path: ':test-configs')
     Classpaths.inheritSlf4j(project, 'slf4j-simple', 'testRuntimeOnly')
 }
+
+test {
+    jvmArgs '--add-exports', 'java.base/jdk.internal.misc=ALL-UNNAMED'
+    systemProperty 'RealTimeClockLoader.timeSource', 'jdk-internals'
+//    systemProperty 'RealTimeClockLoader.timeSource', 'millis'
```
cpwright, line 34: Ideally one would test both ways, though the millis and instant implementations are trivial.  You might add a comment to that effect.

cpwright:
This change is worth a before and after JMH benchmark for getting millis/micros/nanos in a tight loop.

devinrsmith:
I've got a parallel draft PR that addresses a lot of deficiencies in our interfaces, but doesn't take any of the add-opens approach yet so I think both parts may be valuable. Will look soon. https://github.com/deephaven/deephaven-core/pull/2998

Base/src/main/java/io/deephaven/base/clock/Clock.java
```diff
@@ -3,17 +3,21 @@
  */
 package io.deephaven.base.clock;
 
+import io.deephaven.clock.RealTimeClock;
+import org.jetbrains.annotations.NotNull;
+
 public interface Clock {
```
devinrsmith, line 9: We should probably move this to the `clock` package.

Base/src/main/java/io/deephaven/base/clock/Clock.java
```diff
@@ -3,17 +3,21 @@
...
     long currentTimeMillis();
 
     long currentTimeMicros();
 
-    interface Factory {
-        Clock getClock();
-    }
+    long currentTimeNanos();
+
+    final class Null implements Clock {
```
devinrsmith, line 17: enum is a better pattern IMO, but I noticed there aren't any real users of Null.

Base/src/main/java/io/deephaven/base/clock/Clock.java
```diff
@@ -23,67 +27,38 @@ public long currentTimeMillis() {
...
-        public void set() {
-            cachedNowMicros = realClock.currentTimeMicros();
-        }
+        private SystemClock() {}
 
-        public void reset() {
-            cachedNowMicros = 0;
-        }
+        @NotNull
+        public final RealTimeClock REAL_TIME_CLOCK = RealTimeClock.loadImpl();
```
devinrsmith, line 44: public?

clock-impl/src/main/java/io/deephaven/clock/impl/JdkInternalsRealTimeClock.java
```diff
@@ -0,0 +1,79 @@
+/**
+ * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+ */
+package io.deephaven.clock.impl;
```
devinrsmith, line 4: We'll want to use a different license for clock-impl if we use the code as-is since we are using systemUTC rather directly here. That may not be true if we change our strategy for calling VM

clock-impl/src/main/java/io/deephaven/clock/impl/JdkInternalsRealTimeClock.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Returns the current time in nanoseconds. This is the same moment in time {@link Instant#now()} would return, but
+     * represented (without allocation) as a single {@code long} instead of as an {@link Instant} object.
+     * 
+     * @return the difference, measured in nanoseconds (with precision determined by the result of
+     *         {@link VM#getNanoTimeAdjustment}), between the current time and midnight, January 1, 1970 UTC.
+     */
+    @Override
+    public long currentTimeNanos() {
```
devinrsmith, line 43: We should test out some different strategies. We _may_ be able to get away w/ simply:
```java
return VM.getNanoTimeAdjustment(0);
```

clock/src/main/java/io/deephaven/clock/SystemMillisRealTimeClock.java
```diff
@@ -0,0 +1,21 @@
+package io.deephaven.clock;
+
+/**
+ * Created by rbasralian on 10/13/22
+ */
+public final class SystemMillisRealTimeClock implements RealTimeClock {
```
devinrsmith, line 6: Can be enum w/ single INSTANCE.

clock/src/main/java/io/deephaven/clock/JavaInstantRealTimeClock.java
```diff
@@ -0,0 +1,15 @@
+package io.deephaven.clock;
+
+import java.time.Instant;
+
+/**
+ * Created by rbasralian on 10/13/22
+ */
+public final class JavaInstantRealTimeClock implements RealTimeClock {
```
devinrsmith, line 8: can be enum w/ single INSTANCE.

clock/src/main/java/io/deephaven/clock/RealTimeClock.java
```diff
@@ -0,0 +1,24 @@
+/**
+ * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+ */
+package io.deephaven.clock;
+
+import java.util.Iterator;
+import java.util.Optional;
+import java.util.ServiceLoader;
+
+public interface RealTimeClock {
```
devinrsmith, line 10: We should probably have `RealTimeClock extends Clock`

engine/time/build.gradle
```diff
@@ -27,3 +27,9 @@ dependencies {
             project(path: ':test-configs')
     Classpaths.inheritSlf4j(project, 'slf4j-simple', 'testRuntimeOnly')
 }
+
+test {
+    jvmArgs '--add-exports', 'java.base/jdk.internal.misc=ALL-UNNAMED'
+    systemProperty 'RealTimeClockLoader.timeSource', 'jdk-internals'
```
devinrsmith, line 33: IMO, we should have an "escape" hatch, but the majority of cases shouldn't need to specify the system property IMO.

server/jetty-app/build.gradle
```diff
@@ -41,6 +41,14 @@ if (!hasProperty('excludeHotspotImpl')) {
     extraJvmArgs += ['--add-opens', 'java.management/sun.management=ALL-UNNAMED']
 }
 
+if (!hasProperty('excludeClockImpl')) {
+    dependencies {
+        runtimeOnly project(':clock-impl')
+    }
+    extraJvmArgs += ['--add-exports', 'java.base/jdk.internal.misc=ALL-UNNAMED']
+    extraJvmArgs += ['-DRealTimeClockLoader.timeSource=jdk-internals']
```
devinrsmith, line 49: again, this should be implicit based on the classpath for the majority of use-cases.

server/netty-app/README.md
```diff
@@ -43,6 +43,8 @@ Finally, Gradle can be used to update the build and run the application in a sin
 `--add-opens java.management/sun.management=ALL-UNNAMED`. To disable this, set the gradle property `includeHotspotImpl`
 to `false`.
 
+... also clockImpl and java.base/jdk.internal.misc=ALL-UNNAMED
```
devinrsmith, line 46: better doc.

devinrsmith (changes requested):
I believe there we will want to have some python side changes too for the embedded server case.

devinrsmith:
For reference, here's the underlying JDK code path for VM.getNanoTimeAdjustment:

https://github.com/openjdk/jdk/blob/jdk-17+35/src/hotspot/share/prims/jvm.cpp#L234-L282

https://github.com/openjdk/jdk/blob/jdk-17+35/src/hotspot/os/posix/os_posix.cpp#L1385-L1428

https://github.com/openjdk/jdk/blob/jdk-17+35/src/hotspot/os/windows/os_windows.cpp#L1013-L1059

devinrsmith:
Benchmarking nanoTime() is hard. It may be a little better w/ REALTIME clocks b/c they don't have to be monotonic.

See https://shipilev.net/blog/2014/nanotrusting-nanotime/

devinrsmith:
https://github.com/openjdk/jdk/blob/jdk-17+35/src/hotspot/os/bsd/os_bsd.cpp#L788-L821

rbasralian:
See #2998

PR #3006, opened 2022-10-20
Add code to cpp-client/build.gradle to run tests in docker.

jcferretti:
This does not include yet code to automatically check consistency of generated protos, that should come in a later PR.

Summary of changes:
* Removed `build-dependencies.sh`, now lives in the deephaven-base-images repository (https://github.com/deephaven/deephaven-base-images/tree/main/cpp-client).  The script there is refactored some, to allow for building individual parts.  Also made the script write a new `env.sh` file at the end with environment variables to source for, eg, `CMAKE_PREFIX_PATH`.  Updated `README.md` to reflect the change.
* In a separate PR for the deephaven-base-images repository, and with a lot of help from Devin (thanks!)
  we have added code to create a "cpp-client-base" image (code [here](https://github.com/deephaven/deephaven-base-images/tree/d05e316215e44889d0d0140210ab42882a238f71/cpp-client)) that contains all the dependent libraries already built
  in ubuntu-22.04.  The code added to `cpp-client/build.gradle` to do checks via docker leans on that to avoid
  having to wait too long and create too many layer of the new image.
* Small change to `tests/test_util.cc` to support changing the host:port for the server via setting environment variables `DH_HOST` and `DH_PORT`, needed for automating the tests.


cpp-client/README.md
```diff
@@ -27,7 +27,7 @@ C++ compiler and tool suite (cmake etc).
 6. Build and install dependencies for Deephaven C++ client.
 
    Decide on a directory for the dependencies to live (eg, "$HOME/dhcpp").
-   Create that directory and copy the `$DHSRC/deephaven-core/cpp-client/deephaven/client/build-dependencies.sh`
+   Create that directory and copy the `$DHSRC/deephaven-core/cpp-client/build-dependencies.sh`
```
kosak, line 30: By the way, my own workflow for doing this is

* create that directory
* cd to that directory
* run $DHSRC/deephaven-core/cpp-client/build-dependencies.sh

Put another way, I never copy build-dependencies.sh over to my target directory. I simply set my working directory to the target and run build-dependencies.sh using its full path.

(I realize that in some cases people modify build-dependencies.sh, but even in that case, they can make their own decisions about modifying it and where the modified file should live)

Up to you, but I mention this because (depending on your opinion) it might simplify these instructions slightly, and it might be a *tad* more in line with other projects which don't leave a build script behind in the output directory.

cpp-examples/demos/crypto.cc
```diff
@@ -0,0 +1,184 @@
...
+// demo options
+void printFull(const TableHandleManager &manager);
+void printDiffs(const TableHandleManager &manager);
+
+// utilities
+int64_t readNumber(std::string_view prompt);
+std::string readString(std::string_view prompt);
+
+// a bunch of work to create an ostream adaptor
+template<typename ARROW_OPTIONAL>
```
kosak, line 34: I think you can delete this class `OptionalAdaptor` and the method below `adaptOptional`

cpp-client/tests/test_util.cc
```diff
@@ -114,7 +115,11 @@ TableMakerForTests TableMakerForTests::create() {
 }
 
 Client TableMakerForTests::createClient() {
-  std::string connectionString("localhost:10000");
+  const char* hostptr = std::getenv("DH_HOST");
+  const char* portptr = std::getenv("DH_PORT");
```
kosak, line 119: super-mega-nit: because of how C++ binding works I'd prefer

```
const char *hostptr
const char *portptr
```

i.e. the * being closer to the variable name

cpp-client/tests/test_util.cc
```diff
@@ -114,7 +115,11 @@ TableMakerForTests TableMakerForTests::create() {
 }
 
 Client TableMakerForTests::createClient() {
-  std::string connectionString("localhost:10000");
+  const char* hostptr = std::getenv("DH_HOST");
```
kosak, line 118: Are we OK with "licking the cookie" on a short name like `DH_HOST` rather than say `DEEPHAVEN_HOST`?
I'm not particularly opinionated, just wondering if this was considered. I suppose if it's only for our own tests it doesn't matter much.

cpp-examples/demos/crypto.cc
```diff
@@ -0,0 +1,184 @@
+/*
```
kosak, line 1: Is there anything different about crypto.cc (vs demo3.cc or whatever) aside from the connection string `localhost:10042` ?

kosak: I'm not asking you to do anything, but if you want me to take a pass over this and consolidate these demos a little more I can

kosak (approved):
Some micro-comments.

cpp-client/build-dependencies.sh
```diff
@@ -11,6 +11,21 @@
 # Fail on first error; echo each command before executing.
 set -eux
 
+if [ "$#" -ge 1 -a "$1" == "--clean" ]; then
```
devinrsmith, line 14: We need to figure out how to manage this file, since it's now part of deephaven-base-images.

cpp-client/cpp-client-Dockerfile
```diff
@@ -0,0 +1,66 @@
+#
```
devinrsmith, line 1: We shouldn't need this file anymore, right?

docker/registry/cpp-client-base/gradle.properties
```diff
@@ -0,0 +1,3 @@
+io.deephaven.project.ProjectType=DOCKER_REGISTRY
+deephaven.registry.imageName=cpp-client-base:1
+deephaven.registry.imageId=cpp-client-base@sha256:7d99ae35e5f732ee1e6df65fe82b3dc6bcb2f38529f67081bd7328caadd203ba
```
devinrsmith, line 3: These are not properly reference imageName / imageId. Should start w/ `ghcr.io/deephaven/...`

jcferretti: That explains why they don't work... but it doesn't explain why the directory I used as a template has the same problem and that does work:

https://github.com/deephaven/deephaven-core/blob/4d721823f7d8841c1ff86a7c0ce7d0a220155559/docker/registry/go/gradle.properties

cpp-client/cpp-client-Dockerfile
jcferretti: See the last point in the description of the PR; I decided to leave it as a convenience for users.
I can make that more explicit in the README or in comments if you'd lile.

cpp-client/README.md
jcferretti: Devin wants me to remove build-dependencies.sh completely from this repo since it has to also live in https://github.com/deephaven/deephaven-base-images/blob/main/cpp-client/build-dependencies.sh.
So I've re-done the README and done some of what you asked (but the example still copies locally, since the script doesn't exist anymore in the clone of deephaven-core).  PTAL and see what you think.

cpp-client/tests/test_util.cc
jcferretti: The pre-existing via-docker test code (eg gradle files and supporting code for python, go, etc) use these variables already.
I think it makes sense to be consistent, and I don't think this PR would be the place to rename everything, so my humble take here is that we should take this conversation to a different PR if we think that value for effort that change would make sense.  For reference in making that decision:
```
cfs@guanaco 23:18:28 ~/dh/oss1/deephaven-core
$ find . -type f | xargs grep DH_HOST
./py/client/pydeephaven/session.py:            self.host = os.environ.get("DH_HOST", "localhost")
./py/client/build.gradle:        environmentVariable 'DH_HOST', deephavenDocker.containerName.get()
./py/client/build/buildWheel-docker/src/build.gradle:        environmentVariable 'DH_HOST', deephavenDocker.containerName.get()
./py/client/build/buildWheel-docker/src/pydeephaven/session.py:            self.host = os.environ.get("DH_HOST", "localhost")
./sphinx/build/cppClientDoxygen-docker/build.gradle:                  'DH_HOST=' + deephavenDocker.containerName.get() + ' ' +
./sphinx/build/cppClientDoxygen-docker/tests/test_util.cc:  const char* hostptr = std::getenv("DH_HOST");
grep: ./.gradle/7.5/executionHistory/executionHistory.bin: binary file matches
./cpp-client/build/testCppClient-docker/Dockerfile:ENV DH_HOST=dh-server-for-cpp-933abaa3-6584-4bfa-86f0-993add1082be
./cpp-client/build/testCppClient-docker/tests/test_util.cc:  const char* hostptr = std::getenv("DH_HOST");
./cpp-client/build/testCppClient-dockerfile/Dockerfile:ENV DH_HOST=dh-server-for-cpp-933abaa3-6584-4bfa-86f0-993add1082be
./cpp-client/tests/#test_util.cc#:  const char *hostptr = std::getenv("DH_HOST");
./cpp-client/tests/test_util.cc:  const char* hostptr = std::getenv("DH_HOST");
./cpp-client/build.gradle:        environmentVariable 'DH_HOST', deephavenDocker.containerName.get()
./go/build.gradle:        environmentVariable 'DH_HOST', deephavenDocker.containerName.get()
./go/README.md:Tests assume the server is on `localhost:10000` by default, but can be configured by setting the `DH_HOST` and `DH_PORT` environment variables.
./go/README.md:but can be configured by setting the `DH_HOST` and `DH_PORT` environment variables.
./go/internal/test_tools/test_tools.go:// By default it is localhost, but can be overriden by setting the DH_HOST environment variable.
./go/internal/test_tools/test_tools.go: host := os.Getenv("DH_HOST")
```

cpp-examples/demos/crypto.cc
jcferretti: I'm moving that file out of this PR and onto a different one, we can discuss separately when we (just you and me and not the rest of the ppl in this review) discuss the crypto demo part for tuco y tico etc.  Thanks for the comments btw, I made the changes you suggested earlier.

cpp-client/build-dependencies.sh
jcferretti: You win, I am removing it.

cpp-client/cpp-client-Dockerfile
jcferretti: Nevermind, I am removing it too.

cpp-client/tests/test_util.cc
kosak: It's fine, I was just bringing it up. You folks are the the deciders.

cpp-client/README.md
kosak: Ah, looks fine. Thank you. You can resolve this :-)

cpp-examples/demos/crypto.cc
kosak: SGTM

docker/registry/cpp-client-base/gradle.properties
```diff
@@ -0,0 +1,3 @@
+io.deephaven.project.ProjectType=DOCKER_REGISTRY
+deephaven.registry.imageName=ghcr.io/deephaven/cpp-client-base:1
```
devinrsmith, line 2: this is still wrong, should be `:latest`

cpp-client/build.gradle
```diff
@@ -19,3 +23,75 @@ tasks.getByName('licenseFormat').dependsOn(tasks.create("licenseFormatPy", com.h
...
+        // Setup for test run.
+        //
+        environmentVariable 'DH_HOST', deephavenDocker.containerName.get()
+        environmentVariable 'DH_PORT', '8080'
+    }
+    containerDependencies.dependsOn = [deephavenDocker.healthyTask]
+    containerDependencies.finalizedBy = deephavenDocker.endTask
+    network = deephavenDocker.networkName.get()
+    parentContainers = [ Docker.registryTask(project, 'cpp-client-base') ]
+    entrypoint = ['bash', '-c', 'set -eux; mkdir -p /out; cd /cpp-client/install/bin; ' +
```
devinrsmith, line 94: The mkdir should be setup as part of a previous run command. The entrypoint should be script and params as an array (as such, we shouldn't be doing redirection in the entrypoint command either - that can be incorporated into the script as `${2}` or something).

devinrsmith (changes requested):

cpp-client/build.gradle
```diff
@@ -19,3 +23,75 @@ tasks.getByName('licenseFormat').dependsOn(tasks.create("licenseFormatPy", com.h
...
+        // Build client library.
+        //
+        runCommand('mkdir -p /cpp-client/deephaven/client /cpp-client/deephaven/build')
+        copyFile('deephaven/CMakeLists.txt', '/cpp-client/deephaven')
+        copyFile('deephaven/client', '/cpp-client/deephaven/client/')
+        runCommand('''set -eux; \\
+                      cd /cpp-client/deephaven/build; \\
+                      . /cpp-client/deps/env.sh; \\
+                      cmake -DCMAKE_INSTALL_PREFIX=/cpp-client/install ..; \\
+                      make -j\$NCPUS install; \\
```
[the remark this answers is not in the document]
jcferretti: We discussed over slack; making a change in the code in base-images for build-dependencies.sh, as to ensure that NCPUS gets its value from an evaluation at the time the script is sourced (eg, env.sh will contain `export NCPUS=$(getconf _NPROCESSORS_ONLN)`

jcferretti:
> Just for reference, the base image will be our largest yet, topping in at over 3GB.
> 
> ```
> deephaven/cpp-client-base                                 local-build       2893a2a398e3   16 hours ago     3.11GB
> ```

Devin and myself discussed making a default BUILD_TYPE=Release for image generation (the build-dependencies.sh script itself is not changing, but the way it is invoked from the dockerfile will change).  This results in a more economical image:
```
cfs@guanaco 15:46:55 ~/dh/oss1/deephaven-base-images
$ docker images | grep cpp-client-base
ghcr.io/deephaven/cpp-client-base                         latest        071706e8d34c   26 seconds ago   689MB
```

I tried that locally, I am updating the PR for deephaven-base-images right after writing this.

devinrsmith (approved):

PR #3009, opened 2022-10-20
Remove explicit JVM flags around GC

devinrsmith:
We currently bake-in some GC options that would be better left to default JDK ergonomics or the user.

For example, we are currently unable to (easily) change the garbage collector:

```
$ docker run --env "JAVA_OPTS=-XX:+UseZGC" --rm ghcr.io/deephaven/server:0.17.0             
Error occurred during initialization of VM
Multiple garbage collectors selected
```

### DIFF
--- a/containers/groovy-examples/docker-compose.yml
+++ b/containers/groovy-examples/docker-compose.yml
@@ -8,7 +8,7 @@ services:
     volumes:
       - ./data:/data
     environment:
-      - JAVA_OPTS=-Xmx4g
+      - START_OPTS=-Xmx4g
 
   examples:
     image: ghcr.io/deephaven/examples:latest
--- a/containers/groovy/docker-compose.yml
+++ b/containers/groovy/docker-compose.yml
@@ -8,4 +8,4 @@ services:
     volumes:
       - ./data:/data
     environment:
-      - JAVA_OPTS=-Xmx4g
+      - START_OPTS=-Xmx4g
--- a/containers/python-examples-redpanda/docker-compose.yml
+++ b/containers/python-examples-redpanda/docker-compose.yml
@@ -8,7 +8,7 @@ services:
     volumes:
       - ./data:/data
     environment:
-      - JAVA_OPTS=-Xmx4g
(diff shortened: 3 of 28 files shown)


rcaudy:
I don't want to change this. I'm willing to introduce a way to disable the settings, but we should have reasonable defaults tuned for a low-latency application.

devinrsmith:
The good news is that G1GC is the default for 11/17. The default MaxGCPauseMillis for G1GC is 200. @rcaudy, are we worried about bumping MaxGCPauseMillis from 100 to 200?

If we think explicitly setting G1GC/100 is really best for the majority of cases, instead of baking these values into the lower layer start script, we could set them at the docker image layer:

```Dockerfile
ENV JAVA_OPTS="-XX:+UseG1GC -XX:MaxGCPauseMillis=100"
```

This would allow them be wholly overridden when necessary.

devinrsmith:
https://docs.oracle.com/en/java/javase/17/gctuning/garbage-first-garbage-collector-tuning.html#GUID-0BB3B742-A985-4D5E-A9C5-433A127FE0F6

https://docs.oracle.com/en/java/javase/17/gctuning/garbage-first-garbage-collector-tuning.html#GUID-4914A8D4-DE41-4250-B68E-816B58D4E278

is relevant material.

devinrsmith:
It's reasonable for us to consider removing the explicit `-Xmx4g` from example compose files; and potentially add `-XX:MaxRAMPercentage=70.0` (instead of the default 25.0) as part of `JAVA_OPTS`.

rcaudy:
You need to have local runs pick up the options as well (START_OPTS and JAVA_OPTS).

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/1839
Conceptual: https://github.com/deephaven/deephaven.io/issues/1837
Reference: https://github.com/deephaven/deephaven.io/issues/1840
Blog: https://github.com/deephaven/deephaven.io/issues/1838

Issue #3011, opened 2022-10-21
Caching issue when updating Web UI


mofojed:
**Description**

There seems to still be an issue with caching the Web UI index.html. Even though I've updated the web version, it is returning the same ETag, loading the old version from memory. The Web UI then fails to load because the JS files no longer exist.

**Steps to reproduce**
1. Run `./gradlew server-jetty-app:run`
2. Open http://localhost:10000/ide/ and let the UI load
3. Stop the server
4. Update `WEB_VERSION` in `web/client-ui/Dockerfile` to a new version
5. Run `./gradlew server-jetty-app:run`
6. Open http://localhost:10000/ide/ and let the UI load

**Expected results**
6. New UI should load

**Actual results**
6. UI fails to load because of a caching issue. See screenshots.

**Additional details and attachments**

Before updating Web packages. ETag for `/ide/` is `W/"dw/nfoEwv/cdw/nNJbAiXY"`:
First load:
![image](https://user-images.githubusercontent.com/4505624/197555226-b8a9668d-940d-4028-9e53-2aa46a40d696.png)
After refresh:
![image](https://user-images.githubusercontent.com/4505624/197555671-77098a6a-9d37-4ee8-8821-5c2075e60daa.png)

After updating Web Packages (on my [branch web-v0.20.0](https://github.com/mofojed/deephaven-core/tree/web-v0.20.0)):
First load, ETag still has the same value:
![image](https://user-images.githubusercontent.com/4505624/197555970-2771b6b9-3b7f-4f86-a015-23188dbaf938.png)
This causes the js pages not to load, as the hash for the js files is incorrect:
![image](https://user-images.githubusercontent.com/4505624/197556177-8f3b3ced-abcf-4678-b009-70f83ebdc7f2.png)

After a hard refresh, it loads properly, but still the same ETag `W/"dw/nfoEwv/cdw/nNJbAiXY"`:
![image](https://user-images.githubusercontent.com/4505624/197556311-a0d0f338-fc24-4295-885d-f46b75b10c1a.png)

That ETag is different than the ETags for other files.
 

**Versions**
 - Deephaven: SHA `41e4b2f5790dabd0dceaa1663a043013ab4fac48`
 - OS: Ubuntu
 - Browser: Chrome

mofojed:
May be related to https://github.com/deephaven/deephaven-core/pull/2937/files
[...]

PR #3012, opened 2022-10-21
Update web UI to v0.20.0

mofojed:
Release Notes: https://github.com/deephaven/web-client-ui/releases/tag/v0.20.0

NOTE: Before merging this, I want to investigate #3011 and merge a fix for that, as this will break everybody because there's (still) a caching issue lurking.


mofojed:
Just going to update straight to v0.21.0

PR #3017, opened 2022-10-24
Update check-ci-results path to include all build/test-results

devinrsmith:
go and cpp-client are missing test and log data

```
$ find cpp-client/build/test-results -type f    
cpp-client/build/test-results/cpp-test.log
cpp-client/build/test-results/cpp-test.xml

$ find go/build/test-results -type f
go/build/test-results/go-test.xml
```

The check-ci-results path is too specific, and was built before these client tests existed.

### DIFF
--- a/.github/workflows/check-ci.yml
+++ b/.github/workflows/check-ci.yml
@@ -49,7 +49,7 @@ jobs:
         if: always()
         with:
           name: check-ci-results
-          path: '**/build/test-results/*/TEST-*.xml'
+          path: '**/build/test-results/**'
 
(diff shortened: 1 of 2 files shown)


jcferretti (dismissed):

mofojed (commented):
I would appreciate more details in the PR description for future context:
- Why the change?
- Any tests performed? E.g. are there only xml files in the folder, are we sure there aren't other things in that folder that would cause issues, etc.

mofojed (dismissed):

jcferretti (approved):

PR #3032, opened 2022-10-25
Disable autocomplete for python sessions

### LINKED ISSUE #3024: Disable python autocomplete
Temporary workaround for #3008, disabling all python autocomplete on the server, rather than give bad results.

niloc132:
Fixes #3024

### DIFF
--- a/server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
+++ b/server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
@@ -13,6 +13,7 @@ import io.deephaven.engine.util.DelegatingScriptSession;
 import io.deephaven.engine.util.ScriptSession;
 import io.deephaven.engine.util.VariableProvider;
 import io.deephaven.extensions.barrage.util.GrpcUtil;
+import io.deephaven.integrations.python.PythonDeephavenSession;
 import io.deephaven.internal.log.LoggerFactory;
 import io.deephaven.io.logger.LogBuffer;
 import io.deephaven.io.logger.LogBufferRecord;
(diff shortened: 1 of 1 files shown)


mofojed (approved):

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -259,145 +261,203 @@ public void bindTableToVariable(BindTableToVariableRequest request,
...
+
+        @Override
+        public void onNext(AutoCompleteRequest value) {
+            // This implementation only responds to autocomplete requests with "success, nothing found"
+            if (value.getRequestCase() == AutoCompleteRequest.RequestCase.GET_COMPLETION_ITEMS) {
+                safelyExecuteLocked(responseObserver, () -> {
+                    responseObserver.onNext(AutoCompleteResponse.newBuilder()
+                            .setCompletionItems(
+                                    GetCompletionItemsResponse.newBuilder().setSuccess(true))
+                            .build());
```
mofojed, line 441: Need to set the `requestId` on the response.
```suggestion
                    responseObserver.onNext(AutoCompleteResponse.newBuilder()
                            .setCompletionItems(
                                    GetCompletionItemsResponse.newBuilder().setSuccess(true)
                                            .setRequestId(value.getGetCompletionItems().getRequestId()))
                            .build());
```

mofojed (changes requested):
The oddities you were observing is because the request ID wasn't being set, so then in [IdeSession](https://github.com/deephaven/deephaven-core/blob/main/web/client-api/src/main/java/io/deephaven/ide/shared/IdeSession.java#L196) it couldn't map them back to the original pending autocomplete request.

mofojed (approved):

Issue #3021, opened 2022-10-24
Standard non-compliant code in the query library


chipkent:
Commit feb6c8edab8459c393217dc11e7b14c6897d7a3a introduced a number of changes that do not comply with the standards of the query library.  This code should have been reviewed by @kosak and @chipkent before merging.  This is a user library, so the standard is higher than other parts of the code.

1. All methods should be fully documented and consistent with the other docs in the file.
2. All methods should be essentially 100% unit tested.
3. I think `public static <T> T[] forwardFill(final T[] values)` may cause us problems.  The generic stuff should go at the top of the file.  Most of the generic methods have Obj in the name to avoid nasty collisions.  This needs a very careful look to assess the likelihood of causing problems.  Full unit tests will help identify such collisions.
4. `cumMin` and `cumMax` don't match the naming conventions of similar funcs, e.g. `cumsum` and `cumprod`.

lbooker42:
Closed by #3039

PR #3039, opened 2022-10-27
Clean up non-compliant code in the query library

lbooker42:
Renamed `cumMin`/`cumMax` to match existing functions. Added documentation to `forwardFill` and disambiguated templated function name. 

Addresses #3021 

Full unit tests for `cummin()`, `cummax()`, `forwardfill()` functions added with 100% new code coverage

### DIFF
--- a/engine/function/src/templates/Basic.ftl
+++ b/engine/function/src/templates/Basic.ftl
@@ -870,6 +870,62 @@ public class Basic {
         return result;
     }
 
+    /**
+     * Copies the specified array, replacing elements that represent null in the Deephaven convention by the most
+     * recently encountered non-null value if one exists. Otherwise (if no such value exists), replaces those elements
+     * with null.
+     *
+     * @param values values.
+     * @return A copy of the specified array, with Deephaven null elements replaced as described above. If the specified
+     *         array is null, returns null.
+     */
+    public static <T> T[] forwardFillObj(ObjectVector<T> values) {
+        if (values == null) {
+            return null;
+        }
+
+        final int n = values.intSize("forwardFill");
+        final T[] result = values.toArray();
+
+        T lastGood = null;
+        for (int ii = 0; ii < n; ii++) {
+            if (!isNull(values.get(ii))) {
+                lastGood = values.get(ii);
+            }
+
+            result[ii] = lastGood;
+        }
+        return result;
+    }
+
+    /**
+     * Copies the specified array, replacing elements that represent null in the Deephaven convention by the most
+     * recently encountered non-null value if one exists. Otherwise (if no such value exists), replaces those elements
+     * with null.
+     *
+     * @param values values.
+     * @return A copy of the specified array, with Deephaven null elements replaced as described above. If the specified
+     *         array is null, returns null.
+     */
+    public static Boolean[] forwardFillObj(BooleanVector values) {
+        if (values == null) {
+            return null;
+        }
+
+        final int n = values.intSize("forwardFill");
+        final Boolean[] result = values.toArray();
+
+        Boolean lastGood = null;
+        for (int ii = 0; ii < n; ii++) {
+            if (!isNull(values.get(ii))) {
+                lastGood = values.get(ii);
+            }
+
+            result[ii] = lastGood;
+        }
+        return result;
+    }
+
     <#list primitiveTypes as pt>
     <#if !pt.valueType.isBoolean >
 
--- a/engine/function/src/templates/Basic.ftl
+++ b/engine/function/src/templates/Basic.ftl
@@ -660,7 +660,7 @@ public class Basic {
             return null;
         }
 
-        final int n = values.intSize("reverse");
+        final int n = values.intSize("reverseObj");
         @SuppressWarnings("unchecked") final T[] result = (T[])Array.newInstance(values.getComponentType(), n);
 
         for (int i=0; i<n; i++) {
@@ -1746,7 +1746,7 @@ public class Basic {
             return null;
         }
 
-        final int n = values.intSize("reverse");
+        final int n = values.intSize("forwardFill");
         final ${pt.primitive}[] result = new ${pt.primitive}[n];
 
         ${pt.primitive} lastGood = QueryConstants.${pt.null};
--- a/engine/function/src/templates/Basic.ftl
+++ b/engine/function/src/templates/Basic.ftl
@@ -183,7 +183,7 @@ public class Basic {
      * @return number of non-null values.
      */
     static public <T> long countObj(ObjectVector<T> values) {
-        if (values == null){
+        if (values == null) {
             return NULL_LONG;
         }
 
@@ -311,7 +311,7 @@ public class Basic {
      * @return Deephaven vector.
      */
     @SafeVarargs
-    public static <T> ObjectVector<T> vecObj(T... values){
+    public static <T> ObjectVector<T> vecObj(T... values) {
         if (values == null) {
             return null;
         }
@@ -429,7 +429,7 @@ public class Basic {
 
         final long n = values.size();
 
-        if (n == 0){
+        if (n == 0) {
             return 0;
         }
 
@@ -554,7 +554,7 @@ public class Basic {
 
         final T[] array = (T[])Array.newInstance(value.getClass(), size);
 
-        for (int i = 0; i < size; i++){
+        for (int i = 0; i < size; i++) {
             array[i] = value;
         }
 
@@ -569,7 +569,7 @@ public class Basic {
      */
     @SafeVarargs
     @SuppressWarnings({"unchecked"})
-    public static <T> T[] concat(T[]... values){
+    public static <T> T[] concat(T[]... values) {
 
         int n = 0;
 
@@ -602,9 +602,9 @@ public class Basic {
      */
     @SafeVarargs
     @SuppressWarnings({"unchecked"})
-    public static <T> T[] concat(ObjectVector<T>... values){
+    public static <T> T[] concat(ObjectVector<T>... values) {
 
-        if (values.length == 0){
+        if (values.length == 0) {
             return (T[])Array.newInstance(Object.class, 0);
         }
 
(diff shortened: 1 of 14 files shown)


engine/function/src/templates/Basic.ftl
```diff
@@ -842,6 +842,32 @@ public class Basic {
         return ifelseObj(new BooleanVectorDirect(condition), trueCase, falseCase);
     }
 
+    /**
+     * Returns an array with null values replaced with preceding non-null value
```
kosak: Suggested wording, cribbed a little bit from Arrays.copyOf (any thoughts, Chip?)

Copies the specified array, replacing elements that represent null in the Deephaven convention by the most recently encountered non-null value if one exists. Otherwise (if no such value exists), replaces those elements with null.

[note also that there's always this issue where the description and the @return clause are somewhat duplicative. I chose to put more detail in the description, whereas you chose to put more detail in the @return clause. I think I like it "my way" better? Maybe Chip has an opinion]

engine/function/src/templates/Basic.ftl
```diff
@@ -842,6 +842,32 @@ public class Basic {
...
+     *
+     * @param values values.
+     * @return An array of T whose values are determined by the preceding elements.
```
kosak: @return A copy of the specified array, with Deephaven null elements replaced as described above. If the specified array is null, returns null.

engine/function/src/templates/Basic.ftl
```diff
@@ -842,6 +842,32 @@ public class Basic {
...
+     *         most recently encountered non-null value where one exists
+     *         null if all preceding values were null
+     */
+    public static <T> T[] forwardFillObj(final T[] values) {
+        if(values == null) {
```
kosak, line 854: spacing: `if (`

engine/function/src/templates/Basic.ftl
```diff
@@ -842,6 +842,32 @@ public class Basic {
...
+            return null;
+        }
+
+        final T[] result = Arrays.copyOf(values, values.length);
+
+        T lastGood = null;
+        for(int ii = 0; ii < values.length; ii++) {
```
kosak, line 861: spacing: `for (`

engine/function/src/templates/Basic.ftl
```diff
@@ -842,6 +842,32 @@ public class Basic {
...
+            if(!isNull(values[ii])) {
```
kosak, line 862: spacing: `if (`

engine/function/src/templates/Basic.ftl
```diff
@@ -842,6 +842,32 @@ public class Basic {
...
     }
 
+    /**
+     * Returns an array with null values replaced with preceding non-null value
+     *
+     * @param values values.
+     * @return An array of T whose values are determined by the preceding elements.
+     *         most recently encountered non-null value where one exists
+     *         null if all preceding values were null
+     */
```
rcaudy, line 852:
```suggestion
    /**
     * Returns an array with {@code null} values replaced with preceding non-{@code null} value.
     *
     * @param values Input values
     * @return An array of {@code T} with values determined by the preceding elements in {@code values}.
     *         Each element will take on the most recently encountered non-{@code null} value where one exists,
     *         or {@code null} if all preceding values were {@code null}.
     */
```

engine/function/src/templates/Basic.ftl
```diff
@@ -1688,6 +1714,14 @@ public class Basic {
         return ifelse(new BooleanVectorDirect(condition), trueCase, falseCase);
     }
 
+    /**
+     * Returns an array with null values replaced with preceding non-null value
```
kosak, line 1718: Copies the specified array, replacing elements that represent null in the Deephaven convention by the most recently encountered non-null value if one exists. Otherwise (if no such value exists), replaces those elements with Deephaven null.

[note slightly different wording: I ended with "Deephaven null" because this is an array of primitive type]

engine/function/src/templates/Basic.ftl
```diff
@@ -1688,6 +1714,14 @@ public class Basic {
...
+     *
+     * @param values values.
+     * @return An array of ${pt.primitive} whose values are determined by the preceding elements.
```
kosak, line 1721: @return A copy of the specified array, with Deephaven null elements replaced as described above. If the specified array is null, returns null.

engine/function/src/templates/Basic.ftl
```diff
@@ -1696,6 +1730,14 @@ public class Basic {
         return forwardFill(new ${pt.dbArrayDirect}(values));
     }
 
+    /**
+     * Returns an array with null values replaced with preceding non-null value
```
kosak, line 1734: Same as above.

engine/function/src/templates/Basic.ftl
```diff
@@ -1696,6 +1730,14 @@ public class Basic {
...
+     *
+     * @param values values.
+     * @return An array of ${pt.primitive} whose values are determined by the preceding elements.
```
kosak, line 1737: Same as above.

engine/function/src/templates/Basic.ftl
```diff
@@ -1688,6 +1714,14 @@ public class Basic {
...
+     *         most recently encountered non-null value where one exists
+     *         null if all preceding values were null
+     */
     public static ${pt.primitive}[] forwardFill(${pt.primitive}... values){
         if(values == null){
```
kosak, line 1726: Spacing: `if (`

So..... I've just noticed that there's a lot of these spacing things in `Basic.ftl` (not your fault). Assuming Chip agrees, maybe you can just do a global replace of `if(` to `if (` and `for(` to `for (` and `}{` to `} (` in this or a separate PR (or I can do it if that's more appropriate).

lbooker42: I did this, will make re-approval more difficult but looks better, agreed.

engine/function/src/templates/Numeric.ftl
```diff
@@ -525,7 +525,7 @@ public class Numeric {
 
         final long n = values.size();
 
-        if(n != weights.size()){
+        if (n != weights.size()){
```
kosak, line 528: I'm sorry, can you fix all of these also? I tried to mention it in my previous comment but I got the punctuation wrong.
Basically change `){` to `) {`

engine/function/src/templates/Numeric.ftl
```diff
@@ -757,7 +757,7 @@ public class Numeric {
             return NULL_DOUBLE;
         }
 
-        if(values.size() != weights.size()){
+        if (values.size() != weights.size()){
```
kosak, line 760: ...and here

engine/function/src/templates/Numeric.ftl
```diff
@@ -770,7 +770,7 @@ public class Numeric {
             final ${pt.primitive} v = values.get(i);
             final ${pt2.primitive} w = weights.get(i);
 
-            if(!isNull(v) && !isNull(w)){
+            if (!isNull(v) && !isNull(w)){
```
kosak, line 773: ...and here

engine/function/src/templates/Numeric.ftl
```diff
@@ -1125,7 +1125,7 @@ public class Numeric {
      * @return percentile, or null value in the Deephaven convention if values is null or empty.
      */
     public static double percentile(double percentile, ${pt.primitive}... values) {
-        if(values == null || values.length == 0){
+        if (values == null || values.length == 0){
```
kosak, line 1128: ...etc...

engine/function/src/templates/Basic.ftl
```diff
@@ -1688,25 +1715,43 @@ public class Basic {
...
+     * @param values values.
+     * @return A copy of the specified array, with Deephaven null elements replaced as described above. If the
+     *         specified array is null, returns null.
+     */
+    public static ${pt.primitive}[] forwardFill(${pt.dbArray} values) {
+        if (values == null) {
             return null;
         }
 
         final int n = values.intSize("reverse");
```
kosak, line 1749: "reverse" -> "forwardFill"

kosak (approved):

engine/function/src/templates/Basic.ftl
```diff
@@ -842,6 +842,33 @@ public class Basic {
...
+    /**
+     * Copies the specified array, replacing elements that represent null in the Deephaven convention by the most
+     * recently encountered non-null value if one exists. Otherwise (if no such value exists), replaces those elements
+     * with null.
+     *
+     * @param values values.
+     * @return A copy of the specified array, with Deephaven null elements replaced as described above. If the specified
+     *         array is null, returns null.
+     */
+    public static <T> T[] forwardFillObj(final T[] values) {
```
chipkent, line 854:
```suggestion
    @SafeVarargs
    public static <T> T[] forwardFillObj(final T... values) {
```
This change lets users provide args by hand.  I don't expect them to use this particular function this way, but users are creative -- and the change would be consistent with other methods.

engine/function/src/templates/Numeric.ftl
```diff
@@ -1449,7 +1449,7 @@ public class Numeric {
      * @param values values.
      * @return cumulative max of non-null values.
      */
-    public static ${pt.primitive}[] cumMax(${pt.primitive}... values) {
+    public static ${pt.primitive}[] cummax(${pt.primitive}... values) {
```
chipkent, line 1452: test me

engine/function/src/templates/Numeric.ftl
```diff
@@ -1418,7 +1418,7 @@ public class Numeric {
      * @param values values.
      * @return cumulative min of non-null values.
      */
-    public static ${pt.primitive}[] cumMin(${pt.primitive}... values) {
+    public static ${pt.primitive}[] cummin(${pt.primitive}... values) {
```
chipkent, line 1421: test me

chipkent, line 1421: Also need a DH vector version.

engine/function/src/templates/Numeric.ftl
chipkent, line 1452: Also need a DH vector version.

engine/function/src/templates/Basic.ftl
chipkent, line 854: also need a DH vector version.

chipkent (approved):

Issue #2835, opened 2022-09-12
Subplots not available when plotting OHLC plot


hhashim1:
Description:
An option to create subplots below the OHLC plot is needed to validate the data and calculations. While this can be accomplished by having separate charts, it saves a lot of time and much easier to have all the charts in one chart window. 

OHLC chart would have the OHLC data together with `plot_xy` plots and subplots would have other indicators and bar plots that cannot be charted over OHLC plot/pane. They share the same X axis which is time however they have a different Y axis.

Please see the below screenshot from tradingview. All stock trading softwares allow subplots as they are necessary to view all of the data together.

The below image has 6 subplots. They could be a bar, line, or any other type of plots. Tradingview would be a good tool to look at options when it comes to charting.

![image](https://user-images.githubusercontent.com/62855649/189680453-4cf2995e-ea7c-481b-b3c6-c458461d6f49.png)

hhashim1:
I am unable to make my code work based on your suggestion.

```
figure = Figure(figure(2,1))
bar_chart = figure.figure_title("O:TSLA220902P00273330")\
        .new_chart(row=0,col=0) \
        .plot_ohlc(series_name="t4", t=t4.where(filters=["symbol = 'O:TSLA220902P00273330'"]), x="tdate", open="bar_open", high="bar_high", low="bar_low", close="bar_close")\
        .new_chart(row=1,col=0) \
        .plot_xy(series_name="BB_U_1_0", t=t4.where(filters=["symbol = 'O:TSLA220902P00273330'"]), x="tdate", y="BB_U_1_0")\
        .show()
```

Here is the error:
```
r-Scheduler-Serial-1 | i.d.s.s.SessionState      | Internal Error '0a850b56-f4ba-435a-9467-7229b65c814d' java.lang.RuntimeException: Error in Python interpreter:
Type: <class 'TypeError'>
Value: 'Figure' object is not callable
Line: 1
Namespace: <module>
File: <string>
Traceback (most recent call last):
  File "<string>", line 1, in <module>```

**Comment by mofojed** [comment]
Subplots are now available, but we still cannot mix OHLC plots and XY plots in the same chart. E.g. doing something like:
```
from deephaven import time_table
from deephaven.plot import Figure

tt = time_table("00:00:00.01").update(["X=0.01*ii", "Y=ii*ii", "S=sin(X)", "C=cos(X)", "T=tan(X)"]).tail(1000)

f = Figure() \
[...]

PR #3042, opened 2022-10-27
Allow rows/cols to be passed in to python `Figure` constructor

mofojed:
- Partial #2835 
- Change the Figure API to take `rows` and `cols` as it's first two parameters, allowing writing code like `Figure(2,1)` to create a figure with 2 rows and 1 column
- **BREAKING**: Any code passing in their own instance of j_figure will need to specify `j_figure` as the param name now, e.g. `Figure(j_figure=my_j_figure)`
  - Signature of https://deephaven.io/core/pydoc/code/deephaven.plot.figure.html#deephaven.plot.figure.Figure changing
  - I figure it's better to have the rows/cols as defaults anyway
- Tested with a snippet of code creating multiple charts with varying positions of subplots (using Web UI https://github.com/deephaven/web-client-ui/pull/858):
```

tt = time_table("00:00:00.01").update(["X=0.01*ii", "Y=ii*ii", "S=sin(X)", "C=cos(X)", "T=tan(X)"]).tail(1000)

f = Figure() \
    .plot_xy("Y", t=tt, x="X", y="Y") \
    .plot_xy("S", t=tt, x="X", y="S") \
    .plot_xy("C", t=tt, x="X", y="C") \
    .plot_xy("T", t=tt, x="X", y="Y") \
    .show()

f2 = Figure(2, 1) \
    .new_chart(row=0,col=0) \
    .plot_xy("S", t=tt, x="X", y="S") \
    .new_chart(row=1,col=0) \
    .plot_xy("C", t=tt, x="X", y="C") \
    .show()

f3_r = Figure(2, 2) \
    .new_chart(row=0,col=0)\
    .plot_xy("Y", t=tt, x="X", y="S") \
    .new_chart(row=1,col=0) \
    .plot_xy("S", t=tt, x="X", y="C") \
    .new_chart(row=0,col=1).chart(row_span=2) \
    .plot_xy("T", t=tt, x="X", y="T") \
    .show()

f3_c = Figure(2, 2) \
    .new_chart(row=0,col=0)\
    .plot_xy("Y", t=tt, x="X", y="S") \
    .new_chart(row=0,col=1) \
    .plot_xy("S", t=tt, x="X", y="C") \
    .new_chart(row=1,col=0).chart(col_span=2) \
    .plot_xy("T", t=tt, x="X", y="T") \
    .show()

f4 = Figure(2,2) \
    .new_chart(row=0,col=0) \
    .plot_xy("Y", t=tt, x="X", y="Y") \
    .new_chart(row=0,col=1) \
    .plot_xy("S", t=tt, x="X", y="S") \
    .new_chart(row=1,col=0) \
    .plot_xy("C", t=tt, x="X", y="C") \
    .new_chart(row=1,col=1) \
    .plot_xy("T", t=tt, x="X", y="T") \
    .show()
```

### DIFF
--- a/Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
+++ b/Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
@@ -3,7 +3,7 @@
 #
 
 ######################################################################################################################
-#               This code is auto generated. DO NOT EDIT FILE!
+#               Preample code that gets added to generated code in GeneratePyV2FigureAPI.java
 # Run generatePythonFigureWrapper or "./gradlew :Generators:generatePythonFigureWrapper" to generate
 ######################################################################################################################
 """ This module implements the Figure class for creating plots, charts, line, axis, color, etc. """
@@ -104,9 +104,9 @@ class Figure(JObjectWrapper):
 
     j_object_type = jpy.get_type("io.deephaven.plot.Figure")
 
-    def __init__(self, j_figure: jpy.JType = None):
+    def __init__(self, rows: int = 1, cols: int = 1, j_figure: jpy.JType = None):
         if not j_figure:
-            self.j_figure = _JPlottingConvenience.figure()
+            self.j_figure = _JPlottingConvenience.figure(rows, cols)
         else:
             self.j_figure = j_figure
 
--- a/Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
+++ b/Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
@@ -3,7 +3,7 @@
 #
 
 ######################################################################################################################
-#               Preample code that gets added to generated code in GeneratePyV2FigureAPI.java
+#               This code is auto generated. DO NOT EDIT FILE!
 # Run generatePythonFigureWrapper or "./gradlew :Generators:generatePythonFigureWrapper" to generate
 ######################################################################################################################
 """ This module implements the Figure class for creating plots, charts, line, axis, color, etc. """
@@ -105,6 +105,13 @@ class Figure(JObjectWrapper):
     j_object_type = jpy.get_type("io.deephaven.plot.Figure")
 
     def __init__(self, rows: int = 1, cols: int = 1, j_figure: jpy.JType = None):
+        """ Initializes a Figure object that is used for displaying plots
+
+        Args:
+            rows (int, optional): Number of rows in the figure. Defaults to 1.
+            cols (int, optional): Number of columns in the figure. Defaults to 1.
+            j_figure (jpy.JType, internal): Internal use only.
+        """
         if not j_figure:
             self.j_figure = _JPlottingConvenience.figure(rows, cols)
         else:
--- a/py/server/deephaven/plot/figure.py
+++ b/py/server/deephaven/plot/figure.py
@@ -3,7 +3,7 @@
 #
 
 ######################################################################################################################
-#               This code is auto generated. DO NOT EDIT FILE!
+#               Preample code that gets added to generated code in GeneratePyV2FigureAPI.java
 # Run generatePythonFigureWrapper or "./gradlew :Generators:generatePythonFigureWrapper" to generate
 ######################################################################################################################
 """ This module implements the Figure class for creating plots, charts, line, axis, color, etc. """
@@ -932,13 +932,13 @@ class Figure(JObjectWrapper):
             dim = _convert_j("dim", dim, [int])
 
         if not non_null_args:
-            return Figure(self.j_figure.newAxes())
+            return Figure(j_figure=self.j_figure.newAxes())
         elif non_null_args == {"dim"}:
-            return Figure(self.j_figure.newAxes(dim))
+            return Figure(j_figure=self.j_figure.newAxes(dim))
         elif non_null_args == {"name"}:
-            return Figure(self.j_figure.newAxes(name))
+            return Figure(j_figure=self.j_figure.newAxes(name))
         elif non_null_args == {"name", "dim"}:
-            return Figure(self.j_figure.newAxes(name, dim))
+            return Figure(j_figure=self.j_figure.newAxes(name, dim))
         else:
             raise DHError(f"unsupported parameter combination: {non_null_args}")
 
@@ -974,11 +974,11 @@ class Figure(JObjectWrapper):
             col = _convert_j("col", col, [int])
 
         if not non_null_args:
-            return Figure(self.j_figure.newChart())
+            return Figure(j_figure=self.j_figure.newChart())
         elif non_null_args == {"index"}:
-            return Figure(self.j_figure.newChart(index))
+            return Figure(j_figure=self.j_figure.newChart(index))
         elif non_null_args == {"row", "col"}:
-            return Figure(self.j_figure.newChart(row, col))
+            return Figure(j_figure=self.j_figure.newChart(row, col))
         else:
             raise DHError(f"unsupported parameter combination: {non_null_args}")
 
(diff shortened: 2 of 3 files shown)


py/server/deephaven/plot/figure.py
```diff
@@ -104,9 +104,9 @@ class Figure(JObjectWrapper):
 
     j_object_type = jpy.get_type("io.deephaven.plot.Figure")
 
-    def __init__(self, j_figure: jpy.JType = None):
+    def __init__(self, rows: int = 1, cols: int = 1, j_figure: jpy.JType = None):
         if not j_figure:
-            self.j_figure = _JPlottingConvenience.figure()
+            self.j_figure = _JPlottingConvenience.figure(rows, cols)
```
jmao-denver, line 109: @mofojed figure.py is generated, see
```
######################################################################################################################
#               This code is auto generated. DO NOT EDIT FILE!
# Run generatePythonFigureWrapper or "./gradlew :Generators:generatePythonFigureWrapper" to generate
######################################################################################################################
```
We need to update 'io.deephaven.plot.PlotConvenience' first, and that may lead to other changes in the plot library which is a mixture of hand-written and generated code itself.

jmao-denver (changes requested):

py/server/deephaven/plot/figure.py
mofojed: Right - I totally forgot that file was generated. I'm an idiot, I've edited this before.
I'll see if I can make the proper changes in PlottingConvenience and such.

Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
```diff
@@ -104,9 +104,9 @@ class Figure(JObjectWrapper):
 
     j_object_type = jpy.get_type("io.deephaven.plot.Figure")
 
-    def __init__(self, j_figure: jpy.JType = None):
+    def __init__(self, rows: int = 1, cols: int = 1, j_figure: jpy.JType = None):
         if not j_figure:
-            self.j_figure = _JPlottingConvenience.figure()
+            self.j_figure = _JPlottingConvenience.figure(rows, cols)
         else:
             self.j_figure = j_figure
```
chipkent, line 111: It seems like we should throw an error if `j_figure` plus either `rows` or `cols` is passed in.

Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
```diff
@@ -104,9 +104,9 @@ class Figure(JObjectWrapper):
 
     j_object_type = jpy.get_type("io.deephaven.plot.Figure")
 
-    def __init__(self, j_figure: jpy.JType = None):
+    def __init__(self, rows: int = 1, cols: int = 1, j_figure: jpy.JType = None):
```
chipkent, line 107: Should pydoc the new args.  @jmao-denver how are we doing that for constructors?

Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
```diff
@@ -3,7 +3,7 @@
 #
 
 ######################################################################################################################
-#               This code is auto generated. DO NOT EDIT FILE!
+#               Preample code that gets added to generated code in GeneratePyV2FigureAPI.java
```
chipkent, line 6:
```suggestion
#               Preamble code that gets added to generated code in GeneratePyV2FigureAPI.java
```

py/server/deephaven/plot/figure.py
```diff
@@ -3,7 +3,7 @@
 #
 
 ######################################################################################################################
-#               This code is auto generated. DO NOT EDIT FILE!
+#               Preample code that gets added to generated code in GeneratePyV2FigureAPI.java
```
chipkent, line 6: Header is now incorrect after the change to `plotV2.py`.

Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
chipkent: Changing this header causes problems elsewhere.  With this change, the header in `figure.py` is now wrong.  It is clearly confusing, but it was probably better before this change.  To fix it properly would require a change to the generator.  Not sure if it is worth that right now.

chipkent (changes requested):

Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
jmao-denver: If a wrapper should not be instantiated directly, we choose to omit the docstring for the initializer.  With this change, we should document it now and it will be a bit tricky for the j_figure parameter.

chipkent: I guess offering `figure(...)` may make sense, but I'm not sure without looking at the whole API.  It would break all of our existing docs.

mofojed: I could just add rows/cols as 3rd and 4th parameter instead, and then you can still specify it by adding the keywords, e.g. `Figure(rows=2, cols=2)`, so it's not a breaking change.
I just didn't think it made much sense API-wise having users pass in their own `j_figure` instance, seems like an anti-pattern?

chipkent: It does not make sense for users to pass in their own `j_figure`.  As is, your current change should not break any code.  I think the change @jmao-denver is proposing is adding a factory method, analogous to the methods that create `Table`s (e.g. `empty_table`).  If such a constructor method is added, it should be a package-level function named `figure` and should not be @classmethod.

I'm not sure what my thoughts are on pushing users to this constructor vs a package-level constructor function, but we need docs either way.

jmao-denver: My suggestions really just try to avoid documenting the j_figure parameter for the constructor. What I had in mind is a 'of' classmethod.
```
    def __init__(rows, cols) -> Figure:
         """ ... """

    @classmethod
    def of(j_figure) -> Figure:
        ..
```
then we change the generator code to use this new 'of' method instead of use the j_figure keyword parameter.

chipkent: We don't do that anywhere else.  The constructor still needs `j_figure`, so the extra complexity doesn't buy us anything and makes stuff more complex.  I'm fine if we don't document `j_figure` or document it as internal use only.

jmao-denver (approved):

py/server/deephaven/plot/figure.py
```diff
@@ -104,9 +104,18 @@ class Figure(JObjectWrapper):
...
-    def __init__(self, j_figure: jpy.JType = None):
+    def __init__(self, rows: int = 1, cols: int = 1, j_figure: jpy.JType = None):
+        """ Initializes a Figure object that is used for displaying plots
+
+        Args:
+            rows (int): Number of rows in the figure. Defaults to 1.
+            cols (int): Number of columns in the figure. Defaults to 1.
+            j_figure (jpy.JType): Internal use only.
+        Raises:
+            DHError
```
jmao-denver, line 115: I don't see it in the preamble, stale code?

jmao-denver (changes requested):

jmao-denver (approved):

Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
mofojed: As discussed we'll just update the constructor.

Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
mofojed: When using default values (rows=1, cols=1), we can't easily tell if the user specified `Figure()` vs. `Figure(1,1)`, so wouldn't be able to accurately raise an error.

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/1845
Conceptual: https://github.com/deephaven/deephaven.io/issues/1848
Reference: https://github.com/deephaven/deephaven.io/issues/1846
Blog: https://github.com/deephaven/deephaven.io/issues/1847

PR #3045, opened 2022-10-31
Wrap Jetty's Resource type to force an invalid last-modified header

### LINKED ISSUE #3011: Caching issue when updating Web UI

niloc132:
Jetty always returns the last-modified date of the file, but Gradle unsets the "true" last modified date in favor of ensuring that build files cache only based on their contents. Disabling this feature is possible, but still may present as buggy in Jetty, such as if a 1.0 release occurred, then 2.0, then a 1.1 hotfix - if an installation of 1.1 were later replaced with 2.0, the 1.1 files would be "newer" so would be considered up to date.

This patch removes the last-modified value by only returning -1 when requested. We could go further and return an etag that actually maps to the release version or the contents of the file, but that is left for future work.

Fixes #3011

### DIFF
--- a/server/jetty/src/main/java/io/deephaven/server/jetty/ControlledCacheResource.java
+++ b/server/jetty/src/main/java/io/deephaven/server/jetty/ControlledCacheResource.java
@@ -1,0 +1,122 @@
+package io.deephaven.server.jetty;
+
+import org.eclipse.jetty.util.resource.Resource;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.nio.channels.ReadableByteChannel;
+
+/**
+ * Simple wrapper around the Jetty Resource type, to grant us control over caching features. The current implementation
+ * only removes the last-modified value, but a future version could provide a "real" weak/strong etag.
+ */
+public class ControlledCacheResource extends Resource {
+    private final Resource wrapped;
+
+    public ControlledCacheResource(Resource wrapped) {
+        this.wrapped = wrapped;
+    }
+
+    @Override
+    public boolean isContainedIn(Resource r) throws MalformedURLException {
+        return wrapped.isContainedIn(r);
+    }
+
+    @Override
+    public void close() {
+        wrapped.close();
+    }
+
+    @Override
+    public boolean exists() {
+        return wrapped.exists();
+    }
+
+    @Override
+    public boolean isDirectory() {
+        return wrapped.isDirectory();
+    }
+
+    @Override
+    public long lastModified() {
+        // Always return -1, so that we don't get the build system timestamp. In theory we could return the app startup
+        // time as well, so that clients that connect don't need to revalidate quite as often, but this could have other
+        // side effects such as in load balancing with a short-lived old build against a seconds-older new build.
+        return -1;
+    }
+
+    @Override
+    public long length() {
+        return wrapped.length();
+    }
+
+    @Override
+    public URI getURI() {
+        return wrapped.getURI();
+    }
+
+    @Override
+    public File getFile() throws IOException {
+        return wrapped.getFile();
+    }
+
+    @Override
+    public String getName() {
+        return wrapped.getName();
+    }
+
+    @Override
+    public InputStream getInputStream() throws IOException {
+        return wrapped.getInputStream();
+    }
+
+    @Override
+    public ReadableByteChannel getReadableByteChannel() throws IOException {
+        return wrapped.getReadableByteChannel();
+    }
+
+    @Override
+    public boolean delete() throws SecurityException {
+        return wrapped.delete();
+    }
+
+    @Override
+    public boolean renameTo(Resource dest) throws SecurityException {
+        return wrapped.renameTo(dest);
+    }
+
+    @Override
+    public String[] list() {
+        return wrapped.list();
+    }
+
+    @Override
+    public Resource addPath(String path) throws IOException, MalformedURLException {
+        // Re-wrap any instance that might be returned
+        return new ControlledCacheResource(wrapped.addPath(path));
+    }
+
+    @Override
+    public String toString() {
+        // Jetty's CachedContentFactory.CachedHttpContent requires that toString return the underlying URL found on
+        // disk, or else the mime lookup from content type won't resolve anything.
+        return wrapped.toString();
+    }
+
+    @Override
+    public int hashCode() {
+        // As with toString, delegating this to the wrapped instance, just in case there is some specific, expected
+        // behavior
+        return wrapped.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        // As with toString, delegating this to the wrapped instance, just in case there is some specific, expected
+        // behavior
+        return wrapped.equals(obj);
+    }
+}
--- a/server/jetty/src/main/java/io/deephaven/server/jetty/ControlledCacheResource.java
+++ b/server/jetty/src/main/java/io/deephaven/server/jetty/ControlledCacheResource.java
@@ -14,9 +14,16 @@ import java.nio.channels.ReadableByteChannel;
  * only removes the last-modified value, but a future version could provide a "real" weak/strong etag.
  */
 public class ControlledCacheResource extends Resource {
+    public static ControlledCacheResource wrap(Resource wrapped) {
+        if (wrapped instanceof ControlledCacheResource) {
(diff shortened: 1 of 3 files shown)


server/jetty/src/main/java/io/deephaven/server/jetty/ControlledCacheResource.java
```diff
@@ -0,0 +1,122 @@
...
+import java.nio.channels.ReadableByteChannel;
+
+/**
+ * Simple wrapper around the Jetty Resource type, to grant us control over caching features. The current implementation
+ * only removes the last-modified value, but a future version could provide a "real" weak/strong etag.
+ */
+public class ControlledCacheResource extends Resource {
+    private final Resource wrapped;
+
+    public ControlledCacheResource(Resource wrapped) {
```
mofojed, line 19: Seems like there's a lot of methods we're not overriding, such as `.toURL`, `.getURL`, `.setAssociate`, etc. Are we sure these are the only functions that need to be overridden? How did you decide what to override?

server/jetty/src/main/java/io/deephaven/server/jetty/ControlledCacheResource.java
```diff
@@ -0,0 +1,122 @@
...
+    public boolean isDirectory() {
+        return wrapped.isDirectory();
+    }
+
+    @Override
+    public long lastModified() {
+        // Always return -1, so that we don't get the build system timestamp. In theory we could return the app startup
+        // time as well, so that clients that connect don't need to revalidate quite as often, but this could have other
+        // side effects such as in load balancing with a short-lived old build against a seconds-older new build.
+        return -1;
```
mofojed, line 48: This will still be an issue with clients that have already stored the `Last-modified` header. Still get an issue with the client passing `If-Modified-Since: Fri, 01 Feb 1980 05:00:00 GMT`, I'm guessing somewhere Jetty is still checking against what the reported lastModified date is (even if it's not in a header?) and I'm still getting a 304: 

image

Everybody upgrading from v0.17 will still hit this issue and they'll be broken until they do a hard refresh.

mofojed (changes requested):

server/jetty/src/main/java/io/deephaven/server/jetty/ControlledCacheResource.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Override
+    public String[] list() {
+        return wrapped.list();
+    }
+
+    @Override
+    public Resource addPath(String path) throws IOException, MalformedURLException {
+        // Re-wrap any instance that might be returned
+        return new ControlledCacheResource(wrapped.addPath(path));
```
devinrsmith, line 99: Might make sense to check if `wrapped instanceof ControlledCacheResource` so we can reduce re-wrapping a bunch of times? Not sure if issue in practice, but something like:

```java
cacheResource.addPath("1").addPath("2").addPath("3")...
```

server/jetty/src/main/java/io/deephaven/server/jetty/DropIfModifiedSinceHeader.java
```diff
@@ -0,0 +1,70 @@
...
+         * @param request the {@link HttpServletRequest} to be wrapped.
+         * @throws IllegalArgumentException if the request is null
+         */
+        public HeaderRemovingHttpServletRequestWrapper(HttpServletRequest request, String headerToRemove) {
+            super(request);
+            this.headerToRemove = headerToRemove;
+        }
+
+        @Override
+        public String getHeader(String name) {
```
devinrsmith, line 48: I think `getDateHeader` (and `getIntHeader`) should be overridden as well, yeah?

server/jetty/src/main/java/io/deephaven/server/jetty/ControlledCacheResource.java
devinrsmith: Probably makes sense to test alongside the JS plugin PR which creates zip resources.

server/jetty/src/main/java/io/deephaven/server/jetty/DropIfModifiedSinceHeader.java
niloc132: Ah probably, though from testing it doesn't seem to matter.

server/jetty/src/main/java/io/deephaven/server/jetty/ControlledCacheResource.java
niloc132: I can add that check, but each addPath will only wrap the next layer, right?
 * `cacheResource` is singly wrapped
 * `resource.addPath("1")` returns a singly wrapped instance, `resource1`
 * `resource1.addPath("2")` returns a singly wrapped instance, etc
 
Unless the initially created ControlledCacheResource is multiply wrapped to begin with, _or_ the wrapped resource itself somehow contains already-wrapped resources (for example a ResourceCollection full of already wrapped resources, and then wrap the ResourceCollection again), I can't see a way for this to happen? What am I missing?

Mitigation of the theoretical issue will be replacing the public ctor with a static wrap() method that does the check, and delegate to that when wrapping a sub-resource.

server/jetty/src/main/java/io/deephaven/server/jetty/ControlledCacheResource.java
niloc132: I only overrode the abstract methods, and only modified the one that was specific to last-modified. The rest of the methods are implemented by the superclass Resource to rely on the abstract methods. 

 * setAssociate is kind of a special case - it isnt clear to me how it is used, and has explicit docs that it isn't actually used...
 * isSame already delegates to equals (allowing for a subclass to override isSame separately), but we already override equals
 * getResource delegates to addPath, which we override
 * isAlias calls getAlias, and getAlias has a sane default of null for resources that dont alias anything. It looks like this is meant for soft links, absolute path vs canonical path, but we're only dealing with paths inside of jars, so it shouldnt ever matter.
 * getListHTML generates HTML and has a default impl that we don't want to own
 * copyTo arguably does need us to implement it differently to work properly, but at the same time, we don't really support this - internally it is used for unpacking web archives to read them off of the disk
 * getAllResources delegates to list() and addPath(), both of which we implement, no other implementation overrides this default impl
 * 

That said, toURL is static, so doesnt need to be overridden, and getURL _is_ overridden in this class (since it is abstract).

server/jetty/src/main/java/io/deephaven/server/jetty/ControlledCacheResource.java
niloc132: This should be fixed now, thanks for bringing it up.

server/jetty/src/main/java/io/deephaven/server/jetty/ControlledCacheResource.java
mofojed: Nit-pick - `getURI()` is overridden, not `getURL()`... but looks like `getURL` is deprecated anyway so I think that's okay.

mofojed (approved):

devinrsmith (approved):
LGTM - tested w/ the js plugins PR, successfully removed my ability to get 304s out w/ `If-Modified-Since`.

PR #3056, opened 2022-11-02
Add js-plugins pre-built support

devinrsmith:
Fast and simple js plugin support via the configuration property `deephaven.jsPlugins.resourceBase` - assumes the user has already built a js plugins compatible directory to the necessary specifications (the server transparently passes through the data with no knowledge of the structure).

Niceties _can_ be added as documentation and additional follow-up. For example, the user can extend from the Deephaven official images and use the web-plugin-packager or similar construction in a multi-stage Dockerfile build to add the necessary structure to their own image.

### DIFF
--- a/server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
+++ b/server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
@@ -30,6 +30,7 @@ import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.server.SslConnectionFactory;
 import org.eclipse.jetty.server.handler.HandlerCollection;
+import org.eclipse.jetty.servlet.DefaultServlet;
 import org.eclipse.jetty.servlet.ErrorPageErrorHandler;
 import org.eclipse.jetty.servlet.FilterHolder;
 import org.eclipse.jetty.util.resource.Resource;
@@ -73,6 +74,7 @@ public class JettyBackedGrpcServer implements GrpcServer {
         } catch (IOException ioException) {
             throw new UncheckedIOException(ioException);
         }
+        context.setInitParameter(DefaultServlet.CONTEXT_INIT + "dirAllowed", "false");
 
         // For the Web UI, cache everything in the static folder
         // https://create-react-app.dev/docs/production-build/#static-file-caching
@@ -122,13 +124,11 @@ public class JettyBackedGrpcServer implements GrpcServer {
             });
         }
 
+        // Note: handler order matters
         HandlerCollection handlers = new HandlerCollection();
-
         // Set up /js-plugins/*
         JsPlugins.maybeAdd(handlers::addHandler);
-
         handlers.addHandler(context);
-
         jetty.setHandler(handlers);
     }
 
--- a/server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
+++ b/server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
@@ -29,6 +29,7 @@ import org.eclipse.jetty.server.SecureRequestCustomizer;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.server.SslConnectionFactory;
+import org.eclipse.jetty.server.handler.HandlerCollection;
 import org.eclipse.jetty.servlet.ErrorPageErrorHandler;
 import org.eclipse.jetty.servlet.FilterHolder;
 import org.eclipse.jetty.util.resource.Resource;
@@ -85,17 +86,11 @@ public class JettyBackedGrpcServer implements GrpcServer {
         // Add an extra filter to redirect from / to /ide/
         context.addFilter(HomeFilter.class, "/", EnumSet.noneOf(DispatcherType.class));
 
-        // Direct jetty all use this configuration as the root application
-        context.setContextPath("/");
-
         // Handle grpc-web connections, translate to vanilla grpc
         context.addFilter(new FilterHolder(new GrpcWebFilter()), "/*", EnumSet.noneOf(DispatcherType.class));
 
         // Wire up the provided grpc filter
         context.addFilter(new FilterHolder(filter), "/*", EnumSet.noneOf(DispatcherType.class));
-
-        // Set up /js-plugins/*
-        JsPlugins.maybeAdd(context);
 
         // Set up websockets for grpc-web - depending on configuration, we can register both in case we encounter a
         // client using "vanilla"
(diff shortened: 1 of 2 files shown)


devinrsmith:
Here was my local workflow to get up and running:

```
mkdir /tmp/js-plugins
cd /tmp/js-plugins
wget https://registry.npmjs.org/@deephaven/js-plugin-plotly/-/js-plugin-plotly-0.1.0.tgz
tar xzvf js-plugin-plotly-0.1.0.tgz
rm js-plugin-plotly-0.1.0.tgz
mkdir @deephaven
mv package @deephaven/js-plugin-plotly
echo '{"plugins":[{"name":"@deephaven/js-plugin-plotly","version":"0.1.0","main":"dist/index.js"}]}' > manifest.json
```

```
./gradlew py-server:assemble server-jetty-app:installDist
python3 -m venv /tmp/dh
source /tmp/dh/bin/activate
pip install py/server/build/wheel/deephaven_core-0.18.0-py3-none-any.whl
pip install deephaven-plugin-plotly
START_OPTS="-Ddeephaven.jsPlugins.resourceBase=/tmp/js-plugins" ./server/jetty-app/build/install/server-jetty/bin/start
```

(Executed in web ui)
```python
import plotly.express as px
df = px.data.iris()
fig = px.scatter(df, x="sepal_width", y="sepal_length", color="species", size='petal_length', hover_data=['petal_width'])
```

mofojed:
Tested further by adding the matplotlib from a locally built js plugin as well. So:
```
pip install deephaven-plugin-matplotlib
```
Then cloned the deephaven-js-plugins repo: https://github.com/deephaven/deephaven-js-plugins
In that directory, ran:
```
npm install
npm run build
```
Add a link to the directory:
```
ln -s /home/bender/dev/deephaven/deephaven-js-plugins/matplotlib /tmp/js-plugins/@deephaven/js-plugin-matplotlib
```
Update the manifest:
```
echo '{"plugins":[{"name":"@deephaven/js-plugin-plotly","version":"0.1.0","main":"dist/index.js"},{"name":"@deephaven/js-plugin-matplotlib","version":"0.1.0","main":"dist/index.js"}]}' > manifest.json
```
Starting up the server as mentioned by Devin above.
Then running the code snippet:
```
import matplotlib.pyplot as plt
fig = plt.figure()
ax = fig.subplots()  # Create a figure containing a single axes.
ax.plot([1, 2, 3, 4], [4, 2, 6, 7])  # Plot some data on the axes.
```

However was getting a 404 when trying to load js-plugin matplotlib. Seems to work fine if I copy instead of making a symlink. Is there a reason why a symlink doesn't work here, guessing how Jetty loads resources? Would be handy for dev... though could have a build script to copy it as well.
image

Also - anyway to make it so manifest.json gets updated automatically?

devinrsmith:
I haven't looked into symlinks, may be related to https://github.com/eclipse/jetty.project/issues/8259? I can dig in...

devinrsmith:
As far as "manifest.json gets updated automatically" - with respect to *this* PR, _no_.

As part of the alternative PR (#2908), yes, that PR makes the server responsible for manifest.json.

devinrsmith:
I'm not able to get symlinks working either - the code suggests that it should work, but there is probably some missing context.

I've tried context.clearAliasChecks();

> By default, the context is created with the {@link AllowedResourceAliasChecker} which is configured to allow symlinks. If this alias checker is not required, then {@link #clearAliasChecks()} or {@link #setAliasChecks(List)} should be called.

Maybe Colin knows?

mofojed:
I don't think the symlink thing should block this PR, since the workaround is just copy, or even the following:
1. Clone the deephaven-js-plugins repo
2. Add a manifest.json with the plugins you want to load, e.g. `echo '{"plugins":[{"name":"matplotlib","version":"0.1.0","main":"dist/index.js"},{"name":"plotly","version":"0.1.0","main":"dist/index.js"}]}' > manifest.json`
3. Start up pointing to the `deephaven-js-plugins` directory as the js-plugins root, e.g. `START_OPTS="-Ddeephaven.jsPlugins.resourceBase=/home/bender/dev/deephaven/deephaven-js-plugins" ./server/jetty-app/build/install/server-jetty/bin/start`

That works well... _however_ there still is a caching issue - it's just loading the js plugins from the disk cache. I think we need to add different caching headers there.

devinrsmith:
Ah yes, I think we should add the nocache filter for these paths - will do that.

devinrsmith:
Hmmm... I pushed a new commit, but it's not taking to this PR yet. Odd.

mofojed (changes requested):
Err re-verting approval until PR updates wtih the latest but the code works

mofojed (dismissed):

server/jetty/src/main/java/io/deephaven/server/jetty/JsPlugins.java
```diff
@@ -0,0 +1,40 @@
...
+            Resource resource = ControlledCacheResource.wrap(Resource.newResource(resourceBase));
+            WebAppContext context =
+                    new WebAppContext(null, "/js-plugins/", null, null, null, new ErrorPageErrorHandler(), NO_SESSIONS);
+            context.setBaseResource(resource);
+            context.setInitParameter(DefaultServlet.CONTEXT_INIT + "dirAllowed", "false");
+            // Suppress warnings about security handlers
+            context.setSecurityHandler(new ConstraintSecurityHandler());
+            addHandler.accept(context);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
```
niloc132, line 37: Illegal state ex here instead, indicate that the dir doesnt exist?

devinrsmith: Done

niloc132 (approved):
Approved, review over slack

mofojed (approved):

server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
```diff
@@ -122,13 +124,12 @@ public <T> T getEndpointInstance(Class<T> endpointClass) throws InstantiationExc
             });
         }
 
+        // Note: handler order matters due to pathSpec order
         HandlerCollection handlers = new HandlerCollection();
-
         // Set up /js-plugins/*
         JsPlugins.maybeAdd(handlers::addHandler);
-
+        // Set up /*
```
mofojed, line 131: /*?

devinrsmith: This is the handler for pathspecs matching `/*` - so everything that doesn't match `/js-plugins/*`.

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/1859
Conceptual: https://github.com/deephaven/deephaven.io/issues/1857
Reference: https://github.com/deephaven/deephaven.io/issues/1856
Blog: https://github.com/deephaven/deephaven.io/issues/1858

PR #3057, opened 2022-11-02
Add dh-defaults.vmoptions file

### LINKED ISSUE #3054: Reduce duplication for suggested defaults for java VM options
Currently, our recommended defaults "-XX:+UseG1GC -XX:MaxGCPauseMillis=100 -XX:+UseStringDeduplication".

Follow-up to https://github.com/deephaven/deephaven-core/pull/3009

May be worthwhile to add a new file into our .tar distribution (like we do for compiler directives), and potentially use -XX:VMOptionsFile=filename.

devinrsmith:
Consolidates the recommended VM options into a single place. Allows for these options to be defined in a single place.

Moves dh-compiler-directives.txt and dh-defaults.vmoptions into the etc directory (instead of in lib alongside the jars).

Fixes #3054

### DIFF
--- a/buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
+++ b/buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
@@ -92,16 +92,16 @@ tasks.withType(JavaCompile).configureEach {
 
 def createCompilerDirectives = tasks.register('createCompilerDirectives') {
   def compilerDirectivesFile = project.layout.buildDirectory.file('dh-compiler-directives.txt')
+  def compilerDirectivesText = new JsonBuilder([{
+    match (['*.*'] as List)
+    // Note: there seems to be a bug where this option doesn't actually get picked up
+    // So using '-XX:DisableIntrinsic=_currentThread' explicitly
+    // DisableIntrinsic('_currentThread')
+  }]).toPrettyString()
+  it.inputs.property('compilerDirectivesText', compilerDirectivesText)
   it.outputs.file(compilerDirectivesFile)
-
   doFirst {
-    def builder = new JsonBuilder([{
-      match (['*.*'] as List)
-      // Note: there seems to be a bug where this option doesn't actually get picked up
-      // So using '-XX:DisableIntrinsic=_currentThread' explicitly
-      // DisableIntrinsic('_currentThread')
-    }])
-    compilerDirectivesFile.get().asFile.text = builder.toPrettyString()
+    compilerDirectivesFile.get().asFile.text = compilerDirectivesText
   }
 }
 
@@ -123,18 +123,26 @@ def devJvmArgs = [
 //  '-XX:+PrintCompilation',          // this optional line shows jit operations as they happen
 ]
 
+// These are supposed to be generally applicable and recommended JVM options, but they aren't hard requirements.
+// Overly specific options do *not* belong here. For example, we should _not_ be setting something like `-Xmx4g` here.
+// If you are tempted to try and put system properties here (`-Dkey=value`), think again; there should be a more
+// appropriate place to set those.
+//
+// From the perspective of our application distribution, these options will be used as defaults for JAVA_OPTS
+// (if the user already has JAVA_OPTS set, these VM options will _not_ apply).
+def defaultJvmOptions = [
+  '-XX:+UseG1GC',
+  '-XX:MaxGCPauseMillis=100',
+  '-XX:+UseStringDeduplication',
+]
+
 def createVmOptions = tasks.register('createVmOptions') {
   def vmOptionsFile = project.layout.buildDirectory.file('dh-default.vmoptions')
+  def vmOptionsText = defaultJvmOptions.join('\n')
+  it.inputs.property('vmOptionsText', vmOptionsText)
   it.outputs.file(vmOptionsFile)
   doFirst {
-    // These are supposed to be generally applicable and recommended JVM options, but they aren't hard requirements.
-    // Overly specific options do *not* belong here. For example, we should _not_ be setting something like `-Xmx4g` here.
-    // If you are tempted to try and put system properties here (`-Dkey=value`), think again; there should be a more
-    // appropriate place to set those.
-    //
-    // From the perspective of our application distribution, these options will be used as defaults for JAVA_OPTS
-    // (if the user already has JAVA_OPTS set, these VM options will _not_ apply).
-    vmOptionsFile.get().asFile.text = "-XX:+UseG1GC\n-XX:MaxGCPauseMillis=100\n-XX:+UseStringDeduplication\n"
+    vmOptionsFile.get().asFile.text = vmOptionsText
   }
 }
 
--- a/buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
+++ b/buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
@@ -123,19 +123,41 @@ def devJvmArgs = [
 //  '-XX:+PrintCompilation',          // this optional line shows jit operations as they happen
 ]
 
+def createVmOptions = tasks.register('createVmOptions') {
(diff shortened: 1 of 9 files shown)


buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
```diff
@@ -123,30 +123,54 @@ def devJvmArgs = [
...
+  jvmArgs += compilerArgs(compilerDirectivesFile.singleFile.path) + ["-XX:VMOptionsFile=${vmOptionsFile.singleFile.path}"] + devJvmArgs
 }
 
 tasks.withType(GroovyCompile).configureEach {
   javaLauncher.set groovyCompilerLauncher
 }
 
 plugins.withType(ApplicationPlugin) {
-  applicationDistribution.into('lib') {
+  applicationDistribution.into('etc') {
```
rcaudy, line 168: Why `etc` instead of `lib`?

buildSrc/src/main/resources/unixStartScript.txt
```diff
@@ -47,7 +47,13 @@ APP_BASE_NAME=`basename "\$0"`
 DEFAULT_JVM_OPTS=${defaultJvmOpts}
 
 # Customization for deephaven-core to reference a compiler directives file
-DEFAULT_JVM_OPTS="\${DEFAULT_JVM_OPTS} -XX:+UnlockDiagnosticVMOptions -XX:CompilerDirectivesFile=\"\${APP_HOME}/lib/dh-compiler-directives.txt\""
+# There is no easy way for users to override these options
+DEFAULT_JVM_OPTS="\${DEFAULT_JVM_OPTS} -XX:+UnlockDiagnosticVMOptions -XX:CompilerDirectivesFile=\"\${APP_HOME}/etc/dh-compiler-directives.txt\""
+
+# Customization for deephaven-core to include vm options file by default
+# If users explicitly set JAVA_OPTS, the will be opting out of dh-default.vmoptions
```
rcaudy, line 54:
```suggestion
# If users explicitly set JAVA_OPTS, they will be opting out of dh-default.vmoptions
```

server/netty-app/build.gradle
```diff
@@ -76,12 +76,6 @@ if (hasProperty('quiet')) {
 tasks.withType(JavaExec).configureEach {
     // This appends to the existing jvm args, so that java-open-nio still takes effect
     jvmArgs extraJvmArgs
-
-    jvmArgs '-XX:+UseG1GC',
-            '-XX:MaxGCPauseMillis=100',
-            '-XX:+UseStringDeduplication',
-            // For development in this repository via Gradle, use the top-level data directory
-            "-Dstorage.path=${rootDir}/data"
```
rcaudy, line 84: We don't need/want this anymore?

rcaudy:
This needs testing on each of our supported JVM major versions, I presume?

devinrsmith:
It's officially documented as available on 11 and 17, but I'll make sure to do some explicit tests on a variety of JVMs - will report back.

devinrsmith:
I've successfully tested on Temurin 11, 17, and 19 as well as Graal 22.3.0 11, 17, and 19 on Linux.

buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
devinrsmith: It's a better layout IMO, not mixing these files amongst the lib/*.jars. That said, I can easily revert this change.

```
$ find . -not -name "*.jar" 
.
./server-jetty
./server-jetty/lib
./server-jetty/bin
./server-jetty/bin/start
./server-jetty/bin/start.bat
./server-jetty/etc
./server-jetty/etc/dh-compiler-directives.txt
./server-jetty/etc/dh-default.vmoptions
```

I suspect in the future we may want to simplify how we declare our classpath. Right now it's a really long list, and it might be wise to do:

```
CLASSPATH=$APP_HOME/lib/*
```

in our start script. (Note this construction is not a shell glob - it's passed to the JVM w/ a `*` and the JVM works out the classpath.) It's a bit sloppy to have this match against the files shown here in etc.

I think we might find we also want to publish other "non-essential" bits in the future - maybe python requirements.txt files to say "these are the exact libraries we tested against". I think laying the groundwork for these potential bits makes sense.

server/netty-app/build.gradle
devinrsmith: The defaults should be good enough IMO.

The equivalent was removed from jetty-app, and was accidentally left behind here. storage.path defaults to `<deephaven.dataDir>/storage`, and for local development (prepareCompose, docker compose up) that equates to `${rootDir}/data/storage` given:

```Dockerfile
    volumes:
      - ./data:/data
```

For native local development (`./gradlew ... :run`), the default will use OS conventional paths.

buildSrc/src/main/resources/unixStartScript.txt
devinrsmith: Done

server/netty-app/build.gradle
rcaudy: Might changing this break persisted notebooks for users? Do we need to put out a notice of the change?

rcaudy (approved):
Approved. Do note: https://github.com/deephaven/deephaven-core/pull/3057#discussion_r1013210718

server/netty-app/build.gradle
devinrsmith: This image is used for local development only.

buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
```diff
@@ -123,30 +123,54 @@ def devJvmArgs = [
...
+  it.outputs.file(vmOptionsFile)
+  doFirst {
+    // These are supposed to be generally applicable and recommended JVM options, but they aren't hard requirements.
+    // Overly specific options do *not* belong here. For example, we should _not_ be setting something like `-Xmx4g` here.
+    // If you are tempted to try and put system properties here (`-Dkey=value`), think again; there should be a more
+    // appropriate place to set those.
+    //
+    // From the perspective of our application distribution, these options will be used as defaults for JAVA_OPTS
+    // (if the user already has JAVA_OPTS set, these VM options will _not_ apply).
+    vmOptionsFile.get().asFile.text = "-XX:+UseG1GC\n-XX:MaxGCPauseMillis=100\n-XX:+UseStringDeduplication\n"
```
niloc132, line 137: isn't there a groovy-ism that can make this easier to read slash harder to break the `\n`s as we go? ideally we give it an array of individual options, it puts each on a new line, since that appears to be the expected format (rather than just join on whitespace)

buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
niloc132: I'd probably vote against `*` on the classpath value, gradle's generated list gives us consistent ordering, and I don't think we're free of multiple jars containing the same-named classes yet.

The more I look at it, the more I like the etc/lib split, just in case something tries to load from classpath and gets configs - app.d stuff probably should be kept away from configs in the same way, to ensure that a groovy "class" there can't grab on to other files except by explicitly using the filesystem.

buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
devinrsmith: Done

niloc132 (approved):

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/1919
Conceptual: https://github.com/deephaven/deephaven.io/issues/1921
Reference: https://github.com/deephaven/deephaven.io/issues/1920
Blog: https://github.com/deephaven/deephaven.io/issues/1922

PR #3058, opened 2022-11-03
Fix getOrSet when environment variables are set

devinrsmith:
Follow-up to #3035


devinrsmith:
Ensuring that the environment variables are correctly accounted for - testing against a smoke test for the real docker building process. Will report back here once I can confirm this fixes the issue.

rcaudy (approved):

devinrsmith:
Verified fix.

PR #3064, opened 2022-11-04
Multiplexed websocket instances should remove handlers when finished

### LINKED ISSUE #3063: JS gRPC MultiplexedWebsocketTransport doesn't correctly remove websocket event listeners
Over time, as new streams are created on the same underlying websocket, we only add listeners and never remove them. This will leak memory and some CPU as defunct handlers are left attached.

niloc132:
Also includes a workaround for a NodeJS WebSocket polyfill, which doesn't use the expected CloseEvent/MessageEvent constructors.

Fixes #3063

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/grpc/MultiplexedWebsocketTransport.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/grpc/MultiplexedWebsocketTransport.java
@@ -10,6 +10,7 @@ import elemental2.core.JsError;
 import elemental2.core.Uint8Array;
 import elemental2.dom.CloseEvent;
 import elemental2.dom.Event;
+import elemental2.dom.EventListener;
 import elemental2.dom.MessageEvent;
 import elemental2.dom.URL;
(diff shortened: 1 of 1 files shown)


devinrsmith (approved):
Structurally, looks OK.  I don't really know how the JS layers work. Assuming this covers all the `webSocket.addEventListener` cases, and we think it covers all the cases we need for removeHandlers(), LGTM.

mattrunyon (approved):
Appears to be working. In the browser I set a breakpoint where the websocket is created in dh-core and stored as a global variable. Then calling `getEventListeners(temp1)` you can see the listeners attached. Started at 6. Opening a table went to 8. Closing the panel went to 7. Re-opening went back to 8. Closing again back to 7.

In Node for the snapshot tool, the number of handlers slowly creeps up, but I may be flooding the server or the cleanup is being delayed. Or I have a case I'm not handling in the overload I have to do to patch the Node ws to more closely match the browser. I'm seeing about 75% on a given event are closed, but possible it's entirely my fault

PR #3065, opened 2022-11-04
Set the defaults for start_jvm in code

### LINKED ISSUE #3073: start_jvm.DEFAULT_JVM_ARGS not being used
`deephaven_server.Server` is calling `start_jvm(jvm_args=jvm_args)` with `jvm_args=None` by default. This overrides the more appropriate defaults that `start_jvm` itself is setting. The end result is that the `start_jvm.DEFAULT_JVM_ARGS` are not being set for the python embedded server - no `-Xrs`, nor `-XX:DisableIntrinsic=_currentThread`.

devinrsmith:
Fixes #3073

### DIFF
--- a/py/embedded-server/deephaven_server/start_jvm.py
+++ b/py/embedded-server/deephaven_server/start_jvm.py
@@ -46,14 +46,22 @@ DEFAULT_JVM_ARGS = [
 
 # Provide a util func to start the JVM, will use its own defaults if none are offered
 def start_jvm(
-        jvm_args = DEFAULT_JVM_ARGS,
-        jvm_properties = DEFAULT_JVM_PROPERTIES,
-        java_home: str = os.environ.get('JAVA_HOME', None),
+        jvm_args = None,
+        jvm_properties = None,
+        java_home = None,
         extra_classpath = [],
         propfile: str = None,
         config = None):
     """ This function uses the default DH property file to embed the Deephaven server and starts a Deephaven Python
     Script session. """
+    if not jvm_args:
+        jvm_args = DEFAULT_JVM_ARGS
+
+    if not jvm_properties:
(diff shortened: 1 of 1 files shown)


py/embedded-server/deephaven_server/start_jvm.py
```diff
@@ -46,14 +46,22 @@ def _jars():
...
+        jvm_args = None,
+        jvm_properties = None,
+        java_home = None,
         extra_classpath = [],
         propfile: str = None,
         config = None):
     """ This function uses the default DH property file to embed the Deephaven server and starts a Deephaven Python
     Script session. """
+    if not jvm_args:
+        jvm_args = DEFAULT_JVM_ARGS
```
devinrsmith, line 58: Alternatively, we could use the style `jvm_args = jvm_args or DEFAULT_JVM_ARGS`, thoughts?

jmao-denver: I think it might be more Pythonic to change the server __init__ to 
```
if jvm_args:
    start_jvm(jvm_args = jvm_args)
else:
   start_jvm()
```

devinrsmith: Arguably, we _should_ be plumbing through jvm_properties and jvm_home though too - once you have more than one argument, the approach you are suggesting doesn't work...

jmao-denver (approved):

niloc132:
It looks like you may want to revisit [this](https://github.com/deephaven/deephaven-core/commit/595a79641070eac7c051919b2c47f7e191b8e49e) and revert the changes to start_jvm.py - when the user passes no jvm args, we know they arent setting GC, so they should get our defaults?

PR #3070, opened 2022-11-06
Provide default directory for js plugins

devinrsmith:
I'd like to consider moving towards a model where "if you follow directory conventions, you don't need to explicitly set configuration properties".

As a motivating example, I'd like consumers of our docker images to install a JS plugins without needing to explicitly include a configuration property (file or system property).

```Dockerfile
# syntax=docker/dockerfile:1.4

FROM ghcr.io/deephaven/web-plugin-packager:latest as build
RUN ./pack-plugins.sh @deephaven/js-plugin-matplotlib

FROM ghcr.io/deephaven/server:latest
RUN pip install deephaven-plugin-matplotlib
COPY --link --from=build js-plugins/ /opt/deephaven/config/js-plugins/
```

As it stands now, the user would either need to explicitly add `deephaven.prop`:

```properties
includefiles=dh-defaults.prop

deephaven.jsPlugins.resourceBase=/opt/deephaven/config/js-plugins
```

```Dockerfile
COPY --link deephaven.prop /opt/deephaven/config/deephaven.prop
```

or, add a system property (which I want to try and discourage as much as possible):

```Dockerfile
ENV START_OPTS="-Ddeephaven.jsPlugins.resourceBase=/opt/deephaven/config/js-plugins"
```

This example is based off of https://github.com/deephaven-examples/deephaven-matplotlib-base/pull/3

### DIFF
--- a/server/jetty/src/main/java/io/deephaven/server/jetty/JsPlugins.java
+++ b/server/jetty/src/main/java/io/deephaven/server/jetty/JsPlugins.java
@@ -1,15 +1,20 @@
 package io.deephaven.server.jetty;
 
+import io.deephaven.configuration.ConfigDir;
 import io.deephaven.configuration.Configuration;
 import org.eclipse.jetty.security.ConstraintSecurityHandler;
 import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.servlet.DefaultServlet;
 import org.eclipse.jetty.servlet.ErrorPageErrorHandler;
+import org.eclipse.jetty.util.resource.PathResource;
 import org.eclipse.jetty.util.resource.Resource;
 import org.eclipse.jetty.webapp.WebAppContext;
+import org.jetbrains.annotations.NotNull;
 
 import java.io.IOException;
-import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Optional;
 import java.util.function.Consumer;
 
 import static org.eclipse.jetty.servlet.ServletContextHandler.NO_SESSIONS;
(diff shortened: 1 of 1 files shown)


devinrsmith:
Would be able to update and simplify https://github.com/deephaven/deephaven.io/pull/1947

devinrsmith:
This is a fine feature request, irrespective of this PR IMO.

You can't pass environment variables through `./gradlew ...` btw, that's not been a feature we've supported.

server/jetty/src/main/java/io/deephaven/server/jetty/JsPlugins.java
```diff
@@ -1,40 +1,65 @@
...
+import java.nio.file.Path;
+import java.util.Optional;
 import java.util.function.Consumer;
 
 import static org.eclipse.jetty.servlet.ServletContextHandler.NO_SESSIONS;
 
 class JsPlugins {
 
     public static void maybeAdd(Consumer<Handler> addHandler) {
+        resource()
```
mofojed, line 25: Can we log what directory we're reading from? If I'm running `./gradlew server-jetty-app:run`, I really don't know where this default directory is. Some logging would at least help debugging, or give me a clue to where it's looking.

Less of a concern when running from a `Dockerfile`

devinrsmith: Do you think logging in the no-property-set, no-default-directory-exists, is useful? I think logging that "no property has been set, no js-plugins directory existst here" is a bit extraneous (I imagine if we revamp our configuration framework, this might be a _general_ feature for we'd like for debugging any property).

I think logging the positive is completely fine. Will add this to this PR.

https://deephaven.io/core/groovy/docs/how-to-guides/configuration/js-plugins/ will be updated when this PR lands, describing that JS Plugins inherits the default directory of `<configDir>/js-plugins/`.

The configuration directory depends what system you are running on, see https://deephaven.io/core/docs/how-to-guides/configuration/native-application/#config-directory

devinrsmith:
@mofojed - if you try your previous run with `DEEPHAVEN_CONFIG_DIR` instead of `DEEPHAVEN_DATA_DIR`, I think it will work.

server/jetty/src/main/java/io/deephaven/server/jetty/JsPlugins.java
mofojed: I think it would be useful if the directory doesn't exist so you know what it's looking for. I'm still not certain where the "default" config directory is.

mofojed (approved):
Setting `DEEPHAVEN_CONFIG_DIR` instead of `DEEPHAVEN_DATA_DIR` worked.

server/jetty/src/main/java/io/deephaven/server/jetty/JsPlugins.java
devinrsmith: > The config directory is based on OS-conventions by default. On Linux, this is `$HOME/.config/<application>/`, and on macOS this is `$HOME/Library/Application Support/io.Deephaven-Data-Labs.<application>/`.

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2039
Conceptual: https://github.com/deephaven/deephaven.io/issues/2041
Reference: https://github.com/deephaven/deephaven.io/issues/2040
Blog: https://github.com/deephaven/deephaven.io/issues/2042

Issue #2997, opened 2022-10-13
Python debuggers do not work on code entered through the IDE


chipkent:
In these two repositories, I have been working on getting a debugger working with DH.
https://github.com/chipkent/test-dh-docker-pycharm
https://github.com/chipkent/test-dh-docker-vscode

In both cases, the debugger works for code executed via a `.py` file, but it does not work for code entered via the DH IDE.  As a specific example, code entered via the DH IDE does not trigger breakpoints.

On the positive side this stuff is confirmed to work in the debugger:
1. Breakpoints in plain python files
2. Breakpoints in code executed via `exec`
3. Breakpoints in code called in Deephaven `table.update`

The easiest way to reproduce the problem is to check out https://github.com/chipkent/test-dh-docker-vscode and follow the directions to use devcontainers in the README.  Once the devcontainer is started, add breakpoints in the code, run `run_me.py` under the debugger, and then execute code in the IDE that should trigger the breakpoints.  Using this VS Code path to reproduce will be far less painful than using PyCharm, because of #2994.

See #2994

jmao-denver:
@chipkent A while back, Colin and I were facing the same issue and found this solution. Could you confirm that it works for you?
```
import pydevd
pydevd.settrace(suspend=False, trace_only_current_thread=False)
```

As you can see, you need to install pydevd in your virtual env first. Both PyCharm and VSCode reply on this package in their debuggers. We even talked about including it in the DH server and  enable it by default to make the user experience better but decided to wait for actual user request first.

chipkent:
VS Code uses `debugpy` which, under the covers, vendors `pydevd`.  Using the above `pydevd` commands did not yield working breakpoints.

Some digging turned up this issue (https://github.com/microsoft/debugpy/issues/474), which suggests trying `debugpy.debug_this_thread()` to enable debugging in threads created in non-traditional ways (e.g. without using `threading`).

Riffing on this idea yielded some fruit.  If I run:
```python
x = mypkg.get_et()
```
in the console, no breakpoints are triggered.  But if I run:
```python
import debugpy; debugpy.debug_this_thread(); x = mypkg.get_et()
```
in the console, the relevant breakpoints are triggered.  

Unfortunately, every execution from the console requires `debugpy.debug_this_thread()` for the console-initiated breakpoints to trigger.  This potentially suggests:
1. The DH console is creating threads in a non-traditional way.
2. Every execution in the DH console is happening in a fresh thread.
[...]

PR #3075, opened 2022-11-07
Python debugging support

niloc132:
This commit adds a configuration option to customize engine-created
threads, and provides a default implementation that will register those
threads for debugging with pydevd if python is enabled. 

As of this commit, pydevd debugging seems to work correctly, but
VSCode's debugging doesn't work for all threads yet.

Partial #2997


Util/src/main/java/io/deephaven/util/thread/ThreadInitializationFactory.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Extension point to allow threads that will run user code from within the platform to be controlled by configuration.
+ */
+public interface ThreadInitializationFactory {
+    /* private */ String[] CONFIGURED_INITIALIZATION_TYPES =
+            Configuration.getInstance().getStringArrayFromProperty("thread.initialization");
```
devinrsmith, line 15: Does this work if the property doesn't exist or is empty?

py/server/deephaven/__init__.py
```diff
@@ -23,3 +23,33 @@
 from .stream.kafka import producer as kafka_producer
 from .table_factory import empty_table, time_table, merge, merge_sorted, new_table, DynamicTableWriter
 from .replay import TableReplayer
+import threading
+
+def create_thread_entry(thread_name):
```
devinrsmith, line 28: I think having this live in its own module makes sense. It's ok in deephaven, but maybe `deephaven_internal` is more appropriate?

server/src/main/java/io/deephaven/server/console/python/DebuggingInitializer.java
```diff
@@ -0,0 +1,45 @@
...
+        DeephavenModule py_deephaven = (DeephavenModule) PyModule.importModule("deephaven")
+                .createProxy(PyLib.CallableKind.FUNCTION, DeephavenModule.class);
+
+
+        return () -> {
+            // First call in to create a custom function that has the same name as the Java thread (plus a prefix)
+            PyObject runnableResult = py_deephaven.create_thread_entry(Thread.currentThread().getName());
+            // runnable.run();
+            // Invoke that function directly from Java, so that we have only this one initial frame
+            runnableResult.call("__call__", runnable);
```
devinrsmith, line 31: I think we could do this with a single method. The naming would probably change:

`py_deephaven.run_for_java_thread(Thread.currentThread().getName(), runnable)` or similar

devinrsmith, line 31: This technique of ensuring the java frame includes python in the Runnable stack _does_ rely on the fact that python drops the GIL when calling into java. https://github.com/jpy-consortium/jpy/pull/48 nothing actionable, but good to note it wouldn't work otherwise.

Util/src/main/java/io/deephaven/util/thread/ThreadInitializationFactory.java
```diff
@@ -0,0 +1,44 @@
...
+    /* private */ List<ThreadInitializationFactory> INITIALIZERS = Arrays.stream(CONFIGURED_INITIALIZATION_TYPES)
```
devinrsmith, line 16: I'm wondering, do we think there are occasions for more than 1 ThreadInitializationFactory?

Util/src/main/java/io/deephaven/util/thread/ThreadInitializationFactory.java
niloc132: It fails if it does not exist, but succeeds if empty (with an empty array).

Due to how defaults work (or rather, don't work), I was under the impression we wanted to generally avoid using them, so dh-defaults.prop now contains this property.

server/src/main/java/io/deephaven/server/console/python/DebuggingInitializer.java
niloc132: Potentially, and that's how the first implementation worked. However, if we wanted the top-most frame to have a specific name (so that you can tell what Java thread something was running on by looking at the py stack trace, which doesn't come with thread info), we would need to create a custom def to return in the call - the debugger at least loses the `__name__` and `__qualname__` properties that are assigned to rename defs. As such, we may need to return a pyobject so that it can be called again. 

And, if we just use the runnable.run() above (presently commented out), then we don't need to pass runnable at all, so the signature stays the same. In that case however it appears to the debugger that any pydevd.settrace() call in the web IDE actually pauses in the `create_change_list` function.

py/server/deephaven/__init__.py
niloc132: Yep, reasonable, this is about 70% straw man while we work out the specifics of what we even want it to do.

server/src/main/java/io/deephaven/server/console/python/DebuggingInitializer.java
niloc132: True, but the fact that DHaaL works at all relies on that fact as well, the entire library would deadlock on the gil and ugp if we didn't make this assumption.

Util/src/main/java/io/deephaven/util/thread/ThreadInitializationFactory.java
niloc132: I have assumed for now that each factory should be created once, but it if wants per-thread state, it should do so by putting that state in the `createInitializer` method itself, or in the returned Runnable. By only instantiating these once, we grant control to each implementation what kind of scope it wants to have. 

For example, the current `DebuggingInitializer` never closes the deephaven module (in its current form, it should), but could be modified to instead avoid paying the cost of re-creating that proxy for each thread (not just UGP threads, but also the DeephavenApiServerModule scheduler threads too).

server/src/main/java/io/deephaven/server/console/python/DebuggingInitializer.java
devinrsmith: Would this provide the same benefits?

```python
def run_for_java_thread(name, runnable):
    ...
    def JavaThread():
        runnable.run()
    JavaThread()
```

niloc132: * Changing the function's name (via declaring `JavaThread` not by name, but through eval) so the thread appears: technically yes, but by adding two frames instead of one, so the topmost frame isn't named in this way.
* Adding one py frame earlier in the stack than the current code: technically yes, but again, because it adds two frames instead of one, we might be making it harder to understand instead of easier

Remember, these run once per thread, when the thread is created, so on the order of 10ish times on startup and once when the first python command is executed from the web console, so paying the extra jni/py round trip is pretty cheap.

Util/src/main/java/io/deephaven/util/thread/ThreadInitializationFactory.java
```diff
@@ -0,0 +1,44 @@
...
+     * Chains configured initializers to run before/around any given runnable, returning a runnable intended to be run
+     * by a new thread.
+     */
+    static Runnable wrapRunnable(Runnable runnable) {
+        Runnable acc = runnable;
+        for (ThreadInitializationFactory INITIALIZER : INITIALIZERS) {
+            acc = INITIALIZER.createInitializer(acc);
+        }
+        return acc;
+    }
```
rcaudy, line 41: Just to be sure I understand, the reason for this pattern is to allow for initializer-specific cleanup on exit? Otherwise, a pattern of iterative runnable invocations would seem less error-prone. I suppose either version allows arbitrary hijacking of the intended run method; the current pattern could skip calling the wrapped runnable, but the iterative approach could just never return or throw an exception on termination.

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
```diff
@@ -1810,7 +1807,7 @@ private UpdateGraphProcessorThreadFactory(@NotNull final ThreadGroup threadGroup
         public Thread newThread(@NotNull final Runnable r) {
             return super.newThread(() -> {
                 configureRefreshThread();
-                r.run();
+                ThreadInitializationFactory.wrapRunnable(r).run();
```
rcaudy, line 1810: This pattern is inconsistent with the other usages I've seen, and I think I prefer the other one. That is, I think this should look like:
```
return super.newThread(ThreadInitializationFactory.wrapRunnable(() -> {
                configureRefreshThread();
                r.run();
```

server/src/main/java/io/deephaven/server/runner/DeephavenApiServerModule.java
```diff
@@ -154,14 +149,14 @@ public static UpdateGraphProcessor provideUpdateGraphProcessor() {
...
-            super(DeephavenApiServer.class, name, true);
+            super(DeephavenApiServer.class, name);
         }
 
         @Override
         public Thread newThread(final @NotNull Runnable r) {
             return super.newThread(() -> {
                 MultiChunkPool.enableDedicatedPoolForThisThread();
-                r.run();
+                ThreadInitializationFactory.wrapRunnable(r).run();
```
rcaudy, line 159: Another case where we've opted into a different pattern for order of initializers. I think I like the one you use more.

server/src/main/java/io/deephaven/server/console/python/DebuggingInitializer.java
```diff
@@ -0,0 +1,45 @@
...
+ * will be able to be debugged. If python is disabled, this does nothing.
+ */
+public class DebuggingInitializer implements ThreadInitializationFactory {
+    @Override
+    public Runnable createInitializer(Runnable runnable) {
+        if (!"python".equals(Configuration.getInstance().getStringWithDefault("deephaven.console.type", null))) {
+            // python not enabled, don't accidentally start it
+            return runnable;
+        }
+        DeephavenModule py_deephaven = (DeephavenModule) PyModule.importModule("deephaven")
```
rcaudy, line 22: Note from @rcaudy : I haven't reviewed the bottom of this file past this point, or `py/server/deephaven/__init__.py` yet.

Util/src/main/java/io/deephaven/util/thread/ThreadInitializationFactory.java
niloc132: It isn't that we need to do cleanup on the way out per se, but that we want to insert a stack frame in python. If we end up dropping this frame (which has some weird side effects when you "step out" of the apparent top-level frame), then we could just run one after the other, no wrapping.

The current impl has its option of how it wants to do it - I played with an API that would offer a DSL for either wrapping or prefixing, but didn't really love it, I can try bringing that back if you'd prefer.

niloc132: Okay, in light of the arm64 issues, a wrapper is necessary. The question then will be how we name that, if JavaThread is clear and simple enough, or if we want to use it to get some more info to the user.

devinrsmith (approved):

rcaudy (approved):
Approving for Python server codeowners, no interface code here.

PR #3079, opened 2022-11-14
Authorization: generated interfaces and verified by interceptors

nbauernfeind:
This is the replacement PR for #2931.


authorization/src/main/java/io/deephaven/auth/codegen/impl/ApplicationServiceAuthWiring.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    class DenyAll implements ApplicationServiceAuthWiring {
+
+        public void checkPermissionListFields(
+                AuthContext authContext,
+                io.deephaven.proto.backplane.grpc.ListFieldsRequest request) {
+            throw Exceptions.statusRuntimeException(Code.PERMISSION_DENIED, "Operation not allowed.");
+        }
+    }
```
devinrsmith, line 36: Do we imagine implementations might want to extend from any of the AllowAll or DenyAll classes? While it might be "convenient", I think we should force implementations to always implement each method themselves (forces a compile error on new releases instead of opting into potentially a bad choice).

On a related note, these could all be implemented as enum singletons.

authorization/src/main/java/io/deephaven/auth/codegen/impl/ApplicationServiceAuthWiring.java
```diff
@@ -0,0 +1,37 @@
...
+                AuthContext authContext,
+                io.deephaven.proto.backplane.grpc.ListFieldsRequest request) {}
+    }
+
+    class DenyAll implements ApplicationServiceAuthWiring {
+
+        public void checkPermissionListFields(
+                AuthContext authContext,
+                io.deephaven.proto.backplane.grpc.ListFieldsRequest request) {
+            throw Exceptions.statusRuntimeException(Code.PERMISSION_DENIED, "Operation not allowed.");
```
devinrsmith, line 34: Does it make sense to extract this as a helper method in `ServiceAuthWiring`?

engine/context/src/main/java/io/deephaven/engine/context/ExecutionContext.java
```diff
@@ -164,14 +182,29 @@ public QueryCompiler getQueryCompiler() {
...
 
+        private final AuthContext authContext;
+
         private QueryLibrary queryLibrary = PoisonedQueryLibrary.INSTANCE;
         private QueryScope queryScope = PoisonedQueryScope.INSTANCE;
         private QueryCompiler queryCompiler = PoisonedQueryCompiler.INSTANCE;
 
+        private Builder() {
+            // propagate the auth context from the current context
+            this(getContext().authContext);
```
devinrsmith, line 201: This is a bit of a departure from how the other parts are "poisened". Do we think that callers should be forced to do something like `captureAuthContext()` instead?

server/src/main/java/io/deephaven/server/auth/CommunityAuthenticationModule.java
```diff
@@ -0,0 +1,124 @@
...
+    @Singleton
+    public InputTableServiceContextualAuthWiring bindInputTableServiceContextualAuthWiring() {
+        return new InputTableServiceContextualAuthWiring.AllowAll();
+    }
+
+    @Provides
+    @Singleton
+    public TicketResolverBase.AuthTableTransformation bindAuthTableTransformation() {
+        // in community this is an identity transformation
+        return (table) -> table;
```
devinrsmith, line 122: I would create an Identity enum in AuthTableTransformation and return it here.

server/src/main/java/io/deephaven/server/auth/CommunityAuthenticationModule.java
```diff
@@ -0,0 +1,124 @@
...
+    @Singleton
+    public TableServiceContextualAuthWiring bindTableServiceContextualAuthWiring() {
+        return new TableServiceContextualAuthWiring.AllowAll();
+    }
+
+    @Provides
+    @Singleton
+    public InputTableServiceContextualAuthWiring bindInputTableServiceContextualAuthWiring() {
+        return new InputTableServiceContextualAuthWiring.AllowAll();
+    }
```
devinrsmith, line 116: Why are these services not making it into the map?

server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
```diff
@@ -290,24 +316,138 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(final ServerCall<Re
...
+                                    } else if (status.equals(Status.Code.PERMISSION_DENIED)) {
+                                        log.info().append(serviceName).append(".").append(methodName)
+                                                .append(": permission denied ").endl();
+                                    } else {
+                                        log.error().append(cause).endl();
+                                    }
+
+                                    call.close(sre.getStatus(), sre.getTrailers());
+                                } else {
+                                    log.error().append("Failed to invoke auth method: ").append(e).endl();
```
devinrsmith, line 432: I know we have places that take different approaches, but I think we might want to encourage `log.error(e).append("Failed to invoke auth method").endl()` approach instead. (Probably deserves an audit of our current usages and an effort to consolidate our approaches.)

server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
```diff
@@ -290,24 +316,138 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(final ServerCall<Re
...
+                                log.info().append(serviceName).append(".").append(methodName)
+                                        .append(": request unauthenticated").endl();
+                            } else if (status.equals(Status.Code.PERMISSION_DENIED)) {
+                                log.info().append(serviceName).append(".").append(methodName)
+                                        .append(": permission denied ").endl();
+                            } else {
+                                log.error().append(cause).endl();
+                            }
+
+                            call.close(sre.getStatus(), sre.getTrailers());
```
devinrsmith, line 398: Do the same comments you have in the `onFailure` part about catching exceptions for close also make sense here?

server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
```diff
@@ -290,24 +316,138 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(final ServerCall<Re
...
+
+                                return;
+                            }
+                        }
+
+                        // if we're still here, then it is safe to continue with the request
+                        delegate().onMessage(message);
+                    }
+                };
+            };
```
devinrsmith, line 444: There's a lot of logic here - I wonder if it makes sense to break these up into classes instead of anonymous.

server/src/main/java/io/deephaven/server/table/ops/GrpcTableOperation.java
```diff
@@ -13,66 +14,81 @@
...
     protected interface MultiDependencyFunction<T> {
-        List<TableReference> apply(T request);
+        List<TableReference> getTableReferences(T request);
+    }
+    @FunctionalInterface
+    protected interface PermissionFunction<T> {
+        void check(AuthContext authContext, T request, List<Table> sourceTables);
     }
 
+    final PermissionFunction<T> permission;
```
devinrsmith, line 29: I notice that the ops are referencing `permission` directly - that's not how the other member variables are used here (I would look into making them all private, and expose protected method for callers can call into to match other usage patterns). The nice thing is that you could have `protected void check(T request, List<SessionState.ExportObject<Table>> sourceTables)` and the caller wouldn't need to be responsible for providing the auth context or doing the `toTables`.

server/src/main/java/io/deephaven/server/table/ops/GrpcTableOperation.java
```diff
@@ -111,11 +128,21 @@ public Ticket getResultTicket(final T request) {
...
+        return getDependencies.getTableReferences(request);
+    }
+
+    /**
+     * Get the source tables for this operation.
+     *
+     * @param sourceTables the export object wrapped tables which are guaranteed to be satisfied
+     * @return the unwrapped source tables for this operation
+     */
+    protected List<Table> toTables(final List<SessionState.ExportObject<Table>> sourceTables) {
```
devinrsmith, line 145: Can be static?

server/src/main/java/io/deephaven/server/table/ops/GrpcTableOperation.java
devinrsmith: The other approach would be to remove `PermissionFunction` entirely, and expose the member variable `TableServiceContextualAuthWiring` and the callers could call `auth.myAppropriateMethod(...)` directly (I'm not sure how useful abstracting PermissionFunction is...).

authorization/build.gradle
```diff
@@ -0,0 +1,23 @@
+plugins {
+    id 'java-library'
+    id 'io.deephaven.project.register'
+
+    // TODO: fix this hack and invoke the protoc during the build via docker; maybe convert to a replicate task
+    id 'com.github.johnrengelman.shadow' version '7.1.2'
```
devinrsmith, line 6: Why are you needing this for now?

devinrsmith (commented):
First pass on this new PR.

Structurally, looks good.

I'm hoping we can have some examples / tests that exercise these various code paths.

Also, hoping for a layer that can deny arbitrary code execution.

authorization/build.gradle
niloc132: Possible this is getting picked up from some prelim work I did on adding support for a variety of authentication handlers, which lived in a shared branch we used.

authorization/src/main/java/io/deephaven/auth/codegen/GenerateContextualAuthWiring.java
```diff
@@ -0,0 +1,111 @@
+package io.deephaven.auth.codegen;
```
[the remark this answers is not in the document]
niloc132: I personally would prefer if we kept it as a non-annotation processor but instead as a code generator that adheres to the protoc plugin API like this. It would technically work as an annotation processor, but we'd likely have to reflectively attempt to access the already-generated java types that came from protoc, so we'd have to do it later.

That said, I'd prefer if this lived in its own gradle module, as we'll eventually cook up the CI bits to make this happen in the same docker container that all the rest of the proto stuff goes on in.

authorization/src/main/java/io/deephaven/auth/codegen/GenerateContextualAuthWiring.java
```diff
@@ -0,0 +1,111 @@
...
+            final DescriptorProtos.ServiceDescriptorProto service) {
+        final String serviceName = service.getName() + "ContextualAuthWiring";
+        System.err.println("Generating: " + serviceName);
+
+        final StringBuilder sb = new StringBuilder();
+        sb.append("/**\n");
+        sb.append(" * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending\n");
+        sb.append(" */\n");
+        sb.append("/**\n");
+        sb.append(
```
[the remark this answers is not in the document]
niloc132: +1 for javapoet - you never have to write imports, it does it for you, just be passing class/element/mirror instances to the code printer

authorization/src/main/java/io/deephaven/auth/codegen/impl/FlightServiceAuthWiring.java
```diff
@@ -0,0 +1,140 @@
...
+            AuthContext authContext,
+            org.apache.arrow.flight.impl.Flight.FlightDescriptor request);
+
+    void checkPermissionGetSchema(
+            AuthContext authContext,
+            org.apache.arrow.flight.impl.Flight.FlightDescriptor request);
+
+    void checkPermissionDoGet(
+            AuthContext authContext,
+            org.apache.arrow.flight.impl.Flight.Ticket request);
```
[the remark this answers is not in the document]
niloc132: I don't think that's true - they would be TableReferences, which for one-off calls will always be Tickets, and for batches might be offsets.

I could imagine a way to request a TicketResolver - that plus knowledge of the Session is sufficient to transform either a ticket or reference into an actual object (which may or may not be a table).

authorization/build.gradle
nbauernfeind: I did this to generate a single jar that I could easily invoke from a shell script. To invoke protoc you need flags like this:
```
PATH=.:$PATH protoc --service-auth-wiring_out=authorization/src/main/java/ -I proto/proto-backplane-grpc/src/main/proto/  proto/proto-backplane-grpc/src/main/proto/deephaven/proto/console.proto
```

Which will automatically look for an executable called `protoc-gen-service-auth-wiring` on the path. To ease my pain, these were shell scripts that used the jar-jar. Protoc expects to communicate in stdin vs stdout and I couldn't figure out how to do that via gradle -- so this was my cheap work around given that Colin is going to direct me how to automate the generation.

authorization/src/main/java/io/deephaven/auth/codegen/impl/FlightServiceAuthWiring.java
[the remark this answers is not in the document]
nbauernfeind: The RPC hooks that we have are protobuf only. You get TableReferences. This is why the Table service methods include the List<Table> -- these are post resolution of the TableReferences and cannot be invoked at the time the interceptor runs.

ACLs are applied when a TableReference is resolved either as a source table or to be exported into the session. The table is guaranteed to have ACLs applied by the time DoGet/DoExchange are invoked. Consider how useful it is to have access to the table. If the user can transform the table, could they transform it in a way that allows it to be seen via DoGet/DoExchange?

server/src/main/java/io/deephaven/server/auth/CommunityAuthenticationModule.java
```diff
@@ -0,0 +1,124 @@
...
+    String value();
+}
+
+
+@Module
+public class CommunityAuthenticationModule {
+    @Provides
+    @IntoMap
+    @ServiceGrpcKey(ApplicationServiceGrpc.SERVICE_NAME)
+    public ServiceAuthWiring bindApplicationServiceAuthWiring() {
```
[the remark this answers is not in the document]
nbauernfeind: The downside of that approach is that you need to bind both the ServiceAuthWiring to the Map and the singleton instance of the AllowAll. It felt like a very similar number of lines-of-code and this implementation made it easy for me to test the ranges of functionality by overriding one method at a time.

authorization/src/main/java/io/deephaven/auth/codegen/impl/ApplicationServiceAuthWiring.java
nbauernfeind: Hmm. I would rather place a disclaimer and offer recommended practices rather than force the user to copy paste. An extension of the DenyAll actually feels rather appropriate for the scenario where you allow a smaller subset of functionality.

I'll discuss with Ryan.

server/src/main/java/io/deephaven/server/auth/CommunityAuthenticationModule.java
nbauernfeind: The Map is for at-rpc-invocation time authorization. The InputTable and the Table services require that the TableReferences be resolved into Tables. The resolution isn't guaranteed until the session runs the export's runnable. The runnable isn't registered until after the rpc has been invoked. So, we can't perform this authorization through the rpc interceptor before the tables have even had a chance to resolve (also, consider complexities regarding batch requests). Instead, these types are deeply injected (i.e. not automatically generated) where the operation has the resolved TableReferences and the original rpc request.

authorization/build.gradle
niloc132: Probably could use the application plugin to achieve this too, it will give you a zip/tarball you can use as input into the docker container, and a script inside there which can be executed.

server/src/main/java/io/deephaven/server/auth/CommunityAuthenticationModule.java
niloc132: or use Function.identity()?

authorization/src/main/java/io/deephaven/auth/codegen/impl/TableServiceContextualAuthWiring.java
```diff
@@ -0,0 +1,603 @@
...
+import io.deephaven.proto.util.Exceptions;
+
+import java.util.List;
+
+public interface TableServiceContextualAuthWiring {
+
+    void checkPermissionGetExportedTableCreationResponse(
+            AuthContext authContext,
+            io.deephaven.proto.backplane.grpc.Ticket request,
+            List<Table> tables);
```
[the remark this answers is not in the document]
nbauernfeind: I didn't write anything to determine how many source tables exist in a request. This List<Table> mirrors the List<TableReference> that exists between the TableService and the OperationImpl. Picking out the number of source tables felt error prone and I could avoid that class of bugs.

engine/context/src/main/java/io/deephaven/engine/context/ExecutionContext.java
```diff
@@ -112,7 +115,22 @@ public ExecutionContext withSystemic(boolean isSystemic) {
...
+
+    /**
+     * Returns, or creates, an execution context with the given value for {@code authContext} and existing values for
+     * the other members. This is not intended to be used by user code.
+     *
+     * @param authContext the auth context to use instead
+     * @return the execution context
+     */
+    public ExecutionContext withAuthContext(final AuthContext authContext) {
+        // TODO: should the result be a composition of existing auth context and the provided?
```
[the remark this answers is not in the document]
nbauernfeind: Should we so easily allow the script user to set an auth context of their preference? It is starting to feel appropriate to add a `AuthContext AuthContext#compose(AuthContext other)` where a sub-scope belonging to a user can only further limit the system.

engine/context/src/main/java/io/deephaven/engine/context/ExecutionContext.java
nbauernfeind: How much cooperation is appropriate to ask of the user(s)? I feel that without being able to offer an alternative (as that feels like a security faux pas) that passing along the current value feels friendly and helpful.

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -80,6 +80,8 @@ public ConsoleServiceGrpcImpl(final TicketRouter ticketRouter,
     public void getConsoleTypes(final GetConsoleTypesRequest request,
             final StreamObserver<GetConsoleTypesResponse> responseObserver) {
         GrpcUtil.rpcWrapper(log, responseObserver, () -> {
+            // for legacy reasons, this method is used before the session is created
+            // TODO: how do we authorize these calls; can we fix the ordering of requests?
             if (!REMOTE_CONSOLE_DISABLED) {
```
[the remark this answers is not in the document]
nbauernfeind: Hmm. I dropped the todo here because the web ui requests console types before it creates the session. IIRC the console selection from the user was originally required to open the jsapi connection.

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -219,15 +210,18 @@ private static FieldInfo makeVariableDefinition(String title, String type) {
 
     @Override
     public void cancelCommand(CancelCommandRequest request, StreamObserver<CancelCommandResponse> responseObserver) {
-        // TODO not yet implemented, need a way to handle stopping a command in a consistent way
-        super.cancelCommand(request, responseObserver);
+        GrpcUtil.rpcWrapper(log, responseObserver, () -> {
+            // TODO (#53): consider task cancellation
+            responseObserver.onError(GrpcUtil.statusRuntimeException(Code.UNIMPLEMENTED, "Not yet implemented"));
```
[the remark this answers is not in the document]
nbauernfeind: I will revert, but maybe not for the same reason you think =). The existing super call throws this exception, too:
```java
  public static void asyncUnimplementedUnaryCall(
      MethodDescriptor<?, ?> methodDescriptor, StreamObserver<?> responseObserver) {
    checkNotNull(methodDescriptor, "methodDescriptor");
    checkNotNull(responseObserver, "responseObserver");
    responseObserver.onError(Status.UNIMPLEMENTED
        .withDescription(String.format("Method %s is unimplemented",
            methodDescriptor.getFullMethodName()))
        .asRuntimeException());
  }
```

server/src/main/java/io/deephaven/server/console/ScopeTicketResolver.java
```diff
@@ -110,6 +113,11 @@ private <T> SessionState.ExportObject<T> resolve(
             return scopeVar;
         });
 
+        if (export instanceof Table) {
+            // noinspection unchecked
+            export = (T) authTableTransformation.transform((Table) export);
```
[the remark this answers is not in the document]
nbauernfeind: Ok. We can do Object -> Object transformation and require that implementations perform type checking.

authorization-codegen/build.gradle
```diff
@@ -0,0 +1,18 @@
...
+description 'Deephaven Authorization Code Generator'
+
+dependencies {
+     api project(':engine-api')
+
+     Classpaths.inheritGrpcPlatform(project, 'api')
+     api 'io.grpc:grpc-services'
+
+    implementation 'com.squareup:javapoet:1.9.0'
+}
```
devinrsmith, line 18: I see there aren't any tasks that generate the files automatically. I think that's _fine_; it's a manually process right now?

I think we have three options:

1. Leave as-is, need to know to manually code-gen as necessary.
2. Add gradle tasks that can check, and then run to regenerate.
3. Add test that can generate and compare.

Either way, I think we need at least a comment here or quick README file here that explains process.

buildSrc/src/main/groovy/Classpaths.groovy
```diff
@@ -96,6 +96,7 @@ class Classpaths {
 
     static final String GRPC_GROUP = 'io.grpc'
     static final String GRPC_NAME = 'grpc-bom'
+    static final String GRPC_SERVICES_NAME = 'grpc-services'
```
devinrsmith, line 99: unused

authorization/src/main/java/io/deephaven/auth/codegen/impl/ApplicationServiceAuthWiring.java
```diff
@@ -0,0 +1,26 @@
...
+// ------------------------------------------------------------------
+//
+package io.deephaven.auth.codegen.impl;
+
+import io.deephaven.auth.AuthContext;
+import io.deephaven.auth.ServiceAuthWiring;
+import io.deephaven.proto.backplane.grpc.ListFieldsRequest;
+
+public interface ApplicationServiceAuthWiring extends ServiceAuthWiring {
+    void checkPermissionListFields(AuthContext authContext, ListFieldsRequest request);
```
devinrsmith, line 15: Nothing calls this.

authorization/src/main/java/io/deephaven/auth/codegen/impl/ConfigServiceAuthWiring.java
```diff
@@ -0,0 +1,41 @@
...
+import io.deephaven.auth.ServiceAuthWiring;
+import io.deephaven.proto.backplane.grpc.AuthenticationConstantsRequest;
+import io.deephaven.proto.backplane.grpc.ConfigurationConstantsRequest;
+
+public interface ConfigServiceAuthWiring extends ServiceAuthWiring {
+    void checkPermissionGetAuthenticationConstants(AuthContext authContext,
+            AuthenticationConstantsRequest request);
+
+    void checkPermissionGetConfigurationConstants(AuthContext authContext,
+            ConfigurationConstantsRequest request);
```
devinrsmith, line 20: Nothing calls these.

authorization/src/main/java/io/deephaven/auth/codegen/impl/ConsoleServiceAuthWiring.java
```diff
@@ -0,0 +1,120 @@
...
+import io.deephaven.proto.backplane.script.grpc.AutoCompleteRequest;
+import io.deephaven.proto.backplane.script.grpc.BindTableToVariableRequest;
+import io.deephaven.proto.backplane.script.grpc.CancelCommandRequest;
+import io.deephaven.proto.backplane.script.grpc.ExecuteCommandRequest;
+import io.deephaven.proto.backplane.script.grpc.GetConsoleTypesRequest;
+import io.deephaven.proto.backplane.script.grpc.GetHeapInfoRequest;
+import io.deephaven.proto.backplane.script.grpc.LogSubscriptionRequest;
+import io.deephaven.proto.backplane.script.grpc.StartConsoleRequest;
+
+public interface ConsoleServiceAuthWiring extends ServiceAuthWiring {
```
devinrsmith, line 21: Nothing calls the methods here.

authorization/src/main/java/io/deephaven/auth/codegen/impl/FlightServiceAuthWiring.java
```diff
@@ -0,0 +1,92 @@
...
+// This class is generated by GenerateServiceAuthWiring. DO NOT EDIT!
+// ------------------------------------------------------------------
+//
+package io.deephaven.auth.codegen.impl;
+
+import io.deephaven.auth.AuthContext;
+import io.deephaven.auth.ServiceAuthWiring;
+import org.apache.arrow.flight.impl.Flight;
+
+public interface FlightServiceAuthWiring extends ServiceAuthWiring {
```
devinrsmith, line 14: Nothing calls the methods here.

authorization/src/main/java/io/deephaven/auth/codegen/impl/HealthAuthWiring.java
```diff
@@ -0,0 +1,34 @@
...
+// This class is generated by GenerateServiceAuthWiring. DO NOT EDIT!
+// ------------------------------------------------------------------
+//
+package io.deephaven.auth.codegen.impl;
+
+import io.deephaven.auth.AuthContext;
+import io.deephaven.auth.ServiceAuthWiring;
+import io.grpc.health.v1.HealthCheckRequest;
+
+public interface HealthAuthWiring extends ServiceAuthWiring {
```
devinrsmith, line 14: Nothing calls methods here.

authorization/src/main/java/io/deephaven/auth/codegen/impl/ObjectServiceAuthWiring.java
```diff
@@ -0,0 +1,26 @@
...
+// This class is generated by GenerateServiceAuthWiring. DO NOT EDIT!
+// ------------------------------------------------------------------
+//
+package io.deephaven.auth.codegen.impl;
+
+import io.deephaven.auth.AuthContext;
+import io.deephaven.auth.ServiceAuthWiring;
+import io.deephaven.proto.backplane.grpc.FetchObjectRequest;
+
+public interface ObjectServiceAuthWiring extends ServiceAuthWiring {
```
devinrsmith, line 14: Nothing calls methods here.

authorization/src/main/java/io/deephaven/auth/codegen/impl/SessionServiceAuthWiring.java
```diff
@@ -0,0 +1,86 @@
...
+
+import io.deephaven.auth.AuthContext;
+import io.deephaven.auth.ServiceAuthWiring;
+import io.deephaven.proto.backplane.grpc.ExportNotificationRequest;
+import io.deephaven.proto.backplane.grpc.ExportRequest;
+import io.deephaven.proto.backplane.grpc.HandshakeRequest;
+import io.deephaven.proto.backplane.grpc.ReleaseRequest;
+import io.deephaven.proto.backplane.grpc.TerminationNotificationRequest;
+
+public interface SessionServiceAuthWiring extends ServiceAuthWiring {
```
devinrsmith, line 18: Nothing calls methods here.

authorization/src/main/java/io/deephaven/auth/codegen/impl/StorageServiceAuthWiring.java
```diff
@@ -0,0 +1,73 @@
...
+import io.deephaven.auth.AuthContext;
+import io.deephaven.auth.ServiceAuthWiring;
+import io.deephaven.proto.backplane.grpc.CreateDirectoryRequest;
+import io.deephaven.proto.backplane.grpc.DeleteItemRequest;
+import io.deephaven.proto.backplane.grpc.FetchFileRequest;
+import io.deephaven.proto.backplane.grpc.ListItemsRequest;
+import io.deephaven.proto.backplane.grpc.MoveItemRequest;
+import io.deephaven.proto.backplane.grpc.SaveFileRequest;
+
+public interface StorageServiceAuthWiring extends ServiceAuthWiring {
```
devinrsmith, line 19: Nothing calls methods here.

authorization/src/main/java/io/deephaven/auth/codegen/impl/TableServiceContextualAuthWiring.java
```diff
@@ -0,0 +1,433 @@
...
+import io.deephaven.proto.backplane.grpc.Ticket;
+import io.deephaven.proto.backplane.grpc.TimeTableRequest;
+import io.deephaven.proto.backplane.grpc.UngroupRequest;
+import io.deephaven.proto.backplane.grpc.UnstructuredFilterTableRequest;
+import io.deephaven.proto.backplane.grpc.UpdateByRequest;
+import io.deephaven.proto.backplane.grpc.WhereInRequest;
+import java.util.List;
+
+public interface TableServiceContextualAuthWiring extends ServiceAuthWiring {
+    void checkPermissionGetExportedTableCreationResponse(AuthContext authContext, Ticket request,
```
devinrsmith, line 45: Nothing calls this.

authorization/src/main/java/io/deephaven/auth/codegen/impl/TableServiceContextualAuthWiring.java
```diff
@@ -0,0 +1,433 @@
...
+            List<Table> sourceTables);
+
+    void checkPermissionFetchTable(AuthContext authContext, FetchTableRequest request,
+            List<Table> sourceTables);
+
+    void checkPermissionFetchPandasTable(AuthContext authContext, FetchPandasTableRequest request,
```
devinrsmith, line 51: Nothing calls this.

authorization/src/main/java/io/deephaven/auth/codegen/impl/TableServiceContextualAuthWiring.java
```diff
@@ -0,0 +1,433 @@
...
+    void checkPermissionRunChartDownsample(AuthContext authContext, RunChartDownsampleRequest request,
+            List<Table> sourceTables);
+
+    void checkPermissionCreateInputTable(AuthContext authContext, CreateInputTableRequest request,
+            List<Table> sourceTables);
+
+    void checkPermissionWhereIn(AuthContext authContext, WhereInRequest request,
+            List<Table> sourceTables);
+
+    void checkPermissionExportedTableUpdates(AuthContext authContext,
```
devinrsmith, line 147: Nothing calls this.

server/src/main/java/io/deephaven/server/auth/CommunityAuthenticationModule.java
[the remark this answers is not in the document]
devinrsmith: I'm not sure I understand - I've changed all these to static methods (and the class to an interface, not necessary but maybe preferable), and everything works the same I think.

server/src/main/java/io/deephaven/server/console/SessionToExecutionStateModule.java
```diff
@@ -2,13 +2,15 @@
...
 import io.deephaven.engine.util.ScriptSession;
 
 @Module
 public class SessionToExecutionStateModule {
     @Provides
     ExecutionContext bindExecutionContext(ScriptSession session) {
-        return session.getExecutionContext();
+        // TODO: users should be able to provide an auth context that is used for start up scripts, etc
+        return session.getExecutionContext().withAuthContext(new AuthContext.SuperUser());
     }
```
devinrsmith, line 15: Haven't seen this before, but these can also be static method.

server/src/main/java/io/deephaven/server/partitionedtable/PartitionedTableServiceGrpcImpl.java
```diff
@@ -1,10 +1,11 @@
 package io.deephaven.server.partitionedtable;
 
+import com.google.common.collect.Lists;
```
devinrsmith, line 3: Do we need to use guava Lists?

server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
devinrsmith: I am still worried about the potential fragility here. I think we want a good way to unit test this logic (as opposed to integration tests), and I don't see a good way to do that as-is.

It would be great to be able to test the reflection in isolation.

server/src/main/java/io/deephaven/server/session/TicketResolverBase.java
```diff
@@ -3,11 +3,44 @@
...
+import io.deephaven.engine.table.Table;
+import io.deephaven.engine.table.PartitionedTable;
+
 public abstract class TicketResolverBase implements TicketResolver {
+
+    @FunctionalInterface
+    public interface AuthTableTransformation {
+        /**
+         * Implementations must type check the provided source as any type of object can be stored in an export.
+         *
```
devinrsmith, line 15: Need `<p>`

server/src/main/java/io/deephaven/server/table/inputtables/InputTableServiceGrpcImpl.java
```diff
@@ -3,7 +3,10 @@
  */
 package io.deephaven.server.table.inputtables;
 
+import com.google.common.collect.Lists;
```
devinrsmith, line 6: Do we need guava Lists?

server/src/main/java/io/deephaven/server/table/ops/ApplyPreviewColumnsGrpcImpl.java
```diff
@@ -17,14 +18,16 @@
...
     }
 
     @Override
-    public Table create(ApplyPreviewColumnsRequest request, List<SessionState.ExportObject<Table>> sourceTables) {
+    public Table create(final ApplyPreviewColumnsRequest request,
+            final List<SessionState.ExportObject<Table>> sourceTables) {
         Assert.eq(sourceTables.size(), "sourceTables.size()", 1);
+        checkPermission(request, sourceTables);
         return ColumnPreviewManager.applyPreview(sourceTables.get(0).get());
     }
```
devinrsmith, line 32: I know I mentioned abstracting checkPermission - much better now IMO.

Does it make sense to take it a step further? Have GrpcTableOperation implement create(), and introduce a new abstract method `createImpl` where the table concrete GrpcTableOperation implementation doesn't even have to call checkPermission themselves?

server/src/main/java/io/deephaven/server/table/ops/ApplyPreviewColumnsGrpcImpl.java
```diff
@@ -17,14 +18,16 @@
...
+        super(authWiring::checkPermissionApplyPreviewColumns, BatchTableRequest.Operation::getApplyPreviewColumns,
+                ApplyPreviewColumnsRequest::getResultId, ApplyPreviewColumnsRequest::getSourceId);
     }
 
     @Override
-    public Table create(ApplyPreviewColumnsRequest request, List<SessionState.ExportObject<Table>> sourceTables) {
+    public Table create(final ApplyPreviewColumnsRequest request,
+            final List<SessionState.ExportObject<Table>> sourceTables) {
         Assert.eq(sourceTables.size(), "sourceTables.size()", 1);
+        checkPermission(request, sourceTables);
```
devinrsmith, line 30: Thoughts on having checkPermission be the very first part of create() (generally)?

devinrsmith (commented):
Had a lot of comments about "none of these methods are called" - but I forgot about the reflective part. That said, I think there are some individual methods that aren't being called according to IntelliJ that I think we should probably see usages of.

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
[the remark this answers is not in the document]
niloc132: It is required to create the console session from the JS client, but they can be required to log in first (and soon, in order to use the JS API, will need to authenticate before they can call the method that will fetch this list).

authorization/build.gradle
```diff
@@ -0,0 +1,19 @@
+plugins {
+    id 'java-library'
+    id 'io.deephaven.project.register'
+}
+
+description 'Deephaven Authorization'
+
+dependencies {
+    api project(':Base')
+    api project(':authentication')
```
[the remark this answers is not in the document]
niloc132: (am 30% wondering if that is a joke comment)

authentication != authorization

authorization-codegen/build.gradle
```diff
@@ -0,0 +1,18 @@
+plugins {
+    id 'java-library'
```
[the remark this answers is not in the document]
niloc132: and probably would benefit from being `application` instead of shading? would build faster at least (probably)

authorization/build.gradle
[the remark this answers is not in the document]
devinrsmith: My bad, was going too fast. That said, the dependency is unused (compiles fine w/o it).

authorization-codegen/build.gradle
niloc132: i think this step is probably waiting on ~me, to help with changing how the docker container is built that runs protoc

server/src/main/java/io/deephaven/server/partitionedtable/PartitionedTableServiceGrpcImpl.java
```diff
@@ -108,6 +118,8 @@ public void getTable(GetTableRequest request, StreamObserver<ExportedTableCreati
                     .require(partitionedTable, keys)
                     .onError(responseObserver)
                     .submit(() -> {
+                        authWiring.checkPermissionGetTable(session.getAuthContext(), request,
+                                Lists.newArrayList((Table) partitionedTable.get(), keys.get()));
```
niloc132, line 122: if not mutable, consider the java9+ `List.of(e1, e2)`

server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
```diff
@@ -290,24 +314,144 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(final ServerCall<Re
...
-                        call.close(Status.UNAUTHENTICATED, new Metadata());
-                    } catch (IllegalStateException ignored) {
-                        // could be thrown if the call was already closed. As an interceptor, we can't throw,
-                        // so ignoring this and just returning the no-op listener.
+                    return onFailure.get();
+                }
+            }
+
+            // find the auth method
+            final String serviceName = call.getMethodDescriptor().getServiceName();
```
devinrsmith, line 328: I'm beginning to think we should have a singular generic entrypoint instead of the per-service way it's structured right now. `authorization.check(authContext, call.getMethodDescriptor(), message)`. If we want per-service classes, that could be an implementation detail IMO.

By allowing the hooking of a single entrypoint, instead of N, it's a much easier job to configure IMO.

server/src/main/java/io/deephaven/server/auth/CommunityAuthenticationModule.java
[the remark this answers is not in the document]
nbauernfeind: Thanks for being patient with me. This does indeed work and I thought you were referring to another pattern.

authorization-codegen/build.gradle
nbauernfeind: I've spawned #3133. I am adding to the readme how to regenerate the files.

authorization-codegen/build.gradle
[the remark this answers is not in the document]
nbauernfeind: I've spawned #3133 to address this in future work; for now the shadow jar will be part of my readme instructions.

authorization/src/main/java/io/deephaven/auth/codegen/impl/TableServiceContextualAuthWiring.java
nbauernfeind: We do not have this method implemented. It is also not part of Batch. @niloc132 should we remove this method from the API, or is it needed for future use?

server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
```diff
@@ -290,24 +314,144 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(final ServerCall<Re
...
+
+            // allow the TableServices to pass through as authorization requires source table context
+            boolean isTableService = serviceName != null && serviceName.contains("TableService");
+            if (authWiring == null && !isTableService) {
+                log.error().append("Could not find authorization wiring for service: ").append(serviceName)
+                        .append(" method: ").append(methodName).endl();
+                return onFailure.get();
+            }
+
+            final Provider<Method> authMethodLookup = () -> {
```
rcaudy, line 340: Why are we doing this with reflection instead of code generation?

rcaudy: That is, we could generate a "dispatcher" method.

server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
nbauernfeind: I've replaced this with service method overrides instead of injection.

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -80,6 +80,8 @@ public ConsoleServiceGrpcImpl(final TicketRouter ticketRouter,
     public void getConsoleTypes(final GetConsoleTypesRequest request,
             final StreamObserver<GetConsoleTypesResponse> responseObserver) {
         GrpcUtil.rpcWrapper(log, responseObserver, () -> {
+            // for legacy reasons, this method is used before the session is created
```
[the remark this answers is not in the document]
nbauernfeind: Filed this as #3147.

server/src/main/java/io/deephaven/server/table/ops/ApplyPreviewColumnsGrpcImpl.java
nbauernfeind: I lifted the checkPermission call to the callers of create. There are only two, and they are controlled by the batching / one-shot RPC code paths. I like this better than having a `create` and `createImpl`.

server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
nbauernfeind: We can probably still do this, but it looks like I'm going to skip this iteration in favor for the type safe interfaces for now since I was able to eliminate the reflection or even an additional layer of indirection.

authorization/src/main/java/io/deephaven/auth/ServiceAuthWiring.java
```diff
@@ -0,0 +1,194 @@
...
+    /**
+     * Close the call, but don't throw any exceptions.
+     */
+    private static <ReqT, RespT> void quietlyCloseCall(
+            final ServerCall<ReqT, RespT> call, final Status status, final Metadata trailers) {
+        try {
+            call.close(status, trailers);
+        } catch (IllegalStateException ignored) {
+            // could be thrown if the call was already closed. As an interceptor, we can't throw,
+            // so ignoring this and just returning the no-op listener.
```
niloc132, line 160: comment was copied, but not quite accurate, there is no no-op listener here (though both cases that call this return false, which right away will return a no-op listener)

authorization/src/main/java/io/deephaven/auth/ServiceAuthWiring.java
```diff
@@ -0,0 +1,194 @@
...
+            }
+
+            return true;
+        }
+    }
+
+    /**
+     * Close the call, but don't throw any exceptions. We were not given details about the error.
+     */
+    private static <ReqT, RespT> void quietlyCloseCall(final ServerCall<ReqT, RespT> call) {
```
niloc132, line 147: maybe inline this, since it is only called once, just to set two params?

authorization/src/main/java/io/deephaven/auth/codegen/impl/ConsoleServiceAuthWiring.java
```diff
@@ -0,0 +1,331 @@
...
+     * @throws io.grpc.StatusRuntimeException if the user is not authorized to invoke BindTableToVariable
+     */
+    void onMessageReceivedBindTableToVariable(AuthContext authContext,
+            BindTableToVariableRequest request);
+
+    /**
+     * Authorize a request to open a client-streaming rpc AutoCompleteStream.
+     *
+     * @param authContext the authentication context of the request
+     * @param request the request to authorize
```
niloc132, line 131: generated nit: this param doesnt exist

server/src/main/java/io/deephaven/server/table/TableServiceGrpcImpl.java
```diff
@@ -281,6 +286,13 @@ public void whereIn(WhereInRequest request, StreamObserver<ExportedTableCreation
         oneShotOperationWrapper(BatchTableRequest.Operation.OpCase.WHERE_IN, request, responseObserver);
     }
 
+    @Override
```
niloc132, line 289: pretty sure we should remove this rpc call entirely, as dataframes are fetched as widgets, and so will go through the TicketResolverBase.AuthTransformation mechanism when fetched as an object.

niloc132 (approved):
A few small nits, but i've marked it as approved, nothing we can't reasonably followup on.

server/src/main/java/io/deephaven/server/table/TableServiceGrpcImpl.java
nbauernfeind: filing: https://github.com/deephaven/deephaven-core/issues/3153

server/src/main/java/io/deephaven/server/table/TableModule.java
```diff
@@ -42,6 +45,11 @@
 
 @Module
 public interface TableModule {
+    @Provides
+    static TableServiceContextualAuthWiring provideAuthWiring(AuthorizationProvider authProvider) {
+        return authProvider.getTableServiceContextualAuthWiring();
+    }
```
devinrsmith, line 51: I notice we are a bit inconsistent IMO wrt which auth wirings we `@Provides` - looks like it's only the ones that aren't using the `AuthorizationWrappedGrpcBinding` service?

I wonder if it makes sense to have a single module that `@Provides` all the auth wirings from `AuthorizationProvider`? (And then have the impls that use AuthorizationWrappedGrpcBinding depend on the specific type instead of `AuthorizationProvider`.)

devinrsmith (changes requested):
The test dependency is the only blocker IMO.

engine/context/src/main/java/io/deephaven/engine/context/ExecutionContext.java
```diff
@@ -112,7 +104,21 @@ public ExecutionContext withSystemic(boolean isSystemic) {
...
+     * Returns, or creates, an execution context with the given value for {@code authContext} and existing values for
+     * the other members. This is not intended to be used by user code.
+     *
+     * @param authContext the auth context to use instead; null values are ignored
+     * @return the execution context
+     */
+    public ExecutionContext withAuthContext(final AuthContext authContext) {
+        if (authContext == null || authContext == this.authContext) {
+            return this;
+        }
```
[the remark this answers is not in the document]
nbauernfeind: The check intends to prevent the user from setting it null. It's kind of arbitrary. As long as DHE doesn't assume a null auth context is the same as the instance auth context. The user would only be able to change it while a script is being executed, or from a spawned thread.

server/src/main/java/io/deephaven/server/table/TableModule.java
nbauernfeind: AuthorizationWrappedGrpcBinding injects auth handlers that are invoked as gRPC delegates to the service implementation. The ContextualAuthWiring are injected more deeply into the service and do not verify authorization until Tickets have been resolved which doesn't even happen on a gRPC thread but rather the shared executor service. 

They are injected into different locations.

What would you prefer here? Is there an issue with a static `@Provides` method? In other examples like this you suggested making static methods which is why I did so here.

engine/test-utils/build.gradle
```diff
@@ -9,6 +9,7 @@ dependencies {
     api project(":engine-table")
     api project(":engine-rowset")
 
+    implementation TestTools.projectDependency(project, 'engine-context')
```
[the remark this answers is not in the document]
nbauernfeind: I duplicated the helper method instead of adding another jar.

server/jetty/src/main/java/io/deephaven/server/jetty/JettyServerComponent.java
```diff
@@ -38,6 +39,9 @@ interface Builder extends DeephavenApiServerComponent.Builder<Builder> {
         @BindsInstance
         Builder withJettyConfig(JettyConfig config);
 
+        @BindsInstance
+        Builder withAuthorizationProvider(AuthorizationProvider authorizationProvider);
```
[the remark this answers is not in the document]
devinrsmith: I think we need to make the dagger module less confusing, b/c I think it should "be the way". Happy to punt for now.

server/src/main/java/io/deephaven/server/table/TableModule.java
devinrsmith: We can follow-up later IMO.

devinrsmith (dismissed):
Same as Colin. I think we can follow-up post release.

niloc132 (approved):

devinrsmith (approved):

PR #3084, opened 2022-11-15
Wrap the input tables

### LINKED ISSUE #3034: Wrap the InputTable API in the DH Python API (server-side)
This is a feature most requested by the DevRel team.

jmao-denver:
Fixes #3034

### DIFF
--- a/py/server/deephaven/table_factory.py
+++ b/py/server/deephaven/table_factory.py
@@ -197,7 +197,7 @@ class DynamicTableWriter(JObjectWrapper):
 
 
 class InputTable(Table):
-    """InputTable is a subclass of Table that allows the users to dynamic add/delete/modify data in it. There are two
+    """InputTable is a subclass of Table that allows the users to dynamically add/delete/modify data in it. There are two
     types of InputTable - append-only and keyed.
 
     The append-only input table is not keyed, all rows are added to the end of the table, and deletions and edits are
@@ -238,7 +238,7 @@ class InputTable(Table):
             raise DHError(e, "failed to create a InputTable.") from e
 
     def add(self, table: Table):
-        """Writes rows in the provided table to this input table. If this is keyed input table, added rows with keys
+        """Writes rows from the provided table to this input table. If this is a keyed input table, added rows with keys
         that match existing rows will replace those rows.
 
         Args:
--- a/py/server/deephaven/table_factory.py
+++ b/py/server/deephaven/table_factory.py
@@ -222,8 +222,7 @@ class InputTable(Table):
                 raise ValueError("both column definitions and init table are provided.")
 
             if col_defs:
-                cols = [Column(name=n, data_type=t) for n, t in col_defs.items()]
-                j_arg_1 = _JTableDefinition.of([col.j_column_definition for col in cols])
+                j_arg_1 = _JTableDefinition.of([Column(name=n, data_type=t).j_column_definition for n, t in col_defs.items()])
             else:
                 j_arg_1 = init_table.j_table
 
@@ -237,7 +236,7 @@ class InputTable(Table):
         except Exception as e:
             raise DHError(e, "failed to create a InputTable.") from e
 
-    def add(self, table: Table):
+    def add(self, table: Table) -> None:
         """Writes rows from the provided table to this input table. If this is a keyed input table, added rows with keys
         that match existing rows will replace those rows.
 
@@ -252,8 +251,9 @@ class InputTable(Table):
         except Exception as e:
             raise DHError(e, "add to InputTable failed.") from e
 
-    def delete(self, table: Table):
-        """Deletes the keys contained in the provided table from this keyed input table.
+    def delete(self, table: Table) -> None:
+        """Deletes the keys contained in the provided table from this keyed input table. If this method is called on an
+        append-only input table, a PermissionError will be raised.
 
         Args:
             table (Table): the table with the keys to delete
--- a/py/server/deephaven/table_factory.py
+++ b/py/server/deephaven/table_factory.py
@@ -3,7 +3,7 @@
 #
 
 """ This module provides various ways to make a Deephaven table. """
-from __future__ import annotations
+
 from typing import List, Dict, Any, Union, Sequence
 
 import jpy
--- a/py/server/deephaven/table_factory.py
+++ b/py/server/deephaven/table_factory.py
@@ -3,13 +3,14 @@
 #
 
 """ This module provides various ways to make a Deephaven table. """
-from typing import List, Dict, Any, Union
+from __future__ import annotations
+from typing import List, Dict, Any, Union, Sequence
 
 import jpy
 
 from deephaven import DHError
 from deephaven._wrapper import JObjectWrapper
-from deephaven.column import InputColumn
+from deephaven.column import InputColumn, Column
 from deephaven.dtypes import DType
 from deephaven.jcompat import to_sequence
 from deephaven.table import Table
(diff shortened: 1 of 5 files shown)


py/server/deephaven/table_factory.py
```diff
@@ -3,13 +3,14 @@
 #
 
 """ This module provides various ways to make a Deephaven table. """
-from typing import List, Dict, Any, Union
+from __future__ import annotations
```
devinrsmith, line 6: What's this import used for?

py/server/tests/test_table_factory.py
```diff
@@ -200,11 +200,76 @@ def test_dtw_single_string_arg(self):
...
+            s_array = dtypes.array(dtypes.short, [128, 228, 328])
+            i_array = dtypes.array(dtypes.int32, [32768, 42768, 52768])
+            l_array = dtypes.array(dtypes.long, [2 ** 32, 2 ** 33, 2 ** 36])
+            f_array = dtypes.array(dtypes.float32, [1.0, 1.1, 1.2])
+            d_array = dtypes.array(dtypes.double, [1.0 / 2 ** 32, 1.1 / 2 ** 33, 1.2 / 2 ** 36])
+            str_array = dtypes.array(dtypes.string, ["some", "not so random", "text"])
+            table_writer.write_row(b_array, s_array, i_array, l_array, f_array, d_array, str_array
+                                   )
+        t = table_writer.table
+        self.wait_ticking_table_update(t, row_count=1, timeout=5)
```
devinrsmith, line 229: For clarity, I think this table that we are using as input to the inputs tables should be created via new_table instead of table writer.

py/server/tests/test_table_factory.py
```diff
@@ -200,11 +200,76 @@ def test_dtw_single_string_arg(self):
...
+    def test_input_table(self):
+        col_defs = {
+            "ByteArray": dtypes.byte_array,
+            "ShortArray": dtypes.short_array,
+            "Int32Array": dtypes.int32_array,
+            "LongArray": dtypes.long_array,
+            "Float32Array": dtypes.float32_array,
+            "DoubleArray": dtypes.double_array,
+            "StringArray": dtypes.string_array,
+        }
```
devinrsmith, line 217: Why are these tests oriented to array types instead of primitives? Nothing "wrong" with that, I just would have expected something around primitive types.

py/server/tests/test_table_factory.py
```diff
@@ -200,11 +200,76 @@ def test_dtw_single_string_arg(self):
...
+            append_only_input_table = InputTable(init_table=t)
+            with self.assertRaises(DHError) as cm:
+                append_only_input_table.delete(t)
+            self.assertIn("not allowed.", str(cm.exception))
+
+            keyed_input_table = InputTable(init_table=t, key_cols="StringArray")
+            self.assertEqual(keyed_input_table.size, 2)
+            keyed_input_table.delete(t.select("StringArray"))
+            self.assertEqual(keyed_input_table.size, 0)
+
```
devinrsmith, line 272: We should probably have a test w/ key_cols as array.

py/server/deephaven/table_factory.py
jmao-denver: it was needed for an earlier iteration, no longer needed but auto-import-org didn't remove it for some reason.

py/server/tests/test_table_factory.py
```diff
@@ -206,56 +206,42 @@ def test_dtw_single_string_arg(self):
...
-            str_array = dtypes.array(dtypes.string, ["some", "not so random", "text"])
-            table_writer.write_row(b_array, s_array, i_array, l_array, f_array, d_array, str_array
-                                   )
-        t = table_writer.table
-        self.wait_ticking_table_update(t, row_count=1, timeout=5)
-
+        cols = [
+            bool_col(name="Boolean", data=[True, False]),
+            byte_col(name="Byte", data=(1, -1)),
+            char_col(name="Char", data='-1'),
```
devinrsmith, line 212: How does this work? Shouldn't it be two items?

devinrsmith: Got it - so there are two characters here `-` and `1`. Thanks.

devinrsmith (approved):

py/server/deephaven/table_factory.py
```diff
@@ -186,3 +194,77 @@ def write_row(self, *values: Any) -> None:
             self._j_table_writer.logRowPermissive(values)
         except Exception as e:
             raise DHError(e, "failed to write a row.") from e
+
+
+class InputTable(Table):
+    """InputTable is a subclass of Table that allows the users to dynamic add/delete/modify data in it. There are two
```
chipkent, line 200:
```suggestion
    """InputTable is a subclass of Table that allows the users to dynamically add/delete/modify data in it. There are two
```

py/server/deephaven/table_factory.py
```diff
@@ -186,3 +194,77 @@ def write_row(self, *values: Any) -> None:
...
+        """
+        try:
+            if col_defs is None and init_table is None:
+                raise ValueError("either column definitions or init table should be provided.")
+            elif col_defs and init_table:
+                raise ValueError("both column definitions and init table are provided.")
+
+            if col_defs:
+                cols = [Column(name=n, data_type=t) for n, t in col_defs.items()]
+                j_arg_1 = _JTableDefinition.of([col.j_column_definition for col in cols])
```
chipkent, line 226: These could be a 1-liner

py/server/deephaven/table_factory.py
```diff
@@ -186,3 +194,77 @@ def write_row(self, *values: Any) -> None:
...
+            if key_cols:
+                super().__init__(_JKeyedArrayBackedMutableTable.make(j_arg_1, key_cols))
+            else:
+                super().__init__(_JAppendOnlyArrayBackedMutableTable.make(j_arg_1))
+            self.j_input_table = self.j_table.getAttribute(_J_INPUT_TABLE_ATTRIBUTE)
+            self.key_columns = key_cols
+        except Exception as e:
+            raise DHError(e, "failed to create a InputTable.") from e
+
+    def add(self, table: Table):
```
chipkent, line 240: Are we annotating the `None` return type?  I don't remember what standard we are doing or what is best for the IDE.  Maybe what is there is best.

py/server/deephaven/table_factory.py
```diff
@@ -186,3 +194,77 @@ def write_row(self, *values: Any) -> None:
...
+
+        Raises:
+            DHError
+        """
+        try:
+            self.j_input_table.add(table.j_table)
+        except Exception as e:
+            raise DHError(e, "add to InputTable failed.") from e
+
+    def delete(self, table: Table):
```
chipkent, line 255: Are we annotating the `None` return type?  I don't remember what standard we are doing or what is best for the IDE.  Maybe what is there is best.

py/server/deephaven/table_factory.py
```diff
@@ -186,3 +194,77 @@ def write_row(self, *values: Any) -> None:
...
+            else:
+                super().__init__(_JAppendOnlyArrayBackedMutableTable.make(j_arg_1))
+            self.j_input_table = self.j_table.getAttribute(_J_INPUT_TABLE_ATTRIBUTE)
+            self.key_columns = key_cols
+        except Exception as e:
+            raise DHError(e, "failed to create a InputTable.") from e
+
+    def add(self, table: Table):
+        """Writes rows in the provided table to this input table. If this is keyed input table, added rows with keys
+        that match existing rows will replace those rows.
```
chipkent, line 242:
```suggestion
        """Writes rows from the provided table to this input table. If this is a keyed input table, added rows with keys
        that match existing rows will replace those rows.
```

chipkent: What happens if keyed-rows are duplicated?  Error?  Only the last one is used?

py/server/deephaven/table_factory.py
```diff
@@ -186,3 +194,77 @@ def write_row(self, *values: Any) -> None:
...
+        Raises:
+            DHError
+        """
+        try:
+            self.j_input_table.add(table.j_table)
+        except Exception as e:
+            raise DHError(e, "add to InputTable failed.") from e
+
+    def delete(self, table: Table):
+        """Deletes the keys contained in the provided table from this keyed input table.
```
chipkent, line 256: Doc string is written as if the input table is always a keyed input table, but it could be keyed or append.

chipkent (changes requested):

devinrsmith (dismissed):

py/server/deephaven/table_factory.py
chipkent: What happens if you have a keyed table and the added table contains duplicates of keyed rows?  Does that result in an error?  Is there no error, but the last value is used?  etc.

jmao-denver: From reading the Java implementation code and adding a test case, the answer to the questions is: no error will be used, and yes the last duplicate value is used.

chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/1892
Conceptual: https://github.com/deephaven/deephaven.io/issues/1894
Reference: https://github.com/deephaven/deephaven.io/issues/1893
Blog: https://github.com/deephaven/deephaven.io/issues/1891

PR #3086, opened 2022-11-15
Add support for seek row through JS API

### LINKED ISSUE #2838: Add support for seek row through JS API
As a developer writing a JS client, I want a seekRow feature so that I can jump to a specific row in the table.

mofojed:
- Implement Seek row functionality
- Fixes #2838 

Tested on the JSAPI page with the basic table: http://localhost:10000/jsapi/table_basic.html
- Tested each column to ensure the value was found correctly, e.g.
```
await table.seekRow(0, table.columns[0], 'Number', 6)
await table.seekRow(0, table.columns[1], 'Number', 9)
await table.seekRow(0, table.columns[2], 'String', 'World')
```
Also tested with the random values table with all data types:
```

size = 100
scale = 1000

    "MyString=(i%11==0? null : `a`+(int)(scale*(Math.random()*2-1)))",
    "MyInt=(i%12==0 ? null : (int)(scale*(Math.random()*2-1)))",
    "MyFloat=(float)(i%14==0 ? null : i%10==0 ? 1.0F/0.0F: i%5==0 ? -1.0F/0.0F : (float) scale*(Math.random()*2-1))",
    "MyDouble=(double)(i%16==0 ? null : i%10==0 ? 1.0D/0.0D: i%5==0 ? -1.0D/0.0D : (double) scale*(Math.random()*2-1))",
    "MyBoolean = (i%17==0 ? null : (int)(10*Math.random())%2==0)",
    "MyChar = (i%18==0 ? null : new Character((char) (((26*Math.random())%26)+97)) )",
    "MyShort=(short)(i%19==0 ? null : (int)(scale*(Math.random()*2-1)))",
    "MyByte=(Byte)(i%19==0 ? null : new Byte( Integer.toString((int)(Byte.MAX_VALUE*(Math.random()*2-1)))))",
    "MyBigDecimal=(i%21==0 ? null : new java.math.BigDecimal(scale*(Math.random()*2-1)))",
    "MyBigInteger=(i%22==0 ? null : new java.math.BigInteger(Integer.toString((int)(scale*(Math.random()*2-1)))))"
])
```
- Tested each column, ensured the returned row matched:
```
await table.seekRow(0, table.columns[0], 'String', 'a891')
await table.seekRow(0, table.columns[1], 'Number', 217)
await table.seekRow(0, table.columns[2], 'Long', 814)
await table.seekRow(0, table.columns[2], 'Long', dh.LongWrapper.ofString('814'))
await table.seekRow(0, table.columns[3], 'Number', -946.1679077148438)
await table.seekRow(0, table.columns[4], 'Number', -515.6901244808021)
await table.seekRow(0, table.columns[5], 'Boolean', false)
await table.seekRow(0, table.columns[6], 'String', 'x')
await table.seekRow(0, table.columns[7], 'Number', -125)
await table.seekRow(0, table.columns[8], 'Number', 112)
await table.seekRow(0, table.columns[9], 'String', '-847.1314821279607940596179105341434478759765625')
await table.seekRow(0, table.columns[10], 'String', '757')
```


ClientSupport/src/main/java/io/deephaven/clientsupport/gotorow/SeekRow.java
```diff
@@ -0,0 +1,339 @@
...
+import io.deephaven.engine.rowset.RowSetFactory;
+import io.deephaven.engine.table.ColumnSource;
+import io.deephaven.engine.table.Table;
+import io.deephaven.internal.log.LoggerFactory;
+import io.deephaven.io.logger.Logger;
+import io.deephaven.time.DateTime;
+
+import java.util.Random;
+
+public class SeekRow implements Function<Table, Long> {
```
mofojed, line 19: This `Function` copied verbatim from Enterprise

server/src/main/java/io/deephaven/server/table/TableServiceGrpcImpl.java
```diff
@@ -276,11 +248,115 @@ public void updateBy(UpdateByRequest request, StreamObserver<ExportedTableCreati
...
+                                    seekValue,
+                                    request.getInsensitive(),
+                                    request.getContains(),
+                                    request.getIsBackward()));
+                            safelyExecuteLocked(responseObserver, () -> {
+                                SeekRowResponse.Builder rowResponse = SeekRowResponse.newBuilder();
+                                responseObserver.onNext(rowResponse.setResultRow(result).build());
+                                responseObserver.onCompleted();
+                            });
+                        } catch (Exception e) {
```
niloc132, line 351: shouldn't need a try/catch, any exception that is thrown will already be passed to the onError(responseObserver) above

web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
```diff
@@ -995,6 +992,71 @@ public Promise<JsColumnStatistics> getColumnStatistics(Column column) {
...
+            Object seekValue,
+            @JsOptional Boolean insensitive,
+            @JsOptional Boolean contains,
+            @JsOptional Boolean isBackwards) {
+        SeekRowRequest seekRowRequest = new SeekRowRequest();
+        seekRowRequest.setSourceId(state().getHandle().makeTicket());
+        seekRowRequest.setStartingRow(String.valueOf(startingRow));
+        seekRowRequest.setColumnName(column.getName());
+        seekRowRequest.setSeekValue(objectToLiteral(valueType, seekValue));
+        if (insensitive != null) {
```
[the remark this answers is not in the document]
mofojed: Yes

server/src/main/java/io/deephaven/server/table/TableServiceGrpcImpl.java
mofojed: Neat

server/src/main/java/io/deephaven/server/table/TableServiceGrpcImpl.java
```diff
@@ -276,11 +248,115 @@ public void updateBy(UpdateByRequest request, StreamObserver<ExportedTableCreati
         oneShotOperationWrapper(BatchTableRequest.Operation.OpCase.UPDATE_BY, request, responseObserver);
     }
 
+    private Object getSeekValue(Literal literal, Class<?> dataType) {
+        if (literal.hasStringValue()) {
+            return literal.getStringValue();
```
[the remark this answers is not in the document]
mofojed: It actually works fine with the String value. See in SeekRow.java:243, it does a String comparison if a String is passed in. I don't think it should assert here - if you pass in a value as a String it can still seek to it. (Remember, I didn't build this API - I'm just porting it.) With an assertion here, this case wouldn't work:

```
await table.seekRow(0, table.columns[1], 'String', '9')
```

Ditto with nanotime - you can pass in a date as a String and it will compare against the data type.
```
var strDate = dh.DateWrapper.ofJsDate(new Date()).valueOf()
await table.seekRow(0, table.columns[0], 'Datetime', strDate)
```

I can just translate the exception here and throw a `GrpcUtil.statusRuntimeException`?

web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
```diff
@@ -995,6 +992,71 @@ public Promise<JsColumnStatistics> getColumnStatistics(Column column) {
...
+                    break;
+                case Datetime:
+                    literal.setNanoTimeValue(value.toString());
+                    break;
+                case Boolean:
+                    literal.setBoolValue(Boolean.parseBoolean(value.toString()));
+                    break;
+                case Other:
+                default:
+                    throw new UnsupportedOperationException("Invalid value type for seekRow: " + valueType);
```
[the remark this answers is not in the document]
mofojed: I'm confused - this does throw, and the client gets it? Or do you mean convert it to a rejected promise? We do throw in other methods the same way in JsTable, see `JsTable.join` for example.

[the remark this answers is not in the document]
niloc132: Whoops, sorry, was confusing this with something else.

server/src/main/java/io/deephaven/server/table/TableServiceGrpcImpl.java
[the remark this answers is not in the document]
niloc132: I think this is the case we talked about over a call before, where yes, technically, it would compare the values, but it would compare them as strings, not numbers - so 100 < 99. For date values, that will only matter when our ns precision increases the decimal-encoded length, however this is more likely to happen for 'typical' numeric types.

Agreed that the ported "carq" is very permissive, I'm just saying that we should try to avoid that unless we have no other choice, and mark "this is probably wrong, and easy to be made right" cases more clearly for api consumers.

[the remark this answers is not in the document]
mofojed: I'm not sure when the case you're describing would get hit. I've filed an issue for numeric columns as it doesn't seem to go to the "closest" row for them: https://deephaven.atlassian.net/browse/DH-13839

I still want to just get this in to match what's in Enterprise, so we can deliver Go to row functionality to at least match what's already in Swing.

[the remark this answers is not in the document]
mofojed: Ahhh okay I've got it. With a table like:
```
t = empty_table(10_000).update(["x=i", "y=i*0.1", "z=i*1.0", "r=Math.random()*10_000"]).sort("r")
```
Then trying to seek to value `95` on column `Z` it ends up seeking to the row `9,499.0000` instead. Booo. Okay I'll limit it so you need to provide the value as the expected type I guess?

server/src/main/java/io/deephaven/server/table/ops/TableServiceGrpcImpl.java
```diff
@@ -296,12 +305,116 @@ public void updateBy(UpdateByRequest request, StreamObserver<ExportedTableCreati
...
+                                columnName,
+                                seekValue,
+                                request.getInsensitive(),
+                                request.getContains(),
+                                request.getIsBackward()));
+                        safelyExecuteLocked(responseObserver, () -> {
+                            SeekRowResponse.Builder rowResponse = SeekRowResponse.newBuilder();
+                            responseObserver.onNext(rowResponse.setResultRow(result).build());
+                            responseObserver.onCompleted();
+                        });
```
niloc132, line 413: Safely execute is gone now, with a new call to use instead that will prevent accidentally creating the response inside the "safe" part. See GrpcUtil.safelyComplete. Roughly, this will be:

```suggestion
                        SeekRowResponse.Builder rowResponse = SeekRowResponse.newBuilder();
                        safelyComplete(responseObserver, rowResponse.setResultRow(result).build());
```

server/src/main/java/io/deephaven/server/table/ops/TableServiceGrpcImpl.java
```diff
@@ -296,12 +305,116 @@ public void updateBy(UpdateByRequest request, StreamObserver<ExportedTableCreati
...
+            if (dataType == long.class) {
+                return longValue;
+            }
+            if (dataType == float.class) {
+                return longValue.floatValue();
+            }
+            if (dataType == double.class) {
+                return longValue.doubleValue();
+            }
+            if (dataType == BigInteger.class) {
```
niloc132, line 340: If we expect these to work, we'll need BigIntegerWrapper/etc support too?

It seems odd to expect a long (i.e. likely from reading from one column of the table) to search a biginteger/bigdecimal field (which would be a different column).

Is this speculative, how we expect it might work in the future? Or does this work already and I misunderstand?

mofojed: True, I guess I should be converting from a String in the BigInteger/BigWrapper case (since Literal doesn't seem to have an option for either, best to pass as string and avoid truncation).
I'll remove the conversion from long/double.

mofojed: Side note - why do the methods `Literal.getLongValue` and `Literal.getDoubleValue` have the wrong return type? It's returning a boxed type but the signature is for the primitive type, e.g.:
```
  @java.lang.Override
  public long getLongValue() {
    if (valueCase_ == 4) {
      return (java.lang.Long) value_;
    }
    return 0L;
  }
```
Seems odd, and since I'm doing `Long longValue = literal.getLongValue()`, so it's getting autounboxed/boxed in that case?

niloc132: (Note that we're talking about the JVM implementation now, io.deephaven.proto.backplane.grpc.Literal is not used in the JS client via GWT)

Yes, Java will autounbox in cases like this. Since value_ is an Object, it can't be returned as a `long` primitive, it has to be cast to java.lang.Long, then the compiler will auto-unbox it to a long. Unless the value is null (which will not fail the cast), as that will throw a NullPointerException rather than return a value.

server/src/main/java/io/deephaven/server/table/ops/TableServiceGrpcImpl.java
```diff
@@ -55,12 +59,17 @@
...
 import io.deephaven.server.session.TicketRouter;
+import io.deephaven.server.table.ops.GrpcTableOperation;
+import io.deephaven.time.DateTime;
 import io.deephaven.server.table.ExportedTableUpdateListener;
 import io.grpc.StatusRuntimeException;
 import io.grpc.stub.ServerCallStreamObserver;
 import io.grpc.stub.StreamObserver;
 
 import javax.inject.Inject;
+import java.lang.Object;
```
niloc132, line 70: Object should never need an import?

mofojed: Weird. Cleaned up and removed another unused import as well.

niloc132 (approved):

Issue #3088, opened 2022-11-16
Improve / document whereIn with empty columns


devinrsmith:
`whereIn` / `whereNotIn` does not have documentation around what to do if `columnsToMatch` is empty.

The current behavior is logically consistent (given an engine frame of mind): if the right table is empty, the results will be empty; if the right table is non-empty, the results will be equal to the left table.

There are a few "reasonable" assumptions a user could make if we don't explicitly document the behavior:

1. The way it works today
2. The columns are implicitly all the right table columns (this is what I assumed)
3. The columns are implicitly all the left table columns
4. The columns are implicitly the intersection of left & right table columns
5. An exception (columns must be explicit)

At the TableSpec (and gRPC after #990) level, the current behavior is the require explicit column names.

There's also potential to introduce more explicitly typed methods `TOPS whereIn(TABLE)`, like `selectDistinct` does.

Part of epic #3026

PR #3090, opened 2022-11-16
Plumb whereIn/whereNotIn through gRPC

### LINKED ISSUE #990: TableService implementation of whereIn/whereNotIn
See `io.deephaven.api.TableOperations#whereIn` and `io.deephaven.api.TableOperations#whereNotIn`

devinrsmith:
Additionally, removes WhereNotInTable and collapses logic into WhereInTable parameters.

Follow-up #3088

Fixes #990


rcaudy (dismissed):

server/src/main/java/io/deephaven/server/table/ops/WhereInGrpcImpl.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public final Table create(final WhereInRequest request, final List<SessionState.ExportObject<Table>> sourceTables) {
+        Assert.eq(sourceTables.size(), "sourceTables.size()", 2);
+        Assert.gtZero(request.getColumnsToMatchCount(), "request.getColumnsToMatchCount()");
+        final Table left = sourceTables.get(0).get();
+        final Table right = sourceTables.get(1).get();
+        final List<JoinMatch> columnsToMatch = JoinMatch.from(request.getColumnsToMatchList());
+        final AwareFunctionalLock lock =
+                left.isRefreshing() || right.isRefreshing() ? updateGraphProcessor.sharedLock() : null;
+        if (lock != null) {
```
niloc132, line 53: why explicit lock+unlock over computeLocked? at least as a reader the computeLocked is more concise, more obviously correct ("any chance that something could throw on the way out of lock() that would leave it locked, should we just do the lock inside the try/catch").

Example
io.deephaven.server.table.ops.JoinTablesGrpcImpl#create

```
        final Table result;
        if (!lhs.isRefreshing() && !rhs.isRefreshing()) {
            result = realTableOperation.apply(lhs, rhs, columnsToMatch, columnsToAdd, request);
        } else {
            result = updateGraphProcessor.sharedLock().computeLocked(
                    () -> realTableOperation.apply(lhs, rhs, columnsToMatch, columnsToAdd, request));
        }
        return result;
```
(could just as easily return from inside the if/else)

niloc132 (approved):
One minor remark, discussed in slack also about getting js api bindings closer to working

server/src/main/java/io/deephaven/server/table/ops/WhereInGrpcImpl.java
devinrsmith: I'm not a fan of lambda abstractions for singular operations (I see where the abstraction comes in handy for join where we have a lot of them), and the branching that results around the application. That said, the version I has has branching logic around the lock... something I personally prefer, but I think we can do better...

@rcaudy @niloc132 , would you be opposed to adding a new method to FunctionalLock:

```java
public interface FunctionalLock extends Lock {
    ...
    default SafeCloseable closeableLock() {
        final SafeCloseableLock closeable = new SafeCloseableLock(this);
        lock();
        return closeable;
    }
```

This would allow the use of a try resource, which I think is much more ergonomic:

```java
    @Override
    public final Table create(final WhereInRequest request, final List<SessionState.ExportObject<Table>> sourceTables) {
        ...
        try (final SafeCloseable _lock = lock(left, right)) {
            return request.getInverted() ? left.whereNotIn(right, columnsToMatch) : left.whereIn(right, columnsToMatch);
        }
    }

    private SafeCloseable lock(Table left, Table right) {
        return left.isRefreshing() || right.isRefreshing() ? updateGraphProcessor.sharedLock().closeableLock() : null;
    }
```

rcaudy: I don't hate it. Not a bad option, for cases where lambdas are annoying.

server/src/main/java/io/deephaven/server/table/ops/WhereInGrpcImpl.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public final Table create(final WhereInRequest request, final List<SessionState.ExportObject<Table>> sourceTables) {
+        Assert.eq(sourceTables.size(), "sourceTables.size()", 2);
+        Assert.gtZero(request.getColumnsToMatchCount(), "request.getColumnsToMatchCount()");
+        final Table left = sourceTables.get(0).get();
+        final Table right = sourceTables.get(1).get();
+        final List<JoinMatch> columnsToMatch = JoinMatch.from(request.getColumnsToMatchList());
+        try (final SafeCloseable _lock = lock(left, right)) {
```
niloc132, line 51: why a new idiom here, when we have one already?

devinrsmith: See commentary earlier for color - it's more concise and easier to read IMO.

Util/src/main/java/io/deephaven/util/locks/SafeCloseableUnlock.java
```diff
@@ -0,0 +1,19 @@
+package io.deephaven.util.locks;
+
+import io.deephaven.util.SafeCloseable;
+
+import java.util.Objects;
+import java.util.concurrent.locks.Lock;
+
+final class SafeCloseableUnlock implements SafeCloseable {
```
rcaudy: Seems like a lot of boilerplate for something that could be a method reference to `this::unlock`.

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphLock.java
```diff
@@ -386,6 +387,16 @@ public <RESULT_TYPE, EXCEPTION_TYPE extends Exception> RESULT_TYPE computeLocked
...
+        @Override
+        public SafeCloseable lockCloseable() {
+            return delegate.lockCloseable();
+        }
+
+        @Override
+        public SafeCloseable lockInterruptiblyCloseable() throws InterruptedException {
+            return delegate.lockInterruptiblyCloseable();
+        }
+
```
rcaudy, line 399: I'm not sure this is correct.

Util/src/main/java/io/deephaven/util/locks/SafeCloseableUnlock.java
rcaudy: Never mind, I see you did this.

rcaudy (approved):

PR #3098, opened 2022-11-21
Improve DebugAwareFunctionalLock

devinrsmith:


engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphLock.java
```diff
@@ -14,8 +14,8 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import java.util.ArrayDeque;
 import java.util.Deque;
+import java.util.concurrent.ConcurrentLinkedDeque;
```
rcaudy, line 18: Please get rid of the unused import introduced in the last PR:
`import io.deephaven.util.SafeCloseable;`

rcaudy (commented):
This is not an unreasonable approach. It's better than what we have. I think we could fairly easily address the shared lock use case as well, but it would be more work than this.

rcaudy (approved):

PR #3099, opened 2022-11-22
Add the convenience function garbage_collect()

### LINKED ISSUE #3074: Add a Deephaven garbage collection util in Python
As a user, I want to easily call for manual garbage collection

We currently have to call for garbage collection twice. Once in Python and once in Java. In searching our slack, I found a handful of instances where people would call just the Python gc and not know they needed to call the Java gc too.

We should add a util to wrap this functionality so it is nicer to use

### Current
```python
import gc

gc.collect()
jpy.get_type("java.lang.System").gc()
```

### Proposed
```python
from deephaven import gc

gc() # Or something like this
```

jmao-denver:
Fixes #3074

### DIFF
--- a/py/server/deephaven/utils.py
+++ b/py/server/deephaven/utils.py
@@ -1,0 +1,24 @@
+#
+#     Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+#
+"""This module provides the helper and convenience functions that are not part of the Deephaven server but
+nonetheless uniquely useful to user applications built on top of Deephaven."""
+import gc
+
+import jpy
+
+from deephaven import DHError
+
+
+def garbage_collect() -> None:
+    """Runs full garbage collection in Python first and then requests the JVM to run its garbage collector. Since there
+    is no way to force the Java garbage collector to run, the effect of calling this function is non-deterministic.
+
+    Raises:
+        DHError
+    """
+    try:
+        gc.collect()
+        jpy.get_type("java.lang.System").gc()
+    except Exception as e:
+        raise DHError(e, "failed to initiate system-wide garbage collection.")
--- a/py/server/deephaven/utils.py
+++ b/py/server/deephaven/utils.py
@@ -11,14 +11,18 @@ from deephaven import DHError
 
 
 def garbage_collect() -> None:
-    """Runs full garbage collection in Python first and then requests the JVM to run its garbage collector. Since there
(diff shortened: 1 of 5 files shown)


py/server/deephaven/utils.py
```diff
@@ -0,0 +1,24 @@
...
+def garbage_collect() -> None:
+    """Runs full garbage collection in Python first and then requests the JVM to run its garbage collector. Since there
+    is no way to force the Java garbage collector to run, the effect of calling this function is non-deterministic.
+
+    Raises:
+        DHError
+    """
+    try:
+        gc.collect()
+        jpy.get_type("java.lang.System").gc()
```
devinrsmith, line 22: I've often found that it takes two calls to .gc() to free some of the python<->java references. I'm not sure if it can be executed twice in the same script invocation, or if they need to be in separate executions. We should be able to do a little testing; and if applicable, should invoke gc() here twice.

jmao-denver: Needing to call gc() twice makes sense and it is  allowed in my testing. I have also updated the doc string.

py/server/deephaven/utils.py
```diff
@@ -0,0 +1,28 @@
+#
```
chipkent, line 1: I am very strongly against `utils` modules.  They are dumping grounds that never end up well.  This should be a `gc` module, a `system` module, in the `deephaven` module, or something else that is specific.

py/server/deephaven/utils.py
```diff
@@ -0,0 +1,28 @@
...
+#     Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+#
+"""This module provides the helper and convenience functions that are not part of the Deephaven server but
+nonetheless uniquely useful to user applications built on top of Deephaven."""
```
chipkent, line 5: This docstring reinforces my prior comment.

py/server/tests/test_utils.py
```diff
@@ -0,0 +1,17 @@
+#
```
chipkent, line 1: module name

chipkent (changes requested):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/1909
Conceptual: https://github.com/deephaven/deephaven.io/issues/1908
Reference: https://github.com/deephaven/deephaven.io/issues/1910
Blog: https://github.com/deephaven/deephaven.io/issues/1911

Issue #2721, opened 2022-08-08
InputTable#add stuck waiting for sequence


nbauernfeind:
Carson ran into an issue with an InputTable when working on one of the client API's.

The stack trace had an unresolved export waiting on an input table update:
```
"DeephavenApiServer-Scheduler-Serial-1" #35 daemon prio=5 os_prio=0 cpu=2232.18ms elapsed=337.81s tid=0x00007f28580f6800 nid=0xa0 in Object.wait()  [0x00007f288858e000]
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(java.base@11.0.16/Native Method)
	- waiting on <no object reference available>
	at java.lang.Object.wait(java.base@11.0.16/Object.java:328)
	at io.deephaven.engine.table.impl.util.BaseArrayBackedMutableTable$ArrayBackedMutableInputTable.waitForSequence(BaseArrayBackedMutableTable.java:278)
	- waiting to re-lock in wait() <0x0000000705bb9970> (a java.util.Collections$SynchronizedRandomAccessList)
	at io.deephaven.engine.table.impl.util.BaseArrayBackedMutableTable$ArrayBackedMutableInputTable.add(BaseArrayBackedMutableTable.java:201)
	at io.deephaven.server.table.inputtables.InputTableServiceGrpcImpl.lambda$addTableToInputTable$1(InputTableServiceGrpcImpl.java:73)
	at io.deephaven.server.table.inputtables.InputTableServiceGrpcImpl$$Lambda$1049/0x0000000840789c40.run(Unknown Source)
	at io.deephaven.server.session.SessionState$ExportBuilder.lambda$submit$2(SessionState.java:1299)
	at io.deephaven.server.session.SessionState$ExportBuilder$$Lambda$988/0x000000084074fc40.call(Unknown Source)
	at io.deephaven.server.session.SessionState$ExportObject.doExport(SessionState.java:847)
	at io.deephaven.server.session.SessionState$ExportObject$$Lambda$1050/0x000000084078a040.run(Unknown Source)
	at java.util.concurrent.Executors$RunnableAdapter.call(java.base@11.0.16/Executors.java:515)
	at java.util.concurrent.FutureTask.run(java.base@11.0.16/FutureTask.java:264)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(java.base@11.0.16/ThreadPoolExecutor.java:1128)
[...]

PR #3101, opened 2022-11-22
Add gRPC Aggregation coverage

### LINKED ISSUE #991: TableService aggregation coverage
Currently missing:
* weighted sum
* count distinct
* distinct
* unique
* sortedFirst
* sortedLast
* percentile/median field averageMedian

devinrsmith:
Fixes #991


proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -579,6 +592,162 @@ message ComboAggregateRequest {
...
+    AggSpec spec = 1;
+    repeated string match_pairs = 2;
+  }
+
+  message AggregationCount {
+    // The output column name
+    string column_name = 1;
+  }
+
+  message AggregationRowKey {
```
rcaudy, line 733: @niloc132 how do you feel about this sort of thing? Should we be defining separate messages?

table-api/src/main/java/io/deephaven/api/agg/spec/AggSpecUnique.java
```diff
@@ -17,6 +18,12 @@
 @BuildableStyle
 public abstract class AggSpecUnique extends AggSpecBase {
 
+    public static final boolean INCLUDE_NULLS_DEFAULT = false;
```
rcaudy, line 21: You should be consistent in your changes, here. You've added a new pattern for `AggSpecUnique`, but not similar specs like `AggSpecCountDistinct`.

table-api/src/main/java/io/deephaven/api/agg/spec/AggSpecUnique.java
```diff
@@ -8,6 +8,7 @@
 import org.immutables.value.Value.Immutable;
 
 import javax.annotation.Nullable;
+import java.util.Optional;
```
rcaudy, line 11: Some imports unused now.

table-api/src/main/java/io/deephaven/api/agg/spec/AggSpecUnique.java
```diff
@@ -56,25 +64,32 @@ public final String description() {
...
      *
      * @return The non-unique sentinel value
      */
-    @Nullable
-    @Default
-    public Object nonUniqueSentinel() {
-        return null;
+    public abstract Optional<Object> nonUniqueSentinel();
+
+    public final boolean includeNullsOrDefault() {
```
rcaudy, line 77: @niloc132 We should discuss this as a group. Do we really want clients to accept server side defaults in scenarios where we have a wire-compatible APIs that mean something else on the server? I really don't think that's best.

table-api/src/main/java/io/deephaven/api/agg/Aggregation.java
```diff
@@ -655,6 +671,21 @@ static Sentinel Sentinel() {
         return Sentinel.of();
     }
 
+    /**
+     * Calls every single visit method of {@code visitor} with a {@code null} object.
+     *
+     * @param visitor the visitor
+     */
+    static void visitAll(Visitor visitor) {
```
rcaudy, line 679: I think that this method and `AggSpec.visitAll` should move to test code. I can't think of a "production" scenario.

qst/build.gradle
```diff
@@ -9,6 +9,7 @@ dependencies {
     api project(':table-api')
 
     Classpaths.inheritImmutables(project)
+    compileOnly 'com.google.code.findbugs:jsr305:3.0.2'
```
rcaudy, line 12: What do we need this new dep for?

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/GrpcUtil.java
```diff
@@ -137,7 +137,7 @@ public static void safelyExecute(final FunctionalInterfaces.ThrowingRunnable<Exc
         try {
             runner.run();
         } catch (final Exception err) {
-            log.debug().append("Unanticipated gRPC Error: ").append(err).endl();
+            log.error().append("Unanticipated gRPC Error: ").append(err).endl();
```
rcaudy, line 140: What were we trying to suppress when it was `debug`?

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -579,6 +592,162 @@ message ComboAggregateRequest {
...
+
+  // This is a catch-all blank message. It is for AggSpecs that don't have any fields. This construction allows for
+  // AggSpec specialization in the future with breaking backwards compatibility.
+  message AggSpecBlank {}
+
+  message AggSpecApproximatePercentile {
+    // Percentile. Must be in range [0.0, 1.0].
+    double percentile = 1;
+
+    // T-Digest compression factor. Must be greater than or equal to 1. Defaults to 100. 1000 is extremely large.
```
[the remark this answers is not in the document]
devinrsmith: I don't know. The implementation uses double through and through.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/GrpcUtil.java
devinrsmith: I'm not sure. I don't think @nbauernfeind remembers either.

server/src/main/java/io/deephaven/server/table/ops/AggSpecAdapter.java
```diff
@@ -95,27 +133,23 @@ private static io.deephaven.api.agg.spec.AggSpecApproximatePercentile adapt(
...
     }
 
     private static io.deephaven.api.agg.spec.AggSpecPercentile adapt(AggSpecPercentile percentile) {
-        return percentile.hasAverageEvenlyDivided()
-                ? AggSpec.percentile(percentile.getPercentile(), percentile.getAverageEvenlyDivided())
-                : AggSpec.percentile(percentile.getPercentile());
+        return AggSpec.percentile(percentile.getPercentile(), percentile.getAverageEvenlyDivided());
     }
 
     private static SortColumn adapt(AggSpecSortedColumn sortedColumn) {
```
rcaudy, line 155: Still some shared messages. Should we change that?

server/src/main/java/io/deephaven/server/table/ops/TableServiceGrpcImpl.java
```diff
@@ -309,6 +312,12 @@ public void fetchPandasTable(FetchPandasTableRequest request,
     public void batch(final BatchTableRequest request,
             final StreamObserver<ExportedTableCreationResponse> responseObserver) {
         GrpcUtil.rpcWrapper(log, responseObserver, () -> {
+            GrpcErrorHelper.checkRepeatedFieldNonEmpty(request, BatchTableRequest.OPS_FIELD_NUMBER);
```
rcaudy, line 315: I wonder if this fixed Carson's issue?

table-api/src/main/java/io/deephaven/api/agg/spec/AggSpecApproximatePercentile.java
```diff
@@ -5,55 +5,78 @@
...
     public abstract double percentile();
 
     /**
-     * T-Digest compression factor. Must be greater than or equal to 1. Defaults to 100. 1000 is extremely large.
-     * 
-     * @return The T-Digest compression factor
+     * T-Digest compression factor. Must be greater than or equal to 1. 1000 is extremely large.
+     *
+     * <p>
+     * When not specified, the server will choose a compression value.
```
rcaudy, line 75: The engine, not the server. Or "the implementation".

table-api/src/main/java/io/deephaven/api/agg/spec/AggSpecCountDistinct.java
```diff
@@ -3,23 +3,37 @@
...
 
 /**
  * Specifies an aggregation that outputs the count of distinct values for each group.
  */
 @Immutable
-@BuildableStyle
+@SimpleStyle
 public abstract class AggSpecCountDistinct extends AggSpecBase {
 
+    public static final boolean COUNT_NULLS_DEFAULT = SemanticDefaults.COUNT_NULLS_DEFAULT;
```
rcaudy, line 17: Please clean out `SemanticDefaults`.

java-client/session/src/main/java/io/deephaven/client/impl/AggSpecBuilder.java
```diff
@@ -143,72 +152,104 @@ public void visit(io.deephaven.api.agg.spec.AggSpecSortedLast sortedLast) {
...
+        public AggSpecNonUniqueSentinel visit(double x) {
+            return AggSpecNonUniqueSentinel.newBuilder().setDoubleValue(x).build();
+        }
+
+        @Override
+        public AggSpecNonUniqueSentinel visit(Object x) {
+            if (x instanceof String) {
+                return AggSpecNonUniqueSentinel.newBuilder().setStringValue((String) x).build();
+            }
+            throw new IllegalArgumentException(
```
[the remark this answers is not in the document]
devinrsmith: protobuf has a well-known google.protobuf.Timestamp type. Although, in other places we've just used sint64 start_time_nanos = 2 [jstype=JS_STRING];. I wonder if we should break all usages of `start_time_nanos` into google.protobuf.Timestamp api for our Dec release? (There is also a google.protobuf.Duration message we could use where applicable.)

Personally, I think using the protobuf well-known Timestamp might be better (and provide more type-safe generated code).

I'm going to create a ticket for "codec" support, which I think would cover big ints and others.

qst/build.gradle
devinrsmith: Not needed anymore, removed.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/GrpcUtil.java
devinrsmith: I'm going to suggest logging these as error, and re-evaluate later if we see too much logspam.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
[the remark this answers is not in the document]
devinrsmith: I'm going to suggest leaving as double, given that's ultimately what the engine implementation uses.

server/src/main/java/io/deephaven/server/table/ops/AggSpecAdapter.java
devinrsmith: I vote no.

server/src/main/java/io/deephaven/server/table/ops/TableServiceGrpcImpl.java
devinrsmith: I couldn't find an explicit issue you were referencing. Maybe thought of https://github.com/deephaven/deephaven-core/issues/2721, but I don't think that's relevant.

java-client/session/src/main/java/io/deephaven/client/impl/AggSpecBuilder.java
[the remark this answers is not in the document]
devinrsmith: https://github.com/deephaven/deephaven-core/issues/3212

engine/table/src/main/java/io/deephaven/engine/table/impl/by/AggregationProcessor.java
```diff
@@ -1618,8 +1612,41 @@ private static IterativeChunkedAggregationOperator makeUniqueOperator(
...
+        return obj == null ? NULL_LONG : obj.longValue();
+    }
+
+    private static double doubleValue(AnnotatedObject obj) {
+        return obj == null ? NULL_DOUBLE : obj.doubleValue();
+    }
+
+    private static float floatValue(AnnotatedObject obj) {
+        return obj == null ? NULL_FLOAT : obj.floatValue();
+    }
```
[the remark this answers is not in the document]
devinrsmith: I'm not sure the best place a general purpose utility library around this should live (given dependency on table-api and Util).  Best to keep it closer to engine-table IMO until we know we need it wider. I've extracted it into an engine-table package private utility class for now.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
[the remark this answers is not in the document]
rcaudy: Please leave it as double.

java-client/session/src/main/java/io/deephaven/client/impl/AggSpecBuilder.java
```diff
@@ -174,13 +175,15 @@ public void visit(io.deephaven.api.agg.spec.AggSpecTDigest tDigest) {
...
+        final AggSpecUnique.Builder builder = AggSpecUnique.newBuilder()
+                .setIncludeNulls(unique.includeNulls())
+                .setNonUniqueSentinel(adapt(unique.nonUniqueSentinel().orElse(null)));
         out = spec(Builder::setUnique, builder);
     }
 
-    private static AggSpecNonUniqueSentinel adapt(AnnotatedObject obj) {
-        return obj.visit(AggSpecNonUniqueSentinelAdapter.INSTANCE);
+    private static AggSpecNonUniqueSentinel adapt(UnionObject obj) {
+        return obj == null ? AggSpecNonUniqueSentinel.newBuilder().setNullValue(NullValue.NULL_VALUE).build()
```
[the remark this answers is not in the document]
devinrsmith: Well, it is "arbitrary" in the sense that we _could_ define this proto differently if we wanted to; but it is a well-known type/value as part of protobuf: https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#nullvalue. Programmatically, `NullValue.NULL_VALUE` is the only instance of type `google.protobuf.NullValue`.

For example, instead of:

```proto
  message AggSpecNonUniqueSentinel {
    oneof type {
      google.protobuf.NullValue null_value = 1;
```

we could do

```
  message AggSpecNonUniqueSentinel {
    oneof type {
      int32 null_value = 1;
```

and

```java
AggSpecNonUniqueSentinel.newBuilder().setNullValue(0).build()
```

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -579,6 +594,222 @@ message ComboAggregateRequest {
...
+    // Whether to include null values as a distinct value for determining if there is only one unique value to output
+    bool include_nulls = 1;
+
+    // The output value to use for groups that don't have a single unique input value
+    AggSpecNonUniqueSentinel non_unique_sentinel = 2;
+  }
+
+  message AggSpecNonUniqueSentinel {
+    oneof type {
+      google.protobuf.NullValue null_value = 1;
```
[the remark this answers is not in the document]
devinrsmith: Why won't the oneof be compatible with it? Seems to work for me.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -579,6 +594,222 @@ message ComboAggregateRequest {
...
+      string string_value = 2;
+      sint32 int_value = 3;
+      sint64 long_value = 4 [jstype=JS_STRING];
+      float float_value = 5;
+      double double_value = 6;
+      bool bool_value = 7;
+      // Expected to be in range [Byte.MIN_VALUE, Byte.MAX_VALUE]
```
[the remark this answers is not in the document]
devinrsmith: Something we could debate. Should we throw an error if the client tries to send a value that represents one of our null primitive values? (for int, long, float, double, etc) @rcaudy

java-client/session/src/main/java/io/deephaven/client/impl/AggSpecBuilder.java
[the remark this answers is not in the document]
devinrsmith: I'll give this a shot.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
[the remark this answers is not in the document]
devinrsmith: From a typing perspective, `null_value` doesn't have any type information associated with it, where-as `byte_value = Byte.MIN_VALUE` *does*. I'm not sure if that's a useful distinction we want to support or make w/ this structure.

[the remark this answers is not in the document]
devinrsmith: I might argue that errors wrt "null" primitive values should be the responsibility of the engine impl - if the engine allows explicitly setting `Integer.MIN_VALUE` as the sentinel value, we shouldn't try to stop that over the wire IMO.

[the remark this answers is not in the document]
rcaudy: I'm ambivalent, here. I think it could be expedient to just accept the null values as in range, and assume the user wanted that, but I won't complain if we want to validate it and require them to use the `null` type.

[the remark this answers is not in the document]
devinrsmith: If we want to validate non-null primitives, we should either validate it in `io.deephaven.api.agg.spec.AggSpecUnique` or `io.deephaven.engine.table.impl.by.AggregationProcessor#makeUniqueOperator` IMO.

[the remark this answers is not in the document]
devinrsmith: And actually, it's more nuanced and should _probably_ be handled by `AggregationProcessor`. The implementation works right now be converting types based on the column type it's associated with based on `Number`. We don't have enough information at wire time, or `AggSpecUnique` time, to known whether the user is just trying to "save" bytes over the wire and take advantage of the implementations implicit coercions.

devinrsmith:
I need to regenerate the JS side before merging. As soon as I get thumbs up from Colin, I'll regenerate JS and push for final approval.

niloc132 (dismissed):
Approved based on the structure of the proto, going to have to actually try to use it next.

rcaudy (approved):

PR #3110, opened 2022-11-23
Wrap the Table slice op

### LINKED ISSUE #3089: Add Table slice operation to the Python Table wrapper
The slice() method is available in the Java interface but missing in the Python wrapper

jmao-denver:
Fixes #3089

### DIFF
--- a/py/server/tests/test_table.py
+++ b/py/server/tests/test_table.py
@@ -782,6 +782,33 @@ class TableTestCase(BaseTestCase):
         self.assertIn("AAPL-GOOG", html_output)
         self.assertIn("2000", html_output)
 
+    def test_slice(self):
+        with ugp.shared_lock():
+            t = time_table("00:00:00.01")
+        rt = t.slice(0, 3)
+        self.assert_table_equals(t.head(3), rt)
+
+        self.wait_ticking_table_update(t, row_count=5, timeout=5)
+        with ugp.exclusive_lock():
+            rt = t.slice(t.size, -2)
+            self.assertEqual(0, rt.size)
+        self.wait_ticking_table_update(rt, row_count=1, timeout=5)
+        self.assertGreaterEqual(rt.size, 1)
+
+        rt = t.slice(-3, 0)
+        self.assert_table_equals(t.tail(3), rt)
+
+        rt = t.slice(-3, -2)
+        self.wait_ticking_table_update(rt, row_count=1, timeout=5)
+        self.assertEqual(1, rt.size)
+
+        rt = t.slice(1, 3)
+        self.wait_ticking_table_update(rt, row_count=2, timeout=5)
+        self.assertEqual(2, rt.size)
+
+        with self.assertRaises(DHError):
+            rt = t.slice(3, 2)
+
 
 def global_fn() -> str:
     return "global str"
--- a/py/server/tests/test_table.py
+++ b/py/server/tests/test_table.py
@@ -789,7 +789,7 @@ class TableTestCase(BaseTestCase):
         self.assert_table_equals(t.head(3), rt)
 
         self.wait_ticking_table_update(t, row_count=5, timeout=5)
-        with ugp.exclusive_lock():
+        with ugp.shared_lock():
             rt = t.slice(t.size, -2)
             self.assertEqual(0, rt.size)
         self.wait_ticking_table_update(rt, row_count=1, timeout=5)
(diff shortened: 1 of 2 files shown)


rcaudy (dismissed):

py/server/tests/test_table.py
```diff
@@ -782,6 +782,33 @@ def test_python_field_access(self):
...
+
+        self.wait_ticking_table_update(t, row_count=5, timeout=5)
+        with ugp.exclusive_lock():
+            rt = t.slice(t.size, -2)
+            self.assertEqual(0, rt.size)
+        self.wait_ticking_table_update(rt, row_count=1, timeout=5)
+        self.assertGreaterEqual(rt.size, 1)
+
+        rt = t.slice(-3, 0)
+        self.assert_table_equals(t.tail(3), rt)
```
chipkent, line 799: You have implemented a nice assert here using `tail`.  All of the methods here can easily be tested the same way using combinations of `head` and `tail`.

py/server/tests/test_table.py
```diff
@@ -782,6 +782,33 @@ def test_python_field_access(self):
...
+        rt = t.slice(-3, -2)
+        self.wait_ticking_table_update(rt, row_count=1, timeout=5)
+        self.assertEqual(1, rt.size)
+
+        rt = t.slice(1, 3)
+        self.wait_ticking_table_update(rt, row_count=2, timeout=5)
+        self.assertEqual(2, rt.size)
+
+        with self.assertRaises(DHError):
+            rt = t.slice(3, 2)
```
chipkent, line 810: Is this really the behavior we want?  It seems like we may want to just return an empty table.  In fact, if an index is negative, that is what we do.

chipkent (changes requested):

py/server/tests/test_table.py
jmao-denver: The Java doc is very explicit about this behavior. It does seem inconsistent though but I would rather have an exception than an empty table which would take up some resource and would not be of any use. @rcaudy your thoughts?

rcaudy: I'm not particularly interested in re-litigating the semantics of this operation while wrapping from Python, and I'm not unhappy with the current behavior.

We throw an exception for non-sensical inputs in 2 scenarios:
* `firstPositionInclusive < 0 && lastPositionExclusive > 0` : This is a range counted from the end for first and from the beginning for end.
* Both positive or both negative and `lastPositionExclusive < firstPositionInclusive` : This is an "inverted" range that can never be non-empty.
We could just call these impossible ranges "empty", but that seems user-hostile in (likely, IMO) cases where the user made a mistake in preparing their inputs. In the current approach, they can choose to handle the exception.

The "exception" is that we always allow:
* `firstPositionInclusive >= 0 && `lastPositionExclusive < 0` : This is a range defined relative to the end and the beginning, but not necessarily an empty range. This is just as valid as ranges larger than the current table size in either direction, and the output is only empty if there aren't currently any rows in it.

chipkent (changes requested):
I'm leaving this open because of the ongoing discussion item.

py/server/tests/test_table.py
rcaudy: I wouldn't call giving someone a useless result when they almost certainly screwed up their inputs "graceful". It's also just strictly worse when you imagine a scenario with a ticking parent table; the user could proceed to set up a whole scaffold of child operations that wait forever for data that can never change. This is the kind of thing enterprise users hate us for when we them the way you're proposing.

Regardless, if you want to discuss changing this, it can be a separate issue targeting the query engine; the Python wrapper should stay very thin here.

rcaudy: This is similar to `TableTools.merge` with no non-null inputs. We throw an exception, rather than giving the users an empty result that will never, ever tick.

rcaudy: I agree with Raffi's position as stated above. I do see it as reasonable to support a start relative to end and a last relative to start.

jjbrosnan: I agree with Raffi.

- I've never really viewed it as a benefit that Python returns an empty object when you give it bad indices to slice.
- I have always viewed it as a benefit to be able to slice relative to the end of an object.

I don't think throwing an error when incorrectly attempting to slice a table will detract from the user experience, so long as the error makes it obvious.

chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/1917
Conceptual: https://github.com/deephaven/deephaven.io/issues/1915
Reference: https://github.com/deephaven/deephaven.io/issues/1918
Blog: https://github.com/deephaven/deephaven.io/issues/1916

PR #3111, opened 2022-11-25
Closes #3105:  Extract Test helpers into publishable packages

### LINKED ISSUE #3105: Extract Test Utilities into separate published package
Writing unit tests for projects using Deephaven can be difficult.  Deephaven already has sophisticated test utilities built into the test configurations of existing packages.  We can extract these into publishable artifacts for the community to use.

Enterprise did this and also made a couple of additional improvements.  Some of these will probably spin off into separate tickets.

1. Expose TstUtils
2. Expose test column sources
3. Expose test data generators

abaranec:


devinrsmith (commented):
We should look into https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures. I think there may be explicit metadata that gets published in this regards that is useful to downstream consumers.

The main question is "do we really want to do this?"

I see the value, in theory, in exposing DH test libraries. In practice, I'm think our testing scaffolding could use a lot of work. In particular, it's unfortunate that we still rely on junit 4.x.

rcaudy:
> We should look into https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures. I think there may be explicit metadata that gets published in this regards that is useful to downstream consumers.
> 
> The main question is "do we really want to do this?"
> 
> I see the value, in theory, in exposing DH test libraries. In practice, I'm think our testing scaffolding could use a lot of work. In particular, it's unfortunate that we still rely on junit 4.x.

1. Can you give more actionable feedback re: text fixtures?
2. We want to make it easier for end-users to test their usage of Deephaven. This is a port of enterprise changes that were made a year ago. The only reason not to do this, IMO, is if it precludes later improvement.

devinrsmith (approved):
Nothing offensive. Mostly mechanical changes I can get behind.

I've added a couple commits of my own. Will want @rcaudy to check that part specifically.

devinrsmith:
Note: we'll want to hold off on merge until nightlies pass; I don't think there will be any issues.

rcaudy:
I reviewed Devin's changes. No complaints.

PR #3113, opened 2022-11-28
Bump to gradle 7.6

devinrsmith:
Groovy compilation is stricter now, failing out with "[Static type checking]" on places where generics were being used. Explicitly casting fixes the issue.

Solved for issue incorrect usage of `@Input` annotation, see https://docs.gradle.org/7.6/userguide/validation_problems.html#incorrect_use_of_input_annotation


devinrsmith:
Will merge if nightlies pass.

PR #3114, opened 2022-11-28
Jedi Autocomplete

JamesXNelson:


docker/server/src/main/docker/Dockerfile
```diff
@@ -10,12 +10,12 @@ COPY licenses/ /
...
 RUN set -eux; \
-    python -m pip install -q --no-index --no-cache-dir -r requirements.txt; \
+    python -m pip install -q --no-cache-dir -r requirements.txt; \
     rm requirements.txt
 
 COPY wheels/ /wheels
 RUN set -eux; \
-    python -m pip install -q --no-index --no-cache-dir /wheels/*.whl; \
+    python -m pip install -q --no-cache-dir /wheels/*.whl; \
     rm -r /wheels
```
JamesXNelson, line 19: need to remove this once base images are merged w/ jedi in them

py/server/deephaven/completer/__init__.py
```diff
@@ -0,0 +1,95 @@
+#
+# Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+#
+
+""" This module allows the user to configure if and how we use jedi to perform autocompletion.
+See https://github.com/davidhalter/jedi for information on jedi.
+
+# To disable autocompletion
+from deephaven.completer import CompleterSettings, CompleterMode
+CompletionSettings.mode = 'off'
```
jmao-denver, line 10: Should it be jedi_settings?

py/server/deephaven/completer/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+    off = 'off'
+    safe = 'safe'
+    strong = 'strong'
+
+    def __str__(self) -> str:
+        return self.value
+
+
+class Completer(object):
+
```
jmao-denver, line 32: This seems to be a 'private' class for creating a singleton, if so, change it to _Completer ?

py/server/deephaven/completer/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+        return self.__mode != CompleterMode.off
+
+    def can_jedi(self) -> bool:
+        return self.__can_jedi
+
+
+jedi_settings = Completer()
+
+if jedi_settings.mode == CompleterMode.off:
+    print('No jedi library installed on system path, disabling autocomplete')
```
jmao-denver, line 86: Is this safe? Does it go into the server log?

py/server/deephaven/completer/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+
+if jedi_settings.mode == CompleterMode.strong:
+    # start a strong-mode completer thread in bg
+    pass
+
```
jmao-denver, line 91:
```suggestion
```

jmao-denver (changes requested):

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
```diff
@@ -0,0 +1,222 @@
...
+                        .build();
+
+                safelyExecuteLocked(responseObserver,
+                        () -> responseObserver.onNext(AutoCompleteResponse.newBuilder()
+                                .setCompletionItems(builtItems)
+                                .build()));
+                String totalNano = Long.toString(System.nanoTime() - startNano);
+                // lets track how long completions take, as it's known that some
+                // modules like numpy can cause slow completion, and we'll want to know what was causing them
+                log.info().append("Jedi completions took ")
```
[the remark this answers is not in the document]
niloc132: also, logging should be pulled out of here, since we probably a) don't want to count IO in the jedi time, and b) don't want to fail to log time if IO blew up

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
```diff
@@ -0,0 +1,222 @@
...
+                                totalNano.length() > 6 ?
+                                        totalNano.substring(0, totalNano.length() - 6) :
+                                        "0"
+                        )
+                        .append('.')
+                        .append(totalNano.substring(6, Math.min(8, totalNano.length())))
+                        .append("ms").endl();
+
+            } catch (Throwable e) {
+                log.error().append("Jedi completion failure ").append(e).endl();
```
niloc132, line 191: this catch should notify client of failure too?

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
```diff
@@ -0,0 +1,222 @@
...
+                                    StreamObserver<AutoCompleteResponse> responseObserver) {
+        final ScriptSession scriptSession = exportedConsole.get();
+        try (final SafeCloseable ignored = scriptSession.getExecutionContext().open()) {
+            final VariableProvider vars = scriptSession.getVariableProvider();
+
+            PyObject completer = (PyObject) scriptSession.getVariable("jedi_settings");
+            boolean canJedi = completer.callMethod("is_enabled").getBooleanValue();
+            if (!canJedi) {
+                log.trace().append("Ignoring completion request because jedi is disabled").endl();
+                return;
```
niloc132, line 113: should respond with no completions

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
```diff
@@ -0,0 +1,222 @@
...
+            }
+            final VersionedTextDocumentIdentifier doc = request.getTextDocument();
+            final long startNano = System.nanoTime();
+            String text = completer.call("get_doc", doc.getUri()).getStringValue();
+            try {
+                final Position pos = request.getPosition();
+                String completionVar = "completions";
+                scriptSession.setVariable("__jedi_source__", text);
```
niloc132, line 121: that's unfortunate... but i guess not surprising at the end of the day.

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
```diff
@@ -0,0 +1,222 @@
...
+                final ScriptSession.Changes changes = scriptSession.evaluateScript(
+                        completionVar + " = jedi.Interpreter(jedi_settings.get_doc('" + doc.getUri() + "'), [globals()]).complete(" +
+                                (pos.getLine() + 1) + "," + pos.getCharacter() + ")"
+                );
+                final PyListWrapper completes = vars.getVariable(completionVar, null);
+                List<CompletionItem.Builder> completionResults = new ArrayList<>();
+                List<CompletionItem.Builder> completionResults_ = new ArrayList<>();
+                List<CompletionItem.Builder> completionResults__ = new ArrayList<>();
+                for (PyObject completion : completes) {
+                    String completionName = completion.getAttribute("name").getStringValue();
```
niloc132, line 131: future cleanup, i don't see an easy/obvious way to address this right away: with so many calls into and out of python, we're going to be taking and releasing the GIL a lot, plus N JNI calls per completion item (and i dont see anything obvious that limits the results passed back to java) though i'm more worried about the GIL. 

Instead, it would be nice if we could translate each item into a dict while in python, then hijack the copyPythonDictToJavaMap or the like to translate the data to java before giving up the GIL to begin with.

Note that high GIL contention is also going to throw off the perf numbers collected below - seems likely we should measure just the .complete() call above, and possibly separately measure the marshalling, IO. I think that's the only action item I'd have about this...

niloc132: Another option for GIL management is to use the new PyLib.ensureGil(() -> {...}); wrapped around all of the code that actually reads and builds results.

py/server/deephaven/completer/__init__.py
JamesXNelson: Yes. fixing w:

```
# To disable autocompletion
from deephaven.completer import jedi_settings
jedi_settings.mode = 'off'
```

py/server/deephaven/completer/__init__.py
JamesXNelson: Java actually holds onto this object and calls into its methods, but the consumer would probably only care about `jedi_settings` and not `Completer`.  Will rename it.

py/server/deephaven/completer/__init__.py
JamesXNelson: I should probably just move the logging out of python and into the calling code's logger

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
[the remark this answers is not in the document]
JamesXNelson: cleaned up the logging and set it to TRACE.

Also, if I'm only logging > 100ms, I can probably make stronger guesses about string length and simplify the toString ugliness too...
Also, where should I move logging to make it not-count?
Also, who is counting / cares about how long this method takes?

I did the logging part after the work part completed, so if there is a better location / way to do this, you'll have to point me at it explicitly.

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
JamesXNelson: just removed the catch. I don't think I need to eat Throwable here, just let outer catch have it

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
JamesXNelson: done

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
JamesXNelson: I actually got about 85% of the way to moving it all into python.  I'll just finish that and delete this mess. I knew crossing the gil like that was _terrible_, I just did it as proof of concept.  The whole point of the python module in this PR is to move the result processing into python, and just send back a glob (arrays of arrays) of strings and ints for java to process all at once.

I'll wrap that up since it's close and this, as-is, is terrible.

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
```diff
@@ -0,0 +1,234 @@
...
+        final ScriptSession scriptSession = exportedConsole.get();
+        try (final SafeCloseable ignored = scriptSession.getExecutionContext().open()) {
+
+            PyObject completer = (PyObject) scriptSession.getVariable("jedi_settings");
+            boolean canJedi = completer.callMethod("is_enabled").getBooleanValue();
+            if (!canJedi) {
+                log.trace().append("Ignoring completion request because jedi is disabled").endl();
+                // send back an empty response. We may want to include a "don't ask again" flag in the response...
+                safelyExecuteLocked(responseObserver,
+                        () -> responseObserver.onNext(AutoCompleteResponse.newBuilder().build()));
```
niloc132, line 111: need to have the completed items response in here, with success=false, and requestId for the client to match the failure to

docker/server-jetty/build.gradle
```diff
@@ -10,6 +10,7 @@ def targetArch = Architecture.targetArchitecture(project)
 
 def baseMapAmd64 = [
         'server-base': 'server-jetty',
+        'all-ai-base': 'server-all-ai-jetty',
```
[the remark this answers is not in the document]
JamesXNelson: I'm going to leave it in unless someone complains.  Just makes doing deployments that build deephaven-core easier / possible.

py/server/deephaven/completer/__init__.py
JamesXNelson: It goes to the console log in the gui

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
```diff
@@ -0,0 +1,234 @@
...
+            for (PyObject result : results.asList()) {
+                if (!result.isList()) {
+                    throw new UnsupportedOperationException("Expected list-of-lists from jedi_settings.do_completion, " +
+                            "got bad result " + result.call("repr") + " from full results: " + results.call("repr"));
+                }
+                // we expect [ "completion text", start_column ] as our result.
+                // in the future we may want to get more interesting info from jedi to pass back to client
+                final List<PyObject> items = result.asList();
+                String completionName = items.get(0).getStringValue();
+                int start = items.get(1).getIntValue();
```
niloc132, line 146: Hmm. I think we're still paying for each GIL trip, to do the getStringValue, getIntValue calls. Plus, each List.get call is a python `__getitem__` invoke.

I still think we can punt on this, but will want to revisit

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
JamesXNelson: This was completed, you can see the brains in `completer/__init__.py`

py/server/deephaven/completer/__init__.py
JamesXNelson: Still need direction on this.

Also, should I move it to its own file, or leave things in `__init__.py`.  Most others, init seems very sparse, if you'd rather structure it differently, please lmk

jmao-denver: If we don't think there will likely be more modules in the completer sub-package, I would just have _completer.py and keep the class name as it is.

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
JamesXNelson: strange that it wouldn't just say "sending a list of primitives, lets do this all at once".

Maybe if I sent two lists, one of string, one of int, I could convince jpy to give me String[], int[] in two shots...
But, yes... for another day. I'll open some issues to track next steps.

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
JamesXNelson: fixed

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
JamesXNelson: I've confirmed the timing is as you suspect, and will open a ticket w/ an idea how to improve it.

py/server/deephaven/completer/__init__.py
JamesXNelson: done. I just deleted it since there was a similar method already printing effectively the same message to java logs.

py/server/deephaven/completer/__init__.py
JamesXNelson: done

devinrsmith (changes requested):
`py/server/setup.py` needs to be updated with the appropriate jedi dependency.

I'm not seeing method completion like I would expect to see. Not sure if I'm doing something wrong.

devinrsmith:
I've got some suggestions from https://github.com/JamesXNelson/deephaven-core/pull/9

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -267,162 +259,23 @@ public StreamObserver<AutoCompleteRequest> autoCompleteStream(
         return GrpcUtil.rpcWrapper(log, responseObserver, () -> {
             final SessionState session = sessionService.getCurrentSession();
             if (PythonDeephavenSession.SCRIPT_TYPE.equals(scriptSessionProvider.get().scriptType())) {
-                return new NoopAutoCompleteObserver(responseObserver);
+                PyObject[] settings = new PyObject[1];
+                safelyExecute(() -> {
+                    final ScriptSession scriptSession = scriptSessionProvider.get();
+                    scriptSession.evaluateScript("from deephaven.completer import jedi_settings");
+                    settings[0] = (PyObject) scriptSession.getVariable("jedi_settings");
```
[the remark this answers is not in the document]
JamesXNelson: As we discussed offline, your patch which handles this is ideal, but don't wanna pull in too much when we're this close to the finish line.

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
```diff
@@ -0,0 +1,240 @@
...
+                final TextDocumentItem doc = openDoc.getTextDocument();
+                PyObject completer = (PyObject) scriptSession.get().getVariable("jedi_settings");
+                completer.callMethod("open_doc", doc.getText(), doc.getUri(), doc.getVersion());
+                break;
+            }
+            case CHANGE_DOCUMENT: {
+                ChangeDocumentRequest request = value.getChangeDocument();
+                final VersionedTextDocumentIdentifier text = request.getTextDocument();
+
+                PyObject completer = (PyObject) scriptSession.get().getVariable("jedi_settings");
```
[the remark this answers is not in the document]
JamesXNelson: The canonical location is `deephaven.completer.jedi_settings`, but how I'm accessing it from the scope here IS dangerous / bad.  Your work to wrap it in a proxy object looks _way_ better.

py/server/deephaven/completer/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+off: do not use any autocomplete
+safe mode: uses static analysis of source files. Can't execute any code.
+strong mode: looks in your globals() for answers to autocomplete and analyzes your runtime python objects
+later, we may add slow mode, which uses both static and interpreted completion modes.
+"""
+
+from deephaven.completer._completer import Completer
+from jedi import preload_module, Interpreter
+
+jedi_settings = Completer()
```
chipkent, line 22: Needs sphinx doc string.

py/server/deephaven/completer/__init__.py
```diff
@@ -0,0 +1,25 @@
+#
```
chipkent, line 1: If this needs to be public, I would prefer it is called `autocomplete`, since that is more clear to the user looking at the package.

py/server/deephaven/completer/_completer.py
```diff
@@ -0,0 +1,111 @@
...
+from __future__ import annotations
+from enum import Enum
+from typing import Any
+from jedi import Interpreter, Script
+
+
+class CompleterMode(Enum):
+    off = 'off'
+    safe = 'safe'
+    strong = 'strong'
```
chipkent, line 11: I assume these should be caps, since they are enum values.

py/server/deephaven/completer/_completer.py
```diff
@@ -0,0 +1,111 @@
+# only python 3.8 needs this, but it must be the first expression in the file, so we can't predicate it
+from __future__ import annotations
+from enum import Enum
+from typing import Any
+from jedi import Interpreter, Script
+
+
+class CompleterMode(Enum):
```
chipkent, line 8: Everything here needs appropriate docs.

chipkent (changes requested):

devinrsmith (approved):
Couple of follow ups we can handle in another PR:

1. config to disable auto-complete
1. move code to deephaven_internal
1. add autocomplete dependency for deephaven-server

chipkent (approved):
Fix my comments in future commits

jmao-denver (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/1930
Conceptual: https://github.com/deephaven/deephaven.io/issues/1932
Reference: https://github.com/deephaven/deephaven.io/issues/1933
Blog: https://github.com/deephaven/deephaven.io/issues/1931

Issue #2985, opened 2022-10-11
Some conditions can result in JS API breaking rather than emitting an error


niloc132:
We don't have precise steps to reproduce this, but the failure mode probably is that the server ends a session (likely due to lack of refresh from the client), then the client receives an unexpected message on the TerminationNotification stream which it attempts to handle as a positive "the server is shutting down, here's the error". Instead, there is no message payload, and a null pointer exception occurs. 

Client error:
```
2022-10-10T15:38:09.590Z U      TypeError: Cannot read properties of null (reading 'getAbnormalTermination')    Uncaught TypeError: Cannot read properties of null (reading 'getAbnormalTermination')
        at hle (http://localhost:10000/ide/assets/vendor.222459d5.js:88:6696)
        at JP.dispatchEvent (http://localhost:10000/ide/assets/vendor.222459d5.js:88:8340)
        at Mg.handleUncaughtError (http://localhost:10000/ide/assets/index.d662c235.js:1:6978)
```
Server log at this time:
```
deephaven-server-1    | 2022-10-10T15:37:36.678334300Z 2022-10-10T15:37:36.677Z | ESC[32mEFAULT.refreshThreadESC[0;39m | ESC[34m INFOESC[0;39m | ESC[33m.e.u.UpdateGraphProcessorESC[0;39m | Minimal Update Graph Processor cycle times: 25.029ms / 39 cycles = 0.642ms/cycle average), safePointTime=0ms
deephaven-server-1    | 2022-10-10T15:38:09.688597900Z 2022-10-10T15:38:09.688Z | ESC[32mheduler-Concurrent-1ESC[0;39m | ESC[34m INFOESC[0;39m | ESC[33mi.d.s.s.SessionState     ESC[0;39m | SessionState{b28fcf14-3c2e-4e90-8de0-cc6a9574d1ed}: releasing outstanding exports
deephaven-server-1    | 2022-10-10T15:38:09.689732600Z 2022-10-10T15:38:09.689Z | ESC[32mheduler-Concurrent-1ESC[0;39m | ESC[34m INFOESC[0;39m | ESC[33mi.d.s.s.SessionState     ESC[0;39m | SessionState{b28fcf14-3c2e-4e90-8de0-cc6a9574d1ed}: outstanding exports released
[...]

PR #3123, opened 2022-11-29
Improve Jetty/gRPC shutdown

niloc132:
A series of four commits to address #2985 and other adjacent issues. I expect it will make more sense to merge these together, but providing a single branch might help review and testing. 

These commits seek to ensure that the gRPC server shuts down in a predictable way from the server application's perspective and from any gRPC client's perspective.

The first commit addresses a bug where a client doesn't return early enough when reporting an error the first time, and will report a second error.

The second commit reverts some earlier cleanup around the HealthCheck service and offers an alternative solution to cleanly shut down gRPC's own threadpool only once shutdown is complete, instead of right away.  (This has been discussed with James, who originally wrote that patch)

The third commit correctly ends any autocomplete stream when a session ends - among other things, this will enable the server to halt those streams during server shutdown.

Finally, the biggest commit, which likely should land separately, modifies both our "multiplexed" websocket format to support a "GO_AWAY" message (attempting to reflect h2 semantics), and reorders Jetty shutdown to give client's a change to handle a message. Normal h2 streams in Jetty are notified that the socket will end soon, and are given a chance to finish their work, but websockets are immediately closed - this change avoids closing those websockets until later in the shutdown process, giving browser clients a chance to finish their work.


grpc-java/grpc-servlet-websocket-jakarta/src/main/java/io/grpc/servlet/web/websocket/MultiplexedWebSocketServerStream.java
```diff
@@ -54,12 +67,55 @@ public class MultiplexedWebSocketServerStream extends AbstractWebSocketServerStr
...
+     * Enum to describe the process of closing a transport. After the server has begun to close but the client hasn't
+     * yet acknowledged, it is permitted for the client to start a new stream, but after the server acknowledges, no new
+     * streams can be started. Note that shutdown will proceed anyway, and will eventually stop all streams.
+     */
+    enum ClosedState {
+        OPEN, CLOSING, CLOSED
+    }
+
+    private ClosedState closed = ClosedState.OPEN;
+    private CompletableFuture<Void> closingFuture = new CompletableFuture<>();
```
devinrsmith, line 80: Can be final?

grpc-java/grpc-servlet-websocket-jakarta/src/main/java/io/grpc/servlet/web/websocket/MultiplexedWebSocketServerStream.java
```diff
@@ -41,6 +44,16 @@
  * runOnTransportThread while in onMessage, as we're already in the transport thread.
  */
 public class MultiplexedWebSocketServerStream extends AbstractWebSocketServerStream {
+    public static final String GRACEFUL_CLOSE = MultiplexedWebSocketServerStream.class.getName() + ".graceful_close";
+
+    /**
+     * Callback to initiate a gradeful shutdown of this websocket instance, as an alternative to just closing the
```
devinrsmith, line 50: graceful

grpc-java/grpc-servlet-websocket-jakarta/src/main/java/io/grpc/servlet/web/websocket/MultiplexedWebSocketServerStream.java
```diff
@@ -54,12 +67,55 @@ public class MultiplexedWebSocketServerStream extends AbstractWebSocketServerStr
...
+     * a stream of Integer.MAX_INTEGER to the client to signal that new requests will not be accepted, and future
+     * incoming streams will be closed by the server right away. In keeping with h2, until the client ACKs the close, we
+     * will permit incoming streams that are were sent before we closed, but they likely will not have a large window to
+     * get their work done before they are closed the rest of the way.
+     */
+    private CompletableFuture<Void> stopAcceptingNewStreams() {
+        if (closed != ClosedState.OPEN) {
+            return closingFuture;
+        }
+        closed = ClosedState.CLOSING;
```
devinrsmith, line 99: Should `closed` technically be marked as `volatile`? Even though we expect `closed` to go OPEN -> CLOSING -> CLOSED with happens-before/happens-after semantics, I don't think the java memory model guarantees other threads will see the change?

web/client-api/src/main/java/io/deephaven/web/client/api/grpc/MultiplexedWebsocketTransport.java
```diff
@@ -121,9 +121,77 @@ public void sendFallback(Transport transport) {
...
+        private void retain() {
+            activeCount++;
+        }
+
+        private void release() {
+            activeCount--;
+            if (activeCount == 0 && closing) {
+                webSocket.close();
+            }
+        }
```
devinrsmith, line 191: Are there any threading concerns we have to worry about? I'm assuming this is a different sync model wrt JS?

grpc-java/grpc-servlet-websocket-jakarta/src/main/java/io/grpc/servlet/web/websocket/MultiplexedWebSocketServerStream.java
```diff
@@ -54,12 +67,55 @@ public class MultiplexedWebSocketServerStream extends AbstractWebSocketServerStr
...
             List<? extends ServerStreamTracer.Factory> streamTracerFactories, int maxInboundMessageSize,
             Attributes attributes) {
         super(transportListener, streamTracerFactories, maxInboundMessageSize, attributes);
     }
 
+    /**
+     * Stops this multiplexed transport from accepting new streams. Instead, it will reply with its version of GO_AWAY,
+     * a stream of Integer.MAX_INTEGER to the client to signal that new requests will not be accepted, and future
+     * incoming streams will be closed by the server right away. In keeping with h2, until the client ACKs the close, we
+     * will permit incoming streams that are were sent before we closed, but they likely will not have a large window to
```
nbauernfeind, line 92: s/are were/were

web/client-api/src/main/java/io/deephaven/web/client/api/grpc/MultiplexedWebsocketTransport.java
niloc132: No threads on JS, correct, various java keywords have no effect when transpiled to js, as they have no js analog.

grpc-java/grpc-servlet-websocket-jakarta/src/main/java/io/grpc/servlet/web/websocket/MultiplexedWebSocketServerStream.java
niloc132: Ah, I thought I had this set to only be possible from one thread, but I will double check and get back to you, or at least try to tell a convincing story...

niloc132: Yes, that could be helpful, but I think not strictly necessary - `stopAcceptingNewStreams()` can only be called once, and the other tests require a full round trip to the client after that to make sense. I'm not sure what the threading "guarantees" there would be, except that by the time we've sent a message and received a corresponding reply, that change should have propagated. 

The exception to this is a newly opened socket, but I think that onOpen is already racing WebsocketFactory's loop in that case. Will investigate.

Open to adding it anyway, but in that case we have many more fields that should h

grpc-java/grpc-servlet-websocket-jakarta/src/main/java/io/grpc/servlet/web/websocket/MultiplexedWebSocketServerStream.java
```diff
@@ -104,6 +177,14 @@ public void onMessage(ByteBuffer message) throws IOException {
 
         // if this is the first message on this websocket, it is the request headers
         if (stream == null) {
+            if (this.closed == ClosedState.CLOSED) {
+                // Not accepting more calls, client sent this before they saw that, be clear that this is closed
```
niloc132, line 181: this comment is wrong.

In this `if`, we know that the client is aware that they shouldnt open new streams but did it anyway, so we are closing them.

If closed == CLOSING, the client doesn't yet know, so like h2 GO_AWAY, we will permit them to start this new stream - they'll know very soon though, and will have a short window to wrap it up (and, in all likelihood, their session has already ended, so they may not be able to do what they want to do anyway, but that's an application-level concern, not a transport issue).

devinrsmith (approved):
Looks good. Were you able to get a testing procedure that makes you relatively confident in these changes? Not sure if you preferred to separate out the commits or not. I'm fine either way.

niloc132:
> Looks good. Were you able to get a testing procedure that makes you relatively confident in these changes? 

The testing process for the original bug was to open more than one browser/tab/window, and then shutdown the server, ideally with `kill -SIGTERM` (ctrl-c to gradle tends to kill it without giving it time to shutdown cleanly?), ensuring that all received the shutdown message properly.

To ensure that GO_AWAY behaved as expected, I would put a breakpoint in the client implementation of the websocket, and prevent it from acknowledging the message, to ensure that the server would stay up until its timeout. It appears that in some cases the browser will hold open other connections (likely keep-alive for plain http/1.1 calls), and the server waits for those to all terminate as well - if the client has been running for more than a few minutes or has been closed, this doesn't happen.

> Not sure if you preferred to separate out the commits or not. I'm fine either way.

If you don't see an issue with landing them together, I can at least keep the branch in my fork so we can roll back individual commits if it comes to that, else I can submit several small PRs to land one at a time. For the sake of simplicity, given how long until we release I lean towards merging this PR instead, but it isn't a strong preference.

nbauernfeind (approved):

devinrsmith (dismissed):

devinrsmith (approved):

nbauernfeind (approved):

PR #3126, opened 2022-11-30
Improve BigInteger handling in Parquet files (Closes #3125)

### LINKED ISSUE #3125: Improve BigDecimal/BigInteger handling in Parquet
When computing PrecisionAndScale for BigDecimal values, invalid parameters are generated.  BigInteger can be efficiently encoded as a DecimalLogicalType using a fixed precision of 1 and scale of 0 using a variant of the BigDecimalParquetBytesCodec

abaranec:
BigIntegers can be encoded as DecimalLogicalType with a precision of 1 and scale of 0 using a variant of BigDecimalParquetBytesCodec.


jcferretti (dismissed):

engine/table/src/main/java/io/deephaven/engine/table/impl/CodecLookup.java
```diff
@@ -76,7 +77,8 @@ private static boolean noCodecRequired(@NotNull final Class<?> dataType) {
                 // appropriate precision and scale calculated from column data,
                 // unless the user explicitly requested something else
                 // via instructions.
-                dataType == BigDecimal.class;
+                dataType == BigDecimal.class ||
+                dataType == BigInteger.class;
```
rcaudy, line 81: You might need/want to add a comment here like the one for `BigDecimal`.

rcaudy (commented):
I didn't really review this, just glanced at the shape of the work.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetSchemaReader.java
```diff
@@ -327,6 +328,9 @@ public Optional<Class<?>> visit(final LogicalTypeAnnotation.EnumLogicalTypeAnnot
             @Override
             public Optional<Class<?>> visit(
                     final LogicalTypeAnnotation.DecimalLogicalTypeAnnotation decimalLogicalType) {
+                if (decimalLogicalType.getPrecision() == 1 && decimalLogicalType.getScale() == 0) {
```
jcferretti, line 331: Can we add a comment here that the 1 is a "marker" value that we ourselves set up before?
An unsuspecting reader (me, in the past, before getting an explanation) may ask themselves "why is it that the precision would be exactly 1?"

engine/table/src/main/java/io/deephaven/engine/util/BigDecimalUtils.java
```diff
@@ -62,18 +64,23 @@ public static PrecisionAndScale computePrecisionAndScale(
...
+        if (rowSet.isEmpty()) {
+            return EMPTY_TABLE_PRECISION_AND_SCALE;
+        }
+
         // we first compute max(precision - scale) and max(scale), which corresponds to
         // max(digits left of the decimal point), max(digits right of the decimal point).
         // Then we convert to (precision, scale) before returning.
         int maxPrecisionMinusScale = 0;
         int maxScale = 0;
-        try (final ChunkSource.GetContext context = source.makeGetContext(sz);
```
jcferretti, line 71: Can we modify the block comment right before this point to make it explicit that we are not reading the whole table, just making an estimate based on the first PRECISION_SCALE_N_TO_INSPECT values?

jcferretti (changes requested):

engine/table/src/main/java/io/deephaven/engine/util/BigDecimalUtils.java
```diff
@@ -71,27 +73,37 @@ public static PrecisionAndScale computePrecisionAndScale(
         // We first use the initial PRECISION_SCALE_N_TO_INSPECT values to estimate the max(precision - scale) and
         // max(scale), which corresponds to max(digits left of the decimal point), max(digits right of the decimal
         // point). Then we convert to (precision, scale) before returning.
-        int maxPrecisionMinusScale = 0;
-        int maxScale = 0;
+        int maxPrecisionMinusScale = -1;
+        int maxScale = -1;
         try (final ChunkSource.GetContext context = source.makeGetContext(PRECISION_SCALE_N_TO_INSPECT);
```
jcferretti, line 78: Rename PRECISION_SCALE_N_TO_INSPECT ?





[1:41](https://illumon.slack.com/archives/C04DJDAM0CT/p1670006500747539)
Now is more of a working buffer size than a limit

jcferretti (changes requested):

jcferretti (approved):

jcferretti (approved):

PR #3140, opened 2022-12-02
Add dependency on autocomplete feature for PyPi deephaven-server

devinrsmith:
Follow-up to #3114


chipkent (approved):

niloc132 (approved):
Approved. I'd rather optional dependencies remain optional, but this seems like more trouble than it is worth at this time.

PR #3143, opened 2022-12-05
TableReplayer fix

jjbrosnan:
Fixes double-notify assertion failure when Replayer time window has no overlap with times in historical table

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/replay/ReplayTable.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/replay/ReplayTable.java
@@ -64,7 +64,8 @@ public class ReplayTable extends QueryTable implements Runnable {
             nextTimeNanos = nanoTimeSource.getLong(nextRowKey);
             if (nextTimeNanos == QueryConstants.NULL_LONG || nextTimeNanos < currentTimeNanos) {
                 throw new RuntimeException(
-                    "The historical table contains a null or decreasing time value at row number " + nextRowKey + ".");
+                        "The historical table contains a null or decreasing time value at row number " + nextRowKey
+                                + ".");
             }
         } else {
             // NB: It would be best to ensure that if this is hit during construction, we're never added to the UGP.
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/replay/ReplayTable.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/replay/ReplayTable.java
@@ -15,12 +15,15 @@ import org.jetbrains.annotations.NotNull;
 import java.util.Map;
 
 public class ReplayTable extends QueryTable implements Runnable {
-
+    /**
+     * Creates a new ReplayTable based on a row set, set of column sources, time column, and a replayer
+     */
     private final Replayer replayer;
     private final ColumnSource<Long> nanoTimeSource;
     private final RowSet.Iterator rowSetIterator;
 
     private long nextRowKey = RowSequence.NULL_ROW_KEY;
+    private long currentTimeNanos = QueryConstants.NULL_LONG;
     private long nextTimeNanos = QueryConstants.NULL_LONG;
     private boolean done;
 
@@ -49,10 +52,20 @@ public class ReplayTable extends QueryTable implements Runnable {
         }
     }
 
+    /**
+     * Advance the row key and time iterators if there are any left in the table.
+     * 
+     * @throws RuntimeException if time is null, or if the next time is before the current time.
+     */
     private void advanceIterators() {
         if (rowSetIterator.hasNext()) {
             nextRowKey = rowSetIterator.nextLong();
+            currentTimeNanos = nextTimeNanos;
             nextTimeNanos = nanoTimeSource.getLong(nextRowKey);
+            if (nextTimeNanos == QueryConstants.NULL_LONG || nextTimeNanos < currentTimeNanos) {
+                throw new RuntimeException(
+                    "The historical table contains a null or decreasing time value at row number " + nextRowKey + ".");
+            }
         } else {
             // NB: It would be best to ensure that if this is hit during construction, we're never added to the UGP.
             // If this is hit during update processing, it would be great to remove ourselves from the UGP right away.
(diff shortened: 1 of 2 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/replay/Replayer.java
```diff
@@ -186,6 +186,9 @@ public void setTime(long updatedTime) {
      */
     @Override
     public Table replay(Table dataSource, String timeColumn) {
+        if (dataSource.isRefreshing()) {
+            dataSource = TableTools.emptyTable(0).snapshot(dataSource);
+        }
```
chipkent, line 191: @rcaudy This is an interesting case.  Are you ok with this limitation, or is there a way to use a live table?

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/ReplayTable.java
```diff
@@ -4,69 +4,83 @@
...
     private boolean done;
-    private final Replayer replayer;
 
-    public ReplayTable(RowSet rowSet, Map<String, ? extends ColumnSource<?>> result, String timeColumn,
-            Replayer replayer) {
-        super(RowSetFactory.empty().toTracking(), result);
-        Require.requirement(replayer != null, "replayer != null");
-        // noinspection unchecked
-        replayer.registerTimeSource(rowSet, (ColumnSource<DateTime>) result.get(timeColumn));
+    public ReplayTable(
```
chipkent, line 27: Add javadoc, since we are making a change?

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/ReplayTable.java
```diff
@@ -4,69 +4,83 @@
...
+            @NotNull final RowSet rowSet,
+            @NotNull final Map<String, ? extends ColumnSource<?>> columns,
+            @NotNull final String timeColumn,
+            @NotNull final Replayer replayer) {
+        super(RowSetFactory.empty().toTracking(), columns);
+        this.replayer = Require.neqNull(replayer, "replayer");
+        // NB: This will behave incorrectly if our row set or any data in columns can change. Our source table *must*
+        // be static. We also seem to be assuming that timeSource has no null values in rowSet. It would be nice to use
+        // a column iterator for this, but that would upset unit tests by keeping pooled chunks across cycles.
```
chipkent, line 36: 1) We don't want initials in comments or code.
2) The comment implies that we have a requirement that the input table is static.  Can we assert that here from the column sources?
3) @rcaudy How big of a deal is the column iterator comment?
4) See other comments about null times.  We should raise an error on null times.
5) I don't think we are asserting the time order.  We should also raise an error on out-of-order times, if we do not support them.

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/ReplayTable.java
```diff
@@ -4,69 +4,83 @@
...
+        }
+    }
+
+    private void advanceIterators() {
+        if (rowSetIterator.hasNext()) {
+            nextRowKey = rowSetIterator.nextLong();
+            nextTimeNanos = nanoTimeSource.getLong(nextRowKey);
         } else {
+            // NB: It would be best to ensure that if this is hit during construction, we're never added to the UGP.
+            // If this is hit during update processing, it would be great to remove ourselves from the UGP right away.
```
chipkent, line 58: 1) Initials in comments
2) Should this change be made?  @rcaudy

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/ReplayTable.java
```diff
@@ -4,69 +4,83 @@
...
-            }
-        }
-        final RowSet added = indexBuilder.build();
-        if (added.size() > 0) {
+        final RowSet added = advanceToCurrentTime();
+        if (added.isNonempty()) {
             getRowSet().writableCast().insert(added);
             notifyListeners(added, RowSetFactory.empty(), RowSetFactory.empty());
+        } else {
+            added.close();
```
chipkent, line 83: @rcaudy does `added` need to be closed in the non-empty case?

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/Replayer.java
chipkent: Other comments sound like we are stuck with this limitation for the time being.

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/ReplayTable.java
```diff
@@ -4,69 +4,83 @@
...
+            try (final RowSet initial = advanceToCurrentTime()) {
+                getRowSet().writableCast().insert(initial);
+            }
+        }
+    }
+
+    private void advanceIterators() {
+        if (rowSetIterator.hasNext()) {
+            nextRowKey = rowSetIterator.nextLong();
+            nextTimeNanos = nanoTimeSource.getLong(nextRowKey);
```
chipkent, line 55: should be doing something (like puking) when the time is null.

chipkent (changes requested):

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/ReplayTable.java
rcaudy: This change can't be made without a refactoring of the Replayer.

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/Replayer.java
rcaudy: > @rcaudy This is an interesting case.  Are you ok with this limitation, or is there a way to use a live table?

I think this code is entirely the wrong pattern if we wanted to have general updating tables as input. There are other tools for that. I think it's a fine limitation for replay to have.

engine/table/src/main/java/io/deephaven/engine/table/impl/replay/ReplayTable.java
rcaudy: NB stands for Nota Bene. There are no initials here.

We can't tell anything about whether the row set or column sources can change from here. Would need the Table. The calling code handles it for now.

rcaudy: The column iterator comment is just an implementation note for myself or other devs. No big deal at all.

chipkent (changes requested):

PR #3152, opened 2022-12-06
All project types should declare maven repositories

niloc132:
This appears to fix some nightly build issues going back a month or two, as some non-java projects still depend on java artifacts, either through transitively through project dependencies, or through plugins like spotless that resolve dependencies at runtime.


devinrsmith (changes requested):
One quick thing, but can merge after.

devinrsmith (approved):

PR #3155, opened 2022-12-06
Close the inner context of a RegionedContextHolder if it was set

abaranec:


engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/RegionContextHolder.java
```diff
@@ -60,4 +61,11 @@ public <T> T getInnerContext() {
         // noinspection unchecked
         return (T) innerContext;
     }
+
+    @Override
+    public void close() {
+        if(innerContext != null) {
+            innerContext.close();
```
rcaudy, line 68: Do you want to null this out?

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/RegionContextHolder.java
```diff
@@ -28,7 +29,7 @@ public boolean supportsUnboundedFill() {
      * 
      * @param contextObject The context object
      */
-    public void setInnerContext(@Nullable final Object contextObject) {
+    public void setInnerContext(@Nullable final SafeCloseable contextObject) {
```
rcaudy, line 32: Don't we have a `Context` interface for this?

rcaudy (approved):

Issue #3133, opened 2022-11-30
Automate authorization codegen


nbauernfeind:
The authorization hooks were generated using `protoc` plugins that invoke the given generator main via a jar-jar. I did not automate this process during the initial feature build out.

This task is to automate this likely as part of the grpc code generation step.

PR #3156, opened 2022-12-06
Remove unused gRPC method for Pandas DataFrames

### LINKED ISSUE #3153: Remove FetchPandasTable from gRPC Table API
When working on authorization I noticed that FetchPandasTable is not implemented via the gRPC TableService API.
 
@niloc132 says that dataframes are fetched as widgets. These will go through the TicketResolverBase.AuthTransformation mechanism when fetched as an object.

It looks safe to drop this from the API.

niloc132:
This method was never implemented, so impact of removal should be minimal. It was superseded by FetchObject along with a plugin for Pandas, implemented in #1942.

Also updated other generated gRPC bindings that were out of date.

Fixes #3153


niloc132:
Failing due to needing to manually run authorization regen, #3133 will automate this.

nbauernfeind (approved):

nbauernfeind:
Looks like there are still references to this request type in the C++ build. Should we rope in Corey or Cristian to tell us what we need to do? Or do you think these are files that would no longer be generated and just need to be deleted? (Like the FetchPandasTableRequest.java that was still in the jsapi directory.)

nbauernfeind (approved):

devinrsmith (approved):

devinrsmith:
Note: `io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.table_pb.batchtablerequest.operation.OpCase` was not modified - is this expected?

devinrsmith (approved):

PR #3161, opened 2022-12-07
Take buildTableCreationResponse out of safelyExecutes

devinrsmith:


nbauernfeind (approved):

server/src/main/java/io/deephaven/server/table/TableServiceGrpcImpl.java
```diff
@@ -405,8 +403,12 @@ public void getExportedTableCreationResponse(final Ticket request,
                         }
                         authWiring.checkPermissionGetExportedTableCreationResponse(
                                 session.getAuthContext(), request, Collections.singletonList((Table) obj));
-                        responseObserver.onNext(ExportUtil.buildTableCreationResponse(request, (Table) obj));
-                        responseObserver.onCompleted();
```
devinrsmith, line 409: I'm assuming it was on oversight this wasn't in a safelyExecute?

niloc132: It doesn't hurt - the risk is that some work _after_ the call might fail to be performed, but since this is a nonExport() submit, nothing need to be returned and stored for future ticket references.

So we're just mitigating some log messages.

I don't love overly defensive calls like this - but I also really don't love the layers we've had to stack on gRPC to make it usable at all, so I'm good with it.

server/src/main/java/io/deephaven/server/table/TableServiceGrpcImpl.java
```diff
@@ -347,21 +347,19 @@ public void batch(final BatchTableRequest request,
...
-                                    .setResultId(resultId)
-                                    .setSuccess(false)
-                                    .setErrorInfo(finalErrorInfo)
-                                    .build()));
+                    final ExportedTableCreationResponse response = ExportedTableCreationResponse.newBuilder()
+                            .setResultId(resultId)
+                            .setSuccess(false)
+                            .setErrorInfo(errorInfo)
+                            .build();
+                    safelyExecuteLocked(responseObserver, () -> responseObserver.onNext(response));
```
niloc132, line 355: We've got so many of these... I almost want to just have a wrapper/subtype of StreamObserver that does these for us, like onCompleted(Response) to do both onNext and onCompleted safely+locked, and auto-wrap both onNext and onCompleted with the lock and safe wrapper.

niloc132 (approved):

PR #3174, opened 2022-12-09
Jedi cleanup and fixes

devinrsmith:
Uses PyObject proxy for more explicit methods.

Fixes state management, uses per-session instead of per stream state.

Fixes SAFE mode.

Related to #3173

### DIFF
--- a/py/server/deephaven_internal/auto_completer/__init__.py
+++ b/py/server/deephaven_internal/auto_completer/__init__.py
@@ -16,10 +16,10 @@ strong mode: looks in your globals() for answers to autocomplete and analyzes yo
 later, we may add slow mode, which uses both static and interpreted completion modes.
 """
 
-from ._completer import Completer, Mode
+
+from ._completer import Completer, Mode, jedi_settings
 from jedi import preload_module, Interpreter
 
-jedi_settings = Completer()
 # warm jedi up a little. We could probably off-thread this.
 preload_module("deephaven")
 Interpreter("", []).complete(1, 0)
--- a/server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
+++ b/server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
@@ -24,6 +24,7 @@ import io.deephaven.proto.backplane.grpc.Ticket;
 import io.deephaven.proto.backplane.grpc.TypedTicket;
 import io.deephaven.proto.backplane.script.grpc.*;
 import io.deephaven.server.console.completer.JavaAutoCompleteObserver;
+import io.deephaven.server.console.completer.JediCompleter;
 import io.deephaven.server.console.completer.PythonAutoCompleteObserver;
 import io.deephaven.server.session.SessionCloseableObserver;
 import io.deephaven.server.session.SessionService;
@@ -31,6 +32,7 @@ import io.deephaven.server.session.SessionState;
 import io.deephaven.server.session.SessionState.ExportBuilder;
 import io.deephaven.server.session.TicketRouter;
 import io.grpc.stub.StreamObserver;
+import org.jpy.PyModule;
 import org.jpy.PyObject;
 
 import javax.inject.Inject;
@@ -109,14 +111,13 @@ public class ConsoleServiceGrpcImpl extends ConsoleServiceGrpc.ConsoleServiceImp
                     .onError(responseObserver)
                     .submit(() -> {
                         final ScriptSession scriptSession = new DelegatingScriptSession(scriptSessionProvider.get());
-
+                        final StartConsoleResponse response = StartConsoleResponse.newBuilder()
+                                .setResultId(request.getResultId())
+                                .build();
                         safelyExecute(() -> {
-                            responseObserver.onNext(StartConsoleResponse.newBuilder()
-                                    .setResultId(request.getResultId())
-                                    .build());
+                            responseObserver.onNext(response);
                             responseObserver.onCompleted();
                         });
-
                         return scriptSession;
                     });
         });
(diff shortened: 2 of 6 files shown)


py/server/deephaven_internal/auto_completer/__init__.py
```diff
@@ -16,10 +16,10 @@
 later, we may add slow mode, which uses both static and interpreted completion modes.
 """
 
-from ._completer import Completer, Mode
+
+from ._completer import Completer, Mode, jedi_settings
 from jedi import preload_module, Interpreter
 
-jedi_settings = Completer()
```
JamesXNelson, line 22: just curious... why move the instance location? the object is designed such that if customer wants to turn off autocomplete, they need to access this instance to set the mode to off.  Just want to make sure that is clear when you are making decisions about where things should live.

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -255,25 +256,29 @@ public StreamObserver<AutoCompleteRequest> autoCompleteStream(
...
+                return new PythonAutoCompleteObserver(responseObserver, session, () -> jedi(scope));
             }
-
             return new JavaAutoCompleteObserver(session, responseObserver);
         });
     }
 
+    private static JediCompleter jedi(final PyObject scope) {
+        try (final PyModule pyModule = PyModule.importModule("deephaven_internal.auto_completer")) {
+            return pyModule.call("Completer", scope).createProxy(JediCompleter.class);
```
JamesXNelson, line 278: a code comment here would be nice

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -255,25 +256,29 @@ public StreamObserver<AutoCompleteRequest> autoCompleteStream(
...
-                return canJedi ? new PythonAutoCompleteObserver(responseObserver, scriptSessionProvider, session)
-                        : new NoopAutoCompleteObserver(session, responseObserver);
+            final ScriptSession scriptSession = scriptSessionProvider.get();
+            if (PythonDeephavenSession.SCRIPT_TYPE.equals(scriptSession.scriptType())) {
+                final PyObject scope = ((PythonDeephavenSession) scriptSession).scope().mainGlobals().unwrap();
+                // noinspection EmptyTryBlock,unused
+                try (final JediCompleter _unused = jedi(scope)) {
+                    // ensure we can create it
+                } catch (RuntimeException e) {
+                    log.debug(e).append("Unable to create autocomplete; is jedi installed?").endl();
```
JamesXNelson, line 266: will anybody ever see this?
if someone expects autocomplete but it doesn't work, I doubt they are going to start w/ debug logging, or even notice this amid the logspam..

JamesXNelson: if we raise the log level so it's easily noticed, we can include a message about how to disable autocomplete altogether.

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
```diff
@@ -33,12 +49,29 @@ public class PythonAutoCompleteObserver extends SessionCloseableObserver<AutoCom
      * We only log timing for completions that take longer than, currently, 100ms
      */
     private static final long HUNDRED_MS_IN_NS = 100_000_000;
-    private final Provider<ScriptSession> scriptSession;
 
-    public PythonAutoCompleteObserver(StreamObserver<AutoCompleteResponse> responseObserver,
-            Provider<ScriptSession> scriptSession, final SessionState session) {
+    /** Track parsers by their session state, to ensure each session has its own, singleton, parser */
+    private static final Map<SessionState, JediCompleter> parsers = Collections.synchronizedMap(new WeakHashMap<>());
```
JamesXNelson, line 54: extremely unsafe to do this.

jedi is NOT threadsafe and should NOT be accessed by more than one caller at a time.
I plan to eventually put in some kind of threading and locking that ensures we don't accidentally use it in a parallelized manner, and having only a single completer would help with that...

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
```diff
@@ -101,33 +127,18 @@ private void getCompletionItems(GetCompletionItemsRequest request,
...
-                log.trace().append("Ignoring completion request because jedi is disabled").endl();
-                // send back an empty, failed response...
-                safelyExecuteLocked(responseObserver,
-                        () -> responseObserver.onNext(AutoCompleteResponse.newBuilder()
-                                .setCompletionItems(GetCompletionItemsResponse.newBuilder()
-                                        .setSuccess(false)
-                                        .setRequestId(request.getRequestId()))
-                                .build()));
-                return;
-            }
```
JamesXNelson, line 117: what happens if client uses `jedi_settings.mode = 'OFF'` to disable autocomplete in their session?

mofojed (commented):
Don found an autocomplete bug the other day: https://github.com/deephaven/web-client-ui/issues/969

Essentially if the UI calls `closeDocument` without calling `openDocument` first, it kills the connection. While this is an error, it probably shouldn't kill the connection in that case, just throw an error for that issue.

mattrunyon:
@devinrsmith Any chance this could be cleaned up and pushed over the line soonish? I'd like to plumb param help and hover help through Jedi, but I think this should be merged first

devinrsmith:
@mattrunyon - there have been some other changes wrt autocomplete; I'll sync up w/ @niloc132 and @rcaudy to see how we should proceed.

mattrunyon:
@devinrsmith Are those changes merged or in review? Just want to know if there's anything else I should look out for when implementing the other LSP functions

devinrsmith:
The simpler-fixes were merged in (sometime after this PR was "abandoned" as more complex than needed for release). That said, I don't remember the full context, and if we want to revive some of the improvements from this PR, we'll need to prioritize. I'm not sure exactly how it fits in w/ LSP improvements.

mattrunyon:
I'm looking to add support of more request types. Would involve modifying the proto message to include more than just completion requests. And then modifying the Java/Python that wires the completion requests currently to expand to other requests.

I don't think this cleanup is needed for me to wire it through, it just might cause this cleanup to have more conflicts.

PR #3183, opened 2022-12-12
Fix incorrect Table casts of PartitionTables in PartitionedTableService

### LINKED ISSUE #3172: Plotting with a partition table causes an error
**Description**

Plotting with a partition table causes an error

**Steps to reproduce**

Run the following Python code

```python
import random
import string

letters_upp = string.ascii_uppercase

def get_random_string(length):
    # choose from all lowercase letter
    letters = string.ascii_letters
    result_str = ''.join(random.choice(letters) for i in range(length))
    return(result_str)

static_source_1 = empty_table(26300)\
    .update(['X = ii', 'Double1 = randomDouble(1, 100)', 'Int1 = randomInt(1, 100)', 'Timestamp = convertDateTime(`2016-01-01T01:00 NY`) + i * HOUR', 'Date = Timestamp.toDateString()', 'String1 = (java.lang.String)random.choice(letters_upp)', 'String2 = (java.lang.String)get_random_string(random.randint(1,4))'])

figure = Figure()
t_pb = empty_table(100).update_view(formulas = ["X=i", "Z=i%3", "Y=X*X + Z * 10000", "Series_Id = `series_` + Z"]).move_columns(idx=0, cols=["Series_Id"])

table_3_keys = static_source_1.head(2000).where("String1 in `A`, `B`, `C`")\
    .sort("Int1").update(["New_Row = i", "Y = Int1 * (String1 = `A` ? 2 : String1 = `B` ? 0.5 : 1.0)"])
p = figure.plot_xy(series_name="partitioned", t=table_3_keys, by=["String1"], x="New_Row", y="Y").show()
```

**Expected results**

Plot opens

**Actual results**

Plot does not open. Server side error

Error from the server (gets printed out 3 times in the browser)

```
java.lang.ClassCastException: class io.deephaven.engine.table.impl.partitioned.PartitionedTableImpl cannot be cast to class io.deephaven.engine.table.Table (io.deephaven.engine.table.impl.partitioned.PartitionedTableImpl and io.deephaven.engine.table.Table are in unnamed module of loader 'app')
        at io.deephaven.server.partitionedtable.PartitionedTableServiceGrpcImpl.lambda$getTable$11(PartitionedTableServiceGrpcImpl.java:122)
        at io.deephaven.server.session.SessionState$ExportObject.doExport(SessionState.java:884)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at io.deephaven.server.runner.DeephavenApiServerModule$ThreadFactory.lambda$newThread$0(DeephavenApiServerModule.java:164)
        at java.lang.Thread.run(Thread.java:833)
```

**Versions**
 - Deephaven: 0.19.0

nbauernfeind:
Fixes #3172

### DIFF
--- a/server/src/main/java/io/deephaven/server/partitionedtable/PartitionedTableServiceGrpcImpl.java
+++ b/server/src/main/java/io/deephaven/server/partitionedtable/PartitionedTableServiceGrpcImpl.java
@@ -14,8 +14,10 @@ import io.deephaven.proto.backplane.grpc.MergeRequest;
 import io.deephaven.proto.backplane.grpc.PartitionByRequest;
 import io.deephaven.proto.backplane.grpc.PartitionByResponse;
 import io.deephaven.proto.backplane.grpc.PartitionedTableServiceGrpc;
+import io.deephaven.server.auth.AuthorizationProvider;
 import io.deephaven.server.session.SessionService;
 import io.deephaven.server.session.SessionState;
+import io.deephaven.server.session.TicketResolverBase;
 import io.deephaven.server.session.TicketRouter;
 import io.grpc.stub.StreamObserver;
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/deephaven/server/partitionedtable/PartitionedTableServiceGrpcImpl.java
```diff
@@ -161,6 +159,8 @@ public void getTable(GetTableRequest request, StreamObserver<ExportedTableCreati
                                         .get(requestedRow.getRowSet().firstRowKey());
                             });
                         }
+                        authWiring.checkPermissionGetTable(
+                                session.getAuthContext(), request, List.of(table, keys.get()));
```
niloc132, line 163: to confirm, this checkPermissionGetTable shouldn't validate the original PartitionedTable, but checks just the results? I don't _think_ the keys by themselves are necessarily helpful for this, except as an intermediate step?

nbauernfeind: hmm.. @devinrsmith do you believe the permission handler would prefer a PartitionedTable or the result table? Can we propagate some sort of table attribute to the leaf nodes?

niloc132 (approved):

PR #3195, opened 2022-12-12
Bump grpc to 1.51.0

devinrsmith:
https://github.com/grpc/grpc/releases/tag/v1.51.0

Note: no bump to boringssl needed


niloc132:
Please check if we need other grpc runtimes or tools updated too, and possibly clarify commit message that this is about grpc-java specifically.

PR #3210, opened 2022-12-14
Fix LivenessScope for SelectColumnLayer.doApplyUpdate, in order to ensure reference counting for parallel update in transform use cases

### LINKED ISSUE #3209: Select/Update bug in Liveness for transform use-cases
This manifests itself more-obviously in parallel cases, but we are allowing the default `LivenessManager` (that is, whatever the `LivenessScopeStack` has on top) to manage `LivenessArtifact` instances created from `SelectColumnLayer.doApplyUpdate`, in addition to the `liveResultOwner`.

rcaudy:
Fixes #3209


lbooker42 (approved):
Looks good, tested against a failure in my own branch and this resolves it.

PR #3213, opened 2022-12-16
Barrage: Encode Packed Booleans Instead of Reinterpretted Bytes

### LINKED ISSUE #43: Barrage -> Do not reinterpret bool as byte

nbauernfeind:
Fixes #43.

This round trips with java flight, c++ flight, and deephaven's web ui.

### DIFF
--- a/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/BooleanChunkInputStreamGenerator.java
+++ b/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/BooleanChunkInputStreamGenerator.java
@@ -31,7 +31,7 @@ public class BooleanChunkInputStreamGenerator extends BaseChunkInputStreamGenera
 
     public static BooleanChunkInputStreamGenerator convertBoxed(
             final ObjectChunk<Boolean, Values> inChunk, final long rowOffset) {
-        // This code path is utilized for arrays and vectors of DateTimes, which cannot be reinterpreted.
+        // This code path is utilized for arrays / vectors, which cannot be reinterpreted.
         WritableByteChunk<Values> outChunk = WritableByteChunk.makeWritableChunk(inChunk.size());
         for (int i = 0; i < inChunk.size(); ++i) {
             final Boolean value = inChunk.get(i);
@@ -79,8 +79,8 @@ public class BooleanChunkInputStreamGenerator extends BaseChunkInputStreamGenera
             if (sendValidityBuffer()) {
                 size += getValidityMapSerializationSizeFor(subset.intSize(DEBUG_NAME));
             }
-            size += getNumLongsForBitPackOfSize(subset.intSize(DEBUG_NAME)) * 8L;
-            return LongSizedDataStructure.intSize("BaseChunkInputStream.getRawSize", size);
+            size += getNumLongsForBitPackOfSize(subset.intSize(DEBUG_NAME)) * (long) Long.BYTES;
+            return LongSizedDataStructure.intSize(DEBUG_NAME, size);
         }
 
         @Override
@@ -94,7 +94,7 @@ public class BooleanChunkInputStreamGenerator extends BaseChunkInputStreamGenera
             int validityLen = sendValidityBuffer() ? getValidityMapSerializationSizeFor(subset.intSize(DEBUG_NAME)) : 0;
             listener.noteLogicalBuffer(validityLen);
             // payload
-            listener.noteLogicalBuffer(getNumLongsForBitPackOfSize(subset.intSize(DEBUG_NAME)) * 8L);
+            listener.noteLogicalBuffer(getNumLongsForBitPackOfSize(subset.intSize(DEBUG_NAME)) * (long) Long.BYTES);
         }
 
         @Override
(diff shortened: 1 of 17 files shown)


extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/BooleanChunkInputStreamGenerator.java
```diff
@@ -0,0 +1,280 @@
...
+import java.util.Iterator;
+
+import static io.deephaven.util.QueryConstants.*;
+
+public class BooleanChunkInputStreamGenerator extends BaseChunkInputStreamGenerator<ByteChunk<Values>> {
+    private static final String DEBUG_NAME = "BooleanChunkInputStreamGenerator";
+
+    public static BooleanChunkInputStreamGenerator convertBoxed(
+            final ObjectChunk<Boolean, Values> inChunk, final long rowOffset) {
+        // This code path is utilized for arrays and vectors of DateTimes, which cannot be reinterpreted.
```
lbooker42, line 34: `DateTimes`?

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/BooleanChunkInputStreamGenerator.java
```diff
@@ -0,0 +1,280 @@
...
+                    flush.run();
+                }
+                bytesWritten += getValidityMapSerializationSizeFor(subset.intSize(DEBUG_NAME));
+            }
+
+            // write the included values
+            subset.forAllRowKeys(row -> {
+                final byte byteValue = chunk.get((int) row);
+                if (byteValue != NULL_BYTE) {
+                    context.accumulator |= (byteValue & 0x1L) << context.count;
```
lbooker42, line 141: Is `byteValue`always in [0,1]?  If `byteValue` allowed to be `>1` this may encode 0 for `true` values.  I'm not familiar with the underlying storage for our boolean data.

lbooker42 (commented):
This looks good, only a very few comments / questions.

web/client-api/src/main/java/io/deephaven/web/client/api/barrage/BarrageUtils.java
```diff
@@ -364,6 +364,21 @@ private static ColumnData readArrowBuffer(ByteBuffer data, Iter<FieldNode> nodes
...
             case "java.lang.Boolean":
+                // noinspection IntegerDivisionInFloatingPointContext
+                assert positions.length().toFloat64() >= ((size + 63) / 64);
+                // booleans are stored as a bitset, but internally we represent booleans as bytes
+                data.position((int) positions.offset().toFloat64());
+                BitSet wireValues = readBitSetWithLength(data, (int) (positions.length().toFloat64()));
+                byte[] boolArray = new byte[size];
+                for (int i = 0; i < size; ++i) {
+                    if (!hasNulls || valid.get(i)) {
+                        boolArray[i] = (byte) (wireValues.get(i) ? 1 : 0);
```
niloc132, line 375: Optionally we can go further here and actually replace the underlying store to not go back to byte[] - there's no benefit either way. Downstream of this, there is more code that will read those byte[]s and return them to true/false/null.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/BooleanChunkInputStreamGenerator.java
nbauernfeind: This is an excellent catch. I did not look at both conversion methods. You are right, byte values greater than 0 are all true.

server/src/main/java/io/deephaven/server/barrage/BarrageMessageProducer.java
```diff
@@ -294,8 +291,12 @@ public int hashCode() {
 
     private final Stats stats;
 
-    private final ColumnSource<?>[] sourceColumns; // might be reinterpreted
+    /** the possibly reinterpretted source column */
+    private final ColumnSource<?>[] sourceColumns;
+    /** which source columns are object columns and thus need proactive garbage collection */
     private final BitSet objectColumns = new BitSet();
+    /** internally, booleans are reinterpretted to bytes; however we need to be packed bitsets over Arrow */
+    private final Class<?>[] realColumnType;
```
rcaudy, line 299: I would record and use component types, too.

rcaudy (commented):
Minor commentary only, most addresses verbally. Ignored JS API code.

web/client-api/src/main/java/io/deephaven/web/client/api/barrage/WebBarrageUtils.java
```diff
@@ -364,6 +364,21 @@ private static ColumnData readArrowBuffer(ByteBuffer data, Iter<FieldNode> nodes
...
+                assert positions.length().toFloat64() >= ((size + 63) / 64);
+                // booleans are stored as a bitset, but internally we represent booleans as bytes
+                data.position((int) positions.offset().toFloat64());
+                BitSet wireValues = readBitSetWithLength(data, (int) (positions.length().toFloat64()));
+                Boolean[] boolArray = new Boolean[size];
+                for (int i = 0; i < size; ++i) {
+                    if (!hasNulls || valid.get(i)) {
+                        boolArray[i] = wireValues.get(i);
+                    } else {
+                        boolArray[i] = null;
```
niloc132, line 377: technically we can skip this, the array has nulls already in it

niloc132 (approved):

niloc132 (approved):

PR #3216, opened 2022-12-17
Add arrow module in Py API

jmao-denver:
Enable conversion between DH tables and Arrow tables on the server side


jmao-denver:
The 'missing pyarrow' fixes should be removed once the base images are updated with 'pyarrow' preinstalled as with numpy/pandas etc.

devinrsmith (dismissed):
This can't merge as is - the `--no-index` is a "safety" check that ensures everything flows through the base images.

If this is necessary, let's plumb https://github.com/deephaven/deephaven-base-images.

pyarrow is not a small dependency - it's 20-35MB depending on architecture, and it has a dependency on numpy.

Note, pyarrow is an architecture specific deliverable - but it looks like it has good coverage w/ amd64 and arm64 across Linux, Mac, and Windows with python 3.7-3.11.

py/server/tests/test_arrow.py
```diff
@@ -0,0 +1,69 @@
...
+            double_col(name="Double", data=[]),
+            string_col(name="String", data=[]),
+            datetime_col(name="Datetime", data=[]),
+        ]
+        dh_table = new_table(cols=cols)
+        pa_table = dharrow.to_arrow(dh_table)
+        dh_table_rt = dharrow.to_table(pa_table)
+        self.assert_table_equals(dh_table, dh_table_rt)
+
+
```
chipkent, line 83: These tests are all deephaven-centric.  We need another test that tests all supported and unsupported arrow types (e.g. utf8).

jmao-denver:
@nbauernfeind @lbooker42 has looked that the changes in Java and thought they are reasonable. Could you also review these changes?

py/server/tests/test_arrow.py
```diff
@@ -0,0 +1,69 @@
...
+        dh_table = dharrow.to_table(arrow_table)
+        pa_table = dharrow.to_arrow(dh_table)
+
+        dh_table_rt = dharrow.to_table(pa_table)
+
+        self.assert_table_equals(dh_table, dh_table_rt)
+
+    def test_round_trip_types(self):
+        cols = [
+            # bool_col(name="Boolean", data=[True, False]),
```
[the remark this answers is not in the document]
nbauernfeind: Boolean support is officially up for review now. This likely works on top of that change.

After a merge + rebase of boolean support we should bolster the tests below to include boolean columns, too.

py/server/tests/test_arrow.py
```diff
@@ -0,0 +1,98 @@
...
+    def tearDownClass(cls) -> None:
+        del cls.test_table
+
+    def test_round_trip(self):
+        arrow_table = papq.read_table("tests/data/day_trades.parquet")
+
+        dh_table = dharrow.to_table(arrow_table)
+        pa_table = dharrow.to_arrow(dh_table)
+        dh_table_rt = dharrow.to_table(pa_table)
+        self.assert_table_equals(dh_table, dh_table_rt)
```
nbauernfeind, line 46: This test feels like a land mine. If `to_table` creates an empty table then this test will pass, but would clearly be incorrect.

Could we compare the converted table to the version that our ParquetTools reader would generate instead?

py/server/tests/test_arrow.py
```diff
@@ -0,0 +1,98 @@
...
+        pa_table = dharrow.to_arrow(self.test_table)
+        pa_table_cols = dharrow.to_arrow(self.test_table, cols=cols)
+        dh_table = dharrow.to_table(pa_table, cols=cols)
+        dh_table_1 = dharrow.to_table(pa_table_cols)
+        self.assert_table_equals(dh_table_1, dh_table)
+
+    def test_crypto_data(self):
+        arrow_table = papq.read_table("tests/data/crypto_trades_1000.parquet")
+
+        with self.assertRaises(DHError) as cm:
```
nbauernfeind, line 84: Hmm, is there a reason we want to expect this? Is the issue that the file has boolean columns or some other type that is mismatching somehow?

py/server/deephaven/arrow.py
```diff
@@ -0,0 +1,140 @@
...
+
+    try:
+        pa_buffer = dh_schema.serialize()
+        j_barrage_table_builder.setSchema(pa_buffer.to_pybytes())
+
+        record_batches = pa_table.to_batches()
+        for rb in record_batches:
+            pa_buffer = rb.serialize()
+            j_barrage_table_builder.addRecordBatch(pa_buffer.to_pybytes())
+        j_barrage_table_builder.onCompleted()
```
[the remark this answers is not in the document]
nbauernfeind: > but it ends up being excessively complex and inefficient

Unfortunately we do not use Arrow Buffers internal to DHC. The alternative implementation that avoids copying data twice requires building column source wrappers that are backed by Arrow's native buffers. If we want to support converting a DH native table (e.g. converting a ticking table to PyArrow's format) then we must do the copying. Note that our column sources support all sorts of fancy features like redirecting to yield a sorted table.

Our internal column source organization is specifically what makes pulling py arrow data into DH an expensive operation (as well as vice versa).

jcferretti:
From the point of view of the question "is there overlap/work with the C++ wrapping / python client", I think the answer is not directly, and perhaps a way to understand that question would be to ask "should we worry about API consistency from the point of view of exposing arrow calls in python-server, python-client and C++-client?
I think that is a fair concern and we can look into that but my guess is right now is not the best time to do that, we have an open question on the C++ wrapping side about exactly how arrow is going to be exposed for python when we integrate both C++ libraries that depend on arrow code and python code that depends on pyarrow.  Without having clarity on the implementation level potential restrictions there, we can't (shouldn't?) say much now I think.
So I think is better to move this PR forward as is and perhaps open a separate ticket to check arrow API consistency between py-server, py-client and C++-client later.

py/server/deephaven/arrow.py
```diff
@@ -0,0 +1,140 @@
...
+        if isinstance(arrow_type, pa.TimestampType):
+            dh_type = "io.deephaven.time.DateTime"
+
+    if not dh_type:
+        raise DHError(message=f'unsupported arrow data type : {arrow_type}')
+
+    return {"deephaven:type": dh_type}
+
+
+def to_table(pa_table: pa.Table, cols: List[str] = None) -> Table:
```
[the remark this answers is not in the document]
jmao-denver: So far our convention is to not use Optional. If we decide to adopt the use of Optional then it should be a wholesale change across the package.

py/server/deephaven/arrow.py
```diff
@@ -0,0 +1,140 @@
...
+        cols (List[str]): the Arrow table column names, default is None which means including all columns
+
+    Returns:
+        a new table
+
+    Raises:
+        DHError
+    """
+    if cols:
+        pa_table = pa_table.select(cols)
```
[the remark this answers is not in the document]
jmao-denver: This will allocate a bit memory for the new table's metadata but not for the data, similar to what we do in DH.

py/server/tests/test_arrow.py
jmao-denver: Good point, added another assert on table size to tighten the test a bit.

py/server/tests/test_arrow.py
jmao-denver: this is just a placeholder for a potential bug we ran into, changed the name to make it more obvious.

py/server/tests/test_arrow.py
jmao-denver: Also added a test to specifically check against the table read in through DH's parquet tool

jmao-denver:
#3223 is created to track the lack of support of Arrow date32() type as captured in one of the unit tests.

py/server/deephaven/arrow.py
```diff
@@ -0,0 +1,140 @@
...
+    dh_type = arrow_to_dh.get(arrow_type)
+    if not dh_type:
+        # if this is a case of timestamp with tz specified
+        if isinstance(arrow_type, pa.TimestampType):
+            dh_type = "io.deephaven.time.DateTime"
+
+    if not dh_type:
+        raise DHError(message=f'unsupported arrow data type : {arrow_type}')
+
+    return {"deephaven:type": dh_type}
```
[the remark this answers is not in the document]
jmao-denver: 1. yea, we could make it part of dtypes, but I prefer to wait until we support more arrow data types and/or have some ways to use arrow data directly in DHC Python API
2. the ''  is really there as a place holder to keep the list complete and indicates that we should support the type in the future

nbauernfeind (approved):
LGTM if looks good to @chipkent

py/server/setup.py
```diff
@@ -57,6 +57,7 @@ def normalize_version(version):
         'deephaven-plugin',
         'numpy',
         'pandas',
+        'pyarrow<10', # Turbodbc can't be built on MacOS 13.0.1 with pyarrow==10.0.1, filed an issue with the maintainer
```
chipkent, line 60: We should have a tracking ticket on our side to resolve this.

py/server/deephaven/arrow.py
```diff
@@ -0,0 +1,148 @@
...
+    pa.timestamp('s', tz=None): '',
+    pa.timestamp('ms', tz=None): '',
+    pa.timestamp('us', tz=None): '',
+    pa.timestamp('ns', tz=None): '',
+    pa.date32(): 'java.time.LocalDate',
+    pa.date64(): 'java.time.LocalDate',
+    pa.duration('s'): '',
+    pa.duration('ms'): '',
+    pa.duration('us'): '',
+    pa.duration('ns'): '',
```
chipkent, line 44: Should we map to our Periods or to ns?

py/server/deephaven/arrow.py
```diff
@@ -0,0 +1,148 @@
...
+    pa.uint8(): '',
+    pa.uint16(): 'char',
+    pa.uint32(): '',
+    pa.uint64(): '',
+    pa.float16(): '',
+    pa.float32(): 'float',
+    pa.float64(): 'double',
+    pa.time32('s'): '',
+    pa.time32('ms'): '',
+    pa.time64('us'): '',
```
chipkent, line 33: Any reason these can't be mapped to DateTime?  Are we only supporting types that have full round-trip?  Are we not supporting type mapping?  etc.

py/server/deephaven/arrow.py
```diff
@@ -0,0 +1,148 @@
...
+    pa.duration('s'): '',
+    pa.duration('ms'): '',
+    pa.duration('us'): '',
+    pa.duration('ns'): '',
+    pa.month_day_nano_interval(): '',
+    pa.binary(): '',
+    pa.string(): 'java.lang.String',
+    pa.utf8(): 'java.lang.String',
+    pa.large_binary(): '',
+    pa.large_string(): '',
```
chipkent, line 50: Can this not just be a `java.lang.String` on the DH side?

py/server/deephaven/arrow.py
```diff
@@ -0,0 +1,148 @@
...
+    if not dh_type:
+        # if this is a case of timestamp with tz specified
+        if isinstance(arrow_type, pa.TimestampType) and arrow_type.tz:
+            dh_type = "io.deephaven.time.DateTime"
+
+    if not dh_type:
+        raise DHError(message=f'unsupported arrow data type : {arrow_type}, refer to '
+                              f'deephaven.arrow.SUPPORTED_ARROW_TYPES for the list of supported Arrow types.')
+
+    return {"deephaven:type": dh_type}
```
chipkent, line 75: 1) Together, these two are inconsistent.  We do support timestamps if they have timezones, but that is not reflected in `SUPPORTED_ARROW_TYPES`.

2) Are we not supporting cases without timezones to force users to specify the timezone or for some other reason?

3) I think that no specified timezone is traditionally interpreted as the machine local timezone.  https://arrow.apache.org/docs/python/timestamps.html

py/server/deephaven/arrow.py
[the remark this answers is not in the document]
chipkent: It seems like Arrow->DH necessarily creates a static DH table.  In that case, we should have column source support for Arrow buffers.  That is certainly a follow-on project, but the inefficiency here isn't great.  At a minimum, we should have a ticket for this effort.  The copy will certainly make some users unhappy at some point.

py/server/deephaven/arrow.py
```diff
@@ -0,0 +1,148 @@
...
+        pa_schema_buffer = j_arrow_builder.getSchema()
+        with pa.ipc.open_stream(pa.py_buffer(pa_schema_buffer)) as reader:
+            schema = reader.schema
+
+        record_batches = []
+        while j_arrow_builder.hasNext():
+            pa_rb_buffer = j_arrow_builder.next()
+            message = pa.ipc.read_message(pa_rb_buffer)
+            record_batch = pa.ipc.read_record_batch(message, schema=schema)
+            record_batches.append(record_batch)
```
chipkent, line 144: Does this handle UGP locking for ticking tables, or does the method need a locking annotation?  If locking is happening, there should be a comment, since it isn't transparent.

py/server/tests/test_arrow.py
```diff
@@ -0,0 +1,183 @@
...
+    def test_arrow_types_unsigned_integers(self):
+        with self.subTest("unsigned integers"):
+            pa_types = [
+                pa.uint16(),
+            ]
+            pa_data = [
+                pa.array([2 ** 16 - 1, 0]),
+            ]
+
+    @unittest.skip("Not correctly converted by DH")
```
chipkent, line 88: This looks like a problem.  We have a number of supported data types that have disabled unit tests and apparently do not work.

py/server/deephaven/arrow.py
```diff
@@ -0,0 +1,148 @@
...
+    pa.float32(): 'float',
+    pa.float64(): 'double',
+    pa.time32('s'): '',
+    pa.time32('ms'): '',
+    pa.time64('us'): '',
+    pa.time64('ns'): 'io.deephaven.time.DateTime',
+    pa.timestamp('s', tz=None): '',
+    pa.timestamp('ms', tz=None): '',
+    pa.timestamp('us', tz=None): '',
+    pa.timestamp('ns', tz=None): '',
```
chipkent, line 38: As noted later, it does look like we support timestamps if a timezone is present.

chipkent (changes requested):

py/server/deephaven/arrow.py
jmao-denver: @nbauernfeind Your thoughts?

py/server/deephaven/arrow.py
jmao-denver: @nbauernfeind?

py/server/deephaven/arrow.py
jmao-denver: @nbauernfeind ?

py/server/deephaven/arrow.py
jmao-denver: 1. #3273 is created for the possible follow-on project.
2. The engine code handles the locking for ticking table efficiently and automatically,  nothing needs to be done in Python wrapper code.

py/server/deephaven/arrow.py
nbauernfeind: We can, but this requires a separate set of java changes and should not be bundled here.

py/server/deephaven/arrow.py
nbauernfeind: We can, but this requires a separate set of java changes and should not be bundled here.

Currently, we only support types that can be round-tripped and have a unique on-to mapping. To support multiple source types mapping to the same destination type requires java-parsing-side effort.

py/server/deephaven/arrow.py
nbauernfeind: We probably can, but this requires a separate set of java changes and should not be bundled here.

py/server/deephaven/arrow.py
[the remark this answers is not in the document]
nbauernfeind: There is DHE work (DH-12088) that supports columns that wrap arrow buffers. I'm tasked to port that, but from a 10k ft view, it's still quite a bit of work.

py/server/tests/test_arrow.py
nbauernfeind: I appreciate having tests that I can enable once the java side supports new (or more-correct) conversions. Let's leave them in.

py/server/deephaven/arrow.py
nbauernfeind: It's possible that we only support the `ns` resolution version of this timestamp.

py/server/setup.py
jmao-denver: I have filed a ticket with Turbodbc and have helped find a solution to the problem.  I will be notified once the fix is merged and tagged for release.

py/server/deephaven/arrow.py
jmao-denver: 1. as one of the skipped test cases shows, the Arrow->DH conversion of timestamp isn't done correctly. The problem is tracked by #3223. So the SUPPORTED_ARROW_TYPES is correct even though the issue with 'timestamp' isn't strictly 'not supported'.

py/server/deephaven/arrow.py
chipkent: This looks like it is covered in https://github.com/deephaven/deephaven-core/issues/3223

py/server/deephaven/arrow.py
chipkent: This looks like it is covered in https://github.com/deephaven/deephaven-core/issues/3223

py/server/deephaven/arrow.py
chipkent: I have concrerns that `timestamps` return values, but the values are wrong.  Users are not very tolerant of corrupt data.

py/server/deephaven/arrow.py
chipkent: This looks like it is covered in https://github.com/deephaven/deephaven-core/issues/3223

chipkent (approved):

nbauernfeind (approved):

chipkent (approved):

devinrsmith (approved):
Docker / setup.py additions look correct to me; haven't reviewed the arrow-specific parts.

PR #3219, opened 2022-12-20
Fix default_groovy.env

devinrsmith:
DEEPHAVEN_SERVER_IMAGE was set to an outdated image. This sets it correctly. Tested with:

```
./gradlew prepareCompose
docker compose --env-file default_groovy.env up
```


mofojed (approved):

mofojed (approved):
Would also like a PR description - why did the default image name change? Why is server-slim no longer appropriate? How did you test?

PR #3220, opened 2022-12-20
Add JAVA_OPTS and START_OPTS environment variable parsing for gradle

devinrsmith:
Add gradle logic to explicitly read JAVA_OPTS and START_OPTS environment variables to provide developers a model that more closely matches with the native application script.

For example:

```sh
START_OPTS="-Xmx1g" ./gradlew server-jetty-app:run
```

### DIFF
--- a/buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
+++ b/buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
@@ -213,7 +213,4 @@ tasks.withType(CreateStartScripts).configureEach {
   // Note that we don't call compilerArgs() at this time, there is no way to template those strings.
   // Instead, we hard code the expected paths in the above templates to match the path etc/dh-compiler-directives.txt
 //  defaultJvmOpts += compilerArgs(compilerDirectivesFile.singleFile.path)
-
-  // The start script has its own logic for parsing the START_OPTS environment variable
-  // defaultJvmOpts += START_OPTS
 }
--- a/buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
+++ b/buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
@@ -116,12 +116,24 @@ def compilerArgs = { String compilerDirectivesFile ->
   ]
 }
 
-// Utility to add jvm args to all executions, whether intellij or from a application script or gradle javaexec
-def devJvmArgs = [
-//  '-XX:Tier4CompileThreshold=1000', // this optional line makes it easier to trigger the c2 error on the above methods
-//  '-XX:+PrintFlagsFinal',           // this optional line makes it easier to explore the final values for compiler args
-//  '-XX:+PrintCompilation',          // this optional line shows jit operations as they happen
-]
+def parseJvmArgumentsFromEnv = { String envName, List<String> orElse ->
+  def s = System.getenv(envName)
+  if (s == null) {
+    return orElse
+  }
+  // Note: this logic could be improved to handle quotes and other cases, but hopefully this suffices for now.
+  return s.split('\\s+').findAll { it -> !it.trim().isEmpty() } as List<String>
+}
+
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
```diff
@@ -116,12 +116,24 @@ def compilerArgs = { String compilerDirectivesFile ->
...
-//  '-XX:Tier4CompileThreshold=1000', // this optional line makes it easier to trigger the c2 error on the above methods
-//  '-XX:+PrintFlagsFinal',           // this optional line makes it easier to explore the final values for compiler args
-//  '-XX:+PrintCompilation',          // this optional line shows jit operations as they happen
-]
+def parseJvmArgumentsFromEnv = { String envName, List<String> orElse ->
+  def s = System.getenv(envName)
+  if (s == null) {
+    return orElse
+  }
+  // Note: this logic could be improved to handle quotes and other cases, but hopefully this suffices for now.
```
mofojed, line 124: So if you provide args, instead of adding to the JVM args it just replaces all of them - is that what we want?

buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
```diff
@@ -192,8 +211,9 @@ tasks.withType(CreateStartScripts).configureEach {
 //  windowsStartScriptGenerator.template = windowsStartScript
 
   // Note that we don't call compilerArgs() at this time, there is no way to template those strings.
-  // Instead, we hard code the expected paths in the above templates to match the path lib/dh-compiler-directives.txt
+  // Instead, we hard code the expected paths in the above templates to match the path etc/dh-compiler-directives.txt
 //  defaultJvmOpts += compilerArgs(compilerDirectivesFile.singleFile.path)
 
-  defaultJvmOpts += devJvmArgs
+  // The start script has its own logic for parsing the START_OPTS environment variable
+  // defaultJvmOpts += START_OPTS
```
mofojed, line 218: Why leave this in/commented out?

mofojed (changes requested):
You're killing me with the lack of description in the PR. Need to document why this change is being made, how it's supposed to work, how you tested it out.

I tested with this and it seems to work:
```
START_OPTS="-Ddeephaven.jsPlugins.resourceBase=/home/bender/dev/deephaven/deephaven-js-plugins" ./gradlew server-jetty-app:run
```

buildSrc/src/main/groovy/io.deephaven.java-toolchain-conventions.gradle
devinrsmith: I'm not sure I understand the question. The limitation right now is that "complex" cases aren't handled. Something like

```shell
START_OPTS="-Dmy.prop=\"my value with spaces\"" ...
```

is not currently parsed correctly.

mofojed: There's default JVM Args:
```
  '-XX:+UseG1GC',
  '-XX:MaxGCPauseMillis=100',
```
Now if you specify `JAVA_OPTS="-XX:+PrintGCDetails"`, you'll lose those default JVM args instead of it being added to those args. Just wondering if that's the behaviour we want - I would've expected JVM args I provide to be in addition to the default args (I think)

devinrsmith: Yep, this is the behavior we want IMO.

https://deephaven.io/core/docs/how-to-guides/configuration/native-application/#native-application-script

> JAVA_OPTS is an environment variable that may contain pre-defined Deephaven recommended arguments. The user is able to set their own JAVA_OPTS environment variable to override the Deephaven defaults.

If you want to "append", then START_OPTS is the way to go, and that will be ok for `START_OPTS="-XX:+PrintGCDetails"`

The reason you might want to override JAVA_OPTS is to set another GC (which won't work w/ START_OPTS b/c it conflicts w/ JAVA_OPTS): `JAVA_OPTS="-XX:+UseZGC"`.

mofojed (approved):

PR #3222, opened 2022-12-21
Port DH-11271: Rearchitect Static Cross Join to Deal with Rehash More Efficiently

nbauernfeind:
The import changes were caused by a run of `replicateAllSafe`. This JoinStateManager is not replicated which is why there were no downstream changes. This lack of downstream replication matches the DHE commit.

I ran the CrossJoin tests locally and kicked off nightlies.
https://github.com/nbauernfeind/deephaven-core/actions/runs/3745171275/jobs/6359316873


cpwright (approved):

lbooker42 (approved):
This is a faithful port of the DHE changes, approved.

Issue #3235, opened 2022-12-23
gRPC error: ColumnAggregations should have at least two pairs


niloc132:
When implementing the JS API for aggregations, the consistent UI is to let the user build a list of aggregations (at least one), and so in turn the API behind the UI will take at least one aggregation and translate that list into an aggregation object to send to the server.

However, the server emits only `Internal Error: Details Logged w/ID '5d9df6ae-e6b6-482e-90fe-9ae7771c20c3'` when this happens. The server error is actually:

```
java.lang.IllegalArgumentException: class io.deephaven.api.agg.ColumnAggregations should have at least two pairs
        at io.deephaven.api.agg.ColumnAggregations.checkSize(ColumnAggregations.java:40)
        at io.deephaven.api.agg.ImmutableColumnAggregations.validate(ImmutableColumnAggregations.java:94)
        at io.deephaven.api.agg.ImmutableColumnAggregations.access$400(ImmutableColumnAggregations.java:27)
        at io.deephaven.api.agg.ImmutableColumnAggregations$Builder.build(ImmutableColumnAggregations.java:189)
        at io.deephaven.api.agg.ImmutableColumnAggregations$Builder.build(ImmutableColumnAggregations.java:123)
        at io.deephaven.api.agg.Aggregation.of(Aggregation.java:82)
        at io.deephaven.server.table.ops.AggregationAdapter.adapt(AggregationAdapter.java:71)
        at io.deephaven.server.table.ops.AggregationAdapter$Adapter.adapt(AggregationAdapter.java:231)
        at io.deephaven.server.table.ops.AggregationAdapter$Adapters.adapt(AggregationAdapter.java:100)
        at io.deephaven.server.table.ops.AggregationAdapter.adapt(AggregationAdapter.java:59)
        at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
        at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
        at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
[...]

PR #3242, opened 2022-12-26
Improve error message for empty ColumnAggregation(s)

devinrsmith:
Provides better developer experience, see #3235


rcaudy (commented):
This makes things better. I'm not sure it fully addresses the developer UX concerns.

rcaudy (approved):

PR #3246, opened 2022-12-28
Add chart series support for point shapes and sizes

mofojed:
- Add point shape and size to series info, so markers can be plotted with the correct shapes and sizes
- Tested with Web UI branch https://github.com/mofojed/web-client-ui/tree/figure-point-shapes
- Used the following code in Python:
```
import math

math_funcs = ["sin", "cos"]
shapes = ["SQUARE", "CIRCLE", "UP_TRIANGLE", "DOWN_TRIANGLE", "RIGHT_TRIANGLE", "LEFT_TRIANGLE", "DIAMOND"]
# Unsupported shapes: ["ELLIPSE", "HORIZONTAL_RECTANGLE", "VERTICAL_RECTANGLE"]

# Create a generic table that has enough columns to display all the shapes
# Re-uses some of the funcs
t = empty_table(50).update(["x=i*0.1"])
for i in range(len(shapes)):
    t = t.update([f"y{i}=Math.{math_funcs[i % len(math_funcs)]}(x+{math.floor(i / len(math_funcs))})"])

# Generate the table and figure based on the shapes created
p = Figure()
for i in range(len(shapes)):
    p = p.plot_xy(series_name=shapes[i], t=t, x="x", y=f"y{i}").point(shape=shapes[i], size=len(shapes) - i)

p = p.show();
```

### DIFF
--- a/plugin/figure/src/main/java/io/deephaven/figure/FigureWidgetTranslator.java
+++ b/plugin/figure/src/main/java/io/deephaven/figure/FigureWidgetTranslator.java
@@ -153,6 +153,14 @@ public class FigureWidgetTranslator {
         return clientFigure.build();
     }
 
+    private static void assignOptionalStringField(Object value, Consumer<String> setter, Runnable clear) {
+        if (value != null) {
+            setter.accept(value.toString());
+        } else {
+            clear.run();
+        }
+    }
+
     private static <T> void assignOptionalField(T value, Consumer<T> setter, Runnable clear) {
         if (value != null) {
             setter.accept(value);
@@ -289,6 +297,8 @@ public class FigureWidgetTranslator {
                             assignOptionalField(toCssColorString(s.getLineColor()), clientSeries::setLineColor,
                                     clientSeries::clearLineColor);
                             // clientSeries.setLineStyle(s.getLineStyle().toString());
+                            assignOptionalField(toCssColorString(s.getSeriesColor()), clientSeries::setShapeColor,
+                                    clientSeries::clearShapeColor);
                             assignOptionalField(s.getPointLabelFormat(), clientSeries::setPointLabelFormat,
                                     clientSeries::clearPointLabelFormat);
                             assignOptionalField(s.getXToolTipPattern(), clientSeries::setXToolTipPattern,
@@ -300,18 +310,12 @@ public class FigureWidgetTranslator {
                             // with the x and y we have so far mapped to this
 
                             if (s instanceof AbstractXYDataSeries) {
+                                // TODO: Individual point shapes/sizes/labels
+                                // Right now just gets one set for the whole series
                                 AbstractXYDataSeries abstractSeries = (AbstractXYDataSeries) s;
-
-                                // TODO color label
-                                Double pointSize = abstractSeries.getPointSize(0);
-                                if (pointSize != null) {
-                                    clientSeries.setShapeSize(pointSize);
-                                }
-
-                                Shape pointShape = abstractSeries.getPointShape(0);
-                                if (pointShape != null) {
-                                    clientSeries.setShape(pointShape.toString());
-                                }
+                                assignOptionalStringField(abstractSeries.getPointShape(0), clientSeries::setShape, clientSeries::clearShape);
+                                assignOptionalField(abstractSeries.getPointSize(0), clientSeries::setShapeSize, clientSeries::clearShapeSize);
+                                assignOptionalField(abstractSeries.getPointLabel(0), clientSeries::setShapeLabel, clientSeries::clearShapeLabel);
 
                                 if (s instanceof IntervalXYDataSeriesArray) {
                                     // interval (aka histogram)
@@ -348,6 +352,13 @@ public class FigureWidgetTranslator {
                                     errorList.add("OpenAPI presently does not support series of type " + s.getClass());
                                 }
                             } else if (s instanceof AbstractCategoryDataSeries) {
+                                // TODO: Individual point shapes/sizes/labels
+                                // Right now just gets one set for the whole series
+                                AbstractCategoryDataSeries abstractSeries = (AbstractCategoryDataSeries) s;
+                                assignOptionalStringField(abstractSeries.getPointShape(0), clientSeries::setShape, clientSeries::clearShape);
+                                assignOptionalField(abstractSeries.getPointSize(0), clientSeries::setShapeSize, clientSeries::clearShapeSize);
+                                assignOptionalField(abstractSeries.getLabel(0), clientSeries::setShapeLabel, clientSeries::clearShapeLabel);
+
                                 if (s instanceof CategoryDataSeriesPartitionedTable) {// bar and pie from a table
                                     CategoryDataSeriesPartitionedTable series = (CategoryDataSeriesPartitionedTable) s;
                                     clientAxes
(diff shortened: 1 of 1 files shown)


rcaudy (commented):
What about MultiSeries?

plugin/figure/src/main/java/io/deephaven/figure/FigureWidgetTranslator.java
```diff
@@ -300,6 +310,16 @@ private FigureDescriptor.ChartDescriptor translate(ChartImpl chart) {
                             // with the x and y we have so far mapped to this
 
                             if (s instanceof AbstractXYDataSeries) {
+                                // TODO: Individual point shapes/sizes/labels
```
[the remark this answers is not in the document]
mofojed: Added

mofojed:
MultiSeries does set the point shapes (line 465, line 502) but I'm not sure how to test it. I think I need `plot_by` functionality which hasn't been plumbed through yet. Or do you know of a way to trigger it?

rcaudy (commented):
We can merge like this, but we should figure out about multi-series before merging.

rcaudy:
> MultiSeries does set the point shapes (line 465, line 502) but I'm not sure how to test it. I think I need `plot_by` functionality which hasn't been plumbed through yet. Or do you know of a way to trigger it?

It looks to me like `plotBy`, `catPlotBy`, and `ohlcPlotBy` will make a `MultiSeries`.

mofojed:
`plot_by` doesn't seem to be available in python, so I converted the code to Groovy and add a `plotBy` to get a multi-series. It looks like the point shapes are appearing as expected (see the screenshot and code snippet in my original comment).

mofojed:
@niloc132 pointed out there is actually a plot-by-key in python: https://deephaven.io/core/docs/reference/cheat-sheets/cheat-sheet/#plot-by-some-key

I tried to map the same groovy snippet above to using plotBy in Python, but server is throwing an NPE:
```
import math

math_funcs = ["sin", "cos"]
shapes = ["SQUARE", "CIRCLE", "UP_TRIANGLE", "DOWN_TRIANGLE", "RIGHT_TRIANGLE", "LEFT_TRIANGLE", "DIAMOND"]
# Unsupported shapes: ["ELLIPSE", "HORIZONTAL_RECTANGLE", "VERTICAL_RECTANGLE"]

# Create a generic table that has enough columns to display all the shapes
# Re-uses some of the funcs
t = empty_table(50).update([f"name=shapes[i % 5]", "x=i*0.1"])
for i in range(len(shapes)):
    t = t.update([f"y{i}=Math.{math_funcs[i % len(math_funcs)]}(x+{math.floor(i / len(math_funcs))})"])

# Generate the table and figure based on the shapes created
p = Figure()
for i in range(len(shapes)):
    p = p.plot_xy(series_name=shapes[i], t=t, x="x", y=f"y{i}").point(shape=shapes[i], size=len(shapes) - i, visible=True)

p = p.show();

p2 = Figure().plot_xy(series_name="Multi", t=t, x="x", y="y1", by=["name"]).point(visible=True)
for i in range(len(shapes)):
    p2 = p2.point(shape=shapes[i], multi_series_key=[shapes[i]])

p2 = p2.show()
```
Throws:
```
heduler-Concurrent-1 | i.d.s.s.SessionState      | Internal Error '2d7baeac-6803-4c68-bef1-e1b31c0be75d' java.lang.NullPointerException: null
        at io.deephaven.extensions.barrage.util.ExportUtil.buildTableCreationResponse(ExportUtil.java:20)
        at io.deephaven.extensions.barrage.util.ExportUtil.buildTableCreationResponse(ExportUtil.java:13)
        at io.deephaven.server.partitionedtable.PartitionedTableServiceGrpcImpl.lambda$getTable$8(PartitionedTableServiceGrpcImpl.java:166)
        at io.deephaven.server.session.SessionState$ExportObject.doExport(SessionState.java:886)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at io.deephaven.server.runner.DeephavenApiServerModule$ThreadFactory.lambda$newThread$0(DeephavenApiServerModule.java:165)
```

Not sure what I'm doing wrong there, but I tried a different example in python and it worked:
```

table_3_keys = static_source_1.head(2000).where("String1 in `A`, `B`, `C`")\
    .sort("Int1").update(["New_Row = i", "Y = Int1 * (String1 = `A` ? 2 : String1 = `B` ? 0.5 : 1.0)"])
p = Figure().plot_xy(series_name="partitioned", t=table_3_keys, by=["String1"], x="New_Row", y="Y").point(shape="SQUARE", visible=True, multi_series_key=["A"]).point(shape="DIAMOND", visible=True, multi_series_key=["B"]).point(shape="UP_TRIANGLE", visible=True, multi_series_key=["C"]).show()
```

mofojed:
Figured out what the issue was in my snippet above. The problem was on this line:
```
t = empty_table(50).update([f"name=shapes[i % 5]", "x=i*0.1"])
```
This was creating a table where `name` was of type `PyObject`. I needed to change it to a native String, I did it with string concatenation:
```
t = empty_table(50).update([f"name=`` + shapes[i % 5]", "x=i"])
```

Not sure if there's anything we can do about that, pretty annoying that it can't handle the first syntax correctly.

rcaudy (approved):

plugin/figure/src/main/java/io/deephaven/figure/FigureWidgetTranslator.java
```diff
@@ -334,9 +354,17 @@ private FigureDescriptor.ChartDescriptor translate(ChartImpl chart) {
                                     // warn about other unsupported series types
                                     errorList.add("OpenAPI presently does not support series of type " + s.getClass());
                                 }
-
-                                // TODO color label size shape
                             } else if (s instanceof AbstractCategoryDataSeries) {
+                                // TODO #3293: Individual point shapes/sizes/labels
+                                // Right now just gets one set for the whole series
+                                AbstractCategoryDataSeries abstractSeries = (AbstractCategoryDataSeries) s;
+                                assignOptionalStringField(abstractSeries.getPointShape(0), clientSeries::setShape,
```
niloc132, line 361: Is one special cased in the groovy/py plot apis? I thought that this API was always supposed to take an array of points.

I think instead you want getDefaultValue(), and if not null we will assign that, otherwise if the user assigns "dot square line dot square line" they'll just see "dot dot dot dot dot", and maybe we can send a "sorry, can't support that" error

niloc132: Why is color not in this patch too? Removed from the TODO, but not copied over even for a default? Even if we think that users shouldnt set colors, we might want the ability to use a table column for colors, like format columns do, so it is probably worth partially plumbing...

mofojed: @niloc132 What `getDefaultValue()` are you talking about? 
Color is already handled on line 297.

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2112
Conceptual: https://github.com/deephaven/deephaven.io/issues/2114
Reference: https://github.com/deephaven/deephaven.io/issues/2113
Blog: https://github.com/deephaven/deephaven.io/issues/2111

PR #3252, opened 2022-12-29
Fix TableViewportSubscription race condition

### LINKED ISSUE #3251: Using the Web GUI's "Select Distinct" function results in an error.
If you use "Select Distinct Values" in the Web GUI, it results in the table no longer displaying properly.

Create a table, and use the Web GUI to "Select Distinct Values". For instance, this table:

```python

seattle_weather = read_csv("https://media.githubusercontent.com/media/deephaven/examples/main/GSOD/csv/seattle.csv")
```

The error message in the table is `Error: java.lang.AssertionError: realized table is closed, expected status DONE, instead is ACTIVE`.

**Expected results**

**Actual results**

**Versions**

 - Engine Version: latest
 - Web UI Version: latest
 - Java Version: 11
 - Barrage Version: Unsure
 - OS: OS X Monterey
 - Browser: Chrome
 - Docker: 20.10.13

mofojed:
- Race condition was causing UI error with selectDistinct
- TableViewportSubscription could initialize after it had already been closed
- Check the status to see if we're already DONE before initializing
- Fixes #3251

Race condition occurs when `subscription.close()` is called before the subscription has copied the original table. You can reproduce this in a JS API page in the developer console by doing a `setInterval`, e.g.:
```
setInterval(() => {
    var subscription = table.setViewport(0, 10)
    subscription.close()
    console.log('table size', table.size)
}, 50)
```


web/client-api/src/main/java/io/deephaven/web/client/api/subscription/scratch
```diff
@@ -0,0 +1,13 @@
+Trigger Select Distinct
```
devinrsmith, line 1: Extraneous file, yeah? If it's useful for reviewer to see, might post as comment.

devinrsmith (commented):
I think @niloc132 should review this.

web/client-api/src/main/java/io/deephaven/web/client/api/subscription/scratch
mofojed: Whoops, yea I did not mean to include that. The race condition can be identified a little more clearly than mentioned in the doc, I'll update the description.

web/client-api/src/main/java/io/deephaven/web/client/api/subscription/TableViewportSubscription.java
```diff
@@ -109,9 +109,13 @@ public TableViewportSubscription(double firstRow, double lastRow, Column[] colum
         this.original = existingTable;
         this.originalState = original.state();
         copy = existingTable.copy(false).then(table -> new Promise<>((resolve, reject) -> {
-
             // Wait until the state is running to copy it
             originalState.onRunning(newState -> {
+                if (this.status == Status.DONE) {
+                    JsLog.debug("TableViewportSubscription closed before originalState.onRunning completed, ignoring");
+                    table.close();
```
niloc132, line 116: i dont think you need to call this - in the case where this.status == DONE (assigned in internalClose(), we have already closed this.table?

niloc132: Then instead of closing here, we should reject the promise.

niloc132: Nope, this doesn't work after all, table.close() probably makes the most sense.

niloc132 (approved):

PR #3256, opened 2023-01-03
Audit All `@apiNote` and Remove Duplicate Usages in Single Javadoc

nbauernfeind:
Javadoc does not render multiple `@apiNote` tags for a single javadoc. This fixes instances where there are worthy notes that were being dropped.

### DIFF
--- a/engine/api/src/main/java/io/deephaven/engine/table/Table.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/Table.java
@@ -982,8 +982,8 @@ public interface Table extends
      * to avoid deeply nested structures.
      *
      * @apiNote It's best to avoid many chained calls to {@link #mergeBefore(Table...)} and
-     *          {@link #mergeAfter(Table...)}, as this may result in deeply-nested data structures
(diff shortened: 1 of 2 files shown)


engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -983,7 +983,7 @@ RollupTable rollup(Collection<? extends Aggregation> aggregations, boolean inclu
      *
      * @apiNote It's best to avoid many chained calls to {@link #mergeBefore(Table...)} and
      *          {@link #mergeAfter(Table...)}, as this may result in deeply-nested data structures
-     * @apiNote See TableTools.merge(Table...)
+     *          See TableTools.merge(Table...)
```
rcaudy, line 986: We probably need a period on the previous sentence.

rcaudy: Applies to the other fix in this file, too.

rcaudy (approved):

Issue #3249, opened 2022-12-29
Python wrappers for rollups and trees


rcaudy:
With the merge of #3243, we have complete (although not entirely well tested) implementations for rollups and trees accessible from the Java API. Add wrappers from Python.

There are two areas to cover;
1. Adding the appropriate Table operations
2. Wrapping the result types (RollupTable and TreeTable, but children of HierarchicalTable)

We should focus on the former at high priority. The latter is less urgent, and will require Python API design work,

PR #3259, opened 2023-01-04
Wrap the rollup and tree Table operations

jmao-denver:
The new RollupTable and TreeTable classes are very hollow right  now in that no further operations are supported on them. However they abide by the JObjectWrapper protocol and should be easily made visible through DHC's gRPC API in the WebUI.

Partial #3249

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -56,6 +56,10 @@ _JPythonScriptSession = jpy.get_type("io.deephaven.integrations.python.PythonDee
 _test_vectorization = False
 _vectorized_count = 0
 
+# Rollup Table and Tree Table
+_JRollupTable = jpy.get_type("io.deephaven.engine.table.hierarchical.RollupTable")
+_JTreeTable = jpy.get_type("io.deephaven.engine.table.hierarchical.TreeTable")
+
 
 def _j_py_script_session() -> _JPythonScriptSession:
     j_execution_context = _JExecutionContext.getContext()
@@ -212,6 +216,42 @@ def _td_to_columns(table_definition):
             )
         )
     return cols
+
+
+class RollupTable(JObjectWrapper):
+    """ A RollupTable is generated as a result of applying the :meth:`~Table.rollup` method on a Table.
+
+    Note: It should not be instantiated directly by user code.
+    """
+    j_object_type = _JRollupTable
+
+    @property
+    def j_object(self) -> jpy.JType:
+        return self.j_rollup_table
+
+    def __init__(self, j_rollup_table: jpy.JType, aggs: Sequence[Aggregation], include_constituents: bool,
+                 by: Sequence[str]):
+        self.j_rollup_table = j_rollup_table
+        self.aggs = aggs
+        self.include_constituents = include_constituents
+        self.by = by
+
+
+class TreeTable(JObjectWrapper):
+    """ A TreeTable is generated as a result of applying the :meth:`~Table.tree` method on a Table.
+
+    Note: It should not be instantiated directly by user code.
+    """
+    j_object_type = _JTreeTable
+
+    @property
+    def j_object(self) -> jpy.JType:
+        return self.j_tree_table
+
+    def __init__(self, j_tree_table: jpy.JType, id_col: str, parent_col: str):
+        self.j_tree_table = j_tree_table
+        self.id_col = id_col
+        self.parent_col = parent_col
 
 
 class Table(JObjectWrapper):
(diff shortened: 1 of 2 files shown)


py/server/deephaven/table.py
```diff
@@ -1661,6 +1701,80 @@ def slice(self, start: int, stop: int) -> Table:
         except Exception as e:
             raise DHError(e, "table slice operation failed.") from e
 
+    def rollup(self, aggs: Union[Aggregation, Sequence[Aggregation]], by: Union[str, Sequence[str]] = None,
+               include_constituents: bool = False) -> RollupTable:
+        """Creates a rollup table.
+
+         A rollup table aggregates by the specified columns, and then creates a hierarchical table which re-aggregates
+         using one less aggregation column on each level. The column that is no longer part of the aggregation key is
```
rcaudy, line 1709:
```suggestion
         using one less by column on each level. The column that is no longer part of the aggregation key is
```

py/server/deephaven/table.py
```diff
@@ -1661,6 +1701,80 @@ def slice(self, start: int, stop: int) -> Table:
...
+            DHError
+        """
+        try:
+            if promote_orphans:
+                with auto_locking_ctx(self):
+                    j_table = _JTreeTable.promoteOrphans(self.j_table, id_col, parent_col)
+            else:
+                j_table = self.j_table
+
+            with auto_locking_ctx(self):
```
rcaudy, line 1772: We don't need a lock here.

py/server/deephaven/table.py
```diff
@@ -1661,6 +1701,80 @@ def slice(self, start: int, stop: int) -> Table:
...
+            a new RollupTable
+
+        Raises:
+            DHError
+        """
+        try:
+            aggs = to_sequence(aggs)
+            by = to_sequence(by)
+            j_agg_list = j_array_list([agg.j_aggregation for agg in aggs])
+            with auto_locking_ctx(self):
```
rcaudy, line 1730: We don't need a lock here.

rcaudy (dismissed):

py/server/deephaven/table.py
```diff
@@ -1661,6 +1701,76 @@ def slice(self, start: int, stop: int) -> Table:
...
+    def rollup(self, aggs: Union[Aggregation, Sequence[Aggregation]], by: Union[str, Sequence[str]] = None,
+               include_constituents: bool = False) -> RollupTable:
+        """Creates a rollup table.
+
+         A rollup table aggregates by the specified columns, and then creates a hierarchical table which re-aggregates
+         using one less by column on each level. The column that is no longer part of the aggregation key is
+         replaced with null on each level.
+
+         Note some aggregations can not be used in creating a rollup tables, these include: group, partition, median,
+         pct, weighted_avg
```
chipkent, line 1713: What makes each of these aggs not work?  (e.g. median, pct, and weighed_avg)

py/server/tests/test_table.py
```diff
@@ -809,22 +824,38 @@ def test_slice(self):
...
             rt = t.slice(3, 2)
 
+    def test_rollup(self):
+        test_table = empty_table(100)
+        test_table = test_table.update(
+            ["grp_id=(int)(i/5)", "var=(int)i", "weights=(double)1.0/(i+1)"]
+        )
+        with self.assertRaises(DHError) as cm:
+            rollup_table = test_table.rollup(aggs=self.aggs)
+        self.assertRegex(str(cm.exception), r".+ is not supported for rollup")
```
chipkent, line 834: It seems like you should be testing each "bad" agg individually, instead of all at once.

chipkent (changes requested):

py/server/deephaven/table.py
jmao-denver: @rcaudy Maybe you can explain?

py/server/deephaven/table.py
```diff
@@ -214,6 +218,42 @@ def _td_to_columns(table_definition):
     return cols
 
 
+class RollupTable(JObjectWrapper):
```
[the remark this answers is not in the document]
jmao-denver: Yes but that effort is quite substantial and that's why this PR only partially addresses the ticket.

py/server/deephaven/table.py
```diff
@@ -214,6 +218,42 @@ def _td_to_columns(table_definition):
...
+    Note: It should not be instantiated directly by user code.
+    """
+    j_object_type = _JRollupTable
+
+    @property
+    def j_object(self) -> jpy.JType:
+        return self.j_rollup_table
+
+    def __init__(self, j_rollup_table: jpy.JType, aggs: Sequence[Aggregation], include_constituents: bool,
+                 by: Sequence[str]):
```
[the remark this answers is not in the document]
jmao-denver: 1. I assume the args are just static properties on the RollupTable, not sure how they would change out of sync, @rcaudy ?
2. That's quite unlikely as they are for visualization only.

py/server/deephaven/table.py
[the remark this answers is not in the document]
rcaudy: Indeed. That can be a January deliverable, instead of a December one.

py/server/deephaven/table.py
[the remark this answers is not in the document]
rcaudy: They are immutable properties of the `RollupTable`. This is safe.

py/server/deephaven/table.py
```diff
@@ -214,6 +218,42 @@ def _td_to_columns(table_definition):
...
+
+    Note: It should not be instantiated directly by user code.
+    """
+    j_object_type = _JTreeTable
+
+    @property
+    def j_object(self) -> jpy.JType:
+        return self.j_tree_table
+
+    def __init__(self, j_tree_table: jpy.JType, id_col: str, parent_col: str):
```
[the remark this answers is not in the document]
rcaudy: Same as in https://github.com/deephaven/deephaven-core/pull/3259#discussion_r1063798959

py/server/deephaven/table.py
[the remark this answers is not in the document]
chipkent: Create a separate ticket and item in the planning doc.

py/server/deephaven/table.py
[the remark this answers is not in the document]
chipkent: I'm just saying that what is here has the potential of being brittle -- not that it isn't correct.

py/server/deephaven/table.py
chipkent: Should we then have a note that is less specific, with the hope that these things do get fixed?  Then we don't have to remember that these notes need to be updated.

chipkent (approved):
Approved, but my comments to existing conversations should be considered before merging.

py/server/deephaven/table.py
rcaudy: I'm not sure. This is Python specific messaging; the Java code has its own, right down in the authoritative code.

rcaudy: I'm not sure. This is Python specific messaging; the Java code has its own, right down in the authoritative code.

py/server/deephaven/table.py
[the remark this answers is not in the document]
jmao-denver: The ticket will stay open as this PR only partially fixes it.

py/server/deephaven/table.py
jmao-denver: I am OK with leaving it as is because the individual failure tests will catch the change if we do get these fixed in the future.

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2002
Conceptual: https://github.com/deephaven/deephaven.io/issues/2004
Reference: https://github.com/deephaven/deephaven.io/issues/2003
Blog: https://github.com/deephaven/deephaven.io/issues/2005

PR #3265, opened 2023-01-05
Ensure the stamp key is nulled out if the trigger table is empty

devinrsmith:


engine/table/src/main/java/io/deephaven/engine/table/impl/sources/CharacterSingleValueSource.java
```diff
@@ -65,6 +65,13 @@ public final void set(char value) {
     }
     // endregion UnboxedSetter
 
+    @Override
+    public final void setNull() {
+        // region null set
```
rcaudy, line 70: What did you need a region for? I don't see anything doing a region replace, since there are no replicator changes.

devinrsmith: I wasn't sure - I copied the same pattern I see being used for `setNull(int index)`. Should I get rid of these comments?

rcaudy (approved):

PR #3278, opened 2023-01-09
Specialize our safelyExecute calls to avoid logs

### LINKED ISSUE #3277: `safelyExecute` calls log too much
I've been running into this over the last few months, but it is a bigger issue with trees/rollups now, so i'd like to talk about either reverting this change or actually fixing callsites to log important things instead of everything: in https://github.com/deephaven/deephaven-core/pull/3101/files#r1034167610, safelyExecute and safelyExecuteLocked were turned up from debug to error.

The problem is that we use these calls to essentially say "something might go wrong here, but I don't want to hear about it, error handling is already in place elsewhere", which is probably helpful early in development, but if we ship code where this is acceptable, we've already screwed up. The specific case where this matters a lot is racing closed streams from the server, where the client doesn't have a "nice" way to say "thanks, go away now, i'm no longer interested" against the server sending results, so today the client just kills the stream, and the server then uses these error handling cases.

A few examples:
* Browser window closes or refreshes - all streams stop, and any onNext/onError/onCompleted will now throw. There probably isn't a constructive reason to log this.
* User changes their mind about what they want, usually during a long-running operation - without a mechanism in gRPC to terminate the call, it makes the most sense to just cancel the http call itself, avoid any client-side grpc handling, proto deserialization, etc. Likewise, there probably isnt a good reason to log this.
* The same will apply to non-browser clients too, if they terminate, or for some reason get tired of waiting and cancel a long-running call.

Note that in order for there to be a race, the client either needs to be very spammy, or needs to be doing long enough calls that they can cancel before they are completed, which mostly means that we don't see this case when testing simple stuff, only real users hit this with heavy loads or with industrial-strength impatience.

So, I'd like to re-suppress those log messages. We probably did initially un-suppress them for good reason, as suggested before too much logic in the lambdas, so I further suggest that we replace the majority of our usage of these methods with something like this:
* `safelyOnError(observer, error)`
* `safelyOnNext(observer, message)` with generics to make sure the right message is passed in
* `safelyComplete(observer)` 
* `safelyComplete(observer, message)` which will be onNext+complete, for unary calls and so on, might as well make an easy-mode for this which we do a lot

There is at least a case or two that don't follow this, but I suspect they might be errors and shouldn't be considered "safely execute" calls anyway

My inclination is to make all of these the "locked" to be more defensive, but no strong opinion there.

niloc132:
Fixes #3277

### DIFF
--- a/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/GrpcUtil.java
+++ b/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/GrpcUtil.java
@@ -146,7 +146,7 @@ public class GrpcUtil {
      *
      * @param runner the runnable to execute safely
      */
-    public static void safelyExecuteLocked(final Object lockedObject,
+    private static void safelyExecuteLocked(final Object lockedObject,
             final FunctionalInterfaces.ThrowingRunnable<Exception> runner) {
         try {
             // noinspection SynchronizationOnLocalVariableOrMethodParameter
@@ -156,6 +156,27 @@ public class GrpcUtil {
         } catch (final Exception err) {
             log.error().append("Unanticipated gRPC Error: ").append(err).endl();
         }
+    }
+
+    public static <T> void safelyOnNext(StreamObserver<T> observer, T message) {
+        safelyExecuteLocked(observer, () -> observer.onNext(message));
+    }
+
+    public static <T> void safelyComplete(StreamObserver<T> observer, T message) {
+        safelyExecuteLocked(observer, () -> {
+            observer.onNext(message);
+            observer.onCompleted();
+        });
+    }
+
+    public static void safelyComplete(StreamObserver<?> observer) {
+        safelyExecuteLocked(observer, observer::onCompleted);
+    }
+
+    public static void safelyError(StreamObserver<?> observer, Throwable error) {
+        safelyExecuteLocked(observer, () -> {
+            observer.onError(error);
+        });
     }
 
     /**
--- a/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/GrpcUtil.java
+++ b/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/GrpcUtil.java
@@ -17,7 +17,6 @@ import io.grpc.stub.StreamObserver;
 import java.io.IOException;
 import java.util.UUID;
 import java.util.concurrent.Callable;
-import java.util.function.Function;
 
 public class GrpcUtil {
     private static final Logger log = LoggerFactory.getLogger(GrpcUtil.class);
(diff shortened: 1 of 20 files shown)


server/src/main/java/io/deephaven/server/session/SessionState.java
```diff
@@ -761,6 +761,7 @@ private synchronized void setState(final ExportNotification.State state) {
                 if (errorId == null) {
                     assignErrorId();
                 }
+                //TODO do we want logs about this?
                 safelyExecute(() -> errorHandler.onError(state, errorId, caughtException, dependentHandle));
             }
```
niloc132, line 766: TODO added because this is the only remaining callsite for safelyExecute - like the python autocomplete code, I'm thinking we'll replace it with a try/catch. But do we want to log the error in this case, or are we only intending to propagate it to the client?

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/GrpcUtil.java
```diff
@@ -146,7 +146,7 @@ public static void safelyExecute(final FunctionalInterfaces.ThrowingRunnable<Exc
      *
      * @param runner the runnable to execute safely
      */
-    public static void safelyExecuteLocked(final Object lockedObject,
+    private static void safelyExecuteLocked(final Object lockedObject,
```
devinrsmith, line 149: Did you want to change the logging from error back to debug?

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/GrpcUtil.java
```diff
@@ -158,6 +158,27 @@ public static void safelyExecuteLocked(final Object lockedObject,
...
     }
 
+    public static <T> void safelyOnNext(StreamObserver<T> observer, T message) {
+        safelyExecuteLocked(observer, () -> observer.onNext(message));
+    }
+
+    public static <T> void safelyComplete(StreamObserver<T> observer, T message) {
+        safelyExecuteLocked(observer, () -> {
+            observer.onNext(message);
+            observer.onCompleted();
```
devinrsmith, line 168: I don't think we've done this before... but do we have a responsibility to call `observer.onError` if `observer.onNext` throws?

devinrsmith (commented):
Looks great. Stronger interface allowing callers to not do their own lambdas is much cleaner.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/GrpcUtil.java
niloc132: No, if it throws, the connection is gone, no need to tell it that the connection is gone.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/GrpcUtil.java
niloc132: Yes, but have to finish killing these calls first

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/GrpcUtil.java
devinrsmith: I wonder if it makes sense to add a comment to that effect.

At least the javadoc for StreamObserver#onNext has:

```
If an exception is thrown by an implementation the caller is expected to terminate the
stream by calling {@link #onError(Throwable)} with the caught exception prior to
propagating it.
```

server/src/main/java/io/deephaven/server/session/SessionState.java
nbauernfeind: IMO, SessionState changes belong in a stream table to make client debugging a lot easier. I actually have a branch that does this. It has been helpful to debug api client issues for the c++ and go clients. I have not yet put it up for a PR as the web-ui is extremely annoying (re-creating exports of the lastBy version of the table makes the table unusable without also filtering out the web-ui's session). State change logging is extremely verbose and hard to follow from the log -- so the table is a real winner in that regard. If we only log errors here (and not requests and other state changes) then there is not enough context here to understand what failed or why. Most errors will have already logged if they were not due to API issues (i.e. uncaught exceptions will be logged, but illegal parameters and other StatusRuntimeException will not be logged). 

I prefer safely ignoring any onError failure here.

niloc132:
Pushed a commit that I think covers all the java changes I expect to make at this time - still going to add javadoc for the new calls.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/GrpcUtil.java
niloc132: I believe @nbauernfeind and I went back and forth on this, but eventually dug into the implementation to confirm, since we found the docs to be vague. I double checked it now: in short:
 * ServerCalls.ServerCallStreamObserverImpl.onNext checks if already canceled/aborted/completed and throws back to you if so. Clearly a caller of onNext need not call onError in this case.
 * Next, ServerCallImpl.sendHeaders is called, which delegates to sendHeadersInternal to encode, and eventually delegates down to AbstractServerStream.Sink.writeHeaders. From experience, I can tell you that this method is not permitted to throw, bad things will happen and state will be broken.
 * Last, ServerCallImpl.sendMessage delegates to sendMessageInternal catches its own exceptions from sending a payload out to transport, and calls close() on the stream.

Note that I'm only here concerned with calling onNext and onError on the instance of the StreamObserver provided by grpc to our service impl. If you are in a situation where you have both a responseStreamObserver and requestStreamObserver, and fail when calling one side's onNext, you are clearly obligated to call the other side's onError. We don't write a lot of code like that, but it is my guess that this case is what the javadoc is referring to.

devinrsmith (dismissed):

server/src/main/java/io/deephaven/server/session/SessionState.java
```diff
@@ -761,7 +761,11 @@ private synchronized void setState(final ExportNotification.State state) {
                 if (errorId == null) {
                     assignErrorId();
                 }
-                safelyExecute(() -> errorHandler.onError(state, errorId, caughtException, dependentHandle));
+                try {
+                    errorHandler.onError(state, errorId, caughtException, dependentHandle);
+                } catch (final Exception err) {
+                    log.debug().append("Unexpected error while reporting state failure: ").append(err).endl();
```
nbauernfeind, line 767: Oops, I think we want this log to shout as `log.error()` -- since it implies that the error-handler itself is broken.

nbauernfeind (approved):

Issue #3248, opened 2022-12-29
Test rollups and trees


rcaudy:
With the merge of #3243, we have a complete implementation for rollups and trees, along with a number of other features and improvements. Testing, however, remains incomplete, and unit testing focused on the newly-available features is negligible.

This ticket should cover:
* Further integration testing of rollups
* Integration testing of trees
* Development of a server-side snapshot test tool that will snapshot a hierarchical table and produce a table containing the resulting chunk data
* Unit testing for rollups and trees
* Fixing bugs found in testing

rcaudy:
@niloc132 managed to get this error by rapidly collapsing and expanding rollup nodes:
```
2023-01-13T16:48:07.630Z | heduler-Concurrent-1 | ERROR | icalTableViewSubscription | Internal Error '1c0206a4-bba4-4564-bd0c-bfe6c1b90876' io.deephaven.engine.table.impl.hierarchical.HierarchicalTableSnapshotException: HierarchicalTable.snapshot error
        at io.deephaven.engine.table.impl.hierarchical.HierarchicalTableImpl.visitExpandedNode(HierarchicalTableImpl.java:1274)
        at io.deephaven.engine.table.impl.hierarchical.HierarchicalTableImpl.visitExpandedNode(HierarchicalTableImpl.java:1126)
        at io.deephaven.engine.table.impl.hierarchical.HierarchicalTableImpl.visitExpandedNode(HierarchicalTableImpl.java:1257)
        at io.deephaven.engine.table.impl.hierarchical.HierarchicalTableImpl.visitExpandedNode(HierarchicalTableImpl.java:1126)
        at io.deephaven.engine.table.impl.hierarchical.HierarchicalTableImpl.visitExpandedNode(HierarchicalTableImpl.java:1257)
        at io.deephaven.engine.table.impl.hierarchical.HierarchicalTableImpl.traverseExpansionsAndFillSnapshotChunks(HierarchicalTableImpl.java:1120)
        at io.deephaven.engine.table.impl.hierarchical.HierarchicalTableImpl.lambda$snapshotData$5(HierarchicalTableImpl.java:1086)
        at io.deephaven.engine.table.impl.remote.ConstructSnapshot.callDataSnapshotFunction(ConstructSnapshot.java:1193)
        at io.deephaven.engine.table.impl.remote.ConstructSnapshot.callDataSnapshotFunction(ConstructSnapshot.java:1035)
[...]

PR #3284, opened 2023-01-09
Iteration and fixes from testing rollups and trees

### LINKED ISSUE #3275: ConstructSnapshot log spam should be reduced
We log a lot of rarely helpful information from `ConstructSnapshot`, and it's especially noisy from use cases like trees, rollups, and partitioned table transforms/proxy ops.

We should either disable it or move it to debug level.

Examples:
```
[UpdateGraphProcessor-unitTestRefresh-1] INFO io.deephaven.engine.table.impl.remote.ConstructSnapshot - UpdateView non-concurrent Snapshot Function took 0ms
[UpdateGraphProcessor-unitTestRefresh-1] INFO io.deephaven.engine.table.impl.remote.ConstructSnapshot - UpdateView Already held lock, proceeding to locked snapshot
```

rcaudy:
Part of #3248
Closes #3275 

 * Iterate on the gRPC API for hierarchical tables
 * Fix server-side bugs for rollups/trees
 * Address *some* unit testing for rollups/trees
 * Refactor barrage schema utilities
 * Clean up column formatting
 * Dial `ConstructSnapshot` spam down to zero


engine/table/src/main/java/io/deephaven/engine/table/impl/TableDefaults.java
```diff
@@ -307,7 +307,7 @@ default Table renameAllColumns(UnaryOperator<String> renameFunction) {
     @ConcurrentMethod
     @FinalDefault
     default Table formatRowWhere(String condition, String formula) {
-        return formatColumnWhere(ColumnFormattingValues.ROW_FORMAT_NAME, condition, formula);
+        return formatColumnWhere("*", condition, formula);
```
cpwright, line 310: This seems less clear than a shared constant.

engine/table/src/main/java/io/deephaven/engine/table/impl/hierarchical/TreeTableFilter.java
```diff
@@ -106,12 +111,6 @@ public MemoizedOperationKey getMemoKey() {
      */
     private final ColumnSource parentIdSource;
 
-    /**
```
cpwright, line 109: Remind me what magic you did to make the SwapListener not necessary for referential integrity.

engine/table/src/main/java/io/deephaven/engine/util/ColumnFormatting.java
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Returns the base column name from a formatting column name.
+     *
+     * @param columnName the column name
+     * @return the base column name formatted by {@code columnName}
+     */
+    static String getFormatBaseColumn(@NotNull final String columnName) {
+
+        if (columnName.startsWith(Constants.ROW_FORMAT_NAME)) {
+            return "*";
```
cpwright, line 111: So you have this "__ROW" constant; but then you return "*" and have hardcoded star elsewhere.

nbauernfeind (commented):
LGTM. No Comments!

engine/table/src/main/java/io/deephaven/engine/table/impl/hierarchical/TreeTableFilter.java
rcaudy: The SwapListener is held during initialization, but disappears from the DAG in its `end` method on success, by swapping the referent of the parent's `SimpleReference` to the listener from itself to the eventual listener.

Specifically, L145-147:
```
            referenceForSource.swapDelegate(initialDelegate, eventualListener instanceof LegacyListenerAdapter
                    ? (LegacyListenerAdapter) eventualListener
                    : new WeakSimpleReference<>(eventualListener));
```

engine/table/src/main/java/io/deephaven/engine/table/impl/TableDefaults.java
rcaudy: 1. `"*"` seems to be the wildcard we expect users to enter manually, if they use `formatColumns` instead of `formatRowWhere`
2. The previous version of the code was mapping the constant to the wildcard a layer deeper, and I (think) I was trying to standardize a bit.
3. I've exposed and used a constant for this value, as the "user-visible" version.

engine/table/src/main/java/io/deephaven/engine/util/ColumnFormatting.java
rcaudy: Well, less me than "the code base". This is one of those scenarios where I tried to clean things up a bit without going "all in." In this particular case, I just moved the code to a single place from `SelectColumnFactory`.

To the best of my understanding, we use the "__ROW" prefix internally, since "*" isn't a valid column name, but use the "*" name more externally and expect users to input that.

cpwright (approved):

nbauernfeind (approved):

nbauernfeind (approved):

Issue #1763, opened 2021-12-29
JS API Hierarchical Table support


pete-petey:
Update the dh.TreeTable type to use the new gRPC services around hierarchical tables. These types behave largely like typical tables, with a few exceptions:
 * custom columns are not supported at this time
 * size is not available until a viewport has been set (due to the fact that "size" doesn't make sense without expand/collapse state), and will be reported as -1 until then

Typical tables can be transformed into hierarchical tables through the use of table.rollup() and table.treeTable(). Hierarchical tables in scope can also be fetched by name.

niloc132:
This should be safe to close now.

PR #3290, opened 2023-01-11
JS API for Hierarchical Tables

niloc132:


web/client-api/src/main/java/io/deephaven/web/client/api/barrage/def/ColumnDefinition.java
```diff
@@ -3,6 +3,10 @@
  */
 package io.deephaven.web.client.api.barrage.def;
 
+import io.deephaven.web.client.api.Column;
```
mofojed, line 6: This seems kind of backwards; would expect `Column` to import from `ColumnDefinition`, rather than `api.barrage.def.ColumnDefinition` importing from higher up on the package chain. Somewhat circular.

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsRollupConfig.java
```diff
@@ -43,23 +59,188 @@ public JsRollupConfig(JsPropertyMap<Object> source) {
...
+                }
+                // case JsAggregationOperation.SORTED_FIRST: {
+                // // TODO support this
+                // }
+                // case JsAggregationOperation.SORTED_LAST: {
+                // // TODO support this
+                // }
+                // case JsAggregationOperation.WSUM: {
+                // // TODO support this
+                // }
```
mofojed, line 210: Add a ticket ref for the TODO.
Also `default: JsLog.warn(...)` ?

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -209,455 +232,454 @@ public TreeRow(int offsetInSnapshot, Object[] dataColumns, Object rowStyleColumn
...
-                old.run();
-                r.run();
-            };
-        } else {
-            queuedOperations = () -> {
-                nextRequestOps[nextRequestOps.length] = operation;
-                r.run();
-            };
+    private BitSet makeColumnSubscriptionBitset() {
+        // Build the bitset for the columns that are needed to get the data, style, and maintain structure
```
mofojed, line 662: Should be a JavaDoc:
```suggestion
    /**
     * Build the bitset for the columns that are needed to get the data, style, and maintain structure
     * @return BitSet for the columns
     */
    private BitSet makeColumnSubscriptionBitset() {
```

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -209,455 +232,454 @@ public TreeRow(int offsetInSnapshot, Object[] dataColumns, Object rowStyleColumn
...
         // if requested to fire the event, or if the data has changed in some way, fire the event
-        final boolean fireEvent = alwaysFireEvent || !vd.containsSameDataAs(currentViewportData);
-
-        this.currentViewportData = vd;
-
-        Set<Key> presentKeys = new HashSet<>();
-        for (int i = 0; i < result.getTableDetails().length; i++) {
-            TableDetails detail = result.getTableDetails()[i];
-            TreeNodeState treeNodeState = expandedMap.get(detail.getKey());
+        final boolean fireEvent = true;// alwaysFireEvent || !viewportData.containsSameDataAs(currentViewportData);
```
mofojed, line 647: Revert this line
```suggestion
        final boolean fireEvent = alwaysFireEvent || !viewportData.containsSameDataAs(currentViewportData);
```

web/client-api/src/main/java/io/deephaven/web/client/api/tree/enums/JsAggregationOperation.java
```diff
@@ -22,5 +24,9 @@ public class JsAggregationOperation {
             UNIQUE = "Unique",
             // Array operation isn't legal in all contexts, just omit it for now
             // ARRAY = "Array",
+            // These need some other parameter to function, not supported yet
+            // SORTED_FIRST="SortedFirst",
+            // SORTED_LAST="SortedLast",
+            // WSUM = "WeightedSum";
             SKIP = "Skip";
```
mofojed, line 31: You removed this aggregation, should it be removed from here as well? Web UI does reference it in one location, but that could be removed too... your call: https://github.com/deephaven/web-client-ui/blob/main/packages/iris-grid/src/sidebar/aggregations/AggregationUtils.ts#L48

web/client-api/src/main/java/io/deephaven/web/public/rollup.html
```diff
@@ -73,183 +73,138 @@ <h3>Rollup Table Configuration</h3>
...
-        tableElt.innerText = t;
-        tables.appendChild(tableElt);
-        tableElt.onclick = e => {
-          e.stopPropagation();
-          queryInfo.getTable(t).then(table => displayTableDetails(table));
-          document.querySelectorAll('.selected').forEach(elt => {
-            elt.classList.remove('selected');
-          });
-          e.target.classList.add('selected');
+    var connection;
```
mofojed, line 76: These pages be linked from `index.html`

web/client-api/src/main/java/io/deephaven/web/public/treetable.html
```diff
@@ -17,56 +17,51 @@
...
-    });
-  });
-  function addQuery(queryInfo) {
-    if (queryInfo.name === 'SimpleQueries') {
-      if (queryInfo.status === 'Running') {
-        queryInfo.getTreeTable('TreeTable').then(treeTable => {
-          // Note that this class is in treegrid.js
-          var tableElt = new TreeGrid(treeTable).element;
-          document.body.appendChild(tableElt);
+    var connection;
```
mofojed, line 20: Page should be linked from `index.html`

web/client-ui/Dockerfile
```diff
@@ -2,9 +2,9 @@ FROM deephaven/node:local-build
 WORKDIR /usr/src/app
 
 # Most of the time, these versions are the same, except in cases where a patch only affects one of the packages
-ARG WEB_VERSION=0.25.0
-ARG GRID_VERSION=0.25.0
-ARG CHART_VERSION=0.25.0
+ARG WEB_VERSION=0.24.1-hierarchical-table.0
+ARG GRID_VERSION=0.24.1-hierarchical-table.1
+ARG CHART_VERSION=0.24.1-hierarchical-table.0
```
mofojed, line 7: I'll cut a release shortly so you can update this version

mofojed (changes requested):
Would like more details in the PR description for posterity. What the change is, Fixes #1763, and possibly some code snippets that were run to test.

I think there's a bug somewhere too. Running a snippet like this just gives me a blank table:
```
t2 = emptyTable(100).update("I=`` + i", "J = `` + `` + (i/10)")
h2 = t2.tree("I", "J")
```
Results in:
image

Also - are there any python snippets? Specifically, how would I convert something like this to python? I've only tested with Groovy so far:
```
t = emptyTable(100).update("I=i", "J = `` + (int)(i/10)")
h = t.rollup([io.deephaven.api.agg.Aggregation.of(io.deephaven.api.agg.spec.AggSpec.sum(), "I")], "J")
```

web/client-api/src/main/java/io/deephaven/web/client/api/barrage/def/ColumnDefinition.java
niloc132: Well, Column is the combined data from multiple ColumnDefinitions, so it doesnt have a single ColumnDefinition backing it.

web/client-api/src/main/java/io/deephaven/web/client/api/barrage/def/ColumnDefinition.java
```diff
@@ -137,4 +140,85 @@ public void setDescription(String description) {
     public String getDescription() {
         return description;
     }
+
+    public Column makeJsColumn(int index, Map<Boolean, Map<String, ColumnDefinition>> map) {
```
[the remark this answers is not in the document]
niloc132: There is no circular import, Column doesnt depend on ColumnDefinition at all. 

ColumnDefinition is broadly a strongly typed version of the barrage data that is sent over the wire, and Column is somewhere between 1 and ... 8? different ColumnDefinitions smushed into "here's the state of the table as far as JS is concerned".

Additionally, ColumnDefinition is a holdover from pre-barrage, so likely should be removed entirely when we reimplement barrage/flight for JS, but we don't expect to make time for that soon.

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsRollupConfig.java
```diff
@@ -43,23 +59,188 @@ public JsRollupConfig(JsPropertyMap<Object> source) {
...
             includeOriginalColumns = source.getAsAny("includeOriginalColumns").asBoolean();
+            if (!includeOriginalColumns) {
+                JsLog.warn("includeOriginalColumns=false will be ignored");
+            }
         }
         if (source.has("includeDescriptions")) {
             includeDescriptions = source.getAsAny("includeDescriptions").asBoolean();
+            if (!includeDescriptions) {
+                JsLog.warn("includeDescriptions=false will be ignored");
+            }
```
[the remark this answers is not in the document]
niloc132: Good question. I think we can paper over this change if clients (web ui) wants us to, and we can go further and include constituent columns in the table definition that werent grouped or aggregated as well. My inclination was to track the state and at least warn the developer that we were ignoring their settings, but we could actually remove the field from docs too if you don't see value.

web/client-api/src/main/java/io/deephaven/web/client/api/tree/enums/JsAggregationOperation.java
niloc132: I assumed it wasnt safe/nice to outright remove it, though I will make sure we warn if it is used. I'll deprecated it here for later removal?

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -209,455 +235,459 @@ public TreeRow(int offsetInSnapshot, Object[] dataColumns, Object rowStyleColumn
...
-            return;
+
+        // track columns by name and freeze the array to avoid defensive copies
+        this.visibleColumns = JsObject.freeze(columns);
+        for (int i = 0; i < visibleColumns.length; i++) {
+            Column column = visibleColumns[i];
+            columnsByName.put(column.getName(), column);
+        }
+
+
```
mofojed, line 413: Extra new line not really necessary?
```suggestion
```

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -209,455 +235,459 @@ public TreeRow(int offsetInSnapshot, Object[] dataColumns, Object rowStyleColumn
...
+    // The source JsTable behind the original HierarchicalTable, lazily built at this time
+    private final JsLazy<Promise<JsTable>> sourceTable;
+
+    // The current filter and sort state
     private List<FilterCondition> filters = new ArrayList<>();
     private List<Sort> sorts = new ArrayList<>();
+    private TicketAndPromise filteredTable;
+    private TicketAndPromise sortedTable;
+
+    private final Object[][] keyTableData;
```
mofojed, line 286: Could use a comment here about what this is for and why the size is 2 bigger than keyColumns, as it's not clear from just the name.

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -209,455 +235,459 @@ public TreeRow(int offsetInSnapshot, Object[] dataColumns, Object rowStyleColumn
...
+                                    headers -> connection.flightServiceClient().doExchange(headers),
+                                    (first, headers) -> connection.browserFlightServiceClient().openDoExchange(first,
+                                            headers),
+                                    (next, headers, c) -> connection.browserFlightServiceClient().nextDoExchange(next,
+                                            headers,
+                                            c::apply),
+                                    new FlightData());
+
+                    FlightData subscriptionRequestWrapper = new FlightData();
+                    Builder doGetRequest = new Builder(1024);
```
mofojed, line 548: Why 1024? Maybe named const to make it clear why that number picked

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -209,455 +235,459 @@ public TreeRow(int offsetInSnapshot, Object[] dataColumns, Object rowStyleColumn
...
+                    BarrageSubscriptionRequest.addSubscriptionOptions(doGetRequest, serializationOptionsOffset);
+                    BarrageSubscriptionRequest.addViewport(doGetRequest, viewportOffset);
+                    doGetRequest.finish(BarrageSubscriptionRequest.endBarrageSubscriptionRequest(doGetRequest));
+
+                    subscriptionRequestWrapper.setAppMetadata(
+                            WebBarrageUtils.wrapMessage(doGetRequest, BarrageMessageType.BarrageSubscriptionRequest));
+                    doExchange.send(subscriptionRequestWrapper);
+
+                    // hang on to this to get server-sent snapshots, cancel when needed
+                    // doExchange.end();
```
mofojed, line 573: Seems you don't need the `// doExchagne.end();` line here
```suggestion
                    // hang on to the doExchange stream to get server-sent snapshots, cancel when needed
                    // doExchange.end();
```

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -676,38 +706,41 @@ public void setExpanded(Object row, boolean isExpanded) {
...
+            viewTicket = null;
+        }
+        if (stream != null) {
+            stream.then(stream -> {
+                stream.end();
+                stream.cancel();
+                return null;
+            });
+            stream = null;
+        }
```
mofojed, line 736: This block could be a private method `closeStream()`, the exact same block is in multiple locations.

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -676,38 +706,41 @@ public void setExpanded(Object row, boolean isExpanded) {
...
+                t.close();
+                return null;
+            });
+            keyTable = null;
+        }
+
+        if (viewTicket != null) {
+            viewTicket.release();
+            viewTicket = null;
+        }
```
mofojed, line 728: `viewTicket`, `filteredTable`, `sortedTable` all have blocks like this that are in applyFilter, applySort, close; could refactor them into private methods `releaseViewTicket`, `releaseFilteredTable`, `releaseSortedTable` to clean that up a bit.

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
mofojed: Okay cool can you remove the comment after then and capture in a TODO then?

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -722,10 +755,18 @@ public void setViewport(double firstRow, double lastRow, @JsOptional JsArray<Col
             @JsOptional Double updateInterval) {
         this.firstRow = firstRow;
         this.lastRow = lastRow;
-        this.columns = columns != null ? Js.uncheckedCast(columns.slice()) : baseTable.getColumns();
+        this.columns = columns != null ? Js.uncheckedCast(columns.slice()) : visibleColumns;
```
mofojed, line 758: May want to freeze the columns.slice() here for consistency; visibleColumns is already frozen. Doesn't get passed to UI though so doesn't really matter.

mofojed (commented):
Some clean up, only thing that may need attention is the if/else-if for rowDepthColumn vs. expandBycolumn

web/client-ui/Dockerfile
mofojed, line 7: Pull this out of your change - I'll publish v0.26.0 shortly and update the version in core in a separate ticket (linking to release notes etc).

mofojed (changes requested):

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -209,455 +235,459 @@ public TreeRow(int offsetInSnapshot, Object[] dataColumns, Object rowStyleColumn
...
-    public Promise<JsTreeTable> finishFetch() {
-        RollupDefinition rollupDefinition = baseTable.getTableDef().getAttributes().getRollupDefinition();
-        if (rollupDefinition == null) {
-            return Promise.resolve(this);
-        }
+            if (definition.isHierarchicalRowDepthColumn()) {
+                rowDepthCol = column;
+            } else if (definition.isHierarchicalRowExpandedColumn()) {
+                rowExpandedCol = column;
+            } else if (definition.isHierarchicalExpandByColumn()) {
```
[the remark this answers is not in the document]
niloc132: They are, but from our perspective they are mutually exclusive. Technically this isn't 100% true, but it is accurate enough to not need another overhaul. Ryan and I discussed a different metadata format, but ultimately decided against it, at least for now

web/client-ui/Dockerfile
mofojed: v0.26.0 has been published and added to core already

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
niloc132: Ah, all the flatbuffer builders across the codebase are 1024, because that's what the docs show. Apparently this is explicit in other languages too, the api is pretty much the same in all languages.

https://google.github.io/flatbuffers/flatbuffers_guide_tutorial.html

web/client-ui/Dockerfile
niloc132: I've removed these patches, so will just lean on upstream.

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
niloc132: Yeah, I feel like I'm still going to hate it, since there is a "chain" of them, and depending on how far up the "chain" you are, the rest need to be released too. I'm going to see about an enum instead.

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
niloc132: Done, #3310

mofojed:
For posterity, the snippets I used (Groovy):
```
t = emptyTable(100).update("I=i", "J = `` + (int)(i/10)")
h = t.rollup([io.deephaven.api.agg.Aggregation.of(io.deephaven.api.agg.spec.AggSpec.sum(), "I")], "J")
```

```
t2 = emptyTable(100).update("I=`` + i", "J = i == 0 ? null : `` + (int)(i/10)")
h2 = t2.tree("I", "J")
```

niloc132:
> For posterity, the snippets I used (Groovy):
> 
> ```
> t = emptyTable(100).update("I=i", "J = `` + (int)(i/10)")
> h = t.rollup([io.deephaven.api.agg.Aggregation.of(io.deephaven.api.agg.spec.AggSpec.sum(), "I")], "J")
> ```
> 
> ```
> t2 = emptyTable(100).update("I=`` + i", "J = i == 0 ? null : `` + (int)(i/10)")
> h2 = t2.tree("I", "J")
> ```

Thanks, I should fix the example python

mofojed (commented):
I think we should remove `JsTreeTable.getTotalsTable` as well. Right now aggregations are appearing on rollups created with code, and UI is trying to apply a totals table config which is wrong.
Not sure it can actually add aggregations to the tree table when the rollup/tree is created from code rather than from the UI?

mofojed (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2067
Conceptual: https://github.com/deephaven/deephaven.io/issues/2065
Reference: https://github.com/deephaven/deephaven.io/issues/2064
Blog: https://github.com/deephaven/deephaven.io/issues/2066

PR #3292, opened 2023-01-11
Add input_table()

### LINKED ISSUE #3291: Add input_table() factory function for InputTable to be consistent with existing ones.
It is weird to not have an input_table() function for creating a InputTable while empty_table/time_table exist for creating Table.

jmao-denver:
Also replace InputTable at the top level with input_table

Fixes #3291

### DIFF
--- a/py/server/deephaven/__init__.py
+++ b/py/server/deephaven/__init__.py
@@ -21,6 +21,6 @@ from .csv import read as read_csv
 from .csv import write as write_csv
 from .stream.kafka import consumer as kafka_consumer
 from .stream.kafka import producer as kafka_producer
-from .table_factory import empty_table, time_table, merge, merge_sorted, new_table, DynamicTableWriter, InputTable
+from .table_factory import empty_table, time_table, merge, merge_sorted, new_table, DynamicTableWriter, input_table
 from .replay import TableReplayer
 from ._gc import garbage_collect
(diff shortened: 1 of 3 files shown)


py/server/deephaven/__init__.py
```diff
@@ -21,6 +21,6 @@
 from .csv import write as write_csv
 from .stream.kafka import consumer as kafka_consumer
 from .stream.kafka import producer as kafka_producer
-from .table_factory import empty_table, time_table, merge, merge_sorted, new_table, DynamicTableWriter, InputTable
+from .table_factory import empty_table, time_table, merge, merge_sorted, new_table, DynamicTableWriter, input_table
```
mofojed, line 24: We should clearly mark this as a breaking change, and include in the release notes. Is there a process for that in core (@rcaudy)?

For web-client-ui, we label the PR `breaking` and include in the PR description details about the breakage/migration instructions: https://github.com/deephaven/web-client-ui/pull/881
These then get included in the release notes: https://github.com/deephaven/web-client-ui/releases/tag/v0.22.0

mofojed (approved):
Looks good, would just like to know about how we mark the breaking change clearly. I guess it hasn't actually been documented yet so doesn't matter, @jjbrosnan will just need to be aware for the docs he is currently writing.

jjbrosnan (approved):

py/server/deephaven/__init__.py
rcaudy: Sounds like a fine process. Looks like @jmao-denver has tagged this "ReleaseNotesNeeded". I added "breaking" just now.

PR #3299, opened 2023-01-13
SessionService#release - Quietly and Gracefully handle OutOfOrder/Premature Releases/Cancels

nbauernfeind:
@niloc132 was running into this in some work of his where his cancel/release would race with him defining what the export is actually going to export.

### DIFF
--- a/server/src/main/java/io/deephaven/server/session/SessionState.java
+++ b/server/src/main/java/io/deephaven/server/session/SessionState.java
@@ -514,6 +514,9 @@ public class SessionState {
         private volatile ExportNotification.State state = ExportNotification.State.UNKNOWN;
         private volatile int exportListenerVersion = 0;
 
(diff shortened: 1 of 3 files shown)


niloc132 (approved):

server/src/main/java/io/deephaven/server/session/SessionState.java
```diff
@@ -514,6 +514,9 @@ public static boolean isExportStateTerminal(final ExportNotification.State state
         private volatile ExportNotification.State state = ExportNotification.State.UNKNOWN;
         private volatile int exportListenerVersion = 0;
 
+        /** Indicates whether this export has already been well defined. This prevents export object reuse. */
+        private volatile boolean hasHadWorkSet = false;
```
rcaudy, line 518: Need not be `volatile`, you access it in a `synchronized` method.

rcaudy (approved):

PR #3308, opened 2023-01-14
Update start script to support extending the classpath

rbasralian:
Update start script to support extending the classpath based on the `EXTRA_CLASSPATH` environment variable. Use `/apps/libs/*` as default `EXTRA_CLASSPATH` in test images.

### DIFF
--- a/buildSrc/src/main/resources/unixStartScript.txt
+++ b/buildSrc/src/main/resources/unixStartScript.txt
@@ -165,11 +165,9 @@ case "\$( uname )" in                #(
   NONSTOP* )        nonstop=true ;;
 esac
 
-if [ -z "\$EXTRA_CLASSPATH" ]; then
-    CLASSPATH=$classpath
-else
-    CLASSPATH="\$EXTRA_CLASSPATH":$classpath
-fi
+# Deephaven edit.
+# Prepend EXTRA_CLASSPATH and colon if EXTRA_CLASSPATH is non-empty.
+CLASSPATH=\${EXTRA_CLASSPATH:+"\$EXTRA_CLASSPATH":}$classpath
 
 <% if ( mainClassName.startsWith('--module ') ) { %>
 MODULE_PATH=$modulePath
--- a/buildSrc/src/main/resources/unixStartScript.txt
+++ b/buildSrc/src/main/resources/unixStartScript.txt
@@ -165,11 +165,11 @@ case "\$( uname )" in                #(
   NONSTOP* )        nonstop=true ;;
 esac
 
-EXTRA_CLASSPATHS="\${EXTRA_CLASSPATHS:-/apps/libs/*:}"
-echo "Extra classpaths: \$EXTRA_CLASSPATHS"
-
-CLASSPATH="\$EXTRA_CLASSPATHS":$classpath
-echo "Final classpath: \$CLASSPATH"
+if [ -z "\$EXTRA_CLASSPATH" ]; then
+    CLASSPATH=$classpath
+else
+    CLASSPATH="\$EXTRA_CLASSPATH":$classpath
+fi
 
 <% if ( mainClassName.startsWith('--module ') ) { %>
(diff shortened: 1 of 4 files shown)


devinrsmith:
Note: we don't have "custom extended classpath" support anymore, at least out of the box with our existing docker images. It would be relatively easy to add back in; and we'd want to explicitly mention it as part of https://deephaven.io/core/docs/how-to-guides/configuration/native-application/.

devinrsmith:
Instead of naming a specific directory, I'd suggest using an environment variable `EXTRA_CLASSPATH`, or something like that, as it would better fit in with the nature of how we use environment variables wrt the start script.

devinrsmith (changes requested):
Ah - I thought this was an "issue" at first.

I think we can merge something like this in. But I think it's best to use an optional environment variable "EXTRA_CLASSPATH" if set (otherwhise empty) or something like that.

There's also a little blurb of DH specific documentation in this file, we should mention that we are adding the ability to change the classpath.

devinrsmith:
I _think_ it makes sense to prepend the extra classpath (as your PR suggests) instead of appending the extra classpath - but it's a worthwhile point of potential discussion.

rbasralian:
> I _think_ it makes sense to prepend the extra classpath (as your PR suggests) instead of appending the extra classpath - but it's a worthwhile point of potential discussion.

I definitely think prepending is better. It lets you do some risky stuff (e.g. replacing our dependencies), but that is better than _not_ letting someone do that when they want to. (It's also in line with how DHE works, and we've never contemplated changing it there.)


I did change it to take an `EXTRA_CLASSPATHS` environment variable, but kept `/apps/libs/*` as the default, since our documentation already claims that it works, and there's no harm in having a useful default.

devinrsmith:
I think it's best to not have a default, and instead update the documentation to reflect the extra environment variable as necessary if you want to extend the classpath. (As a better follow-up, I think we need new documentation around how to do this "properly": https://github.com/deephaven/deephaven.io/issues/2091.)

I also think we shouldn't be printing anything out in this script.

devinrsmith:
I also mentioned in DM, but the start script is for *native* users too, not just docker. In which case, I think setting a default at the script level is inappropriate; but we _can_ set a default higher up in the docker image layers.

buildSrc/src/main/resources/unixStartScript.txt
```diff
@@ -165,7 +165,12 @@ case "\$( uname )" in                #(
   NONSTOP* )        nonstop=true ;;
 esac
 
-CLASSPATH=$classpath
+if [ -z "\$EXTRA_CLASSPATH" ]; then
+    CLASSPATH=$classpath
+else
+    CLASSPATH="\$EXTRA_CLASSPATH":$classpath
+fi
```
devinrsmith, line 172: A bit nit-picky, but can you add this comment and try this combined version with your test setup? (I've verified to it seems to work for me.) `$classpath` is potentially the majority of the script length and if we can do without doubling it that would be good.

```shell
# Deephaven edit.
# Prepend EXTRA_CLASSPATH and colon if EXTRA_CLASSPATH is non-empty.
CLASSPATH=\${EXTRA_CLASSPATH:+\$EXTRA_CLASSPATH:}$classpath
```

devinrsmith (changes requested):

devinrsmith (approved):
Please change your commit message to explain `EXTRA_CLASSPATH` - the `/apps/libs/*` is incidental in this case; the `/apps/libs/*` message will be on the deephaven-server-docker side of things.

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2103
Conceptual: https://github.com/deephaven/deephaven.io/issues/2101
Reference: https://github.com/deephaven/deephaven.io/issues/2102
Blog: https://github.com/deephaven/deephaven.io/issues/2100

PR #3309, opened 2023-01-14
Port DH-8170: OuterJoinTools including bugfix DH-12658

nbauernfeind:
Note that fullOuterJoin is an alternative implementation that should A) be more efficient when table2 is refreshing, and B) behaves like join/naturalJoin/etc w.r.t. columns to add.

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/util/OuterJoinTools.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/OuterJoinTools.java
@@ -13,6 +13,7 @@ import io.deephaven.engine.table.Table;
 import io.deephaven.engine.table.impl.CrossJoinHelper;
 import io.deephaven.engine.table.impl.QueryTable;
 import io.deephaven.engine.table.impl.select.MatchPairFactory;
+import io.deephaven.engine.table.impl.select.NullSelectColumn;
 import io.deephaven.engine.table.impl.select.SelectColumn;
 import io.deephaven.engine.table.impl.select.SourceColumn;
 import io.deephaven.engine.table.impl.sources.NullValueColumnSource;
@@ -86,7 +87,8 @@ public class OuterJoinTools {
                     (QueryTable) leftWithSentinel,
                     matchPairs,
                     createColumnsToAdd(leftWithSentinel, matchPairs, MatchPair.ZERO_LENGTH_MATCH_PAIR_ARRAY),
-                    CrossJoinHelper.DEFAULT_NUM_RIGHT_BITS_TO_RESERVE).where(sentinelColumnName + " == null")
+                    CrossJoinHelper.DEFAULT_NUM_RIGHT_BITS_TO_RESERVE)
+                    .where(sentinelColumnName + " == null")
                     .dropColumns(sentinelColumnName);
 
             return TableTools.merge(leftTable, rightTable);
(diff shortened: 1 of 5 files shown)


engine/table/src/main/java/io/deephaven/engine/util/OuterJoinTools.java
```diff
@@ -0,0 +1,410 @@
...
+        // handle the simpler scenario where no conflicts may exist otherwise leftOuterJoin would have failed
+        if (columnsToAdd.length == 0) {
+            final MatchPair[] matchPairs = Arrays.stream(columnsToMatch)
+                    .map(mp -> new MatchPair(mp.rightColumn(), mp.leftColumn()))
+                    .toArray(MatchPair[]::new);
+
+            final Table leftWithSentinel = table1.coalesce().updateView(sentinelColumnName + " = true");
+
+            // we must use leftOuterJoin again as there may be multiple matches per group
+            final Table rightTable = CrossJoinHelper.leftOuterJoin(
```
cpwright, line 84: So this is an interesting take on it; because we're going to end up doing the full cross join and then ditching part of the answer.

We are eliminating one hash table for the naturalJoin (or whereIn), which should be a win; the question is whether the necessary cross join state is worse is open and may depend on the data.

The columns are also out of order; but I guess merge handles that nicely?

engine/table/src/main/java/io/deephaven/engine/util/OuterJoinTools.java
```diff
@@ -0,0 +1,410 @@
...
+        // we must use leftOuterJoin again as there may be multiple matches per group
+        final Table rightOnlyTable = CrossJoinHelper.leftOuterJoin(
+                (QueryTable) table2.coalesce(),
+                (QueryTable) table1.coalesce().view(leftColumns),
+                matchColumns,
+                new MatchPair[] {new MatchPair(sentinelColumnName, sentinelColumnName)},
+                CrossJoinHelper.DEFAULT_NUM_RIGHT_BITS_TO_RESERVE).where(sentinelColumnName + " == null")
+                .view(rightColumns);
+
+        final Table nullLeftTable = TableTools.newTable(
```
cpwright, line 120: I am skeptical that it makes sense to have another join as part of line 123 instead of creating the columns correctly in the view on line 118.

engine/table/src/main/java/io/deephaven/engine/util/OuterJoinTools.java
nbauernfeind: The merge does handle the out of order columns by respecting the ordering in the first table.

I agree the cross-join output may be terrible (and in specific scenarios absolutely is). Is it worse than the extra memory usage? I don't know; it's hard to tell and probably is sometimes and probably isn't sometimes.

Best solution probably involves incrementally improving the cross join mode to be an enum of INNER_JOIN, LEFT_OUTER_JOIN, and FULL_OUTER_JOIN. This is likely the biggest performance win for effort.

I'll chat offline with you about what you/@rcaudy prefer as part of this PR.

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2060
Conceptual: https://github.com/deephaven/deephaven.io/issues/2061
Reference: https://github.com/deephaven/deephaven.io/issues/2063
Blog: https://github.com/deephaven/deephaven.io/issues/2062

PR #3319, opened 2023-01-18
Remove pyarrow<10 restriction.

devinrsmith:
This was blocking deephaven usage with Python 3.11 (pyarrow 9.0.0 does not provide Python 3.11 wheels).

The restriction in-and-of-itself was a bit premature, and only applicable in cases where the user wanted to explicitly install turbodbc.

### DIFF
--- a/py/server/setup.py
+++ b/py/server/setup.py
@@ -57,7 +57,7 @@ setup(
         'deephaven-plugin',
         'numpy',
         'pandas',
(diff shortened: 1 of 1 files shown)


jmao-denver (approved):
The restriction was added in anticipation that the ODBC integration project will depend on Turbodbc which requires pyarrow 9.0. But Turbodbc should be an optional dependency for us and they are working to make it buildable on latest pyarrow and latest MacOS etc.

PR #3333, opened 2023-01-19
Update proto-generated files for cpp-client

niloc132:
Branch is a draft since it appears to have compiler errors, but as far as I can tell, this is what you get from following the current manual steps to generate new cpp.


kosak (approved):
Approved (I pushed a commit to suppress the warning about deprecated declarations)

PR #3335, opened 2023-01-19
Remove the extra python base images

devinrsmith:
The extra python -base images are mostly fluff wrt this repository; previously with netty, the base images were used as part of the build and publish workflow. With jetty, the build and publish workflow is in https://github.com/deephaven/deephaven-server-docker, so keeping them dangling over here has little value.


niloc132 (approved):

rcaudy (commented):
Seems to me like there are inconsistent deletions w.r.t. tensorflow. Can you explain, @devinrsmith ?

devinrsmith:
@rcaudy forgot a few deletions, added them, repushed.

rcaudy (approved):

PR #3339, opened 2023-01-20
Support generating cpp proto files automatically, with compare checks

niloc132:
This is incomplete without the changed deephaven images, and will fail to build.


kosak (approved):

devinrsmith (dismissed):
The tasks work for me. I've verified that the compare will fail if the contents are improperly changed; and that updateProtobuf syncs the generated version correctly afaict.

I would suggest @rcaudy on M1 verify that `./gradlew cpp-client:compareProtobuf` works.

devinrsmith (approved):

PR #3346, opened 2023-01-21
Port DH-12148: Formula Array Access

nbauernfeind:
I had to pull together several PR's to completely port this feature. There was a lot more work than I anticipated lurking beneath the façade of DH-12148.

I felt that ShiftColumnOperation could be improved on from a performance point of view. I've included a test that arbitrarily takes ~30s on the original implementation, and ~8s on the replacement implementation. This is +70% improvement for that arbitrary, but hand picked, case using random updates. There are degenerate scenarios of the original implementation that the new implementation avoids. I dropped the replacement in a separate commit should one want to review the port vs the rewrite.

An interesting bonus of supporting formula array access is that we can drop most usages of `EvalNugget#hasUnstableColumns` as array offset access is now guaranteed to propagate downstream. This helped me identify a bug or two.

There were a couple of snippets/scenarios in the DHE port that just did not work in DHC. I reached out to @cpwright offline to highlight concerns.

[DH-12148](https://deephaven.atlassian.net/browse/DH-12148): Formula Array Access changes [commit](https://github.com/deephaven-ent/iris/commit/7e701a77e6ed26d852755c2886b19e6441bc85fd)
[DH-12030](https://deephaven.atlassian.net/browse/DH-12030) Implement ShiftColumnOperation [commit](https://github.com/deephaven-ent/iris/commit/f7826a675240d08dabf39eaa839aeb9d8a93d112)
[DH-12199](https://deephaven.atlassian.net/browse/DH-12199): ShiftedColumnOperation's ModifiedColumnSet should include all Source Column + Shifted Columns [commit](https://github.com/deephaven-ent/iris/commit/718a502862802610ecf9ed5db09cace635150295)
[DH-12273](https://deephaven.atlassian.net/browse/DH-12273): FormulaArrayAccess changes for where clause [commit](https://github.com/deephaven-ent/iris/commit/30dffe3f6bb3a6727a553c348c7ca8c1ec07da51)

Nightlies are running [here](https://github.com/nbauernfeind/deephaven-core/actions/runs/4348490143).

### DIFF
--- a/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
+++ b/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
@@ -276,7 +276,7 @@ public enum UpdateGraphProcessor implements UpdateSourceRegistrar, NotificationQ
         refreshThread.setDaemon(true);
 
         final int updateThreads =
-                Configuration.getInstance().getIntegerWithDefault("UpdateGraphProcessor.updateThreads", 1);
+                Configuration.getInstance().getIntegerWithDefault("UpdateGraphProcessor.updateThreads", -1);
         if (updateThreads <= 0) {
             this.updateThreads = Runtime.getRuntime().availableProcessors();
         } else {
--- a/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
+++ b/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
@@ -276,7 +276,7 @@ public enum UpdateGraphProcessor implements UpdateSourceRegistrar, NotificationQ
         refreshThread.setDaemon(true);
 
         final int updateThreads =
-                Configuration.getInstance().getIntegerWithDefault("UpdateGraphProcessor.updateThreads", -1);
+                Configuration.getInstance().getIntegerWithDefault("UpdateGraphProcessor.updateThreads", 1);
         if (updateThreads <= 0) {
             this.updateThreads = Runtime.getRuntime().availableProcessors();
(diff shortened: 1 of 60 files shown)


engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
```diff
@@ -276,7 +276,7 @@ public synchronized void take(final AccumulatedCycleStats out) {
         refreshThread.setDaemon(true);
 
         final int updateThreads =
-                Configuration.getInstance().getIntegerWithDefault("UpdateGraphProcessor.updateThreads", -1);
+                Configuration.getInstance().getIntegerWithDefault("UpdateGraphProcessor.updateThreads", 1);
```
lbooker42, line 279: This doesn't look right, it reverses the recent change to multi-thread by default.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/table/BarrageTable.java
```diff
@@ -100,11 +98,6 @@ public abstract class BarrageTable extends QueryTable implements BarrageMessage.
     protected boolean serverReverseViewport;
     protected BitSet serverColumns;
 
-    /** the size of the initial viewport requested from the server (-1 implies full subscription) */
```
lbooker42, line 103: I'm not sad to see this go, it was kludge.

lbooker42 (commented):
Very excited to see `APPEND_ONLY` attribute, I think this could be used to accelerate a large number of operations.

lbooker42 (dismissed):

lbooker42 (dismissed):

lbooker42 (approved):

PR #3349, opened 2023-01-23
Simplify js api build

niloc132:
Historically, the js api and ide projects were separate, under the assumption that distinct JS files might be exposed to different consumers. That never happened, and for DHC the web-client-ide jar is down to only three .java files. This PR moves those files to web-client-api, and then removes the build wiring made redundant with only one JS library.


devinrsmith (approved):
Gradle changes look good to me. Was able to execute web-client-api tasks locally, and checked that server came up successfully.

PR #3353, opened 2023-01-24
DH-14010: Ingest Community performance logs into Enterprise

arman-ddl:
Not ready for review; currently just for reference.


engine/table/src/main/java/io/deephaven/engine/table/impl/util/InternalTableLoggerWrapper.java
```diff
@@ -0,0 +1,18 @@
...
+import io.deephaven.engine.table.TableDefinition;
+import io.deephaven.engine.table.impl.QueryTable;
+import io.deephaven.engine.tablelogger.UpdatePerformanceLogLogger;
+import io.deephaven.io.logger.Logger;
+import io.deephaven.tablelogger.TableLogger;
+
+public interface InternalTableLoggerWrapper<T extends TableLogger> {
+    T getTableLogger();
+
+    QueryTable getQueryTable();
```
arman-ddl, line 12: Per Charles:
> you need to separate the interface I think
> the thing that gives you a table logger is very different than what gives you a query table
> like in DHC, I would not really expect those to bet he same thing at all

engine/table/src/main/java/io/deephaven/engine/table/impl/util/InternalTableLoggerWrapper.java
```diff
@@ -0,0 +1,18 @@
+package io.deephaven.engine.table.impl.util;
+
+import io.deephaven.engine.table.TableDefinition;
+import io.deephaven.engine.table.impl.QueryTable;
+import io.deephaven.engine.tablelogger.UpdatePerformanceLogLogger;
+import io.deephaven.io.logger.Logger;
+import io.deephaven.tablelogger.TableLogger;
+
+public interface InternalTableLoggerWrapper<T extends TableLogger> {
```
cpwright, line 9: I would get rid of this class.  Logging the table and getting the result are two completely separate things, which may not be particularly related.  You should have provider for the TableLogger and another provider for the QueryTable.  In Community, those might need to be related; but in Enterprise they would not be.

Additionally, the idea of just "get the query table" doesn't make sense in Enterprise, because you should not want to monitor the performance from the executing query.

devinrsmith:
Note: we don't merge / PR into release branches, so if this is something we want to merge in, let's make sure we sync up.

cpwright:
This would go into main, not the release branch; which will eventually be picked up for Vermilion.  The Enterprise feature's merge will wait on 0.23 being released (which is somewhat unfortunate, as it could rot between now and the end of the month - but is the most reasonable path).

arman-ddl:
Temporarily closing to avoid nightly runs.

PR #3360, opened 2023-01-26
JS API support for expand all

### LINKED ISSUE #3322: JS API support for expand all
As a client using tree tables in the the JS API, I want an expand all API so that I can expand all the nodes under a specific child in the tree table (or all the nodes in the entire tree).

emilyhuxng:
Closes #3322 

Added a boolean for expand all and the action added is dependant on that boolean.

### DIFF
--- a/web/WebDevelopersGuide.md
+++ b/web/WebDevelopersGuide.md
@@ -786,13 +786,13 @@ and count of children at each level of the hierarchy, and differences in the dat
  * `boolean isIncludeConstituents` - True if this is a roll-up and will provide the original rows that make up each grouping.
 
 ###### Methods
- * `expand(Number|TreeRow, boolean)` - Expands the given node, so that its children are visible when they are in the viewport. 
- The parameter can be the row index, or the row object itself. The second parameter is an boolean value, false by default, specifying if the row should be fully expanded. Equivalent to `setExpanded(row, true)` with an optional third boolean parameter. 
- * `collapse(Number|TreeRow)` - Collapses the given node, so that its children and descendents are not visible in the
+ * `expand(Number|TreeRow, boolean expandDescendants = false)` - Expands the given node, so that its children are visible when they are in the viewport. 
+ The parameter can be the row index, or the row object itself. The second parameter is an boolean value, false by default, specifying if the row and all descendants should be fully expanded. Equivalent to `setExpanded(row, true)` with an optional third boolean parameter. 
+ * `collapse(Number|TreeRow)` - Collapses the given node, so that its children and descendants are not visible in the
  size or the viewport. The parameter can be the row index, or the row object itself. Equivalent to 
  `setExpanded(row, false, false)`.
  * `setExpanded(Number|TreeRow, boolean, boolean)` - Specifies if the given node should be expanded or collapsed. If this node has
- children, and the value is changed, the size of the table will change. If node is to be expanded and the third parameter, expandAll, is true, then its children will also be expanded.
+ children, and the value is changed, the size of the table will change. If node is to be expanded and the third parameter, expandDescendants, is true, then its children will also be expanded.
  * `isExpanded(Number|TreeRow):boolean` - Returns true if the given row is expanded, false otherwise. Equivalent to
  `TreeRow.isExpanded`, if an instance of the row is available.
  * `setViewport(Number firstRow, Number lastRow, Column[]= columns)`
--- a/web/WebDevelopersGuide.md
+++ b/web/WebDevelopersGuide.md
@@ -786,13 +786,13 @@ and count of children at each level of the hierarchy, and differences in the dat
  * `boolean isIncludeConstituents` - True if this is a roll-up and will provide the original rows that make up each grouping.
 
 ###### Methods
- * `expand(Number|TreeRow)` - Expands the given node, so that its children are visible when they are in the viewport. 
- The parameter can be the row index, or the row object itself. Equivalent to `setExpanded(row, true)`.
+ * `expand(Number|TreeRow, boolean)` - Expands the given node, so that its children are visible when they are in the viewport. 
+ The parameter can be the row index, or the row object itself. The second parameter is an boolean value, false by default, specifying if the row should be fully expanded. Equivalent to `setExpanded(row, true)` with an optional third boolean parameter. 
  * `collapse(Number|TreeRow)` - Collapses the given node, so that its children and descendents are not visible in the
  size or the viewport. The parameter can be the row index, or the row object itself. Equivalent to 
- `setExpanded(row, false)`.
- * `setExpanded(Number|TreeRow, boolean)` - Specifies if the given node should be expanded or collapsed. If this node has
- children, and the value is changed, the size of the table will change.
+ `setExpanded(row, false, false)`.
+ * `setExpanded(Number|TreeRow, boolean, boolean)` - Specifies if the given node should be expanded or collapsed. If this node has
+ children, and the value is changed, the size of the table will change. If node is to be expanded and the third parameter, expandAll, is true, then its children will also be expanded.
  * `isExpanded(Number|TreeRow):boolean` - Returns true if the given row is expanded, false otherwise. Equivalent to
  `TreeRow.isExpanded`, if an instance of the row is available.
  * `setViewport(Number firstRow, Number lastRow, Column[]= columns)`
--- a/web/WebDevelopersGuide.md
+++ b/web/WebDevelopersGuide.md
@@ -787,12 +787,12 @@ and count of children at each level of the hierarchy, and differences in the dat
 
 ###### Methods
  * `expand(Number|TreeRow, boolean expandDescendants = false)` - Expands the given node, so that its children are visible when they are in the viewport. 
- The parameter can be the row index, or the row object itself. The second parameter is an boolean value, false by default, specifying if the row and all descendants should be fully expanded. Equivalent to `setExpanded(row, true)` with an optional third boolean parameter. 
+ The parameter can be the row index, or the row object itself. The second parameter is a boolean value, false by default, specifying if the row and all descendants should be fully expanded. Equivalent to `setExpanded(row, true)` with an optional third boolean parameter. 
  * `collapse(Number|TreeRow)` - Collapses the given node, so that its children and descendants are not visible in the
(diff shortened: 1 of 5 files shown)


web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -248,13 +248,14 @@ public int depth() {
                 return depthColumn[offsetInSnapshot];
             }
 
-            public void appendKeyData(Object[][] keyTableData, boolean expanded) {
+            public void appendKeyData(Object[][] keyTableData, boolean expanded, boolean expandAll) {
```
mofojed, line 251: I'd probably change this to just take the action directly instead of expanded/expandAll.

mofojed (commented):
We'll need to update documentation to the WebDevelopersGuide.md as well.
Should get @niloc132 's review to make sure this API is okay or if we should do something different (such as a separate expandAll method instead of just a new param).

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -726,19 +727,19 @@ private BitSet makeColumnSubscriptionBitset() {
...
-        setExpanded(row, false);
+        setExpanded(row, false, false);
     }
 
     @JsMethod
-    public void setExpanded(Object row, boolean isExpanded) {
+    public void setExpanded(Object row, boolean isExpanded, @JsOptional Boolean expandAll) {
         // TODO check row number is within bounds
-
+        boolean expandAllValue = expandAll != null && expandAll == true;
```
niloc132, line 742: Can simplify null check and true check to just equals true - however in regular java at least you can't directly compare a null to a boolean primitive value (java will unbox it), instead you have to compare with the constant Boolean.TRUE.

This seems wasteful, but in GWT, since all your code is actually JS, all boxed booleans are just plain js values, that is in JS, `true === Boolean.TRUE`, there is no boxing and unboxing for Booleans (or Doubles).

So, this can be written simply as 
```suggestion
        boolean expandAllValue = expandAll == Boolean.TRUE;
```

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
niloc132: Agreed, both these ternaries could be shifted instead to the only callsite.

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
niloc132: Then you could easily put this comparison into the actual "which expand byte should I send" check, rather than having two sets of checks, something like

```
final byte action;
if (!isExpanded) {
  // if you want to be picky, you could assert that expandAll is null or false here, to make sure no one is trying a "collapse with descendents"
  action = ACTION_COLLAPSE;
} else if (expandAll == Boolean.TRUE) {
  action = ACTION_EXPAND_WITH_DESCENDENTS;
} else {
  action = ACTION_EXPAND;
}
```

niloc132 (commented):
We're going to need doc updates too, but with the typescript codegen on the horizon it might make more sense to just write better docs here instead?

web/WebDevelopersGuide.md
```diff
@@ -786,13 +786,13 @@ and count of children at each level of the hierarchy, and differences in the dat
...
+ * `expand(Number|TreeRow, boolean)` - Expands the given node, so that its children are visible when they are in the viewport. 
+ The parameter can be the row index, or the row object itself. The second parameter is an boolean value, false by default, specifying if the row should be fully expanded. Equivalent to `setExpanded(row, true)` with an optional third boolean parameter. 
  * `collapse(Number|TreeRow)` - Collapses the given node, so that its children and descendents are not visible in the
  size or the viewport. The parameter can be the row index, or the row object itself. Equivalent to 
- `setExpanded(row, false)`.
- * `setExpanded(Number|TreeRow, boolean)` - Specifies if the given node should be expanded or collapsed. If this node has
- children, and the value is changed, the size of the table will change.
+ `setExpanded(row, false, false)`.
+ * `setExpanded(Number|TreeRow, boolean, boolean)` - Specifies if the given node should be expanded or collapsed. If this node has
+ children, and the value is changed, the size of the table will change. If node is to be expanded and the third parameter, expandAll, is true, then its children will also be expanded.
```
mofojed, line 795:
```suggestion
 children, and the value is changed, the size of the table will change. If node is to be expanded and the third parameter, expandDescendants, is true, then its children will also be expanded.
```

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -748,7 +756,40 @@ public void setExpanded(Object row, boolean isExpanded) {
...
-        r.appendKeyData(keyTableData, isExpanded);
+        r.appendKeyData(keyTableData, action);
+        if (keyTable != null) {
+            keyTable.then(t -> {
+                t.close();
+                return null;
+            });
+            keyTable = null;
+        }
+        replaceSubscription(RebuildStep.HIERARCHICAL_TABLE_VIEW);
```
mofojed: This is used in `setExpanded`, `expandAll`, and `collapseAll`, you could refactor it into it's own method, e.g.
```
private void replaceKeyTable() {
    if (keyTable != null) {
        keyTable.then(t -> {
            t.close();
            return null;
        });
        keyTable = null;
    }
    replaceSubscription(RebuildStep.HIERARCHICAL_TABLE_VIEW);
} 
```

And then here just call that method:
```suggestion
        replaceKeyTable();
```

web/client-api/src/main/java/io/deephaven/web/public/treegrid.js
```diff
@@ -2,14 +2,25 @@
  * Simple es6 class encapsulating some basic rendering for a tree table in a grid, allowing
  * expanding and collapsing nodes, and filtering and sorting columns.
  */
+let console = window.console;
+
+console.log("hello");
```
mofojed, line 7: Remove unnecessary console statements.
```suggestion
```

web/client-api/src/main/java/io/deephaven/web/public/treegrid.js
```diff
@@ -34,7 +45,7 @@ class TreeGrid {
       if (cellElt.classList.contains('expanded')) {
         this.treeTable.collapse(rowNumber);
       } else {
-        this.treeTable.expand(rowNumber);
+        this.treeTable.expand(rowNumber, false);
```
mofojed, line 48: Param is optional, don't need to provide `false` here.
```suggestion
        this.treeTable.expand(rowNumber);
```

mofojed (changes requested):

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -725,19 +725,42 @@ private BitSet makeColumnSubscriptionBitset() {
...
+        Js.<JsArray<Double>>cast(keyTableData[i++]).setAt(0, (double) 0);
+        Js.<JsArray<Double>>cast(keyTableData[i++]).setAt(0, action);
+        if (keyTable != null) {
+            keyTable.then(t -> {
+                t.close();
+                return null;
+            });
+            keyTable = null;
+        }
+        replaceSubscription(RebuildStep.HIERARCHICAL_TABLE_VIEW);
```
mofojed, line 740: This is common in this method and end of `setExpanded`, can refactor (as mentioned in previous comments)

web/WebDevelopersGuide.md
```diff
@@ -786,13 +786,13 @@ and count of children at each level of the hierarchy, and differences in the dat
  * `boolean isIncludeConstituents` - True if this is a roll-up and will provide the original rows that make up each grouping.
 
 ###### Methods
- * `expand(Number|TreeRow)` - Expands the given node, so that its children are visible when they are in the viewport. 
- The parameter can be the row index, or the row object itself. Equivalent to `setExpanded(row, true)`.
- * `collapse(Number|TreeRow)` - Collapses the given node, so that its children and descendents are not visible in the
+ * `expand(Number|TreeRow, boolean expandDescendants = false)` - Expands the given node, so that its children are visible when they are in the viewport. 
+ The parameter can be the row index, or the row object itself. The second parameter is an boolean value, false by default, specifying if the row and all descendants should be fully expanded. Equivalent to `setExpanded(row, true)` with an optional third boolean parameter.
```
mofojed, line 790:
```suggestion
 The parameter can be the row index, or the row object itself. The second parameter is a boolean value, false by default, specifying if the row and all descendants should be fully expanded. Equivalent to `setExpanded(row, true)` with an optional third boolean parameter. 
```

web/WebDevelopersGuide.md
```diff
@@ -786,13 +786,13 @@ and count of children at each level of the hierarchy, and differences in the dat
...
+ The parameter can be the row index, or the row object itself. The second parameter is an boolean value, false by default, specifying if the row and all descendants should be fully expanded. Equivalent to `setExpanded(row, true)` with an optional third boolean parameter. 
+ * `collapse(Number|TreeRow)` - Collapses the given node, so that its children and descendants are not visible in the
  size or the viewport. The parameter can be the row index, or the row object itself. Equivalent to 
- `setExpanded(row, false)`.
- * `setExpanded(Number|TreeRow, boolean)` - Specifies if the given node should be expanded or collapsed. If this node has
- children, and the value is changed, the size of the table will change.
+ `setExpanded(row, false, false)`.
+ * `setExpanded(Number|TreeRow, boolean, boolean)` - Specifies if the given node should be expanded or collapsed. If this node has
+ children, and the value is changed, the size of the table will change. If node is to be expanded and the third parameter, expandDescendants, is true, then its children will also be expanded.
```
mofojed, line 795:
```suggestion
 children, and the value is changed, the size of the table will change. If node is to be expanded and the third parameter, `expandDescendants`, is true, then its children will also be expanded.
```

mofojed (changes requested):
Minor cleanup suggestions to the docs. Functionality-wise looks good.

mofojed (approved):

PR #3371, opened 2023-01-27
JS API authentication via Flight v2

niloc132:
JS API changes necessary for #2564, moving from the deprecated SessionService.newSession call to Flight v2 authentication. Since the gRPC-web client in use doesn't support interceptors or even reading headers from requests, we only use FlightService.handshake to keep the session alive and rotate tokens, with a hand-written implementation rather than the generated version.

Documentation will follow when the rest of the ticket is closed.

Partial #2564


server/src/main/java/io/deephaven/server/arrow/FlightServiceGrpcImpl.java
```diff
@@ -121,6 +121,7 @@ private Optional<AuthContext> login(String type, long version, ByteString payloa
                 AuthenticationRequestHandler.HandshakeResponseListener listener) throws AuthenticationException {
             AuthenticationRequestHandler handler = authRequestHandlers.get(type);
             if (handler == null) {
+                log.info().append("No AuthenticationRequestHandler registered for type ").append(type).endl();
```
niloc132, line 124: Need to revert these, they end up triggering when `Bearer` is specified

server/src/main/java/io/deephaven/server/session/SessionService.java
```diff
@@ -240,6 +243,7 @@ public SessionState getSessionForAuthToken(final String token) throws Authentica
         final String payload = offset < 0 ? "" : token.substring(offset + 1);
         AuthenticationRequestHandler handler = authRequestHandlers.get(key);
         if (handler == null) {
+            log.info().append("No AuthenticationRequestHandler registered for type ").append(key).endl();
```
niloc132, line 246: revert this as well

niloc132: Actually this one is okay, since we throw otherwise

niloc132: No, changed my mind again, going to revert this until we do more testing - it turns out that we permit Bearer to be an auth type outside of what DH uses for refresh, but only if we can't handle the token _as a_ refresh token, so clients that were connected before restart can trigger this.

web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
```diff
@@ -6,34 +6,45 @@
...
-            LOGIN_TYPE_SAML = "saml",
-            LOGIN_TYPE_REFRESH = "refresh",
-            LOGIN_TYPE_PSK = "psk",
-            LOGIN_TYPE_OIDC = "oidc",
             LOGIN_TYPE_ANONYMOUS = "anonymous";
 
-    private final String serverUrl;
+    private final IdeConnection ideConnection;
 
+    @JsConstructor
```
niloc132, line 39: unnecessary

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -414,37 +404,63 @@ private void startExportNotificationsStream() {
...
+                // Metadata
+                JsArray<String> authorization = Js.<BrowserHeaders>uncheckedCast(headers).get("authorization");
+                if (authorization.length > 0) {
+                    JsArray<String> existing = metadata().get("authorization");
+                    if (!existing.getAt(0).equals(authorization.getAt(0))) {
+                        // use this new token
+                        metadata().set("authorization", authorization);
+                        CustomEventInit init = CustomEventInit.create();
+                        init.setDetail(new JsRefreshToken(authorization.getAt(0), sessionTimeoutMs));
+                        info.fireEvent(EVENT_REFRESH_TOKEN_UPDATED, init);
```
niloc132, line 434: i have not tested this part yet

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -414,37 +404,63 @@ private void startExportNotificationsStream() {
...
+                if (status.isOk()) {
+                    // schedule an update based on our currently configured delay
+                    scheduledAuthUpdate = DomGlobal.setTimeout(ignore -> {
+                        authUpdate();
+                    }, sessionTimeoutMs / 2);
+
+                    resolve.onInvoke((Void) null);
+                } else {
+                    // token is no longer valid, signal deauth for re-login
+                    // TODO
```
niloc132, line 449: will link the current issue for follow-up here

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -414,37 +404,63 @@ private void startExportNotificationsStream() {
...
-            req.setPayload(handshakeResponse.getSessionToken_asU8());
-            sessionServiceClient.refreshSessionToken(req, metadata, (fail, success) -> {
-                if (fail != null) {
-                    // TODO set a flag so others know not to try until we re-trigger initial auth
-                    // TODO re-trigger auth; but for now let's try again using our last successful auth
-                    checkStatus((ResponseStreamWrapper.Status) fail);
-                    authUpdate(handshakeResponse);
-                    return;
+    /**
+     * Manages auth token update and rotation. A typical grpc/grpc-web client would support something like and
```
nbauernfeind, line 408: s/like and/like an

vbabich (dismissed):

web/client-api/src/main/java/io/deephaven/web/client/api/barrage/stream/HandshakeStreamFactory.java
```diff
@@ -0,0 +1,138 @@
...
+                new HandshakeRequest());
+    }
+
+    private static Map<String, JsArray<Function>> listenerMap() {
+        Map<String, JsArray<Function>> listeners = new HashMap<>();
+        listeners.put("data", new JsArray<>());
+        listeners.put("end", new JsArray<>());
+        listeners.put("status", new JsArray<>());
+        listeners.put("headers", new JsArray<>());
+        return listeners;
```
mofojed, line 136: These constants for the type is used a few times in the class, should probably be extracted to a constant

web/client-api/src/main/java/io/deephaven/web/client/ide/IdeClient.java
```diff
@@ -19,13 +19,10 @@ public Promise<IdeSession> getExistingSession_old(String websocketUrl, String au
         return IdeClient.getExistingSession(websocketUrl, authToken, serviceId, language);
     }
 
+    @Deprecated
```
mofojed, line 22: Should mark as deprecated in WebDevelopersGuide as well.

web/client-api/src/main/java/io/deephaven/web/client/ide/IdeConnection.java
```diff
@@ -7,52 +7,55 @@
...
-    }
-
+    private final JsRunnable deathListenerCleanup;
     private final String serverUrl;
 
-    private final JsRunnable deathListenerCleanup;
+    private final ConnectToken token = new ConnectToken();
+
+    @JsConstructor
+    public IdeConnection(String serverUrl, Boolean fromJava) {
```
mofojed, line 29: I think need to mark as `fromJava` as `@JsOptional` or else GWT will complain before we even get to the warning below.

web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
```diff
@@ -47,43 +57,74 @@ private <R> Promise<String[][]> getConfigs(Consumer<JsBiConsumer<Object, R>> rpc
...
     }
 
     public Promise<Void> login(LoginCredentials credentials) {
-        return connection.get().whenServerReady("login").then(ignore -> Promise.resolve((Void) null));
+        Objects.requireNonNull(credentials.getType(), "type must be specified");
+        ConnectToken token = ideConnection.getToken();
+        if (LOGIN_TYPE_PASSWORD.equals(credentials.getType())) {
+            Objects.requireNonNull(credentials.getUsername(), "username must be specified for password login");
+            Objects.requireNonNull(credentials.getToken(), "token must be specified for password login");
+            token.setType("Basic");
```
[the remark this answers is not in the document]
niloc132: I didn't plan to, since I wasn't intending to expose this directly to JS API consumers.

web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
```diff
@@ -47,43 +57,74 @@ private <R> Promise<String[][]> getConfigs(Consumer<JsBiConsumer<Object, R>> rpc
...
+            token.setValue(credentials.getToken());
+            if (credentials.getUsername() != null) {
+                JsLog.warn("username ignored for login type " + credentials.getType());
+            }
+        }
+        Promise<Void> login =
+                ideConnection.connection.get().whenServerReady("login").then(ignore -> Promise.resolve((Void) null));
+
+        // fetch configs and check session timeout
+        login.then(ignore -> getServerConfigValues()).then(configs -> {
```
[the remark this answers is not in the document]
niloc132: Sure - it has a safe fallback value, so we're just guarding against "the server is crazy".

web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
```diff
@@ -93,6 +134,14 @@ public Promise<UserInfo> getUserInfo() {
     }
 
     public JsStorageService getStorageService() {
-        return new JsStorageService(connection.get());
+        return new JsStorageService(ideConnection.connection.get());
+    }
+
+    public Promise<IdeConnection> getAsIdeConnection() {
+        return Promise.resolve(ideConnection);
```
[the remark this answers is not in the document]
niloc132: Because I'm trying to allow for that not always being the way that this will work - easier to just always be async and adding a slight delay for a one-time call, than try to add it later?

My guess is that we will eventually default to assuming that the server isn't a worker (but only config/storage/both), and we will need to confirm that it can be used as a worker.

web/client-api/src/main/java/io/deephaven/web/client/api/barrage/stream/HandshakeStreamFactory.java
niloc132: And then rewritten to never do this crap again. But yes.

web/client-api/src/main/java/io/deephaven/web/client/ide/IdeClient.java
niloc132: I'm holding off in favor of moving this all to type script in the next week or two, then deleting the .md file?

web/client-api/src/main/java/io/deephaven/web/client/ide/IdeConnection.java
niloc132: No, there is no runtime typecheck, but I'll add it, its better to have it. I'm using this to cheat my way around js constructor semantics, and will remove this in 2ish releases

web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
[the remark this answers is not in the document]
mofojed: Okay, was thinking for the Java code but whatever.

web/client-api/src/main/java/io/deephaven/web/client/ide/IdeClient.java
mofojed: Okay

mofojed (approved):

Issue #3367, opened 2023-01-26
`Instant`/`LocalDate`/`LocalTime`/`ZonedDateTime` support for Parquet, KafkaTools and JsonNodeUtil


nbauernfeind:
When porting over `Instant`, `LocalDate`, `LocalTime`, and `ZonedDateTime` support from DHE, I skipped initial integration with Parquet/Kafka/Json due to low familiarity with their implementations (additionally, these integrations were not included in the initial work).

This ticket is to track adding support for these types to those ingestion/exhaustion tools.

Issue #3359, opened 2023-01-26
`ArraySource#move` support moving of full blocks


nbauernfeind:
When moving data from a source to a dest that align with block multiples we can re-use and move full blocks. See `CharacterArraySource#move` for reference.

PR #3385, opened 2023-01-31
Port DH-11692: Add native support for java.time types

nbauernfeind:
I did a bit more replication than was in the DHE implementation; but hopefully for the better.

DH-11692 via commits:
- https://github.com/deephaven-ent/iris/commit/563e69e16a2206b0c90ec220b0bd75f4a68e3bac
- https://github.com/deephaven-ent/iris/commit/f680b26788f1d4ab998e374278eff81449c85b7f
- https://github.com/deephaven-ent/iris/commit/4f46299190e396093860e2a9becd1382f21ffd39
- https://github.com/deephaven-ent/iris/commit/7baba9f419c6ab9f2fb0b16c6ad77da414b28237

Follow up work:
- #3367
- #3359

Nightlies: https://github.com/nbauernfeind/deephaven-core/actions/runs/4058490234


engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -1240,128 +1234,182 @@ public static Clock currentClock() {
...
-     * Provides a String representing the previous business date in the New York time zone using the NYSE calendar, or,
-     * if a custom {@link #lastBusinessDayNyOverride} has been set, the date provided by that override.
+     * Converts a {@link DateTime} to a {@link ZonedDateTime}.
      *
-     * @return A String in yyyy-MM-dd format.
+     * @param dateTime The a {@link DateTime} to convert.
+     * @param timeZone The {@link TimeZone} to use for the conversion.
+     * @return A {@link ZonedDateTime} using the specified time zone. or null if dateTime was null
      */
-    public static String lastBusinessDateNy() {
```
cpwright, line 1318: Removing lastBusinessDateNy() is probably defensible if you do not want to own calendars at all (but then Enterprise needs to reimplement something).  You'll also lose a lot of plot utility w/o business calendars.

I think not having currentDateNy() which does not rely on calendars is pretty bad.  @rbasralian do you think this is a tenable position with respect to usability?

None of the override stuff I particular see us needing.

engine/table/src/main/java/io/deephaven/engine/table/impl/util/TableTimeConversions.java
```diff
@@ -0,0 +1,341 @@
...
+ * {@link Instant}, {@link ZonedDateTime}, {@link LocalDate}, {@link LocalTime}, and {@code long}.
+ *
+ * <p>
+ * For example, lets say that you wanted to select multiple days from a table, but filter them to a specific times of
+ * the day.
+ * </p>
+ *
+ * <pre>
+ * baseTable = db.i("Market", "Trades")
+ *               .where("Date &#62; 2021-10-01")
```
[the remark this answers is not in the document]
nbauernfeind: Hmm. In the pre-block this seems to render just fine using `<`.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
nbauernfeind: As discussed offline, for the foreseeable future, DHE will own supporting clients in this regard.

cpwright (approved):

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/RegionedColumnSourceInstant.java
```diff
@@ -0,0 +1,96 @@
...
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalTime;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+
+/**
+ * Regioned column source implementation for columns of {@link Instant}s.
+ */
+final class RegionedColumnSourceInstant
```
rcaudy, line 25: Can you make this class and its peers a little more mutually consistent? (`elementIndex`->`rowKey`, `final @NotNull` -> `@NotNull final`, spacing, etc)

Util/src/main/java/io/deephaven/util/type/TypeUtils.java
```diff
@@ -512,14 +513,15 @@ public static boolean isCharacter(@NotNull final Class<?> c) {
...
+     * Whether the class is a Date, DateTime, or Instant.
      *
      * @param type The class.
-     * @return true if the type is a DateTime or {@link Date}.
+     * @return true if the type is a {@link Date}, DateTime or {@link Instant}.
      */
     public static boolean isDateTime(Class<?> type) {
-        return Date.class.isAssignableFrom(type) || type.getAnnotation(IsDateTime.class) != null
-                && type.getAnnotation(IsDateTime.class).value();
+        return Date.class.isAssignableFrom(type)
```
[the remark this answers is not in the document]
nbauernfeind: This method is used to create time filters and to determine if the column is displayable to the user. (If it is non-displayable, then it gets `toString`d.) It is only used by server-side UI helpers.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/LongChunkedAddOnlyMinMaxOperator.java
```diff
@@ -42,7 +49,15 @@ class LongChunkedAddOnlyMinMaxOperator implements IterativeChunkedAggregationOpe
         this.minimum = minimum;
         this.name = name;
         // region resultColumn initialization
-        resultColumn = type == DateTime.class ? new DateTimeArraySource() : new LongArraySource();
+        if (type == DateTime.class) {
```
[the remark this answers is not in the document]
nbauernfeind: It probably would be nice. As noted offline we're probably in trouble anywhere we lose the ZoneId (say via `ComponentType` usages). I've opened an issue for creating a `ColumnType` that we can pass around instead of `Class<?>`. This is probably a lot better than introducing an optional `ZoneId` to pass through multiple layers.

See #3455.

Util/src/main/java/io/deephaven/util/codec/ZonedDateTimeCodec.java
```diff
@@ -0,0 +1,96 @@
...
+import io.deephaven.util.QueryConstants;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.nio.ByteBuffer;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+
+public class ZonedDateTimeCodec implements ObjectCodec<ZonedDateTime> {
```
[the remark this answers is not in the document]
nbauernfeind: Hmm, it's hard for me to tell. I'm not sure how to utilize it, but it does seem like there could exist parquet metadata that would utilize this codec if referenced by class name.

Util/src/main/java/io/deephaven/util/codec/ZonedDateTimeCodec.java
```diff
@@ -0,0 +1,96 @@
...
+    // Sadly, this is copied from DBTimeUtils since that lives in the DB package and this cannot.
+    private static long toEpochNano(@Nullable final ZonedDateTime value) {
+        if (value == null) {
+            return QueryConstants.NULL_LONG;
+        }
+
+        return safeComputeNanos(value.toEpochSecond(), value.getNano());
+    }
+
+    private static long safeComputeNanos(long epochSecond, long nanoOfSecond) {
```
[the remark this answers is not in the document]
nbauernfeind: Note that this exception is preventing `epochSecond` from overflowing as well. I've added `Math.addExact` for the summation, though.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/AggregationProcessor.java
```diff
@@ -507,6 +510,9 @@ final void addMinOrMaxOperator(final boolean isMin, @NotNull final String inputN
                     return;
                 }
             }
+            if (rawInputSource instanceof ConvertableTimeSource.Zoned) {
+                ZoneId id = ((ConvertableTimeSource.Zoned) rawInputSource).getZone();
+            }
```
rcaudy, line 515: Seems like a change that didn't pan out?

engine/table/src/test/java/io/deephaven/engine/table/impl/sources/AbstractBooleanColumnSourceTest.java
```diff
@@ -21,8 +21,11 @@
 import io.deephaven.engine.rowset.RowSetBuilderSequential;
 import io.deephaven.engine.rowset.RowSetFactory;
 import io.deephaven.engine.rowset.chunkattributes.RowKeys;
+import io.deephaven.engine.table.ChunkSink;
```
rcaudy, line 24: All of these new imports are unused. That's OK, if these are a replication artifact.

Util/src/main/java/io/deephaven/util/type/TypeUtils.java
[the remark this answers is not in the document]
rcaudy: Why not ZonedDateTime, or LocalXYZ?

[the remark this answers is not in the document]
rcaudy: We discussed that `Date` probably does not belong here, other code that thinks `Serializable` works should be changed, and you need to debug `Instant` and or `ZonedDateTime`.

nbauernfeind:
I can't seem to reply to your final comment. I added ZonedDateTime/Instant to the list. I removed Date. For now LocalDate and LocalTime are stubbed with FixedSizeBinary formats, but these are not implemented between server/client neither for worker-to-worker nor for the jsapi. 

At this moment, barrage strips the ZoneId from ZonedDateTime columns and sends as UTC. The ser/deser code acts on Class<?> and it would be much nicer to use ColumnDataType instead of passing extra args.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ReinterpretUtils.java
```diff
@@ -98,6 +99,68 @@ public static ColumnSource<Long> zonedDateTimeToLongSource(final @NotNull Column
...
+    public static WritableColumnSource<?> writableBooleanToByteSource(WritableColumnSource<?> source) {
+        if (source.allowsReinterpret(byte.class)) {
+            return (WritableColumnSource<Byte>) source.reinterpret(byte.class);
+        }
+        return null;
+    }
+
+    /**
+     * Given an {@link Instant} column source turn it into a long column source, either via reinterpretation or
+     * wrapping.
```
rcaudy, line 132: Remove `or wrapping` from this one and `writableInstantToLongSource`.
Maybe consider mentioning that the source is writable on all 5 new methods? Feel not strictly necessary, since the types are specified.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/ReinterpretUtils.java
```diff
@@ -98,6 +99,68 @@ public static ColumnSource<Long> zonedDateTimeToLongSource(final @NotNull Column
...
+     * @return the long source or null if it could not be reinterpretted
+     */
+    @NotNull
+    public static WritableColumnSource<Long> writableZonedDateTimeToLongSource(
+            final @NotNull WritableColumnSource<?> source) {
+        if (source.allowsReinterpret(long.class)) {
+            return (WritableColumnSource<Long>) source.reinterpret(long.class);
+        }
+        return null;
+    }
```
rcaudy, line 162: You have 3 identical method bodies. Converge them?

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/InMemoryColumnSource.java
```diff
@@ -18,6 +18,7 @@
 import org.jetbrains.annotations.Nullable;
 
 import java.time.Instant;
+import java.time.ZonedDateTime;
```
rcaudy, line 21: Unused

engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/SelectColumnLayer.java
```diff
@@ -110,7 +110,13 @@ final public class SelectColumnLayer extends SelectOrViewColumnLayer {
 
     private ChunkSource<Values> getChunkSource() {
         if (chunkSource == null) {
-            chunkSource = ReinterpretUtils.maybeConvertToPrimitive(selectColumn.getDataView());
+            ColumnSource<?> dataSource = selectColumn.getDataView();
+            if (dataSource.getType() != writableSource.getType()) {
+                // this should only occur when using primitives internally and the user has requested a non-primitive
+                chunkSource = ReinterpretUtils.maybeConvertToPrimitive(dataSource);
+            } else {
```
rcaudy, line 117: We should follow the reinterpret with an `Assert` that the resulting types are now equal.

rcaudy (approved):

PR #3395, opened 2023-02-01
Use Classgraph to Import Unknown Packages Not Yet Loaded

### LINKED ISSUE #1129: Importing groovy.time.* results in an error
The following code (in a Deephaven Groovy console):

`import groovy.time.*`

Results in the following error and stack trace:
```
r-Scheduler-Serial-1 | .u.GroovyDeephavenSession | Invalid import: "import groovy.time.*;"
r-Scheduler-Serial-1 | i.d.g.s.SessionState      | Internal Error '12e4617c-5ec4-41d3-a2c5-219105f4df9b' java.lang.RuntimeException: Attempting to import a path that does not exist: import groovy.time.*;
        at io.deephaven.db.util.GroovyDeephavenSession.updateScriptImports(GroovyDeephavenSession.java:487)
        at io.deephaven.db.util.GroovyDeephavenSession.grepScriptImports(GroovyDeephavenSession.java:498)
        at io.deephaven.db.util.GroovyDeephavenSession.evaluate(GroovyDeephavenSession.java:229)
        at io.deephaven.db.util.AbstractScriptSession.evaluateScript(AbstractScriptSession.java:115)
        at io.deephaven.db.util.ScriptSession.evaluateScript(ScriptSession.java:71)
        at io.deephaven.grpc_api.console.ConsoleServiceGrpcImpl.lambda$null$8(ConsoleServiceGrpcImpl.java:176)
        at io.deephaven.grpc_api.session.SessionState$ExportBuilder.lambda$submit$1(SessionState.java:1303)
        at io.deephaven.grpc_api.session.SessionState$ExportObject.doExport(SessionState.java:866)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at io.deephaven.grpc_api.runner.DeephavenApiServerModule$ThreadFactory.lambda$newThread$0(DeephavenApiServerModule.java:146)
        at java.lang.Thread.run(Thread.java:748)
```

**Steps to reproduce**
1. Launch deephaven-core with Groovy
2. `import groovy.time.*`
3. An error will occur

**Expected results**

Not much to happen.  The import should work, and you should be able to use the groovy time library.

**Actual results**

An error with the stack trace given above.

**Versions**
 - Deephaven: community
 - OS: Big Sur
 - Browser: Chrome
 - Docker: 20.10.7

nbauernfeind:
Fixes #1129.

### DIFF
--- a/engine/table/build.gradle
+++ b/engine/table/build.gradle
@@ -52,6 +52,7 @@ dependencies {
     // TODO(deephaven-core#3204): t-digest 3.3 appears to have higher errors than 3.2
     implementation 'com.tdunning:t-digest:3.2'
     implementation 'com.squareup:javapoet:1.13.0'
+    implementation 'io.github.classgraph:classgraph:4.8.103'
 
     implementation project(':plugin')
     implementation depCommonsLang3
--- a/engine/table/build.gradle
+++ b/engine/table/build.gradle
@@ -52,7 +52,7 @@ dependencies {
     // TODO(deephaven-core#3204): t-digest 3.3 appears to have higher errors than 3.2
     implementation 'com.tdunning:t-digest:3.2'
     implementation 'com.squareup:javapoet:1.13.0'
-    implementation 'io.github.classgraph:classgraph:4.8.103'
+    implementation 'io.github.classgraph:classgraph:4.8.154'
 
     implementation project(':plugin')
     implementation depCommonsLang3
--- a/engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
@@ -438,7 +438,7 @@ public class GroovyDeephavenSession extends AbstractScriptSession<GroovySnapshot
     }
 
     private static boolean visibleToClassGraph(String packageImport) {
-        try (ScanResult scanResult = new ClassGraph().enableAllInfo().acceptPackages(packageImport).scan()) {
+        try (ScanResult scanResult = new ClassGraph().enableClassInfo().acceptPackages(packageImport).scan()) {
             final Optional<ClassInfo> firstClassFound = scanResult.getAllClasses().stream().findFirst();
             // force load the class so that the jvm is aware of the package
             firstClassFound.ifPresent(ClassInfo::loadClass);
--- a/engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
@@ -273,10 +273,9 @@ public class GroovyDeephavenSession extends AbstractScriptSession<GroovySnapshot
         return e;
     }
 
-    private static String classForNameString(String className) throws ClassNotFoundException {
+    private static Class<?> loadClass(String className) throws ClassNotFoundException {
         try {
-            Class.forName(className);
-            return className;
+            return Class.forName(className, false, GroovyDeephavenSession.class.getClassLoader());
         } catch (ClassNotFoundException e) {
             if (className.contains(".")) {
                 // handle inner class cases
@@ -285,7 +284,7 @@ public class GroovyDeephavenSession extends AbstractScriptSession<GroovySnapshot
                 String tail = className.substring(index + 1);
                 String newClassName = head + "$" + tail;
 
-                return classForNameString(newClassName);
+                return loadClass(newClassName);
             } else {
                 throw e;
             }
@@ -294,7 +293,7 @@ public class GroovyDeephavenSession extends AbstractScriptSession<GroovySnapshot
 
     private static boolean classExists(String className) {
         try {
-            classForNameString(className);
+            loadClass(className);
             return true;
         } catch (ClassNotFoundException e) {
             return false;
@@ -303,7 +302,7 @@ public class GroovyDeephavenSession extends AbstractScriptSession<GroovySnapshot
 
     private static boolean functionExists(String className, String functionName) {
         try {
-            Method[] ms = Class.forName(classForNameString(className)).getMethods();
+            Method[] ms = loadClass(className).getMethods();
 
             for (Method m : ms) {
                 if (m.getName().equals(functionName)) {
@@ -319,7 +318,7 @@ public class GroovyDeephavenSession extends AbstractScriptSession<GroovySnapshot
 
     private static boolean fieldExists(String className, String fieldName) {
         try {
-            Field[] fs = Class.forName(classForNameString(className)).getFields();
+            Field[] fs = loadClass(className).getFields();
 
             for (Field f : fs) {
                 if (f.getName().equals(fieldName)) {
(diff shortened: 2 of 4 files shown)


engine/table/build.gradle
```diff
@@ -52,6 +52,7 @@ dependencies {
     // TODO(deephaven-core#3204): t-digest 3.3 appears to have higher errors than 3.2
     implementation 'com.tdunning:t-digest:3.2'
     implementation 'com.squareup:javapoet:1.13.0'
+    implementation 'io.github.classgraph:classgraph:4.8.103'
```
devinrsmith, line 55: I like classgraph. We need to decide if this additional dependency is worth it. It doesn't have any external dependencies itself, which is nice.

I'd bump to the latest version.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -435,6 +437,15 @@ public static String isValidImportString(Logger log, String importString) {
         }
     }
 
+    private static boolean visibleToClassGraph(String packageImport) {
+        try (ScanResult scanResult = new ClassGraph().enableAllInfo().acceptPackages(packageImport).scan()) {
```
devinrsmith, line 441: Is it possible that a user could run `import com.example.*.SomeClass`? This is something that `acceptPackages` allows, but `import <xyz>` doesn't (I don't think...).

nbauernfeind: Doesn't look like it:
```
java.lang.RuntimeException: Attempting to import a path that does not exist: import groovy.*.TimeCategory;
```

Where as both of these do work:
```groovy
import groovy.time.*
import groovy.time.TimeCategory
```

engine/table/build.gradle
nbauernfeind: I've bumped. I'd be interested in any other approach you have in mind. The reported bug is an annoying thing for the user to need to handle.

Maybe we should make groovy-support an extension?

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
devinrsmith, line 441: I think we can limit the scan features to `ClassGraph().enableClassInfo()`; I don't think we necessarily want, or need, the other features (field info, method info, annotation info, private classes, etc).

devinrsmith (commented):
LGTM, one more point.

devinrsmith (dismissed):
Let's check w/ @rcaudy too

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -435,6 +437,15 @@ public static String isValidImportString(Logger log, String importString) {
         }
     }
 
+    private static boolean visibleToClassGraph(String packageImport) {
+        try (ScanResult scanResult = new ClassGraph().enableClassInfo().acceptPackages(packageImport).scan()) {
+            final Optional<ClassInfo> firstClassFound = scanResult.getAllClasses().stream().findFirst();
+            // force load the class so that the jvm is aware of the package
+            firstClassFound.ifPresent(ClassInfo::loadClass);
```
rcaudy, line 444: Is this going to invoke static initializers for the class? Could this have undesirable side-effects?
(I think the answer is "no", since you have not called `io.github.classgraph.ClassGraph#initializeLoadedClasses` on the `ClassGraph` instance, but it might be worth testing/verifying. Also, `ClassGraph` falls back to `Class.forName()`, which defaults to `true` for initializing classes, which is kind of broken if you ask me, and maybe we should offer them (`ClassGraph`) a PR to fix it.)
That said, our `classExists` is not taking any such precautions. Maybe we need to make these things more consistent

devinrsmith: If you dig in deeper, it actually doesn't fall back to forName (I think) - by default, it calls w/ `initializeLoadedClasses=false` and the explicit forName call.

devinrsmith:
```java
    Class<?> loadClass(final boolean ignoreExceptions) {
        if (classRef == null) {
            final String className = getClassInfoNameOrClassName();
            if (scanResult != null) {
                classRef = scanResult.loadClass(className, ignoreExceptions);
            } else {
                // Fallback, if scanResult is not set
                try {
                    classRef = Class.forName(className);
                } catch (final Throwable t) {
                    if (!ignoreExceptions) {
                        throw new IllegalArgumentException("Could not load class " + className, t);
                    }
                }
            }
        }
        return classRef;
    }
```

I think `scanResult` will always be non-null.

```java
    public Class<?> loadClass(final String className, final boolean returnNullIfClassNotFound)
            throws IllegalArgumentException {
        if (closed.get()) {
            throw new IllegalArgumentException("Cannot use a ScanResult after it has been closed");
        }
        if (className == null || className.isEmpty()) {
            throw new NullPointerException("className cannot be null or empty");
        }
        try {
            return Class.forName(className, scanSpec.initializeLoadedClasses, classGraphClassLoader);
        } catch (final ClassNotFoundException | LinkageError e) {
            if (returnNullIfClassNotFound) {
                return null;
            } else {
                throw new IllegalArgumentException("Could not load class " + className + " : " + e, e);
            }
        }
    }
```

nbauernfeind:
I can verify that class graph does not initialize any classes until they are actually loaded:
```java
package io.deephaven.server;

public class TestStatic {
    static {
        System.out.println("Initializing TestStatic");
    }
}
```

I ran three commands one at a time, in this order:
```groovy
import io.deephaven.server.barrage.*
```
```groovy
import io.deephaven.server.*
```
```groovy
import io.deephaven.server.TestStatic
```

Only the last of which printed my static initializer:
```
Initializing TestStatic
```

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
rcaudy: I'm suggesting we change our code to do:
```
Class.forName(className, false, GroovyDeephavenSession.class.getClassLoader());
```

nbauernfeind:
Including Ryan's request to avoid initializing the class, this `TestStatic` class is not initialized until after it is actually used.

Results in no output:
```groovy
import io.deephaven.server.TestStatic
```

Can finish loading the class like so:
```groovy
println(TestStatic.class.toString())
```
Yields:
```
- Initializing TestStatic
- class io.deephaven.server.TestStatic
```

rcaudy (approved):

PR #3400, opened 2023-02-02
Fixes SubscribeToLogs example to work with authenticated channel

devinrsmith:


server/src/test/java/io/deephaven/server/console/ConsoleServiceTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+        @Override
+        public void beforeStart(ClientCallStreamObserver<LogSubscriptionRequest> requestStream) {
+            this.stream = requestStream;
+        }
+
+        @Override
+        public void onNext(LogSubscriptionData value) {
+            latch.countDown();
+            ++count;
```
nbauernfeind, line 43: This is likely to be invoked from another thread. You need to make `count`, and `error`, `volatile` to avoid a racy test. (Or synchronize `onNext` / `onError`.)

devinrsmith: I think the semantics provided by the CountDownLatch are good enough - I can refactor a bit to get rid of `count`; and, `onError` / `onComplete` are guaranteed to only be called once, and visibility of `error` should be guarded by `done`. I'll re-push, let me know what you think.

nbauernfeind (approved):

Issue #3323, opened 2023-01-18
Python: Cannot apply both size and shape to points in a plot-by series


mofojed:
**Description**

If you try and apply both a shape and a size to points in a plot-by series, only the shape is applied.

(Note: found this while implementing JS API/UI for plot shapes and sizes, https://github.com/deephaven/deephaven-core/pull/3246 and https://github.com/deephaven/web-client-ui/pull/981)

**Steps to reproduce**
1. Execute the following snippet:
```
from deephaven import empty_table
from deephaven.plot import Figure
import math

math_funcs = ["sin", "cos"]
shapes = ["SQUARE", "CIRCLE", "UP_TRIANGLE", "DOWN_TRIANGLE", "RIGHT_TRIANGLE", "LEFT_TRIANGLE", "DIAMOND"]
size = 50
# Unsupported shapes: ["ELLIPSE", "HORIZONTAL_RECTANGLE", "VERTICAL_RECTANGLE"]

# Create a generic table that has enough columns to display all the shapes
# Re-uses some of the funcs
t = empty_table(size).update([f"name=`` + shapes[(int)Math.floor(i / Math.ceil(size / {len(shapes)}))]", "x=i*0.1"])
for i in range(len(shapes)):
    t = t.update([f"y{i}=Math.{math_funcs[i % len(math_funcs)]}(x+{math.floor(i / len(math_funcs))})"])

# Create a plot by figure
p2 = Figure().plot_xy(series_name="Multi", t=t, by=["name"], x="x", y="y1").line(visible=True)
for i in range(len(shapes)):
    # apply both shapes and sizes to points
    p2 = p2.point(shape=shapes[i], multi_series_key=[shapes[i]], size=len(shapes) - i, visible=True)
p2 = p2.show()
```

**Expected results**
1. `p2` should appear, with each series having different shapes and sizes

**Actual results**
1. `p2` appears, but each series only has different shapes

**Additional details and attachments**

![image](https://user-images.githubusercontent.com/4505624/213205815-cd1dc055-1dd9-4112-a00a-d67044f7599c.png)

**Versions**
Engine Version: 0.20.0
Web UI Version: 0.23.0
Java Version: 11.0.16
Barrage Version: 0.5.0

mofojed:
[...]

PR #3402, opened 2023-02-02
Fix cannot apply both size and shape to points in a plot-by series

jnumainville:
Fixes https://github.com/deephaven/deephaven-core/issues/3323

Verified by running the code from the issue above:

```
import math

math_funcs = ["sin", "cos"]
shapes = ["SQUARE", "CIRCLE", "UP_TRIANGLE", "DOWN_TRIANGLE", "RIGHT_TRIANGLE", "LEFT_TRIANGLE", "DIAMOND"]
size = 50
# Unsupported shapes: ["ELLIPSE", "HORIZONTAL_RECTANGLE", "VERTICAL_RECTANGLE"]

# Create a generic table that has enough columns to display all the shapes
# Re-uses some of the funcs
t = empty_table(size).update([f"name=`` + shapes[(int)Math.floor(i / Math.ceil(size / {len(shapes)}))]", "x=i*0.1"])
for i in range(len(shapes)):
    t = t.update([f"y{i}=Math.{math_funcs[i % len(math_funcs)]}(x+{math.floor(i / len(math_funcs))})"])

# Create a plot by figure
p2 = Figure().plot_xy(series_name="Multi", t=t, by=["name"], x="x", y="y1").line(visible=True)
for i in range(len(shapes)):
    # apply both shapes and sizes to points
    p2 = p2.point(shape=shapes[i], multi_series_key=[shapes[i]], size=len(shapes) - i, visible=True)
p2 = p2.show()
```
With the following graph as an output:
Screenshot 2023-02-02 at 2 47 05 PM

The different series have different shapes and sizes, as expected.

### DIFF
--- a/Generators/src/main/java/io/deephaven/plot/util/GeneratePyV2FigureAPI.java
+++ b/Generators/src/main/java/io/deephaven/plot/util/GeneratePyV2FigureAPI.java
@@ -959,8 +959,6 @@ public class GeneratePyV2FigureAPI {
                 }
 
                 sb.append("\n");
-
-
             }
 
             if (needsMskCheck) {
--- a/Generators/src/main/java/io/deephaven/plot/util/GeneratePyV2FigureAPI.java
+++ b/Generators/src/main/java/io/deephaven/plot/util/GeneratePyV2FigureAPI.java
@@ -913,14 +913,14 @@ public class GeneratePyV2FigureAPI {
                 final boolean mskUsed = Arrays.stream(an).anyMatch("multi_series_key"::equals);
 
                 validateArgNames(an, alreadyGenerated, signatures, pyArgMap);
-                final String[] quoted_an = Arrays.stream(an).map(s -> "\"" + s + "\"").toArray(String[]::new);
+                final String[] quotedAn = Arrays.stream(an).map(s -> "\"" + s + "\"").toArray(String[]::new);
 
                 // prevent removal of multi_series_key until after it's been fully used
-                final String[] filtered_quoted_an = Arrays.stream(quoted_an)
+                final String[] filteredQuotedAn = Arrays.stream(quotedAn)
                         .filter(s -> !s.equals("\"multi_series_key\""))
                         .toArray(String[]::new);
 
-                if (quoted_an.length == 0) {
+                if (quotedAn.length == 0) {
                     sb.append(INDENT)
                             .append(INDENT)
                             .append("if set()")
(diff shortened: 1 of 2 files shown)


Generators/src/main/java/io/deephaven/plot/util/GeneratePyV2FigureAPI.java
```diff
@@ -929,12 +944,33 @@ private void generatePyFuncCallSequential(final StringBuilder sb,
...
+                if (mskUsed) {
+                    sb.append(INDENT)
+                            .append(INDENT)
+                            .append(INDENT)
+                            .append("multi_series_key_used = True\n");
+                }
+
+                sb.append("\n");
+
+
```
mofojed, line 963:
```suggestion
```

mofojed (approved):

jnumainville:

mofojed (approved):

py/server/deephaven/plot/figure.py
```diff
@@ -503,6 +506,9 @@ def chart(
             non_null_args = non_null_args.difference({"pie_label_format"})
             f_called = True
 
+        if multi_series_key_used:
+            non_null_args = non_null_args.difference({"multi_series_key"})
```
jmao-denver, line 510: Adding comments explaining why this is necessary would be nice as it is not obvious from the code itself.

jmao-denver (dismissed):
The change does make the code a little harder to follow but I don't see a clearly better alternative.

py/server/deephaven/plot/figure.py
jnumainville: Will add comment after Chip reviews

Generators/src/main/java/io/deephaven/plot/util/GeneratePyV2FigureAPI.java
```diff
@@ -890,21 +891,35 @@ private void generatePyFuncCallSequential(final StringBuilder sb,
...
                 final Key key = item.first;
                 final String[] an = item.second;
 
+                final boolean mskUsed = Arrays.stream(an).anyMatch("multi_series_key"::equals);
+
                 validateArgNames(an, alreadyGenerated, signatures, pyArgMap);
                 final String[] quoted_an = Arrays.stream(an).map(s -> "\"" + s + "\"").toArray(String[]::new);
 
+                // prevent removal of multi_series_key until after it's been fully used
+                final String[] filtered_quoted_an = Arrays.stream(quoted_an)
```
chipkent, line 919: Variable names are snake case instead of camel case.

py/server/deephaven/plot/figure.py
chipkent: I think the comment in the generator explains things: `// prevent removal of multi_series_key until after it's been fully used`.  @jmao-denver do we need more?

chipkent (changes requested):

py/server/deephaven/plot/figure.py
jmao-denver: No. That should be enough.

chipkent (approved):

PR #3404, opened 2023-02-03
Fix HierarchicalTableImpl bug when replacing an inputed linkage directive with a real one

rcaudy:


engine/table/src/main/java/io/deephaven/engine/table/impl/hierarchical/HierarchicalTableImpl.java
```diff
@@ -848,8 +853,24 @@ private VisitAction getAction() {
...
+                children = new ArrayList<>();
+            }
+            if (removedChildDirective == null) {
+                childDirective.slotInParent = children.size();
+                children.add(childDirective);
+            } else {
+                Assert.eq(children.get(removedChildDirective.slotInParent),
+                        "children.get(removedChildDirective.slotInParent)",
+                        removedChildDirective, "removedChildDirective");
+                Assert.eq(removedChildDirective.action, "removedChildDirective.action", Linkage, "Linkage");
```
niloc132, line 870: i think technically you don't need the "Linkage" string, since the tostring of the enum is its own name

niloc132 (dismissed):

niloc132 (approved):

mofojed (approved):

rcaudy:
Part of #3248

PR #3407, opened 2023-02-06
Add ForwardedRequestCustomizer

devinrsmith:
This allows jetty to correctly handle x-forwarded-* headers.

```sh
$ curl -H "x-forwarded-proto: https" -H "x-forwarded-host: example.com" -I http://localhost:10000
HTTP/1.1 302 Found
Date: Mon, 06 Feb 2023 02:41:34 GMT
Location: https://example.com/ide/
Content-Length: 0
Server: Jetty(11.0.13)
```

Without this patch, we would get `Location: http://localhost:10000/ide/`


niloc132 (approved):
Approved, but where do we consume these, what's the use case or handler for this?

devinrsmith:
I've got a couple of different reverse proxy setups - depending on if you requested the full `/ide/` path, or something like `/ide`, you'd get a redirect and it would ship you back to http. I'm not 100% sure what the "correct" thing to do, or if there is even a definitive standard in this space. Some proxies might "fix" a misbehaving server, but other proxies might assume the "server knows best", and so proceed w/ the "bad" Location.

PR #3408, opened 2023-02-06
Compute `UpdateBy` results correctly for in-memory columns

lbooker42:
@pete-petey uncovered a bug when attempting to compute output values for in-memory sources.  The reproducing python code is the following:
```
import os
import deephaven.pandas as dhpd
os.system('pip install Nasdaq-Data-Link')

import nasdaqdatalink
mydata = dhpd.to_table(nasdaqdatalink.get("WIKI/MSFT")).update("Sym = `MSFT`")

from deephaven.updateby import rolling_sum_tick
rev_ticks = 10
fwd_ticks = 0

rolling_volume = mydata.update_by([
    rolling_sum_tick(cols="Rolling_Volume = Adj_Volume", rev_ticks=10, fwd_ticks=0)], by="Sym")
```

The error is that the output columns from UpdateBy are not computed but remain `null`. 

This fix addresses the root problem and provides a test case to ensure that the in-memory input source columns are computed correctly.

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateBy.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateBy.java
@@ -201,7 +201,6 @@ public abstract class UpdateBy {
                 .toArray();
         inputCacheNeeded = cacheableSourceIndices.length > 0;
 
-        // noinspection unchecked
         inputSourceCaches = new SoftReference[inputSources.length];
 
         buckets =
@@ -317,7 +316,11 @@ public abstract class UpdateBy {
                 inputSourceReferenceCounts = null;
             }
 
+
             if (initialStep) {
+                // Set all windows as dirty and need computation
+                Arrays.fill(dirtyWindows, true);
+                // Create the proper JobScheduler for the following parallel tasks
                 if (OperationInitializationThreadPool.NUM_THREADS > 1
                         && !OperationInitializationThreadPool.isInitializationThread()) {
                     jobScheduler = new OperationInitializationPoolJobScheduler();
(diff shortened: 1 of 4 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateByWindow.java
```diff
@@ -231,7 +231,7 @@ abstract void computeAffectedRowsAndOperators(final UpdateByWindowBucketContext
     void assignInputSources(final UpdateByWindowBucketContext context, final ColumnSource<?>[] inputSources) {
         context.inputSources = inputSources;
         context.inputSourceGetContexts = new ChunkSource.GetContext[inputSources.length];
-        context.inputSourceChunks = new WritableChunk[inputSources.length];
+        context.inputSourceChunks = new Chunk[inputSources.length];
```
lbooker42, line 234: This did not need to be `WritableChunk`, corrected.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateBy.java
```diff
@@ -358,12 +358,25 @@ private void onError(@NotNull final Exception error) {
          * {@code completedAction} when the work is complete
          */
         private void computeCachedColumnRowsets(final Runnable resumeAction) {
+            // Determine which windows need to be computed.
```
rcaudy, line 361: I think `processUpdate` could return earlier if there are no dirty buckets.
This could happen if there were only shifts, I think. Also if there were only mods to rows that influenced nothing.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateBy.java
```diff
@@ -409,8 +421,6 @@ private void computeCachedColumnRowsets(final Runnable resumeAction) {
                                         }
                                         // at least one dirty bucket will need this source
                                         srcNeeded = true;
-                                        // this window must be computed
-                                        dirtyWindows[winIdx] = true;
```
rcaudy, line 413: We might be able to short-circuit some of the checks in this loop if we checked `dirtyWindows[winIdx]` before looking at the buckets. Also, I found it very confusing even on re-read that the window bucket context is called `winCtx` and the check method is `isWIndowDirty` not `isWindowBucketDirty` or similar.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateBy.java
rcaudy, line 361: I think this code belongs in its own method now, with a more appropriate name, and called before `computeCachedColumnRowsets`.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateByWindow.java
lbooker42: This was a serendipitous finding while writing the test.  `ObjectSparseArraySource.getChunk()` will return `ObjectChunk.getEmptyChunk()` when passed `Rowset.EMPTY`.  This is non-writable and will fail the array store auto-cast.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateBy.java
```diff
@@ -201,7 +201,6 @@ protected UpdateBy(
                 .toArray();
         inputCacheNeeded = cacheableSourceIndices.length > 0;
 
-        // noinspection unchecked
```
rcaudy, line 204: I think IntelliJ is lying to you, it depends on the inspection profile.

rcaudy (approved):

PR #3415, opened 2023-02-07
Fixed `seekRow` Error on Char Column

Zhou-Ziheng:
Previously, calling `seekRow` with a `Char` Column would throw an error. This is now fixed.


mofojed (approved):
I don't think release notes are needed for this

PR #3416, opened 2023-02-07
Wrap the withAttributes/getAttrbutes methods

### LINKED ISSUE #3376: Add a wrapper of WithAttributes in Table
Deephaven's JS-plugins example (https://github.com/deephaven/deephaven-js-plugins/tree/main/table-example) relies on the access of the Table wrapper's j_table property and its JPY wrapped method setAttribute() which has been refactored to withAttributes() in the AttributeMap interface. Despite that the JS-plugins feature is for very advanced users, forcing the users to know the Java interface in Python goes against our goal of making our Python API as transparent and Pythonic as possible.

jmao-denver:
Fixes #3376

also include some code cleaning up and improvement in jcompat.py.

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -320,12 +320,12 @@ class Table(JObjectWrapper):
     def j_object(self) -> jpy.JType:
         return self.j_table
 
-    def get_table_attributes(self) -> Dict[str, Any]:
+    def attributes(self) -> Dict[str, Any]:
         """Returns all the attributes defined on the table."""
         j_map = jpy.cast(self.j_table, _JLiveAttributeMap).getAttributes()
         return j_map_to_dict(j_map)
 
-    def with_table_attributes(self, attrs: Dict[str, Any]) -> Table:
+    def with_attributes(self, attrs: Dict[str, Any]) -> Table:
         """Returns a new Table that has the provided attributes defined on it and shares the underlying data and schema
         with this table.
 
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -346,7 +346,7 @@ class Table(JObjectWrapper):
             j_map = j_hashmap(attrs)
             return Table(j_table=jpy.cast(self.j_table, _JLiveAttributeMap).withAttributes(j_map))
         except Exception as e:
-            raise DHError(e, "failed to TODO") from e
+            raise DHError(e, "failed to create a table with attributes.") from e
 
     def to_string(self, num_rows: int = 10, cols: Union[str, Sequence[str]] = None) -> str:
         """Returns the first few rows of a table as a pipe-delimited string.
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 import contextlib
 import inspect
 from enum import Enum, auto
-from typing import Union, Sequence, List, Any, Optional, Callable
+from typing import Union, Sequence, List, Any, Optional, Callable, Dict
 
 import jpy
 import numpy as np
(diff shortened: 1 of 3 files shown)


jmao-denver:
@jjbrosnan @mattrunyon I include you because once it is merged, we also need to modify the JS plug-in example which gave rise to this change

py/server/deephaven/table.py
```diff
@@ -319,6 +320,34 @@ def meta_table(self) -> Table:
     def j_object(self) -> jpy.JType:
         return self.j_table
 
+    def get_table_attributes(self) -> Dict[str, Any]:
```
rcaudy, line 323: Do we want to offer the calls that allow users to fetch a single attribute, or subset of attribtues?

py/server/deephaven/table.py
```diff
@@ -319,6 +320,34 @@ def meta_table(self) -> Table:
...
+            attrs (Dict[str, Any]): a dict of table attribute names and their values
+
+        Returns:
+            a new Table
+
+        Raises:
+            DHError
+        """
+        try:
+            j_map = j_hashmap(attrs)
```
rcaudy, line 346: Are we unwrapping objects as they're copied into the map? Should we?

py/server/deephaven/table.py
jmao-denver: Unless we have or going to have a huge number of such attributes, I don't think it is necessary.

py/server/deephaven/table.py
jmao-denver: Yes and we should, otherwise we'll run into a problem with JPY when the value is a wrapped Java object.

py/server/deephaven/table.py
mattrunyon, line 323: Why `get_table_attributes` and `with_table_attributes` instead of `get_attributes` and `with_attributes`? It's a method on a table so the `_table_` part seems redundant

jmao-denver: Python users are conditioned to access object attributes in a Pythonic way, i.e. through the __dict__ attribute (the built-in vars() function, getattr(), setattr()), __getattr__, __getattribute__, and the more advanced descriptors. Because the name get_attributes() is a bit too generic, so in order to highlight the difference and to avoid questions like 'why do you need get_attributes', I came up with those names without too much conviction. Suggestions for better names are welcome.

chipkent, line 323: This method seems to be studdering by having `table` in it.  Also, `get` is very java like.  Should it just be `attributes`?

py/server/deephaven/table.py
```diff
@@ -319,6 +320,34 @@ def meta_table(self) -> Table:
     def j_object(self) -> jpy.JType:
         return self.j_table
 
+    def get_table_attributes(self) -> Dict[str, Any]:
+        """Returns all the attributes defined on the table."""
+        j_map = jpy.cast(self.j_table, _JLiveAttributeMap).getAttributes()
+        return j_map_to_dict(j_map)
```
chipkent, line 326: We are bleeding java objects!  `getAttributes()` returns `Map<String,Object>`, and `j_map_to_dict` converts to a dict without wrapping the result types.

1. What are the attribute types?
2. How do we handle them in Python without bleeding Java objects?

chipkent (changes requested):

py/server/deephaven/table.py
jmao-denver: 1. Even though from the current usages, the types of attributes are only primitive ones and strings, it does leave doors open for more complex objects, so type hint [Any] is appropriate.
2. We'll wrap the Java objects (non primitive/strings) if a wrapper exists, otherwise it will just be JPY wrapped ones and should be used as such. But most likely if it is something important for the users we should already have wrapped it in Python.

py/server/deephaven/table.py
jjbrosnan: After looking this over, I agree with Chip's suggestion to use `attributes` here.

py/server/deephaven/table.py
```diff
@@ -319,6 +320,34 @@ def meta_table(self) -> Table:
...
+
+    def with_table_attributes(self, attrs: Dict[str, Any]) -> Table:
```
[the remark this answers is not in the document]
jjbrosnan: `with_attributes` or `w_attributes` both could work. Probably the former, since we're not abbreviating other methods.

jjbrosnan (commented):
Comments on method names. I concur with Chip.

chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2218
Conceptual: https://github.com/deephaven/deephaven.io/issues/2217
Reference: https://github.com/deephaven/deephaven.io/issues/2220
Blog: https://github.com/deephaven/deephaven.io/issues/2219

PR #2903, opened 2022-09-22
Add `RollingSum` operator to `updateBy()`

lbooker42:
This is very large commit containing the back-end support for all upcoming windowed/rolling operators and the full implementation of `RollingSum`.

Closes #2624

PR #3418, opened 2023-02-08
Optimize/simplify Ring buffer data structures

### LINKED ISSUE #3391: Improve/simplify `RingBuffer` and `PairwiseBuffer` data structures
Implement the suggestions from the review conducted in PR #2903 by @kosak.

- Documentation improvements
- Simplified indexing for push/pop and dirty entry tracking
- Simplified `dirty` entry evaluation for Pairwise buffers

lbooker42:
Follow suggestions from @kosak, and improve the `RingBuffer` and `PairwiseBuffer` data structures as found on merged PR: #2903

This will close #3391  when completed


engine/table/src/main/java/io/deephaven/engine/table/impl/SortListener.java
```diff
@@ -710,7 +710,7 @@ public void add(long numWrites, long numRequestedAdds) {
 
             // assert that the buffers are not overwritten
             Assert.eqTrue(writes.offer(numWrites), "writes.offer(numWrites)");
-            Assert.eqTrue(requestedAdds.add(numRequestedAdds), "requestedAdds.add(numRequestedAdds)");
+            Assert.eqTrue(requestedAdds.offer(numRequestedAdds), "requestedAdds.offer(numRequestedAdds)");
```
lbooker42, line 713: This change has two purposes: to call attention to the only pre-`updateBy` usage of these data structures and because I think the code was inconsistent and should be changed.

lbooker42: Discussed the impact of the change on this code w/ @rcaudy, agreed this change has zero impact.

Base/src/main/java/io/deephaven/base/ringbuffer/CharRingBuffer.java
```diff
@@ -3,106 +3,117 @@
...
+        mask = storage.length - 1;
+        tail = size;
+        head = 0;
+    }
+
+    private void copyRingBufferToArray(char[] dest) {
+        final int size = size();
+        final int storageHead = (int) (head & mask);
+
+        // firstCopyLen is either the size of the ring buffer, the distance from head to the end, or the size
```
kosak, line 83: change "to the end" to "to the end of the storage array"

kosak (approved):
A bunch of smallish changes, but otherwise approved.

Base/src/main/java/io/deephaven/base/ringbuffer/CharRingBuffer.java
```diff
@@ -144,15 +155,14 @@ public void ensureRemaining(int count) {
...
+     * structure becoming corrupted and unusable. This call must be used in conjunction with
+     * {@link #ensureRemaining(int)} or {@link #remaining()} to verify remaining capacity if sufficient.
      *
      * @param e the value to add to the buffer
      */
     public void addUnsafe(char e) {
-        storage[tail] = e;
-        tail = (tail + 1) % storage.length;
-        size++;
+        storage[(int) (tail++ & mask)] = e;
```
kosak, line 165: I forgot to ask. Have we decided on a policy (ignore vs explicitly check) for int64 wraparound?

lbooker42: Ignore, considering that we will reset with every call to grow().

kosak: IMO, that doesn't sound like a valid rationale to me. A ring buffer only grows a few times in its lifetime, but on the other hand can be in a steady state (adding/removing data) for an arbitrarily long time without ever growing.

I'm actually ok with not checking, but with the right rationale: namely that that int64s are really big numbers and they're not going to wrap during any particular run of the program.

I'm also ok with checking.

lbooker42: I added some code to the safe `add()` operations and to `remaining()` and `ensureRemaining()`.  I'm probably trying to hard to keep the unsafe code streamlined, I can be convinced to put this in `addUnsafe()` and trust the predictor.  Lemme know what you think.

Base/src/main/java/io/deephaven/base/ringbuffer/CharRingBuffer.java
```diff
@@ -121,6 +135,7 @@ public int capacity() {
     }
 
     public int remaining() {
+        maybeFixIndices();
```
kosak, line 138: I don't personally much like all these calls to `maybeFixIndices()` sprinkled throughout the code (5 different places by my count). It makes it difficult for future readers/maintainers to have confidence exactly what code needs a call to `maybeFixIndices`. For example they may wonder (as I do) why a non-mutating method like `remaining()` needs a fix-indices check. In my opinion the answer to all these questions is "the only code that might need fix-indices is code that changes `tail`". Along those lines the check needs to be in only one place: inside `addUnsafe`. I realize that doing so makes `addUnsafe` look longer and less tight, but I want to double- and quadruple- down on the argument that highly-predictable branches are basically free. "Most" user code is going to do the wrap check somewhere anyway, so from a performance standpoint it's moot whether the tail got checked inside `add()` or `addUnsafe()`. For code where it does matter (some hypothetical tight loop that doesn't call `add()`, only `addUnsafe()`), I think we can rely on the branch predictor to smoothly skip over the cost of the `if`.

If you're really unconvinced, you could also run your (existing! not new) benchmarks and/or tests with and without the wrap check to confirm that there is no wall-clock-observable difference between the presence and absence of the wrap check. Since we know that the wrap check is only a theoretical hazard anyway, we could run our code with no checks at all to see if it's any faster.

If you do buy all of the above, you could inline maybeFixIndices() inside addUnsafe() rather than being a separate function call to make it feel a little tighter.

lbooker42: I'm convinced.

Base/src/main/java/io/deephaven/base/ringbuffer/AggregatingCharRingBuffer.java
```diff
@@ -0,0 +1,448 @@
...
+        }
+
+        // Store our computed range
+        calcHead = head;
+        calcTail = tail;
+
+        return treeStorage[1];
+    }
+
+    void normalizeAndEvaluate(final long head1, final long tail1, final long head2, final long tail2) {
```
kosak, line 277: Personally I feel like this method and its various helper methods are too complicated and hard to analyze. I think the job they are trying to accomplish -- namely, keep track of up to three ranges and possibly merge them as you go up levels in the tree -- can be expressed more simply.

I did a prototype implementation in https://github.com/deephaven/deephaven-core/pull/3378 that you may want to take a look at. I took a couple of shortcuts e.g. by calling sort rather than manually ordering my ranges, but overall the resultant code is so short I think you may find the approach valuable.

Base/src/main/java/io/deephaven/base/ringbuffer/AggregatingObjectRingBuffer.java
```diff
@@ -0,0 +1,436 @@
...
+ * Aggregation is performed by calling the aggregation function on pairs of values from the circular buffer. The results
+ * of the aggregations are stored into a parallel tree of result values where the root node contains the overall
+ * aggregation of all the leaf nodes. Performance is improved by performing aggregation over only those values which
+ * have changed since the most recent evaluation.
+ */
+
+public class AggregatingObjectRingBuffer<T> extends ObjectRingBuffer<T> {
+    private final ObjectFunction pairwiseFunction;
+    private final T identityVal;
+    private Object[] treeStorage;
```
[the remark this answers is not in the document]
lbooker42: Corrected.

Base/src/main/java/io/deephaven/base/ringbuffer/AggregatingCharRingBuffer.java
```diff
@@ -0,0 +1,712 @@
...
+                offset + (r3h / 2), offset + (r3t / 2));
+    }
+
+    /**
+     * This function accepts three unique ranges in the evaluation tree and iteratively computes aggregation results of
+     * value pairs in the range which are stored in parent nodes. Each iteration will shrink the ranges by a factor of
+     * 2. As the ranges shrink, an overlap will occur and this function will transfer to
+     * {@link #evaluateTwoRanges(int, int, int, int)} for better performance.
+     */
+    private void evaluateThreeRanges(int r1h, int r1t, int r2h, int r2t, int r3h, int r3t) {
```
kosak, line 621: By the way, this new organization of the code is a lot more readable for me. So thank you for that.

Base/src/main/java/io/deephaven/base/ringbuffer/AggregatingCharRingBuffer.java
```diff
@@ -0,0 +1,712 @@
...
+        while (true) {
+            if (r1t >= r2h) {
+                // r1 and r2 overlap, collapse them together and call the two range version.
```
kosak, line 624: period rather than comma, or "so"

```
r1 and r2 overlap. Collapse them together
```

or

```
r1 and r2 overlap, so collapse them together
```

Base/src/main/java/io/deephaven/base/ringbuffer/AggregatingCharRingBuffer.java
```diff
@@ -0,0 +1,712 @@
...
+                evaluateTwoRanges(r1h, r2t, r3h, r3t);
+                return;
+            }
+            if (r2t >= r3h) {
+                // r2 and r3 overlap, collapse them together and call the two range version.
```
kosak, line 629: Likewise re period and comma

```
r2 and r3 overlap. Collapse them together
```

or

``
r2 and r3 overlap, so collapse them together
```

File: Base/src/main/java/io/deephaven/base/ringbuffer/AggregatingCharRingBuffer.java [hunk 10]
```diff
@@ -0,0 +1,712 @@
...
+                evaluateTwoRanges(r1h, r1t, r2h, r3t);
+                return;
+            }
+
+            // No collapse is possible, compute the unique ranges.
```
kosak, line 634: Likewise, period or comma with "so"

Base/src/main/java/io/deephaven/base/ringbuffer/AggregatingCharRingBuffer.java
```diff
@@ -0,0 +1,712 @@
...
+    /**
+     * This function accepts two unique ranges in the evaluation tree and iteratively computes aggregation results of
+     * value pairs in the range which are stored in parent nodes. Each iteration will shrink the ranges by a factor of
+     * 2. As the ranges shrink, an overlap will occur and this function will transfer to
+     * {@link #evaluateRange(int, int)} for better performance.
+     */
+    private void evaluateTwoRanges(int r1h, int r1t, int r2h, int r2t) {
+        while (true) {
+            if (r1t >= r2h) {
+                // r1 and r2 overlap, collapse them together and call the single range version.
```
kosak, line 658: period / comma with so

kosak (approved):

PR #3420, opened 2023-02-08
Port DH-12458: ColumnSources That Wrap Native Apache Arrow

### LINKED ISSUE #2923: Remove explicit constraint on commons-codec
During investigation around https://github.com/deephaven/deephaven-core/pull/2922, it was discovered that arrow-vector:7.0.0 (and 8.0.0) depends on commons-codec:1.10 which has a vulnerability. As such, we've constrained commons-codec explicitly to 1.15. This explicit constraint should be removed once arrow has been upgraded to 9.0.0+.

nbauernfeind:
This is a pretty close port of DHE's Arrow wrapped column sources. There are some differences; I didn't understand the point of writing a generator for the ArrowWrapperTool which wouldn't be re-used in multiple contexts.

Ports:
[DH-12458](https://deephaven.atlassian.net/browse/DH-12458) via https://github.com/deephaven-ent/iris/commit/8fbc8493c2a4ab6408d9ead555c6c1150fd5db64
[DH-13953](https://deephaven.atlassian.net/browse/13953) via https://github.com/deephaven-ent/iris/commit/3a2b79c124593b9c5879b3c3f30efd30d7c28c81

Fixes #2923.

Follow-Up Work: #3633


extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
```diff
@@ -0,0 +1,340 @@
...
+
+        @Override
+        public void close() {
+            if (!shared) {
+                helper.addToPool(shareable);
+            }
+        }
+    }
+
+    public static final class Shareable extends SharedContext {
```
rcaudy, line 255: Can be a `ResettableContext` instead of a `SharedContext`. Also, see the changes Andy made in a round 2 commit.

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
```diff
@@ -0,0 +1,340 @@
...
+     * Reads arrow data from file and returns a query table
+     */
+    public static QueryTable readArrow(final @NotNull String path) {
+        final Helper helper = new Helper(path, rootAllocator);
+        final Shareable context = helper.newShareable();
+        final ArrowFileReader reader = context.getReader();
+        try {
+            int biggestBlock = 0;
+            final List<ArrowBlock> recordBlocks = reader.getRecordBlocks();
+            final int[] blocks = new int[recordBlocks.size()];
```
rcaudy, line 110: You told me blocks can be bigger than Integer.MAX_VALUE, because they are not stored in arrays. Maybe you are lying, or maybe the code is wrong, or maybe the Java arrow impl can't handle it?

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/sources/AbstractArrowColumnSource.java
```diff
@@ -0,0 +1,74 @@
...
+                int blockNumber = getBlockNo(nextKey);
+                long lastKeyInBlock = nextKey | (highBit - 1);
+                if (recordBatchKeys == null || lastKeyInBlock != lastKeyInBlockCache) {
+                    recordBatchKeys = okIt.getNextRowSequenceThrough(lastKeyInBlock);
+                    lastKeyInBlockCache = lastKeyInBlock;
+                }
+                context.ensureLoadingBlock(blockNumber);
+                recordBatchKeys.forAllRowKeys(rowKeyConsumer);
+            }
+        }
```
rcaudy, line 64: The caching is useless.
```suggestion
        try (RowSequence.Iterator okIt = rowSequence.getRowSequenceIterator()) {
            while (okIt.hasMore()) {
                final long nextKey = okIt.peekNextKey();
                final int blockNumber = getBlockNo(nextKey);
                final long lastKeyInBlock = nextKey | (highBit - 1);
                final RowSequence recordBatchKeys = okIt.getNextRowSequenceThrough(lastKeyInBlock);
                context.ensureLoadingBlock(blockNumber);
                recordBatchKeys.forAllRowKeys(rowKeyConsumer);
            }
        }
```

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/sources/ArrowBooleanColumnSource.java
```diff
@@ -0,0 +1,53 @@
...
+    public final Boolean get(final long rowKey) {
+        try (ArrowWrapperTools.FillContext fc = (ArrowWrapperTools.FillContext) makeFillContext(1)) {
+            fc.ensureLoadingBlock(getBlockNo(rowKey));
+            return extract(getPositionInBlock(rowKey), fc.getVector(field));
+        }
+    }
+
+    private Boolean extract(final int posInBlock, final BitVector vector) {
+        return vector.getObject(posInBlock);
+    }
```
rcaudy, line 52: Could we do better by doing bulk copies of the vector and validity buffer and reconciling nulls ourselves afterwards? (I'm not sure.)

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/sources/AbstractArrowColumnSource.java
```diff
@@ -0,0 +1,74 @@
...
+        return new ArrowWrapperTools.FillContext(arrowHelper, sharedContext);
+    }
+
+    protected final void fillChunk(
+            @NotNull final ArrowWrapperTools.FillContext context,
+            @NotNull final RowSequence rowSequence,
+            @NotNull final LongConsumer rowKeyConsumer) {
+        long lastKeyInBlockCache = QueryConstants.NULL_LONG;
+        RowSequence recordBatchKeys = null;
+        try (RowSequence.Iterator okIt = rowSequence.getRowSequenceIterator()) {
```
rcaudy, line 52: We can skip making this iterator if the input `rowSequence` maps entirely to one record batch. You could delegate to a method "append to chunk from single record batch", either with the entire row sequence, or the row sequence sliced down (`recordBachKeys`).

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/sources/AbstractArrowColumnSource.java
```diff
@@ -0,0 +1,74 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.util.function.LongConsumer;
+
+/**
+ * Base class for arrow column sources
+ * 
+ * @param <T>
+ */
+public abstract class AbstractArrowColumnSource<T> extends AbstractColumnSource<T> {
```
rcaudy, line 20: This feels *so* similar to the regioned/page code. I wonder if we could re-use that?

rcaudy:
I wonder if we could abstract this in order to allow some of it to be re-usable for in-memory arrow data. It would be nice for the ODBC or ADBC integrations.

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
```diff
@@ -0,0 +1,340 @@
...
+            for (int i = 0; i < recordBlocks.size(); i++) {
+                reader.loadRecordBatch(recordBlocks.get(i));
```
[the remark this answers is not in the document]
nbauernfeind: We double checked the code paths to ensure that arrow buffers are indeed being freed when a new record batch is loaded.

I do believe I can improve this by seeking ourselves and only reading the metadata. Not closing this comment until I've done that and can verify improved latencies.

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
nbauernfeind: It appears that while the initial portion of the parsing allows for `long` sized `RecordBatch`es, that the actual `Vector` implementations that interact with underlying `ArrowBuf` expect `int` based indexing. This is consistent with "best-practices" that arrow implementations limit record batch lengths to MAX_INT.

extensions/arrow/build.gradle
```diff
@@ -3,33 +3,16 @@ plugins {
...
-
-configurations {
-    implementation.extendsFrom irisDb
-    testImplementation.extendsFrom irisDbTest
-}
-
 description = 'An extension integrating Arrow sources with Deephaven\'s Table Model'
 
 dependencies {
+    api project(':Configuration')
```
rcaudy, line 9: Why is this API?

extensions/arrow/src/test/java/io/deephaven/extensions/arrow/ArrowVarCharVectorTest.java
```diff
@@ -34,7 +34,7 @@ public class ArrowVarCharVectorTest {
     private static QueryTable loadTable() {
         //noinspection ConstantConditions;
         final File dataFile = new File(ArrowVarCharVectorTest.class.getResource("/var_char_vector.arrow").getFile());
-        return ArrowWrapperTools.readArrow(dataFile.getPath());
+        return ArrowWrapperTools.readFeather(dataFile.getPath());
```
rcaudy, line 37: Can you find test data and confirm if we can handle FeatherV1, FeatherV2, or both?

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
```diff
@@ -70,162 +81,169 @@
 import org.apache.arrow.vector.types.pojo.Field;
 import org.jetbrains.annotations.NotNull;
 
+import static org.apache.arrow.vector.ipc.message.MessageSerializer.IPC_CONTINUATION_TOKEN;
+
 /**
  * <pre>
```
rcaudy, line 87: Doesn't need a pre block anymore.

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
```diff
@@ -70,162 +81,169 @@
...
-                    biggestBlock = schemaRowCount;
+            // TODO: load only the metadata to speed up initial read time
+
+            int metadataBufLen = 1024;
+            byte[] rawMetadataBuf = new byte[metadataBufLen];
+            for (int ii = 0; ii < blocks.length; ++ii) {
+                final ArrowBlock block = recordBlocks.get(ii);
+                while (block.getMetadataLength() > metadataBufLen) {
+                    metadataBufLen *= 2;
+                    rawMetadataBuf = new byte[metadataBufLen];
```
rcaudy, line 137: Clamp to `io.deephaven.base.ArrayUtil#MAX_ARRAY_SIZE`

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
```diff
@@ -70,162 +81,169 @@
...
                 }
+
+                channel.position(block.getOffset());
+
+                final ByteBuffer metadataBuf = ByteBuffer.wrap(rawMetadataBuf, 0, block.getMetadataLength());
+                int numRead = channel.read(metadataBuf);
+                if (numRead != block.getMetadataLength()) {
+                    throw new IOException("Unexpected end of input trying to read batch.");
```
rcaudy, line 145: Maybe we could offer more information (file name, number read and expected, block index, mother's maiden name, etc).

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
```diff
@@ -70,162 +81,169 @@
...
+                }
+
+                metadataBuf.flip();
+                if (metadataBuf.getInt() == IPC_CONTINUATION_TOKEN) {
+                    // if the continuation token is present, skip the length
+                    metadataBuf.getInt();
```
rcaudy, line 151: `metadaBuf.position(metadataBuf.position() + Integer.BYTES)`?

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
```diff
@@ -70,162 +81,169 @@
...
+                }
+                final Message message = Message.getRootAsMessage(metadataBuf.asReadOnlyBuffer());
+                final RecordBatch batch = (RecordBatch) message.header(new RecordBatch());
```
rcaudy, line 154: Could we be re-using a flatbuffer object?

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
```diff
@@ -300,41 +337,69 @@ void ensureLoadingBlock(final int blockNumber) {
...
+
+        private ReaderCleanup(final Shareable shareable) {
+            super(shareable, CleanupReferenceProcessorInstance.DEFAULT.getReferenceQueue());
+            this.reader = shareable.getReader();
+        }
+
+        @Override
+        public void cleanup() {
+            try {
+                reader.close();
```
rcaudy, line 399: Please test that this works.

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/sources/AbstractArrowColumnSource.java
nbauernfeind: I'm not sure we can use the regioned page code. A RecordBatch might have size up to MAX_ARRAY_LEN (if the feather file is playing nice to java-users); 2^20 is probably not large enough for arbitrary files.

rcaudy: Regions max out much larger than 2^20.

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
```diff
@@ -70,162 +82,204 @@
...
             }
-            return getQueryTable(context, highBit, builder.build().toTracking(), helper);
+
+            final VectorSchemaRoot root = context.getReader().getVectorSchemaRoot();
+            final Map<String, AbstractColumnSource<?>> sources = root.getSchema()
+                    .getFields().stream().collect(Collectors.toMap(
+                            Field::getName,
+                            f -> generateColumnSource(root.getVector(f), f, highBit, arrowTableContext),
+                            (u, v) -> {
+                                throw new IllegalStateException(String.format("Duplicate key %s", u));
```
rcaudy, line 186: Same comment as earlier, exceptions would be more helpful with more context.

rcaudy (commented):
I've tried to refresh my memory from the prior review, and looked through commits 2-7. I don't really have any complaints about the direction of things.

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/sources/ArrowBooleanColumnSource.java
nbauernfeind: This ends up being a wash due to the need to reconcile nulls. The non-direct-buffer copying version is simpler and will be easier to maintain. I've filed #3633 which would load data directly into a page cache via Chunks (via Barrage code paths) and is more likely to yield better performance.

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/sources/AbstractArrowColumnSource.java
nbauernfeind: I've filed #3633 - I think this is a good idea, but after digging into the Parquet implementation I feel that we should do this only if we have users (or customers) who really need a more efficient implementation. For this, I think we lose most of the original DHE implementation, increases scope, and might make it difficult to meet our porting goal on time.

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2390
Conceptual: https://github.com/deephaven/deephaven.io/issues/2389
Reference: https://github.com/deephaven/deephaven.io/issues/2391
Blog: https://github.com/deephaven/deephaven.io/issues/2388

Issue #3421, opened 2023-02-09
DH URI / BarrageTableResolver authentication support


devinrsmith:
As authentication becomes more real, we'll want some way for DH URIs and BarrageTableResolver to support authentication.

There is potential for it to be sourced from their existing credentials if applicable (for example, between an already authenticated user on this server who wants to use the same identity on another server); or for the user to explicitly set the authentication context.

It's unclear if we want to support the ability to provide authentication constants in the DH URI itself. While this would be possible, I'm not sure if it's advisable. I could convince myself it's "OK" for psk and Basic username / "password".

nbauernfeind:
Java's `URI` model allows for extracting the query parameter portion of a URI. This could be useful for PSK style auth.

We could also possibly add for some sort of API to set an "AuthenticationProvider" that would provide auth details for a given URI.

PR #3423, opened 2023-02-09
Migrates the java client authentication to Flight Auth v2

### LINKED ISSUE #3285: Java Client: Migrate off of deprecated newSession/refreshSessionToken to Flight Auth
DHE, as a client-side user of DHC, needs to use Flight Auth. The current version of `io.deephaven.client.impl.SessionImpl` still uses `SessionService::NewSession` and `SessionService::RefreshSessionToken` which are deprecated as they do not fit into either Flight Auth v1 or Flight Auth v2 molds.

There are tests in `FlightMessageRoundTripTest` that demonstrate how to use Flight Auth v1 and Flight Auth v2. However, they skirt around the need to worry about a session timing out and do not automatically refresh tokens (these are unit tests that aim to catch edge cases and are not concerned with true passage of time).

Fixing up `SessionImpl` to use the appropriate implementation for authentication will make client-behavior, under the reign of Flight Auth, more obvious, clear, and frankly, usable by DHE.

devinrsmith:
There are also plumbing improvements that allow constructing sessions with specific authentication values, and the sharing of authenticated channels without extraneous duplication of logic.

Fixes #3285


devinrsmith:
This work was done in support of testing #3410. I wanted to take a bit further wrt DH URIs and BarrageTableResolver, but have time-bounded this work for now and filed https://github.com/deephaven/deephaven-core/issues/3421

java-client/barrage-dagger/src/main/java/io/deephaven/client/impl/BarrageSubcomponent.java
```diff
@@ -33,6 +32,9 @@ interface Builder extends BarrageSessionFactoryBuilder {
 
         Builder allocator(@BindsInstance BufferAllocator bufferAllocator);
 
+        Builder authenticationTypeAndValue(
+                @BindsInstance @Nullable @Named("authenticationTypeAndValue") String authenticationTypeAndValue);
```
nbauernfeind, line 36: Is this sufficient? Does this allow for bi-directional authentication schemes?

java-client/example-utilities/src/main/java/io/deephaven/client/examples/AuthenticationOptions.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Option(names = {"--psk"}, description = "The pre-shared key")
+    String psk;
+
+    @Option(names = {"--explicit"}, description = "The explicit authentication type and value")
+    String explicit;
+
+    public String toAuthenticationTypeAndValue() {
+        if (mtls != null && mtls) {
+            // todo
```
nbauernfeind, line 19: Is this a lost comment? Or should we add an issue tag?

java-client/flight/src/main/java/io/deephaven/client/impl/FlightSession.java
```diff
@@ -20,6 +20,10 @@ public class FlightSession implements AutoCloseable {
 
     public static FlightSession of(SessionImpl session, BufferAllocator incomingAllocator,
             ManagedChannel channel) {
+        // Note: this pattern of FlightClient owning the ManagedChannel does not mesh well with the idea that some
```
nbauernfeind, line 23: This comment is golden. Thanks; well said.

java-client/session/src/main/java/io/deephaven/client/impl/SessionImpl.java
```diff
@@ -69,156 +63,77 @@
...
+        this.bearerChannel = Objects.requireNonNull(bearerChannel);
+        this.bearerHandler = Objects.requireNonNull(bearerHandler);
         this.exportTicketCreator = new ExportTicketCreator();
-        this.states = new ExportStates(this, sessionService, config.channel().table().withCallCredentials(credentials),
-                exportTicketCreator);
+        this.states = new ExportStates(this, bearerChannel.session(), bearerChannel.table(), exportTicketCreator);
         this.serialManager = TableHandleManagerSerial.of(this);
         this.batchManager = TableHandleManagerBatch.of(this, config.mixinStacktrace());
+        this.pingJob = config.executor().scheduleAtFixedRate(
+                () -> bearerChannel.config().getConfigurationConstants(
```
nbauernfeind, line 130: Cristian pointed out that when the server was unavailable that the client retried in an excessive repetitive manner. Have you verified that this no longer happens? I believed it would require setting alternate grpc call options.

Also, Colin pointed out that possibly ConfigService might run on another host (say, for a cluster of instances) - should we add a ping grpc call directly to the `SessionService` in the event that we're hitting envoy and being routed to different hosts?

java-client/example-utilities/src/main/java/io/deephaven/client/examples/AuthenticationOptions.java
devinrsmith: Meant to remove. We may eventually want proper string constants somewhere publicly accessible for authentication methods we support out of the box.

java-client/session/src/main/java/io/deephaven/client/impl/SessionImpl.java
devinrsmith: I've taken out that more "complicated" retry logic, and figured that using a fixed job that "pings" multiple times during the timeout interval is "good enough". (I don't see much value in some sort of backoff schedule... either we successfully reauth before the timeout interval, or we don't.)

I'd very much be in favor of a simple no-op "ping" RPC in `SessionService`; although, I think @niloc132 might have opinions? The general advice might be "use Handshake", but the code as written right now doesn't depend on Flight. Right now I'm essentially (ab)using `ConfigService#getConfigurationConstants` since I know the user is authorized for it.

java-client/barrage-dagger/src/main/java/io/deephaven/client/impl/BarrageSubcomponent.java
devinrsmith: No.

I think we want a working example we can interact with to come up with a better interface. Either way, the structure in main right now only supports anonymous authentication, and this adds on the ability for one-and-done type auth.

That said, I'm not generally happy w/ the structure of BarrageSubcomponent, FlightSubcomponent, SessionSubcomponent (I know I was the one who made them). We may want to remove some of these layers of abstractions or provide better interfaces for users to hook into.

nbauernfeind (approved):

PR #3434, opened 2023-02-13
Add read_sql in the dbc package

### LINKED ISSUE #3430: add read_sql() to the relational data ingestion API
Context:
1. read_sql() is provided in Pandas as a way to read SQL query into a DataFrame. 
2. ConnectorX mirrored the read_sql() interface but  support other data formats including Arrow and has better performance and memory efficiency than other similar products such as Pandas, Turbodbc etc. 
3. In contrast to Turbodbc, ConnectorX provides binary wheels for pip installation and doesn't require additional ODBC setup, which should make the user experience much smoother. 
4. In comparison with ADBC, ConnectorX supports many more data sources with ODBC still a WIP.

jmao-denver:
Fixes #3430

1. add the read_sql() function. ***Note that it is a package-level function so the handling of the optional dependency on ConnectorX is different from the adbc/odbc modules***
2. clean up existing code a bit

### DIFF
--- a/py/server/deephaven/dbc/__init__.py
+++ b/py/server/deephaven/dbc/__init__.py
@@ -1,5 +1,35 @@
 #
 #     Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
 #
+"""The dbc package includes the modules and functions for using external databases with Deephaven."""
 
-"""The dbc package includes the modules for using external databases with Deephaven."""
+from deephaven import DHError
+from deephaven.table import Table
+import deephaven.arrow as dharrow
+
+
+def read_sql(conn: str, query: str) -> Table:
+    """Executes the provided SQL query via ConnectorX and returns a Deephaven table.
+
+    Args:
+        conn (str): a connection string URI, please refer to https://sfu-db.github.io/connector-x/databases.html for
+            database specific format
+        query (str): SQL query statement
+
+    Returns:
+        a new Table
+
+    Raises:
+        DHError
+    """
+    try:
+        import connectorx as cx
+    except ImportError:
+        raise DHError(message="import ConnectorX failed, please install it first.")
+
+    try:
+        pa_table = cx.read_sql(conn=conn, query=query, return_type="arrow")
+    except Exception as e:
+        raise DHError(e, message="failed to get a Arrow table from ConnectorX.") from e
+
+    return dharrow.to_table(pa_table)
--- a/py/server/tests/test_dbc.py
+++ b/py/server/tests/test_dbc.py
@@ -5,7 +5,8 @@ import unittest
 
 import turbodbc
 
-from deephaven.dbc import odbc as dhodbc, adbc as dhadbc
+from deephaven import DHError
+from deephaven.dbc import odbc as dhodbc, adbc as dhadbc, read_sql
 from tests.testbase import BaseTestCase
 
 
@@ -31,6 +32,13 @@ class DbcTestCase(BaseTestCase):
                 # This is not ideal but ADBC might have a bug regarding cursor.rowcount it currently returns -1
                 # instead of the actual size
                 self.assertEqual(table.size, 10)
+
+    def test_read_sql(self):
+        postgres_url = "postgresql://test:test@postgres:5432/test"
+        query = "SELECT t_ts, t_id, t_instrument, t_exchange, t_price, t_size FROM CRYPTO_TRADES LIMIT 10"
+        dh_table = read_sql(conn=postgres_url, query=query)
+        self.assertEqual(len(dh_table.columns), 6)
+        self.assertEqual(dh_table.size, 10)
 
 
 if __name__ == '__main__':
--- a/py/server/deephaven/dbc/adbc.py
+++ b/py/server/deephaven/dbc/adbc.py
@@ -9,11 +9,9 @@ ADBC defines a standard API to fetch data in Arrow format from databases that su
 from databases that only support ODBC/JDBC. By relying on ADBC, Deephaven is able to ingest data efficiently from a
 wide variety of data sources. """
 
-from typing import Any
-
-from deephaven.table import Table
-from deephaven import arrow as dharrow
 from deephaven import DHError
+from deephaven import arrow as dharrow
+from deephaven.table import Table
 
 try:
     import adbc_driver_manager.dbapi
@@ -26,8 +24,8 @@ def read_cursor(cursor: adbc_driver_manager.dbapi.Cursor) -> Table:
     """Converts the result set of the provided cursor into a Deephaven table.
 
     Args:
-        cursor (Any): an ADBC DB-API cursor. Prior to it being passed in, its execute() method must be called to
-            run a query operation that produces an Arrow table
+        cursor (adbc_driver_manager.dbapi.Cursor): an ADBC DB-API cursor. Prior to it being passed in, its execute()
+            method must be called to run a query operation that produces an Arrow table
 
     Returns:
         a new Table
--- a/py/server/deephaven/dbc/odbc.py
+++ b/py/server/deephaven/dbc/odbc.py
@@ -8,8 +8,6 @@ Turbodbc is DB-API 2.0 compliant, provides access to relational databases via th
 importantly it has optimized, built-in Apache Arrow support when fetching ODBC result sets. This enables Deephaven to
 achieve maximum efficiency when ingesting relational data. """
 
-from typing import Any
-
 from deephaven import DHError
 from deephaven import arrow as dharrow
 from deephaven.table import Table
@@ -25,8 +23,8 @@ def read_cursor(cursor: turbodbc.cursor.Cursor) -> Table:
     """Converts the result set of the provided cursor into a Deephaven table.
 
     Args:
-        cursor (Any): a Turbodbc cursor. Prior to it being passed in, its execute() method must be called to run a query
-            operation that produces a result set
+        cursor (turbodbc.cursor.Cursor): a Turbodbc cursor. Prior to it being passed in, its execute() method must be
+            called to run a query operation that produces a result set
 
     Returns:
         a new Table


Integrations/build.gradle
```diff
@@ -67,7 +67,7 @@ def runInDocker = { String name, String sourcePath, List<String> command, Closur
             // set up the container, env vars - things that aren't likely to change
             from 'deephaven/server-netty:local-build'
             runCommand '''set -eux; \\
-                      pip3 install unittest-xml-reporting==3.0.4; \\
+                      pip3 install unittest-xml-reporting==3.0.4 connectorx; \\
```
chipkent, line 70: connectorx should at a bare minimum be listed as an optional dependency.  I don't see that happening anywhere in the PR.
https://setuptools.pypa.io/en/latest/userguide/dependency_management.html

py/server/deephaven/dbc/__init__.py
```diff
@@ -1,5 +1,35 @@
...
+    Returns:
+        a new Table
+
+    Raises:
+        DHError
+    """
+    try:
+        import connectorx as cx
+    except ImportError:
+        raise DHError(message="import ConnectorX failed, please install it first.")
```
chipkent, line 28: Should this give the pip install command?

py/server/deephaven/dbc/__init__.py
```diff
@@ -1,5 +1,35 @@
...
+
+    try:
+        pa_table = cx.read_sql(conn=conn, query=query, return_type="arrow")
+    except Exception as e:
+        raise DHError(e, message="failed to get a Arrow table from ConnectorX.") from e
```
chipkent, line 33:
```suggestion
        raise DHError(e, message="failed to get an Arrow table from ConnectorX.") from e
```

py/server/deephaven/dbc/__init__.py
```diff
@@ -1,5 +1,35 @@
...
+from deephaven.table import Table
+import deephaven.arrow as dharrow
+
+
+def read_sql(conn: str, query: str) -> Table:
+    """Executes the provided SQL query via ConnectorX and returns a Deephaven table.
+
+    Args:
+        conn (str): a connection string URI, please refer to https://sfu-db.github.io/connector-x/databases.html for
+            database specific format
```
chipkent, line 16: This method is written in a very connectorx centric way, but it is a very flexible API.  For example, it could easily be made to work with ADBC and turbodbc.  I can see adding a third argument that indicates the connection type (and defaults to connectorx).  Then users have one powerful API that works with all of our connectors.  Is there a reason not to do this?

chipkent (changes requested):

Integrations/build.gradle
jmao-denver: I believe this is one of the packages (including turbodbc, adbc) we deemed as 'need to be installed by the user' in his/her Python environment. We could add it to the optional required packages for deephaven-core the same as we did for 'jedi', but then we need to include turbodbc, adbc as well. The pip command syntax for installing addition features of a package isn't the most friendly to remember. I think we might be better off either making it a default requirement or leaving it out entirely. @devinrsmith , your thoughts too?

py/server/deephaven/dbc/__init__.py
jmao-denver: Probably not, I think it is safe to assume a Python user knows?

Integrations/build.gradle
devinrsmith: Our "Integration" testing image includes packages we don't explicitly mark as optional - see https://github.com/deephaven/deephaven-server-docker/blob/main/contexts/server-base/requirements.txt.

(It probably makes sense to move `unittest-xml-reporting` to there as well.)

That said, we _could_ add optional features in our setup.py - I'm not sure the best way to organize them, if dbc is separate from adbc is separate from odbc?

An alternative that I don't think we've considered is to have "proper" additional python packages:

`deephaven-core-odbc`, `deephaven-core-autocomplete`, etc. In this way, we _could_ create optional packages that have concrete requirements.

docker/registry/protoc-base/gradle.properties
```diff
@@ -1,5 +1,3 @@
 io.deephaven.project.ProjectType=DOCKER_REGISTRY
 deephaven.registry.imageName=ghcr.io/deephaven/protoc-base:latest
-deephaven.registry.imageId=ghcr.io/deephaven/protoc-base@sha256:dcbde68539a04be744963d1f3c90c30be1c7f75d291635d2f1efc3ee0c0167ed
-# TODO(deephaven-base-images#55): arm64 native image for protoc-base
-deephaven.registry.platform=linux/amd64
```
devinrsmith, line 5: Sorry, this hasn't been incorporated into the update script. Please make sure this doesn't get removed.

docker/registry/cpp-client-base/gradle.properties
```diff
@@ -1,5 +1,3 @@
 io.deephaven.project.ProjectType=DOCKER_REGISTRY
 deephaven.registry.imageName=ghcr.io/deephaven/cpp-client-base:latest
-deephaven.registry.imageId=ghcr.io/deephaven/cpp-client-base@sha256:5b6e1749437170c0b0a3ebe5755c38fdbe3645b7c409cf8a72bc6a6acf50eebe
-# TODO(deephaven-base-images#54): arm64 native image for cpp-client-base
-deephaven.registry.platform=linux/amd64
```
devinrsmith, line 5: Sorry, this hasn't been incorporated into the update script. Please make sure this doesn't get removed.

py/server/tests/test_dbc.py
```diff
@@ -32,6 +34,14 @@ def test_read_adbc(self):
                 # instead of the actual size
                 self.assertEqual(table.size, 10)
 
+    @unittest.skipIf(platform.machine() != "x86_64", reason="connectorx not available on Linux/Arm64")
```
devinrsmith, line 37: Can you add a link to the connectorx issue?

docker/server-jetty/src/main/server-jetty/requirements.txt
```diff
@@ -1,5 +1,6 @@
-adbc-driver-manager==0.1.0
-adbc-driver-postgresql==0.1.0
+adbc-driver-manager==0.2.0
+adbc-driver-postgresql==0.2.0
+connectorx==0.3.1
```
devinrsmith, line 3: I think adding this requirement will cause M1 integration tests to fail. I think we can hack in the same platform check that we've added for the base images...

devinrsmith (changes requested):

py/server/deephaven/dbc/__init__.py
chipkent: The message is not clear that the package is `connectorx`.  If they try `ConnectorX`, the install will fail.  As it is, this message will probably require a google, which is not ideal.

py/server/deephaven/dbc/__init__.py
chipkent: It really isn't much additional complexity.

py/server/tests/test_dbc.py
```diff
@@ -32,6 +34,15 @@ def test_read_adbc(self):
                 # instead of the actual size
                 self.assertEqual(table.size, 10)
 
+    @unittest.skipIf(platform.machine() != "x86_64", reason="connectorx not available on Linux/Arm64"
+                                                            "https://github.com/sfu-db/connector-x/issues/386")
```
devinrsmith, line 38: How does this work? Is this ticket part of 'reason'? Not sure I've seen this pattern used before.

devinrsmith (approved):

py/server/tests/test_dbc.py
jmao-denver: yes, that's just concatenation of two literal strings.

py/server/deephaven/dbc/__init__.py
jmao-denver: I am in support of a simply interface for ODBC/ADBC but the flag pattern isn't so nice. How about read_odbc and read_adbc with the same signature as read_sql?  This will still enable us to check for the requirement of installation of turbodbc and adbc.

py/server/deephaven/dbc/__init__.py
```diff
@@ -2,34 +2,85 @@
...
+                raise DHError(message="import connectorx failed, please install it first.")
+
+            try:
+                pa_table = cx.read_sql(conn=conn, query=query, return_type="arrow")
+                return dharrow.to_table(pa_table)
+            except Exception as e:
+                raise DHError(e, message="failed to get a Arrow table from ConnectorX.") from e
+        elif driver == "odbc":
+            from deephaven.dbc.odbc import read_cursor
+            import turbodbc
```
chipkent, line 40: This needs to be in an try-except block.

chipkent: .... actually, the import above may cover this case.

py/server/deephaven/dbc/__init__.py
```diff
@@ -2,34 +2,85 @@
...
+            with turbodbc.connect(connection_string=conn) as conn:
+                with conn.cursor() as cursor:
+                    cursor.execute(query)
+                    return read_cursor(cursor)
+        elif driver == "adbc":
+            from deephaven.dbc.adbc import read_cursor
+            if not conn:
+                import adbc_driver_sqlite.dbapi as dbapi
+            elif conn.strip().startswith("postgresql"):
```
chipkent, line 49:
```suggestion
            elif conn.strip().startswith("postgresql:"):
```
Should we include the `:` in case drivers start with the same prefix to a point?

py/server/deephaven/dbc/__init__.py
```diff
@@ -2,34 +2,85 @@
...
+                import adbc_driver_postgresql.dbapi as dbapi
+            else:
+                raise DHError(message="not supported ADBC connection string")
```
chipkent, line 52:
```suggestion
                raise DHError(message="unsupported ADBC connection string")
```
Should this print the string?

py/server/deephaven/dbc/__init__.py
```diff
@@ -2,34 +2,85 @@
...
+            import turbodbc.connection
+            if isinstance(conn, turbodbc.connection.Connection):
+                from deephaven.dbc.odbc import read_cursor
+                with conn.cursor() as cursor:
+                    cursor.execute(query)
+                    return read_cursor(cursor)
+        except ImportError:
+            pass
+
+        raise DHError(message="invalid conn argument")
```
chipkent, line 81:
```suggestion
        raise DHError(message="unsupported conn argument")
```
Should this print out the type of the conn arg?

py/server/deephaven/dbc/__init__.py
```diff
@@ -2,34 +2,85 @@
...
+            else:
+                raise DHError(message="not supported ADBC connection string")
+
+            with dbapi.connect(conn) as dbconn:
+                with dbconn.cursor() as cursor:
+                    cursor.execute(query)
+                    return read_cursor(cursor)
+        else:
+            raise DHError(message=f"unsupported driver {driver}")
+    else:
```
chipkent, line 60: I'm conflicted on this section.  I love that it "just works".  On the other hand, we are ignoring the driver value.  Should we be using the driver value here?  Should we just document that we ignore the driver value if conn is not a string?

py/server/tests/test_dbc.py
```diff
@@ -34,14 +34,57 @@ def test_read_adbc(self):
                 # instead of the actual size
                 self.assertEqual(table.size, 10)
 
-    @unittest.skipIf(platform.machine() != "x86_64", reason="connectorx not available on Linux/Arm64")
-    def test_read_sql(self):
-        postgres_url = "postgresql://test:test@postgres:5432/test"
+    @unittest.skipIf(platform.machine() != "x86_64", reason="connectorx not available on Linux/Arm64"
+                                                            "https://github.com/sfu-db/connector-x/issues/386")
+    def test_read_sql_connectorx(self):
```
chipkent, line 39: Should this be a separate test or a subtest under `test_read_sql`?  It seems like it is a subtest.

docker/server-jetty/src/main/server-jetty/requirements.txt
```diff
@@ -1,5 +1,6 @@
-adbc-driver-manager==0.1.0
-adbc-driver-postgresql==0.1.0
+adbc-driver-manager==0.2.0
+adbc-driver-postgresql==0.2.0
+connectorx==0.3.1; platform.machine == 'x86_64'
```
chipkent, line 3: @devinrsmith should confirm this is the right syntax.

docker/server/src/main/server-netty/requirements.txt
```diff
@@ -1,5 +1,6 @@
-adbc-driver-manager==0.1.0
-adbc-driver-postgresql==0.1.0
+adbc-driver-manager==0.2.0
+adbc-driver-postgresql==0.2.0
+connectorx==0.3.1; platform.machine == 'x86_64'
```
chipkent, line 3: @devinrsmith should confirm this is the right syntax.

chipkent (changes requested):

py/server/deephaven/dbc/__init__.py
jmao-denver: yes, that's intentional

py/server/deephaven/dbc/__init__.py
jmao-denver: yes, it is safer

py/server/tests/test_dbc.py
jmao-denver: It it made a separate test because we need to use ```@unittest.skipIf```

docker/server-jetty/src/main/server-jetty/requirements.txt
jmao-denver: He already did and approved images related changes

devinrsmith (dismissed):

chipkent (approved):

devinrsmith (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2235
Conceptual: https://github.com/deephaven/deephaven.io/issues/2237
Reference: https://github.com/deephaven/deephaven.io/issues/2236
Blog: https://github.com/deephaven/deephaven.io/issues/2238

PR #3435, opened 2023-02-14
move API docs from md file to java classes

vegegoku:


web/client-api/src/main/java/io/deephaven/web/client/api/Sort.java
```diff
@@ -7,7 +7,13 @@
...
 
+/**
+ * Describes a Sort present on the table. No visible constructor, created through the use of Column.sort(), will be tied
+ * to that particular column data. Sort instances are immutable, and use a builder pattern to make modifications. All
+ * methods return a new Sort instance.
+ */
 public class Sort {
+
     @JsProperty(namespace = "dh.Sort")
     public static final String ASCENDING = "ASC",
```
niloc132, line 18: note to future me, make this into a tstypedef since only three strings are legal anyway...

web/client-api/src/main/java/io/deephaven/web/client/api/i18n/JsTimeZone.java
```diff
@@ -12,6 +12,28 @@
...
+ * throughout the year (potentially changing each year). These instances mostly are useful at this time to pass to the
+ * `DateTimeFormat.format()` methods, though also support a few properties at this time to see details about each
+ * instance.
+ *
+ *
+ * The following timezone codes are supported when getting a timezone object - instances appearing in the same line will
+ * return the same details:
+ *
+ *
+ * - `GMT`/`UTC` - `Asia/Tokyo` - `Asia/Seoul` - `Asia/Hong_Kong` - `Asia/Singapore` - `Asia/Calcutta`/`Asia/Kolkata` -
```
niloc132, line 26: Probably should explicitly break this into ul/li items, since the spotless pass is murdering it

web/client-api/src/main/java/io/deephaven/web/client/api/filter/FilterCondition.java
```diff
@@ -13,12 +13,32 @@
...
 @JsType(namespace = "dh")
 public class FilterCondition {
 
     @JsIgnore
     private final Condition descriptor;
 
+    /**
+     * @param function
+     * @param args
+     * @return a filter condition invoking a static function with the given parameters. Currently supported Deephaven
```
niloc132, line 30: will need to be reworked into something spotless doesnt mangle

niloc132 (commented):
Nice work. I'm going to send you a PR on this branch for a few small things to clean up, but here's the main idea of follow-up I'd like to see before we merge this:
 * internal only remarks should be removed, so we just see the docs intended for JS authors. They can be separated into their own `/* */` comment blocks if we think they are important to keep.
 * We should provide more "enums" for special constants, rather than just listing them all
 * Take a (gradual) pass and populate `@param`s and other empty properties. I'm not sure if we want to stay with the `@return` property usually having the body for simple methods/properties or not...
 * I'm unsure if we want to have `Read-only` any more, since TS should pick that up automatically, and at least in theory any TS doc tool will add that for us. We'll need to test before we go ahead and remove those. Ditto "optional parameter..."
 * Some markdown needs to be reworked into HTML, so that spotless doesnt rewrite it into nonsense.

web/client-api/src/main/java/io/deephaven/web/client/api/Column.java
```diff
@@ -78,6 +102,10 @@ public Format getFormat(TableData.Row row) {
         return row.getFormat(this);
     }
 
+    /**
+     * @deprecated do not use. Internal index of the column in the table, to be used as a key on the Row.
```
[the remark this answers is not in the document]
mattrunyon: Colin said there's no alternative, so nevermind on this

mattrunyon (approved):

PR #3450, opened 2023-02-20
Fix automcomplete suggestions overwriting existing text

### LINKED ISSUE #3363: Autocomplete suggestions overwrite existing text instead of inserting
**Description**

When using Python autocomplete, the suggestions overwrite existing text rather than insert it.

**Steps to reproduce**
1. Enter the following snippet of code in the console, but do not submit the command:
```
from deephaven import emtpy_table
t = empty_table(10).update(["x=i"])
```
2. Put the cursor after `.update` and change it to `.update_`, then select `.update_view` from the suggestion menu (see video).

**Expected results**
2. The snippet should update to `.update_view`, i.e.
```
from deephaven import emtpy_table
t = empty_table(10).update_view(["x=i"])
```

**Actual results**
2. The update_view overwrites part of the formula, and the resulting text is:
```
from deephaven import emtpy_table
t = empty_table(10).update_view=i"])
```

autocomplete-overwrite

**Versions**
Engine Version: 0.21.0
Web UI Version: 0.26.1
Java Version: 17.0.5
Barrage Version: 0.5.0

mattrunyon:
Fixes #3363 

`TextEdit.range` should be the range of characters to be replaced, not the range of the inserted characters. For completions that should just be the original range

We should maybe change our returned `TextEdit` to use `newText` instead of `text` to match the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textEdit). That would require a 1 line change in the web UI, and it probably needs to support both options since the monaco component is re-used in enterprise

### DIFF
--- a/server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
+++ b/server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
@@ -37,7 +37,7 @@ public class PythonAutoCompleteObserver extends SessionCloseableObserver<AutoCom
     private final Provider<ScriptSession> scriptSession;
 
     public PythonAutoCompleteObserver(StreamObserver<AutoCompleteResponse> responseObserver,
-                                      Provider<ScriptSession> scriptSession, final SessionState session) {
+            Provider<ScriptSession> scriptSession, final SessionState session) {
         super(session, responseObserver);
         this.scriptSession = scriptSession;
     }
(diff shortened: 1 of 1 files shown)


niloc132 (approved):

devinrsmith (approved):
Looks good. Do you have a follow-up ticket for the `newText` issue you referenced?

mattrunyon:
https://github.com/deephaven/deephaven-core/issues/3452

PR #3458, opened 2023-02-22
Pass arrays from vector type columns unpacked to Python func

### LINKED ISSUE #2577: Query language fails to call python methods properly with grouped columns
The query language is not properly calling python functions when columns are grouped.  The example below illustrates the problem.  Array values seem to be expanded and passed in as separate arguments, instead of being passed in as a single array argument.

```python

t1 = time_table("00:00:01").update(["TimeBin = upperBin(Timestamp, '00:00:03')", "Person = random() > 0.5 ? `A` : random() > 0.5 ? `B` : `C`"])

t2 = t1 \
    .drop_columns("Timestamp") \
    .select_distinct(["TimeBin", "Person"]) \
    .sort(["TimeBin", "Person"]) \
    .group_by("TimeBin")

# single positional argument
def make_pairs_1(a):
    return 1

# take any inputs as an array
def make_pairs_2(*a):
    return 2

# works
x11 = t1.update("Pair=(int)make_pairs_1(Person)")
# works
x12 = t1.update("Pair=(int)make_pairs_2(Person)")
# fails
x21 = t2.update("Pair=(int)make_pairs_1(Person)")
# works
x22 = t2.update("Pair=(int)make_pairs_2(Person)")
```

jmao-denver:
Fixes #2577

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
@@ -958,9 +958,11 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
                         expressions[ei] = new CastExpr(
                                 new ClassOrInterfaceType("java.lang.Object"),
                                 expressions[ei]);
+                        expressionTypes[ei] = Object.class;
+                    } else {
+                        expressionTypes[ei] = convertVector(expressionTypes[ei],
+                                parameterizedTypes[ei] == null ? null : parameterizedTypes[ei][0]);
                     }
-                    expressionTypes[ei] = convertVector(expressionTypes[ei],
-                            parameterizedTypes[ei] == null ? null : parameterizedTypes[ei][0]);
                 }
             }
         }
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
@@ -944,6 +944,27 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
     private Expression[] convertParameters(final Executable executable,
             final Class<?>[] argumentTypes, final Class<?>[] expressionTypes,
             final Class<?>[][] parameterizedTypes, Expression[] expressions) {
+
+        // For Python callables, all Vector columns should be converted into arrays
+        if (executable.getDeclaringClass() == PyCallableWrapper.class) {
+            for (int ei = 0; ei < expressionTypes.length; ei++) {
+                if (isTypedVector(expressionTypes[ei])) {
+                    expressions[ei] =
+                            new MethodCallExpr(new NameExpr("VectorConversions"), "nullSafeVectorToArray",
+                                    new NodeList<>(expressions[ei]));
+                    // To prevent JPY from unpacking a non-primitive array to make sure the result array is treated
+                    // as a single argument
+                    if (expressionTypes[ei] == ObjectVector.class) {
+                        expressions[ei] = new CastExpr(
+                                new ClassOrInterfaceType("java.lang.Object"),
+                                expressions[ei]);
+                    }
+                    expressionTypes[ei] = convertVector(expressionTypes[ei],
+                            parameterizedTypes[ei] == null ? null : parameterizedTypes[ei][0]);
+                }
+            }
+        }
+
         final int nArgs = argumentTypes.length; // Number of declared arguments
         for (int ai = 0; ai < (executable.isVarArgs() ? nArgs - 1 : nArgs); ai++) {
             if (argumentTypes[ai] != expressionTypes[ai] && argumentTypes[ai].isPrimitive()
--- a/py/server/tests/test_table.py
+++ b/py/server/tests/test_table.py
@@ -873,6 +873,35 @@ class TableTestCase(BaseTestCase):
         self.assertEqual(attrs, rt_attrs)
         self.assertTrue(rt.j_table is not self.test_table.j_table)
 
+    def test_grouped_column_as_arg(self):
+        t1 = empty_table(100).update(
+            ["id = i % 10", "Person = random() > 0.5 ? true : random() > 0.5 ? false : true"]).sort(
+            ["id", "Person"]).group_by("id")
+        t2 = empty_table(100).update(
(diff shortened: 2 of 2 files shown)


py/server/tests/test_table.py
```diff
@@ -859,6 +859,33 @@ def test_tree(self):
         tree_table = self.test_table.tail(10).tree(id_col='a', parent_col='c', promote_orphans=True)
         self.assertIsNotNone(tree_table)
 
+    def test_grouped_column_as_arg(self):
```
chipkent, line 862: I would add another set of test tables using java arrays instead of DB arrays.  e.g. `"Person = new Object[]{...}"` and `"Person = new String[]{...}"` etc.

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -944,6 +944,27 @@ public static boolean isTypedVector(Class<?> type) {
     private Expression[] convertParameters(final Executable executable,
             final Class<?>[] argumentTypes, final Class<?>[] expressionTypes,
             final Class<?>[][] parameterizedTypes, Expression[] expressions) {
+
+        // For Python callables, all Vector columns should be converted into arrays
```
chipkent, line 948: I have a few questions / comments after reading this:
1) This looks like it converts all vectors to arrays, right?
2) I can see doing the array conversion as being convenient, but I can also see it as being low performance.  Do we understand the tradeoffs of always doing it vs having a method in python that gets called?
3) This looks like nodes are being added to the query call expressions.  Will this change effectively break vectorization for python until we do extensive work to make the rest of vectorization work?
4) Have any tests been run to quantify (3)?

chipkent (changes requested):

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
jmao-denver: 1. yes
2. it is actually more performant because JPY knows how to handle Java arrays not vectors. So converting vectors to arrays first should reduce the boundary-crossing when the Python code iterates over the data.
3. Python function calls with array/vector argument can't be vectorized.

docker-compose-common.yml
```diff
@@ -14,26 +14,31 @@ services:
       # with max memory.
       #
       # To turn on debug logging, add: -Dlogback.configurationFile=logback-debug.xml
-      - START_OPTS=-Xmx4g -Ddeephaven.console.type=${DEEPHAVEN_CONSOLE_TYPE} -Ddeephaven.application.dir=${DEEPHAVEN_APPLICATION_DIR}
+      # - JAVA_TOOL_OPTIONS=-Xmx4g -Ddeephaven.console.type=${DEEPHAVEN_CONSOLE_TYPE} -Ddeephaven.application.dir=${DEEPHAVEN_APPLICATION_DIR}
```
devinrsmith, line 17: I don't think we want any changes to this file?

docker/server-jetty/src/main/docker/Dockerfile
```diff
@@ -28,13 +28,26 @@ FROM $BASE
 COPY --from=install / /
 VOLUME /data
 VOLUME /cache
-HEALTHCHECK --interval=3s --retries=3 --timeout=11s CMD /opt/grpc_health_probe/grpc_health_probe -addr=localhost:10000 -connect-timeout=10s || exit 1
+HEALTHCHECK --interval=3s --retries=3 --timeout=11s CMD /bin/grpc_health_probe -addr=localhost:10000 -connect-timeout=10s || exit 1
```
devinrsmith, line 31: I'm confused why this change was made...

docker-compose-common.yml
chipkent: agree

docker/server-jetty/src/main/docker/Dockerfile
chipkent: agree

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
chipkent: Does the python code see the input as a numpy array or just a buffer that needs to be converted to a numpy array?

docker-compose-common.yml
jmao-denver: sorry, dumb mistake

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
jmao-denver: The Python code sees it as a JPY wrapped Java array not a numpy array. If the array is a primitive one, its buffer can be used to create a numpy array with zero-copy.

chipkent: This seems very usable without always converting to numpy.

image

My only concern is the lack of vectorization, but I think that is a separate effort.

chipkent (dismissed):

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -944,6 +944,27 @@ public static boolean isTypedVector(Class<?> type) {
...
+                                    new NodeList<>(expressions[ei]));
+                    // To prevent JPY from unpacking a non-primitive array to make sure the result array is treated
+                    // as a single argument
+                    if (expressionTypes[ei] == ObjectVector.class) {
+                        expressions[ei] = new CastExpr(
+                                new ClassOrInterfaceType("java.lang.Object"),
+                                expressions[ei]);
+                    }
+                    expressionTypes[ei] = convertVector(expressionTypes[ei],
+                            parameterizedTypes[ei] == null ? null : parameterizedTypes[ei][0]);
```
rcaudy, line 963: I'm nervous that in the case of `ObjectVector` we are creating an expression that explicitly conflicts with the expression type. Maybe we need the code to be this:
```suggestion
                    if (expressionTypes[ei] == ObjectVector.class) {
                        expressions[ei] = new CastExpr(
                                new ClassOrInterfaceType("java.lang.Object"),
                                expressions[ei]);
                        expressionTypes[ei] = Object.class;
                    } else {
                        expressionTypes[ei] = convertVector(expressionTypes[ei],
                                parameterizedTypes[ei] == null ? null : parameterizedTypes[ei][0]);
                    }
```

PR #3460, opened 2023-02-23
Provide a session listener callback interface

### LINKED ISSUE #3457: Allow server initialization code to provide a listener to track the number of clients connected.
For implementing
https://deephaven.atlassian.net/browse/DH-14196
DnD needs the ability to register a callback on server initialization, to listen for client session creation and expiration (eg, client "connect" and "disconnect").  The listener needs be ability to ask for the count of remaining "connected" clients.
The plan on the DnD side is to use the listener to register a timer action when the count goes from > 0 to zero.  The action will shutdown the server after a timeout if no other client connects.  It will also cancel the action if the count goes from 0 to > 0 before the timeout.

niloc132:
Fixes #3457

### DIFF
--- a/server/src/main/java/io/deephaven/server/session/SessionService.java
+++ b/server/src/main/java/io/deephaven/server/session/SessionService.java
@@ -170,8 +170,7 @@ public class SessionService {
      * @return the most recent token expiration
      */
     public TokenExpiration refreshToken(final SessionState session) {
-        TokenExpiration tokenExpiration = checkTokenAndRotate(session, false);
-        return tokenExpiration;
+        return checkTokenAndRotate(session, false);
     }
 
     @SuppressWarnings("SynchronizationOnLocalVariableOrMethodParameter")
@@ -360,11 +359,8 @@ public class SessionService {
                 // token expiration time.
                 outstandingCookies.poll();
 
-                synchronized (next.session) {
-                    final TokenExpiration tokenExpiration = next.session.getExpiration();
-                    if (tokenExpiration != null && tokenExpiration.deadlineMillis <= nowMillis) {
-                        next.session.onExpired();
-                    }
+                if (next.session.isExpired()) {
+                    next.session.onExpired();
                 }
             } while (true);
 
--- a/server/src/main/java/io/deephaven/server/session/SessionService.java
+++ b/server/src/main/java/io/deephaven/server/session/SessionService.java
@@ -29,6 +29,7 @@ import java.util.Deque;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
@@ -62,10 +63,13 @@ public class SessionService {
 
     private final Map<String, AuthenticationRequestHandler> authRequestHandlers;
 
+    private final SessionListener sessionListener;
+
     @Inject
     public SessionService(final Scheduler scheduler, final SessionState.Factory sessionFactory,
             @Named("session.tokenExpireMs") final long tokenExpireMs,
-            Map<String, AuthenticationRequestHandler> authRequestHandlers) {
+            Map<String, AuthenticationRequestHandler> authRequestHandlers,
+            Set<SessionListener> sessionListeners) {
         this.scheduler = scheduler;
         this.sessionFactory = sessionFactory;
         this.tokenExpireMs = tokenExpireMs;
@@ -87,6 +91,8 @@ public class SessionService {
         if (ProcessEnvironment.tryGet() != null) {
             ProcessEnvironment.getGlobalFatalErrorReporter().addInterceptor(this::onFatalError);
         }
+
+        this.sessionListener = new DelegatingSessionListener(sessionListeners);
     }
 
     private synchronized void onFatalError(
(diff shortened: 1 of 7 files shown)


server/src/main/java/io/deephaven/server/session/SessionService.java
```diff
@@ -355,6 +366,7 @@ public void run() {
                 synchronized (next.session) {
                     final TokenExpiration tokenExpiration = next.session.getExpiration();
                     if (tokenExpiration != null && tokenExpiration.deadlineMillis <= nowMillis) {
+                        sessionListener.onSessionEnd(next.session);
```
jcferretti, line 369: "Never fire a user callback that you don't have any control over while holding a lock"
Suggest you set a boolean in the synchronized block and fire the callback outside of it.

jcferretti: Also we talked over slack and you mentioned catching any exceptions the callback may throw, which makes a lot of sense and I didn't think about (writing it down here so that we don't forget).

server/src/main/java/io/deephaven/server/session/SessionService.java
```diff
@@ -163,11 +170,13 @@ public SessionState newSession(final AuthContext authContext) {
      * @return the most recent token expiration
      */
     public TokenExpiration refreshToken(final SessionState session) {
-        return refreshToken(session, false);
+        TokenExpiration tokenExpiration = checkTokenAndRotate(session, false);
+        sessionListener.onSessionRefresh(session);
```
devinrsmith, line 174: Do we still want to invoke "onSessionRefresh" when tokenExpiration is null (ie, already expired)?

devinrsmith: Also, what about cases where the token wasn't actually refreshed?

server/src/main/java/io/deephaven/server/session/SessionService.java
devinrsmith: There may be ordering-guarantees we should provide, and that may or may not mean we need to execute this code with the current lock (I'm not sure without digging in more).

Essentially, I think on a per-SessionState basis, we should offer some sort of "linearity". onSessionCreate happens before onSessiotRefresh(*) happens before onSessionEnd (if we don't already). There may be concurrency concerns, depending on how SessionService methods are invoked.

devinrsmith (changes requested):
Failing tests.

I am interested in what the enterprise implementation will look like; and the guarantees we expect to provide wrt how the listener is invoked.

server/src/main/java/io/deephaven/server/session/SessionService.java
niloc132: Yeah, given the time constraints on backporting this, I'm going to drop the token refresh feature entirely.

It turns out we already have SessionState "on close callbacks" (which implement Closable). Those callbacks are guaranteed to only run once already, and run from outside the onCloseCallbacks lock, but still within the session lock here.  This relaxes the constraint of the onSessionEnd triggering _before_ the session expire happens, so exports are gone and other callbacks may have been invoked.

Finally, onExpired, which will trigger the close callbacks, is called from three places, and only one is under a lock on the SessionState itself. If we need to be locked for that, the other two should be fixed too, but I think instead we'd prefer to move the onExported call itself out, as @jcferretti proposed above.

devinrsmith (approved):
Great - I think this a good simplification, assuming `io.deephaven.server.session.SessionState#addOnCloseCallback` works as expected, and meets DHE needs.

server/src/main/java/io/deephaven/server/session/SessionService.java
nbauernfeind: I agree; whatever we're trying to do here should be handled with an on session close callback. It is guaranteed to be called if the user terminates the session, or whenever it is discovered to be expired. It also is not invoked under a lock.

niloc132: > It also is not invoked under a lock.

Turns out that's not quite true, see the method we're commenting on, looks like this in my current set of changes (i.e. unchanged from main):
```java
                synchronized (next.session) {
                    final TokenExpiration tokenExpiration = next.session.getExpiration();
                    if (tokenExpiration != null && tokenExpiration.deadlineMillis <= nowMillis) {
                        next.session.onExpired();
                    }
                }
```
onExpired does take a lock to find out what callbacks should be invoked, then invokes them outside of the lock, but we're still under the lock here.

server/src/main/java/io/deephaven/server/session/SessionService.java
```diff
@@ -163,11 +170,12 @@ public SessionState newSession(final AuthContext authContext) {
      * @return the most recent token expiration
      */
     public TokenExpiration refreshToken(final SessionState session) {
-        return refreshToken(session, false);
+        TokenExpiration tokenExpiration = checkTokenAndRotate(session, false);
+        return tokenExpiration;
```
nbauernfeind, line 174: Any reason not to return the result without placing it in a variable first?

nbauernfeind (approved):

devinrsmith (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2213
Conceptual: https://github.com/deephaven/deephaven.io/issues/2214
Reference: https://github.com/deephaven/deephaven.io/issues/2212
Blog: https://github.com/deephaven/deephaven.io/issues/2215

PR #3474, opened 2023-03-01
Add dropColumns as TableOperations / TableSpec

devinrsmith:
This is in support of #3473, where for adapting implementation purposes, it's cleaner and easier to use dropColumns.

### DIFF
--- a/qst/src/main/java/io/deephaven/qst/table/DropColumnsTable.java
+++ b/qst/src/main/java/io/deephaven/qst/table/DropColumnsTable.java
@@ -29,13 +29,6 @@ public abstract class DropColumnsTable extends TableBase implements SingleParent
         return visitor;
     }
 
-    @Check
-    final void checkNonEmpty() {
-        if (dropColumns().isEmpty()) {
-            throw new IllegalArgumentException("Expected dropColumns() to be non-empty");
-        }
-    }
-
     public interface Builder {
         Builder parent(TableSpec parent);
 
--- a/qst/src/main/java/io/deephaven/qst/table/DropColumnsTable.java
+++ b/qst/src/main/java/io/deephaven/qst/table/DropColumnsTable.java
@@ -36,7 +36,7 @@ public abstract class DropColumnsTable extends TableBase implements SingleParent
         }
     }
 
-    interface Builder {
+    public interface Builder {
         Builder parent(TableSpec parent);
 
         Builder addDropColumns(ColumnName element);
--- a/qst/src/main/java/io/deephaven/qst/table/DropColumnsTable.java
+++ b/qst/src/main/java/io/deephaven/qst/table/DropColumnsTable.java
@@ -1,0 +1,50 @@
+/**
+ * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+ */
+package io.deephaven.qst.table;
+
+import io.deephaven.annotations.NodeStyle;
+import io.deephaven.api.ColumnName;
+import org.immutables.value.Value.Check;
+import org.immutables.value.Value.Default;
+import org.immutables.value.Value.Immutable;
+
+import java.util.List;
+
+@Immutable
+@NodeStyle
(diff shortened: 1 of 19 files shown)


devinrsmith:
For a bit more context, the SQL query:

```python
my_time = time_table("00:00:01").update_view(
    ["I=ii%10==0?null:ii", "B=ii%11==0?null:ii%3==0"]
)
```

```sql
SELECT 
  * 
FROM 
  my_time 
ORDER BY "B", "Timestamp" DESC
```

produces the logical plan

```
LogicalSort(sort0=[$2], sort1=[$0], dir0=[ASC], dir1=[DESC])
  LogicalProject(Timestamp=[$0], I=[$1], B=[$2])
    LogicalTableScan(table=[[my_time]])
```

notice that the hierarchy produced has sort being executed *after* the project, yet still refers to the columns by reference.

With the DHC implementation, I'd like to adapt all of the tables into "reference" space, execute the plan, and then remove the extra references.

In this implementation, instead of using a `view` for the projection, I'm able to use an `update_view` so that the sort can still proceed in reference space, and then finally by dropping the references as the last step.

```
digraph {
"op_0" ["label"="ticketTable(scan/my_time)"]
"op_1" ["label"="view(ref_0=Timestamp,ref_1=I,ref_2=B)"]
"op_2" ["label"="updateView(Timestamp=ref_0,I=ref_1,B=ref_2)"]
"op_3" ["label"="sort([ASCENDING(ref_2),DESCENDING(ref_0)])"]
"op_4" ["label"="dropColumns([ref_0,ref_1,ref_2])"]
"op_1" -> "op_0"
"op_2" -> "op_1"
"op_3" -> "op_2"
"op_4" -> "op_3"
}
```

devinrsmith:
I'd potentially like to follow this up with `rename_columns`.

qst/src/main/java/io/deephaven/qst/table/DropColumnsTable.java
```diff
@@ -0,0 +1,50 @@
...
+        visitor.visit(this);
+        return visitor;
+    }
+
+    @Check
+    final void checkNonEmpty() {
+        if (dropColumns().isEmpty()) {
+            throw new IllegalArgumentException("Expected dropColumns() to be non-empty");
+        }
+    }
```
rcaudy, line 37: This feels controversial.

table-api/src/main/java/io/deephaven/api/TableOperations.java
```diff
@@ -1300,4 +1300,31 @@ TOPS updateBy(UpdateByControl control, Collection<? extends UpdateByOperation> o
...
+     */
+    TOPS dropColumns(Collection<String> columnNames);
+
+    /**
+     * Drop the {@code columnNames} from {@code this} table.
+     *
+     * @param columnNames the columns to drop
+     * @return the table
+     */
+    TOPS dropColumns(ColumnName... columnNames);
```
rcaudy, line 1326: You're missing `@ConcurrentMethod` annotations.
Also, it seems to me like "this" isn't the right way to say what dropColumns does. We return a new table without the dropped columns.

devinrsmith: Will fix. Made a change to docs, happy to take more specific wording.

devinrsmith:
So - my _current_ SQL implementation does not rely on dropColumns anymore. That said, it might be nice to get these improvements over the line regardless?

qst/src/main/java/io/deephaven/qst/table/DropColumnsTable.java
rcaudy: Hmm. Would be reasonable to add these 3 lines at the top of `QueryTable.dropColumns`:
```
        if (columnNames.length == 0) {
            return prepareReturnThis();
        }
```

rcaudy (commented):
Caught up to head, still need to discuss open issue.

rcaudy (approved):

PR #3478, opened 2023-03-02
When redirecting to Web UI, pass query string parameters

### LINKED ISSUE #3463: Preserve query parameters on redirect to IDE
Currently, when I go to http://localhost:10000/?param I get redirected to http://localhost:10000/ide/ and the `param` gets dropped.
I'd like to have the query parameters preserved on redirect.

niloc132:
If any query string is present, this change will include them in the redirect location.

Fixes #3463

### DIFF
--- a/server/jetty/src/main/java/io/deephaven/server/jetty/HomeFilter.java
+++ b/server/jetty/src/main/java/io/deephaven/server/jetty/HomeFilter.java
@@ -8,6 +8,7 @@ import jakarta.servlet.FilterChain;
 import jakarta.servlet.ServletException;
 import jakarta.servlet.ServletRequest;
 import jakarta.servlet.ServletResponse;
+import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


vbabich (approved):

server/jetty/src/main/java/io/deephaven/server/jetty/HomeFilter.java
```diff
@@ -16,8 +17,18 @@ public class HomeFilter implements Filter {
...
-        if (response instanceof HttpServletResponse) {
-            ((HttpServletResponse) response).sendRedirect("/ide/");
+        if (request instanceof HttpServletRequest && response instanceof HttpServletResponse) {
+            HttpServletRequest req = (HttpServletRequest) request;
+            HttpServletResponse resp = (HttpServletResponse) response;
+
+            final String location;
+            String queryString = req.getQueryString();
+            if (queryString != null) {
+                location = "/ide/?" + queryString;
```
devinrsmith, line 27: Assuming any encoding is already accounted for?

niloc132: Correct, getQueryString()'s contract is that it will not do any escaping, so it is already escaped. https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#getQueryString--

devinrsmith (approved):

Issue #3479, opened 2023-03-03
Provide a way in DeephavenApiServer to request the server to stop.


jcferretti:
DeephavenApiServer has a join() method but no stop method.
In DnD we do a weird logic check when we run a script to see if we should call join or not; we only call join when we don't run a script.
It would be cleaner to just ask the server to stop after running a script, and always joining.

In general, providing a join() method but no stop() method is weird.

PR #3480, opened 2023-03-03
Provide a way in DeephavenApiServer to request the server to stop. #3479

jcferretti:
https://illumon.slack.com/archives/C03LXTRL9U3/p1677810466037599

### DIFF
--- a/server/src/main/java/io/deephaven/server/runner/DeephavenApiServer.java
+++ b/server/src/main/java/io/deephaven/server/runner/DeephavenApiServer.java
@@ -39,7 +39,6 @@ import java.util.concurrent.TimeoutException;
  */
 public class DeephavenApiServer {
     private static final Logger log = LoggerFactory.getLogger(DeephavenApiServer.class);
-
     private final GrpcServer server;
     private final UpdateGraphProcessor ugp;
     private final LogInit logInit;
@@ -87,7 +86,6 @@ public class DeephavenApiServer {
     SessionService sessionService() {
         return sessionService;
     }
-
 
     /**
      * Starts the various server components, and returns without blocking. Shutdown is mediated by the ShutdownManager,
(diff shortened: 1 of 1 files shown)


jcferretti:
I am closing this ticket as won't implement.
For reference this is the slack conversation around it:

https://illumon.slack.com/archives/C03LXTRL9U3/p1677858902298489

Quoting Colin from there:
```
So in short, there isn't a reasonable way to do this at this time, and the use case hasn't come up where it would make sense. Will defer to any engine person (ping 
[@ryan](https://illumon.slack.com/team/U211UCJL9)
?), so this is just from my perspective.
The engine is the main problem here, but not the only one. As you noticed, both of our grpc server implementations (netty and jetty) can support this reasonably (the hard part was finding a common API that both can satisfy), and our gRPC services likewise are mostly good at going away when we ask them to.
There are a lot of things in DeephavenApiServer that should happen in a hypothetically stop(). Running briefly down the list by looking at run()...
unwire shutdown tasks, which are statically wired (and so we don't know what other global state lives on :stuck_out_tongue: )
stop the logging that was initialized (otherwise we stand a pretty good chance of breaking if someone calls run() again)
scripts, both app mode and script sessions. esp for python, this is global in our process, until we either a) get subinterpreters working, or b) use pyodide (via wasm) so we can cleanly sandbox, clean it up, run more than one, etc. Groovy has its own, different problems, with already initialized classes that won't initialize again until we start having fun with classloaders
Please note that these "unstoppable things" are distinct from "no really, you have to run this first, and exactly once", which live in MainHelper.init(), and let us poke the JVM itself in a few ways (configs, properties, logging)
If you're not looking for "stop, so i can start", but "stop so i can halt the jvm", unless I'm very much mistaken that should be covered by just going through a normal shutdown, which will trigger Runtime's shutdown hooks, which will then unwind the above as best we can (see calls to ShutdownManager.registerTask), and then leave things in a "well you can't start things again, but at least we did our best to stop. Unless I'm very much mistaken, that should be covered by simply calling System.exit(0), as that is the only way that we call them today (technically there is a maybeInvokeTasks() call, but it isn't called directly).
If you only want to ask the gRPC server to stop, you can ask that to be injected directly, and then call one of its "time to start stopping" or "go away now" methods. Those are already also hooked up to the shutdown manager though if DeephavenApiServer.run() was called, so this is not an extra step you need to do.
If you want to do all of the shutdown stuff directly, but not actually halt the jvm, consider ProcessEnvironment.getGlobalShutdownManager().maybeInvokeTasks(), which is itself explicitly registered as the shutdown hook in Runtime.
This came up briefly around python, and the consensus was that it would be nice, but doesn't really buy us a lot for real use cases yet, so it was explicitly not added.
```

jcferretti:
This was trying to fix
https://github.com/deephaven/deephaven-core/issues/3479

PR #3482, opened 2023-03-03
JS API should only show non-constituent columns in groupedColumns array

### LINKED ISSUE #3477: Rollup rows produces duplicates in groupedColumns array
**Description**

When using the "Rollup rows" feature, the `model.groupedColumns` array contains 2 entries for each group column.
 
**Steps to reproduce**
1. Create a query with at least 2 columns

```python

static_table = empty_table(25).update_view(formulas=[
    "MyString=new String(`a`+i)",
    "MyChar= new Character((char) ((i%26)+97))",
])
```
2. Rollup rows with any number of columns, with constituent columns enabled

**Expected results**

The model.groupedColumns array should contain the columns that were added in the rollup.

**Actual results**

model.groupedColumns contains 2 columns for each column added in the rollup

e.g. Here's the above query with MyChar and MyString added as rollup columns:

image

**Versions**
Engine Version: 0.22.0
Web UI Version: 0.30.1
Java Version: 17.0.6
Barrage Version: 0.5.0

niloc132:
Fixes #3477

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
@@ -358,7 +358,7 @@ public class JsTreeTable extends HasEventHandling {
             if (definition.isVisible()) {
                 columns[columns.length] = column;
             }
-            if (definition.isRollupGroupByColumn()) {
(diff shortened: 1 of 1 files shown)


bmingles (approved):
Verified the extra rows are no longer present in `groupedColumns`. UI behaves as expected on rollup now.

Issue #2474, opened 2022-06-03
Select/Update do not propagate Blink Table attribute


nbauernfeind:
While `Table.view` and `Table.updateView` do propagate the Blink Table attribute, `Table.select` and `Table.update` do not.

Even for a blink table there could be some performance improvements by using select/update over their lazy counterparts.

Is there a counter argument why they should not?

rcaudy:
Right now, there's no attribute copying for select or update at all: 

`io.deephaven.engine.table.impl.MemoizedOperationKey.SelectUpdateViewOrUpdateView#copyType`:
```
                case Select:
                case Update:
                    // turns out select doesn't copy attributes, maybe we should more accurately codify that
                    return BaseTable.CopyAttributeOperation.None;
```

I have no philosophical objection to changing this.

PR #3484, opened 2023-03-03
Select/Update should copy table attributes

### LINKED ISSUE #3476: `QueryTable#select` drops stream table property
A community user discovered that `select` is dropping the stream property. 

```python
from deephaven.stream.kafka import consumer as kc
from deephaven.stream import stream_to_append_only
from deephaven.time import to_period, minus_period, now, millis_to_datetime, is_after

kafka_cfg = {
    'bootstrap.servers': '<some-boot-strap-list>',
    'schema.registry.url': '<some-registry-url>'
}

stream = kc.consume(
    kafka_cfg,
    'topic',
    key_spec=kc.KeyValueSpec.IGNORE,
    value_spec=kc.avro_spec('topic-value', schema_version='1'),
    table_type=kc.TableType.stream()
).where(filters=["(boolean)is_after(millis_to_datetime(timestamp), minus_period(now(), to_period(`30D`)))"]).select(formulas=["user_id", "client_id"])
```

> I want to take that stream output and move it over to an append only table however it fails. So if I add:
```python
result = stream_to_append_only(stream)
```

> this fails with:
```
failed to create an append-only table. : RuntimeError: java.lang.IllegalArgumentException: Input is not a stream table!
```

I reproduced with:
```groovy
def t = TableTools.timeTableBuilder()
  .period("00:00:01")
  .streamTable(true)
  .build()
println(t.isStream())
println(t.select("Timestamp").isStream())
```

nbauernfeind:
Fixes #3476 and #2474.

Nightlies are running: https://github.com/nbauernfeind/deephaven-core/actions/runs/4327218730/

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/BaseTable.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/BaseTable.java
@@ -156,7 +156,7 @@ public abstract class BaseTable<IMPL_TYPE extends BaseTable<IMPL_TYPE>> extends 
         None,
 
         // legacy attributes
-        Flatten, Sort, UpdateView, Join, Filter,
+        Flatten, Sort, Join, Filter,
         // new attributes
         DropColumns, View, Reverse,
         /**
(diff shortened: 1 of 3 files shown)


nbauernfeind:
While we might want to propagate this properties, it's not as obvious what should be propagated and needs more discussion. Closing for now.

PR #3493, opened 2023-03-06
GitHub action to update the web version

mofojed:
- GitHub action that automatically updates the web versions to the latest and create a Pull Request for it
- Can be triggered manually through `workflow_dispatch`, or eventually will trigger it automatically from `web-client-ui` repo when publishing web packages action is completed
- Tested against my own repository. Example PR that was created there: https://github.com/mofojed/deephaven-core/pull/7
- If already at the latest version, then action exits successfully

### DIFF
--- a/.github/workflows/update-web.yml
+++ b/.github/workflows/update-web.yml
@@ -52,7 +52,7 @@ jobs:
           GRID_VERSION: ${{steps.web_versions.outputs.GRID_VERSION}}
           CHART_VERSION: ${{steps.web_versions.outputs.CHART_VERSION}}
         with:
-          # TODO: We need to add a PAT so that we can push to my fork probably
+          # TODO: We need to add a PAT so that we can push to my fork probably... or do we?
 #          path: ./deephaven-core
           base: main
           title: Update web version ${{ env.WEB_VERSION }}
--- a/.github/workflows/update-web.yml
+++ b/.github/workflows/update-web.yml
@@ -21,7 +21,7 @@ jobs:
         id: web_versions
         run: |
           echo "npm version is $(npm --version)"
-          echo "WEB_VERSION=0.34.0" >> $GITHUB_OUTPUT
+          echo "WEB_VERSION=0.29.0" >> $GITHUB_OUTPUT
           echo "GRID_VERSION=$(npm info @deephaven/embed-grid@latest version)" >> $GITHUB_OUTPUT
           echo "CHART_VERSION=$(npm info @deephaven/embed-chart@latest version)" >> $GITHUB_OUTPUT
       - name: Update deephaven-core
--- a/.github/workflows/update-web.yml
+++ b/.github/workflows/update-web.yml
@@ -1,16 +1,15 @@
 # This workflow will automatically create a PR in deephaven-core to update the WEB_VERSION
 name: Update web version
 
-on:
-  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-web.yml
```diff
@@ -0,0 +1,42 @@
+# This workflow will automatically create a PR in deephaven-core to update the WEB_VERSION
+name: Update web version
+
+on:
+  workflow_dispatch:
+  repository_dispatch:
+    types: [update_web]
```
devinrsmith, line 7: So, GitHub says 

> You can use the GitHub API to trigger a webhook event called [repository_dispatch](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#repository_dispatch) when you want to trigger a workflow for activity that happens outside of GitHub

>  You can manually trigger a workflow run using the GitHub API, GitHub CLI, or GitHub browser interface.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows

I've not used repository_dispatch before, but I'm wondering if it's still preferable to use "workflow_dispatch" via the API b/c of the wording "for activity that happens outside of GitHub".

.github/workflows/update-web.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  update-web:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout latest deephaven-core
+        uses: actions/checkout@v3
```
devinrsmith, line 13: The "Checkout latest deephaven-core" isn't technically what's going on.

https://github.com/actions/checkout

> When checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event. Otherwise, uses the default branch.

If we always want to run this against main, we should set `ref: main`

I think this impl is _ok_ as-is, but then I would just choose "Checkout" or something simpler for the name of this step.

.github/workflows/update-web.yml
```diff
@@ -0,0 +1,42 @@
...
+          sed -i "s/^ARG CHART_VERSION=.*/ARG CHART_VERSION=$CHART_VERSION/" ./web/client-ui/Dockerfile
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        env:
+          WEB_VERSION: ${{steps.web_versions.outputs.WEB_VERSION}}
+        with:
+          base: main
+          title: Update web version ${{ env.WEB_VERSION }}
+          body: Release notes https://github.com/deephaven/web-client-ui/releases/tag/v${{ env.WEB_VERSION }}
+          branch: update-web-test-${{ env.WEB_VERSION }}
```
devinrsmith, line 41: I would change this name, remove `test`. I think it might be preferable to have some folder-like prefix:

```
deephaven-bot/update-web-${{ env.WEB_VERSION }}
```

or something like that.

.github/workflows/update-web.yml
mofojed: Good call. I was going based off this post: https://github.com/orgs/community/discussions/26323#discussioncomment-3251443
But looks like it was recently updated to just use workflow_dispatch: https://github.com/orgs/community/discussions/26323#discussioncomment-3251448
I'll just remove the `repository_dispatch` trigger. Triggering the `workflow_dispatch` will still be done from the web-client-ui repo.

devinrsmith (approved):

PR #3495, opened 2023-03-07
Remove the explicit versioning of dependencies for the Python client testing

### LINKED ISSUE #3494: Unpin the versions of the Python client dependencies
1. based on a discussion I had with Colin, the key dependencies: arrow, protobuf, grpcio, pandas are very unlikely to introduce breaking changes, so to unpin them in the dependency requirements of the Python client is probably very safe.
2. when an incompatibility shows up in CI or dev's own testing, we have the option to either change our API code (preferred) or pin the affected package(s)
3. hopefully we never (say never) need to manually upgrade the dependencies again.

jmao-denver:
Fixes #3494

### DIFF
--- a/py/client/build.gradle
+++ b/py/client/build.gradle
@@ -86,17 +86,21 @@ tasks.getByName('check').dependsOn(Docker.registerDockerTask(project, 'testPyCli
         from('tests') {
             into 'project/tests'
         }
+        from ('requirements.txt') {
+            into 'project/'
+        }
     }
     containerDependencies.dependsOn = [deephavenDocker.healthyTask]
     containerDependencies.finalizedBy = deephavenDocker.endTask
     network = deephavenDocker.networkName.get()
     dockerfile {
         from('deephaven/python:local-build')
-        runCommand '''set -eux; \\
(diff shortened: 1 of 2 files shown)


devinrsmith (dismissed):

py/client/build.gradle
```diff
@@ -86,17 +86,21 @@ tasks.getByName('check').dependsOn(Docker.registerDockerTask(project, 'testPyCli
...
         from('deephaven/python:local-build')
-        runCommand '''set -eux; \\
-                      pip3 install unittest-xml-reporting==3.0.4 pyarrow==7.0.0 protobuf==3.20.1 grpcio==1.46.0 bitstring==3.1.9 pandas==1.2.5 timeout-decorator==0.5.0;\\
-                      mkdir -p /out/report'''
         copyFile('project', '/project')
         workingDir('/project')
+        runCommand '''set -eux; \\
+                      mkdir -p /out/report; \\
+                      pip3 install --upgrade pip; \\
+                      pip3 install -r requirements.txt'''
```
niloc132, line 103: For readers of this code, would it make sense to change the name of this file, so it is clear that this is only for testing?

jmao-denver: good suggestion, added requirements-dev.txt

niloc132 (approved):

PR #3500, opened 2023-03-07
If the server/proxy doesn't provide an error string, make one up

### LINKED ISSUE #3498: `coreClient.login` rejects with a "" error on failed auth
When trying to log in anonymously with anon auth disabled, `coreClient.login` rejects with an empty string error and doesn't trigger the `HACK_CONNECTION_FAILURE` event.

Expected:
Either reject with a proper error or trigger the `HACK_CONNECTION_FAILURE` with failure details.

niloc132:
Fixes #3498

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
@@ -457,9 +457,9 @@ public class WorkerConnection {
                     // TODO deephaven-core#2564 fire an event for the UI to re-auth
                     checkStatus(status);
                     if (status.getDetails() == null || status.getDetails().isEmpty()) {
-                        reject.onInvoke(status.getDetails());
-                    } else {
                         reject.onInvoke("Error occurred while authenticating, gRPC status " + status.getCode());
+                    } else {
+                        reject.onInvoke(status.getDetails());
                     }
                 }
             });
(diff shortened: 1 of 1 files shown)


web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -456,7 +456,11 @@ private Promise<Void> authUpdate() {
                     // token is no longer valid, signal deauth for re-login
                     // TODO deephaven-core#2564 fire an event for the UI to re-auth
                     checkStatus(status);
-                    reject.onInvoke(status.getDetails());
+                    if (status.getDetails() == null || status.getDetails().isEmpty()) {
+                        reject.onInvoke(status.getDetails());
```
vbabich, line 460: Should this be the opposite? Reject with `status.getDetails()` if it's not null or empty.

niloc132: Yes, it definitely should. I must have somehow not rebuilt when I un-reverted this...

niloc132: Fixed, tested by running the IDE with PSK enabled on the latest server, to confirm I didn't get the 16 error code, but the proper message `Authentication details invalid`.

vbabich (approved):

Issue #3501, opened 2023-03-07
JS API should support re-authenticating and rebuilding the session


niloc132:
As a user, if my session has expired and the client has the appropriate login information already, 

This may not be possible in all circumstances, but the JS API has enough information to re-request tables in cases such as server restart or extended connection loss, but it isn't being used today. Instead, reconnect only works as long as the session on the server is still active.

Among other things, this would require permitting the user to re-authenticate in some way. This will not work in all situations or for all objects - for example IdeConnection instances can't be rebuilt in this way.

PR #3502, opened 2023-03-07
JS API should attempt to reconnect to the gRPC server

### LINKED ISSUE #225: JS API should only request logs it hasn't seen yet
The JS API internally plays back old logs, and the server will also keep around old logs (to a point) to replay for new connections or reconnects. We should pick one or the other so that we provide the sorts of guarantees about usage that we would like.

Another approach could be to offer a Table instance which holds the logs, so that it follows the same API as every other table.

### LINKED ISSUE #730: Complete, test JS API reconnects

niloc132:
As long as a user's session on the server remains active, this should enable a client to reconnect to that session, and continue using already-exported objects. Some of the work is restored here for #3501 to be finished as well, but the client would need a way to create a new session without user interaction.

The dh.IdeConnection.HACK_CONNECTION_FAILURE event is now deprecated, and should be removed. Clients should instead use the EVENT_DISCONNECTED and EVENT_RECONNECTED events to signal to the user that an object isn't available.

Also fixes a bug where the last seen log item is replayed incorrectly.

Fixes #730
Fixes #225


web/client-api/src/main/java/io/deephaven/web/client/api/grpc/MultiplexedWebsocketTransport.java
```diff
@@ -305,6 +309,8 @@ private void removeHandlers() {
         cleanup.run();
         cleanup = JsRunnable.doNothing();
 
+        // release our reference to the transport, last one out will close the socket (if needed)
+        // TODO ensure this is called once per instance
```
niloc132, line 313: Need to re-check this, I don't quite trust it, but also haven't seen it fail.

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -473,12 +520,15 @@ private void subscribeToTerminationNotification() {
...
+                            connectionLost();
                         }
                         return;
                     }
                     assert success != null;
 
                     // welp; the server is gone -- let everyone know
+                    connectionLost();
+
                     info.notifyConnectionError(new ResponseStreamWrapper.Status() {
```
niloc132, line 532: Removing the notifyConnectionError here, and instead going with a disconnect event exposes a missing case that we need to consider for DHC, how to signal that the server has shut down _and told the client this_, but how to potentially handle the server coming back again afterwards. We probably need a new event, or metadata on the disconnect event, but likewise should probably welcome the server back by prompting another login (right now we'll get the "reauth failed" event) by the user.

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -366,7 +366,7 @@ public void record(LogBufferRecord record) {
             }
             // since the subscribe() method auto-replays all existing logs, filter to just once this consumer probably
             // hasn't seen
-            if (record.getTimestampMicros() < request.getLastSeenLogTimestamp()) {
+            if (record.getTimestampMicros() <= request.getLastSeenLogTimestamp()) {
```
nbauernfeind, line 369: This is just a note, not really feedback or a request other than to consider whether or not we should change the API.

Interesting. The listener could miss log messages that occurred at the same timestamp, but I see how the alternative guarantees at least one duplicate message. Maybe a better API would count the log messages and you just provide an offset. (This is probably why Kafka uses an offset instead of timestamp for new subscriptions.)

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -285,53 +287,66 @@ private void connectToWorker() {
...
+                                .peek(cts -> {
+                                    cts.forActiveTables(t -> {
+                                        assert t.state().isAncestor(cts)
+                                                : "Invalid binding " + t + " (" + t.state() + ") does not contain "
+                                                        + cts;
+                                    });
+                                })
+                                .toArray(ClientTableState[]::new);
+                        // clear caches
+                        // TODO verify we still need to do this, it seems like it might signify a leak
```
nbauernfeind, line 318: does this need a deephaven-core issue number?

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
nbauernfeind: Subscribers of TerminationNotification get a tombstone from the server with `abnormalTermination` set to `false`. We can probably use this to indicate a normal / expected shutdown? Though, I personally get these on my machines as the DH instances restart whenever they are redeployed. Either way, maybe this is the path you want to use?

web/client-api/src/main/java/io/deephaven/web/client/ide/IdeConnection.java
```diff
@@ -22,6 +22,9 @@
  */
 @JsType(namespace = "dh")
 public class IdeConnection extends QueryConnectable<IdeConnection> {
+    public static final String EVENT_DISCONNECT = "disconnect";
+    public static final String EVENT_RECONNECT = "reconnect";
```
devinrsmith, line 26: I don't see these being used - are these meant to be referenced by client code?

devinrsmith (commented):
I'm a bit out of my depth when it comes to reviewing JS API code.

Is there an sample client code to show EVENT_DISCONNECTED / EVENT_RECONNECTED event handling? Is there any "easy" way to simulate server disconnections?

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
niloc132: We do get "expected shutdown", but if you are testing an application or some server side debugging, you're going to edit, restart, edit, restart - an expected shutdown isn't the same as "never come back without user interaction".

Success or failure we can tell the client that the disconnect happened, but the question then is do we say "the server is down (here's the error|this was on purpose)" or just say "its down, it might be up again soon, please dismiss the message"?

web/client-api/src/main/java/io/deephaven/web/client/ide/IdeConnection.java
niloc132: Yes, we declare constants for events that can be fired on the class that can fire them.

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
niloc132: Yes, or I should just finish fixing this.

mofojed:
I opened a draft PR for UI to update for this: https://github.com/deephaven/web-client-ui/pull/1149
I'm not sure how best to test this, but I did just try with just killing the connection to see what happens. Couple notes:
- No error `details` are passed with the `EVENT_DISCONNECT`. Would be useful if there are possible different disconnect scenarios (such as "socket going down and we might come back" vs. "worker shutdown")
- The disconnect event seems to be getting fired 3 times:
image
image
image

niloc132:
> No error details are passed with the EVENT_DISCONNECT. Would be useful if there are possible different disconnect scenarios (such as "socket going down and we might come back" vs. "worker shutdown")

"worker shutdown" is definitely a "we might come back" scenario? At the very least, DnD PQs will have to anticipate this, and will expect (one day) to be able to attempt reconnect here (though _at present_ the JS API can't support this, only due to re-auth concerns).


> The disconnect event seems to be getting fired 3 times:...

This is consistent with dhe workers, both PQs and ad hoc, right? Its actually N times, based on how long you are disconnected for, how fast the reconnect retries to connect again. The original idea was that the UI could get a sense of how long it had been since the last try.

On the other hand, Tables (and formerly treetables, need to look closer at DHC here) only fire once, and suppress all later events until reconnected.

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
niloc132: Ah that's fair. Dup messages were confusing me, and the comment made it appear that this was "filtering to [ones] this consumer probably hadn't seen" meant something other than "will definitely always dup the last message seen, and maybe more". @devinrsmith any thoughts? I think you touched this last?

I know we will sometimes have gaps since this is implemented as a ring buffer, and maybe we tacitly accept that?

web/client-api/src/main/java/io/deephaven/web/client/ide/IdeConnection.java
niloc132: (sorry for the dup, i had this open and it didnt show the old message, so thought i hadnt responded...)

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
niloc132: Added a new event on IdeConnection, EVENT_SHUTDOWN that will emit just the status message from the shutdown, rather than the full grpc status. It will only fire if the client gets notification that the server is explicitly going away, as opposed to losing that stream (or any other stream).

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
niloc132: Okay, so I'll revert this and fix the comment instead?

mofojed (approved):
Looks good with my Web UI changes: https://github.com/deephaven/web-client-ui/pull/1149

niloc132:
I added one more commit to remove a logged error message that can occur when the user's session is resumed via a new connection.

mofojed (dismissed):

PR #3506, opened 2023-03-08
Enable Support for Multiple Update Graph Processors

nbauernfeind:
Here is some sample Groovy usage:
```groovy
import io.deephaven.util.SafeCloseable
import io.deephaven.engine.updategraph.impl.PeriodicUpdateGraph
import io.deephaven.engine.context.ExecutionContext

oobUpdateGraph = PeriodicUpdateGraph.newBuilder("OOB").numUpdateThreads(1).build()
try (SafeCloseable ignored = ExecutionContext.getContext().withUpdateGraph(oobUpdateGraph).open()) {
    t = TableTools.timeTable("PT00:00:00.1").tail(1)
}
```

Observe that the table still ticks when the REPL is holding the exclusive DEFAULT PUG lock:
```groovy
Thread.sleep(20000)
println("done")
```

Note that as-is the UGP stops and does not propagate a stop-related error notification to source table listeners. We might want to do something to be clear that the tables are done. We cannot purely set reference count to zero as this does not propagate downstream. I did try out such a change, but it some tests that were not expecting liveness referents / artifacts to fail `manage` calls.

Nightlies: https://github.com/nbauernfeind/deephaven-core/actions/runs/4789971295


Generators/Generators.gradle
```diff
@@ -25,6 +25,8 @@ dependencies {
     runtimeOnly project(path: ':configs')
     runtimeOnly project(path: ':test-configs')
     runtimeOnly project(':extensions-kafka')
+    // dh-default.prop references io.deephaven.server.console.python.DebuggingInitializer
+    runtimeOnly project(':server')
```
devinrsmith, line 29: We should prefer to fix our "default" props, or have a different props file for Generators instead of adding this dependency IMO. (I've seen io.deephaven.server.console.python.DebuggingInitializer errors pop up in other places, so it may be worthwhile thinking about a more modular / dependency-sensitive approach to prop files instead of inheriting a single defaults file...)

devinrsmith:
I haven't reviewed this PR, but I have done a quick migration of `io.deephaven.app.GcApplication` to use its own UGP; I was able to verify that the GcApplication was able to tick while the main UGP was busy.

chipkent (approved):

chipkent (approved):

engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -89,6 +91,39 @@ public interface Table extends
...
+
+        for (final Table other : others) {
+            if (other == null || !other.isRefreshing()) {
+                continue;
+            }
+            if (foundUpdateContext == null) {
+                foundUpdateContext = other.getUpdateContext();
+            } else if (foundUpdateContext != other.getUpdateContext()) {
+                throw new IllegalStateException("Tables from different update contexts cannot be mixed. Found: "
+                        + foundUpdateContext + " and " + other.getUpdateContext());
```
rcaudy, line 117: Another benefit to naming UGPs...

engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -89,6 +91,39 @@ public interface Table extends
...
+    @ConcurrentMethod
+    UpdateContext getUpdateContext();
+
+    /**
+     * Ensure that the table is compatible with the currently active update context.
+     *
+     * @param others Other source tables to check for consistency
+     */
+    @FinalDefault
+    default void checkUpdateContextConsistency(final Table... others) {
```
rcaudy, line 103: We don't like defaults in `Table` anymore. Maybe put this in `TableDefaults`.

engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -89,6 +91,39 @@ public interface Table extends
...
+            }
+            if (foundUpdateContext == null) {
+                foundUpdateContext = other.getUpdateContext();
+            } else if (foundUpdateContext != other.getUpdateContext()) {
+                throw new IllegalStateException("Tables from different update contexts cannot be mixed. Found: "
+                        + foundUpdateContext + " and " + other.getUpdateContext());
+            }
+        }
+
+        if (foundUpdateContext != null && foundUpdateContext != UpdateContext.get()) {
```
rcaudy, line 121: Why do we care about this? We care about using the logical clock from the tables' UGP, and/or the locks from their UGP. I think only new roots should be finding their UGP from the context, derived tables should be inheriting.

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -550,6 +550,7 @@ public Object[] getRecord(long rowNo, String... columnNames) {
 
     @Override
     public PartitionedTable partitionBy(final boolean dropKeys, final String... keyColumnNames) {
+        checkUpdateContextConsistency();
```
rcaudy, line 553: I think this should look something like opening a context.

server/src/main/java/io/deephaven/server/table/ops/JoinTablesGrpcImpl.java
```diff
@@ -86,8 +84,11 @@ public Table create(final T request, final List<SessionState.ExportObject<Table>
         if (!lhs.isRefreshing() && !rhs.isRefreshing()) {
             result = realTableOperation.apply(lhs, rhs, columnsToMatch, columnsToAdd, request);
         } else {
-            result = updateGraphProcessor.sharedLock().computeLocked(
-                    () -> realTableOperation.apply(lhs, rhs, columnsToMatch, columnsToAdd, request));
+            final AwareFunctionalLock sharedLock = lhs.isRefreshing()
+                    ? lhs.getUpdateContext().getSharedLock()
+                    : rhs.getUpdateContext().getExclusiveLock();
```
rcaudy, line 89: This should always be the shared lock. And you want to be sure both parents have the same UGP. I think you want to use `getUpdateContext(lhs, rhs)` earlier to identify the lock to use.

server/src/main/java/io/deephaven/server/table/ops/GrpcTableOperation.java
```diff
@@ -147,4 +149,42 @@ public void checkPermission(final T request, final List<SessionState.ExportObjec
...
+    /**
+     * Get the appropriate update context for this operation. Returns the first update context found from a refreshing
+     * table.
+     *
+     * @param source the required "left hand side" source table
+     * @param otherSources any other source tables
+     * @return the update context
+     */
+    @NotNull
+    protected static UpdateContext getUpdateContext(final Table source, final Table... otherSources) {
```
rcaudy, line 162: This should probably check consistency.

rcaudy (commented):
High level review. Suggested some refactoring, and a change to the policy that determines UGP for derived tables.

rcaudy:
We discussed in-person about maybe just having the idea of an `UpdateGraph` object that hides the "processor" details, with a `clock()`, `sharedLock()`, `exclusiveLock()`, and other appropriate interfaces exposed. `LogicalClock` should probably have an interface that does not expose the mutating methods.

mofojed:
@rcaudy Is this targeting June release now?

engine/test-utils/src/main/java/io/deephaven/engine/testutil/testcase/RefreshingTableTestCase.java
```diff
@@ -57,36 +55,34 @@ public static int scaleToDesiredTestLength(final int maxIter) {
     public void setUp() throws Exception {
         super.setUp();
 
-        UpdateGraphProcessor.DEFAULT.enableUnitTestMode();
-        UpdateGraphProcessor.DEFAULT.resetForUnitTests(false);
+        UpdateContext.updateGraphProcessor().enableUnitTestMode();
```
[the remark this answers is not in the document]
nbauernfeind: I've filed this as: #3886

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateContext.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        /**
+         * This enables Unit Test Mode. Unit tests mode allows complete control over the update graph processor. This is
+         * useful for testing boundary conditions to validate the behavior of table operations.
+         *
+         * @param allowUnitTestMode true to allow unit test mode
+         * @return this builder
+         */
+        public Builder allowUnitTestMode(boolean allowUnitTestMode) {
```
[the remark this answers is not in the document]
nbauernfeind: As discussed offline, we agree there is no reason to limit this for all Update Graphs but rather just the DEFAULT.

chipkent (dismissed):

rcaudy (dismissed):
Approving. Please don't merge unless nightlies are green.

py/server/deephaven/execution_context.py
```diff
@@ -36,6 +37,10 @@ class ExecutionContext(JObjectWrapper, ContextDecorator):
     def j_object(self) -> jpy.JType:
         return self.j_exec_ctx
 
+    @property
+    def update_graph(self) -> _JUpdateGraph:
```
chipkent, line 41: Two problems:
1) The API is exposing Java objects to the user.
2) The type hint is referring to a private variable.

py/server/deephaven/table.py
```diff
@@ -537,6 +539,15 @@ def is_refreshing(self) -> bool:
             self._is_refreshing = self.j_table.isRefreshing()
         return self._is_refreshing
 
+    @property
+    def update_graph(self) -> _JUpdateGraph:
```
chipkent, line 543: Two problems:
1) The API is exposing Java objects to the user.
2) The type hint is referring to a private variable.

py/server/deephaven/table.py
```diff
@@ -2298,6 +2309,11 @@ def table(self) -> Table:
             self._table = Table(j_table=self.j_partitioned_table.table())
         return self._table
 
+    @property
+    def update_graph(self) -> _JUpdateGraph:
```
chipkent, line 2313: Two problems:
1) The API is exposing Java objects to the user.
2) The type hint is referring to a private variable.

py/server/deephaven/table.py
```diff
@@ -2554,6 +2570,11 @@ def is_refreshing(self) -> bool:
         """Whether this proxy represents a refreshing partitioned table."""
         return self.target.is_refreshing
 
+    @property
+    def update_graph(self) -> _JUpdateGraph:
```
chipkent, line 2574: Two problems:
1) The API is exposing Java objects to the user.
2) The type hint is referring to a private variable.

py/server/deephaven/table_listener.py
```diff
@@ -237,25 +238,29 @@ def modified_columns(self) -> List[str]:
         return list(cols) if cols else []
 
 
-def _do_locked(f: Callable, lock_type="shared") -> None:
-    """Executes a function while holding the UpdateGraphProcessor (UGP) lock.  Holding the UGP lock
+def _do_locked(ug: Union[_JUpdateGraph, Table], f: Callable, lock_type="shared") -> None:
```
chipkent, line 241: Two problems:
1) The API is exposing Java objects to the user.
2) The type hint is referring to a private variable.

py/server/deephaven/update_graph.py
```diff
@@ -0,0 +1,222 @@
...
+
+_JUpdateGraph = jpy.get_type("io.deephaven.engine.updategraph.UpdateGraph")
+
+auto_locking = True
+"""Whether to automatically acquire the Update Graph (UG) shared lock for an unsafe operation on a refreshing 
+table when the current thread doesn't own either the UG shared or the UG exclusive lock. The newly obtained lock will 
+be released after the table operation finishes. Auto locking is turned on by default."""
+
+
+def has_exclusive_lock(ug: Union[_JUpdateGraph, "Table", "PartitionedTable", "PartitionTableProxy"]) -> bool:
```
chipkent, line 25: Two problems:
1) The API is exposing Java objects to the user.
2) The type hint is referring to a private variable.

chipkent: Many places in this file.

chipkent (changes requested):

chipkent (dismissed):

rcaudy (approved):

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2731
Conceptual: https://github.com/deephaven/deephaven.io/issues/2728
Reference: https://github.com/deephaven/deephaven.io/issues/2730
Blog: https://github.com/deephaven/deephaven.io/issues/2729

PR #3509, opened 2023-03-08
Support the use of Python callable's attributes in query strings

### LINKED ISSUE #3508: Can't use attributes on Python callable objects in queries
```

# read_csv will accept a url or a path to a local file
seattle_weather = read_csv("https://media.githubusercontent.com/media/deephaven/examples/main/GSOD/csv/seattle.csv")

from deephaven.time import year, TimeZone
from deephaven import agg

TZ_NY = TimeZone.NY
hi_lo_by_year = seattle_weather.view(formulas=["Year = (int)year(ObservationDate, TimeZone.NY)", "TemperatureF"])\
    .where(filters=["Year >= 2000"])\
    .agg_by([\
        agg.avg(cols=["Avg_Temp = TemperatureF"]),\
        agg.min_(cols=["Lo_Temp = TemperatureF"]),\
        agg.max_(cols=["Hi_Temp = TemperatureF"])
        ],\
     by=["Year"])
```
Running the above code causes an exception
```
Full expression           : (int)year(ObservationDate, TimeZone.NY)
Expression having trouble : io.deephaven.engine.table.impl.lang.QueryLanguageParser$VisitArgs@43adc7c1
Exception message         : Cannot resolve field name.
    Expression : TimeZone.NY
    Scope      : TimeZone
    Scope Type : io.deephaven.engine.util.PyCallableWrapper
    Field Name : NY
```
The workaround is to assign the class attribute to a local variable first (```TZ_NY``` in the code) and replace ```TimeZone.NY``` with it in the query string, but it is not natural and confusing to the users.

jmao-denver:
Fixes #3508

### DIFF
--- a/py/server/tests/test_table.py
+++ b/py/server/tests/test_table.py
@@ -5,9 +5,10 @@ import unittest
 from types import SimpleNamespace
 from typing import List, Any
 
-from deephaven import DHError, read_csv, empty_table, SortDirection, AsOfMatchRule, time_table, ugp
+from deephaven import DHError, read_csv, empty_table, SortDirection, AsOfMatchRule, time_table, ugp, new_table, dtypes
 from deephaven.agg import sum_, weighted_avg, avg, pct, group, count_, first, last, max_, median, min_, std, abs_sum, \
     var, formula, partition
+from deephaven.column import datetime_col
 from deephaven.execution_context import make_user_exec_ctx
 from deephaven.html import to_html
 from deephaven.jcompat import j_hashmap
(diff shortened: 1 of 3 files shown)


rcaudy (dismissed):
Please be sure to run nightlies before merging.

py/server/tests/test_table.py
```diff
@@ -902,6 +903,34 @@ def make_pairs_3(tid, a, b):
             self.assertEqual(x2.size, 10)
             self.assertEqual(x3.size, 10)
 
+    def test_class_attrs_in_query(self):
+        input_cols = [
+            datetime_col(name="DTCol", data=[dtypes.DateTime(1), dtypes.DateTime(10000000)]),
+        ]
+        test_table = new_table(cols=input_cols)
+        from deephaven.time import year, TimeZone
+        rt = test_table.update("Year = (int)year(DTCol, TimeZone.NY)")
```
rcaudy, line 912: Maybe add a test case where you cast an argument in addition to the result of the expression.

jmao-denver:
> Please be sure to run nightlies before merging.

Nightly check has finished successfully.

rcaudy (dismissed):

rcaudy (dismissed):

py/server/tests/test_table.py
```diff
@@ -902,6 +903,42 @@ def make_pairs_3(tid, a, b):
...
+    def test_callable_attrs_in_query(self):
+        input_cols = [
+            datetime_col(name="DTCol", data=[dtypes.DateTime(1), dtypes.DateTime(10000000)]),
+        ]
+        test_table = new_table(cols=input_cols)
+        from deephaven.time import year, TimeZone
+        rt = test_table.update("Year = (int)year(DTCol, TimeZone.NY)")
+        self.assertEqual(rt.size, test_table.size)
+
+        class Foo:
```
chipkent, line 915: The java code above is working with Python attributes.  From a little googling, python attributes can be instance methods, static methods, and class methods.  We are only testing instance methods.  Tests should be included for static methods and class methods to make sure they also work.

chipkent (changes requested):

py/server/tests/test_table.py
jjbrosnan: This tests an instance attribute, but it's not testing a method. These tests cover class, instance, and static methods *and* variables if the PR aims to make all of them usable from query strings.

jjbrosnan (commented):
Response to Chip's comment about tests.

chipkent (approved):

jjbrosnan (approved):

PR #3511, opened 2023-03-09
Move rpcWrapper to gRPC Interceptor; Open Session's ExecutionContext

nbauernfeind:
I've cleaned up the `rpcWrapper` by moving it into the session object injection itself.
I've also opened the session's ExecutionContext in the context of all gRPC requests. 

DHE can now fetch the AuthContext during Ticket resolution as originally intended.

Nightlies are running here: https://github.com/nbauernfeind/deephaven-core/actions/runs/4370246882


rcaudy (dismissed):
LGTM, merge once Colin is also happy.

niloc132 (commented):
I think the title should be adjusted to say that the rpcWrapper is moved to the interceptor, or perhaps call listener?

niloc132 (approved):

PR #3512, opened 2023-03-09
Add SearchDisplayMode layout hints

mofojed:
- Ported over from Enterprise, where these are already implemented
- Not sure how best to handle the enum from the Python API, just using a String right now. Will get Jianfeng/Chip to review.
- Tested in Python and Groovy, ensuring you could bring up the Search Bar (from the side menu) in table `canSearch`, and it was not available in `cannotSearch`.
Python:
```python
t = empty_table(100).update(["a=`abc`"])
canSearch = t.layout_hints(search_display_mode="Show")
cannotSearch = t.layout_hints(search_display_mode="Hide")
```
Groovy:
```groovy
import io.deephaven.engine.util.LayoutHintBuilder.SearchDisplayModes

t = emptyTable(5).updateView("a=`abc`")
canSearch = t.setLayoutHints(LayoutHintBuilder.get().setSearchBarAccess(SearchDisplayModes.Show).build())
cannotSearch = t.setLayoutHints(LayoutHintBuilder.get().setSearchBarAccess(SearchDisplayModes.Hide).build())
```

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -1852,7 +1852,7 @@ class Table(JObjectWrapper):
 
     def layout_hints(self, front: Union[str, List[str]] = None, back: Union[str, List[str]] = None,
                      freeze: Union[str, List[str]] = None, hide: Union[str, List[str]] = None,
-                     column_groups: List[dict] = None) -> Table:
+                     column_groups: List[dict] = None, search_display_mode: str = None) -> Table:
         """ Sets layout hints on the Table
 
         Args:
@@ -1867,6 +1867,8 @@ class Table(JObjectWrapper):
                 name (str): The group name
                 children (List[str]): The
                 color (Optional[str]): The hex color string or Deephaven color name
+            search_display_mode (str): set the search bar to explicitly be accessible or inaccessible, or use the system default.
+                Values can be one of `Show`, `Hide`, and `Default`
 
         Returns:
             a new table with the layout hints set
@@ -1893,6 +1895,10 @@ class Table(JObjectWrapper):
                 for group in column_groups:
                     _j_layout_hint_builder.columnGroup(group.get("name"), j_array_list(group.get("children")),
                                                        group.get("color", ""))
+
+            if search_display_mode is not None:
+                _j_layout_hint_builder.setSearchBarAccess(search_display_mode)
+
         except Exception as e:
             raise DHError(e, "failed to create layout hints") from e
 
--- a/engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/LayoutHintBuilder.java
@@ -33,6 +33,27 @@ public class LayoutHintBuilder {
     private Map<String, ColumnGroup> columnGroups;
 
     /**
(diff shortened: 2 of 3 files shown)


py/server/deephaven/table.py
```diff
@@ -1852,7 +1852,7 @@ def format_row_where(self, cond: str, formula: str) -> Table:
 
     def layout_hints(self, front: Union[str, List[str]] = None, back: Union[str, List[str]] = None,
                      freeze: Union[str, List[str]] = None, hide: Union[str, List[str]] = None,
-                     column_groups: List[dict] = None) -> Table:
+                     column_groups: List[dict] = None, search_display_mode: str = None) -> Table:
```
mofojed, line 1855: @jmao-denver is there a better way to use an `enum` from in python? I tried using `jpy.get_type("io.deephaven.engine.util.LayoutHintBuilder.SearchDisplayModes")` but I get a Class not found error.

niloc132 (approved):
Approved js api changes

py/server/deephaven/table.py
jmao-denver: it is an inner class, the JNI syntax is to use '$' notation.
```
io.deephaven.engine.util.LayoutHintBuilder$SearchDisplayModes
```
Please see this file for an example wrapping of an Enum.
```
py/server/deephaven/plot/linestyle.py
```

py/server/deephaven/table.py
jmao-denver, line 1: add a test case in test_table.py: test_layout_hints()?

py/server/deephaven/table.py
jmao-denver: I am actually not very sure about introducing an Enum in this case. I am fine with the parameter being a string with 3 valid, self-evident values.

py/server/deephaven/table.py
```diff
@@ -1867,6 +1867,8 @@ def layout_hints(self, front: Union[str, List[str]] = None, back: Union[str, Lis
                 name (str): The group name
                 children (List[str]): The
                 color (Optional[str]): The hex color string or Deephaven color name
+            search_display_mode (str): set the search bar to explicitly be accessible or inaccessible, or use the system default.
+                Values can be one of `Show`, `Hide`, and `Default`
```
jmao-denver, line 1872: Should expand on what 'Default' means?

jmao-denver (changes requested):

py/server/deephaven/table.py
mofojed: Okay, I've used an Enum, let me know if you want me to revert to string. I think @chipkent has previously said he prefers enums, not sure. Downside is you need a `from deephaven.table import SearchDisplayMode` before this.

Or I could make it accept both a string or the enum?

py/server/deephaven/table.py
mofojed: I've added tests, and updated the other `test_layout_hints()` to actually verify the layout_hint attribute. I'm not sure how to run these tests locally though - I was running `./gradlew :py-server:check` but it didn't seem to run these tests. How do I run just these tests locally?

jmao-denver (approved):
LGTM

chipkent (approved):

PR #3514, opened 2023-03-09
Stop UGP NotificationProcessor From Silently Swallowing Errors

nbauernfeind:
In the Array Column Access PR nightly `testSerial` was timing out. After isolating the issue to a specific FuzzerTest, I was able to observe unwanted behavior when running from IJ.

In particular, `RefreshingTableTestCase` throws an `AssertionError` via `TestCase.fail()` if a table propagates a bad update. I've added an additional check to validate `APPEND_ONLY` tables. This error was causing the notification processing threads to exit, eventually starving the notification queue of threads to process them. The UGP was waiting for Godot, causing CI's `testSerial` to be forced to exit after 6 hrs.

This did not occur in my testing because the error was not common enough to starve my executor from processing threads. The machine I was using has 8 cores, but I suspect CI is at 2.

I've come up with a fix that should catch losing notification threads in non-test scenarios, too.

This is a new issue caused by enabling multi-threaded processing by default.
Running nightlies in case there were any tests silently failing: https://github.com/nbauernfeind/deephaven-core/actions/runs/4379371283

### DIFF
--- a/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
+++ b/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
@@ -1392,6 +1392,10 @@ public enum UpdateGraphProcessor implements UpdateSourceRegistrar, NotificationQ
             while (outstandingNotificationsCount() > 0) {
                 doWork();
             }
+            // A successful and a failed notification may race the release of pendingNormalNotificationsCheckNeeded,
+            // causing this thread to miss a false isHealthy. Since isHealthy is set prior to decrementing
+            // outstandingNotificationsCount, we're guaranteed to read the correct value after exiting the while loop.
+            Assert.eqTrue(isHealthy, "isHealthy");
         }
 
         @Override
--- a/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
+++ b/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
@@ -1092,15 +1092,15 @@ public enum UpdateGraphProcessor implements UpdateSourceRegistrar, NotificationQ
      */
     private void flushNotificationsAndCompleteCycle() {
         // We cannot proceed with normal notifications, nor are we satisfied, until all update source refresh
-        // notifications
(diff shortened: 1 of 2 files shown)


engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
```diff
@@ -1314,27 +1322,35 @@ public ConcurrentNotificationProcessor(@NotNull final ThreadFactory threadFactor
...
-                Notification satisfiedNotification = null;
-                synchronized (satisfiedNotifications) {
-                    while (running && (satisfiedNotification = satisfiedNotifications.poll()) == null) {
-                        try {
-                            satisfiedNotifications.wait();
-                        } catch (InterruptedException ignored) {
+            Notification satisfiedNotification = null;
+            try {
+                while (running) {
+                    satisfiedNotification = null;
```
rcaudy, line 1328: This should be at the end of the loop, not the beginning, but still before the accounting.

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
```diff
@@ -1301,6 +1308,7 @@ private class ConcurrentNotificationProcessor implements NotificationProcessor {
         private final Semaphore pendingNormalNotificationsCheckNeeded = new Semaphore(0, false);
 
         private volatile boolean running = true;
+        private volatile boolean anyThreadHasExited = false;
```
rcaudy, line 1311: I think you should invert this.

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphProcessor.java
```diff
@@ -1259,6 +1259,13 @@ private interface NotificationProcessor {
          * Called before pending notifications are drained.
          */
         void beforeNotificationsDrained();
+
+        /**
+         * @return true if this processor is healthy and in an expected state
+         */
+        default boolean isHealthy() {
```
rcaudy, line 1266: We said we would take this out of the API, and move the assertion to `doWork()` post-acquire for the concurrent NP.

rcaudy (approved):

rcaudy (approved):

PR #3526, opened 2023-03-10
QueryLanguageParser: Select Best Method by Cost

### LINKED ISSUE #3461: CI failure for booleanAsByte
We seem to be resolving the wrong overload for `BooleanUtils.booleanAsByte` at least some of the time.
https://scans.gradle.com/s/ig3cve3w726uw/tests/:engine-table:testOutOfBand/io.deephaven.engine.table.impl.sources.regioned.TestChunkedRegionedOperations/testSparseTableSmallChunks?top-execution=1

nbauernfeind:
There is some janky logic used to determine whether one candidate method is better than another candidate. It did not consider the costs of boxing/unboxing/widening/etc parameters. This PR adds a rough cost based on the kinds of conversions that were required to make the method a candidate.

Fixes #3461.

I'm going to try running nightlies a few times. I've seen it hit one of the three out of band tests pretty consistently recently.
Trial 1: https://github.com/nbauernfeind/deephaven-core/actions/runs/4390904494
Trial 2: https://github.com/nbauernfeind/deephaven-core/actions/runs/4390904618
Trial 3: https://github.com/nbauernfeind/deephaven-core/actions/runs/4390904723

(Note Trial 2 had an unrelated fuzzer test that I am looking into; I fixed the check error but didn't re-run trials.)

Latest nightlies: https://github.com/nbauernfeind/deephaven-core/actions/runs/4408689379/

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
@@ -467,7 +467,7 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
 
     private Method getMethod(final Class<?> scope, final String methodName, final Class<?>[] paramTypes,
             final Class<?>[][] parameterizedTypes) {
-        final ArrayList<Method> acceptableMethods = new ArrayList<>();
+        final ArrayList<CandidateExecutable<Method>> acceptableMethods = new ArrayList<>();
 
         if (scope == null) {
             for (final Class<?> classImport : staticImports) {
@@ -488,21 +488,19 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
                     variablesUsed.add(methodName);
                 }
             }
+        } else if (scope == org.jpy.PyObject.class || scope == PyCallableWrapper.class) {
+            // This is a Python method call, assume it exists and wrap in PythonScopeJpyImpl.CallableWrapper
+            for (Method method : PyCallableWrapper.class.getDeclaredMethods()) {
+                possiblyAddExecutable(acceptableMethods, method, "call", paramTypes, parameterizedTypes);
+            }
         } else {
-            if (scope == org.jpy.PyObject.class || scope == PyCallableWrapper.class) {
-                // This is a Python method call, assume it exists and wrap in PythonScopeJpyImpl.CallableWrapper
-                for (Method method : PyCallableWrapper.class.getDeclaredMethods()) {
-                    possiblyAddExecutable(acceptableMethods, method, "call", paramTypes, parameterizedTypes);
-                }
-            } else {
-                for (final Method method : scope.getMethods()) {
+            for (final Method method : scope.getMethods()) {
+                possiblyAddExecutable(acceptableMethods, method, methodName, paramTypes, parameterizedTypes);
+            }
+            // If 'scope' is an interface, we must explicitly consider the methods in Object
+            if (scope.isInterface()) {
+                for (final Method method : Object.class.getMethods()) {
                     possiblyAddExecutable(acceptableMethods, method, methodName, paramTypes, parameterizedTypes);
-                }
-                // If 'scope' is an interface, we must explicitly consider the methods in Object
-                if (scope.isInterface()) {
-                    for (final Method method : Object.class.getMethods()) {
-                        possiblyAddExecutable(acceptableMethods, method, methodName, paramTypes, parameterizedTypes);
-                    }
                 }
             }
         }
@@ -512,14 +510,14 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
                     + (scope != null ? " in " + scope : ""));
         }
 
-        Method bestMethod = null;
-        for (final Method method : acceptableMethods) {
+        CandidateExecutable<Method> bestMethod = null;
+        for (final CandidateExecutable<Method> method : acceptableMethods) {
             if (bestMethod == null || isMoreSpecificMethod(bestMethod, method)) {
                 bestMethod = method;
             }
         }
 
-        return bestMethod;
+        return bestMethod.executable;
     }
 
     private static boolean isPotentialImplicitCall(Class<?> methodClass) {
@@ -631,7 +629,7 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
     @SuppressWarnings({"ConstantConditions"})
     private Constructor<?> getConstructor(final Class<?> scope, final Class<?>[] paramTypes,
             final Class<?>[][] parameterizedTypes) {
-        final ArrayList<Constructor<?>> acceptableConstructors = new ArrayList<>();
+        final ArrayList<CandidateExecutable<Constructor<?>>> acceptableConstructors = new ArrayList<>();
 
         for (final Constructor<?> constructor : scope.getConstructors()) {
             possiblyAddExecutable(acceptableConstructors, constructor, scope.getName(), paramTypes, parameterizedTypes);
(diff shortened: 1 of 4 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -698,88 +716,113 @@ private static <EXECUTABLE_TYPE extends Executable> void possiblyAddExecutable(
...
-        if (e1.isVarArgs() && e2.isVarArgs()) {
+        if (e1.executable.isVarArgs() && e2.executable.isVarArgs()) {
             e1ParamTypes[e1ParamTypes.length - 1] = e1ParamTypes[e1ParamTypes.length - 1].getComponentType();
             e2ParamTypes[e2ParamTypes.length - 1] = e2ParamTypes[e2ParamTypes.length - 1].getComponentType();
         }
 
         for (int i = 0; i < e1ParamTypes.length; i++) {
-            if (!canAssignType(e1ParamTypes[i], e2ParamTypes[i]) && !isTypedVector(e2ParamTypes[i])) {
+            int costToAssign = costToAssignType(e1ParamTypes[i], e2ParamTypes[i]);
+            if (costToAssign == -1 && !isTypedVector(e2ParamTypes[i])) {
```
rcaudy, line 825: `-1` should be replaced by the appropriate constant

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -792,21 +835,36 @@ private static <EXECUTABLE_TYPE extends Executable> Boolean isMoreSpecificExecut
...
-            return classA.isAssignableFrom(classB);
+        if (boxedDest == src) {
+            return CandidateExecutable.BOX_CONVERSION;
+        } else if (boxedSrc == dest) {
+            return CandidateExecutable.UNBOX_CONVERSION;
+        } else if (boxedDest.isAssignableFrom(boxedSrc)) {
+            return CandidateExecutable.WIDEN_CONVERSION;
         }
+
+        return CandidateExecutable.NO_CONVERSION_POSSIBLE;
```
rcaudy, line 863:
```
        }

        if (src == NULL_CLASS) {
            return dest.isPrimitive() ? CandidateExecutable.NO_CONVERSION_POSSIBLE : CandidateExecutable.ALREADY_EQUAL;
        }

        Class<?> boxedDest = io.deephaven.util.type.TypeUtils.getBoxedType(dest);
        Class<?> boxedSrc = io.deephaven.util.type.TypeUtils.getBoxedType(src);

        if (boxedSrc == dest) {
            return CandidateExecutable.BOX_CONVERSION;
        }
        if (boxedDest == src) {
            return CandidateExecutable.UNBOX_CONVERSION;
        }
        if (boxedDest.isAssignableFrom(boxedSrc)) {
            return CandidateExecutable.WIDEN_CONVERSION;
        }

        return CandidateExecutable.NO_CONVERSION_POSSIBLE;
```

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -661,15 +658,33 @@ private String paramsTypesToString(Class<?>[] paramTypes) {
...
+            this.executable = executable;
+            this.cost = cost;
+        }
+
+        private static final int NO_CONVERSION_POSSIBLE = -1;
+        private static final int ALREADY_EQUAL = 0;
+        private static final int BOX_CONVERSION = 1;
+        private static final int WIDEN_CONVERSION = 2;
+        private static final int UNBOX_CONVERSION = 5; // may NPE, so is more expensive
+        private static final int VECTOR_CONVERSION = 10;
```
rcaudy, line 675: Vector conversion is arbitrarily expensive. Maybe make it 100?
I think unboxing goes before boxing, which goes before widening, to match the JVM. The NPE is specific to `Boolean`, we think. Maybe update comment accordingly.
```
        private static final int UNBOX_CONVERSION = 5; // may NPE, so is more expensive
        private static final int VECTOR_CONVERSION = 10;
```

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -698,88 +716,113 @@ private static <EXECUTABLE_TYPE extends Executable> void possiblyAddExecutable(
...
+                cost += CandidateExecutable.VECTOR_CONVERSION;
                 paramType = convertVector(paramType, parameterizedTypes[candidateParamTypes.length - 1] == null ? null
                         : parameterizedTypes[candidateParamTypes.length - 1][0]);
             }
 
+            int costToAssignVarArgs = costToAssignType(candidateParamType, paramType);
             boolean canAssignVarArgs = candidateParamTypes.length == paramTypes.length && paramType.isArray()
-                    && canAssignType(candidateParamType, paramType);
+                    && costToAssignVarArgs != CandidateExecutable.NO_CONVERSION_POSSIBLE;
+            if (canAssignVarArgs) {
```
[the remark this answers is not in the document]
nbauernfeind: As discussed in VC, this was confusing to begin with. It should not be possible for both to be true. The first check is for when there is a singular argument that matches the var-args type. The second check attempts to cast all of the remaining arguments to see if they will qualify as var-args. I don't believe it is possible for both checks to be true simultaneously.

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -846,23 +848,27 @@ private static int costToAssignFrom(final Class<?> dest, final Class<?> src) {
...
                 return CandidateExecutable.WIDEN_CONVERSION;
             }
             return CandidateExecutable.NO_CONVERSION_POSSIBLE;
-        } else if (!dest.isPrimitive() && src == NULL_CLASS) {
-            return CandidateExecutable.ALREADY_EQUAL;
         }
 
-        Class<?> boxedDest = io.deephaven.util.type.TypeUtils.getBoxedType(dest);
-        Class<?> boxedSrc = io.deephaven.util.type.TypeUtils.getBoxedType(src);
+        if (src == NULL_CLASS) {
```
rcaudy, line 858: Should this be an unboxing conversion for non-boolean? I guess that's a question of what other code does, downstream.

engine/table/src/test/java/io/deephaven/engine/table/impl/lang/TestQueryLanguageParser.java
```diff
@@ -929,7 +929,7 @@ public void testVariables() throws Exception {
 
         expression = "!myBoolean";
         resultExpression = "not(myBoolean)";
-        check(expression, resultExpression, Boolean.class, new String[] {"myBoolean"});
+        check(expression, resultExpression, boolean.class, new String[] {"myBoolean"});
```
rcaudy, line 932: Maybe we should add a `Boolean` flavor of these tests, too?

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -1041,13 +1041,16 @@ private Expression[] convertParameters(final Executable executable,
...
-                                .valueOf(argumentTypes[ai].getSimpleName().toUpperCase())),
-                        expressions[ai]);
+                expressions[ai] = new CastExpr(new PrimitiveType(PrimitiveType.Primitive
+                        .valueOf(argumentTypes[ai].getSimpleName().toUpperCase())), expressions[ai]);
             } else if (unboxArguments && argumentTypes[ai].isPrimitive() && !expressionTypes[ai].isPrimitive()) {
-                expressions[ai] = new MethodCallExpr(expressions[ai],
-                        argumentTypes[ai].getSimpleName() + "Value", new NodeList<>());
+                if (expressionTypes[ai] == NULL_CLASS) {
+                    expressions[ai] = new CastExpr(new PrimitiveType(PrimitiveType.Primitive
+                            .valueOf(argumentTypes[ai].getSimpleName().toUpperCase())), expressions[ai]);
```
rcaudy, line 1049: Couldn't this be a constant expression?

rcaudy (approved):

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -41,6 +41,7 @@
 import com.github.javaparser.ast.expr.IntegerLiteralExpr;
 import com.github.javaparser.ast.expr.LambdaExpr;
 import com.github.javaparser.ast.expr.LiteralExpr;
+import com.github.javaparser.ast.expr.LiteralStringValueExpr;
```
rcaudy, line 44: Unused import

rcaudy (approved):

PR #3532, opened 2023-03-11
Provide config service methods to java clients

### LINKED ISSUE #3528: Java Client doesn't have a ConfigService as it does other services
At this time, it appears that the ConfigService grpc is only for internal use, the API isnt exposed at all, except as a plain grpc call.

Instead, consider a ConfigService in io.deephaven.client.impl:
```java
public interface ConfigService {
  CompletableFuture<Map<String, ConfigValue>> getAuthenticationConstants();
  CompletableFuture<Map<String, ConfigValue>> getConfigurationConstants();
}
```

niloc132:
Refactors existing java-client grpc response observers into one shared implementation, so that all can be canceled.

Fixes #3528

### DIFF
--- a/java-client/session/src/main/java/io/deephaven/client/impl/SessionImpl.java
+++ b/java-client/session/src/main/java/io/deephaven/client/impl/SessionImpl.java
@@ -14,7 +14,6 @@ import io.deephaven.proto.backplane.grpc.ConfigurationConstantsRequest;
 import io.deephaven.proto.backplane.grpc.ConfigurationConstantsResponse;
 import io.deephaven.proto.backplane.grpc.DeleteTableRequest;
 import io.deephaven.proto.backplane.grpc.FetchObjectRequest;
-import io.deephaven.proto.backplane.grpc.FieldsChangeUpdate;
 import io.deephaven.proto.backplane.grpc.HandshakeRequest;
 import io.deephaven.proto.backplane.grpc.ListFieldsRequest;
 import io.deephaven.proto.backplane.grpc.ReleaseRequest;
@@ -23,10 +22,7 @@ import io.deephaven.proto.backplane.grpc.TypedTicket;
 import io.deephaven.proto.backplane.script.grpc.BindTableToVariableRequest;
 import io.deephaven.proto.backplane.script.grpc.ExecuteCommandRequest;
 import io.deephaven.proto.backplane.script.grpc.StartConsoleRequest;
-import io.deephaven.proto.backplane.script.grpc.StartConsoleResponse;
 import io.deephaven.proto.util.ExportTicketHelper;
-import io.grpc.stub.ClientCallStreamObserver;
-import io.grpc.stub.ClientResponseObserver;
 import io.grpc.stub.StreamObserver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -47,8 +43,6 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import java.util.function.BiConsumer;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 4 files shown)


java-client/session/src/main/java/io/deephaven/client/impl/UnaryGrpcFuture.java
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * ClientResponseObserver implementation that can map results into a CompletableFuture. To keep generics under control,
+ * has two factory methods, {@link #of(Object, BiConsumer, Function)} and {@link #ignoreResponse(Object, BiConsumer)}.
+ * 
+ * @param <ReqT> The type of the gRPC request
+ * @param <RespT> The type of the gRPC response
+ * @param <V> The type of the result to make available in the future
+ */
+public class UnaryGrpcFuture<ReqT, RespT, V> implements ClientResponseObserver<ReqT, RespT> {
```
devinrsmith, line 19: Is there value in making this public? (I know we are very loose in other places w/ public...)

java-client/session/src/main/java/io/deephaven/client/impl/UnaryGrpcFuture.java
```diff
@@ -0,0 +1,93 @@
...
+     * @param <V> The type of the result to make available in the future
+     * @return A CompletableFuture that will succeed, fail, or be canceled. If it succeeds, will contain a result from
+     *         the mapping function.
+     */
+    public static <ReqT, RespT, V> CompletableFuture<V> of(ReqT request, BiConsumer<ReqT, StreamObserver<RespT>> call,
+            Function<RespT, V> mapping) {
+        UnaryGrpcFuture<ReqT, RespT, V> observer = new UnaryGrpcFuture<>(mapping);
+        call.accept(request, observer);
+        return observer.future();
+    }
```
devinrsmith, line 38: What are your thoughts on having the callers always get `CompletableFuture<RespT>` back here (ie, no `mapping` arg), and they can call `.thenApply(...)` as appropriate?

devinrsmith (changes requested):
Generally nice abstraction. One big issue, `subscribeToFields` broken.

java-client/session/src/main/java/io/deephaven/client/impl/UnaryGrpcFuture.java
niloc132: I actually started off that way, as it seems nicer to be "fluent" like that.  The problem there is that if they return the result of `.thenApply()` or whatnot, then the caller is left with the downstream CF, and canceling that won't cancel the upstream one (so won't stop the grpc call).

java-client/session/src/main/java/io/deephaven/client/impl/UnaryGrpcFuture.java
niloc132: My assumption was that just as DeephavenChannel is very-public (presumably to support the various use cases that java-client doesnt have implemented?), we want this tool available for downstream users to make unary calls easy to build with CFs.

java-client/session/src/main/java/io/deephaven/client/impl/SessionImpl.java
```diff
@@ -257,17 +261,15 @@ public CompletableFuture<Void> deleteFromInputTable(HasTicketId destination, Has
...
 
     @Override
     public Cancel subscribeToFields(Listener listener) {
-        final ListFieldsRequest request = ListFieldsRequest.newBuilder().build();
-        final ListFieldsObserver observer = new ListFieldsObserver(listener);
-        bearerChannel.application().listFields(request, observer);
-        return observer;
+        CompletableFuture<Void> future = UnaryGrpcFuture
+                .ignoreResponse(ListFieldsRequest.getDefaultInstance(), channel().application()::listFields);
+        return () -> future.cancel(false);
```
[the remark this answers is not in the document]
niloc132: Agreed, I'll undo this.

java-client/session/src/main/java/io/deephaven/client/impl/UnaryGrpcFuture.java
devinrsmith: Good call. While the caller _could_ still add their own `.whenComplete(...)` and pass along the cancel, that's a bunch of extra machinery I agree we don't really want.

devinrsmith (approved):

abaranec (approved):

PR #3536, opened 2023-03-13
Deephaven-in-docker gradle wiring support for env vars

### LINKED ISSUE #3535: Lower the JVM max memory limit for running the Python client tests
When repeatedly running the Python client tests against the DH server in container on Mac, I ran into the OOM issue where the Docker is resourced at 8g maximum, and the JVM's Xmx=6g. @niloc132 and I debugged the issue together and concluded that 6g or 4g are way too high for the server and it is the cause of the OOM because Docker advertises the 8g to the container without taking into account other containers running at the same time and when JVM seeks to bump its heap memory size beyond what's available, Docker simply kills it. Nothing in the server logs tells the story and that silent crash could really leave the user confused. 

Based on the above observation and hypothesis, Colin thought it would be a good idea to lower the JVM max heap size to a much smaller number (1g seems to be enough considering the dataset size that the tests require). With that, the OOM can't be reproduced confirming the theory. This might have the additional effect of decreasing resource usage for parallel tests.

niloc132:
Enables tests to control the environment variables of a running deephaven server in docker, to configure the jvm or deephaven itself.

This patch also modifies py/go/cpp tests to only ask for a max of 512mb of heap, to avoid issues running in docker desktop.

Fixes #3535

### DIFF
--- a/buildSrc/src/main/groovy/io/deephaven/tools/docker/DeephavenInDockerExtension.groovy
+++ b/buildSrc/src/main/groovy/io/deephaven/tools/docker/DeephavenInDockerExtension.groovy
@@ -9,6 +9,7 @@ import com.bmuschko.gradle.docker.tasks.network.DockerRemoveNetwork
 import groovy.transform.CompileStatic
 import org.gradle.api.Project
 import org.gradle.api.Task
+import org.gradle.api.provider.MapProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.TaskProvider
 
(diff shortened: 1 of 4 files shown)


devinrsmith (approved):
LGTM.

As a general note: by default, JVM ergonomics optimizes for a different GC collector when the heap size is small. For the deephaven server, we are explicitly forcing non-ergonomic defaults via `JAVA_OPTS=-XX:+UseG1GC -XX:MaxGCPauseMillis=100 -XX:+UseStringDeduplication`. Not that I think there is pressing needs, but if we find ourselves needing to further optimize integration tests cases for small heap sizes, we may wish to override the server defaults and explicitly set `JAVA_OPTS` instead of `START_OPTS`.

niloc132:
Yeah, I deliberately set START_OPTS to let those settings continue to be used, as that is what our docker-compose example does as well, under the assumption that we _probably_ want tests to be tweaking the server no more than is obvious from examples.

I considered briefly just offering a list for jvm args, maybe a map for system properties, but it seemed not-too-dirty to just directly expose all env vars so we could decide when to set JAVA_OPTS, EXTRA_CLASSPATH as well as and when we need them.

jmao-denver (approved):
LGTM

PR #3542, opened 2023-03-14
Send CORS headers for browser clients using gRPC-web

### LINKED ISSUE #3525: Support CORS for grpc-web / websocket calls
As a developer, I want to be able to import the API dynamically into a script on a different URL. Currently, such http requests are blocked by CORS.

niloc132:
Fixes #3525

### DIFF
--- a/server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
+++ b/server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
@@ -3,12 +3,15 @@
  */
 package io.deephaven.server.jetty;
 
+import io.deephaven.server.browserstreaming.BrowserStreamInterceptor;
 import io.deephaven.server.runner.GrpcServer;
 import io.deephaven.ssl.config.CiphersIntermediate;
 import io.deephaven.ssl.config.ProtocolsIntermediate;
 import io.deephaven.ssl.config.SSLConfig;
 import io.deephaven.ssl.config.TrustJdk;
 import io.deephaven.ssl.config.impl.KickstartUtils;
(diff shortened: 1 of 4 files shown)


nbauernfeind (approved):

mofojed (commented):
So just allowing all origins, not allowing it to be configured? Seems like it'd be fine since we need to auth anyway.

vbabich (approved):

server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
```diff
@@ -96,6 +103,57 @@ public JettyBackedGrpcServer(
         // Add an extra filter to redirect from / to /ide/
         context.addFilter(HomeFilter.class, "/", EnumSet.noneOf(DispatcherType.class));
 
+        // If requested, permit CORS requests
+        FilterHolder holder = new FilterHolder(CrossOriginFilter.class);
```
devinrsmith, line 107: Are there considerations where one might _not_ want CORS support? If so, we could consider adding an optional configuration piece in `io.deephaven.server.jetty.JettyConfig`.

niloc132: Fair, though I can't think of any (we don't use browser cookies for anything here, so a malicious website using a user's browser to make calls on their behalf wouldnt really do anything).

My personal vote is that this land as is, then dig in more with parameterizing all of this for the march release (which DnD is almost inevitably going to move to for this release anyway...).

devinrsmith: LGTM

PR #3548, opened 2023-03-15
Allow using LivenessScope outside of the context of with: statement.

jcferretti:
Draft, not ready for review, discussing with Jianfeng over slack.


jmao-denver (approved):
LGTM

PR #3551, opened 2023-03-16
Apply spotless after quick-check failures

niloc132:
Automatically applies spotless after a quick check fails, and presents the diff as github PR comment suggestions for the author to fix.

Example pull request showing this in action (with the suggestion manually "unresolved" so that it is visible when looking at the changes): https://github.com/niloc132/deephaven-core/pull/4

image

### DIFF
--- a/.github/workflows/quick-ci.yml
+++ b/.github/workflows/quick-ci.yml
@@ -2,9 +2,7 @@ name: Quick CI
 
 on:
   pull_request:
-    branches: [ main ]
   push:
-    branches: [ 'main', 'check/**', 'release/v*' ]
 
 jobs:
   quick:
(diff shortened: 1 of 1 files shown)


.github/workflows/quick-ci.yml
```diff
@@ -48,3 +48,15 @@ jobs:
           name: quick-ci-jvm-err
           path: '**/*_pid*.log'
           if-no-files-found: ignore
+      - name: On failure, try to reformat and propose fixes
+        uses: burrunan/gradle-cache-action@v1
+        if: failure()
+        with:
+          job-id: spotless-formatting
+          arguments: spotlessJavaApply
```
devinrsmith, line 56: Any reason to prefer `spotlessJavaApply` instead of `spotlessApply`?

niloc132: Ah good call, that's better. I'm also adding updateProto to this as a separate step, so the changes are marked separately.

mofojed (commented):
Any reason not to merge this?

Side note - I prefer to format on save, any reason we don't on Community?

devinrsmith:
> Side note - I prefer to format on save, any reason we don't on Community?

@mofojed what do you mean "format on save"?

PR #3552, opened 2023-03-16
Rename snapshot() to to_arrow and add the new snapshot/When ops in Python client

### LINKED ISSUE #2454: Python client and server have naming conflicts for `Table.snapshot`
The python client and server both have `Table.snapshot` methods, but these methods are not the same.  This makes code developed for the client or server non-portable to the other.  This problem needs to be resolved by having distinct names for the distinct snapshot operations.

When this change is made, the `pyclient` `README.md` and the `pyclient/docs/source/examples.rst` must be changed to be consistent with the new naming.

jmao-denver:
Fixes #2454

### DIFF
--- a/py/client/tests/test_query.py
+++ b/py/client/tests/test_query.py
@@ -64,6 +64,28 @@ class QueryTestCase(BaseTestCase):
         result_table = query.exec()
         self.assertTrue(result_table.size == 10)
 
+    def test_snapshot(self):
+        test_table = self.session.time_table(period=1000000)
+        while test_table.snapshot().size < 100:
+            time.sleep(0.001)
+        query = self.session.query(test_table)
+        (query.update(formulas=["Col1 = i", "Col2 = i * 2"])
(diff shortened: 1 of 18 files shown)


py/client/tests/test_query.py
```diff
@@ -64,6 +64,28 @@ def test_update_by(self):
         result_table = query.exec()
         self.assertTrue(result_table.size == 10)
 
+    def test_snapshot(self):
+        test_table = self.session.time_table(period=1000000)
+        while test_table.snapshot().size < 100:
+            time.sleep(0.001)
```
chipkent, line 70: Are there any cases where we could accidentally hang forever?

chipkent (approved):

py/client/tests/test_query.py
jmao-denver: Not likely. Even when the server somehow falls into a dead-lock situation, the gRPC/network layer should eventually time out and exception out.

PR #3572, opened 2023-03-19
Fix a bug in EMA tests

jmao-denver:
Not sure how it happened - the flip of the 'if' branches, could be a mistake caused by distraction.


rcaudy (commented):
I don't understand. rt should never be bigger than t. I think you should just take the shared lock and assert that they're equal.

rcaudy (approved):

PR #3584, opened 2023-03-22
Go client authentication support

chipkent:
Support authentication in the go client.

### DIFF
--- a/go/internal/test_tools/test_tools.go
+++ b/go/internal/test_tools/test_tools.go
@@ -68,7 +68,7 @@ func CheckError(t *testing.T, msg string, err error) {
 }
 
 // GetHost returns the host to connect to for the tests.
-// By default it is localhost, but can be overriden by setting the DH_HOST environment variable.
+// By default it is localhost, but can be overridden by setting the DH_HOST environment variable.
 func GetHost() string {
 	host := os.Getenv("DH_HOST")
 	if host == "" {
@@ -79,12 +79,23 @@ func GetHost() string {
 }
 
 // GetPort returns the port to connect to for the tests.
-// By default it is 10000, but can be overriden by setting the DH_PORT environment variable.
+// By default it is 10000, but can be overridden by setting the DH_PORT environment variable.
 func GetPort() string {
 	port := os.Getenv("DH_PORT")
 	if port == "" {
 		return "10000"
 	} else {
 		return port
+	}
+}
+
+// GetAuth returns the auth string to connect to for the tests.
+// By default it is Anonymous, but can be overridden by setting the DH_AUTH environment variable.
+func GetAuth() string {
+	auth := os.Getenv("DH_AUTH")
+	if auth == "" {
+		return "Anonymous"
+	} else {
+		return auth
 	}
 }
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -20,10 +20,10 @@ import (
 	"sync"
 
 	apppb2 "github.com/deephaven/deephaven-core/go/internal/proto/application"
+	configpb2 "github.com/deephaven/deephaven-core/go/internal/proto/config"
 	consolepb2 "github.com/deephaven/deephaven-core/go/internal/proto/console"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
-	"google.golang.org/grpc/metadata"
 )
 
 // ErrClosedClient is returned as an error when trying to perform a network operation on a client that has been closed.
@@ -51,6 +51,7 @@ type Client struct {
 
 	appServiceClient apppb2.ApplicationServiceClient
 	ticketFact       ticketFactory
+	tokenMgr         *tokenManager
 }
 
 // NewClient starts a connection to a Deephaven server.
@@ -187,17 +188,6 @@ func (client *Client) Close() error {
 	return err
 }
 
-// withToken attaches the current session token to a context as metadata.
-func (client *Client) withToken(ctx context.Context) (context.Context, error) {
-	tok, err := client.getToken()
-
-	if err != nil {
-		return nil, err
-	}
-
-	return metadata.NewOutgoingContext(ctx, metadata.Pairs("authorization", string(tok))), nil
-}
-
 // RunScript executes a script on the deephaven server.
 //
 // The script language depends on the argument passed to WithConsole when creating the client.
@@ -207,7 +197,7 @@ func (client *Client) RunScript(ctx context.Context, script string) error {
 		return ErrNoConsole
 	}
 
-	ctx, err := client.consoleStub.client.withToken(ctx)
+	ctx, err := client.consoleStub.client.tokenMgr.withToken(ctx)
 	if err != nil {
 		return err
 	}
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -171,8 +171,6 @@ func (client *Client) Close() error {
 
 	client.isOpen = false
 
-	client.sessionStub.Close()
-
 	if client.grpcChannel != nil {
 		client.grpcChannel.Close()
 		client.grpcChannel = nil
--- a/go/pkg/client/client.go
+++ b/go/pkg/client/client.go
@@ -58,12 +58,20 @@ type Client struct {
 //
 // The client should be closed using Close() after it is done being used.
 //
-// Keepalive messages are sent automatically by the client to the server at a regular interval (~30 seconds)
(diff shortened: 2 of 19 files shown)


go/pkg/client/client.go
```diff
@@ -76,19 +85,31 @@ func NewClient(ctx context.Context, host string, port string, options ...ClientO
 
 	client.ticketFact = newTicketFactory()
 
-	client.sessionStub, err = newSessionStub(ctx, client)
+	client.flightStub, err = newFlightStub(client, host, port)
 	if err != nil {
 		client.Close()
```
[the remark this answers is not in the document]
chipkent: Defer doesn't work, since the client is being returned.  We don't want to close it in that case.  You only want to close in the error cases.  For those error cases, the client may be partially built, so the partially built client needs to be closed. ... which is going to make me check that there are not NPEs for a partial close.

jcferretti (approved):

go/pkg/client/tokens.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	timeout := time.Duration(maxTimeoutMillis/2) * time.Millisecond
+
+	token := &tokenResp{Token: tkn}
+	done := make(chan bool)
+	ticker := time.NewTicker(timeout)
+
+	go func() {
+		for {
```
kosak, line 147: In C++ we did things a little differently.  You may or may not decide they are important enough to do.

Cristian and I were quite worried about the case where a message might be lost, dropped, or massively delayed. Sending handshakes at a fixed interval, as you do here, is only reliable if they are all delivered, and delivered in a timely fashion.

To try to be more robust about this, we used the following approach instead.

1. We snoop *all* request/response traffic and we notice changes in auth token as they happen. This means when there is other traffic we don't bother sending a handshake at all, because we were able to snoop the token from some other response. We only do a handshake if there has been a `timeout`-sized period of no traffic at all.
2. The code basically has two states: normal (I have received some kind of response in the last timeout period) and handshaking (I've *not* received any kind of response in the last timeout period). When it enters the "handshaking" state, the code sends handshakes pretty aggressively (like once every 5 seconds). This allows the token refresh thing to be somewhat robust against network freezes or connection drops. But then, *any* response from the server (whether a normal response or a handshake response) tends to put the code back into the "normal" state, so in the common case you only send one handshake and then drop back to the normal state.
3. In order to account for the vagaries of round trip time and connection drops, the code keeps track of timeouts based on the local time when a (request or handshake) was *sent* rather than when a response was *received*. For example, if I sent a (request or handshake) at 12:00:00 and I receive a response at 12:00:20, I will do my timeout calculations based on when I sent the request, not when I received the response. In other words if the timeout is 5 minutes, I will consider going to handshake mode at 12:05:00, not 12:05:20.

Basically the C++ code follows this kind of pseudocode:

At any request
```
Note current local time
send the request
```

At any response to a request or a handshake
```
note the new auth header (maybe don't bother copying it if it hasn't changed)
reset the `handshake deadline` to the (time of the corresponding request) plus 5 minutes
```

The handshake thread:
```
forever:
  sleep until handshake deadline
  if current time < handshake deadline  # deadline has advanced in the meantime
     continue loop
 note current handshake request time
 send handshake message asynchronously
 set handshake deadline to now + 5 seconds
 # the above is pessimistic, because the response to this handshake message,
 # if received, will undergo standard auth header and deadline processing and
 # will advance the deadline forward ~5 minutes
```

go/pkg/client/client.go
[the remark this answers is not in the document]
kosak: Would it work to write code like this at the start
```
	workingClient := &Client{grpcChannel: grpcChannel, isOpen: true}
        defer {
            if workingClient != nil
                 workingClient.Close()
        }
	workingClient.suppressTableLeakWarning = opts.suppressTableLeakWarning
        etc etc
```

and then at the bottom

```
         client = workingClient
         workingClient = nil
         return client, nil
```

go/pkg/client/client.go
```diff
@@ -76,19 +85,31 @@ func NewClient(ctx context.Context, host string, port string, options ...ClientO
...
 		return nil, err
 	}
 
-	client.consoleStub, err = newConsoleStub(ctx, client, opts.scriptLanguage)
+	cfgClient := configpb2.NewConfigServiceClient(grpcChannel)
+
+	if auth == "" {
+		auth = "Anonymous"
```
kosak, line 97: Suggest moving magic string literals like this (and "http.session.durationMs") to some utility class so they can be defined in one place

go/internal/test_tools/test_tools.go
```diff
@@ -88,3 +88,14 @@ func GetPort() string {
 		return port
 	}
 }
+
+// GetAuth returns the auth string to connect to for the tests.
+// By default it is Anonymous, but can be overridden by setting the DH_AUTH environment variable.
+func GetAuth() string {
+	auth := os.Getenv("DH_AUTH")
+	if auth == "" {
+		return "Anonymous"
```
kosak, line 97: Suggest moving magic string literals like this (and "http.session.durationMs") to some utility class so they can be defined in one place

go/pkg/client/tokens.go
```diff
@@ -0,0 +1,186 @@
...
+		return nil, err
+	}
+
+	ac, err := cfg.GetConfigurationConstants(withAuthToken(ctx, tkn), &configpb2.ConfigurationConstantsRequest{})
+
+	if err != nil {
+		return nil, err
+	}
+
+	sessionDurationStr, ok := ac.ConfigValues["http.session.durationMs"]
```
kosak, line 128: Suggest moving magic string literals like this (and "Anonymous", elsewhere) to some utility class so they can be defined in one place

go/pkg/client/tokens.go
```diff
@@ -0,0 +1,186 @@
...
+
+	if !ok {
+		return nil, errors.New("server configuration constants do not contain: http.session.durationMs")
```
kosak, line 131:
```
"server configuration constants do not contain: " + MySpecialConstantsWhatever.durationWhatever
```

go/pkg/client/tokens.go
kosak: @jcferretti proposed (and I agree) that we should defer this work to a future PR.

go/pkg/client/client.go
```diff
@@ -43,54 +43,78 @@ type Client struct {
...
+// To see what authentication methods are available on the Deephaven server, navigate to: http://<host>:<port>/jsapi/authentication/.
+//
 // The option arguments can be used to specify other settings for the client.
 // See the With<XYZ> methods (e.g. WithConsole) for details on what options are available.
-func NewClient(ctx context.Context, host string, port string, options ...ClientOption) (*Client, error) {
+func NewClient(ctx context.Context, host string, port string, auth string, options ...ClientOption) (client *Client, err error) {
+	defer func() {
+		if err != nil && client != nil {
+			e := client.Close()
+			log.Println("Error when closing failed client: ", e)
```
kosak, line 78: It does not seem right to me that we unconditionally print a message here.  I would think in the common case, if we get here, `client.Close()` will actually succeed and e will be nil. Shouldn't this be

```
if e != nil {
  log.Println("blah")
}
```

go/internal/test_tools/test_tools.go
kosak, line 97: Should this be `Constants.DefaultAuth` (or whatever) rather than the literal string "Anonymous"?

kosak (commented):
couple more small things

go/internal/test_tools/test_tools.go
chipkent: I don't think this should be using the constant.  This is set to match the DH instance that is spun up to test against.  We have no guarantee that the instance will be using the default connection mechanism, so I'm not completely comfortable linking them.

chipkent: My opinion is weak on this, so if you think it should be changed, I will.

kosak: I'm not deeply invested in either choice. I assumed it was an oversight. If not, that's fine.

kosak (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2337
Conceptual: https://github.com/deephaven/deephaven.io/issues/2339
Reference: https://github.com/deephaven/deephaven.io/issues/2338
Blog: https://github.com/deephaven/deephaven.io/issues/2340

chipkent:
This PR will break example Go clients, and these clients are not being automatically run in our docs tests.

PR #3589, opened 2023-03-23
Migrate to flight auth

### LINKED ISSUE #3513: Python client should migrate to flight auth
The SessionService.NewSession grpc call is deprecated. Instead either use the FlightService.Handshake method (aka Flight Auth v1) or set the authorization header on the first request to get a bearer token to use on later requests.

Likewise, instead of SessionService.RefreshSessionToken, when the server's response includes another authorization header, use that for future requests.

See https://github.com/deephaven/deephaven-core/tree/main/authentication for more details.

jmao-denver:
Fixes #3513

### DIFF
--- a/py/client/pydeephaven/_config_service.py
+++ b/py/client/pydeephaven/_config_service.py
@@ -1,0 +1,22 @@
+#
+# Copyright (c) 2016-2023 Deephaven Data Labs and Patent Pending
+#
+from typing import Any, Dict
+
+from pydeephaven.dherror import DHError
+from pydeephaven.proto import config_pb2, config_pb2_grpc
+
+
+class ConfigService:
+    def __init__(self, session):
+        self.session = session
+        self._grpc_app_stub = config_pb2_grpc.ConfigServiceStub(session.grpc_channel)
+
+    def get_configuration_constants(self) -> Dict[str, Any]:
+        try:
+            response = self._grpc_app_stub.GetConfigurationConstants(config_pb2.ConfigurationConstantsRequest(),
+                                                                     metadata=self.session.grpc_metadata
+                                                                     )
+            return dict(response.config_values)
+        except Exception as e:
+            raise DHError("failed to get the configuration constants.") from e
--- a/py/client/pydeephaven/session.py
+++ b/py/client/pydeephaven/session.py
@@ -1,19 +1,22 @@
 #
-# Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+# Copyright (c) 2016-2023 Deephaven Data Labs and Patent Pending
 #
-
+import base64
 import os
 import threading
 from typing import List
 
-import pyarrow
-from bitstring import BitArray
-
 import grpc
+import pyarrow
+import pyarrow.flight as paflight
+from bitstring import BitArray
+from pyarrow import ArrowNotImplementedError
+from pyarrow._flight import ClientMiddlewareFactory, ClientMiddleware, ClientAuthHandler
 
-from pydeephaven._arrow_flight_service import ArrowFlightService
-from pydeephaven._console_service import ConsoleService
 from pydeephaven._app_service import AppService
+from pydeephaven._arrow_flight_service import ArrowFlightService
+from pydeephaven._config_service import ConfigService
+from pydeephaven._console_service import ConsoleService
 from pydeephaven._session_service import SessionService
 from pydeephaven._table_ops import TimeTableOp, EmptyTableOp, MergeTablesOp, FetchTableOp
 from pydeephaven._table_service import TableService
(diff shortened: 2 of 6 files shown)


py/client/pydeephaven/session.py
```diff
@@ -60,7 +115,15 @@ def __init__(self, host: str = None, port: int = None, never_timeout: bool = Tru
             self.port = int(os.environ.get("DH_PORT", 10000))
 
         self.is_connected = False
-        self.session_token = None
+
+        if auth_type == "Anonymous":
+            self._auth_token = auth_type
+        elif auth_type == "Basic":
+            auth_token_base64 = base64.b64encode(auth_token.encode("ascii")).decode("ascii")
```
chipkent, line 122: @niloc132 If the Go client should be split up into two args, it probably also needs to base64 encode...right?

py/client/pydeephaven/session.py
```diff
@@ -86,11 +152,12 @@ def __exit__(self, exc_type, exc_val, exc_tb):
     def tables(self):
         with self._r_lock:
             fields = self._fetch_fields()
-            return [field.field_name for field in fields if field.application_id == 'scope' and field.typed_ticket.type == 'Table']
+            return [field.field_name for field in fields if
+                    field.application_id == 'scope' and field.typed_ticket.type == 'Table']
 
     @property
     def grpc_metadata(self):
```
chipkent, line 159: None of these methods have type hints.  The client needs a major lookover.

py/client/pydeephaven/session.py
```diff
@@ -124,6 +191,13 @@ def app_service(self):
...
         return self._app_service
 
+    @property
+    def config_service(self):
+        if not self._config_service:
+            self._config_service = ConfigService(self)
+
+        return self._config_service
+
     def make_ticket(self, ticket_no=None):
```
chipkent, line 201: private?
The client API needs to look at all of the methods and confirm if they should be public or not.

I _think_ that the tickets are not exposed to users in the Go client.  Should they be in python?  ... again, something for another PR.

py/client/pydeephaven/session.py
```diff
@@ -153,12 +227,27 @@ def _fetch_fields(self):
 
     def _connect(self):
         with self._r_lock:
-            self.grpc_channel, self.session_token, self._timeout = self.session_service.connect()
-            self.is_connected = True
+            try:
+                self._flight_client = paflight.connect(location=(self.host, self.port), middleware=[
+                    _DhClientAuthMiddlewareFactory(self)])
+                self._auth_handler = _DhClientAuthHandler(self)
+                self._flight_client.authenticate(self._auth_handler)
```
chipkent, line 234: Is the initial auth only used once?  I think it is only being used once.  @niloc132 mentioned problems if it gets reused.

py/client/pydeephaven/session.py
```diff
@@ -169,9 +258,10 @@ def _keep_alive(self):
     def _refresh_token(self):
```
chipkent, line 258: See https://github.com/deephaven/deephaven-core/issues/3593.  @jcferretti can comment further.

py/client/pydeephaven/session.py
```diff
@@ -201,6 +291,7 @@ def close(self) -> None:
                 self.grpc_channel.close()
                 self.is_connected = False
                 self._last_ticket = 0
+                self._flight_client.close()
                 # self._executor.shutdown()
```
chipkent, line 295: needed?

py/client/pydeephaven/session.py
```diff
@@ -124,6 +191,13 @@ def app_service(self):
 
         return self._app_service
 
+    @property
+    def config_service(self):
```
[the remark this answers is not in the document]
chipkent: I see the same problem with other services.  It needs significant cleanup... probably in another PR.

py/client/pydeephaven/session.py
```diff
@@ -153,12 +227,27 @@ def _fetch_fields(self):
 
     def _connect(self):
         with self._r_lock:
```
chipkent, line 229: Where the lock is created, I would comment about what the lock is for.

chipkent (changes requested):

py/client/pydeephaven/_config_service.py
```diff
@@ -0,0 +1,21 @@
...
+from pydeephaven.dherror import DHError
+from pydeephaven.proto import config_pb2, config_pb2_grpc
+
+
+class ConfigService:
+    def __init__(self, session):
+        self.session = session
+        self._grpc_app_stub = config_pb2_grpc.ConfigServiceStub(session.grpc_channel)
+
+    def get_configuration_constants(self) -> dict:
```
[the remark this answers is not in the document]
jmao-denver: Yes, the key should be str, but the value, though it is str right now,  could be any type as the comment for the value field in config.proto explicitly made clear.

py/client/pydeephaven/session.py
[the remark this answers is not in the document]
jmao-denver: The original reason why these are not private and don't have docstrings is that 
1. they don't show up in the sphinx generated docs. 
2. the static checker in PyCharm would complain about accessing a private property in other modules

py/client/pydeephaven/session.py
jmao-denver: see the reason given above.

py/client/pydeephaven/session.py
jmao-denver: see response above.

py/client/pydeephaven/session.py
jmao-denver: yes.

py/client/pydeephaven/session.py
```diff
@@ -35,12 +83,19 @@ class Session:
...
 
         Args:
             host (str): the host name or IP address of the remote machine, default is 'localhost'
             port (int): the port number that Deephaven server is listening on, default is 10000
+            auth_type (str): the authentication type string, can be "Anonymous', 'Basic", or any custom-built
+                authenticator in the server, such as "io.deephaven.authentication.psk.PskAuthenticationHandler",
+                default is 'Anonymous'.
+            auth_token (str): the authentication token string. When auth_type is 'Basic', it must be
+                "user:password"; when auth_type is "Anonymous', it will be ignored; when auth_type is a custom-built
+                authenticator, it must conform to the specific requirement of the authenticator
```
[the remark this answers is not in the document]
niloc132: They syntax of the Authorization header is in two parts, the "scheme" and "parameters", see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization#syntax

See also
https://httpwg.org/specs/rfc9110.html#field.authorization
and
https://httpwg.org/specs/rfc9110.html#credentials

Sending it as one value is probably fine, but most use cases will tend to think of it as two, likely a constant for the kind of authentication they are using, then the specific token that identifies them.

py/client/pydeephaven/session.py
niloc132: Only the `Basic` type is expected/required to be base64 encoded, this isn't necessarily required for other formats. For example, if you base64 the already-hex encoded uuids that are used for bearer tokens, or the already-base64 encoded openId connect payloads, the server won't be able to read them.

chipkent (approved):

chipkent (dismissed):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2369
Conceptual: https://github.com/deephaven/deephaven.io/issues/2366
Reference: https://github.com/deephaven/deephaven.io/issues/2368
Blog: https://github.com/deephaven/deephaven.io/issues/2367

PR #3591, opened 2023-03-24
Generalize internal table loggers

arman-ddl:
* _Will change target branch to main after draft phase_
* _Replaces https://github.com/deephaven/deephaven-core/pull/3353_

## 03/30/2023

* Removed `TableLoggerWrapper` and am instead checking types from engine table logger interface clients
* Renamed `EngineTableLoggerProvider` to `EngineTableLoggers`

### Summary
* Created interfaces for loggers and changed `MemoryTableLogger` logic into implementation details; e.g. `QueryOperationPerformanceLogLoggerMemoryImpl` extends `MemoryTableLogger` and implements `QueryOperationPerformanceLogLogger` (`MemoryTableLogger` shouldn't be a wrapper because memory logger logic will not be needed for e.g. DnD implementation)
* Created a `TableLoggerWrapper` class with `instanceof` logic to handle the leftover coupling on memory logger methods
* DI via `EngineTableLoggerProvider`; defaults to memory loggers for DHC use cases
* Rearranged data retrieval like for `initialSize` and `processUniqueId` for a more consistent internal table logger factory interface

### Questions
* Is `TableLoggerWrapper` a sufficient bridge for memory logger coupling for this PR?
* Between `UpdatePerformanceTracker`, `ServerStateTracker`, and `MemoryTableLoggers` being private constructor singletons outside of `DeephavenApiServerComponent`'s dependency graph, I'm not sure whether I should or how to leverage Dagger; advice?
* Advice for better way to leverage `initialSize` and `processUniqueId` than current refactoring?


engine/table/src/main/java/io/deephaven/engine/table/impl/util/MemoryTableLoggers.java
```diff
@@ -24,12 +25,22 @@
 public class MemoryTableLoggers {
```
abaranec, line 25: I think we should rename this to something else.  PerformanceMetrics, or DeephavenMetrics or something like that

engine/table/src/main/java/io/deephaven/engine/table/impl/util/MemoryTableLoggers.java
```diff
@@ -41,49 +52,36 @@ public static MemoryTableLoggers getInstance() {
...
 
-    private final ProcessInfo processInfo;
-    private final MemoryTableLogger<QueryPerformanceLogLogger> qplLogger;
-    private final MemoryTableLogger<QueryOperationPerformanceLogLogger> qoplLogger;
-    private final MemoryTableLogger<ProcessInfoLogLogger> processInfoLogger;
-    private final MemoryTableLogger<ProcessMetricsLogLogger> processMetricsLogger;
+    private final MemoryTableLoggerWrapper<QueryPerformanceLogLogger> qplLogger;
+    private final MemoryTableLoggerWrapper<QueryOperationPerformanceLogLogger> qoplLogger;
+    private final MemoryTableLoggerWrapper<ProcessInfoLogLogger> processInfoLogger;
+    private final MemoryTableLoggerWrapper<ProcessMetricsLogLogger> processMetricsLogger;
```
abaranec, line 58:
```suggestion
    private final QueryPerformanceLogLogger qplLogger;
    private final QueryOperationPerformanceLogLogger qoplLogger;
    private final ProcessInfoLogLogger processInfoLogger;
    private final ProcessMetricsLogLogger processMetricsLogger;
```

abaranec: Don't use a wrapper, just use the interfaces directly.

engine/table/src/main/java/io/deephaven/engine/table/impl/util/ServerStateTracker.java
```diff
@@ -35,13 +36,12 @@ public static ServerStateTracker getInstance() {
 
     private final Logger logger;
 
-    private final MemoryTableLogger<ServerStateLog> processMemLogger;
+    private final MemoryTableLoggerWrapper<ServerStateLogLogger> processMemLogger;
```
abaranec, line 39: Don't use wrapper

engine/table/src/main/java/io/deephaven/engine/tablelogger/EngineTableLoggerProvider.java
```diff
@@ -0,0 +1,42 @@
+package io.deephaven.engine.tablelogger;
+
+import io.deephaven.engine.tablelogger.impl.memory.EngineTableLoggerMemoryImplFactory;
+
+/**
+ * Provides the factory for providing engine table loggers.
+ */
+public class EngineTableLoggerProvider {
```
abaranec, line 8: How about a better class name,  maybe MetricsLoggers so that the usage is less confusing

MetricLoggers.Factory. or EngineLoggers.Provider feel better than EngineTableLoggerProvider.Factory

engine/table/src/main/java/io/deephaven/engine/tablelogger/EngineTableLoggerProvider.java
```diff
@@ -0,0 +1,42 @@
...
+    private EngineTableLoggerProvider() {
+        throw new UnsupportedOperationException();
```
abaranec, line 10: There's no need for the UOE here

engine/table/src/main/java/io/deephaven/engine/tablelogger/EngineTableLoggerProvider.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public static void set(final Factory factory) {
+        EngineTableLoggerProvider.factory = factory;
+    }
+
+    /**
+     * Marker interface to associate engine table loggers with one another.
+     */
+    public interface EngineTableLogger {
```
abaranec, line 26: I think this is also unnecessary

engine/table/src/main/java/io/deephaven/engine/tablelogger/ProcessInfoLogLogger.java
```diff
@@ -1,104 +1,19 @@
...
-    protected ISetter createSetter() {
-        outstandingSetters.getAndIncrement();
-        return new DirectSetter();
-    }
-
-    public void log(final String id, final String type, final String key, final String value) throws IOException {
+/**
+ * Logs data that describes JVM parameters. This is useful to check a worker's configuration.
+ */
+public interface ProcessInfoLogLogger extends EngineTableLoggerProvider.EngineTableLogger {
```
abaranec, line 12: Drop the tagging interface on all of these

abaranec (commented):
Overall I think this is correct.  I've picked on the naming a bit and I think we should ditch the extra classes and interfaces for simplicity

engine/table/src/main/java/io/deephaven/engine/table/impl/util/MemoryTableLoggerWrapper.java
```diff
@@ -0,0 +1,42 @@
...
+import io.deephaven.engine.tablelogger.impl.memory.MemoryTableLogger;
+
+/**
+ * This class is an artifact of DHC reliance on the memory table logger implementation before engine table loggers were
+ * generalized. It handles runtime type checking for clients and throws {@link UnsupportedOperationException} where the
+ * previous memory table logger coupling fails.
+ * 
+ * @param <T> Engine table logger type in which can instance can be retrieved
+ */
+public class MemoryTableLoggerWrapper<T extends EngineTableLoggerProvider.EngineTableLogger> {
```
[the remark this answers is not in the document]
arman-ddl: Replaced by io.deephaven.engine.tablelogger.impl.memory.MemoryTableLogger#maybeGetQueryTable.

engine/table/src/main/java/io/deephaven/engine/tablelogger/EngineTableLoggerProvider.java
arman-ddl: Changed top level to EngineLoggers. Kept nested interface as Factory.

engine/table/src/main/java/io/deephaven/engine/tablelogger/EngineTableLoggerProvider.java
arman-ddl: I instinctually add that in my private constructors to block against a) bad developer usage and b) reflection. I will assume you are implying Deephaven convention prefers to leave it empty, so I will remove it.

engine/table/src/main/java/io/deephaven/engine/tablelogger/EngineTableLoggerProvider.java
arman-ddl: Actually, I think I will change it further to EngineTableLoggers for a bit more specificity.

arman-ddl:

devinrsmith:
These commits should all be rebased against main, unless we are planning to cut a patch release with them (in which case, we won't be targeting main as target branch).

abaranec (approved):

devinrsmith (approved):
A note for any other reviewers: it's much easier to view this changeset in IJ, or otherwise outside of GH, b/c GH isn't doing a good job with the diff wrt filename changes.

Nothing objectionable. I think your PR description is out of date. I don't see `TableLoggerWrapper` or DI wrt `EngineTableLoggerProvider`. I'm assuming DHE is planning to hook in early and call `io.deephaven.engine.tablelogger.EngineTableLoggers#set`? While I'd prefer to plumb through proper dagger injection, that would probably be a large task given the amount of places these interfaces are statically referenced.

jcferretti (approved):

PR #3598, opened 2023-03-24
Test to make sure JPY PyObject.getXXXValue() errors out immediately

### LINKED ISSUE #3597: Add  a JPY test case to verify a fix that the next JPY release will include
The fix targets the issue reported in https://github.com/jpy-consortium/jpy/issues/98

jmao-denver:
Fixes #3597

Linked to: https://github.com/jpy-consortium/jpy/pull/99

### DIFF
--- a/py/jpy-integration/src/javaToPython/java/io/deephaven/jpy/integration/ErrorOutTest.java
+++ b/py/jpy-integration/src/javaToPython/java/io/deephaven/jpy/integration/ErrorOutTest.java
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+ */
+package io.deephaven.jpy.integration;
+
+import io.deephaven.jpy.JpyModule;
+import io.deephaven.jpy.PythonTest;
+import org.jpy.PyInputMode;
+import org.jpy.PyObject;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Test;
+
+
+public class ErrorOutTest extends PythonTest {
+    // ----- error out -----
+    @Test(expected = RuntimeException.class)
+    @Ignore("Need to wait for the next JPY release to fix https://github.com/jpy-consortium/jpy/issues/98")
+    public void errorIntOutFromNone() {
+        String expr = "[None, []]";
+        PyObject in = PyObject.executeCode(expr, PyInputMode.EXPRESSION);
+        in.asList().get(0).getIntValue();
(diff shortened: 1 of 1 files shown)


chipkent (dismissed):

py/jpy-integration/src/javaToPython/java/io/deephaven/jpy/integration/ErrorOutTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+
+public class ErrorOutTest extends PythonTest {
+    // ----- error out -----
+    @Test(expected = RuntimeException.class)
+    @Ignore("Need to wait for the next JPY release to fix https://github.com/jpy-consortium/jpy/issues/98")
+    public void errorIntOutFromNone() {
+        String expr = "[None, []]";
+        PyObject in = PyObject.executeCode(expr, PyInputMode.EXPRESSION);
+        in.asList().get(0).getIntValue();
```
devinrsmith, line 23: I'd like to make sure we have tests that demonstrate getIntValue w/ python ints in the cases:

* 0-32 bits
* 33-64 bits
* >64 bits

Given the new code, I expect the first two cases to "work", and the last case to properly fail.

py/jpy-integration/src/javaToPython/java/io/deephaven/jpy/integration/ErrorOutTest.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Test(expected = RuntimeException.class)
+    @Ignore("Need to wait for the next JPY release to fix https://github.com/jpy-consortium/jpy/issues/98")
+    public void errorLongOutFromNone() {
+        String expr = "[None, []]";
+        PyObject in = PyObject.executeCode(expr, PyInputMode.EXPRESSION);
+        in.asList().get(0).getLongValue();
```
devinrsmith, line 31: I'd like to make sure we have tests that demonstrate getLongValue w/ python ints in the cases:

* 0-64 bits
* >64 bits

Given the new code, I expect the first case to work and the last case to properly fail.

devinrsmith:
Given the lossy nature of getIntValue(), we should audit all of our own uses of it and switch over to more precise calls:

```java
Math.toIntExact(valueIn.getLongValue())
```

or something equivalent to that (considering int, short, byte, and char consumer).

devinrsmith:
Follow-up https://github.com/deephaven/deephaven-core/issues/3716

devinrsmith (approved):

PR #3600, opened 2023-03-24
Adds is_refreshing support for parquet read

### LINKED ISSUE #3596: Parquet directory rerfesh service support
The partition table framework supports discovering and adding new partitions on-the-fly. The plumbing for this has not been done for parquet directories. Potentially relevant interfaces, `io.deephaven.engine.table.impl.locations.TableLocationProvider`, `io.deephaven.engine.table.impl.locations.impl.TableLocationKeyFinder`, `io.deephaven.engine.table.impl.locations.util.TableDataRefreshService`.

devinrsmith:
Additionally, applies black formatting to parquet.py.

Fixes #3596

### DIFF
--- a/extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
+++ b/extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
@@ -11,6 +11,8 @@ import io.deephaven.base.verify.Require;
 import io.deephaven.engine.table.ColumnDefinition;
 import io.deephaven.engine.table.Table;
 import io.deephaven.engine.table.TableDefinition;
+import io.deephaven.engine.table.impl.locations.util.TableDataRefreshService;
+import io.deephaven.engine.updategraph.UpdateGraphProcessor;
 import io.deephaven.engine.util.TableTools;
 import io.deephaven.vector.*;
 import io.deephaven.stringset.StringSet;
@@ -380,6 +382,9 @@ public class ParquetTools {
         final BasicFileAttributes sourceAttr = readAttributes(sourcePath);
         if (sourceAttr.isRegularFile()) {
             if (sourceFileName.endsWith(PARQUET_FILE_EXTENSION)) {
+                if (instructions.isRefreshing()) {
+                    throw new IllegalArgumentException("Unable to have a refreshing single parquet file");
+                }
                 final ParquetTableLocationKey tableLocationKey = new ParquetTableLocationKey(source, 0, null);
                 final Pair<List<ColumnDefinition<?>>, ParquetInstructions> schemaInfo = convertSchema(
                         tableLocationKey.getFileReader().getSchema(),
@@ -471,9 +476,14 @@ public class ParquetTools {
                 StandaloneTableKey.getInstance(),
                 locationKeyFinder,
                 new ParquetTableLocationFactory(readInstructions),
-                null);
-        return new PartitionAwareSourceTable(tableDefinition, "Read multiple parquet files with " + locationKeyFinder,
-                RegionedTableComponentFactoryImpl.INSTANCE, locationProvider, null);
+                readInstructions.isRefreshing() ? TableDataRefreshService.getSharedRefreshService() : null);
+        return new PartitionAwareSourceTable(
+                tableDefinition,
+                readInstructions.isRefreshing() ? "Read refreshing parquet files with " + locationKeyFinder
+                        : "Read multiple parquet files with " + locationKeyFinder,
+                RegionedTableComponentFactoryImpl.INSTANCE,
+                locationProvider,
+                readInstructions.isRefreshing() ? UpdateGraphProcessor.DEFAULT : null);
     }
 
     /**
@@ -487,11 +497,14 @@ public class ParquetTools {
     public static Table readPartitionedTableInferSchema(
             @NotNull final TableLocationKeyFinder<ParquetTableLocationKey> locationKeyFinder,
             @NotNull final ParquetInstructions readInstructions) {
-        final RecordingLocationKeyFinder<ParquetTableLocationKey> recordingLocationKeyFinder =
-                new RecordingLocationKeyFinder<>();
-        locationKeyFinder.findKeys(recordingLocationKeyFinder);
-        final List<ParquetTableLocationKey> foundKeys = recordingLocationKeyFinder.getRecordedKeys();
+        final RecordingLocationKeyFinder<ParquetTableLocationKey> initialKeys = new RecordingLocationKeyFinder<>();
+        locationKeyFinder.findKeys(initialKeys);
+        final List<ParquetTableLocationKey> foundKeys = initialKeys.getRecordedKeys();
         if (foundKeys.isEmpty()) {
+            if (readInstructions.isRefreshing()) {
+                throw new IllegalArgumentException(
+                        "Unable to create refreshing partitioned parquet table when there are no initial parquet files");
+            }
             return TableTools.emptyTable(0);
         }
         // TODO (https://github.com/deephaven/deephaven-core/issues/877): Support schema merge when discovering multiple
@@ -513,8 +526,8 @@ public class ParquetTools {
                     getUnboxedTypeIfBoxed(partitionValue.getClass()), null, ColumnDefinition.ColumnType.Partitioning));
         }
         allColumns.addAll(schemaInfo.getFirst());
-        return readPartitionedTable(recordingLocationKeyFinder, schemaInfo.getSecond(),
-                TableDefinition.of(allColumns));
+        return readPartitionedTable(readInstructions.isRefreshing() ? locationKeyFinder : initialKeys,
+                schemaInfo.getSecond(), TableDefinition.of(allColumns));
     }
 
     /**
--- a/py/server/deephaven/parquet.py
+++ b/py/server/deephaven/parquet.py
@@ -15,14 +15,17 @@ from deephaven.table import Table
 
 _JParquetTools = jpy.get_type("io.deephaven.parquet.table.ParquetTools")
 _JFile = jpy.get_type("java.io.File")
-_JCompressionCodecName = jpy.get_type("org.apache.parquet.hadoop.metadata.CompressionCodecName")
+_JCompressionCodecName = jpy.get_type(
+    "org.apache.parquet.hadoop.metadata.CompressionCodecName"
+)
 _JParquetInstructions = jpy.get_type("io.deephaven.parquet.table.ParquetInstructions")
 _JTableDefinition = jpy.get_type("io.deephaven.engine.table.TableDefinition")
 
 
 @dataclass
 class ColumnInstruction:
-    """  This class specifies the instructions for reading/writing a Parquet column. """
+    """This class specifies the instructions for reading/writing a Parquet column."""
+
     column_name: str = None
     parquet_column_name: str = None
     codec_name: str = None
@@ -30,10 +33,25 @@ class ColumnInstruction:
     use_dictionary: bool = False
 
 
-def _build_parquet_instructions(col_instructions: List[ColumnInstruction] = None, compression_codec_name: str = None,
-                                max_dictionary_keys: int = None, is_legacy_parquet: bool = False,
-                                for_read: bool = True):
-    if not any([col_instructions, compression_codec_name, max_dictionary_keys, is_legacy_parquet]):
+def _build_parquet_instructions(
+    col_instructions: List[ColumnInstruction] = None,
+    compression_codec_name: str = None,
+    max_dictionary_keys: int = None,
+    is_legacy_parquet: bool = None,
+    target_page_size: int = None,
+    is_refreshing: bool = None,
+    for_read: bool = True,
+):
+    if not any(
+        [
+            col_instructions,
+            compression_codec_name,
+            max_dictionary_keys is not None,
+            is_legacy_parquet is not None,
+            target_page_size is not None,
+            is_refreshing is not None,
+        ]
+    ):
         return None
 
     builder = _JParquetInstructions.builder()
@@ -56,19 +74,31 @@ def _build_parquet_instructions(col_instructions: List[ColumnInstruction] = None
     if max_dictionary_keys is not None:
         builder.setMaximumDictionaryKeys(max_dictionary_keys)
 
-    if is_legacy_parquet:
-        builder.setIsLegacyParquet(True)
+    if is_legacy_parquet is not None:
+        builder.setIsLegacyParquet(is_legacy_parquet)
+
+    if target_page_size is not None:
+        builder.setTargetPageSize(target_page_size)
+
+    if is_refreshing is not None:
+        builder.setIsRefreshing(is_refreshing)
 
     return builder.build()
 
 
-def read(path: str, col_instructions: List[ColumnInstruction] = None, is_legacy_parquet: bool = False) -> Table:
-    """ Reads in a table from a single parquet, metadata file, or directory with recognized layout.
+def read(
+    path: str,
+    col_instructions: List[ColumnInstruction] = None,
+    is_legacy_parquet: bool = None,
+    is_refreshing: bool = None,
+) -> Table:
+    """Reads in a table from a single parquet, metadata file, or directory with recognized layout.
 
     Args:
         path (str): the file or directory to examine
         col_instructions (List[ColumnInstruction]): instructions for customizations while reading
         is_legacy_parquet (bool): if the parquet data is legacy
+        is_refreshing (bool): if the parquet data represents a refreshing source
 
     Returns:
         a table
@@ -78,9 +108,12 @@ def read(path: str, col_instructions: List[ColumnInstruction] = None, is_legacy_
     """
 
     try:
-        read_instructions = _build_parquet_instructions(col_instructions=col_instructions,
-                                                        is_legacy_parquet=is_legacy_parquet,
-                                                        for_read=True)
+        read_instructions = _build_parquet_instructions(
+            col_instructions=col_instructions,
+            is_legacy_parquet=is_legacy_parquet,
+            is_refreshing=is_refreshing,
+            for_read=True,
+        )
 
         if read_instructions:
             return Table(j_table=_JParquetTools.readTable(path, read_instructions))
@@ -95,7 +128,7 @@ def _j_file_array(paths: List[str]):
 
 
 def delete(path: str) -> None:
-    """ Deletes a Parquet table on disk.
+    """Deletes a Parquet table on disk.
 
     Args:
         path (str): path to delete
@@ -109,10 +142,15 @@ def delete(path: str) -> None:
         raise DHError(e, f"failed to delete a parquet table: {path} on disk.") from e
 
 
-def write(table: Table, path: str, col_definitions: List[Column] = None,
-          col_instructions: List[ColumnInstruction] = None, compression_codec_name: str = None,
-          max_dictionary_keys: int = None) -> None:
-    """ Write a table to a Parquet file.
+def write(
+    table: Table,
+    path: str,
+    col_definitions: List[Column] = None,
+    col_instructions: List[ColumnInstruction] = None,
+    compression_codec_name: str = None,
+    max_dictionary_keys: int = None,
+) -> None:
+    """Write a table to a Parquet file.
 
     Args:
         table (Table): the source table
@@ -128,33 +166,47 @@ def write(table: Table, path: str, col_definitions: List[Column] = None,
         DHError
     """
     try:
-        write_instructions = _build_parquet_instructions(col_instructions=col_instructions,
-                                                         compression_codec_name=compression_codec_name,
-                                                         max_dictionary_keys=max_dictionary_keys,
-                                                         for_read=False)
+        write_instructions = _build_parquet_instructions(
+            col_instructions=col_instructions,
+            compression_codec_name=compression_codec_name,
+            max_dictionary_keys=max_dictionary_keys,
+            for_read=False,
+        )
 
         table_definition = None
         if col_definitions is not None:
-            table_definition = _JTableDefinition.of([col.j_column_definition for col in col_definitions])
+            table_definition = _JTableDefinition.of(
+                [col.j_column_definition for col in col_definitions]
+            )
 
         if table_definition:
             if write_instructions:
-                _JParquetTools.writeTable(table.j_table, path, table_definition, write_instructions)
+                _JParquetTools.writeTable(
+                    table.j_table, path, table_definition, write_instructions
+                )
             else:
                 _JParquetTools.writeTable(table.j_table, _JFile(path), table_definition)
         else:
             if write_instructions:
-                _JParquetTools.writeTable(table.j_table, _JFile(path), write_instructions)
+                _JParquetTools.writeTable(
+                    table.j_table, _JFile(path), write_instructions
+                )
             else:
                 _JParquetTools.writeTable(table.j_table, path)
     except Exception as e:
         raise DHError(e, "failed to write to parquet data.") from e
 
 
-def batch_write(tables: List[Table], paths: List[str], col_definitions: List[Column],
-                col_instructions: List[ColumnInstruction] = None, compression_codec_name: str = None,
-                max_dictionary_keys: int = None, grouping_cols: List[str] = None):
-    """ Writes tables to disk in parquet format to a supplied set of paths.
+def batch_write(
+    tables: List[Table],
+    paths: List[str],
+    col_definitions: List[Column],
+    col_instructions: List[ColumnInstruction] = None,
+    compression_codec_name: str = None,
+    max_dictionary_keys: int = None,
+    grouping_cols: List[str] = None,
+):
+    """Writes tables to disk in parquet format to a supplied set of paths.
 
     If you specify grouping columns, there must already be grouping information for those columns in the sources.
     This can be accomplished with .groupBy(<grouping columns>).ungroup() or .sort(<grouping column>).
@@ -176,18 +228,28 @@ def batch_write(tables: List[Table], paths: List[str], col_definitions: List[Col
         DHError
     """
     try:
-        write_instructions = _build_parquet_instructions(col_instructions=col_instructions,
-                                                         compression_codec_name=compression_codec_name,
-                                                         max_dictionary_keys=max_dictionary_keys,
-                                                         for_read=False)
+        write_instructions = _build_parquet_instructions(
+            col_instructions=col_instructions,
+            compression_codec_name=compression_codec_name,
+            max_dictionary_keys=max_dictionary_keys,
+            for_read=False,
+        )
 
-        table_definition = _JTableDefinition.of([col.j_column_definition for col in col_definitions])
+        table_definition = _JTableDefinition.of(
+            [col.j_column_definition for col in col_definitions]
+        )
 
         if grouping_cols:
-            _JParquetTools.writeParquetTables([t.j_table for t in tables], table_definition, write_instructions,
-                                              _j_file_array(paths), grouping_cols)
+            _JParquetTools.writeParquetTables(
+                [t.j_table for t in tables],
+                table_definition,
+                write_instructions,
+                _j_file_array(paths),
+                grouping_cols,
+            )
         else:
-            _JParquetTools.writeTables([t.j_table for t in tables], table_definition,
-                                       _j_file_array(paths))
+            _JParquetTools.writeTables(
+                [t.j_table for t in tables], table_definition, _j_file_array(paths)
+            )
     except Exception as e:
         raise DHError(e, "write multiple tables to parquet data failed.") from e
--- a/extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetInstructions.java
+++ b/extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetInstructions.java
@@ -64,6 +64,7 @@ public abstract class ParquetInstructions implements ColumnToCodecMappings {
 
     private static final int MIN_DEFAULT_PAGE_SIZE = 64 << 10;
     private static volatile int defaultTargetPageSize = 1 << 20;
+    private static final boolean DEFAULT_IS_REFRESHING = false;
 
     /**
      * Set the default target page size (in bytes) used to section rows of data into pages during column writing. This
@@ -124,6 +125,11 @@ public abstract class ParquetInstructions implements ColumnToCodecMappings {
 
     public abstract int getTargetPageSize();
 
+    /**
+     * @return if the data source is refreshing
+     */
+    public abstract boolean isRefreshing();
+
     @VisibleForTesting
     public static boolean sameColumnNamesAndCodecMappings(final ParquetInstructions i1, final ParquetInstructions i2) {
         if (i1 == EMPTY) {
@@ -182,6 +188,11 @@ public abstract class ParquetInstructions implements ColumnToCodecMappings {
         @Override
         public int getTargetPageSize() {
             return defaultTargetPageSize;
+        }
+
+        @Override
+        public boolean isRefreshing() {
+            return DEFAULT_IS_REFRESHING;
         }
     };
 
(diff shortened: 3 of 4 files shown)


extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -380,6 +382,9 @@ private static Table readTableInternal(@NotNull final File source,
         final BasicFileAttributes sourceAttr = readAttributes(sourcePath);
         if (sourceAttr.isRegularFile()) {
             if (sourceFileName.endsWith(PARQUET_FILE_EXTENSION)) {
+                if (instructions.isRefreshing()) {
+                    throw new IllegalArgumentException("Unable to have a refreshing single parquet file");
```
rcaudy, line 386: Is this worth throwing an exception for? We can just ignore the setting.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -471,9 +476,14 @@ public static Table readPartitionedTable(
...
                 locationKeyFinder,
                 new ParquetTableLocationFactory(readInstructions),
-                null);
-        return new PartitionAwareSourceTable(tableDefinition, "Read multiple parquet files with " + locationKeyFinder,
-                RegionedTableComponentFactoryImpl.INSTANCE, locationProvider, null);
+                readInstructions.isRefreshing() ? TableDataRefreshService.getSharedRefreshService() : null);
+        return new PartitionAwareSourceTable(
+                tableDefinition,
+                readInstructions.isRefreshing() ? "Read refreshing parquet files with " + locationKeyFinder
+                        : "Read multiple parquet files with " + locationKeyFinder,
```
rcaudy, line 483: Just a readability preference:
```suggestion
                readInstructions.isRefreshing()
                        ? "Read refreshing parquet files with " + locationKeyFinder
                        : "Read multiple parquet files with " + locationKeyFinder,
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -487,11 +497,14 @@ public static Table readPartitionedTable(
...
-                new RecordingLocationKeyFinder<>();
-        locationKeyFinder.findKeys(recordingLocationKeyFinder);
-        final List<ParquetTableLocationKey> foundKeys = recordingLocationKeyFinder.getRecordedKeys();
+        final RecordingLocationKeyFinder<ParquetTableLocationKey> initialKeys = new RecordingLocationKeyFinder<>();
+        locationKeyFinder.findKeys(initialKeys);
+        final List<ParquetTableLocationKey> foundKeys = initialKeys.getRecordedKeys();
         if (foundKeys.isEmpty()) {
+            if (readInstructions.isRefreshing()) {
+                throw new IllegalArgumentException(
+                        "Unable to create refreshing partitioned parquet table when there are no initial parquet files");
```
rcaudy, line 506:
```suggestion
                        "Unable to infer schema for a refreshing partitioned parquet table when there are no initial parquet files");
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -513,8 +526,8 @@ public static Table readPartitionedTableInferSchema(
                     getUnboxedTypeIfBoxed(partitionValue.getClass()), null, ColumnDefinition.ColumnType.Partitioning));
         }
         allColumns.addAll(schemaInfo.getFirst());
-        return readPartitionedTable(recordingLocationKeyFinder, schemaInfo.getSecond(),
-                TableDefinition.of(allColumns));
+        return readPartitionedTable(readInstructions.isRefreshing() ? locationKeyFinder : initialKeys,
```
rcaudy, line 529: This is a little unfortunate if it causes us to scan for keys an extra time initially. I suppose it depends on the key finder implementation from a performance standpoint.

rcaudy (commented):
I think we need additional changes:
* `io.deephaven.parquet.table.layout.KeyValuePartitionLayout` is inferring partitioning column types from the set of values encountered. This will create issues if we change our mind about types on a subsequent scan. We should consider whether we can look for this scenario and provide a better error message.
* `io.deephaven.parquet.table.layout.ParquetKeyValuePartitionedLayout` and `io.deephaven.parquet.table.layout.ParquetFlatPartitionedLayout` should probably record the location keys they have created in a map by path, in order to avoid recreating them. This will reduce some of the costs associated with repeat scans.
* We should check for trailing magic numbers and refuse to create `ParquetTableLocationKey` instances for incomplete files.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
devinrsmith: Will do.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
devinrsmith: I would think that the user is expecting a ticking partition table whenever they set `instructions.isRefreshing()`. 

I'm not the biggest fan of how `readTable` branches to all these different paths depending on if it's a file, or a directory, or if there is a file in a directory, etc... but it _does_ make it a convenient entrypoint...

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
devinrsmith: I think some of these interfaces could be improved. For example, I think there is room for the user to explicitly provide the schema instead of inferring it from the first file. 

This situation could also be improved by updating or changing TableLocationKeyFinder - we shouldn't need to read all of the table locations if we only want just the first one (for inferring schema).

devinrsmith:
> io.deephaven.parquet.table.layout.KeyValuePartitionLayout is inferring partitioning column types from the set of values encountered. This will create issues if we change our mind about types on a subsequent scan. We should consider whether we can look for this scenario and provide a better error message.

I'm assuming that all TLKs found by an instance of  `tableLocationKeyFinder.findKeys(...)` should have consistent `TableLocationKey#getPartitionKeys`?

> io.deephaven.parquet.table.layout.ParquetKeyValuePartitionedLayout and io.deephaven.parquet.table.layout.ParquetFlatPartitionedLayout should probably record the location keys they have created in a map by path, in order to avoid recreating them. This will reduce some of the costs associated with repeat scans.

I wish there was a better way to do this instead of `TableLocationKeyFinder` - it seems like we could have a listener based interface or some such that automatically produces / notifies when a new TableLocationKey is created / found. (In the case of file-based TableLocationKey, could be based off of java.nio.file.WatchService... of course, you _could_ still do a polling based impl w/ a map... regardless, the caller-based `TableLocationKeyFinder#findKeys` seems a bit funky to me.)

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
rcaudy: This `readTable` is intended to be "easy to use". It has received a lot of positive feedback. The explicit, "tell us exactly what to do" option still exists in the public interface.

I suppose I can get behind your exception. I just think instructions tend to get copy-pasted a lot.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
rcaudy: I think we already have an exposed public method that allows the user to specify. I also think it's not that easy for a user to do.

The key finders were built for that static use case. I agree that we could have a richer interface (`findFirstKey`, `findNewKeys(Set)`).

rcaudy:
> > io.deephaven.parquet.table.layout.KeyValuePartitionLayout is inferring partitioning column types from the set of values encountered. This will create issues if we change our mind about types on a subsequent scan. We should consider whether we can look for this scenario and provide a better error message.
> 
> I'm assuming that all TLKs found by an instance of `tableLocationKeyFinder.findKeys(...)` should have consistent `TableLocationKey#getPartitionKeys`?
> 
> > io.deephaven.parquet.table.layout.ParquetKeyValuePartitionedLayout and io.deephaven.parquet.table.layout.ParquetFlatPartitionedLayout should probably record the location keys they have created in a map by path, in order to avoid recreating them. This will reduce some of the costs associated with repeat scans.
> 
> I wish there was a better way to do this instead of `TableLocationKeyFinder` - it seems like we could have a listener based interface or some such that automatically produces / notifies when a new TableLocationKey is created / found. (In the case of file-based TableLocationKey, could be based off of java.nio.file.WatchService... of course, you _could_ still do a polling based impl w/ a map... regardless, the caller-based `TableLocationKeyFinder#findKeys` seems a bit funky to me.)

Yes, we require identical partition layouts across TLKs for the same table. No way for the system to work properly without that, the partitions are columns of the data.

The deeper interfaces expect a `TableLocationProvider` (TLP) to support a listener model, and the typical listener then pushes into a buffer object that allows polling by the "real" engine constructs. The `PollingTableLocationProvider` implementation is much simpler than a lot of the other refreshing TLPs that we don't have in community right now. `WatchService`, in my opinion, is kind of a hot mess; it's lossy, with significant implementation differences across OSes. We can absolutely implement per-layout alternative TLPs that can discover parquet files, but it's likely a premature optimization. In one of my other replies I suggested something like `findMoreKeys(Map)`.

rcaudy:
> I'm very surprised that we are washing through string CSV building for `io.deephaven.parquet.table.layout.KeyValuePartitionLayout`; as such, I think any "costs associated with repeat scans" will be better focused on cleaning that up first?

No way. I'm very, very happy with using the CSV parser here. It's basically perfect for the purpose, since it supports columnar string->type inference and conversion out of the box. I would hate to have to build that from scratch, and I like standardizing on the CSV parser's decisions. The only reasonable alternative would be to show one column at a time to some lower layer of the parser to avoid constructing the CSV text.

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/util/ExecutorTableDataRefreshService.java
```diff
@@ -87,6 +87,8 @@ private void doRefresh() {
                 refresh();
             } catch (TableDataException e) {
                 subscriptionAggregator.activationFailed(this, e);
+            } catch (Throwable t) {
```
rcaudy, line 90: Are you sure we want to catch and swallow `Error` here? Maybe better if it was `Exception`?

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/TableLocationKeyFinder.java
```diff
@@ -14,6 +14,17 @@
...
+    /**
+     * May create a new {@link TableLocationKeyFinder} with additional safety checks.
+     *
+     * @param finder the finder
+     * @return the finder, potentially with additionaly safety checks
+     * @param <TLK> the table location key type
+     */
+    static <TLK extends TableLocationKey> TableLocationKeyFinder<TLK> safetyCheck(TableLocationKeyFinder<TLK> finder) {
+        return new TableLocationKeySafetyImpl<>(finder);
+    }
```
rcaudy, line 26: Some JavaDoc cleanup:
```suggestion
    /**
     * May create a new TableLocationKeyFinder with additional safety checks.
     *
     * @param finder The finder
     * @return The finder, potentially with additional safety checks added
     * @param <TLK> The {@link TableLocationKey} type
     */
    static <TLK extends TableLocationKey> TableLocationKeyFinder<TLK> safetyCheck(
            @NotNull TableLocationKeyFinder<TLK> finder) {
        return new TableLocationKeySafetyImpl<>(finder);
    }
```

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/TableLocationKeySafetyImpl.java
```diff
@@ -0,0 +1,65 @@
...
+                        "TableLocationKeyFinder %s has produced a duplicate TableLocationKey %s.", finder, tlk));
+            }
+            consumer.accept(tlk);
+        }
+
+        public void finished() {
+            for (TLK tlk : tlks) {
+                if (!seen.contains(tlk)) {
+                    throw new TableDataException(
+                            String.format("TableLocationKeyFinder %s has remove a previously seen TableLocationKey %s.",
```
rcaudy, line 58:
```suggestion
                            String.format("TableLocationKeyFinder %s has removed a previously seen TableLocationKey %s.",
```

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/TableLocationKeySafetyImpl.java
```diff
@@ -0,0 +1,65 @@
...
+            if (!seen.add(tlk)) {
+                throw new TableDataException(String.format(
+                        "TableLocationKeyFinder %s has produced a duplicate TableLocationKey %s.", finder, tlk));
+            }
+            consumer.accept(tlk);
+        }
+
+        public void finished() {
+            for (TLK tlk : tlks) {
+                if (!seen.contains(tlk)) {
```
rcaudy, line 56: This is likely not the best approach. Data was removed, and we observed it, but the downstream Table may not care, as it may have filtered the removed partition by key. We'll eventually blow up if we try to interact with the partition further. Unfortunately, we won't poll the size or anything, since Parquet files don't grow. Note that all bets are off if files are replaced.

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/TableLocationKeySafetyImpl.java
```diff
@@ -0,0 +1,65 @@
...
+
+        public Checked(Consumer<TLK> consumer) {
+            this.consumer = Objects.requireNonNull(consumer);
+            this.seen = new HashSet<>();
+        }
+
+        @Override
+        public void accept(TLK tlk) {
+            if (partitionsKeys == null) {
+                partitionsKeys = new HashSet<>(tlk.getPartitionKeys());
```
rcaudy, line 41: Use ` LinkedHashSet`. Ordering matters.

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/TableLocationKeySafetyImpl.java
```diff
@@ -0,0 +1,65 @@
...
+import io.deephaven.engine.table.impl.locations.TableDataException;
+import io.deephaven.engine.table.impl.locations.TableLocationKey;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Consumer;
+
+final class TableLocationKeySafetyImpl<TLK extends TableLocationKey> implements TableLocationKeyFinder<TLK> {
```
rcaudy, line 12: I'm a little bit skeptical of this whole endeavor. Also, it needs JavaDoc. I think we might be better off checking the partition safety downstream, say, in the `PollingTableLocationProvider`.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileReader.java
```diff
@@ -27,6 +27,13 @@ public class ParquetFileReader {
     private static final int FOOTER_LENGTH_SIZE = 4;
     private static final String MAGIC_STR = "PAR1";
     static final byte[] MAGIC = MAGIC_STR.getBytes(StandardCharsets.US_ASCII);
+
+    public static class ParquetFileReaderException extends RuntimeException {
```
rcaudy, line 31: This would be better as a top-level class. Nested exception classes are very annoying.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetFileHelper.java
```diff
@@ -0,0 +1,20 @@
...
+import java.nio.file.PathMatcher;
+
+import static io.deephaven.parquet.table.ParquetTableWriter.PARQUET_FILE_EXTENSION;
+
+final class ParquetFileHelper {
+
+    private static final PathMatcher MATCHER;
+
+    static {
+        MATCHER = FileSystems.getDefault().getPathMatcher(String.format("glob:*%s", PARQUET_FILE_EXTENSION));
```
rcaudy, line 14: This is a whole lot of overkill to check that a filename ends in the right extension. You've turned something that allocates a String into something that allocates the same String and then does regex match, at the cost of additional allocation. At least, that's what `sun.nio.fs.UnixFileSystem#getPathMatcher` looks like.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetFlatPartitionedLayout.java
```diff
@@ -5,42 +5,61 @@
...
     public String toString() {
         return ParquetFlatPartitionedLayout.class.getSimpleName() + '[' + tableRootDirectory + ']';
     }
 
     @Override
-    public void findKeys(@NotNull final Consumer<ParquetTableLocationKey> locationKeyObserver) {
-        try (final DirectoryStream<Path> parquetFileStream =
-                Files.newDirectoryStream(tableRootDirectory.toPath(), "*" + PARQUET_FILE_EXTENSION)) {
+    public synchronized void findKeys(@NotNull final Consumer<ParquetTableLocationKey> locationKeyObserver) {
+        try (final DirectoryStream<Path> parquetFileStream = FileSystems.getDefault().provider()
```
rcaudy, line 49: Why do you prefer `FileSystems.getDefault().provider()` to `Files`? I think that's an incorrect choice, we should be using the directories FS, not the default.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetFlatPartitionedLayout.java
```diff
@@ -5,42 +5,61 @@
...
+                .newDirectoryStream(tableRootDirectory.toPath(), ParquetFileHelper::fileNameMatches)) {
             for (final Path parquetFilePath : parquetFileStream) {
-                locationKeyObserver.accept(new ParquetTableLocationKey(parquetFilePath.toFile(), 0, null));
+                ParquetTableLocationKey locationKey = cache.get(parquetFilePath);
+                if (locationKey == null) {
+                    try {
+                        new ParquetFileReader(parquetFilePath.toString(), new LocalFSChannelProvider());
```
rcaudy, line 55: I know I asked for this, but you're correct that it's expensive. We should probably discuss a little bit.

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/TableLocationKeySafetyImpl.java
devinrsmith: I have verified all the exceptional cases get caught, so it does _something_.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetFileHelper.java
devinrsmith: This is an improvement in efficiency wrt how `ParquetFlatPartitionedLayout` was implemented IMO.

Previously, the implementation called `java.nio.file.Files#newDirectoryStream(java.nio.file.Path, java.lang.String)`, which does a regex compilation every call. In this version, we essentially do the regex once.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetFlatPartitionedLayout.java
devinrsmith: Given we have `File tableRootDirectory`, we _know_ that `FileSystems.getDefault()` will be used. That said, I'm happy to make this more generic in case this ever gets made more generic via `Path tableRootDirectory`.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetFlatPartitionedLayout.java
devinrsmith: Agreed.

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/TableLocationKeySafetyImpl.java
devinrsmith: This is an package private class; that said, there is javadoc on `io.deephaven.engine.table.impl.locations.impl.TableLocationKeyFinder#safetyCheck`.

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/util/ExecutorTableDataRefreshService.java
devinrsmith: So, I've found that errors are silently dropped if we don't catch them here. Originally, `io.deephaven.engine.table.impl.locations.impl.TableLocationKeySafetyImpl#TableLocationKeySafetyImpl` was throwing IllegalArgument or State exceptions, but from the logs / web UI, it seemed like everything was _ok_. Potentially related to discussions in #3608

rcaudy: Silently dropped? Are you certain? It should result in JVM termination. Are we missing an appropriate uncaught exception handler?

rcaudy: It might be that we need to actually check for `Error` in the `ScheduledFuture`.

rcaudy: Regardless, I agree that we need to catch `Throwable` for now.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetFileHelper.java
rcaudy: I agree that this is an improvement for `ParquetFlatPartitionedLayout`, but not for other cases where we're using the matcher. You could easily get the best of both worlds if you change the implementation to be:
```
import java.nio.file.Path;

import static io.deephaven.parquet.table.ParquetTableWriter.PARQUET_FILE_EXTENSION;

final class ParquetFileHelper {

    public static boolean fileNameMatches(Path path) {
        return path.getFileName().endsWith(PARQUET_FILE_EXTENSION);
    }
}
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetFlatPartitionedLayout.java
rcaudy: Looks like you went back to `Files`, with the appropriate overload (`java.nio.file.Files#newDirectoryStream(java.nio.file.Path, java.nio.file.DirectoryStream.Filter<? super java.nio.file.Path>)`), so I think I'm happy. I agree with your point that a `Path` made from `File.toPath` will always have the default FS.

rcaudy (commented):
Better, but some of the outstanding comments stand.

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/TableLocationKeySafetyImpl.java
rcaudy: None that specifies what the safety checks are/do.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetFlatPartitionedLayout.java
rcaudy: I think we can check for the trailer more cheaply than by constructing a reader. If we're going to construct the reader, it would be better to get full benefits of doing so (caching-wise), and use the correct channel provider; you can do this by optimistically constructing the PTLK and calling `io.deephaven.parquet.table.location.ParquetTableLocationKey#getFileReader`.

devinrsmith: Yeah - I was looking into that a bit. I wasn't sure if there's a lot of extraneous overhead due to the CachedChannelProvider / TrackedSeekableChannelsProvider / TrackedFileHandleFactory in that path.

Additionally, this code path explicitly catches IOException and elevates to TableDataException. We can extract that inner part I guess, so we can catch the IOException ourselves.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetFileHelper.java
devinrsmith: Almost, missing a toString().

devinrsmith:
The only new safety check manifests itself if all the partitions change values at once (otherwise, there is internal logic in `io.deephaven.parquet.table.layout.KeyValuePartitionLayout` that ensures no inconsistencies).

```
io.deephaven.engine.table.impl.locations.TableDataException: PollingTableLocationProvider[StandaloneTableKey] has produced an inconsistent TableLocationKey with unexpected partition keys. expected=[TimestampHour2] actual=[TimestampHour].                         
        at io.deephaven.engine.table.impl.locations.impl.AbstractTableLocationProvider.verifyPartitionKeys(AbstractTableLocationProvider.java:212)                                                                                                                    
        at io.deephaven.engine.table.impl.locations.impl.AbstractTableLocationProvider.handleTableLocationKey(AbstractTableLocationProvider.java:113)                                                                                                                 
        at io.deephaven.parquet.table.layout.KeyValuePartitionLayout.lambda$findKeys$0(KeyValuePartitionLayout.java:163)                                                                                                                                              
        at io.deephaven.engine.rowset.RowSet.lambda$forAllRowKeys$0(RowSet.java:306)                                               
        at io.deephaven.engine.rowset.impl.singlerange.SingleRange.ixForEachLong(SingleRange.java:105)                                                                                                                                                                
        at io.deephaven.engine.rowset.impl.WritableRowSetImpl.forEachRowKey(WritableRowSetImpl.java:339)                                                                                                                                                              
        at io.deephaven.engine.rowset.RowSet.forAllRowKeys(RowSet.java:305)                                                        
        at io.deephaven.parquet.table.layout.KeyValuePartitionLayout.findKeys(KeyValuePartitionLayout.java:159)                                                                                                                                                       
        at io.deephaven.engine.table.impl.locations.impl.PollingTableLocationProvider.refresh(PollingTableLocationProvider.java:51)                                                                                                                                   
        at io.deephaven.engine.table.impl.locations.util.ExecutorTableDataRefreshService$ScheduledTableLocationProviderRefresh.refresh(ExecutorTableDataRefreshService.java:121)                                                                                      
        at io.deephaven.engine.table.impl.locations.util.ExecutorTableDataRefreshService$ScheduledSubscriptionTask.doRefresh(ExecutorTableDataRefreshService.java:87)
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetKeyValuePartitionedLayout.java
```diff
@@ -17,7 +17,7 @@ public class ParquetKeyValuePartitionedLayout extends KeyValuePartitionLayout<Pa
     public ParquetKeyValuePartitionedLayout(@NotNull final File tableRootDirectory,
             final int maxPartitioningLevels) {
         super(tableRootDirectory,
-                path -> path.getFileName().toString().endsWith(ParquetTableWriter.PARQUET_FILE_EXTENSION),
+                ParquetFileHelper::fileNameMatches,
```
rcaudy, line 20: This change has rendered an import unused.

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/AbstractTableLocationProvider.java
```diff
@@ -196,6 +204,16 @@ public TableLocation getTableLocationIfPresent(@NotNull final TableLocationKey t
         }
     }
 
+    private void verifyPartitionKeys(@NotNull TableLocationKey locationKey) {
+        if (partitionKeys == null) {
+            partitionKeys = new ArrayList<>(locationKey.getPartitionKeys());
+        } else if (!equals(partitionKeys, locationKey.getPartitionKeys())) {
```
rcaudy, line 210: This is an improvement. We should consider also verifying the types of the partition values. It's a little bit tricky, though, since we'd want to allow for assignable types based on the schema (either inferred or supplied).

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -604,15 +618,26 @@ private static ParquetSchemaReader.ColumnDefinitionConsumer makeSchemaReaderCons
...
     }
 
+    /**
+     * Make a {@link ParquetFileReader} for the supplied {@link File}.
+     *
+     * @param parquetFile The {@link File} to read
+     * @return The new {@link ParquetFileReader}
+     * @throws IOException if an IO exception occurs
+     */
+    public static ParquetFileReader getParquetFileReaderIo(@NotNull File parquetFile) throws IOException {
```
rcaudy, line 634: I want to paint this bike shed a different color, I just don't have a great suggestion. `getParquetFileReaderChecked`?

devinrsmith: Sure.

py/server/deephaven/parquet.py
```diff
@@ -123,38 +162,55 @@ def write(table: Table, path: str, col_definitions: List[Column] = None,
...
+                    table.j_table, path, table_definition, write_instructions
+                )
             else:
                 _JParquetTools.writeTable(table.j_table, _JFile(path), table_definition)
         else:
             if write_instructions:
-                _JParquetTools.writeTable(table.j_table, _JFile(path), write_instructions)
+                _JParquetTools.writeTable(
+                    table.j_table, _JFile(path), write_instructions
+                )
```
jmao-denver, line 196: I think it was formatting like that that turned Chip off.

rcaudy (dismissed):

devinrsmith:
Created https://github.com/deephaven/deephaven-core/issues/3638

py/server/deephaven/parquet.py
```diff
@@ -56,19 +71,31 @@ def _build_parquet_instructions(col_instructions: List[ColumnInstruction] = None
...
 
     return builder.build()
 
 
-def read(path: str, col_instructions: List[ColumnInstruction] = None, is_legacy_parquet: bool = False) -> Table:
+def read(
+    path: str,
+    col_instructions: List[ColumnInstruction] = None,
+    is_legacy_parquet: bool = None,
+    is_refreshing: bool = None,
```
jmao-denver, line 90: For boolean type parameters I think it is best to have a default value, which won't leave the users to wonder if unset what they are.

py/server/deephaven/parquet.py
```diff
@@ -151,9 +191,16 @@ def write(table: Table, path: str, col_definitions: List[Column] = None,
...
-                col_instructions: List[ColumnInstruction] = None, compression_codec_name: str = None,
-                max_dictionary_keys: int = None, grouping_cols: List[str] = None):
+def batch_write(
+    tables: List[Table],
+    paths: List[str],
+    col_definitions: List[Column],
+    col_instructions: List[ColumnInstruction] = None,
+    compression_codec_name: str = None,
+    max_dictionary_keys: int = None,
+    target_page_size: int = None,
```
jmao-denver, line 201: It is fine to have a default value of None, but maybe we should document if unset, what the target_page_size is.

jmao-denver (changes requested):
Some questions about parameter default values.

jmao-denver (approved):
The Python changes look good to me.

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2397
Conceptual: https://github.com/deephaven/deephaven.io/issues/2396
Reference: https://github.com/deephaven/deephaven.io/issues/2395
Blog: https://github.com/deephaven/deephaven.io/issues/2394

PR #3606, opened 2023-03-27
Ensures the avro `map` type maps to `java.util.Map` type.

### LINKED ISSUE #3602: Avro `map` type incorrectly treated as `GenericRecord`
The avro `map` type (https://avro.apache.org/docs/1.11.1/specification/#maps) is treated as a `org.apache.avro.generic.GenericRecord` with respect to our kafka ingestion logic. This seems incorrect, as a `map` is a shallow structure with string keys and typed value. For example:

```json
{
  "type": "map",
  "values": "string"
}
```

DH ostensibly has a test schema, https://github.com/deephaven/deephaven-core/blob/v0.22.1/extensions/kafka/src/test/resources/avro-examples/withmaps.avsc, but this appears to be unused.

This issue manifests itself when values from the column are accessed:

```
java.lang.ClassCastException: class java.util.HashMap cannot be cast to class org.apache.avro.generic.GenericRecord (java.util.HashMap is in module java.base of loader 'bootstrap'; org.apache.avro.generic.GenericRecord is in unnamed module of loader 'app')
```

```
FAULT.refreshThread | rumentedTableListenerBase | Uncaught exception for entry= AddsToRingsListener, added.size()=32, modified.size()=0, removed.size()=0, shifted.size()=0, modifiedColumnSet={EMPTY}:
java.lang.ArrayStoreException: arraycopy: element type mismatch: can not cast one of the elements of java.lang.Object[] to the type of the destination array, org.apache.avro.generic.GenericRecord
```

devinrsmith:
Fixes #3602

### DIFF
--- a/extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
+++ b/extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
@@ -396,7 +396,7 @@ public class KafkaTools {
                 break;
             }
             case MAP:
-                columnsOut.add(ColumnDefinition.fromGenericType(mappedNameForColumn, GenericRecord.class));
+                columnsOut.add(ColumnDefinition.fromGenericType(mappedNameForColumn, Map.class));
                 break;
             case NULL:
             default:


jcferretti (approved):

devinrsmith:
Note: the avro returned type (see `io.deephaven.kafka.ingest.GenericRecordUtil#getPath`) is `HashMap<org.apache.avro.util.Utf8, ...>`. While we could argue we should translate this into a `String` key, it's not unreasonable for us to expect the user to handle this if necessary.

rcaudy (approved):

Issue #3453, opened 2023-02-21
Improve LSP support


mattrunyon:
As a user, I want better LSP support so that the web IDE provides more niceties that I use in my desktop IDE.

We currently have support for `Completion Request`, but we could add support for several other items with Jedi. This list will be in the order that I find most useful/impactful, but all of them enhance the user experience.

- [ ] [Completion request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion) - We support this, but could improve to also include
  - [ ]  `CompletionItemKind` - use LSP enum and then we will convert to monaco enum on client b/c monaco uses a different order for some reason
  - [ ] `documentation` in the response provides an arrow the user can click to see more docs. Or we can [use this hack](https://stackoverflow.com/questions/54795603/always-show-the-show-more-section-in-monaco-editor) to show the documentation always. VSCode shows this documentation by default.
  - [ ] `detail` - this is often used to show the package or some potential item in the autocomplete item on the right side. VSCode for web stuff shows the package name if it is something that will be auto imported on autocomplete. Pycharm has some info about the package the item belongs to regardless of if it is imported already
  - [ ] `deprecated` - Not sure if we're marking anything as deprecated in the Python code, but would be a nice-to-have in the future if we start deprecating APIs. Shows the autocomplete item with strikethrough to indicate deprecation
- [ ] [Signature help](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_signatureHelp) - Provides docstring and param info when you start calling a function
- [ ] [Hover request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_hover) - Provides docstring and other relevant info when hovering over a symbol
- [ ] [Document formatting request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_formatting) - Would be nice to have the code follow some pep8 formatter when possible. The user should be able to disable formatting on the client. This would trigger formatting on save
- [ ] [Document on Type Formatting Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_onTypeFormatting) - Auto indent is usually ok, but this might provide some more info
[...]

PR #3607, opened 2023-03-27
Add signatureHelp and hover to Python autocomplete

mattrunyon:
Corresponds with https://github.com/deephaven/web-client-ui/pull/1178

This addresses the first 3 items in #3453 except deprecation tags.

Adds `signatureHelp` and `hover` requests to the Python autocomplete server

Requires you build and force reinstall the Python server locally unless you are using `pip -e` flag for your local Python server.

Put the plumbing in place for autocomplete cancellation and diagnostic requests

You can either use the web-client-ui main branch dev server or change `web/client-ui/Dockerfile` to `WEB_VERSION=0.32.1-lsp.3` locally to test the web UI.


web/client-api/src/main/java/io/deephaven/web/shared/ide/lsp/SignatureInformation.java
```diff
@@ -0,0 +1,89 @@
...
+import static io.deephaven.web.shared.fu.JsArrays.setArray;
+
+@JsType(namespace = "dh.lsp")
+public class SignatureInformation implements Serializable {
+    public String label;
+    public MarkupContent documentation;
+    private ParameterInformation[] parameters;
+    public int activeParameter;
+
+    public SignatureInformation() {}
```
niloc132, line 27: default constructor need not be declared if it has no params or body like this

web/client-api/src/main/java/io/deephaven/web/shared/ide/lsp/SignatureInformation.java
```diff
@@ -0,0 +1,89 @@
...
+            System.arraycopy(args, 0, typed, 0, length);
+            parameters = typed;
+        } else {
+            throw new IllegalArgumentException("Not a ParameterInformation[] or js []" + args);
+        }
+    }
+
+    @JsProperty
+    public Object getParameters() {
+        return parameters; // Should this clone the array?
```
niloc132, line 46: should probably JsObject.freeze() it when it is assigned

server/src/main/java/io/deephaven/server/console/completer/JavaAutoCompleteObserver.java
```diff
@@ -93,58 +82,74 @@ public void onNext(AutoCompleteRequest value) {
...
+                log.trace().append("Completion canceled").append(exception).endl();
+            }
             safelyOnNext(responseObserver,
                     AutoCompleteResponse.newBuilder()
-                            .setCompletionItems(mangledResults)
+                            .setSuccess(false)
+                            .setRequestId(request.getRequestId())
                             .build());
-        } catch (Exception exception) {
+        } catch (Throwable exception) {
```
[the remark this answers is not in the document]
mattrunyon: That's just what was in `getCompletionItems` error handling. I think the only difference is rethrowing the exception if it's an Error which is from `PythonAutoCompleteObserver`.

If the parser fails or some other step in `getCompletionItems` fails, would it be an `Error` or just a `Throwable`?

web/client-api/src/main/java/io/deephaven/web/shared/ide/lsp/SignatureInformation.java
```diff
@@ -0,0 +1,89 @@
...
+    private ParameterInformation[] parameters;
+    public int activeParameter;
+
+    public SignatureInformation() {}
+
+    @JsProperty
+    public void setParameters(Object args) {
+        if (args == null || args instanceof ParameterInformation[]) {
+            parameters = (ParameterInformation[]) args;
+        } else if (args instanceof JavaScriptObject) {
```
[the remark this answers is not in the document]
mattrunyon: I don't have any plans to port it right now. If that does become the case we can always add this back if needed?

What pattern here are you referencing? What would be the way without the pattern. Just cast? There shouldn't be a case where we create instances of these from JS.

web/client-api/src/main/java/io/deephaven/web/shared/ide/lsp/ParameterInformation.java
```diff
@@ -0,0 +1,45 @@
...
+    @Override
+    @JsIgnore
+    public String toString() {
+        return "ParameterInformation{" +
+                "label=" + label +
+                ", documentation='" + documentation + '\'' +
+                '}';
+    }
+
+    @Override
```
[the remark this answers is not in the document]
mattrunyon: Ok. I added just since the others had them, but I'll remove for the classes I added since they're not used

server/src/main/java/io/deephaven/server/console/completer/JavaAutoCompleteObserver.java
```diff
@@ -93,58 +82,74 @@ public void onNext(AutoCompleteRequest value) {
...
+
+            switch (request.getRequestCase()) {
+                case GET_COMPLETION_ITEMS: {
+                    response.setCompletionItems(getCompletionItems(request.getGetCompletionItems(), exportedConsole,
+                            parser, responseObserver));
+                    break;
+                }
+                case GET_SIGNATURE_HELP: {
+                    // Not implemented. Return empty signature list
+                    response.setSignatures(GetSignatureHelpResponse.newBuilder().build());
```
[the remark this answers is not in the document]
mattrunyon: I think I got them all

server/src/main/java/io/deephaven/server/console/completer/JavaAutoCompleteObserver.java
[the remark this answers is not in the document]
niloc132: It would be an Exception, unless it was something terrible like "out of memory" or "i actually cant talk to python at all" or "this thread is interrupted because the server is shutting down" - things we can't reasonably recover from.

Exception and Error both extend from Throwable, if we have to notice exiting a block in error you would catch throwable, but if you are just handling reasonable failures, it should be an exception

web/client-api/src/main/java/io/deephaven/web/client/ide/LspTranslate.java
```diff
@@ -82,4 +79,56 @@ private static Position toJs(
...
+                src.getParametersList();
+        for (int i = 0; i < paramsList.getLength(); i++) {
+            params.push(toJs(paramsList.getAt(i)));
+        }
+        item.setParameters(params);
+        return item;
+    }
+
+    private static ParameterInformation toJs(
+            final io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.console_pb.ParameterInformation src) {
```
niloc132, line 114: in your place, i would probably make these constructors for the various js types, but this isnt a requirement/expectation

web/client-api/src/main/java/io/deephaven/web/shared/ide/lsp/Hover.java
```diff
@@ -0,0 +1,26 @@
...
+import jsinterop.annotations.JsType;
+
+import java.io.Serializable;
+
+@JsType(namespace = "dh.lsp")
+public class Hover implements Serializable {
+    public MarkupContent contents;
+    public DocumentRange range;
+
+
```
niloc132, line 16: extra newlines

web/client-api/src/main/java/io/deephaven/web/shared/ide/lsp/MarkupContent.java
```diff
@@ -0,0 +1,35 @@
...
+
+@JsType(namespace = "dh.lsp")
+public class MarkupContent implements Serializable {
+    public String kind;
+    public String value;
+
+    public MarkupContent() {}
+
+    @JsIgnore
+    public MarkupContent(MarkupContent source) {
```
niloc132, line 21: this copy constructor appears unused?

niloc132 (dismissed):

py/server/deephaven_internal/auto_completer/_completer.py
```diff
@@ -70,34 +118,43 @@ def can_jedi(self) -> bool:
...
-            Script(txt)
-            if self.__mode == Mode.SAFE
-            else Interpreter(txt, [self.__scope])
-        )
-        completions = completer.complete(line, col)
+        completer = self.get_completer(uri)
+        # Path completions don't seem useful with our setup
+        # It also doesn't suggest nested paths/files when the string is a parent path
+        # Might just be a client issue, but either way not useful right now
+        completions = filter(lambda c: c.type != "path", completer.complete(line, col))
```
jmao-denver, line 147: list comprehension is probably more readable

jmao-denver (changes requested):
Small quibbles

jmao-denver (approved):
The Python changes LGTM

devinrsmith:
I could have swore we didn't throw any huge errors when `jedi` is not installed, but it looks like that is not the case. This is not this PRs fault, but definitely something we should cleanup. Will create follow-up ticket. Should also plan to either close #3174 or re-do.

devinrsmith:
Follow up https://github.com/deephaven/deephaven-core/issues/3640

server/src/main/java/io/deephaven/server/console/completer/JavaAutoCompleteObserver.java
```diff
@@ -93,56 +82,72 @@ public void onNext(AutoCompleteRequest value) {
...
+                case GET_HOVER: {
+                    // Not implemented. Return empty hover help
+                    response.setHover(
+                            GetHoverResponse.newBuilder().setContents(MarkupContent.getDefaultInstance()).build());
+                    break;
+                }
+                case GET_DIAGNOSTIC: {
+                    // Not implemented. Return empty diagnostics
+                    response.setDiagnostic(GetPullDiagnosticResponse.getDefaultInstance());
+                    break;
```
devinrsmith, line 133: It looks like the pattern here is to return success if the server knows about the type, but doesn't implement it. That said,  I wonder if it's better to *not* set any response on the GET_SIGNATURE_HELP / GET_HOVER / GET_DIAGNOSTIC cases. 

The way the implementation works now, for example, on a GET_DIAGNOSTIC request, the client needs to know that an empty `kind` field means it's not implemented?

```
message GetPullDiagnosticResponse {
    string kind = 1;
    optional string result_id = 2;
    repeated Diagnostic items = 3;
}
```

Note: as the server is previously implemented, and implemented now, a newer client won't be able to talk to an older server (REQUEST_NOT_SET will throw an exception, which seems acceptable as long as the client knows how to properly handle it).

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
```diff
@@ -70,128 +70,87 @@ public void onNext(AutoCompleteRequest value) {
...
             }
             case REQUEST_NOT_SET: {
                 throw Exceptions.statusRuntimeException(Code.INVALID_ARGUMENT, "Autocomplete command missing request");
             }
+            default: {
+                // Maintain compatibility with older clients
+                // The only autocomplete type supported before the consoleId was moved to the parent request was
+                // GetCompletionItems
+                final io.deephaven.proto.backplane.grpc.Ticket consoleId =
+                        value.hasConsoleId() ? value.getConsoleId() : value.getGetCompletionItems().getConsoleId();
```
devinrsmith, line 91: Should we have a safer version of this? If `console_id` is not set, ensure that request type is `get_completion_items`? (Could extract code and use on the java side too)

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
```diff
@@ -70,128 +70,87 @@ public void onNext(AutoCompleteRequest value) {
...
-                if (totalCompletionNanos > HUNDRED_MS_IN_NS && log.isTraceEnabled()) {
-                    log.trace().append("Found ")
-                            .append(finalItems.size())
-                            .append(" jedi completions from doc ")
-                            .append(doc.getVersion())
-                            .append("\tjedi_time=").append(toMillis(totalJediNanos))
-                            .append("\tbuild_response_time=").append(toMillis(totalResponseBuildNanos))
-                            .append("\ttotal_complete_time=").append(toMillis(totalCompletionNanos))
-                            .endl();
+            switch (request.getRequestCase()) {
```
devinrsmith, line 128: I think if we ever see the `default` case (in the way the code is currently structured), we should throw an illegal argument exception.

server/src/main/java/io/deephaven/server/console/completer/JavaAutoCompleteObserver.java
devinrsmith: I think if we ever see the `default` case (in the way the code is currently structured), we should throw an illegal argument exception.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/console.proto
```diff
@@ -131,21 +143,36 @@ message AutoCompleteRequest {
...
+        GetPublishDiagnosticResponse diagnostic_publish = 7;
     }
 }
 
 message BrowserNextResponse {
 }
 
 message OpenDocumentRequest {
-    io.deephaven.proto.backplane.grpc.Ticket console_id = 1;
+    io.deephaven.proto.backplane.grpc.Ticket console_id = 1 [deprecated=true];
```
devinrsmith, line 175: Should we have a server safety layer that ensures _if_ the deprecated fields and new fields are set, they are equal; otherwise blowup? Alternatively, we could say either the new fields must be set, or the deprecated fields must be set, but not both.

devinrsmith (commented):
I did not get deep into the LSP side of things. My main concerns are about server/client semantics.

server/src/main/java/io/deephaven/server/console/completer/JavaAutoCompleteObserver.java
mattrunyon: The LSP spec is to first negotiate what the client and server support. Then the client should not send requests or expect results if the server does not support the request type in that negotiation. The client can also tell the server what it supports (like markdown or plaintext only). For now we are just assuming the web IDE which supports markdown

We could instead add a `NOT_SUPPORTED` response type to be more explicit in these cases. In that case we probably wouldn't throw in the `default` and instead just return a `NOT_SUPPORTED` response

On the web client I'm checking if the JS API has either of the new requests before registering the provider, so the web IDE won't try to send requests unless the JS API has the method we expect.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/console.proto
mattrunyon: I think the only reason for deprecation here (as opposed to just ripping out these fields) is if a user had an older web client but newer engine version. Realistically I think the only place that may happen right now is with web developers, so I would be fine just removing these fields instead of deprecating, but I know @niloc132 had other thoughts on that

PR #3610, opened 2023-03-28
UnionSourceManager: Do Not Generate Shifts on Empty Constituents

### LINKED ISSUE #3594: Merging of Stream Tables should not generate shifts.
The merging of stream tables is still a stream:
```groovy

t1 = TableTools.timeTableBuilder().period("00:00:01").streamTable(true).build()
t2 = TableTools.timeTableBuilder().period("00:00:00.1").streamTable(true).build()
t3 = TableTools.timeTableBuilder().period("00:00:00.01").streamTable(true).build()
r1 = TableTools.merge(t1, t2, t3)
println(r1.isStream())
```

However, the merge still produces shifts if any of the tables exceed their "max size". These shifts move empty tables (because they are stream tables) and should be omitted. I suspect that the added rows is also not flat as expected for a stream table.

A good work-around is to `flatten()` the merged stream. The result is a stream, and the upstream shifts will disappear as flatten will squish the stream tables.

nbauernfeind:
Fixes #3594.

Nightlies are running at: https://github.com/nbauernfeind/deephaven-core/actions/runs/4545546046

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/BaseTable.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/BaseTable.java
@@ -630,6 +630,9 @@ public abstract class BaseTable<IMPL_TYPE extends BaseTable<IMPL_TYPE>> extends 
             Assert.assertion(getRowSet().sizePrev() == 0 || getRowSet().lastRowKeyPrev() < update.added().firstRowKey(),
                     "getRowSet().lastRowKeyPrev() < update.added().firstRowKey()");
         }
+        if (isStream()) {
+            Assert.assertion(update.shifted().empty(), "update.shifted.empty()");
+        }
 
         // First validate that each rowSet is in a sane state.
         if (VALIDATE_UPDATE_INDICES) {
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/BaseTable.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/BaseTable.java
@@ -631,6 +631,9 @@ public abstract class BaseTable<IMPL_TYPE extends BaseTable<IMPL_TYPE>> extends 
                     "getRowSet().lastRowKeyPrev() < update.added().firstRowKey()");
         }
         if (isStream()) {
+            Assert.eq(update.added().size(), "added size", getRowSet().size(), "current table size");
+            Assert.eq(update.removed().size(), "removed size", getRowSet().sizePrev(), "previous table size");
+            Assert.assertion(update.modified().isEmpty(), "update.modified.isEmpty()");
             Assert.assertion(update.shifted().empty(), "update.shifted.empty()");
         }
 
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/BaseTable.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/BaseTable.java
@@ -619,7 +619,7 @@ public abstract class BaseTable<IMPL_TYPE extends BaseTable<IMPL_TYPE>> extends 
         Assert.neqNull(update.shifted(), "shifted");
 
         if (isFlat()) {
-            Assert.assertion(getRowSet().isFlat(), "build().isFlat()", getRowSet(), "build()");
+            Assert.assertion(getRowSet().isFlat(), "getRowSet().isFlat()", getRowSet(), "getRowSet()");
         }
         if (isAppendOnly() || isAddOnly()) {
             Assert.assertion(update.removed().isEmpty(), "update.removed.empty()");
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
@@ -499,7 +499,7 @@ public class UnionSourceManager {
             }
 
             if (changes == null || changes.empty()) {
-                // constituent is either static or did not change this cycle
+                // Constituent is either static or did not change this cycle
                 if (slotAllocationChanged) {
                     currFirstRowKeys[nextCurrentSlot + 1] = checkOverflow(nextSlotPrevFirstRowKey + shiftDelta);
                     resultRows.insertWithShift(currFirstRowKey, constituent.getRowSet());
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
@@ -499,10 +499,12 @@ public class UnionSourceManager {
             }
 
             if (changes == null || changes.empty()) {
+                // constituent is either static or did not change this cycle
                 if (slotAllocationChanged) {
                     currFirstRowKeys[nextCurrentSlot + 1] = checkOverflow(nextSlotPrevFirstRowKey + shiftDelta);
                     resultRows.insertWithShift(currFirstRowKey, constituent.getRowSet());
-                    if (shiftDelta != 0) {
+                    // don't bother shifting if the constituent is empty
+                    if (shiftDelta != 0 && constituent.size() > 0) {
                         downstreamShiftBuilder.shiftRange(prevFirstRowKey, prevLastRowKey, shiftDelta);
                     }
                 }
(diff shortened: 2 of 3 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/BaseTable.java
```diff
@@ -630,6 +630,9 @@ public final void notifyListeners(final TableUpdate update) {
             Assert.assertion(getRowSet().sizePrev() == 0 || getRowSet().lastRowKeyPrev() < update.added().firstRowKey(),
                     "getRowSet().lastRowKeyPrev() < update.added().firstRowKey()");
         }
+        if (isStream()) {
+            Assert.assertion(update.shifted().empty(), "update.shifted.empty()");
```
rcaudy, line 634: We could also assert:
1. No modifies
2. Removed size == previous size (other validation and rules ensure removed is a subset of previous)
3. Added size == current size (other validation and rules ensure added is a subset of current)
```suggestion
            Assert.eq(update.added().size(), "added size", getRowSet().size(), "current table size");
            Assert.eq(update.removed().size(), "removed size", getRowSet().sizePrev(), "previous table size");
            Assert.assertion(update.modified().isEmpty(), "update.modified.isEmpty()");
            Assert.assertion(update.shifted().empty(), "update.shifted.empty()");
```

engine/table/src/main/java/io/deephaven/engine/table/impl/BaseTable.java
```diff
@@ -630,6 +630,9 @@ public final void notifyListeners(final TableUpdate update) {
             Assert.assertion(getRowSet().sizePrev() == 0 || getRowSet().lastRowKeyPrev() < update.added().firstRowKey(),
                     "getRowSet().lastRowKeyPrev() < update.added().firstRowKey()");
         }
+        if (isStream()) {
```
rcaudy, line 633: While you're in here, look at the `isFlat()` checks. `build()` should be `getRowSet()`.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
```diff
@@ -499,10 +499,12 @@ private void processExisting(@NotNull final Table constituent) {
             }
 
             if (changes == null || changes.empty()) {
+                // constituent is either static or did not change this cycle
```
rcaudy, line 502:
```suggestion
                // Constituent is either static or did not change this cycle
```

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
```diff
@@ -499,10 +499,12 @@ private void processExisting(@NotNull final Table constituent) {
...
                 if (slotAllocationChanged) {
                     currFirstRowKeys[nextCurrentSlot + 1] = checkOverflow(nextSlotPrevFirstRowKey + shiftDelta);
                     resultRows.insertWithShift(currFirstRowKey, constituent.getRowSet());
-                    if (shiftDelta != 0) {
+                    // don't bother shifting if the constituent is empty
+                    if (shiftDelta != 0 && constituent.size() > 0) {
```
rcaudy, line 507: We should move the line above (`insertWithShift`) into this `if` block.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
```diff
@@ -499,10 +499,12 @@ private void processExisting(@NotNull final Table constituent) {
             }
 
             if (changes == null || changes.empty()) {
+                // constituent is either static or did not change this cycle
                 if (slotAllocationChanged) {
                     currFirstRowKeys[nextCurrentSlot + 1] = checkOverflow(nextSlotPrevFirstRowKey + shiftDelta);
                     resultRows.insertWithShift(currFirstRowKey, constituent.getRowSet());
-                    if (shiftDelta != 0) {
+                    // don't bother shifting if the constituent is empty
```
rcaudy, line 506:
```suggestion
                    // Don't bother shifting or re-inserting if the constituent is empty
```

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
```diff
@@ -556,8 +558,13 @@ private void processExisting(@NotNull final Table constituent) {
...
             } else if (shiftDelta != 0) {
-                Assert.assertion(slotAllocationChanged, "slotAllocationChanged");
-                downstreamShiftBuilder.shiftRange(prevFirstRowKey, prevLastRowKey, shiftDelta);
+                try (final RowSet prevRowSet = constituent.getRowSet().copyPrev()) {
+                    // if prev row set is identical to rows removed, then we do not need to shift
+                    if (!prevRowSet.equals(changes.removed())) {
+                        Assert.assertion(slotAllocationChanged, "slotAllocationChanged");
+                        downstreamShiftBuilder.shiftRange(prevFirstRowKey, prevLastRowKey, shiftDelta);
+                    }
+                }
```
rcaudy, line 567: This could be cheaper, and the cheaper version can be written more cleanly:
```
            } else if (shiftDelta != 0 && constituent.getRowSet().sizePrev() != changes.removed().size()) {
                // Note that changes.removed() must be a subset of the constituent's previous row set.
                // If constituent is removing all of its previous rows, then we do not need to propagate a shift.
                Assert.assertion(slotAllocationChanged, "slotAllocationChanged");
                downstreamShiftBuilder.shiftRange(prevFirstRowKey, prevLastRowKey, shiftDelta);
            }
```

rcaudy (changes requested):

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
```diff
@@ -556,8 +558,13 @@ private void processExisting(@NotNull final Table constituent) {
                     resultRows.insertWithShift(currFirstRowKey, constituent.getRowSet());
                 }
             } else if (shiftDelta != 0) {
```
rcaudy, line 560: Please look at L534:
`final boolean needToProcessShifts = changes.shifted().nonempty() && constituent.getRowSet().isNonempty();`
Maybe we should also be checking that previous was non-empty (using `sizePrev()`)?

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
nbauernfeind: Is `shiftDelta` guaranteed to be non-zero here? We clear `resultRows` past the first shifted row when we set `slotAllocationChanged` to true. If it is valid for shiftDelta to be zero, then we must add the constituent to the result set unconditionally.

I'll drop the `shiftDelta` of zero check. I believe with your partition table related changes that constituents can be removed causing negative shifts; and thus a shift of zero is possible. However, `shiftRange` does ignore shifts of zero.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
nbauernfeind: Seems like a reasonable condition to add; shifts are expensive.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
```diff
@@ -529,7 +531,10 @@ private void processExisting(@NotNull final Table constituent) {
                 nextSlotCurrFirstRowKey = nextSlotPrevFirstRowKey;
             }
 
-            final boolean needToProcessShifts = changes.shifted().nonempty() && constituent.getRowSet().isNonempty();
+            // Ignore shifts if the constituent was empty or became empty
+            final boolean needToProcessShifts = changes.shifted().nonempty()
+                    && constituent.getRowSet().isNonempty()
```
rcaudy, line 536: On re-reading, maybe we should be using the same secondary conditions as when we apply a shift delta. That is, apply shifts if we have any and removed size isn't the same as previous size.

rcaudy (approved):

PR #3619, opened 2023-03-29
Provide the ability to list available Kafka topics from the console

jcferretti:


jcferretti:
Capture_058

py/server/deephaven/stream/kafka/utils.py
jmao-denver, line 1: 1. It might make more sense to have this function moved to the ```__init__.py```
2. A test case?

jmao-denver (changes requested):

py/server/deephaven/stream/kafka/utils.py
```diff
@@ -0,0 +1,29 @@
...
+import jpy
+
+from deephaven import DHError
+from deephaven.jcompat import j_properties
+
+
+_JKafkaTools = jpy.get_type("io.deephaven.kafka.KafkaTools")
+
+def topics(kafka_config: Dict):
+    """Returns a list of topic names available from Kafka.
```
jmao-denver, line 18: Should we return a Python list instead of a JPY wrapped Java list?

jcferretti:
After followup to review comments:
Capture_059

py/server/deephaven/jcompat.py
```diff
@@ -72,6 +72,13 @@ def j_map_to_dict(m) -> Dict[Any, Any]:
     return {e.getKey(): wrap_j_object(e.getValue()) for e in m.entrySet().toArray()}
 
 
+def j_list_to_list(jlist) -> List[Any]:
+    """Converts a java list to a python list."""
+    if not jlist:
+        return None
+
```
jmao-denver, line 79: return [] is better so that callers won't get tripped on ```len(j_list_to_list(...)```, even though in Java null and [] can mean very different things. Also the type hint of List[Any] basically says it is always a List.

py/server/deephaven/stream/kafka/__init__.py
```diff
@@ -3,3 +3,33 @@
...
+from deephaven import DHError
+from deephaven.jcompat import j_list_to_list, j_properties
+
+_JKafkaTools = jpy.get_type("io.deephaven.kafka.KafkaTools")
+
+def topics(kafka_config: Dict) -> List[str]:
+    """Returns a list of topic names available from Kafka.
+
+    Args:
+        kafka_config (Dict): configuration for the associated Kafka consumer.
```
jmao-denver, line 20: Not sure where the 'associated Kafka consumer' comes from.  Should it be a configuration for the targeted Kafka cluster?

jmao-denver (changes requested):
Minor things

py/server/deephaven/stream/kafka/__init__.py
jcferretti: This is exactly the same that we say in the documentation for produce and consume:
```
def produce(
    table: Table,
    kafka_config: Dict,
    topic: str,
    key_spec: KeyValueSpec,
    value_spec: KeyValueSpec,
    last_by_key_columns: bool = False,
) -> Callable[[], None]:
    """Produce to Kafka from a Deephaven table.                                                                                                 
                                                                                                                                                
    Args:                                                                                                                                       
        table (Table): the source table to publish to Kafka                                                                                     
        kafka_config (Dict): configuration for the associated kafka producer     
[...]
```

and


```
def consume(
        kafka_config: Dict,
        topic: str,
        partitions: List[int] = None,
        offsets: Dict[int, int] = None,
        key_spec: KeyValueSpec = None,
        value_spec: KeyValueSpec = None,
        table_type: TableType = TableType.stream(),
) -> Table:
    """Consume from Kafka to a Deephaven table.                                                                                                 
                                                                                                                                                
    Args:                                                                                                                                       
        kafka_config (Dict): configuration for the associated Kafka consumer and also the resulting table.                                      
            Once the table-specific properties are stripped, the remaining one is used to call the constructor of                               
            org.apache.kafka.clients.consumer.KafkaConsumer; pass any KafkaConsumer specific desired configuration here
[...]
```

py/server/deephaven/jcompat.py
jcferretti: Forgot to change that one, sry, ty.

py/server/deephaven/stream/kafka/__init__.py
```diff
@@ -3,3 +3,33 @@
...
+        The content is used to call the constructor of
+        org.apache.kafka.clients.Admin; pass any Admin specific desired configuration here
```
jmao-denver, line 22: Makes sense and looks good, just need to indent these two lines.

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -43,7 +43,8 @@ def produce(
 
     Args:
         table (Table): the source table to publish to Kafka
-        kafka_config (Dict): configuration for the associated kafka producer
+        kafka_config (Dict): configuration for the associated kafka producer.  This is used to call the constructor of
+        org.apache.kafka.clients.producer.KafkaProducer; pass any KafkaProducer specific desired configuration here
```
jmao-denver, line 47: indent please.

jmao-denver (approved):
LGTM

PR #3628, opened 2023-03-29
Improve the level of detail logged for kafka ingestion rates

jcferretti:
add calls to poll and bytes ingested per period stats

### DIFF
--- a/extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaStreamConsumer.java
+++ b/extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaStreamConsumer.java
@@ -13,5 +13,12 @@ import java.util.function.Consumer;
  * Consumer for lists of ConsumerRecords coming from Kafka. The StreamFailureConsumer is extended so that we can report
  * errors emanating from our consumer thread.
  */
-public interface KafkaStreamConsumer extends Consumer<List<? extends ConsumerRecord<?, ?>>>, StreamFailureConsumer {
+public interface KafkaStreamConsumer extends StreamFailureConsumer {
+    /**
+     * Consume a list of ConsumerRecords coming from Kafka.
+     *
+     * @param records the records to consume
+     * @return the total number of message bytes processed
+     */
+    long consume(List<? extends ConsumerRecord<?, ?>> records);
 }
--- a/extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaStreamConsumer.java
+++ b/extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaStreamConsumer.java
@@ -7,7 +7,6 @@ import io.deephaven.stream.StreamFailureConsumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 
 import java.util.List;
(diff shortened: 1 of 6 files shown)


extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaStreamConsumer.java
```diff
@@ -13,5 +13,12 @@
...
  */
-public interface KafkaStreamConsumer extends Consumer<List<? extends ConsumerRecord<?, ?>>>, StreamFailureConsumer {
+public interface KafkaStreamConsumer extends StreamFailureConsumer {
+    /**
+     * Consume a list of ConsumerRecords coming from Kafka.
+     *
+     * @param records the records to consume
+     * @return the total number of message bytes processed
+     */
+    long consume(List<? extends ConsumerRecord<?, ?>> records);
```
devinrsmith, line 23: I don't have strong opinions - but do you prefer `consume` vs `accept`?

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaStreamConsumer.java
```diff
@@ -13,5 +13,12 @@
  * Consumer for lists of ConsumerRecords coming from Kafka. The StreamFailureConsumer is extended so that we can report
  * errors emanating from our consumer thread.
  */
-public interface KafkaStreamConsumer extends Consumer<List<? extends ConsumerRecord<?, ?>>>, StreamFailureConsumer {
+public interface KafkaStreamConsumer extends StreamFailureConsumer {
+    /**
+     * Consume a list of ConsumerRecords coming from Kafka.
+     *
+     * @param records the records to consume
+     * @return the total number of message bytes processed
```
devinrsmith, line 21: Not sure if worth clarifying a bit more - but it's the sum of *serialized* key and value sizes of all the records. I _think_ this means that the in-memory representation may be much bigger (depending on if there is wire compression). But I also _think_ it means that there are bytes unaccounted for (headers, framing, etc).

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaStreamConsumer.java
jcferretti: I do in that accept is generally expected/understood to return void, and I prefer to avoid a potential confusion about this being what it is not (a java.util.function Consumer type interface; there is no ObtToLongFunction in java.util.function so....

jcferretti:
Capture_062
Capture_061

Generating with:

```
cfs@coipo 19:35:35 ~/dh/oss1/deephaven-core/redpanda/examples/python
$ pwd
/home/cfs/dh/oss1/deephaven-core/redpanda/examples/python
(confluent-kafka) 
cfs@coipo 19:36:02 ~/dh/oss1/deephaven-core/redpanda/examples/python
$ j=1000000; while [ $j -gt 0 ]; do python3 ./kafka-produce.py prints 0 MSFT double:20.0; j=$((j - 1)); done

```

devinrsmith (approved):

Issue #3609, opened 2023-03-27
Update gRPC expression / filter / literal structures


devinrsmith:
As part of https://github.com/deephaven/deephaven-core/pull/3441, the structures around `io.deephaven.api.expression.Expression`, `io.deephaven.api.filter.Filter`, and `io.deephaven.api.literal.Literal` will be updated to be more inclusive of the types of mappings expected from SQL. Given that the SQL support is currently server-side only, the gRPC translation of these structures was separated out.

devinrsmith:
During investigation into https://github.com/deephaven/deephaven-core/issues/4606, noticed that this _wasn't_ actually fixed fully; at least not plumbed through with the java client. See comments for this ticket in BatchTableRequestBuilder.

devinrsmith:
Currently, the RPC for filtering a table is FilterTableRequest, which is built on top of Condition. This has served the purposes for web UI purposes, but is lacking a tight integration with the `io.deephaven.api.filter.Filter` APIs. As the java client is becoming more used now, there are some growing pains where we are running into issues with the current implementation (https://github.com/deephaven/deephaven-core/pull/4748).

PR #3630, opened 2023-03-30
SQL Calcite

devinrsmith:
This branch is currently targeted towards a branch that assumes #3609 has already been merged in. It should _not_ be merged as-is until #3609 has been merged and this PR (or a new PR) retargeted to main.


niloc132 (commented):
There are many TODOs that need to either be fixed, made into SQLTODOs, or reference bugs. Also,` System.out.println`s.

Many classes could use javadoc, if only to briefly describe what they are doing to deal with their calcite counterparts - I think the VC summary you gave me would probably do the job for most of these.

I'm not sure that either project specifically belongs where it is - would these make more sense in extensions or java-client?

engine/sql/src/main/java/io/deephaven/engine/sql/Sql.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+        final TableSpec tableSpec = SqlAdapter.parseSql(sql, headers);
+
+        System.out.println(GraphVizMain.toDot(tableSpec));
+
+        final TableCreatorTicketSupport creator = new TableCreatorTicketSupport() {
+            @Override
+            public Table of(TicketTable ticketTable) {
+                final String scanTicket = new String(ticketTable.ticket(), StandardCharsets.UTF_8);
+                if (!scanTicket.startsWith("scan/")) {
```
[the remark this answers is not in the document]
devinrsmith: I've updated the interface to give clear ownership to the calling side; the SqlAdapter layer won't be creating any TicketTables that the caller needs to know the format of.

sql/IMPLEMENTATION.md
```diff
@@ -0,0 +1,21 @@
...
+
+The integration is currently only exposed as a script-session function call - it is _not_ exposed via a "proper" SQL api
+such as ODBC, JDBC, or Flight SQL.
+
+## Column references
+
+Calcite treats all references as indexes (with callers responsible for knowing the proper context in which those indexes
+exist).
+
+TODO: discuss index ref vs named ref
```
rcaudy, line 21: Be sure to label your todos.

sql/src/test/java/io/deephaven/sql/SqlAdapterTest.java
```diff
@@ -0,0 +1,309 @@
...
+import io.deephaven.qst.table.TicketTable;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
```
rcaudy, line 16: Moving from numbers to descriptive names would make this better and more maintainable.

sql/src/test/resources/io/deephaven/sql/query-12.sql
```diff
@@ -0,0 +1,10 @@
+SELECT
+  count(*) as my_count,
+  max(AuthorId) as max_author_id,
+  min(Id) as min_id,
+  FIRST_VALUE(Id) as first_id,
+  LAST_VALUE(Id) as last_id,
+  avg(Id) as avg_id,
+  avg(Id) as avg_id
+FROM
+  books
```
rcaudy, line 10: Are table names case-insensitive? Uppercased for canonicalization?

sql/src/test/resources/io/deephaven/sql/query-12.sql
```diff
@@ -0,0 +1,10 @@
+SELECT
+  count(*) as my_count,
```
rcaudy, line 2: Are we losing name case-sensitivity?

sql/src/main/java/io/deephaven/sql/SqlRootContext.java
```diff
@@ -0,0 +1,72 @@
...
+        // string is parsed again. From a unit testing perspective, and a server cacheability perspective, we would
+        // prefer to assign consistent (internal) column names for the same SQL query string.
+        final String columnPrefix = String.format("%s%d_", context.prefix(), repeatableId(node));
+        return new IndexRef(this, columnPrefix, 0);
+    }
+
+    private int repeatableId(RelNode node) {
+        final Integer id = repeatableId.get(node);
+        if (id == null) {
+            throw new IllegalArgumentException("node does not belong to this context");
```
rcaudy, line 53: This should never happen, but it's cryptic. Should be clearer, and include more info (node, repeatable ID map, etc).

sql/src/main/java/io/deephaven/sql/IndexRef.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rex.RexInputRef;
+
+import java.util.Objects;
+
+class IndexRef implements FieldAdapter, RelNodeAdapter {
+
+    public static String columnName(String prefix, int index) {
+        return String.format("%s%d", prefix, index);
```
rcaudy, line 17: This might not be the best use case for `format`.

sql/src/main/java/io/deephaven/sql/LogicalSortAdapter.java
```diff
@@ -0,0 +1,124 @@
...
+                    break;
+                default:
+                    throw new UnsupportedOperationException("Unsupported sort direction " + fieldCollation.direction);
+            }
+        }
+        final SortTable sortTable = builder.build();
+        if (sort.fetch != null) {
+            final Expression expression = fieldAdapter.expression(sort, sort.fetch);
+            if (!(expression instanceof Literal)) {
+                throw new UnsupportedOperationException("Deephaven only supports literals for FETCH");
```
rcaudy, line 65: Are there named params? We should support that, too.

server/jetty-app/src/main/resources/logback.xml
```diff
@@ -23,6 +23,9 @@
   <!--<logger name="io.netty" level="INFO"/>-->
   <!--<logger name="io.deephaven" level="INFO"/>-->
 
+  <logger name="io.deephaven.sql.SqlAdapter" level="DEBUG"/>
+  <logger name="io.deephaven.engine.sql.Sql" level="DEBUG"/>
+
```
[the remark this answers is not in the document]
devinrsmith: I've added a `dryRun` option that prints out the graphviz as an INFO log, which I think is a reasonable way to go about it.

sql/src/main/java/io/deephaven/sql/SqlAdapter.java
```diff
@@ -0,0 +1,183 @@
...
+     * @param scope the scope
+     * @return the table spec
+     */
+    public static TableSpec parseSql(
+            String sql,
+            Scope scope) {
+        // 0: Configuration
+        // SQLTODO(parse-sql-configuration)
+        //
+        // Allow for customization of calcite parsing details. Would this mean that calcite should / would become part
```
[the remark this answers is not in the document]
devinrsmith: Agreed.

sql/src/main/java/io/deephaven/sql/LogicalSortAdapter.java
devinrsmith: Added new feature `SQLTODO(custom-expression)`

sql/src/test/resources/io/deephaven/sql/query-12.sql
```diff
@@ -0,0 +1,10 @@
+SELECT
```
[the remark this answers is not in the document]
devinrsmith: I'm hoping any improvements in this regard can be handled in a SQL-agnostic way, see `SQLTODO(qst-optimization)`

sql/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+    Classpaths.inheritAssertJ(project)
+    testImplementation 'org.junit.jupiter:junit-jupiter'
+    testImplementation project(':qst:graphviz')
+}
+
+test {
+    useJUnitPlatform()
+}
+
+// todo: should this be a java 8 project?
```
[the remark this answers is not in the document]
niloc132: This note was more about "Why would this be java 8, if it isnt meant to be designed for clients that can't update"

sql/src/test/java/io/deephaven/sql/SqlAdapterTest.java
niloc132: or maybe named constants in the test class

rcaudy (commented):
Finished re-review of files I looked at on Tuesday

sql/src/main/java/io/deephaven/sql/RexNodeExpressionAdapterImpl.java
```diff
@@ -0,0 +1,246 @@
...
+            throw new UnsupportedOperationException("Unsupported RAND with seed");
+        }
+        return RawString.of("java.lang.Math.random()");
+    }
+
+    private Expression currentTimestamp(RexCall call) {
+        if (!call.operands.isEmpty()) {
+            throw new IllegalArgumentException("Expected 0 argument operator");
+        }
+        return RawString.of("DateTime.now()");
```
rcaudy, line 116: This will be broken by the time PR

sql/src/main/java/io/deephaven/sql/RelNodeVisitorAdapter.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Override
+    public RelNode visit(LogicalIntersect intersect) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public RelNode visit(LogicalMinus minus) {
+        throw new UnsupportedOperationException();
+    }
```
rcaudy, line 109: `whereIn` and `whereNotIn`?

sql/src/main/java/io/deephaven/sql/RelNodeVisitorAdapter.java
```diff
@@ -0,0 +1,142 @@
...
+        return values;
+    }
+
+    @Override
+    public RelNode visit(LogicalFilter filter) {
+        out = visitor.visit(filter);
+        return filter;
+    }
+
+    @Override
```
rcaudy, line 73: Maybe this is updateView?

rcaudy:
I think it's very important that we write a comprehensive, user-facing document explaining what we support, what we want to, what we never will, etc.

sql/src/main/java/io/deephaven/sql/RelNodeVisitorAdapter.java
devinrsmith: added todo

sql/src/main/java/io/deephaven/sql/ExtractAnds.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public Collection<Filter> visit(FilterOr ors) {
+        return Collections.singleton(ors);
+    }
+
+    @Override
+    public Collection<Filter> visit(FilterAnd ands) {
+        return ands.filters();
```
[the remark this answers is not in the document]
devinrsmith: I think there's potential to make better simplifiers, and that would probably include making "ExtractAnds" better. I don't mean to cover all the cases right now though.

sql/src/main/java/io/deephaven/sql/RelNodeVisitorAdapter.java
devinrsmith: A LogicalCalc is:

> A relational expression which computes project expressions and also filters.

I think it only comes into play when Calcite optimizations are turned on.

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -233,20 +233,25 @@ protected Changes createDiff(PythonSnapshot from, PythonSnapshot to, RuntimeExce
...
                 final PyObject toValue = change.call(PyObject.class, "__getitem__", int.class, 2);
-                applyVariableChangeToDiff(diff, name, maybeUnwrap(fromValue), maybeUnwrap(toValue));
+                applyVariableChangeToDiff(diff, name, maybeUnwrap(module, fromValue), maybeUnwrap(module, toValue));
             }
             return diff;
         }
     }
 
-    private Object maybeUnwrap(Object o) {
+    private static PythonScriptSessionModule module() {
```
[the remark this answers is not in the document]
devinrsmith: Fixed.

sql/src/main/java/io/deephaven/sql/TypeAdapter.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Override
+    public void visit(BooleanType booleanType) {
+        out = create(SqlTypeName.BOOLEAN);
+    }
+
+    @Override
+    public void visit(ByteType byteType) {
+        out = create(SqlTypeName.TINYINT);
```
rcaudy, line 65: I think this is an incorrect mapping. We have a symmetry problem, no matter what, without richer types.

sql/src/main/java/io/deephaven/sql/TypeAdapter.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Override
+    public void visit(LongType longType) {
+        out = create(SqlTypeName.BIGINT);
+    }
+
+    @Override
+    public void visit(FloatType floatType) {
+        out = create(SqlTypeName.FLOAT);
```
rcaudy, line 90: Is this correct?

devinrsmith: Good call. Should be REAL.

py/server/deephaven/experimental/sql.py
```diff
@@ -0,0 +1,64 @@
...
+        yield
+    finally:
+        j_py_script_session.popScope()
+
+
+def eval(
+    sql: str,
+    dry_run: bool = False,
+    globals: Optional[Mapping[str, Any]] = None,
+    locals: Optional[Mapping[str, Any]] = None,
```
jmao-denver, line 30: I am not clear about the significance of separating the scope into globals and locals. Would something like 'extra_scope' be simpler, unless we do modify the current globals and/or locals?

devinrsmith: This is meant to more most closely match with the scoping model that Python itself uses: https://docs.python.org/3/library/functions.html#exec. I think for 90+% of use cases, the user won't be setting any scoping parameters (so, functionally, it's just as simple to have the defaults for `globals`/`locals` as it is to have the defaults for a single `extra_scope`).

Along this vein, I think there is good potential in the future to re-implement PythonScopeJpyImpl in this fashion as well to keep `globals` / `locals` dicts instead of single `dict` (and potentially, to have the majority of the logic inside python itself to save back and forth boundary crossings); this would potentially allow for python scoping in java to be more efficient (wouldn't need to create a new `dict` each time).

While `extra_scope` would be "simpler" for some use cases, I think there may be advanced use cases that could potentially take advantage of the distinction in the future:

```sql
SELECT * FROM locals.my_table, globals.my_table WHERE locals.my_table.id = globals.my_table.id
```

devinrsmith:
```python
def some_advanced_eval(globals=None, locals=None):
  return eval("...", globals=globals, locals=locals)
```

```sql
SELECT custom_python("some_advanced_eval(globals(), locals())")
```

devinrsmith: It's not necessarily meant to be "clean" - it's meant to be possible to do advanced things. For example,

`org.jpy.PyLib#executeCode(String code, int start, Object globals, Object locals)`

`org.jpy.PyLib#executeScript(String file, int start, Object globals, Object locals)`

both take globals / locals instead of a single dict; it more closely matches how c python operates, https://docs.python.org/3/c-api/veryhigh.html#c.PyEval_EvalCode

devinrsmith: If we think passing in globals / locals via parameters is too speculative right now, I might prefer to get rid of it entirely:

```python
def eval(
    sql: str,
    dry_run: bool = False,
) -> Union[Table, jpy.JType]:
...
```

It's not unreasonable to think that this sort of functionality may be more general in the future, and it's up to the user to call it:

```python
from deephaven.experimental import sql, custom_scope

my_globals = ...
my_locals = ...

with custom_scope(my_globals, my_locals):
  my_table = sql.eval("...")
  
# as opposed to:
# my_table = sql.eval("...", globals=my_globals, locals=my_locals)

File: py/server/tests/test_sql.py [hunk 21]
```diff
@@ -0,0 +1,127 @@
...
+            return sql.eval("SELECT * FROM test_table LIMIT 3")
+
+        result_table = do_sql()
+        self.assertEqual(result_table.size, 3)
+
+    def test_current_timestamp(self):
+        result_table = sql.eval("SELECT CURRENT_TIMESTAMP")
+        self.assertEqual(result_table.size, 1)
+
+
```
chipkent, line 125: We tested a few other scoping scenarios for the table methods.  Consider testing some of those here as well.  See for example the stuff around this line in the test file.
https://github.com/deephaven/deephaven-core/blob/main/py/server/tests/test_table.py#L730

py/server/deephaven/experimental/sql.py
chipkent: On the python table operations side, there were a lot of discussions between @jmao-denver , @rcaudy, and me about how to express the scoping.  Some of the ones that we decided not to follow look like what is currently implemented here.  As a result, I'm for removing the parameters so that normal and SQL behave the same.  If we get some sort of use case that needs further functionality to explicitly define the values, we will need to make the same change in both places.

py/server/deephaven/experimental/sql.py
```diff
@@ -0,0 +1,64 @@
...
+) -> Union[Table, jpy.JType]:
+    """Experimental SQL evaluation. Subject to change.
+
+    If both globals and locals is omitted (the default), the sql is executed with the globals and locals in the
+    environment where eval() is called.
+
+    Args:
+        sql (str): the sql
+        dry_run (bool, optional): if it should be a dry run, False by default
```
chipkent, line 39:
```suggestion
        sql (str): SQL query string
        dry_run (bool, optional): if the query should be a dry run, default is False
```
Changed to be consistent with wording in other docs.

py/server/deephaven/experimental/sql.py
```diff
@@ -0,0 +1,64 @@
...
+    j_py_script_session = _j_py_script_session()
+    # Can do `globals | locals` in Python 3.9+
+    j_py_script_session.pushScope({**globals, **locals})
+    try:
+        yield
+    finally:
+        j_py_script_session.popScope()
+
+
+def eval(
```
chipkent, line 26: `eval` is a built-in python function, so we are better off renaming this to `sql_eval` or something else.

py/server/deephaven/experimental/sql.py
```diff
@@ -0,0 +1,64 @@
...
+    sql: str,
+    dry_run: bool = False,
+    globals: Optional[Mapping[str, Any]] = None,
+    locals: Optional[Mapping[str, Any]] = None,
+) -> Union[Table, jpy.JType]:
+    """Experimental SQL evaluation. Subject to change.
+
+    If both globals and locals is omitted (the default), the sql is executed with the globals and locals in the
+    environment where eval() is called.
```
chipkent, line 35:
```suggestion
    """Evaluates an SQL query and returns the result.
    
    Variables in the query string are taken from the environment when this method is called.
```
Reworded to be more consistent with our other docs.
The experimental package is understood to be experimental and subject to change, so no need to repeat that here.

chipkent (changes requested):

py/server/deephaven/experimental/sql.py
devinrsmith: I'm hoping we always reference it qualified by the sql module:

```python
from deephaven.experimental import sql

sql.eval(...)
```

Or, I could get behind the more explicit name `evaluate(...)`.

py/server/deephaven/experimental/sql.py
devinrsmith: Right now, it's only named tables that we can use. There may be potential in the future to use constants or other variables as input to SQL, but that hasn't been integrated yet.

py/server/tests/test_sql.py
devinrsmith: The scoping for SQL is a bit more limited, as it's limited to only referencing table names. Custom functions are on the roadmap, and we'd be able to get more tests of this kind then.

py/server/deephaven/experimental/sql.py
rcaudy: The user can always do that themselves:
`from deephaven.experimental.sql import eval as sql_eval`

rcaudy (dismissed):
I'm content to merge this, once the Python crowd is on board.

jmao-denver (approved):
LGTM

py/server/deephaven/experimental/sql.py
chipkent: Overwriting a built-in function is generally considered to be bad practice.

py/server/tests/test_sql.py
chipkent: The test cases I'm referring to as examples make sure that query scope is properly obtained from the correct Python scope.

py/server/deephaven/experimental/sql.py
devinrsmith: I've changed the function name to "evaluate"

py/server/tests/test_sql.py
devinrsmith: I've added a new test, but it's not really testing anything new - is there another specific SQL test you want to see?

rcaudy (approved):

rcaudy (approved):

chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2707
Conceptual: https://github.com/deephaven/deephaven.io/issues/2706
Reference: https://github.com/deephaven/deephaven.io/issues/2705
Blog: https://github.com/deephaven/deephaven.io/issues/2708

PR #3632, opened 2023-03-30
Provide reconnect wiring for other JS API types

### LINKED ISSUE #3616: JS API reconnect doesn't work with figures, trees, or partitionedtables
Client side objects don't emit events to indicate that the connection was lost or re-established.

niloc132:
Tested by creating a rollup, a plot with a table, and a plot with a partitioned table (i.e. "plot by"), taking the browser offline for up to a minute, and then going online again, and confirming that all widgets correctly connected and resumed ticking. Note that the current version of web-client-ui will make the page unusable if the old "HACK_CONNECTION_FAILURE" event is emitted - either test with a newer web UI or disable that event.

Fixes #3616

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/lifecycle/HasLifecycle.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/lifecycle/HasLifecycle.java
@@ -3,54 +3,86 @@
  */
 package io.deephaven.web.client.api.lifecycle;
 
+import elemental2.dom.CustomEvent;
 import elemental2.dom.CustomEventInit;
+import elemental2.promise.Promise;
 import io.deephaven.web.client.api.HasEventHandling;
 import io.deephaven.web.client.api.JsTable;
 import io.deephaven.web.client.fu.JsLog;
+import io.deephaven.web.client.fu.LazyPromise;
 import io.deephaven.web.client.state.ClientTableState;
 
 import static io.deephaven.web.client.api.JsTable.EVENT_RECONNECTFAILED;
 
 /**
- * An abstraction over the lifecycle methods used to reconnect JsTable, so we can have non-JsTable bound states get the
- * same revivification treatment.
+ * An abstraction over simple lifecycle methods, to see if objects are currently connected to their server-side
+ * counterparts.
  */
-public interface HasLifecycle {
-
-    // extend HasEventHandling to get these three provided for you.
-    void suppressEvents();
-
-    void unsuppressEvents();
-
-    boolean isSuppress();
-
-    void revive(ClientTableState state);
+public abstract class HasLifecycle extends HasEventHandling {
+    /**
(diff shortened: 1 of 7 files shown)


web/client-api/src/main/java/io/deephaven/web/client/api/lifecycle/HasLifecycle.java
```diff
@@ -3,54 +3,86 @@
...
 
-    void disconnected();
+    /**
+     * Indicate that a new session has been created on the server, and this object should re-create its corresponding
+     * server-side object if possible. Override this to implement custom behavior, being sure to call reconnect() when
+     * finished.
+     *
+     * @return a promise that will resolve when this object is reconnected
+     */
+    public Promise<?> refetch() {
```
mofojed, line 56: Is it worth making `HasLifeCycle` take a template and type this `Promise<?>`? I see if you call `JsTable.refetch()` right now it's just returning a `Promise<Object>` rather than typing correctly to `Promise<JsTable>`.

web/client-api/src/main/java/io/deephaven/web/client/api/widget/plot/JsFigure.java
```diff
@@ -18,6 +18,8 @@
 import io.deephaven.web.client.api.JsPartitionedTable;
```
mofojed, line 18: Unused import `HasEventHandling`

web/client-api/src/main/java/io/deephaven/web/client/api/lifecycle/HasLifecycle.java
```diff
@@ -3,54 +3,86 @@
...
 package io.deephaven.web.client.api.lifecycle;
 
+import elemental2.dom.CustomEvent;
 import elemental2.dom.CustomEventInit;
+import elemental2.promise.Promise;
 import io.deephaven.web.client.api.HasEventHandling;
 import io.deephaven.web.client.api.JsTable;
 import io.deephaven.web.client.fu.JsLog;
+import io.deephaven.web.client.fu.LazyPromise;
 import io.deephaven.web.client.state.ClientTableState;
```
mofojed, line 13: Unused imports

web/client-api/src/main/java/io/deephaven/web/client/api/lifecycle/HasLifecycle.java
niloc132: Yeah I'll override it in table so that it throws (tables have a more complex reconnect than just "hi i'm back"), but since the ? is covariant we can override this to make it more specific in subclasses (see JsFigure and JsPartitionedTable). This type won't be visible to JS, and this method particularly isn't intended to be exposed, but I am reticent to make this type recursively generic for fear of the meaning being lost or confused in translation to TS...

PR #3637, opened 2023-03-31
Specify Use of Java NIO for extensions-arrow

nbauernfeind:
Fixes java17/java18 issue not being able to initialize arrow memory utils. The issue was introduced in #3420.

### DIFF
--- a/extensions/arrow/build.gradle
+++ b/extensions/arrow/build.gradle
@@ -33,3 +33,5 @@ spotless {
         )
     }
 }
+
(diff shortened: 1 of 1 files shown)


devinrsmith (approved):
I'm assuming you have verified this locally (both, the tests do fail without it, and are fixed with it).

Probably worthwhile to add a reference and commit message to the PR # where this issue was introduced.

nbauernfeind:
Yes, tested that `./gradlew extensions-arrow:test -PtestRuntimeVersion=17` fails without the change and passes with it.

Issue #3638, opened 2023-03-31
Apply consistent formatting for python files


devinrsmith:
Just like we have automatic code formatting for .java files, it would be great to have the same for .python files.

Spotless has support for python files https://github.com/diffplug/spotless/tree/main/plugin-gradle#python.

The only current python format spotless currently implements is based around https://github.com/psf/black. That said, black seems like a great candidate as it seems like a very popular choice.

devinrsmith:
https://github.com/deephaven/deephaven-core/issues/5352 would be nice-to-have before this feature.

devinrsmith:
https://github.com/astral-sh/ruff has also been mentioned as black-compatible formatter / linter.

PR #3639, opened 2023-03-31
Add spotless black formatting

devinrsmith:
Note - this is a quick draft to see what all the python differences would look like. This PR is *NOT* ready from a CI / gradle perspective.

See #3638


devinrsmith:
This is editing some auto-generated files, we'd need to fix that.

devinrsmith:
Other potential formatters (no current spotless support):

https://github.com/google/yapf
https://github.com/hhatto/autopep8

Issue #3648, opened 2023-04-03
 Move getMeta into TableOperations 


jmao-denver:
Currently ```getMeta``` operation is the ```io.deephaven.engine.table.Table``` interface. If we elevate it into ```io.deephaven.api.TableOperations```.  the Java client will get it for free.

Issue #3620, opened 2023-03-29
Support getting meta table in the Python client


jmao-denver:
Since there is not a gRPC call for that, here is an expensive way to achieve it:

1. bind the table to a non-conflicting name in the global scope
2. run a console script to assign the meta table to the chosen name in 1.
3. open the table through the fetch table call
4. run another console script to set the name to None

Since every table in the Python client already has a Arrow schema associated with it, it is not clear whether what this meta table feature brings to the users is truly needed.

nbauernfeind:
Chip asked to add `getMeta` to the gRPC Table API. This seems like a reasonable request as it is a `Table -> Table` operation. See #2243.

jmao-denver:
closed by #3649

PR #3649, opened 2023-04-03
Add getMeta gRPC call and meta_table in py client

### LINKED ISSUE #2243: Support `getMeta` in client APIs
The client APIs do not currently support `getMeta`.  This is a very useful method when developing code.  To support the client APIs, `getMeta` must first be implemented in the gRPC layer.

jmao-denver:
Fixes #2243 #3620 

related to #3648 

TODO: Add getMeta in the  Go (@chipkent ) and CPP (@kosak ) clients


py/client/pydeephaven/table.py
```diff
@@ -72,6 +73,12 @@ def _parse_schema(self, schema_header):
         reader = pa.ipc.open_stream(schema_header)
         self.schema = reader.schema
 
+    @property
+    def meta_table(self) -> Table:
```
devinrsmith, line 77: Does this make sense as a `property`?

authorization/src/main/java/io/deephaven/auth/codegen/impl/TableServiceContextualAuthWiring.java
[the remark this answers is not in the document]
jmao-denver: Yeah, I had the same question and asked @niloc132 about it. @nbauernfeind your thoughts?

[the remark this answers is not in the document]
devinrsmith: The scripts are in `authorization-codegen/`. `authorization/README.md` w/ details.

py/client/pydeephaven/table.py
jmao-denver: I think you can argue either way. The server API also has it as a property.

devinrsmith: I'm not fully aware of the pythonic semantics one might assume when something is marked as a property. That said, naively I would be surprised as an end-user if I call a property method and it ends up needing to do a round-trip RPC.

jmao-denver: @property is the Pythonic way to implement a getter method, though not recommended for getters that take long time to finish.

niloc132 (dismissed):

py/client/pydeephaven/table.py
devinrsmith: Isn't this a prime example of a method that might take a long time to finish?

devinrsmith (changes requested):
The authorization wiring is incorrect, it needs to be done with the code generation. Right now the method names are incorrect.

py/client/pydeephaven/table.py
jmao-denver: yes, and I guess it also depends on the real use case of it. OTOH we also want to keep the client API/server API consistent.

server/src/main/java/io/deephaven/server/table/ops/MetaTableGrpcImpl.java
```diff
@@ -0,0 +1,31 @@
...
+import io.deephaven.base.verify.Assert;
+import io.deephaven.engine.table.Table;
+import io.deephaven.proto.backplane.grpc.BatchTableRequest;
+import io.deephaven.proto.backplane.grpc.MetaTableRequest;
+import io.deephaven.server.session.SessionState;
+
+import javax.inject.Inject;
+import java.util.List;
+
+public class MetaTableGrpcImpl extends GrpcTableOperation<MetaTableRequest> {
```
chipkent, line 16: Does this java stuff need docs?

devinrsmith (approved):

PR #3656, opened 2023-04-04
Added `Delta` to `UpdateBy` operators

lbooker42:


replication/static/src/main/java/io/deephaven/replicators/ReplicateUpdateBy.java
```diff
@@ -132,6 +132,18 @@ public static void main(String[] args) throws IOException {
...
+        files = ReplicatePrimitiveCode.charToAllButBoolean(
+                "engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/delta/CharDeltaOperator.java",
+                exemptions);
+        for (final String f : files) {
+            if (f.contains("Int")) {
+                fixupInteger(f);
+            }
+            // if (f.contains("Long")) {
+            // augmentLongWithReinterps(f);
+            // }
```
nbauernfeind, line 145: remove commented code?

nbauernfeind:
As discussing offline, I am curious/concerned about the behavior when overflow occurs. The silent overflow could yield a bad user-experience; maybe we should explicitly error on overflow? Alternatively, we could auto-widen the result on behalf of the user; I just don't like that would imply `long` -> `BigInteger` which feels off.

nbauernfeind (dismissed):

nbauernfeind (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2454
Conceptual: https://github.com/deephaven/deephaven.io/issues/2457
Reference: https://github.com/deephaven/deephaven.io/issues/2455
Blog: https://github.com/deephaven/deephaven.io/issues/2456

PR #3659, opened 2023-04-04
Improve query library performance by using vector iteration

### LINKED ISSUE #3574: Update function libraries to use iteration with Vectors
Once #3570 merges, we can and should change more or less all Vector iteration in our libraries to use `iterator` or `stream` methods, rather than looping on `get` calls. This is potentially far more efficient for some `Vector` types, especially those downstream of a `groupBy`, `aggBy` with `Group`, `updateBy` with `RollingGroup`, or `rangeJoin` with `Group`.

### For example:
This code in `Basic.ftl`:
```
    static public <T> long countObj(ObjectVector<T> values) {
        if (values == null) {
            return NULL_LONG;
        }

        final long n = values.size();
        long count = 0;

        for (long i = 0; i < n; i++) {
            T c = values.get(i);

            if (!isNull(c)) {
                count++;
            }
        }

        return count;
    }
```
should be:
```
    static public <T> long countObj(ObjectVector<T> values) {
        if (values == null) {
            return NULL_LONG;
        }

        if (values.isEmpty()) {
            return 0;
        }

        long count = 0;
        for (final Iterator<T> vi = values.iterator(); vi.hasNext(); ){
            final T c = vi.next();

            if (!isNull(c)) {
                count++;
            }
        }

        return count;
    }
```
We could also write it like this, but it's probably worse:
```
    static public <T> long countObj(ObjectVector<T> values) {
        if (values == null) {
            return NULL_LONG;
        }

        if (values.isEmpty()) {
            return 0;
        }

        return values.iterator().stream().filter(c -> !isNull(c)).count();
    }
```

### A related change we should also make: Avoid using `Vector` `get` when array access will do:
This code in `Basic.ftl`:
```
    static public <T> T[] replaceIfNull(ObjectVector<T> values, T replacement) {
        final int n = values.intSize("replaceIfNull");
        T[] result = values.toArray();

        for (int i = 0; i < n; i++) {
            result[i] = replaceIfNull(values.get(i), replacement);
        }

        return result;
    }
 ```
should be:
```
    static public <T> T[] replaceIfNull(ObjectVector<T> values, T replacement) {
        final int n = values.intSize("replaceIfNull");
        T[] result = values.toArray();

        for (int i = 0; i < n; i++) {
            result[i] = replaceIfNull(result[i], replacement);
        }

        return result;
    }
```

### PR DESCRIPTION
**Pull request by chipkent**
Performance of the query library has been improved by moving from direct access to iteration.

Resolves #3574

### CODE REVIEW

File: engine/function/src/templates/Basic.ftl [hunk 1]
```diff
@@ -122,11 +123,12 @@ public class Basic {
...
     static public <T> T[] replaceIfNull(ObjectVector<T> values, T replacement) {
-        final int n = values.intSize("replaceIfNull");
         T[] result = values.toArray();
+        int i = 0;
 
-        for (int i = 0; i < n; i++) {
-            result[i] = replaceIfNull(values.get(i), replacement);
+        for (final T v : values) {
+            result[i] = replaceIfNull(v, replacement);
+            i++;
```
rcaudy, line 131: You should revert this change, and instead replace just the unnecessary lookup in `values` with a lookup in `result`:
```
            result[i] = replaceIfNull(result[i], replacement);
```

engine/function/src/main/java/io/deephaven/function/Logic.java
```diff
@@ -49,8 +49,7 @@ static public Boolean and(boolean... values) {
      * @return logical and of all the values in the array. By convention, returns true if the array is empty.
      */
     static public Boolean and(ObjectVector<Boolean> values) {
-        for (int ii = 0; ii < values.size(); ++ii) {
-            Boolean b = values.get(ii);
+        for (final Boolean b : values) {
```
rcaudy, line 52: One thing I did a poor job of explaining in the issue is that these iterators must be closed, although for all the current implementations it’s enough to use them until they are exhausted. This loop *is* using the iterator to exhaustion, so it's "OK". That said, if we wanted to be extra careful/pedantic, we could instead use a try-with-resources block. This applies to pretty much every usage, so just commenting once for discussion.

engine/function/src/main/java/io/deephaven/function/Logic.java
```diff
@@ -86,11 +85,10 @@ static public Boolean and(Boolean[] values, Boolean nullValue) {
      * @return logical and of all the values in the array. By convention, returns true if the array is empty.
      */
     static public Boolean and(ObjectVector<Boolean> values, Boolean nullValue) {
-        for (int ii = 0; ii < values.size(); ++ii) {
-            Boolean b = values.get(ii);
-            b = b == null ? nullValue : b;
+        for (final Boolean b : values) {
```
rcaudy, line 88: We should consider whether we should check the empty case before making the iterator, and apply whatever convention we decide on universally. Some of the iterators are a little expensive to make, although we usually make cheaper ones for small/empty vectors. Even the cheapest iterators are still an allocation, although hopefully "escaped". This applies to a lot of methods, so just commenting once for discussion.

engine/function/src/templates/Basic.ftl
```diff
@@ -760,9 +757,18 @@ public class Basic {
...
 
-        for (int i=0; i < n_c; i++) {
-            result[i] = condition.get(i) == null ? null : (condition.get(i) ? trueCase.get(i) : falseCase.get(i));
-        }
+        final Iterator<Boolean> ci = condition.iterator();
+        final Iterator<T> ti = trueCase.iterator();
+        final Iterator<T> fi = falseCase.iterator();
+        int i = 0;
+
+        while(ci.hasNext() && ti.hasNext() && fi.hasNext()) {
```
rcaudy, line 765: Since you've established that the sizes are equal, I would just check `ci.hasNext()` and skip the others.

engine/function/src/templates/Basic.ftl
```diff
@@ -983,9 +991,11 @@ public class Basic {
     static public ${pt.primitive}[] replaceIfNull(${pt.vector} values, ${pt.primitive} replacement) {
         final int n = values.intSize("replaceIfNull");
         ${pt.primitive}[] result = new ${pt.primitive}[n];
+        int i = 0;
 
-        for (int i = 0; i < n; i++) {
-            result[i] = replaceIfNull(values.get(i), replacement);
+        for (final ${pt.primitive} v : values) {
```
rcaudy, line 996: More evidence that I did a bad job writing up the issue. This isn't the right pattern to follow, now that we're into the primitives. This approach will box every value (correctly, but we don't want to do it) and then force the language to unbox naively (which will NPE if we boxed to `null`). You'll need to use the iterator explicitly, and invoke `nextChar`, etc to get the primitives, or use `forEachRemaining` with the appropriate primitive consumer.

The `char` version should then look like:
```
try (final CloseablePrimitiveIteratorOfChar iterator = values.iterator()) {
    while (iterator.hasNext()) {
        is (!isNull(iterator.nextChar())) {
            count++;
        }
    }
}
```

This applies more or less to all of the remaining methods you've updated.

engine/function/src/templates/Basic.ftl
```diff
@@ -1671,9 +1680,18 @@ public class Basic {
...
 
-        for (int i=0; i < n_c; i++) {
-            result[i] = condition.get(i) == null ? ${pt.null} : (condition.get(i) ? trueCase.get(i) : falseCase.get(i));
-        }
+        final Iterator<Boolean> ci = condition.iterator();
+        final ${pt.vectorIterator} ti = trueCase.iterator();
+        final ${pt.vectorIterator} fi = falseCase.iterator();
+        int i = 0;
+
+        while(ci.hasNext() && ti.hasNext() && fi.hasNext()) {
```
rcaudy, line 1688: Same comment as in `ifElseObj`.

rcaudy (commented):
Stopping early, because we need to do more work for the primitive cases.

engine/function/src/main/java/io/deephaven/function/Logic.java
rcaudy: We decided to use try-with-resources everywhere.

engine/function/src/main/java/io/deephaven/function/Logic.java
rcaudy: We decided not to check `isEmpty()` unless we have a reason to in the function.

engine/function/src/templates/Basic.ftl
```diff
@@ -1459,10 +1513,12 @@ public class Basic {
         final ArrayList<${pt.boxed}> orderedList = new ArrayList<>();
```
rcaudy, line 1513: We should audit the usages of `toArray()`. I think we need to be more cautious about our contracts re: copying.
1. L126: `replaceIfNullObj` as noted in a separate comment
2. L304: `arrayObj` might be returning the vector's own backing array. Is this OK? The JavaDoc doesn't specify. Do you need/want a version of `toArray` that guarantees it will always allocate a new result?
3. L530: `distinctObj` has the same kind of issue as `arrayObj` for single-element vectors.
4. L920: `forwardFillObj` as noted in a separate comment.
5. L1190: `array` has the same kind of issue as `arrayObj`
6. L1439: `distinct` has the same kind of issue as `distinctObj`
7. L1510: `ObjectVector<Boolean>` version of `distinct` has the same kind of issue as `distinctObj`

engine/function/src/templates/Sort.ftl
```diff
@@ -30,7 +30,7 @@ public class Sort {
         if (values == null) {
             return null;
         }
-        if (values.size() == 0) {
+        if (values.isEmpty()) {
```
rcaudy, line 33: Same comment about auditing `toArray()` as in `Basic.ftl`. The cases where we return an empty `toArray()` are fine. Sorting an un-copied `toArray()` is not. I see this issue on L38 (in `sortObj`) and L101 (in `sortDescendingObj`).

We copy aggressively in the primitive versions, at L162 and L220 respectively. This is safe, but might be an extra allocation and copy in some cases.

engine/function/src/templates/Numeric.ftl
```diff
@@ -532,10 +548,13 @@ public class Numeric {
         double sum = 0;
         double sum2 = 0;
         double count = 0;
+        final ${pt.vectorIterator} vi = values.iterator();
```
rcaudy, line 551: Missing try-with-resources.

engine/function/src/templates/Numeric.ftl
```diff
@@ -764,11 +783,12 @@ public class Numeric {
         // see https://stats.stackexchange.com/questions/25895/computing-standard-error-in-weighted-mean-estimation
         double sumw = 0;
         double sumw2 = 0;
-        final long n = values.size();
+        final ${pt.vectorIterator} vi = values.iterator();
```
rcaudy, line 786: Missing try-with-resources.

engine/function/src/templates/Numeric.ftl
```diff
@@ -1225,13 +1253,17 @@ public class Numeric {
         double sum1 = 0;
         double sum01 = 0;
         double count = 0;
-        final long n = values0.size();
+        final ${pt.vectorIterator} v0i = values0.iterator();
```
rcaudy, line 1256: Missing try-with-resources.

engine/function/src/templates/Numeric.ftl
```diff
@@ -1306,15 +1338,19 @@ public class Numeric {
...
-        final long n = values0.size();
-
-        for (long i = 0; i < n; i++) {
-            if (!isNull(values0.get(i)) && !isNull(values1.get(i))) {
-                sum0 += values0.get(i);
-                sum0Sq += values0.get(i) * values0.get(i);
-                sum1 += values1.get(i);
-                sum1Sq += values1.get(i) * values1.get(i);
-                sum01 += values0.get(i) * values1.get(i);
+        final ${pt.vectorIterator} v0i = values0.iterator();
```
rcaudy, line 1341: Missing try-with-resources.

engine/function/src/templates/Numeric.ftl
```diff
@@ -1463,22 +1502,28 @@ public class Numeric {
...
 
-        if (values.size() == 0) {
+        if (values.isEmpty()) {
             return new ${pt.primitive}[0];
         }
 
         final int n = values.intSize("cummin");
         ${pt.primitive}[] result = new ${pt.primitive}[n];
-        result[0] = values.get(0);
+        final ${pt.vectorIterator} vi = values.iterator();
```
rcaudy, line 1511: Missing try-with-resources.

engine/function/src/templates/Numeric.ftl
```diff
@@ -1530,17 +1575,20 @@ public class Numeric {
...
         }
 
-        if (values.size() == 0) {
+        if (values.isEmpty()) {
             return new ${pt.primitive}[0];
         }
 
         final int n = values.intSize("cummax");
         ${pt.primitive}[] result = new ${pt.primitive}[n];
+        final ${pt.vectorIterator} vi = values.iterator();
```
rcaudy, line 1584: Missing try-with-resources.

engine/function/src/templates/Numeric.ftl
```diff
@@ -1598,22 +1646,28 @@ public class Numeric {
...
 
-        if (values.size() == 0) {
+        if (values.isEmpty()) {
             return new ${pt.primitive}[0];
         }
 
         final int n = values.intSize("cumsum");
         ${pt.primitive}[] result = new ${pt.primitive}[n];
-        result[0] = values.get(0);
+        final ${pt.vectorIterator} vi = values.iterator();
```
rcaudy, line 1655: Missing try-with-resources.

engine/function/src/templates/Numeric.ftl
```diff
@@ -1671,22 +1725,28 @@ public class Numeric {
...
 
-        if (values.size() == 0) {
+        if (values.isEmpty()) {
             return new ${pt.primitive}[0];
         }
 
         final int n = values.intSize("cumprod");
         ${pt.primitive}[] result = new ${pt.primitive}[n];
-        result[0] = values.get(0);
+        final ${pt.vectorIterator} vi = values.iterator();
```
rcaudy, line 1734: Missing try-with-resources.

engine/function/src/templates/Numeric.ftl
```diff
@@ -2154,10 +2214,13 @@ public class Numeric {
         }
 
         double vsum = 0;
+        final ${pt.vectorIterator} vi = values.iterator();
```
rcaudy, line 2217: Missing try-with-resources.

engine/function/src/templates/Numeric.ftl
```diff
@@ -2231,10 +2294,13 @@ public class Numeric {
 
         double vsum = 0;
         double wsum = 0;
+        final ${pt.vectorIterator} vi = values.iterator();
```
rcaudy, line 2297: Missing try-with-resources.

engine/function/src/templates/Numeric.ftl
```diff
@@ -1055,14 +1079,18 @@ public class Numeric {
...
+        try (final ${pt.vectorIterator} vi = values.iterator()) {
+            while ( vi.hasNext() ) {
+                final ${pt.primitive} c = vi.${pt.iteratorNext}();
+                if (!isNull(c) && (c < val || (c == val && count == 0) )) {
+                    val = c;
+                    index = i;
+                    count++;
+                }
+
+                i++;
```
rcaudy, line 1093: Sticking this comment on an arbitrary line. We have two `toArray()` issues in this file:
1. L1140, in `median`: We intend to copy, but we might not be
2. L1180, in `percentile`: Same issue.

engine/function/src/templates/Basic.ftl
```diff
@@ -1294,8 +1343,11 @@ public class Basic {
 
         final Set<${pt.boxed}> keys = new HashSet<${pt.boxed}>();
 
-        for (long ii = 0; ii < n; ii++) {
-            keys.add(values.get(ii));
+        try (final ${pt.vectorIterator} vi = values.iterator()) {
```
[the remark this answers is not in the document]
chipkent: I removed it.  We will have to mention this in the release notes.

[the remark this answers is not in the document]
rcaudy: Will we? I expect it will "just work".

engine/function/src/templates/Basic.ftl
```diff
@@ -1459,10 +1513,12 @@ public class Basic {
         final ArrayList<${pt.boxed}> orderedList = new ArrayList<>();
         final Set<${pt.boxed}> counts = new HashSet<>();
 
-        for (long ii = 0; ii < n; ii++) {
-            ${pt.primitive} val = values.get(ii);
-            if ((includeNull || val != QueryConstants.${pt.null}) && counts.add(val)) {
-                orderedList.add(val);
+        try (final ${pt.vectorIterator} vi = values.iterator()) {
```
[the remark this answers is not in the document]
chipkent: Removed the vector method and made the array method more efficient.

rcaudy (commented):
I'm caught up. Once we merge my PR to this PR, I think we can merge (or ask Corey to review for extra safety).

rcaudy (approved):

PR #3661, opened 2023-04-05
Simplify export listeners in the server

### LINKED ISSUE #3590: Investigate gRPC's feedback when IO errors occur
In some places, we effectively ignore any exceptions that come out of StreamObserver.onNext, while in others we use those errors to ensure we don't call onNext again. 

The concern is that if onNext throws, gRPC might not deliver the error to us in any other way (such as onError, onClose, or setOnCancelHandler), so we must catch it and remove any chance that we would write to the stream again. If this is true, we might want a better tool than `safelyOnNext`, where the failure to deliver the message results in cleaning up resources.

On the other hand, if errors of this nature are being reported (and checked for!) consistently via setOnCancelHandler, we should simplify how we handle onNext errors for long-running streams.

niloc132:
Any code observing an export listener no longer is responsible to remove its own handler in case of the stream being canceled, and can use safelyComplete/safelyOnNext to signal clients.

Also fixes a websocket transport issue where errors were not correctly reported to the stream.

Fixes #3590

(temporarily left as draft to discuss a potential race condition)

### DIFF
--- a/server/src/main/java/io/deephaven/server/util/GrpcServiceOverrideBuilder.java
+++ b/server/src/main/java/io/deephaven/server/util/GrpcServiceOverrideBuilder.java
@@ -4,6 +4,7 @@
 package io.deephaven.server.util;
 
 import com.google.rpc.Code;
+import io.deephaven.extensions.barrage.util.GrpcUtil;
 import io.deephaven.proto.util.Exceptions;
 import io.deephaven.server.browserstreaming.BrowserStream;
 import io.deephaven.server.browserstreaming.BrowserStreamInterceptor;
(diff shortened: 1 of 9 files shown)


server/src/main/java/io/deephaven/server/util/GrpcServiceOverrideBuilder.java
```diff
@@ -266,8 +267,7 @@ public void invokeNext(
                     .onError(responseObserver)
                     .submit(() -> {
                         browserStream.get().onMessageReceived(request, streamData);
-                        responseObserver.onNext(null);// TODO simple response payload
-                        responseObserver.onCompleted();
+                        GrpcUtil.safelyComplete(responseObserver, null);
```
nbauernfeind, line 270: I don't think this is OK. I believe Unary requests are required to yield a singular response.

Issue #188, opened 2021-03-23
Improve JS API's usage of subscriptions


niloc132:
To expedite development, JS is still using flatten, and re-creating subscriptions rather than updating them out-of-band. Fortunately, by flattening we reduce the cost of re-creating subscriptions, at least in terms of sending large sparse indexes.

The first fix should be to change both types of subscriptions to share the same code to allow for non-flat indexes - to achieve this, viewports will need to handle receiving a snapshot that only covers part of the new viewport. Ideally, out-of-band subscription updates should be implemented before this work is done.

Additionally, subscriptions are using ColumnData rather than just creating the real data that is used by the API - some of this indirection can be cleaned up. Doing this work will also let us make the kinds of data we can read much more flexible, and possibly consider delegating to arrow.

Breaking this down as bulletpoints rather than text above:
 * Stop re-creating subscriptions when the viewport or columns change, but still subscribed to the same table
 * Support the "reverse viewport" feature
 * Stop requiring "flat" tables - some code should be sharable with the non-viewport subscription to achieve this, but it will need to be generalized to support viewports.
 * Drop the ColumnData, TableSnapshot, DeltaUpdates intermediate types, as long as so much of the above is being rewritten, to make it easier to support new types.

niloc132:
See https://github.com/deephaven/deephaven-core/issues/2435

niloc132:
Example query that is slow due to "redirection index" 
```Python
from deephaven import empty_table
from deephaven.plot.figure import Figure

t = empty_table(100000).update(["X=ii", "Y=ii*2"])
figure = Figure()
before = figure\
    .plot_xy(series_name="Y", t=t, x="X", y="Y")\
    .show()
```

niloc132:
https://github.com/deephaven/deephaven-core/issues/1917

niloc132:
Implementation plan:
 * Share flatbuffer lib and generated java with GWT implementation, rather than using flatbuffers.js. This requires a small amount of supersource, and saves about 60KB before gwt minification. It appears this will also save us copying arraybuffers in a few places.
     * Consider removing the remaining copies by aligning header/metadata/data with protobuf tricks.
 * Share low-level barrage read/write code with the server. The Flight code is not shared, since flight's schema/etc implementation require a real JVM in too many ways. This includes the engine-chunk library, and the *ChunkInputStreamGenerator types. Similarly, StreamReaderOptions, FieldNodeInfo will be part of this api. 
[...]

PR #3662, opened 2023-04-05
Restore totals table support in the JS API

### LINKED ISSUE #3314: JS API support for totals tables
The JS API is capable of producing totals and grand totals tables that update as the user interacts with the underlying table. However, these haven't been updated to use the new grpc aggregation APIs.

niloc132:
Fixes #3314

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/util/TotalsTableBuilder.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/TotalsTableBuilder.java
@@ -606,14 +606,14 @@ public class TotalsTableBuilder {
 
         final Set<AggType> defaultOperations = EnumSet.of(builder.defaultOperation);
         final Map<AggType, List<String>> columnsByType = new LinkedHashMap<>();
-        for (final Map.Entry<String, ? extends ColumnSource> entry : source.getColumnSourceMap().entrySet()) {
+        for (final Map.Entry<String, ? extends ColumnSource<?>> entry : source.getColumnSourceMap().entrySet()) {
             final String columnName = entry.getKey();
             if (ColumnFormatting.isFormattingColumn(columnName)) {
                 continue;
             }
 
             final Set<AggType> operations = builder.operationMap.getOrDefault(columnName, defaultOperations);
-            final Class type = entry.getValue().getType();
+            final Class<?> type = entry.getValue().getType();
 
             for (final AggType op : operations) {
                 if (operationApplies(type, op)) {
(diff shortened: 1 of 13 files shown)


mofojed (commented):
@niloc132 At a quick glance, how would @emilyhuxng use this for the databar stuff? We can create a totals table, but we can't join it back to the original table. `JsTable.join` expects another `JsTable`, whereas this is a `JsTotalsTable`. Or am I misthinking this?

niloc132:
Good call, I'll work on relaxing that a bit further. Under the hood, a TotalsTable is just a Table, except we don't let you perform certain operations on it that don't make sense.

niloc132:
@emilyhuxng I've added a new interface which will will permit tables and totals tables to be joined to tables and totals tables.

niloc132:
@mofojed this isn't quite ready, but I wanted to discuss two things, see if we can find a way to finalize and ship it:
 * Count doesn't work the same way it did in DHE, which was to append the column name to `__Count`. From playing a bit with DHE, it will then render the count under each selected column (by default, this is "all of them"), but just as "Count". DHC doesn't let the client create a count aggregation on each column, which makes a bit more sense anyway, but the web UI doesn't presently know how to handle that. Do you want to have synthetic columns created to match DHE behavior, or can the UI handle this?
 * Distinct needs to be able to fetch the array of data that ... is distinct for that column - likely to be a big output, and while in DHE we could download it as a large string, in DHC the server will send an array of whatever type it is, which the JS API can't yet handle (but will be fixed in #188, in the near-ish future). Would it be reasonable to disable this in the short term?

niloc132:
Discussed with Ryan, and we could paper over both of these with custom columns (updateview to normalize arrays to string, and rename Count columns to ColA__Count, ColB__Count...), but ideally we wouldnt do too much of this if we can help it.

mofojed:
For the second, I'd prefer custom columns to match DHE behaviour, until we've improved the JS API usage of subscriptions.
For the first, I thought `COUNT` only counted non-null rows or something, but that doesn't seem to be the case. Certain that it's always the same on every column (basically just row count)? I think I'd still like it to behave like DHE, UI wise we'll need to change up a few things if it essentially just applies to the whole table instead.

niloc132:
In column statistics, the Count operation is only non-nulls (as distinct from Size), but not in aggregations (in either dhe or dhc).

niloc132:
Column renaming and toString-ing is complete, plus also removing requested aggregations that don't make sense and the server will reject.

mofojed (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2720
Conceptual: https://github.com/deephaven/deephaven.io/issues/2721
Reference: https://github.com/deephaven/deephaven.io/issues/2719
Blog: https://github.com/deephaven/deephaven.io/issues/2722

Issue #3392, opened 2023-02-01
Add `UpdateBy` `Rolling` features to QST / GRPC


lbooker42:
Properly add new operators of `UpdateBy` to the QST and GRPC structures.
- `RollingSum`
- `RollingGroup`
- `RollingAvg`
- `RollingMin`
- `RollingMax`

PR #2903 has this code stubbed out and this should be addressed ASAP

PR #3663, opened 2023-04-05
Added `Exponential Moving Sum` to `UpdateBy` operators

lbooker42:
Closes #2760

PR #3671, opened 2023-04-06
 Added `Exponential Moving Min`/`Exponential Moving Max` to `UpdateBy` operators

lbooker42:
This needs to merge after EMS PR (#3663) since this was created from the EMS branch.


replication/static/src/main/java/io/deephaven/replicators/ReplicateUpdateBy.java
```diff
@@ -93,10 +93,45 @@ public static void main(String[] args) throws IOException {
...
+                fixupInteger(f);
+            }
+            if (f.contains("Byte")) {
+                fixupByte(f);
+            }
+        }
+        ReplicatePrimitiveCode.floatToAllFloatingPoints(
+                "engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/em/FloatEMOperator.java");
+
+        files = ReplicatePrimitiveCode.charToAllButBooleanAndFloats(
```
nbauernfeind, line 109: Should we get rid of the EMA/EMS code generation now? Won't these generate files that are otherwise non-existent as of this PR?

java-client/session/src/main/java/io/deephaven/client/impl/UpdateByBuilder.java
```diff
@@ -118,6 +118,18 @@ public UpdateByColumn.UpdateBySpec visit(EmaSpec ema) {
                     .build();
         }
 
+        // TODO: complete properly (DHC ticket #3666)
```
nbauernfeind, line 121: I think this is the wrong issue number? Should these be #3392?

java-client/session/src/test/java/io/deephaven/client/impl/UpdateBySpecBuilderTest.java
```diff
@@ -41,6 +41,18 @@ public UpdateByColumn.UpdateBySpec visit(EmaSpec ema) {
                     .build();
         }
 
+        // TODO: complete properly (DHC ticket #3666)
```
nbauernfeind, line 44: I think this is the wrong issue number? Should these be #3392?

java-client/session/src/main/java/io/deephaven/client/impl/UpdateByBuilder.java
lbooker42: I am trying to close #3392 since it covers operators already merged.

nbauernfeind (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2475
Conceptual: https://github.com/deephaven/deephaven.io/issues/2473
Reference: https://github.com/deephaven/deephaven.io/issues/2476
Blog: https://github.com/deephaven/deephaven.io/issues/2474

PR #3672, opened 2023-04-10
Ported DHE ServerConfigValues

Zhou-Ziheng:


web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
```diff
@@ -113,10 +124,9 @@ public Promise<Void> login(@TsTypeRef(LoginCredentials.class) JsPropertyMap<Obje
 
         // fetch configs and check session timeout
         login.then(ignore -> getServerConfigValues()).then(configs -> {
-            for (String[] config : configs) {
-                if (config[0].equals("http.session.durationMs")) {
-                    ideConnection.connection.get().setSessionTimeoutMs(Double.parseDouble(config[1]));
-                }
+            if (configs.hasValue("http.session.durationMs")) {
```
mofojed, line 127: Should store this string as a constant

web/client-api/src/main/java/io/deephaven/web/client/api/ServerConfigValues.java
```diff
@@ -0,0 +1,13 @@
...
+ */
+package io.deephaven.web.client.api;
+
+/**
+ * A container for configuration values from the server.
+ */
+public interface ServerConfigValues {
+    String GRADLE_VERSION = "gradleVersion";
+    String VCS_VERSION = "vcsVersion";
+    String JAVA_VERSION = "javaVersion";
```
mofojed, line 12: These constants aren't used on Community at all, so doesn't make sense here.

web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
```diff
@@ -134,13 +144,8 @@ public Promise<Void> onConnected(@JsOptional Double timeoutInMillis) {
...
     }
 
-    public Promise<String[][]> getServerConfigValues() {
-        return getConfigs(
-                c -> ideConnection.connection.get().configServiceClient().getConfigurationConstants(
-                        new ConfigurationConstantsRequest(),
-                        ideConnection.connection.get().metadata(),
-                        c::apply),
-                ConfigurationConstantsResponse::getConfigValuesMap);
+    public Promise<JsServerConfigValues> getServerConfigValues() {
```
mofojed, line 147: We can't just change the API, as that will break existing consumers. Of course, doing this change would now match how it is on Enterprise...

mofojed (changes requested):
The motivation for this is to have a convenience class where we can `.getBooleanValue` or something and have the parsing handled by the API, rather than needing to parse everything from a `String`... but I don't think this is right (at least as is) since the constants we have (e.g. `.getVcsVersion`) aren't used in Community.

Was there another reason for doing this? I'm still inclined to keep this as just a `String[][]`, and leave it up to the client to parse depending on the setting; we could introduce a new `String[][] getConfigValues()` on both `IrisClient` and `CoreClient`, and deprecate `getServerConfigValues()` then, so we're at least using the same method on both Community and Enterprise; we could also have our own wrapper defined in `web-client-ui` (in `jsapi-utils` package perhaps) then just use that in both Community and Enterprise...

@niloc132 do you have any comments

web/client-api/src/main/java/io/deephaven/web/client/api/ServerConfigValues.java
niloc132: In DHC, we have a convention for versions to be read at runtime. By default in dhc this includes:
 * `java.version`
 * `deephaven.version`
 * `barrage.version`

PR #3675, opened 2023-04-10
Preserve both TS definition and JS native type for iterator

### LINKED ISSUE #3674: JS API error downloading tables
Broken from a change in the TS JS API stuff

Downloading any table from the table side panel triggers an exception on `snapshotIterator.next()` where the snapshot is an iterator on `UpdateEvent.added`

niloc132:
The GWT compiler was able to tell the non-native interface had no concrete implementations in the codebase, so replaced all invocations on instance methods with errors, so it is necessary to mark the type as native. However, this in turn prevents the TS definition generation from listing the type, since it is only responsible for declared types.

Fixes #3674


niloc132:
Testing this involves both checking the filed bug, and also verifying that the generated TypeScript includes the following:
```
export interface Iterator<T> {
	hasNext():boolean;
	next():IIterableResult<T>;
}
export interface IIterableResult<T> {
	value:T;
	done:boolean;
}
```

mofojed (approved):
Verified: 
- fixes our e2e tests
- downloading CSV from the side bar works (#3674)
- the outputted types includes `IIterableResult` and `Iterator`

PR #3683, opened 2023-04-11
Add `RollingCount` feature to `UpdateBy`

lbooker42:


nbauernfeind (commented):
I am curious if you should prefer storing all of the values into a `RollingByteBuffer` and just store a `0` if null and `1` if non-null (or `NULL_BYTE` isn't terrible if only slightly confusing).

For this particular operator, should we support `Object` columns? `BigDecimal` and `BigInteger` would not need impls if we supported `Object`.

lbooker42:
Great comments, have adopted your suggestions and use `ByteRingBuffer` for value storage and created `Object` operator.

nbauernfeind (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2472
Conceptual: https://github.com/deephaven/deephaven.io/issues/2471
Reference: https://github.com/deephaven/deephaven.io/issues/2469
Blog: https://github.com/deephaven/deephaven.io/issues/2470

PR #3686, opened 2023-04-12
Add `RollingStd` to `UpdateBy` operators

lbooker42:
Add rolling standard deviation calculation to the operator list for `UpdateBy`


engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingstd/CharRollingStdOperator.java
```diff
@@ -0,0 +1,146 @@
...
+                outputValues.set(outIdx, NULL_DOUBLE);
+            } else {
+                final int count = valueBuffer.size() - nullCount;
+                final double valueSquareSum = valueSquareBuffer.evaluate();
+                final double valueSum = valueBuffer.evaluate();
+
+                final double variance = valueSquareSum / (count - 1) - valueSum * valueSum / count / (count - 1);
+                final double std = Math.sqrt(variance);
+
+                outputValues.set(outIdx, std);
```
nbauernfeind, line 115: Is this correct? `count - 1` seems suspicious. A std-deviation of a single value is zero, not null (this impl will div by zero; but I see you protected against divByZero in `BigDecimal`/`BigInteger`).

I believe the variance should be: `valueSquareSum / count - (valueSum / count) ^ 2`

lbooker42: This matches the implementation of `var` found in the `Numeric` classes. See `io.deephaven.function.Numeric#var(io.deephaven.vector.ByteVector)` for an example.

I was under the impression we were returning `unbiased sample variance` as explained here: https://en.wikipedia.org/wiki/Bessel%27s_correction.  @chipkent, is this correct?

chipkent: The wikipedia article is correct.  Everyone thinks it is wrong initially.

nbauernfeind (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2488
Conceptual: https://github.com/deephaven/deephaven.io/issues/2490
Reference: https://github.com/deephaven/deephaven.io/issues/2489
Blog: https://github.com/deephaven/deephaven.io/issues/2487

PR #3691, opened 2023-04-13
Fix layouthints TS for SearchDisplayMode

mofojed:
- Fixes error introduced by conflicts from #3512 and #3567 

Comparison of outputted TS:
Before:
```
	export interface LayoutHints {
		get hiddenColumns():string[]|null|undefined;
		get frozenColumns():string[]|null|undefined;
		get columnGroups():ColumnGroup[]|null|undefined;
		get areSavedLayoutsAllowed():boolean;
		get frontColumns():string[]|null|undefined;
		get backColumns():string[]|null|undefined;
	}
```

After:
```
	export interface LayoutHints {
		get hiddenColumns():string[]|null|undefined;
		get searchDisplayMode():SearchDisplayModeType|null|undefined;
		get frozenColumns():string[]|null|undefined;
		get columnGroups():ColumnGroup[]|null|undefined;
		get areSavedLayoutsAllowed():boolean;
		get frontColumns():string[];
		get backColumns():string[]|null|undefined;
	}

	type SearchDisplayModeType = string;
	export class SearchDisplayMode {
		static readonly SEARCH_DISPLAY_DEFAULT:SearchDisplayModeType;
		static readonly SEARCH_DISPLAY_HIDE:SearchDisplayModeType;
		static readonly SEARCH_DISPLAY_SHOW:SearchDisplayModeType;
	}
```


web/client-api/src/main/java/io/deephaven/web/client/api/JsLayoutHints.java
```diff
@@ -15,8 +15,7 @@
 import java.util.Map;
 import java.util.stream.Collectors;
 
-@TsInterface
-@JsType(name = "LayoutHints", namespace = "dh")
+@TsName(namespace = "dh", name = "LayoutHints")
```
niloc132, line 18: This technically works, but when we effectively have an enum, we're preferring to move those to their own distinct type. So

```java
@JsTypeDef(jsType = "string")
@JsType(namespace = "dh")
public class SearchDisplayMode {
    public static final String SEARCH_DISPLAY_DEFAULT = "Default",
            SEARCH_DISPLAY_HIDE = "Hide",
            SEARCH_DISPLAY_SHOW = "Show";
}
```

Then these annotations can stay as-is, since it is more appropriate for this to be an interface, right?

niloc132 (commented):
I can't comment on the line, but please also add a @TsTypeRef(SearchDisplayMode.class) to the return type of getSearchDisplayMode()

PR #3693, opened 2023-04-13
Create PatternFilter

devinrsmith:
PatternFilter is an explicit filter construction based on a Pattern.

Consolidates all previous usages of PatterFilter, RegexFilter, and StringContainsFilter into a logically equivalent PatternFilter.

In support of more expressive filters for #3441


engine/table/src/main/java/io/deephaven/engine/table/impl/select/WhereFilterFactory.java
```diff
@@ -327,8 +333,11 @@ private static WhereFilter getSelectFilter(String colName, String quickFilter, Q
             return ComparableRangeFilter.makeBigDecimalRange(colName, quickFilter);
         } else if (filterMode != QuickFilterMode.NUMERIC) {
             if (colClass == String.class) {
-                return new StringContainsFilter(MatchFilter.CaseSensitivity.IgnoreCase, MatchFilter.MatchType.Regular,
-                        colName, quickFilter);
+                return new PatternFindFilter(
+                        ColumnName.of(colName),
+                        Pattern.compile(Pattern.quote(quickFilter), Pattern.CASE_INSENSITIVE),
```
devinrsmith, line 338: This matches the behavior before (essentially, simplifying the logic in the values.length == 1 case); I wonder, should we instead be doing:

```java
Pattern.compile(quickFilter, Pattern.CASE_INSENSITIVE | Pattern.LITERAL)
```
?

engine/table/src/main/java/io/deephaven/engine/table/impl/select/WhereFilterFactory.java
```diff
@@ -343,8 +352,11 @@ private static WhereFilter getSelectFilter(String colName, String quickFilter, Q
     private static WhereFilter getSelectFilterForAnd(String colName, String quickFilter, Class<?> colClass) {
         // AND mode only supports String types
         if (colClass.isAssignableFrom(String.class)) {
-            return new StringContainsFilter(MatchFilter.CaseSensitivity.IgnoreCase, MatchFilter.MatchType.Regular,
-                    colName, quickFilter);
+            return new PatternFindFilter(
+                    ColumnName.of(colName),
+                    Pattern.compile(Pattern.quote(quickFilter), Pattern.CASE_INSENSITIVE),
```
devinrsmith, line 357: Ditto wrt LITERAL.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/PatternFindFilter.java
```diff
@@ -0,0 +1,209 @@
...
+        if (column == null) {
+            throw new RuntimeException(String.format("Column '%s' doesn't exist in this table, available columns: %s",
+                    columnName.name(), tableDefinition.getColumnNames()));
+        }
+    }
+
+    @Override
+    public WritableRowSet filter(RowSet selection, RowSet fullSet, Table table, boolean usePrev) {
+        final ColumnSource<String> columnSource = table.getColumnSource(columnName.name());
+        return ChunkFilter.applyChunkFilter(selection, columnSource, usePrev, this);
```
rcaudy, line 107: We think we want to:
1. Cleanup references to `StringContainsFilter`
2. Rename `constructRegex` to say what it does
3. Combine these classes, and just have different chunk filter implementations.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/PatternFindFilter.java
```diff
@@ -0,0 +1,209 @@
...
+                    return Pattern.quote(converter == null ? val : converter.convertStringLiteral(val).toString());
+                });
+
+        // If the match is simple, includes -any- or includes -none- we can just use a simple
+        // regex of or'd values
+        if ((matchType == MatchType.Regular && internalDisjunctive) ||
+                (matchType == MatchType.Inverted && !internalDisjunctive)) {
+            regex = valueStream.collect(Collectors.joining("|"));
+        } else {
+            // Note that internalDisjunctive is -always- false here.
```
rcaudy, line 202: Can we try to understand why this comment holds?

engine/table/src/main/java/io/deephaven/engine/table/impl/select/WhereFilterFactory.java
devinrsmith: Choosing to keep the `.quote` for now, since I know it's equivalent to the previous logic.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/PatternFindFilter.java
devinrsmith: I believe it's because `StringContainsFilter` is really only meant for a WhereFilterFactory parsers of the form:

```
<ColumnName> [icase] [not] includes [any|all]<"String">
```

and was the only "external" caller setting `internalDisjunctive` (and testing).

I think it makes sense to move this logic explicitly to WhereFilterFactory to declutter the PatternFilter.

devinrsmith:
I've updated the code to _not_ include configurability for nullness within the PatternFilter itself.

I think ultimately in the expressive code that might want to invert a full PatternFilter, the logic will essentially need to be:

```
not(pattern_filter) ~= pattern_filter(!invert_pattern) OR isNull

File: engine/table/src/main/java/io/deephaven/engine/table/impl/select/PatternFilter.java [hunk 5]
```diff
@@ -0,0 +1,193 @@
...
+        results.setSize(0);
+        for (int ix = 0; ix < values.size(); ++ix) {
+            if (matcher(values.get(ix))) {
+                results.add(keys.get(ix));
+            }
+        }
+    }
+
+    private boolean matcher(String value) {
+        return value != null && (invertPattern ^ mode.test(pattern, value));
```
rcaudy, line 182: This seems fine. I would probably have elevated this up to 4 different static filter methods (for each combination of invertPattern and mode), and just delegated to the right one.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/WhereFilterFactory.java
```diff
@@ -379,6 +383,60 @@ public static WhereFilter[] getExpressionsWithQuickFilter(
...
+
+    private static String constructStringContainsRegex(
+            String[] values,
+            MatchType matchType,
+            boolean internalDisjunctive,
+            boolean removeQuotes,
+            String columnName) {
+        if (values == null || values.length == 0) {
+            throw new IllegalArgumentException(
+                    "constructStringContainsRegex must be created with at least one value parameter");
```
rcaudy, line 411:
```suggestion
                    "constructStringContainsRegex must be called with at least one value parameter");
```

rcaudy (commented):
Trivial comments. I'm happy enough to merge this, once someone looks at the Python for conventionality.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/PatternFindFilter.java
rcaudy: I agree. I still think a better comment, or just an assertion, would improve the code.

rcaudy (commented):
I'm happy with the current state. Ready to approve pending Python review.

py/server/deephaven/filters.py
```diff
@@ -88,21 +97,163 @@ def not_(filter_: Filter) -> Filter:
...
+
+    def __repr__(self):
+        return self.name
+
+
+def pattern(
+    mode: PatternMode,
+    col: str,
+    regex: str,
+    flags: Union[PatternFlag, List[PatternFlag]] = [],
```
jmao-denver, line 232:
```suggestion
    flags: Union[PatternFlag, List[PatternFlag]] = None,
```
Though not a big deal here, it is generally considered a safer practice to default mutable parameters to None

jmao-denver (changes requested):
Docstring format and other minor things

jmao-denver (approved):
LGTM, @chipkent This is a breaking change, so someone from dev-rel needs to take a look. as well.

devinrsmith:
Does anybody use RegexFilter? Afaict, our documentation refers explicitly to the engine impl, so it's out of date: https://deephaven.io/core/docs/how-to-guides/regex/#filter-with-regex

devinrsmith:
If necessary, we _could_ create a compatibility RegexFilter method (or keep as class)... but I would suggest not doing that unless we think it's valuable.

rcaudy (approved):

devinrsmith:
Updated that documentation is probably needed. We should probably delete and/or update https://deephaven.io/core/docs/how-to-guides/regex/#filter-with-regex

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2467
Conceptual: https://github.com/deephaven/deephaven.io/issues/2465
Reference: https://github.com/deephaven/deephaven.io/issues/2464
Blog: https://github.com/deephaven/deephaven.io/issues/2466

PR #3696, opened 2023-04-13
Add API support for databar column formatting

### LINKED ISSUE #3546: Add API support for `databar` column formatting
Need to develop the API for specifying a `databar` format for a column.
E.g. `t = source.format_columns("GPA = databar(GPA, Min(GPA), Max(GPA), ...options")` (format/API to be discussed and decided).

There are many options that can be specified:
- Negative divider placement: proportional (default), middle, colour only
- Value visibility: Show (default), hide
- Value placement: Adjacent (default), overlap
- Direction: Left-to-right (default), right-to-left
- Colors: Positive color, negative color, solid color, gradient, or transparency
  - Should be able to specify "theme" colors (e.g. theme positive, theme negative)
- Markers (bullet graph): Reference to another columns value - allow multiple markers, define colour for markers.

Should look at the implement of [Table.formatColumns](https://github.com/deephaven/deephaven-core/blob/fa3ababc68e43a560233858e847005d4d2994b79/engine/api/src/main/java/io/deephaven/engine/table/Table.java#L320) and [heatmap](https://github.com/deephaven/deephaven-core/blob/main/engine/table/src/main/java/io/deephaven/gui/color/Colors.java#L480).
Will also need to implement it in the [python API](https://github.com/deephaven/deephaven-core/blob/fa3ababc68e43a560233858e847005d4d2994b79/py/server/deephaven/table.py#LL1795).

When discussing with @rcaudy , he suggested using an `aggBy` on the original table and `min`/`max` on the desired column, then `natural_join` it on the original table, and subscribe to the result table to ensure accurate and consistent minima/maxima for use in the calculation.  A totals table does not have an inherent consistency guarantee there.

Python example showing the type of agg we'd do:
```
from deephaven import agg

size = 100
scale = 1000

])

agg_table = random_values_with_null_NaN.agg_by([agg.min_(cols=["__MyLong_min=MyLong"]), agg.max_(cols=["__MyLong_max=MyLong"])])
joined_table = random_values_with_null_NaN.natural_join(table=agg_table, on=[])
```

You can see this is working because when complete, the special formatting column will not yet be hidden by the JS API and will be visible in the UI. The formatting column should have a special prefix similar to other special formatting columns in [ColumnFormatting.java](https://github.com/deephaven/deephaven-core/blob/fa3ababc68e43a560233858e847005d4d2994b79/engine/table/src/main/java/io/deephaven/engine/util/ColumnFormatting.java#LL16C3-L16C3).

emilyhuxng:
Closes #3546 

Add `formatDatabar` functions.


py/server/deephaven/table.py
```diff
@@ -1842,6 +1842,16 @@ def format_row_where(self, cond: str, formula: str) -> Table:
         except Exception as e:
             raise DHError(e, "failed to color format rows conditionally.") from e
 
+    def format_databar(self, columnName: str, valueColumn: str, min: float, max: float, axis: str = "proportional",
```
mofojed, line 1845: As discussed in chat, I think you should default to `=None` for min/max, so you can use the value if specified, otherwise you just implicitly use the min/max from that column.
Also, I don't think you need `columnName` here? Just a param named `column` should be fine, since we're just formatting one column:
```suggestion
    def format_databar(self, column: str, min: float = None, max: float = None, axis: str = "proportional",
```

engine/table/src/main/java/io/deephaven/engine/table/impl/TableDefaults.java
```diff
@@ -286,6 +292,23 @@ default Table formatColumns(String... columnFormats) {
...
     }
 
+    @Override
+    @ConcurrentMethod
+    default Table formatDatabar(String columnName, String valueColumn, String axis, Double min, Double max,
+            String positiveColor, String negativeColor, String valuePlacement, String direction, Double opacity) {
+        String prefix = valueColumn + DATA_BAR_SUFFIX;
+        return this.naturalJoin(
+                this.aggBy(List.of(AggMin(prefix + "_MIN=" + valueColumn),
+                        AggMax(prefix + "_MAX=" + valueColumn)))
```
mofojed, line 302: Check if `min`/`max` are null first: if they're null, then use these aggregations (implicitly defining min/max based on the column); if they are not null, then just use the values they specified.

mofojed (changes requested):

py/server/deephaven/table.py
emilyhuxng: I thought they should be different for if the user wants the actual data bar column to be a different name.

mofojed: I think the same - we have the value display as part of the data bar. If they want to display it in a different column for some reason, then they should just copy the column (using a custom column).

dsmmcken: How would custom columns allow you to format string column A with a databar using ints from column B? Which is something heatmap does allow.

mofojed: I'm confused. How does that apply to databar? Are we displaying a databar alongside the string value from column A?

dsmmcken: That's a reasonable argument. We can skip having a separate value column, and require the data values and the column being formatted to just be column=

mofojed:
Superceded by https://github.com/deephaven/deephaven-core/pull/4181

PR #3698, opened 2023-04-13
Fix array index error on stale hover request

### LINKED ISSUE #3695: Hover throws an error if the requests are stale
In the autocomplete code, we check if the request is stale and if it is, we return an empty response from the completer to Java. We don't properly handle this case in hover which results in array access errors on the Java side that propagate to the client. The other request types return 2D arrays, so if that is empty it will not iterate and try to access the inner arrays

Add proper handling for hover to check if the array is empty before accessing its elements

mattrunyon:
Fixes #3695 

Doesn't need to modify the other autocomplete handlers since they expect a 2D array, and if the array is empty, it won't try to access any of the inner elements

Tested by hovering over a Python keyword and confirming no error is thrown. This was fine before as well, but I changed the Python hover function to return a string instead of a list containing 1 element that was a string

### DIFF
--- a/server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
+++ b/server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
@@ -280,18 +280,15 @@ public class PythonAutoCompleteObserver extends SessionCloseableObserver<AutoCom
                 // our java is 0-indexed lines and chars. jedi is 1-indexed lines and 0-indexed chars
                 // we'll keep that translation ugliness to the in-java result-processing.
                 pos.getLine() + 1, pos.getCharacter());
-        if (!result.isList()) {
+        if (!result.isString()) {
             throw new UnsupportedOperationException(
-                    "Expected list from jedi_settings.do_hover, got " + result.call("repr"));
+                    "Expected string from jedi_settings.do_hover, got " + result.call("repr"));
         }
 
-        // We expect [ contents ] as our result
         // We don't set the range b/c Jedi doesn't seem to give the word range under the cursor easily
         // Monaco in the web auto-detects the word range for the hover if not set
-        final List<PyObject> hover = result.asList();
-
         return GetHoverResponse.newBuilder()
-                .setContents(MarkupContent.newBuilder().setValue(hover.get(0).getStringValue()).setKind("markdown"))
+                .setContents(MarkupContent.newBuilder().setValue(result.getStringValue()).setKind("markdown"))
                 .build();
     }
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
```diff
@@ -280,18 +280,15 @@ private GetHoverResponse getHover(GetHoverRequest request, PyObject completer) {
...
         }
 
-        // We expect [ contents ] as our result
         // We don't set the range b/c Jedi doesn't seem to give the word range under the cursor easily
         // Monaco in the web auto-detects the word range for the hover if not set
-        final List<PyObject> hover = result.asList();
-
         return GetHoverResponse.newBuilder()
-                .setContents(MarkupContent.newBuilder().setValue(hover.get(0).getStringValue()).setKind("markdown"))
+                .setContents(MarkupContent.newBuilder().setValue(result.getStringValue()).setKind("markdown"))
```
devinrsmith, line 291: What does the web client do if contents is unset? Does it gracefully handle that case? It may be reasonable to have `do_hover` return `Optional[str]` instead and check `result.isNone()`. Alternatively, `AutoCompleteResponse.success == false`?

mattrunyon: Web client ignores it if set to empty string and I believe would ignore if unset. `AutoCompleteResponse.success == false` is used to denote an error or failure on the server. Right now that's if Jedi is disabled/missing or there's a server error handling the request. Out of date request isn't a failure really, just a "we don't care about this any more so we'll say nothing is found"

devinrsmith: I think we should add some documentation to the proto in these cases, b/c every time I forget what the semantics are. In a lot of other cases we throw proper grpc status exceptions when there is a server error. cc @niloc132

chipkent (approved):

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
niloc132: If unset, the MarkupContent.value property will be exposed to JS as `""`.

gRPC status code exceptions would be inappropriate here, as this is a bidirectional stream that effectively many different "unary" and streaming calls are passed on - any grpc status code error would kill the entire stream, rather than sending an individual response back as failure. This decision was made before we had a single websocket to handle all calls, so we could revise it, but it would require more locking and waiting on the server to avoid potential concurrency problems and racing messages (which a single grpc stream gives us for free).

More docs for the proto file would be okay, but @mattrunyon has structured this to follow the lsp spec rather than having the grpc wire format be the source of truth. With that said, backward and forward compatibility are still a concern, as we haven't 100% settled how to handle letting the JS API version drift from the server version.

niloc132 (approved):

PR #3711, opened 2023-04-19
Use embedded pattern flags instead of exposing via python wrapper

devinrsmith:
Upon digging into how this filter might be exposed over gRPC, it seems like keeping things simpler and relying on the embedded flag expressions is a better choice as it allows the full pattern to be specified by a string. For motivation, some manually invoked filters that the web ui client is currently constructing is using embedded flags:

```ts
filter.invoke(
  'matches',
  dh.FilterValue.ofString(`(?s)(?i).*\\Q${value}\\E.*`)
)
```

This is an example of setting the flags `DOTALL` (?s) and `CASE_INSENSITIVE` (?i), and using literal wrappers `\Q` / `\E`, to search if the column contains `${value}`.

If we find there are certain patterns in the future that need explicit flags, we can add them back in at that time.

Follow-up to #3693

### DIFF
--- a/py/server/deephaven/filters.py
+++ b/py/server/deephaven/filters.py
@@ -113,6 +113,9 @@ def pattern(
 ) -> Filter:
     """Creates a regular-expression pattern filter.
 
+    See https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/regex/Pattern.html for documentation on
+    the regex pattern.
+
     Args:
         mode (PatternMode): the mode
         col (str): the column name
--- a/py/server/deephaven/filters.py
+++ b/py/server/deephaven/filters.py
@@ -105,119 +105,10 @@ class PatternMode(Enum):
     """Matches any subsequence of the input against the pattern"""
 
 
-class PatternFlag(Enum):
-    CASE_INSENSITIVE = _JPattern.CASE_INSENSITIVE
-    """Enables case-insensitive matching.
-
-    By default, case-insensitive matching assumes that only characters in the US-ASCII charset are being matched.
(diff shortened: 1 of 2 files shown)


jmao-denver:
1. do we still need the 'inverted' flag? 
2. should we point to where a user can find the regex lang spec?

devinrsmith:
> 1. do we still need the 'inverted' flag?
> 
>     2. should we point to where a user can find the regex lang spec?

1. Yes, the `invert_pattern` is still meaningful as it's _not_ the same as just notting the filter.

2. I'm not sure what our thoughts are on linking to javadoc from pydoc? I'd be down for it, as opposed to trying to re-write the docs directly in pydoc. A link to https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/regex/Pattern.html seems good?

jmao-denver:
> > 1. do we still need the 'inverted' flag?
> >    
> >    2. should we point to where a user can find the regex lang spec?
> 
> 1. Yes, the `invert_pattern` is still meaningful as it's _not_ the same as just notting the filter.
> 2. I'm not sure what our thoughts are on linking to javadoc from pydoc? I'd be down for it, as opposed to trying to re-write the docs directly in pydoc. A link to https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/regex/Pattern.html seems good?

1. I am not against having the 'invert_pattern' flag, I just thought that regex itself has the ability to specify 'invert pattern'
2. I am fine with having the link in the Python doc string, which would clearly convey that the Java's regex syntax is what we support.

jmao-denver (approved):
LGTM

PR #3728, opened 2023-04-24
Add TreeTable.isClosed property

### LINKED ISSUE #3714: TreeTable.isClosed property
In the JS API, `TreeTable` objects don't appear to have an `isClosed` property like `Table` instances. We have a number of util code in the web UI that takes either a TreeTable or a Table. It would make things simpler to not have to check which table type we are working with for checking if the table is closed or not.

@mofojed pointed out that there is a `closed` variable that seems to always be false:
https://github.com/deephaven/deephaven-core/blob/226a428be5e746a2f1fbf07264a742769631c4c2/web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java#L358

niloc132:
Fixes #3714


mofojed (approved):

bmingles (approved):
Changes look good to me

Issue #3670, opened 2023-04-06
update_by ema_tick_decay should have floating point decay parameter


chipkent:
Currently, we have this definition:
```
def ema_tick_decay(time_scale_ticks: int, cols: Union[str, List[str]],
                   op_control: OperationControl = None) -> UpdateByOperation:
```
There is no reason that `time_scale` should be limited to integers.  It is more appropriate as a double.

Issue #2660, opened 2022-07-21
The tick decay rate for EMA should not be constrained to be integer only.


jmao-denver:
This was found by @chipkent in his review of the PR that wraps the new updateBy feature.

PR #3731, opened 2023-04-24
Add python wrappers and tests for new `update_by` operators

### LINKED ISSUE #3721: Python wrappers for new `UpdateBy` operators
Add python wrappers and tests for the following `UpdateBy` operators:
- Delta
- RollingWAvg
- Exponential Moving Sum 
- Exponential Min/Max
- RollingCount
- RollingStd

lbooker42:
Operators wrapped:
- Delta
- RollingWAvg
- Exponential Moving Sum
- Exponential Min/Max
- RollingCount
- RollingStd

When merged, will close #3721


py/server/tests/test_updateby.py
```diff
@@ -25,19 +27,39 @@ def tearDown(self) -> None:
...
+        emmin_time_decay(ts_col="Timestamp", time_scale=10, cols="ema_a = a"),
+        emmin_time_decay(ts_col="Timestamp", time_scale="00:00:00.001", cols="ema_c = c",
+                       op_control=em_op_ctrl),
+        # exponential moving maximum
+        emmax_tick_decay(time_scale_ticks=100, cols="ema_a = a"),
+        emmax_tick_decay(time_scale_ticks=100, cols="ema_a = a", op_control=em_op_ctrl),
+        emmax_time_decay(ts_col="Timestamp", time_scale=10, cols="ema_a = a"),
+        emmax_time_decay(ts_col="Timestamp", time_scale="00:00:00.001", cols="ema_c = c",
+                       op_control=em_op_ctrl),
+    ]
```
jmao-denver, line 59: It reads a bit weird when the class variables are scattered around in the code. The benefit of doing so is obvious in that it has the variables defined right before they are used in the testing methods. From a code organization POV it might be better to fold them all into a setupClass() method (see test_arrow.py for an example). The current per-test fixtures (in setUp()/tearDown()) could also be moved there since they don't need to be refreshed for each test case.
This is just a suggestion as it is not so clear cut better.

lbooker42: I like the suggestion, have refactored to move those lists to a common section.

py/server/deephaven/updateby.py
```diff
@@ -48,6 +49,15 @@ class BadDataBehavior(Enum):
     POISON = _JBadDataBehavior.POISON
     """Allow the bad data to poison the result. This is only valid for use with NaN"""
 
+class DeltaControl(Enum):
+    """An Enum defining ways to handle null values during update-by Delta operations."""
```
chipkent, line 53: This should include some clarification of what "Delta operations" are, since without that, a new user will not have any idea what this is and if they have arrived at the right docs to help them.

py/server/deephaven/updateby.py
```diff
@@ -164,6 +171,211 @@ def ema_time_decay(ts_col: str, time_scale: Union[int, str], cols: Union[str, Li
         raise DHError(e, "failed to create a time-decay EMA UpdateByOperation.") from e
 
 
+def ems_tick_decay(time_scale_ticks: int, cols: Union[str, List[str]],
```
chipkent, line 174: These new methods suffer from the same problem as the ticket I have reported twice.  The timescalee should *not* be int.  It should be float.

I wasn't able to find the other ticket to link to.  Maybe it was fixed?

py/server/deephaven/updateby.py
```diff
@@ -164,6 +171,211 @@ def ema_time_decay(ts_col: str, time_scale: Union[int, str], cols: Union[str, Li
...
+    except Exception as e:
+        raise DHError(e, "failed to create a tick-decay EMS UpdateByOperation.") from e
+
+def ems_time_decay(ts_col: str, time_scale: Union[int, str], cols: Union[str, List[str]],
+                   op_control: OperationControl = None) -> UpdateByOperation:
+    """Creates an EMS (exponential moving sum) UpdateByOperation for the supplied column names, using time as the
+    decay unit.
+
+    The formula used is
+        a = e^(-dt / time_scale_nanos)
```
chipkent, line 212: should `time_scale_nanos` be `time_scale` to match the args?  

Same comment for other methods that may be relevant.

py/server/deephaven/updateby.py
```diff
@@ -164,6 +171,211 @@ def ema_time_decay(ts_col: str, time_scale: Union[int, str], cols: Union[str, Li
...
+        if op_control is None:
+            return UpdateByOperation(j_updateby_op=_JUpdateByOperation.Ems(ts_col, time_scale, *cols))
+        else:
+            return UpdateByOperation(
+                j_updateby_op=_JUpdateByOperation.Ems(op_control.j_op_control, ts_col, time_scale, *cols))
+    except Exception as e:
+        raise DHError(e, "failed to create a time-decay EMS UpdateByOperation.") from e
+
+
+def emmin_tick_decay(time_scale_ticks: int, cols: Union[str, List[str]],
```
chipkent, line 242: These new methods suffer from the same problem as the ticket I have reported twice.  The timescale should *not* be int.  It should be float.

I wasn't able to find the other ticket to link to.  Maybe it was fixed?

py/server/deephaven/updateby.py
```diff
@@ -164,6 +171,211 @@ def ema_time_decay(ts_col: str, time_scale: Union[int, str], cols: Union[str, Li
...
+    except Exception as e:
+        raise DHError(e, "failed to create a tick-decay EM Min UpdateByOperation.") from e
+
+def emmin_time_decay(ts_col: str, time_scale: Union[int, str], cols: Union[str, List[str]],
+                   op_control: OperationControl = None) -> UpdateByOperation:
+    """Creates an EM Min (exponential moving minimum) UpdateByOperation for the supplied column names, using time as the
+    decay unit.
+
+    The formula used is
+        a = e^(-dt / time_scale_nanos)
```
chipkent, line 280: See comment on name mismatch.

py/server/deephaven/updateby.py
```diff
@@ -164,6 +171,211 @@ def ema_time_decay(ts_col: str, time_scale: Union[int, str], cols: Union[str, Li
...
+        if op_control is None:
+            return UpdateByOperation(j_updateby_op=_JUpdateByOperation.EmMin(ts_col, time_scale, *cols))
+        else:
+            return UpdateByOperation(
+                j_updateby_op=_JUpdateByOperation.EmMin(op_control.j_op_control, ts_col, time_scale, *cols))
+    except Exception as e:
+        raise DHError(e, "failed to create a time-decay EM Min UpdateByOperation.") from e
+
+
+def emmax_tick_decay(time_scale_ticks: int, cols: Union[str, List[str]],
```
chipkent, line 310: These new methods suffer from the same problem as the ticket I have reported twice.  The timescalee should *not* be int.  It should be float.

I wasn't able to find the other ticket to link to.  Maybe it was fixed?

py/server/deephaven/updateby.py
```diff
@@ -164,6 +171,211 @@ def ema_time_decay(ts_col: str, time_scale: Union[int, str], cols: Union[str, Li
...
+    except Exception as e:
+        raise DHError(e, "failed to create a tick-decay EM Max UpdateByOperation.") from e
+
+def emmax_time_decay(ts_col: str, time_scale: Union[int, str], cols: Union[str, List[str]],
+                     op_control: OperationControl = None) -> UpdateByOperation:
+    """Creates an EM Max (exponential moving maximum) UpdateByOperation for the supplied column names, using time as the
+    decay unit.
+
+    The formula used is
+        a = e^(-dt / time_scale_nanos)
```
chipkent, line 348: same

py/server/deephaven/updateby.py
```diff
@@ -270,6 +482,32 @@ def forward_fill(cols: Union[str, List[str]]) -> UpdateByOperation:
         raise DHError(e, "failed to create a forward fill UpdateByOperation.") from e
 
 
+def delta(cols: Union[str, List[str]], delta_control: DeltaControl = DeltaControl.VALUE_DOMINATES) -> UpdateByOperation:
+    """Creates a delta UpdateByOperation for the supplied column names.
```
chipkent, line 486: I have zero idea what this does based on the docs.

py/server/deephaven/updateby.py
```diff
@@ -759,4 +997,254 @@ def rolling_prod_time(ts_col: str, cols: Union[str, List[str]], rev_time: Union[
...
+    Raises:
+        DHError
+    """
+    try:
+        cols = to_sequence(cols)
+        return UpdateByOperation(j_updateby_op=_JUpdateByOperation.RollingCount(rev_ticks, fwd_ticks, *cols))
+    except Exception as e:
+        raise DHError(e, "failed to create a rolling count (tick) UpdateByOperation.") from e
+
+def rolling_count_time(ts_col: str, cols: Union[str, List[str]], rev_time: Union[int, str],
```
chipkent, line 1040: Looking at this, I'm noticing a naming difference that I want to point out before we have DnD users.  Here we have `rolling_count_tick` and `rolling_count_time` but earlier we had `ema_tick_decay` and `ema_time_decay` (or whatever it was).  Should we include the `_decay` in the other methods?

I'm not sure what my thoughts are, but I'm pointing it out before we get locked in forever.
@lbooker42 @jmao-denver

chipkent (changes requested):

py/server/deephaven/updateby.py
lbooker42: Here are the already reported issues: 
#2660 
#3670

py/server/deephaven/updateby.py
```diff
@@ -759,4 +997,254 @@ def rolling_prod_time(ts_col: str, cols: Union[str, List[str]], rev_time: Union[
...
+    except Exception as e:
+        raise DHError(e, "failed to create a rolling standard deviation (tick) UpdateByOperation.") from e
+
+def rolling_std_time(ts_col: str, cols: Union[str, List[str]], rev_time: Union[int, str],
+                       fwd_time: Union[int, str] = 0) -> UpdateByOperation:
+    """Creates a rolling standard deviation UpdateByOperation for the supplied column names, using time as the windowing unit. This
+    function accepts nanoseconds or time strings as the reverse and forward window parameters. Negative values are
+    allowed and can be used to generate completely forward or completely reverse windows. A row containing a null in
+    the timestamp column belongs to no window and will not have a value computed or be considered in the windows of
+    other rows.
```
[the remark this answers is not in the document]
lbooker42: Correct, have clarified in the documentation.

py/server/deephaven/updateby.py
```diff
@@ -759,4 +997,254 @@ def rolling_prod_time(ts_col: str, cols: Union[str, List[str]], rev_time: Union[
...
+    Raises:
+        DHError
+    """
+    try:
+        cols = to_sequence(cols)
+        rev_time = _JDateTimeUtils.expressionToNanos(rev_time) if isinstance(rev_time, str) else rev_time
+        fwd_time = _JDateTimeUtils.expressionToNanos(fwd_time) if isinstance(fwd_time, str) else fwd_time
+        return UpdateByOperation(j_updateby_op=_JUpdateByOperation.RollingWAvg(ts_col, rev_time, fwd_time, weight_col, *cols))
+    except Exception as e:
+        raise DHError(e, "failed to create a rolling weighted average (time) UpdateByOperation.") from e
```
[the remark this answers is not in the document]
lbooker42: EM Std is not going to make the EOM release.  There are some complexities in the operator that need managed before it can merge.

table-api/src/main/java/io/deephaven/api/updateby/spec/EmMinMaxSpec.java
```diff
@@ -20,7 +20,7 @@ public static EmMinMaxSpec of(OperationControl control, boolean isMax, WindowSca
     }
 
     public static EmMinMaxSpec of(boolean isMax, WindowScale windowScale) {
-        return ImmutableEmMinMaxSpec.builder().timeScale(windowScale).build();
+        return ImmutableEmMinMaxSpec.builder().isMax(isMax).timeScale(windowScale).build();
```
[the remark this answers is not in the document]
lbooker42: Yes, good catch.  Have added java-side tests of these interfaces as well.

py/server/deephaven/updateby.py
lbooker42: I dislike adding `decay` to Rolling operators; they do not decay like exponential ops. Actually I prefer to have the em function as `ema_ticks`/`ema_time`/`ems_ticks` etc. without `decay` attached.  Any user of these functions should already understand the decay mechanism built in.

jmao-denver: I agree with @lbooker42, I am for removing 'decay' from the function names.

chipkent: I was suggesting removing `_decay` for consistency -- not adding it to everything.  Shorter is better as long as we are clear.

chipkent: I see that my initial writeup was garbage, and I did say "include".

py/server/deephaven/updateby.py
```diff
@@ -483,7 +490,13 @@ def forward_fill(cols: Union[str, List[str]]) -> UpdateByOperation:
...
 
 def delta(cols: Union[str, List[str]], delta_control: DeltaControl = DeltaControl.VALUE_DOMINATES) -> UpdateByOperation:
-    """Creates a delta UpdateByOperation for the supplied column names.
+    """Creates a delta UpdateByOperation for the supplied column names. The Delta operation produces values by computing
+    the difference between the current value and the previous value. When the current value is null, this operation
+    will output null. When the current value is valid, the output will depend on the DeltaControl provided.
+
+        When delta_control is not provided or set to VALUE_DOMINATES, a valid value following a null value returns the
+            valid value.
+        When delta_control is set to NULL_DOMINATES, a valid value following a null value returns null.
```
chipkent: Now that we have some docs, I have a much better understanding of what is going on.  Here is my understanding:

1) `VALUE_DOMINATES` treats the prior null value as 0, so the current value is returned as the current delta.
2) `NULL_DOMINATES` poisons the result and results in `null` for the current delta.

If this interpretation is correct, I'm thinking about the default value and which option produces the least surprise.  I might argue that `NULL_DOMINATES` would be less surprising, but we should get feedback from @jmao-denver @jjbrosnan @alexpeters1208 .

chipkent (changes requested):

py/server/deephaven/updateby.py
lbooker42: You interpret correctly.  This was a port from DHE and I simply propagated the default of `VALUE_DOMINATES` without introspection.

lbooker42:
Added a commit that removes the `_decay` suffix from the python operator wrapping. There will need to be some followup in documentation since merging this would be a breaking change.

py/server/deephaven/updateby.py
lbooker42: This is being addressed in #3745, to merge after this.

chipkent (approved):

PR #3736, opened 2023-04-25
Add python filters is_null, is_not_null, and Filter.not_()

devinrsmith:


py/server/deephaven/filters.py
```diff
@@ -80,21 +88,33 @@ def and_(filters: List[Filter]) -> Filter:
...
+
+def is_null(col: str) -> Filter:
+    """Creates a new filter that evaluates to true when the col is null, and evaluates to false when col is not null.
+
+    Args:
+        col (str): the column name
+
+    Returns:
+        a new is-null Filter
     """
```
jmao-denver, line 116: Would ```Filter.from_("X == null")``` be equivalent to 'is_null("X")?

devinrsmith: It's logically the same, and will produce the same results, but the engine will do different things.

"X == null" will cause java to compile a filter kernel (ConditionFilter), whereas `is_null("X")` is able to construct a much more efficient `MatchFilter`.

devinrsmith: The more idiomatic way to express this via string would be `from_("isNull(X)")`, but that still results in ConditionFilter.

jmao-denver: How about adding an option in ```pattern``` to express the ``match null``` intention?

devinrsmith: "MatchFilter" is not a regex construction, it's a "match from this list of values" construction.

py/server/deephaven/filters.py
```diff
@@ -80,21 +88,33 @@ def and_(filters: List[Filter]) -> Filter:
...
         filter_ (Filter): the filter to negate with
 
     Returns:
-        a new Filter
+        a new not Filter
+    """
+    return filter_.not_()
+
+
+def is_null(col: str) -> Filter:
```
chipkent, line 108: For convenience, I would also consider adding `is_not_null` since it is so common, but you could convince me otherwise.

chipkent (changes requested):

py/server/deephaven/filters.py
devinrsmith: I think that is the consensus after I polled others as well, adding it.

jmao-denver (approved):
LGTM

PR #3763, opened 2023-04-28
Don't coalesce tables to get size or to build preview columns

### LINKED ISSUE #3748: Exporting Table Causes it to Be Coalesced
**Description**
Running "x = ..." with a command that produces an Uncoalesced table would not coalesce the table.

**Expected results**

Viewing an uncoalesced table does not force coalescing.

**Versions**
 - Deephaven: 0.21.0

niloc132:
Fixes #3748

### DIFF
--- a/proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
+++ b/proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
@@ -262,8 +262,8 @@ message ExportedTableCreationResponse {
   bool is_static = 5;
 
   // The current number of rows for this table. If this is negative, the table isn't coalesced, meaning the
-  // size isn't known without scanning partitions. Typically, the client should filter the data by the partitioned
-  // columns first.
+  // size isn't known without scanning partitions. Typically, the client should filter the data by the
+  // partitioning columns first.
   sint64 size = 6 [jstype=JS_STRING];
 
   // TODO: attributes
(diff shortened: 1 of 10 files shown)


niloc132:
To test this, I ran this python:
```python
t1 = empty_table(10).update_view(['I=i', 'Partition=0'])
t2 = empty_table(10).update_view(['I=i', 'Partition=1'])
from deephaven.parquet import write
write(t1, '/tmp/table/P=p0/table.parquet')
write(t2, '/tmp/table/P=p1/table.parquet')
from deephaven.parquet import read
t = read('/tmp/table')
```

I put a breakpoint in UncoalescedTable.coalesce() to confirm it was only called as expected - when the UI tries to pick a partition, it always results in a call to coalesce(), but that appears to be how the individual partition is read.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -261,7 +261,9 @@ message ExportedTableCreationResponse {
   // Whether or not this table might change.
   bool is_static = 5;
 
-  // The current number of rows for this table.
+  // The current number of rows for this table. If this is negative, the table isn't coalesced, meaning the
+  // size isn't known without scanning partitions. Typically, the client should filter the data by the partitioned
```
rcaudy, line 265:
```suggestion
  // size isn't known without scanning partitions. Typically, the client should filter the data by the partitioning
```

rcaudy (approved):
Good enough for me. One suggested comment change in the .proto, but meh.

nbauernfeind (approved):
LGTM

PR #3774, opened 2023-05-02
Wrap Table's awaitUpdate method

### LINKED ISSUE #3732: Wrap Table.awaitUpdate methods in Python
A couple of community users have run into situations where they need to reach the raw JPY wrapped awaitUpdate() to make sure that the Python code can depend on a ticking table being populated with some data.

jmao-denver:
Fixes #3732

### DIFF
--- a/py/server/tests/test_table.py
+++ b/py/server/tests/test_table.py
@@ -979,6 +979,20 @@ class TableTestCase(BaseTestCase):
         self.assertEqual(df.loc[0]['Col'], 1)
         self.assertTrue(rt.columns[0].data_type == dtypes.int32)
 
+    def test_await_update(self):
+        with self.assertRaises(DHError):
+            empty_table(10).await_update()
+
+        time_t = time_table("00:00:00.001")
+        updated = time_t.await_update()
+        self.assertTrue(updated)
+        updated = time_t.update("X = i % 2").where("X = 2").await_update(0)
+        self.assertFalse(updated)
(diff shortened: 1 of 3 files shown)


py/server/tests/test_table.py
```diff
@@ -979,6 +979,20 @@ def do_something(p=None):
         self.assertEqual(df.loc[0]['Col'], 1)
         self.assertTrue(rt.columns[0].data_type == dtypes.int32)
 
+    def test_await_update(self):
+        with self.assertRaises(DHError):
+            empty_table(10).await_update()
+
+        time_t = time_table("00:00:00.001")
+        updated = time_t.await_update()
```
chipkent, line 987: Don't some of the other tests use `await_update` via `jpy`?  If so, these should be changed to use the new functionality instead of directly though `jpy`.

chipkent (changes requested):

py/server/tests/test_table.py
jmao-denver: No, it is a util function in the base test class.

chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2549
Conceptual: https://github.com/deephaven/deephaven.io/issues/2552
Reference: https://github.com/deephaven/deephaven.io/issues/2550
Blog: https://github.com/deephaven/deephaven.io/issues/2551

PR #3782, opened 2023-05-03
Correctly set custom headers when looking up auth constants

### LINKED ISSUE #3781: CoreClient.getAuthConfigValues does not set custom headers provided by the client in ConnectOptions
**Description**

CoreClient.getAuthConfigValues does not set custom headers provided by the client in ConnectOptions. This manifests in DnD when running behind Envoy.

CoreClient uses `connectOptions` provided by the client to make the `IdeConnection`: https://github.com/deephaven/deephaven-core/blob/21c680c18cef6e646542a1d4736ffa2a40ecbbfc/web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java#L46

That allows the client to set custom headers: https://github.com/deephaven/deephaven-core/blob/21c680c18cef6e646542a1d4736ffa2a40ecbbfc/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java#L276

However, the request for `client.getAuthConfigValues` never uses those connection options: https://github.com/deephaven/deephaven-core/blob/21c680c18cef6e646542a1d4736ffa2a40ecbbfc/web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java#LL82C93-L82C93

niloc132:
Fixes #3781

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
@@ -1,7 +1,9 @@
 package io.deephaven.web.client.api;
 
 import com.vertispan.tsdefs.annotations.TsTypeRef;
+import elemental2.core.JsObject;
 import elemental2.promise.Promise;
+import io.deephaven.javascript.proto.dhinternal.browserheaders.BrowserHeaders;
 import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.config_pb.AuthenticationConstantsRequest;
 import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.config_pb.AuthenticationConstantsResponse;
 import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.config_pb.ConfigValue;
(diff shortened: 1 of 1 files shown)


web/client-api/src/main/java/io/deephaven/web/client/api/CoreClient.java
```diff
@@ -77,12 +79,19 @@ public String getServerUrl() {
...
 
     public Promise<String[][]> getAuthConfigValues() {
-        return getConfigs(
-                // Explicitly creating a new client, and not passing auth details, so this works pre-connection
-                c -> new ConfigServiceClient(getServerUrl(), CLIENT_OPTIONS).getAuthenticationConstants(
-                        new AuthenticationConstantsRequest(),
-                        c::apply),
-                AuthenticationConstantsResponse::getConfigValuesMap);
+        return ideConnection.getConnectOptions().then(options -> {
+            BrowserHeaders metadata = new BrowserHeaders();
```
mofojed, line 83: I'm a little confused... you're setting the headers in this metadata, but then this metadata isn't getting used anywhere?

niloc132: Fixed.

mofojed (approved):

mofojed:
Verified in DnD

PR #3788, opened 2023-05-03
added info to AggSpec javadocs

elijahpetty:
Added descriptions and param info to the AggSpec interface javadocs.

### DIFF
--- a/table-api/src/main/java/io/deephaven/api/agg/spec/AggSpec.java
+++ b/table-api/src/main/java/io/deephaven/api/agg/spec/AggSpec.java
@@ -20,6 +20,8 @@ import java.util.Collection;
 public interface AggSpec {
 
     /**
+     * Produces a single-row table with the absolute some of each column.
+     *
      * @return {@link AggSpecAbsSum#of()}
      */
     static AggSpecAbsSum absSum() {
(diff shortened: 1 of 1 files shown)


rcaudy (commented):
None of these docs are particularly accurate. They cover what kind of aggregation is being specified, but they don't correctly match the usage for agg specs, which are not only used in `aggAllBy`, but also `ColumnAggregation`.

table-api/src/main/java/io/deephaven/api/agg/spec/AggSpec.java
```diff
@@ -20,150 +20,224 @@
 public interface AggSpec {
 
     /**
+     * Produces a single-row table with the absolute some of each column.
```
rcaudy, line 23: This is coming from `TableOperations.absSumBy()`, and isn't really accurate for the `AggSpec`.

table-api/src/main/java/io/deephaven/api/agg/spec/AggSpec.java
```diff
@@ -20,150 +20,224 @@
...
+     * Only works for numeric or Comparable input types.
+     *
      * @return {@link AggSpecMedian#of()}
      */
     static AggSpecMedian median() {
         return AggSpecMedian.of();
     }
 
     /**
+     * Specifier for a column aggregation that produces a median value from the input column's values for each group.
```
rcaudy, line 180:
```suggestion
     * Specifier for a {@link ColumnAggregation} that produces a median value from the input column's values for each group.
```

rcaudy, line 180: I like this one as a template to apply to some of the ones that aren't quite up to this standard.

PR #3789, opened 2023-05-03
Added `Exponential Moving Standard Deviation` to `UpdateBy` operators 

lbooker42:


engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/emstd/BigIntegerEmStdOperator.java
```diff
@@ -0,0 +1,146 @@
...
+                    if (input == null) {
+                        handleBadData(this, true);
+                    } else {
+                        final BigDecimal decInput = new BigDecimal(input);
+                        if (curEma == null) {
+                            curEma = decInput;
+                            curVariance = BigDecimal.ZERO;
+                            curVal = null;
+                        } else {
+                            //  incremental variance = alpha * (prevVariance + (1 − alpha) * (x − prevEma)^2)
```
chipkent, line 49: For these equations, I think it is worth putting a note with the pdf URL and title.  This applies to all files.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/emstd/CharEmStdOperator.java
```diff
@@ -0,0 +1,166 @@
...
+import io.deephaven.engine.table.impl.util.RowRedirection;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import static io.deephaven.util.QueryConstants.*;
+
+public class CharEmStdOperator extends BasePrimitiveEmStdOperator {
+    public final ColumnSource<?> valueSource;
+    // region extra-fields
+    // endregion extra-fields
```
chipkent, line 21: needed?

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/emstd/CharEmStdOperator.java
```diff
@@ -0,0 +1,166 @@
...
+                             final double windowScaleUnits,
+                             final ColumnSource<?> valueSource,
+                             final boolean sourceRefreshing
+                             // region extra-constructor-args
+                             // endregion extra-constructor-args
+    ) {
+        super(pair, affectingColumns, rowRedirection, control, timestampColumnName, windowScaleUnits, sourceRefreshing);
+        this.valueSource = valueSource;
+        // region constructor
+        // endregion constructor
```
chipkent, line 158: needed?

table-api/src/main/java/io/deephaven/api/updateby/UpdateByOperation.java
chipkent, line 1: I think the std is the key operation.  At some point, we will want to backfill all of the operations with variance as well.

chipkent (dismissed):

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/emstd/CharEmStdOperator.java
lbooker42: Needed by `ReplicateUpdateBy.java` for `Byte` specialization.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/emstd/BaseBigNumberEmStdOperator.java
```diff
@@ -0,0 +1,248 @@
...
+
+    protected final WritableColumnSource<BigDecimal> emaSource;
+    protected final WritableColumnSource<BigDecimal> maybeEmaInnerSource;
+
+    public interface EmFunction {
+        BigDecimal apply(BigDecimal prevVal, BigDecimal curVal, BigDecimal alpha, BigDecimal oneMinusAlpha);
+    }
+
+    public abstract class Context extends BaseObjectUpdateByOperator<BigDecimal>.Context {
+        ObjectChunk<T, ? extends Values> objectValueChunk;
```
nbauernfeind, line 49: should all of these members be `protected`?

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/emstd/BaseBigNumberEmStdOperator.java
```diff
@@ -0,0 +1,248 @@
...
+                maybeEmaInnerSource.startTrackingPrevValues();
+            }
+        }
+    }
+
+    // region Shifts
+    @Override
+    public void applyOutputShift(@NotNull final RowSet subRowSetToShift, final long delta) {
+        super.applyOutputShift(subRowSetToShift, delta);
+        if (emaSource != null) {
```
nbauernfeind, line 208: Should this instead be: `rowRedirection == null`? I suspect this code path throws a ClassCastException when a rowRedirection is in use.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/emstd/BasePrimitiveEmStdOperator.java
```diff
@@ -0,0 +1,208 @@
...
+                maybeEmaInnerSource.startTrackingPrevValues();
+            }
+        }
+    }
+
+    // region Shifts
+    @Override
+    public void applyOutputShift(@NotNull final RowSet subRowSetToShift, final long delta) {
+        super.applyOutputShift(subRowSetToShift, delta);
+        if (emaSource != null) {
```
nbauernfeind, line 164: `rowRedirection == null`?

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/emstd/BasePrimitiveEmStdOperator.java
```diff
@@ -0,0 +1,208 @@
...
+            if (rowRedirection != null) {
+                ((WritableSourceWithPrepareForParallelPopulation) maybeEmaInnerSource)
+                        .prepareForParallelPopulation(changedRows);
+            } else {
+                ((WritableSourceWithPrepareForParallelPopulation) emaSource).prepareForParallelPopulation(changedRows);
+            }
+        }
+    }
+
+    void handleBadData(@NotNull final Context ctx,
```
nbauernfeind, line 183: `protected` methods?

engine/table/src/test/java/io/deephaven/engine/table/impl/updateby/TestEmStd.java
```diff
@@ -0,0 +1,1277 @@
...
+import static io.deephaven.function.Basic.isNull;
+import static io.deephaven.time.DateTimeUtils.MINUTE;
+import static io.deephaven.time.DateTimeUtils.convertDateTime;
+import static io.deephaven.util.QueryConstants.*;
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.fail;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@Category(OutOfBandTest.class)
+public class TestEmStd extends BaseUpdateByTest {
```
nbauernfeind, line 52: Can you add a test that hits the `rowRedirection != null` code paths?

table-api/src/main/java/io/deephaven/api/updateby/UpdateByOperation.java
```diff
@@ -607,6 +607,149 @@ static UpdateByOperation EmMax(OperationControl control, String timestampColumn,
...
+    /**
+     * Create an {@link EmStdSpec exponential moving standard deviation} for the supplied column name pairs, using ticks
+     * as the decay unit. Uses the default OperationControl settings.
+     * <p>
+     * The formula used is
+     * </p>
+     *
+     * <pre>
+     *     a = e^(-1 / tickDecay)
+     *     variance = a * (prevVariance + (1 − a) * (x − prevEma)^2)
```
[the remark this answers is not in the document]
lbooker42: Double-checked, the parentheses are correct per the function implemented.  We could distribute the `a` value, but this matches both the implementation and the referenced paper.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/emstd/BaseBigNumberEmStdOperator.java
lbooker42: A little bit complicated, but this is correct as is.  `applyOutputShift()` is only called when `rowRedirection == null` because the redirection is managed at a higher layer in `UpdateBy.java`.  Here is the caller, it confirms that the output table is not redirected before calling:
```
                        if (!redirHelper.isRedirected() && upstream.shifted().nonempty()) {
                            // Shift the non-redirected output sources now, after parallelPopulation.
                            upstream.shifted().apply((begin, end, delta) -> {
                                try (final RowSet subRowSet = source.getRowSet().prev().subSetByKeyRange(begin, end)) {
                                    for (UpdateByOperator op : win.getOperators()) {
                                        op.applyOutputShift(subRowSet, delta);
                                    }
                                }
                            });
                        }
```

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/emstd/BasePrimitiveEmStdOperator.java
lbooker42: see above

engine/table/src/test/java/io/deephaven/engine/table/impl/updateby/TestEmStd.java
lbooker42: See above, can't actually be reached when `rowRedirection != null`.  Currently have `testZeroKeyAppendOnlyRedirected()`, `testZeroKeyGeneralRedirected()`, `testBucketedAppendOnlyRedirected()`, `testBucketedGeneralRedirected()` that test the redirected code paths.

nbauernfeind (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2595
Conceptual: https://github.com/deephaven/deephaven.io/issues/2596
Reference: https://github.com/deephaven/deephaven.io/issues/2597
Blog: https://github.com/deephaven/deephaven.io/issues/2598

PR #3796, opened 2023-05-04
Add Expression, Filter, Literal Strings.of tests

devinrsmith:
Also, makes the encapsulate methods in Strings private.

### DIFF
--- a/table-api/src/main/java/io/deephaven/api/Strings.java
+++ b/table-api/src/main/java/io/deephaven/api/Strings.java
@@ -38,6 +38,10 @@ public class Strings {
         return rawString.value();
     }
 
+    public static String of(RawString rawString, boolean invert) {
+        return of(rawString, false, invert);
+    }
+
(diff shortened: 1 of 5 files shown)


table-api/src/main/java/io/deephaven/api/Strings.java
```diff
@@ -59,20 +63,14 @@ public static String of(FilterIn in) {
         return in.toString();
```
rcaudy, line 63: This seems suboptimal, and worse when inverted with !
I think it should look like something we would parse to a `MatchFilter`. Which means `not in` as an operator when inverted.

devinrsmith: Note: we aren't currently using Strings for actually printing out anything useful - it's really as helper for label building. When we print out WhereFilter for the engine, it will use the toString of MatchFilter.

rcaudy (approved):

PR #3802, opened 2023-05-05
Custom autocomplete API

niloc132:
Permits downstream applications to add specific completions to Java autocomplete calls. This is an internal API at this time, and subject to change.

### DIFF
--- a/open-api/lang-tools/src/main/java/io/deephaven/lang/completion/ChunkerCompleter.java
+++ b/open-api/lang-tools/src/main/java/io/deephaven/lang/completion/ChunkerCompleter.java
@@ -50,7 +50,7 @@ public class ChunkerCompleter implements CompletionHandler {
     private ParsedDocument doc;
 
     public ChunkerCompleter(final Logger log, VariableProvider variables) {
-        this(log, variables, new CompletionLookups(() -> new CustomCompletion() {}));
+        this(log, variables, new CompletionLookups(Collections.emptySet()));
     }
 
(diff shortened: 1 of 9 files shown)


open-api/lang-tools/src/main/java/io/deephaven/lang/completion/ChunkerCompleter.java
```diff
@@ -784,7 +784,7 @@ private void invokeComplete(
         // for now, this will be a naive replacement, but later we'll want to look at _where_
         // in the invoke the cursor is; when on the ending paren, we'd likely want to look at
         // whether we are the argument to something, and if so, do a type check, and suggest useful .coercions().
-        String name = node.getName().trim();
+        String name = node.getName();
```
devinrsmith, line 787: Assuming this is just a cleanup?

devinrsmith (approved):
Structurally, looks pretty straightforward. Can't say I know the guts of ChunkerCompleter, but everything (except for trim) looks to be like a no-op by default, so works for me.

open-api/lang-tools/src/main/java/io/deephaven/lang/completion/ChunkerCompleter.java
niloc132: Yep - I was going to go further and remove name from a few spots, but just left it as not including `String name` as a param to the custom autocomplete interface - it isn't really needed, the invoke node will already trim on the way in/out.

niloc132:
Tested by confirming that some basic autocomplete in groovy still works - columns are suggested for existing tables.

PR #3815, opened 2023-05-08
Only test for true values in barrage column attributes

### LINKED ISSUE #3814: NullPointerException in JS API using one_click
Run this code in the python console:
```python
# create link from table to plot on "mystring column"
from deephaven.plot.selectable_dataset import one_click

simple_ticking = time_table("00:00:01").update([
        "MyString=new String(`a`+(char) ((i%3)+97))",
        "MyInt=new Integer(i)",
])

toc = one_click(simple_ticking, by=["MyString"], require_all_filters=True)
p = Figure().plot_xy(series_name="Distance", t=toc, x="Timestamp", y="MyInt").show()
```
Then, click the suggested "Add input filters" button, and enter a value present in the table, like `aa`.

Expected: Data will render in the figure.

Actual: Error in the browser console:
```
dh-core.js:1937 Uncaught (in promise) TypeError: java.lang.NullPointerException
    at NullPointerException.createError_0 [as createError] (dh-core.js:1937:10)
    at NullPointerException.initializeBackingError (dh-core.js:1233:46)
    at NullPointerException.Throwable (dh-core.js:1171:8)
    at NullPointerException.Exception (dh-core.js:1245:16)
    at NullPointerException.RuntimeException (dh-core.js:1262:16)
    at NullPointerException.JsException (dh-core.js:1898:23)
    at new NullPointerException (dh-core.js:1924:18)
    at Object.checkCriticalNotNull (dh-core.js:4711:16)
    at Object.$equals_12 (dh-core.js:1971:13)
    at Object.readColumnDefinitions (dh-core.js:16352:47)
    at Object.readTableDefinition (dh-core.js:16390:19)
    at Object.$applyTableCreationResponse (dh-core.js:32547:42)
    at Object.$lambda$33_2 (dh-core.js:32864:9)
    at Function.onInvoke_281 (dh-core.js:33553:16)
    at lambda (dh-core.js:173:22)
```

This is an error when invoking the JS API method `newTable`, which is used to create the keys to be selected in the PartitionedTable to pick the appropriate table. This will affect other usages of PartitionedTable as well.

niloc132:
Fixes #3814

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/barrage/WebBarrageUtils.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/barrage/WebBarrageUtils.java
@@ -100,12 +100,12 @@ public class WebBarrageUtils {
             cols[i].setStyleColumnName(fieldMetadata.get("styleColumn"));
 
             if (fieldMetadata.containsKey("inputtable.isKey")) {
-                cols[i].setInputTableKeyColumn(Boolean.parseBoolean(fieldMetadata.get("inputtable.isKey")));
+                cols[i].setInputTableKeyColumn("true".equals(fieldMetadata.get("inputtable.isKey")));
             }
 
             cols[i].setDescription(fieldMetadata.get("description"));
 
-            cols[i].setPartitionColumn(fieldMetadata.get("isPartitioning").equals("true"));
+            cols[i].setPartitionColumn("true".equals(fieldMetadata.get("isPartitioning")));
 
             cols[i].setHierarchicalExpandByColumn(
                     "true".equals(fieldMetadata.get("hierarchicalTable.isExpandByColumn")));


web/client-api/src/main/java/io/deephaven/web/client/api/barrage/WebBarrageUtils.java
```diff
@@ -100,12 +100,12 @@ public static ColumnDefinition[] readColumnDefinitions(Schema schema) {
             cols[i].setStyleColumnName(fieldMetadata.get("styleColumn"));
 
             if (fieldMetadata.containsKey("inputtable.isKey")) {
-                cols[i].setInputTableKeyColumn(Boolean.parseBoolean(fieldMetadata.get("inputtable.isKey")));
+                cols[i].setInputTableKeyColumn("true".equals(fieldMetadata.get("inputtable.isKey")));
```
mofojed, line 103: Why not use `Boolean.parseBoolean`? Seems clear and also doesn't restrict to case sensitivity:
```
    public static boolean parseBoolean(String s) {
        return "true".equalsIgnoreCase(s);
    }
```

niloc132: That's fair, could rewrite all of the - I just picked what we were being consistent around and made it 100% so, instead of ~80% (with one equivalent call, and one wrong).

mofojed: Also fair

mofojed (approved):

PR #3820, opened 2023-05-09
Use Jetty image for client integration tests, docs

### LINKED ISSUE #2943: Stop depending on deephaven/server-netty for other aspects of the build
Integration tests and producing docs should all use the jetty builds going forward.

niloc132:
Also optionally exposes port for dh-in-docker.

Fixes #2943

### DIFF
--- a/buildSrc/src/main/groovy/io/deephaven/tools/docker/DeephavenInDockerExtension.groovy
+++ b/buildSrc/src/main/groovy/io/deephaven/tools/docker/DeephavenInDockerExtension.groovy
@@ -1,14 +1,17 @@
 package io.deephaven.tools.docker
 
 import com.bmuschko.gradle.docker.tasks.container.DockerCreateContainer
+import com.bmuschko.gradle.docker.tasks.container.DockerInspectContainer
 import com.bmuschko.gradle.docker.tasks.container.DockerRemoveContainer
 import com.bmuschko.gradle.docker.tasks.container.DockerStartContainer
 import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
 import com.bmuschko.gradle.docker.tasks.network.DockerCreateNetwork
 import com.bmuschko.gradle.docker.tasks.network.DockerRemoveNetwork
+import com.github.dockerjava.api.command.InspectContainerResponse
 import groovy.transform.CompileStatic
 import org.gradle.api.Project
 import org.gradle.api.Task
+import org.gradle.api.provider.ListProperty
 import org.gradle.api.provider.MapProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.TaskProvider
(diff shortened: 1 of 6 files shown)


niloc132:
To test this, I ran `./gradlew smoke -x :docker-server:prepareDocker`, then confirmed that there are no results for ` docker images | grep netty`.

buildSrc/src/main/groovy/io/deephaven/tools/docker/DeephavenInDockerExtension.groovy
```diff
@@ -82,6 +97,14 @@ public abstract class DeephavenInDockerExtension {
             task.containerId.set containerName.get()
         }
 
+        portTask = project.tasks.register('waitForPort', DockerInspectContainer) {task ->
+            task.dependsOn healthyTask
+            task.containerId.set containerName.get()
+            task.onNext { InspectContainerResponse inspect ->
+                getPort().set(Integer.parseInt(((InspectContainerResponse) inspect).getNetworkSettings().ports.bindings.values().first()[0].hostPortSpec))
+            }
+        }
```
devinrsmith, line 106: Is this meant to be an ad-hoc / user-run task? I don't see anything that depends on it.

I'm assuming the end-2-end idea is that we can instead do:

```diff
diff --git a/py/client/build.gradle b/py/client/build.gradle
index 2f61f8b30c..68511d713a 100644
--- a/py/client/build.gradle
+++ b/py/client/build.gradle
@@ -93,7 +93,7 @@ tasks.getByName('check').dependsOn(Docker.registerDockerTask(project, 'testPyCli
             into 'project/'
         }
     }
-    containerDependencies.dependsOn = [deephavenDocker.healthyTask]
+    containerDependencies.dependsOn = [deephavenDocker.portTask]
     containerDependencies.finalizedBy = deephavenDocker.endTask
     network = deephavenDocker.networkName.get()
     dockerfile {
@@ -105,7 +105,7 @@ tasks.getByName('check').dependsOn(Docker.registerDockerTask(project, 'testPyCli
                       pip3 install --upgrade pip; \\
                       pip3 install -r requirements-dev.txt'''
         environmentVariable 'DH_HOST', deephavenDocker.containerName.get()
-        environmentVariable 'DH_PORT', '10000'
+        environmentVariable 'DH_PORT', "${deephavenDocker.getPort()}"
     }
     parentContainers = [ Docker.registryTask(project, 'python') ]
     entrypoint = ['python', '-m', 'xmlrunner', 'discover', 'tests', '-v', '-o', '/out/report']
```

niloc132: That's the idea, but what you have there won't work since the python client tests are run in _another_ docker container, and the getPort() exposed here is the exposed port on the host, not the container.

Instead, this can be used for JS API tests, or java-client tests, in pretty much the way that you described above.

devinrsmith (approved):

jmao-denver (approved):
LGTM

PR #3828, opened 2023-05-11
Add pandas version requirement 1.5.0 or later

### LINKED ISSUE #3826: Require pandas be 1.5.0 or later for Deephaven Python API
due to changes to support Pandas 2.0

jmao-denver:
Fixes #3826

### DIFF
--- a/py/server/setup.py
+++ b/py/server/setup.py
@@ -56,7 +56,7 @@ setup(
         'jpy>=0.13.0',
         'deephaven-plugin',
         'numpy',
-        'pandas',
+        'pandas>=1.5.0',
         'pyarrow',
         # TODO(deephaven-core#3082): Remove numba dependency workarounds
         # It took 6 months for numba to support 3.11 after it was released, we want to make sure deephaven-core will be


devinrsmith (approved):
I'm not against this change, I just don't know if there are any potential conflicts other users may have where they need to run < 1.5.0. Pandas 1.5.0 was released in September 2022, so it's _relatively_ new. I'm not sure if there are feasible ways to make the code work w/ < 1.5 and 2+; and I'm not sure it's worth the effort, but just trying to be explicit and note it.

jmao-denver:
> I'm not against this change, I just don't know if there are any potential conflicts other users may have where they need to run < 1.5.0. Pandas 1.5.0 was released in September 2022, so it's _relatively_ new. I'm not sure if there are feasible ways to make the code work w/ < 1.5 and 2+; and I'm not sure it's worth the effort, but just trying to be explicit and note it.

yeah, I actually looked at what you asked about first, but didn't see a way to support older pandas without making the code very messy. That combined with the fact that  pandas 1.5.0 introduced no breaking changes to the API made me feel quite comfortable with this change.

chipkent:
What is pushing this change?  Just making Pandas 2.0 work?  Specifically, what are we needing that only comes in in 1.5?

jmao-denver:
Community slack user Raaghlur ran into the same problem when upgrading to 0.24.1 as Pete's.

Issue #3827, opened 2023-05-11
Lower latency page loading (ide, iframe)


devinrsmith:
Currently, the jetty server only enables caching of `/ide/assets/*`, and disables caching on `/ide/*` and `/jsapi/*`:

```java
        // For the Web UI, cache everything in the static folder
        // https://create-react-app.dev/docs/production-build/#static-file-caching
        context.addFilter(NoCacheFilter.class, "/ide/*", EnumSet.noneOf(DispatcherType.class));
        context.addFilter(NoCacheFilter.class, "/jsapi/*", EnumSet.noneOf(DispatcherType.class));
        context.addFilter(CacheFilter.class, "/ide/assets/*", EnumSet.noneOf(DispatcherType.class));
        context.addFilter(DropIfModifiedSinceHeader.class, "/*", EnumSet.noneOf(DispatcherType.class));
```

This can cause excessive bandwidth and loading times, as `/jsapi/dh-core.js` and `/jsapi/dh-internal.js` are requested fresh on every page load. Furthermore, nothing under `/iframe/` is cached.

PR #3829, opened 2023-05-11
Add caching directives for /iframe/table/ and /iframe/chart/

devinrsmith:
Partial #3827

### DIFF
--- a/server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
+++ b/server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
@@ -96,8 +96,7 @@ public class JettyBackedGrpcServer implements GrpcServer {
         }
         context.setInitParameter(DefaultServlet.CONTEXT_INIT + "dirAllowed", "false");
 
-        // For the Web UI, cache everything in the static folder
-        // https://create-react-app.dev/docs/production-build/#static-file-caching
+        // Cache all of the appropriate assets folders
         for (String appRoot : List.of("/ide/", "/iframe/table/", "/iframe/chart/")) {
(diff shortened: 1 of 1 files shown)


server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
```diff
@@ -97,9 +98,11 @@ public JettyBackedGrpcServer(
 
         // For the Web UI, cache everything in the static folder
         // https://create-react-app.dev/docs/production-build/#static-file-caching
```
niloc132, line 100: maybe update these comments while we're in here?

niloc132 (approved):
one comment, otherwise approved assuming these paths are all as expected

server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
devinrsmith: I assumed the "static folder" meant something about the react build process; but in this context, should we instead say the "assets folder"? @mofojed

mofojed: Yea, we use Vite now instead of create-react-app, so we should update this comment. Though there doesn't seem to be a Vite equivalent page that is as helpful, closest I could find is: https://vitejs.dev/guide/dep-pre-bundling.html#browser-cache

devinrsmith:
This PR and #3831, improves the lighthouse score from 55 -> 79.

mofojed (approved):

niloc132 (approved):

Issue #3004, opened 2022-10-19
Jetty HTTP testing


devinrsmith:
With our move to jetty, it probably makes sense to do HTTP client tests against the servlets and filters we install in `io.deephaven.server.jetty.JettyBackedGrpcServer`.

For example, testing cache parameters, etags, ide, jsapi, and js-plugins.

devinrsmith:
* [Playwright](https://github.com/microsoft/playwright)
* [Selenium](https://github.com/SeleniumHQ/selenium)
* [Lighthouse](https://github.com/GoogleChrome/lighthouse)

Are potential tools.

PR #3831, opened 2023-05-11
Add HTTP compression support

devinrsmith:
By default, enables HTTP compression. It can be explicitly disabled by setting the configuration property `http.compression` to `false`.

Partial #3827

### DIFF
--- a/server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
+++ b/server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
@@ -216,6 +216,8 @@ public class JettyBackedGrpcServer implements GrpcServer {
             // The default of 32 bytes seems a bit small.
             gzipHandler.setMinGzipSize(1024);
             // The GzipHandler documentation says GET is the default, but the constructor shows both GET and POST.
+            // This should ensure our gRPC messages don't get compressed for now, but we may need to be more explicit in
+            // the future as gRPC can technically operate over GET.
             gzipHandler.setIncludedMethods(HttpMethod.GET.asString());
             // Otherwise, the other defaults seem reasonable.
             gzipHandler.setHandler(handlers);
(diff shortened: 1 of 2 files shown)


server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
```diff
@@ -205,7 +209,21 @@ public <T> T getEndpointInstance(Class<T> endpointClass) {
...
         handlers.addHandler(context);
-        jetty.setHandler(handlers);
+
+        final Handler handler;
+        if (config.httpCompressionOrDefault()) {
+            final GzipHandler gzipHandler = new GzipHandler();
+            // The default of 32 bytes seems a bit small.
+            gzipHandler.setMinGzipSize(1024);
+            // The GzipHandler documentation says GET is the default, but the constructor shows both GET and POST.
+            gzipHandler.setIncludedMethods(HttpMethod.GET.asString());
```
niloc132, line 219: I'm guessing you are setting this to avoid gzipping the grpc streams (which default to POST)? That might eventually not be sufficient, grpc can technically use GET as well (though as we have it written today, it shouldn't come up).

devinrsmith:
Would be good to have some follow-up wrt https://github.com/deephaven/deephaven-core/issues/3004

server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
devinrsmith: Correct. I've updated the comment.

devinrsmith:
This PR and #3829, improves the lighthouse score from 55 - 79.

mofojed (approved):
Awesome, this reduces the payload significantly.

Kind of annoying I couldn't get etags sorted out: https://github.com/deephaven/deephaven-core/issues/3011

I think that would be the only other spot we could optimize.

niloc132 (approved):

PR #3833, opened 2023-05-15
Add additional options for percentile, median, unique, count_distinct and wrap distinct in the Python server API

### LINKED ISSUE #3626: Add additional options for percentile, median, unique, count_distinct and wrap distinct in the Python server API

jmao-denver:
Fixes #3626

### DIFF
--- a/py/server/deephaven/agg.py
+++ b/py/server/deephaven/agg.py
@@ -46,7 +46,7 @@ class Aggregation:
 
 
 def sum_(cols: Union[str, List[str]] = None) -> Aggregation:
-    """Create a Sum aggregation.
+    """Creates a Sum aggregation.
 
     Args:
         cols (Union[str, List[str]]): the column(s) to aggregate on, can be renaming expressions, i.e. "new_col = col";
@@ -59,7 +59,7 @@ def sum_(cols: Union[str, List[str]] = None) -> Aggregation:
 
 
 def abs_sum(cols: Union[str, List[str]] = None) -> Aggregation:
-    """Create an Absolute-sum aggregation.
+    """Creates an Absolute-sum aggregation.
 
     Args:
         cols (Union[str, List[str]]): the column(s) to aggregate on, can be renaming expressions, i.e. "new_col = col";
@@ -72,7 +72,7 @@ def abs_sum(cols: Union[str, List[str]] = None) -> Aggregation:
 
 
 def group(cols: Union[str, List[str]] = None) -> Aggregation:
-    """Create a Group aggregation.
+    """Creates a Group aggregation.
 
     Args:
         cols (Union[str, List[str]]): the column(s) to aggregate on, can be renaming expressions, i.e. "new_col = col";
@@ -85,7 +85,7 @@ def group(cols: Union[str, List[str]] = None) -> Aggregation:
 
 
 def avg(cols: Union[str, List[str]] = None) -> Aggregation:
-    """Create an Average aggregation.
+    """Creates an Average aggregation.
 
     Args:
         cols (Union[str, List[str]]): the column(s) to aggregate on, can be renaming expressions, i.e. "new_col = col";
@@ -98,7 +98,7 @@ def avg(cols: Union[str, List[str]] = None) -> Aggregation:
 
 
 def count_(col: str) -> Aggregation:
-    """Create a Count aggregation. This is not supported in 'Table.agg_all_by'.
+    """Creates a Count aggregation. This is not supported in 'Table.agg_all_by'.
 
     Args:
         col (str): the column to hold the counts of each distinct group
@@ -110,7 +110,7 @@ def count_(col: str) -> Aggregation:
 
 
 def partition(col: str, include_by_columns: bool = True) -> Aggregation:
-    """Create a Partition aggregation. This is not supported in 'Table.agg_all_by'.
+    """Creates a Partition aggregation. This is not supported in 'Table.agg_all_by'.
 
     Args:
         col (str): the column to hold the sub tables
@@ -123,7 +123,7 @@ def partition(col: str, include_by_columns: bool = True) -> Aggregation:
 
 
 def count_distinct(cols: Union[str, List[str]] = None, count_nulls: bool = False) -> Aggregation:
-    """Create a Count Distinct aggregation which computes the count of distinct values within an aggregation group for
+    """Creates a Count Distinct aggregation which computes the count of distinct values within an aggregation group for
     each of the given columns.
 
     Args:
@@ -138,7 +138,7 @@ def count_distinct(cols: Union[str, List[str]] = None, count_nulls: bool = False
 
 
 def first(cols: Union[str, List[str]] = None) -> Aggregation:
-    """Create a First aggregation.
+    """Creates a First aggregation.
 
     Args:
         cols (Union[str, List[str]]): the column(s) to aggregate on, can be renaming expressions, i.e. "new_col = col";
@@ -151,7 +151,7 @@ def first(cols: Union[str, List[str]] = None) -> Aggregation:
 
 
 def formula(formula: str, formula_param: str, cols: Union[str, List[str]] = None) -> Aggregation:
-    """Create a user defined formula aggregation.
+    """Creates a user defined formula aggregation.
 
     Args:
         formula (str): the user defined formula to apply to each group
@@ -166,7 +166,7 @@ def formula(formula: str, formula_param: str, cols: Union[str, List[str]] = None
 
 
 def last(cols: Union[str, List[str]] = None) -> Aggregation:
-    """Create Last aggregation.
+    """Creates Last aggregation.
 
     Args:
         cols (Union[str, List[str]]): the column(s) to aggregate on, can be renaming expressions, i.e. "new_col = col";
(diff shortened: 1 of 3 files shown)


py/server/deephaven/agg.py
```diff
@@ -201,31 +202,39 @@ def max_(cols: Union[str, List[str]] = None) -> Aggregation:
...
-    """Create a Percentile aggregation.
+def pct(percentile: float, cols: Union[str, List[str]] = None, average_evenly_divided: bool = False) -> Aggregation:
+    """Create a Percentile aggregation which computes the percentile value within an aggregation group for each of
+    the given columns.
 
     Args:
         percentile (float): the percentile used for calculation
         cols (Union[str, List[str]]): the column(s) to aggregate on, can be renaming expressions, i.e. "new_col = col";
             default is None, only valid when used in Table agg_all_by operation
+        average_evenly_divided (bool): when the percentile is 0.5 and the group size is an even number, whether to
```
rcaudy, line 230: This is not specific to percentile == 0.5

py/server/deephaven/agg.py
```diff
@@ -269,17 +278,29 @@ def std(cols: Union[str, List[str]] = None) -> Aggregation:
...
+    """Create a Unique aggregation which computes the single unique value within an aggregation group for each of
+    the given columns. If all values in a column are null, or if there is more than one distinct value in a column, the
+    result is null.
 
     Args:
         cols (Union[str, List[str]]): the column(s) to aggregate on, can be renaming expressions, i.e. "new_col = col";
             default is None, only valid when used in Table agg_all_by operation
+        include_nulls (bool): whether null is treated as a value for the purpose of determining if the values in the
+            aggregation group are unique, default is False.
+        null_sentinel (Any): the non-null sentinel value to substitute null when include_nulls is True, default is None
```
rcaudy, line 291: I'm not sure None makes sense here. Will that result in us storing `null`? Are you unwrapping `PyObject`s?

py/server/deephaven/agg.py
```diff
@@ -269,17 +278,29 @@ def std(cols: Union[str, List[str]] = None) -> Aggregation:
     return Aggregation(j_agg_spec=_JAggSpec.std(), cols=cols)
 
 
-def unique(cols: Union[str, List[str]] = None) -> Aggregation:
-    """Create a Unique aggregation.
+def unique(cols: Union[str, List[str]] = None, include_nulls: bool = False, null_sentinel: Any = None) -> Aggregation:
```
rcaudy, line 281: This work should cover similar args to `count_distinct`. Also, is the wrapper missing for distinct?

py/server/deephaven/agg.py
jmao-denver: Hmm, then the Java doc should be revised? 

```
   /**
     * The percentile to calculate. Must be &gt;= 0.0 and &lt;= 1.0.
     *
     * @return The percentile to calculate
     */
    @Parameter
    public abstract double percentile();

    /**
     * Whether to average the highest low-bucket value and lowest high-bucket value, when the low-bucket and high-bucket
     * are of equal size. Only applies to numeric types.
     *
     * @return Whether to average the two result candidates for evenly-divided input sets of numeric types
     */
    @Parameter
    public abstract boolean averageEvenlyDivided();

```

py/server/deephaven/agg.py
jmao-denver: 1. null_sentinel is only applied when include_nulls is True
2. when include_nulls is True, we check to make sure null_sentinel be non-None

py/server/deephaven/agg.py
rcaudy: Where does the JavaDoc say the language I objected to?

jmao-denver: `when the low-bucket and high-bucket
     * are of equal size.`

py/server/deephaven/agg.py
rcaudy: I think there's some confusion here. The "non-unique sentinel" is what you should be wrapping. Not the "null sentinel". Look at `AggSpecUnique`.

py/server/deephaven/agg.py
rcaudy: This is about the implementation details. It's a more reasonable consideration for 50th percentile, but in cases with relatively few inputs, we could still be evenly divided at other percentiles.
The formula for how big the low bucket will be is:
```
            final long targetLo = (int)((totalSize - 1) * percentile) + 1;
```
So if you have 2 values and 70th percentile, we'll be evenly divided. Same for 4 values and 60th percentile.

py/server/deephaven/agg.py
```diff
@@ -321,3 +344,18 @@ def weighted_sum(wcol: str, cols: Union[str, List[str]] = None) -> Aggregation:
         an aggregation
     """
     return Aggregation(j_agg_spec=_JAggSpec.wsum(wcol), cols=cols)
+
+
+def distinct(cols: Union[str, List[str]] = None, include_nulls: bool = False) -> Aggregation:
+    """Create a Distinct aggregation which computes the distinct values within an aggregation group for each of the
```
rcaudy, line 350: You should mention that the output column values are vectors, as in `group`.

py/server/deephaven/agg.py
```diff
@@ -278,33 +280,35 @@ def std(cols: Union[str, List[str]] = None) -> Aggregation:
...
+    result is null or the specified non_unique_sentinel value.
 
     Args:
         cols (Union[str, List[str]]): the column(s) to aggregate on, can be renaming expressions, i.e. "new_col = col";
             default is None, only valid when used in Table agg_all_by operation
         include_nulls (bool): whether null is treated as a value for the purpose of determining if the values in the
             aggregation group are unique, default is False.
-        null_sentinel (Any): the non-null sentinel value to substitute null when include_nulls is True, default is None
+        non_unique_sentinel (Any): the non-null sentinel value when no unique value exists, default is None, must be
+            a non-None value when include_nulls is True.
```
rcaudy, line 294: Something is off in the wording here. The error talks about `null`, this talks about `None`. Which is better? Also, I think "the non-null sentinel" should just say "the sentinel" and "when no unique value exists" might be better as "when no single unique value exists"
```suggestion
        non_unique_sentinel (Any): the sentinel value to use when no single unique value exists, default is None, must be
            a non-None value when include_nulls is True.
```

py/server/deephaven/agg.py
```diff
@@ -278,33 +280,35 @@ def std(cols: Union[str, List[str]] = None) -> Aggregation:
...
     Returns:
         an aggregation
     """
-    if include_nulls:
-        if null_sentinel:
-            null_sentinel = _JUnionObject.of(null_sentinel)
-        else:
+
+    if non_unique_sentinel is None:
+        if include_nulls:
```
rcaudy, line 301: I don't think we actually impose this limitation in the Java code. I suppose in your case it's reasonable, since the argument is optional.

rcaudy (dismissed):

devinrsmith (dismissed):

py/server/deephaven/agg.py
```diff
@@ -201,35 +204,43 @@ def max_(cols: Union[str, List[str]] = None) -> Aggregation:
...
+def median(cols: Union[str, List[str]] = None, average_evenly_divided: bool = True) -> Aggregation:
+    """Creates a Median aggregation which computes the median value within an aggregation group for each of the
+    given columns.
 
     Args:
         cols (Union[str, List[str]]): the column(s) to aggregate on, can be renaming expressions, i.e. "new_col = col";
             default is None, only valid when used in Table agg_all_by operation
+        average_evenly_divided (bool): when the group size is an even number, whether to average the two middle values
+            for the output value, default is True. when set to False, use the smaller value. This flag is only valid for
+            numeric types.
```
chipkent, line 216:
```suggestion
        average_evenly_divided (bool): when the group size is an even number, whether to average the two middle values
            for the output value.  When set to True, average the two middle values. When set to False, use the smaller value. The default is True.  This flag is only valid for
            numeric types.
```

py/server/deephaven/agg.py
```diff
@@ -201,35 +204,43 @@ def max_(cols: Union[str, List[str]] = None) -> Aggregation:
...
+    """Creates a Percentile aggregation which computes the percentile value within an aggregation group for each of
+    the given columns.
 
     Args:
         percentile (float): the percentile used for calculation
         cols (Union[str, List[str]]): the column(s) to aggregate on, can be renaming expressions, i.e. "new_col = col";
             default is None, only valid when used in Table agg_all_by operation
+        average_evenly_divided (bool): when the percentile splits the group into two halves, whether to average the two
+            middle values for the output value, default is False, meaning to use the smaller value. This flag is only
+            valid for numeric types.
```
chipkent, line 234: See other suggestion.  These are worded in a way that is complex syntactically and users may have a hard time following.

py/server/deephaven/agg.py
```diff
@@ -269,21 +280,35 @@ def std(cols: Union[str, List[str]] = None) -> Aggregation:
...
+    the given columns. If all values in a column are null, or if there is more than one distinct value in a column, the
+    result is null or the specified non_unique_sentinel value.
 
     Args:
         cols (Union[str, List[str]]): the column(s) to aggregate on, can be renaming expressions, i.e. "new_col = col";
             default is None, only valid when used in Table agg_all_by operation
+        include_nulls (bool): whether null is treated as a value for the purpose of determining if the values in the
+            aggregation group are unique, default is False.
+        non_unique_sentinel (Any): the non-null sentinel value when no unique value exists, default is None, must be
+            a non-None value when include_nulls is True.
```
chipkent, line 294:
```suggestion
        non_unique_sentinel (Any): the non-null sentinel value when no unique value exists, default is None,]. Must be
            a non-None value when include_nulls is True.
```

chipkent (changes requested):

chipkent (approved):

PR #3839, opened 2023-05-16
Check for overflows when calling jpy getIntValue

### LINKED ISSUE #3716: Properly handle returns from PyObject.getIntValue() where it is used due to its lossy nature with overflow
One solution: Math.toIntExact(valueIn.getLongValue())
```
io.deephaven.integrations.python.PythonValueGetter
io.deephaven.engine.table.impl.lang.QueryLanguageFunctionUtils
```

Same solution needs to be applied to other non-long integer types.

jmao-denver:
Fixes #3716

related to https://github.com/jpy-consortium/jpy/issues/98
linked to: #3598 

Note:
PyObject.getIntValue and family are also used in the generated formula code to cast a PyObject to a specified 'integer' type. Since casting in Java doesn't check for overflow, these use cases are not included as part of the change.

### DIFF
--- a/Integrations/src/main/java/io/deephaven/integrations/python/PythonValueGetter.java
+++ b/Integrations/src/main/java/io/deephaven/integrations/python/PythonValueGetter.java
@@ -49,21 +49,33 @@ class PythonValueGetter {
         if (valueIn == null) {
             return QueryConstants.NULL_INT;
         }
-        return valueIn.getIntValue();
+        long ll = valueIn.getLongValue();
+        if (ll > Integer.MAX_VALUE || ll < Integer.MIN_VALUE) {
+            throw new ArithmeticException("The PyObject value is outside Integer range.");
+        }
+        return (int) ll;
     }
 
     static short getShort(PyObject valueIn) {
         if (valueIn == null) {
-            return QueryConstants.NULL_SHORT;
+            return QueryConstants.NULL_SHORT; // NB: should there be a getShortValue() in jpy?
         }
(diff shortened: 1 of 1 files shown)


Integrations/src/main/java/io/deephaven/integrations/python/PythonValueGetter.java
```diff
@@ -49,21 +49,33 @@ static int getInt(PyObject valueIn) {
         if (valueIn == null) {
             return QueryConstants.NULL_INT;
         }
-        return valueIn.getIntValue();
+        long ll = valueIn.getLongValue();
+        if (ll > Integer.MAX_VALUE || ll < Integer.MIN_VALUE) {
```
devinrsmith, line 53: The more canonical way to do this is:

```java
if ((int) ll != ll) {
...
```

Integrations/src/main/java/io/deephaven/integrations/python/PythonValueGetter.java
```diff
@@ -49,21 +49,33 @@ static int getInt(PyObject valueIn) {
...
+        return (int) ll;
     }
 
     static short getShort(PyObject valueIn) {
         if (valueIn == null) {
-            return QueryConstants.NULL_SHORT;
+            return QueryConstants.NULL_SHORT; // NB: should there be a getShortValue() in jpy?
+        }
+        long ll = valueIn.getLongValue();
+        if (ll > Short.MAX_VALUE || ll < Short.MIN_VALUE) {
```
devinrsmith, line 64: `(short) ll != ll`

devinrsmith (approved):

PR #3845, opened 2023-05-17
Add methods to get underlying data array from writable chunks.

cpwright:


engine/chunk/src/main/java/io/deephaven/chunk/WritableCharChunk.java
```diff
@@ -79,6 +79,29 @@ public WritableCharChunk<ATTR> slice(int offset, int capacity) {
...
+    /**
+     * Get the offset into the underlying data array.
+     *
+     * @return the offset into the underlying data array.
+     */
+    public final int offset() {
+        return offset;
+    }
+    // endregion array
+
```
rcaudy, line 104:
```suggestion
    // region array
    /**
     * Get the data array backing this WritableCharChunk. The first element of this chunk corresponds to
     * {@code array()[arrayOffset()]}.
     * <p>
     * This WritableCharChunk must never be {@link #close() closed} while the array <em>may</em> be in use externally,
     * because it must not be returned to any pool for re-use until that re-use is guaranteed to be exclusive.
     *
     * @return The backing data array
     */
    public final char[] array() {
        return data;
    }

    /**
     * Get this WritableCharChunk's offset into the backing data array. The first element of this chunk corresponds to
     * {@code array()[arrayOffset()]}.
     *
     * @return The offset into the backing data array
     */
    public final int arrayOffset() {
        return offset;
    }
    // endregion array

```

engine/chunk/src/test/java/io/deephaven/chunk/TestCharChunk.java
```diff
@@ -595,6 +595,35 @@ public void testWritableChunkChunk() {
...
+                TestCase.assertEquals((char) ii, array[ii]);
+            }
+
+            array[offset] = 42;
+            TestCase.assertEquals((char)42, chunk.get(0));
+            chunk.set(1, (char) 97);
+            TestCase.assertEquals((char)97, array[offset + 1]);
+        }
+    }
+    // endregion testArray
```
rcaudy, line 625:
```suggestion
    // region testArray
    @Test
    public void testArray() {
        final char[] underlyingData = new char[100];
        for (int ii = 0; ii < underlyingData.length; ++ii) {
            underlyingData[ii] = (char) ii;
        }
        final char[] array;
        final int offset;
        // noinspection rawtypes
        try (WritableCharChunk chunk = WritableCharChunk.writableChunkWrap(underlyingData, 10, 20)) {
            array = chunk.array();
            offset = chunk.arrayOffset();
            TestCase.assertSame(underlyingData, array);
            TestCase.assertEquals(10, offset);

            final int lastOffset = offset + chunk.size();
            for (int ii = offset; ii < lastOffset; ++ii) {
                TestCase.assertEquals((char) ii, array[ii]);
            }

            array[offset] = 42;
            TestCase.assertEquals((char) 42, chunk.get(0));
            chunk.set(1, (char) 97);
            TestCase.assertEquals((char) 97, array[offset + 1]);
        }
    }
    // endregion testArray
```

rcaudy (commented):
Suggested cleanups to JavaDocs and code (since spotless won't touch these files), as well as a rename from `offset()` to `arrayOffset()`. Also changed unit test to use `arrayOffset()` instead of bare access to `offset` afforded by package-private access.

rcaudy (approved):

PR #3850, opened 2023-05-18
Remove shift-restrictions for non-stream tables in PublishToKafa.PublishListener

### LINKED ISSUE #3847: PublishToKafka.PublishListener is too picky about shifts
We currently have:
```
            if (keySerializer != null || isStream) {
                Assert.assertion(upstream.shifted().empty(), "upstream.shifted.empty()");
            }
```
For the stream/blink case, that's fine, the assumption should never be violated, although maybe we should just not care.
For the keyed case, it's really not so good; if users are merging partitions, or sorting, we might see shifts that don't impact our correctness in any way.

rcaudy:
Fixes #3847

### DIFF
--- a/extensions/kafka/src/main/java/io/deephaven/kafka/publish/PublishToKafka.java
+++ b/extensions/kafka/src/main/java/io/deephaven/kafka/publish/PublishToKafka.java
@@ -249,16 +249,13 @@ public class PublishToKafka<K, V> extends LivenessArtifact {
 
         @Override
         public void onUpdate(TableUpdate upstream) {
-            if (keySerializer != null || isStream) {
-                Assert.assertion(upstream.shifted().empty(), "upstream.shifted.empty()");
-            }
             Assert.assertion(!keysModified.containsAny(upstream.modifiedColumnSet()),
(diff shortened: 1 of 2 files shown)


extensions/kafka/src/main/java/io/deephaven/kafka/publish/PublishToKafka.java
```diff
@@ -249,16 +249,13 @@ private PublishListener(
 
         @Override
         public void onUpdate(TableUpdate upstream) {
-            if (keySerializer != null || isStream) {
```
devinrsmith, line 252: I'm assuming `keySerializer != null` was just a bad check here too?

devinrsmith (approved):

jmao-denver (approved):

extensions/kafka/src/main/java/io/deephaven/kafka/publish/PublishToKafka.java
rcaudy: No. The theory was that if you had a key serializer, we expected you to be downstream of an aggregation, in which case there would be no shifts.

cpwright (approved):

PR #3851, opened 2023-05-19
Port DH-11168: Indexing enhancements to Community.

lbooker42:
Port [DH-11168](https://deephaven.atlassian.net/browse/DH-11168): Indexing enhancements to Community.


lbooker42 (commented):
Performed self-review

engine/table/src/main/java/io/deephaven/engine/table/impl/AbstractColumnSource.java
```diff
@@ -108,69 +108,73 @@ public ColumnSource<T> getPrevSource() {
...
+            boolean invertMatch, Object... keys) {
+        final GroupingBuilder groupingBuilder = getGroupingBuilder();
+        final Table filteredGroups = getGroupingBuilder()
+                .clampToIndex(startingWritableRowSet)
+                .matching(!caseInsensitive, invertMatch, keys)
+                .buildTable();
+        final RowSetBuilderRandom allInMatchingGroups = RowSetFactory.builderRandom();
+        final ColumnSource<RowSet> indexSource =
+                filteredGroups.getColumnSource(groupingBuilder.getIndexColumnName(), RowSet.class);
+        filteredGroups.getRowSet().forAllRowKeys(key -> allInMatchingGroups.addRowSet(indexSource.get(key)));
```
rcaudy, line 169: This is a good place to use a column iterator.

engine/table/src/main/java/io/deephaven/engine/table/impl/indexer/RowSetIndexer.java
```diff
@@ -243,24 +251,25 @@ private static Map<Object, RowSet> lookupMapping(
     private static void collectGrouping(
             final RowSet rowSetRowSet,
```
rcaudy, line 252: What did I do?

engine/table/src/main/java/io/deephaven/engine/table/impl/indexer/RowSetIndexer.java
```diff
@@ -243,24 +251,25 @@ private static Map<Object, RowSet> lookupMapping(
...
             @SuppressWarnings("unchecked")
-            final Map<Object, RowSet> sourceGrouping = keyColumns.get(0).getGroupToRange();
-            for (final Map.Entry<Object, RowSet> objectRowSetEntry : sourceGrouping.entrySet()) {
-                final RowSet resultRowSet = rowSetOp.apply(objectRowSetEntry.getValue());
-                if (resultRowSet.size() > 0) {
-                    resultCollector.accept(objectRowSetEntry.getKey(), resultRowSet);
+            final GroupingBuilder builder = keyColumns.get(0).getGroupingBuilder();
+            final Map<Object, RowSet> groupings = builder.buildGroupingMap();
+            groupings.forEach((key, value) -> {
+                final RowSet resultIndex = rowSetOp.apply(value);
```
rcaudy, line 266: I'm suspicious that rowSetOp is always a no-op, or intersecting (which could be done with strict building), and that resultCollector is being used to make a map or a table, potentially unnecessarily.

engine/table/src/main/java/io/deephaven/engine/table/impl/indexer/RowSetIndexer.java
```diff
@@ -243,24 +251,25 @@ private static Map<Object, RowSet> lookupMapping(
...
+                if (value.size() > 0 && resultIndex.size() > 0) {
+                    resultCollector.accept(key, resultIndex);
                 }
-            }
+            });
         } else {
-            final long columnsWithGrouping = keyColumns.stream().filter(cs -> cs.getGroupToRange() != null).count();
+            final long columnsWithGrouping = keyColumns.stream().filter(ColumnSource::hasGrouping).count();
```
rcaudy, line 272: Is this case what DataIndexers are for?

engine/table/src/main/java/io/deephaven/engine/table/impl/dataindex/DataIndexProvider.java
```diff
@@ -0,0 +1,25 @@
+package io.deephaven.engine.table.impl.dataindex;
+
+import io.deephaven.engine.table.Table;
+import io.deephaven.util.annotations.InternalUseOnly;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * An interface that exposes methods to find and retrieve Data Indexes created during the merge process. A Data Index is
```
rcaudy, line 9: This is pretty enterprise-specific.

engine/table/src/main/java/io/deephaven/engine/table/impl/dataindex/DataIndexProvider.java
```diff
@@ -0,0 +1,25 @@
...
+@InternalUseOnly
+public interface DataIndexProvider {
+    /**
+     * Get the data index for the specified set of key columns.
+     *
+     * @param columns the key columns
+     * @return the data index table, or null if one was not available.
+     */
+    @Nullable
+    Table getDataIndex(@NotNull String... columns);
```
rcaudy, line 24: This seems under-specified. Is there a column of row sets? What's it named?

engine/table/src/main/java/io/deephaven/engine/table/impl/dataindex/DataIndexProviderImpl.java
```diff
@@ -0,0 +1,132 @@
...
+                    // Shift the indices onto our range.
+                    indexTable = indexTable.update("Index=Index.shift(" + firstKey + ")");
+
+                    if (dataIndices == null) {
+                        dataIndices = new HashMap<>();
+                    }
+                    dataIndices.put(indexKey, new SoftReference<>(indexTable));
+                }
+                return indexTable;
+            }
```
rcaudy, line 77: I think we should make this lock smaller. This might be a use-case for having a persistent `DataIndexState` object linking an immutable `indexKey` and a mutable `volatile SoftReference`, maybe with a cached "canCompute" `boolean` to avoid asking for the same index more than once if the location can't provide it.

Having a re-usable object will allow us to lock at finer granularity if we need to concurrently build two different indices. Maybe use `KeyedObjectHashMap`.

engine/table/src/main/java/io/deephaven/engine/table/impl/dataindex/DataIndexProviderImpl.java
```diff
@@ -0,0 +1,132 @@
...
+                    return null;
+                }
+
+                locationIndexes[tCount++] = locationIndex;
+            }
+
+            // Ensure we return the final table sorted by firstKey so we preserve encounter order of the original data.
+            return TableTools.merge(locationIndexes)
+                    .groupBy(columns)
+                    .update("Index=io.deephaven.engine.table.impl.dataindex.DataIndexProviderImpl.combineIndices(Index)")
```
rcaudy, line 116: I wonder if we could avoid the compilation.

engine/table/src/main/java/io/deephaven/engine/table/impl/dataindex/DataIndexProviderImpl.java
```diff
@@ -0,0 +1,132 @@
...
+                    .updateView("FirstKey=Index.firstRowJey()")
```
rcaudy, line 117: Same comment regarding compilation, also, what's a row jey?

rcaudy (commented):
Some superficial/high level review.

alexpeters1208 (commented):
Looks like the R client is no longer touched by this PR, as it was in previous iterations. I assume my review request is leftover from that.

engine/table/src/main/java/io/deephaven/engine/table/impl/indexer/DataIndexer.java
```diff
@@ -0,0 +1,269 @@
...
+    /** Remove the item at the specified index from a list. */
+    private static <T> List<T> removeItem(List<T> items, int index) {
+        List<T> result = new ArrayList<T>(items.size() - 1);
+        result.addAll(items.subList(0, index));
+        result.addAll(items.subList(index + 1, items.size() + 1));
+        return result;
+    }
+
+    /** Check a specified map for an index with these column sources. */
+    private static DataIndex findIndex(
```
rcaudy, line 213: Placeholder for Ryan

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/TableLocation.java
```diff
@@ -77,13 +83,52 @@ interface Listener extends BasicTableDataListener {
...
+    boolean hasDataIndexFor(@NotNull String... columns);
+
+    /**
+     * Get the data index table for the specified set of columns. Note that the order of columns does not matter here.
+     *
+     * @param columns the key columns for the index
+     * @return the index table or null if one does not exist.
+     */
+    @Nullable
+    Table getDataIndex(@NotNull String... columns);
```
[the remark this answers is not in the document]
lbooker42: I don't think so, this is the raw table associated with the index at this location.  DataIndex is a table-level concept.

engine/api/src/main/java/io/deephaven/engine/table/DataIndex.java
```diff
@@ -0,0 +1,164 @@
...
+public interface DataIndex {
+    interface PositionLookup {
+        int apply(Object key);
+    }
+
+    interface RowSetLookup {
+        RowSet apply(Object key);
+    }
+
+    /** Get the key column sources for this index. */
```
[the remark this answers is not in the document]
lbooker42: Discuss whether to maintain any string interfaces or only use column source -> column source mapping...

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/RegionedColumnSourceManager.java
```diff
@@ -89,9 +106,33 @@ public class RegionedColumnSourceManager implements ColumnSourceManager {
...
+        locationSource = (ObjectArraySource<TableLocation>) ArrayBackedColumnSource.getMemoryColumnSource(0,
+                TableLocation.class);
+        offsetSource = (LongArraySource) ArrayBackedColumnSource.getMemoryColumnSource(0, long.class);
+        rowSetSource = (ObjectArraySource<RowSet>) ArrayBackedColumnSource.getMemoryColumnSource(0, RowSet.class);
+        final Map<String, ColumnSource<?>> columnLocationMap = new LinkedHashMap<>();
+        columnLocationMap.put(LOCATION_COLUMN_NAME, locationSource);
+        columnLocationMap.put(OFFSET_COLUMN_NAME, offsetSource);
+        columnLocationMap.put(ROWSET_COLUMN_NAME, rowSetSource);
+
+        orderedLocationsTable = new QueryTable(RowSetFactory.empty().toTracking(), columnLocationMap);
```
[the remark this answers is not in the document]
rcaudy: See `io.deephaven.engine.table.impl.util.FunctionGeneratedTableFactory.DelayedErrorNotifier`

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/RegionedColumnSourceManager.java
```diff
@@ -302,35 +382,26 @@ private void processInitial(final RowSetBuilderSequential addedRowSetBuilder, fi
...
-                        state.updateGrouping(
-                                addRowSetInTable == null ? addRowSetInTable = initialRowSet.shift(regionFirstKey)
-                                        : addRowSetInTable);
-                    }
-                }
-            } finally {
-                if (addRowSetInTable != null) {
-                    addRowSetInTable.close();
-                }
+
```
rcaudy, line 385: Placeholder

engine/table/src/main/java/io/deephaven/engine/table/impl/UncoalescedTable.java
```diff
@@ -65,6 +65,8 @@ public UncoalescedTable(@NotNull final TableDefinition definition, @NotNull fina
      */
     protected abstract Table doCoalesce();
 
+    protected void postCoalesceAction(final Table coalesced) {}
```
rcaudy, line 68: Why isn't this part of `doCoalesce()`? You explained. I think you have access to all the stuff you care about via other means.

rcaudy (commented):
Today's review of RCSM and SourceTable stuff.

engine/api/src/main/java/io/deephaven/engine/table/DataIndexTransformer.java
```diff
@@ -0,0 +1,60 @@
...
+    Optional<RowSet> intersectRowSet();
+
+    Optional<RowSet> invertRowSet();
+
+    @Value.Default
+    default boolean sortByFirstRowKey() {
+        return false;
+    }
+
+    Map<ColumnSource<?>, ColumnSource<?>> keyColumnRemap();
```
rcaudy, line 24: Might be worth renaming and documenting to make it very clear which side is "new" and which is "old".

rcaudy (commented):
Comments on top-level interfaces and ticking patterns for index tables.

engine/table/src/main/java/io/deephaven/engine/table/impl/indexer/DataIndexer.java
```diff
@@ -0,0 +1,256 @@
...
+ * 
+ * @apiNote DataIndexers should not be used after the host {@link TrackingRowSet} has been {@link RowSet#close()
+ *          closed}.
+ */
+public class DataIndexer implements TrackingRowSet.Indexer {
+    public static DataIndexer of(TrackingRowSet rowSet) {
+        return rowSet.indexer(DataIndexer::new);
+    }
+
+    private final TrackingRowSet rowSet;
```
rcaudy, line 31: Unused. Unneeded?

engine/table/src/main/java/io/deephaven/engine/table/impl/indexer/DataIndexer.java
```diff
@@ -0,0 +1,256 @@
...
+        return index != null;
+    }
+
+    public boolean hasDataIndex(final ColumnSource<?>... keyColumns) {
+        return hasDataIndex(Arrays.asList(keyColumns));
+    }
+
+    public boolean hasDataIndex(final Collection<ColumnSource<?>> keyColumns) {
+        if (keyColumns.isEmpty()) {
+            return true;
```
rcaudy, line 70: This should be `false`, or you should be building a single-celled table data index when asked (via no-key agg).

engine/table/src/main/java/io/deephaven/engine/table/impl/indexer/DataIndexer.java
```diff
@@ -0,0 +1,256 @@
...
+        }
+    }
+
+    private DataIndexer(@NotNull final TrackingRowSet rowSet) {
+        this.rowSet = rowSet;
+        this.dataIndexes = new WeakHashMap<>();
+    }
+
+    public boolean hasDataIndex(final Table table, final String... keyColumnNames) {
+        final DataIndex index = getDataIndex(table, keyColumnNames);
```
rcaudy, line 60: It surprises me that this doesn't remap to column sources and call `hasDataIndex`.

engine/table/src/main/java/io/deephaven/engine/table/impl/indexer/DataIndexer.java
```diff
@@ -0,0 +1,256 @@
...
+
+        DataIndexCache(@Nullable final DataIndex index) {
+            this.index = index;
+            this.dataIndexes = new WeakHashMap<>();
+        }
+    }
+
+    private DataIndexer(@NotNull final TrackingRowSet rowSet) {
+        this.rowSet = rowSet;
+        this.dataIndexes = new WeakHashMap<>();
```
rcaudy, line 56: Either make this map `final` (and delete the code for allocating it at on-demand or expecting that it could ever be `null`), or remove this initialization.

I think I prefer the deferred initialization.
Follow the pattern in `io.deephaven.engine.table.impl.BaseTable#ensureParents`.

Only materialize it when you plan to mutate it, because `null` implies `empty`.

*Always* synchronize when you interrogate it.
*Never* synchronize while building a data index; but do synchronize while adding it, and be sure that if you lose a race, you return the result that is actually in the cache.
*Lastly*; we need to understand liveness for these things. If they are refreshing, and no longer live, they don't exist. We should be handing out liveness references to our DataIndexes on access. See the memoization cache code in `QueryTable`.

Maybe we want to actually keep our data indexes weakly reachable?

engine/table/src/main/java/io/deephaven/engine/table/impl/indexer/DataIndexer.java
```diff
@@ -0,0 +1,256 @@
...
+
+        if (dataIndexes == null) {
+            return false;
+        }
+
+        final DataIndex dataIndex = findIndex(dataIndexes, keyColumns);
+        // It's possible that a potentially indexed column might be corrupt or incomplete when we try to use it.
+        // Test it now to make sure that we don't falsely claim to have a functional index.
+        if (dataIndex != null && dataIndex.table() != null) {
+            return true;
```
rcaudy, line 81: Not sure if this is the right test.
We want to ensure that it's still live if ticking.
We want to give an ownership relationship to the caller in the `get` case (but not the `has` case).

In the case where `table()` is corrupt, we should probably throw an exception before returning here or in the getIndex case.

engine/table/src/main/java/io/deephaven/engine/table/impl/indexer/DataIndexer.java
```diff
@@ -0,0 +1,256 @@
...
+        // TODO: should we remove the index if it's corrupt? Could it repair itself, probably not.
+        return false;
+    }
+
+    public boolean canMakeDataIndex(final Table table, final String... keyColumnNames) {
+        return canMakeDataIndex(table, Arrays.asList(keyColumnNames));
+    }
+
+    public boolean canMakeDataIndex(final Table table, final Collection<String> keyColumnNames) {
+        return true;
```
rcaudy, line 92: Note to Ryan, understand how this applies in the case of persistent vs runtime impls.

engine/table/src/main/java/io/deephaven/engine/table/impl/indexer/DataIndexer.java
```diff
@@ -0,0 +1,256 @@
...
+     * @return the DataIndex, or null if one does not exist
+     */
+    public DataIndex getDataIndex(final ColumnSource<?>... keyColumns) {
+        return findIndex(dataIndexes, Arrays.asList(keyColumns));
+    }
+
+    /**
+     * Return a DataIndex for the given key columns, or null if no such index exists.
+     *
+     * @param sourceTable the table to index (if a new index is created by this operation)
```
rcaudy, line 108: This isn't allowed... I guess you have other code to push.

engine/api/src/main/java/io/deephaven/engine/table/DataIndex.java
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Get the index as a table.
+     *
+     * @param usePrev whether to use previous values and row keys.
+     *
+     * @return the index as a table.
+     */
+    @NotNull
+    Table table(final boolean usePrev);
```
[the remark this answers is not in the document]
rcaudy: Look at `io.deephaven.engine.table.impl.QueryTable#getSubTable(io.deephaven.engine.rowset.TrackingRowSet, io.deephaven.engine.table.ModifiedColumnSet, java.util.Map<java.lang.String,java.lang.Object>, java.lang.Object...)`;

You want something similar that inherits the parents atrributes (`getAttributes()`) and changes the `ColumnSource` map to replace the RowSet source with one that has the appropriate prev behavior.

chipkent (approved):
Python LGTM

With these changes, we also need to address https://github.com/deephaven/deephaven-core/issues/5021

rcaudy (approved):

rcaudy (approved):

deephaven-internal:

Community: https://github.com/deephaven/deephaven-docs-community/issues/174

PR #3869, opened 2023-05-24
Barrage: Fix BMP Failing OutOfBand BarrageBlinkTable Test by Respecting pre/post Snapshot Update Boundary

nbauernfeind:
I had forgotten about these great tests. It caught one issue due to how an update may be split by the creation of an empty snapshot.

### DIFF
--- a/server/src/main/java/io/deephaven/server/barrage/BarrageMessageProducer.java
+++ b/server/src/main/java/io/deephaven/server/barrage/BarrageMessageProducer.java
@@ -203,7 +203,7 @@ public class BarrageMessageProducer<MessageView> extends LivenessArtifact
     private volatile long lastScheduledUpdateTime = 0;
 
     private final boolean isBlinkTable;
-    /** if the parent is a blink table, then this records number of items seen since last propagation */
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/deephaven/server/barrage/BarrageMessageProducer.java
```diff
@@ -2172,6 +2170,9 @@ public boolean snapshotCompletedConsistently(final long afterClockValue, final b
                     flipSnapshotStateForSubscriptions(snapshotSubscriptions);
                     finalizeSnapshotForSubscriptions(snapshotSubscriptions);
                     promoteSnapshotToActive();
+                    // the snapshot must separate blink updates (due to subscription changes); this requires that
+                    // pre/post the snapshot are independent updates w.r.t. filtering data to within the viewport
+                    blinkTableUpdateSize = 0;
```
rcaudy, line 2175: We discussed this. You should also zero it when you clear `pendingDeltas`.

rcaudy (approved):

PR #3877, opened 2023-05-26
Correctly pass errors from executed commands to gRPC clients

### LINKED ISSUE #3876: ConsoleService.ExecuteCommand gRPC call should correctly populate ExecuteCommandResponse.error_message
If a client has permissions to invoke executeCommand, there is no risk in sending them unredacted errors. The field already exists in the ExecuteCommandResponse, but it isn't populated, so clients need to deal with logs to find error messages.

niloc132:
Changes script session to not throw when the invoked command has an
error, but instead returns a Changes instance which includes the error
object.

Also adjusts the JS IdeSession.runCode method to fail with a changes object, but the toString() method will continue to provide the error message.

Fixes #3876

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
@@ -152,12 +152,6 @@ public abstract class AbstractScriptSession<S extends AbstractScriptSession.Snap
             }
         }
 
-        // re-throw any captured exception now that our listener knows what query scope state had changed prior
-        // to the script session execution error
-        if (evaluateErr != null) {
-            throw evaluateErr;
-        }
-
         return diff;
     }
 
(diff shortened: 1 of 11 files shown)


mofojed (approved):

devinrsmith (approved):
From issue:

> If a client has permissions to invoke executeCommand, there is no risk in sending them unredacted errors.

I wouldn't universally say this is true. This is true with our default setup; but if an integrator has custom `io.deephaven.auth.codegen.impl.ConsoleServiceAuthWiring#onMessageReceivedExecuteCommand`, they don't have "full exec" permissions. That said, it may still be reasonable to show them the full logs.

I think we should have a ticket for controlling if unredacted logs get passed to the client more generally for the other RPCs. I think tying into authorization makes sense.

niloc132:
More or less agreed, two points:
 * If a user can already run code directly, they can already technically wrap with a try/catch, or find some other way to exfiltrate the error message, stack trace.
 * My statement was trying to qualify that _as long as_ they effectively have access to the logs, there is no problem with giving them an unredacted error message from their own code.

PR #3879, opened 2023-05-30
R Client 1.0

alexpeters1208:
The first shipment for the new DHC R client provides support for connecting to a Deephaven server with anon, basic, or custom authentication, pulling Deephaven tables on the server into R, pushing R data onto the server, and running Python or Groovy scripts on the server.


R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,113 @@
...
+    SEXP getStream() {
+        ArrowArrayStream* stream = new ArrowArrayStream;
+
+        std::shared_ptr<arrow::flight::FlightStreamReader> fsr = internal_tbl_hdl.getFlightStreamReader();
+
+        // allocate memory for vector of RecordBatches and populate with fsr
+        std::vector<std::shared_ptr<arrow::RecordBatch>> empty_record_batches;
+        DEEPHAVEN_EXPR_MSG(fsr->ReadAll(&empty_record_batches)); // need to add OK or throw
+
+        // convert to RecordBatchReader
```
chipkent, line 47: DRY.  Comment repeats the code.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,113 @@
...
+    TableHandleWrapper* dropColumns(std::vector<std::string> columnSpecs) {
+        return new TableHandleWrapper(internal_tbl_hdl.dropColumns(columnSpecs));
+    };
+
+    TableHandleWrapper* update(std::vector<std::string> columnSpecs) {
+        return new TableHandleWrapper(internal_tbl_hdl.update(columnSpecs));
+    };
+
+    SEXP getStream() {
+        ArrowArrayStream* stream = new ArrowArrayStream;
```
chipkent, line 39: 1) Your code is not minimizing the scope of variables.  This is not needed until the bottom of the function, but it is defined at the top.
2) `ArrowArrayStream* stream = new ArrowArrayStream();` would be more clear.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,113 @@
...
+        std::vector<std::shared_ptr<arrow::RecordBatch>> empty_record_batches;
+        DEEPHAVEN_EXPR_MSG(fsr->ReadAll(&empty_record_batches)); // need to add OK or throw
+
+        // convert to RecordBatchReader
+        arrow::Result<std::shared_ptr<arrow::RecordBatchReader>> record_batch_reader = arrow::RecordBatchReader::Make(empty_record_batches);
+
+        //export to C struct
+        arrow::ExportRecordBatchReader(record_batch_reader.ValueOrDie(), stream);
+
+        // wrap pointer in external pointer and return
```
chipkent, line 53: DRY. Comment repeats the code.  If you want a useful comment, then comment about why `XPtr` needs to be returned.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,113 @@
...
+        Rcpp::XPtr<ArrowArrayStream> xptr(stream, true);
+        return xptr;
```
chipkent, line 55: I think you can make this one line.
```suggestion
        return Rcpp::XPtr<ArrowArrayStream>(stream, true);
```

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,113 @@
...
+    };
+
+    void print() {
+        std::cout << internal_tbl_hdl.stream(true) << '\n';
+    };
```
chipkent, line 60: Delete or reformulate before shipping.

R/rdeephaven/src/client.cpp
chipkent, line 1: Does everything here need documentation?  e.g. does Rcpp generate docs from the docs here?

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+
+class ClientWrapper {
+public:
+    TableHandleWrapper* emptyTable(int64_t size) {
+        return new TableHandleWrapper(internal_tbl_hdl_mngr.emptyTable(size));
+    };
+    TableHandleWrapper* openTable(std::string tableName) {
+        return new TableHandleWrapper(internal_tbl_hdl_mngr.fetchTable(tableName));
+    };
```
chipkent, line 74: Consider the naming of this function.  What do each of our supported langs use for their names?

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,113 @@
...
+    deephaven::client::TableHandle internal_tbl_hdl;
+};
+
+
+class ClientWrapper {
+public:
+    TableHandleWrapper* emptyTable(int64_t size) {
+        return new TableHandleWrapper(internal_tbl_hdl_mngr.emptyTable(size));
+    };
+    TableHandleWrapper* openTable(std::string tableName) {
```
chipkent, line 72: General comment... whitespace is your friend.  Add spaces to make things look nice, or better yet, use a code formatter.  I'm sure VS Code has one.

https://code.visualstudio.com/docs/cpp/cpp-ide#_code-formatting

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,113 @@
...
+    ClientWrapper(deephaven::client::Client ref) : internal_client(std::move(ref)) {};
+    const deephaven::client::Client internal_client;
+    const deephaven::client::TableHandleManager internal_tbl_hdl_mngr = internal_client.getManager();
+    friend ClientWrapper* newClientWrapper(const std::string &target);
+};
+
+// factory method for calling private constructor, Rcpp does not like <const std::string &target> in constructor
+ClientWrapper* newClientWrapper(const std::string &target) {
+    return new ClientWrapper(deephaven::client::Client::connect(target));
+};
```
chipkent, line 86: 1) It looks like you define this function but don't use it in the R.
2) "Wrapper" should not appear in the external API that users will see.  What should this function be named.
3) Comment indicates that there are problems still.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+RCPP_MODULE(ClientModule) {
+
+    class_<TableHandleWrapper>("TableHandle")
+    .method("select", &TableHandleWrapper::select)
+    .method("view", &TableHandleWrapper::view)
+    .method("drop_columns", &TableHandleWrapper::dropColumns)
+    .method("update", &TableHandleWrapper::update)
+    .method("print", &TableHandleWrapper::print)
+    .method(".get_stream", &TableHandleWrapper::getStream)
```
chipkent, line 105: method name

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+// ######################### RCPP GLUE #########################
+
+using namespace Rcpp;
+
+RCPP_EXPOSED_CLASS(TableHandleWrapper)
+RCPP_EXPOSED_CLASS(BatchReader)
+RCPP_EXPOSED_CLASS(ArrowArrayStream)
+
+RCPP_MODULE(ClientModule) {
```
chipkent, line 97: Better module name.  Is `RDeephaven` or something else more appropriate?

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,113 @@
...
+    .method("drop_columns", &TableHandleWrapper::dropColumns)
+    .method("update", &TableHandleWrapper::update)
+    .method("print", &TableHandleWrapper::print)
+    .method(".get_stream", &TableHandleWrapper::getStream)
+    ;
+
+    class_<ClientWrapper>("Client")
+    .factory<const std::string&>(newClientWrapper)
+    .method("empty_table", &ClientWrapper::emptyTable)
+    .method("open_table", &ClientWrapper::openTable)
```
chipkent, line 111: See method name comments.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,113 @@
...
+    .method("select", &TableHandleWrapper::select)
+    .method("view", &TableHandleWrapper::view)
+    .method("drop_columns", &TableHandleWrapper::dropColumns)
+    .method("update", &TableHandleWrapper::update)
+    .method("print", &TableHandleWrapper::print)
+    .method(".get_stream", &TableHandleWrapper::getStream)
+    ;
+
+    class_<ClientWrapper>("Client")
+    .factory<const std::string&>(newClientWrapper)
```
chipkent, line 109: Is this defining what gets invoked in the `new(Client,..)` syntax.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+#include <arrow/c/abi.h>
+#include <arrow/c/bridge.h>
+
+#include <Rcpp.h>
+
+
+// ######################### DH WRAPPERS, API #########################
+
+class TableHandleWrapper {
```
chipkent, line 18: If we can get this close to shipping, we can consider breaking this into a `.h` and a `.cpp`.  Right now, everything is defined in one place -- Java style.  Not a problem for now.

R/rdeephaven/lib/cpp-dependencies/build-dependencies.sh
```diff
@@ -0,0 +1,390 @@
+#!/bin/bash
```
chipkent, line 1: Is this file somewhere else in the repo, or is this the only location?  If somewhere else, this should be a symlink.

R/rdeephaven/lib/cpp-dependencies/build-dependencies.sh
```diff
@@ -0,0 +1,390 @@
...
+          exit 1
+          ;;
+  esac
+fi
+
+# At the point of this writing, the latest immer release is pretty old.
+# We want something a lot more recent, but don't want to track head as is a moving
+# target and we can't guarantee things will continue to compile/be consistent.
+# So we select a particular SHA.
+: ${IMMER_SHA:=e5d79ed80ec74d511cc4f52fb68feeac66507f2c}
```
chipkent, line 116: How are we going to keep this version in sync?

R/rdeephaven/lib/cpp-dependencies/build-dependencies.sh
```diff
@@ -0,0 +1,390 @@
...
+# there is no guarantee where the CWD is after a prior phase.
+#
+: ${GIT_FLAGS:="--quiet -c advice.detachedHead=false"}
+
+### Protobuf
+if [ "$BUILD_PROTOBUF" = "yes" ]; then
+  echo
+  echo "*** Clone protobuf"
+  cd $SRC
+  git clone $GIT_FLAGS -b v3.20.1 --depth 1 https://github.com/protocolbuffers/protobuf.git
```
chipkent, line 167: How do we keep this version in sync?

R/rdeephaven/lib/cpp-dependencies/build-dependencies.sh
```diff
@@ -0,0 +1,390 @@
...
+  fi
+  echo "*** protobuf DONE"
+fi
+
+### re2
+if [ "$BUILD_RE2" = "yes" ]; then
+  echo
+  echo "*** Clone re2"
+  cd $SRC
+  git clone $GIT_FLAGS -b 2022-04-01 --depth 1 https://github.com/google/re2.git
```
chipkent, line 186: How do we keep this version in sync?

R/rdeephaven/lib/cpp-dependencies/build-dependencies.sh
```diff
@@ -0,0 +1,390 @@
...
+  fi
+  echo "*** ares DONE"
+fi
+
+### zlib
+if [ "$BUILD_ZLIB" = "yes" ]; then
+  echo
+  echo "*** Clone zlib"
+  cd $SRC
+  git clone $GIT_FLAGS -b v1.2.11 --depth 1 https://github.com/madler/zlib
```
chipkent, line 281: How do we keep this version in sync?

R/rdeephaven/lib/cpp-dependencies/build-dependencies.sh
```diff
@@ -0,0 +1,390 @@
...
+  fi
+  echo "*** grpc DONE"
+fi
+
+### arrow
+if [ "$BUILD_ARROW" = "yes" ]; then
+  echo
+  echo "*** Clone arrow"
+  cd $SRC
+  git clone $GIT_FLAGS -b apache-arrow-7.0.0 --depth 1 https://github.com/apache/arrow
```
chipkent, line 321: How do we keep this version in sync?

R/rdeephaven/lib/cpp-dependencies/build-dependencies.sh
chipkent: Currently, all of the dependency versions are sprinkled all through this file.  It may be cleaner to consolidate them all here.

R/rdeephaven/src/Makevars
```diff
@@ -0,0 +1,68 @@
...
+SOURCES = $(wildcard *.cpp)
+# compiled objects to link
+OBJECTS = $(SOURCES:.cpp=.o)
+
+all: $(SHLIB)
+
+clean:
+	find . -type f -name '*.o' -delete
+	find . -type f -name '*.so' -delete
+	rm RcppExports.cpp
```
chipkent, line 68: It seems strange for `RcppExports.cpp` to be removed by clean but committed to git.  I assume that the file should be removed from git.

R/rdeephaven/lib/build-cpp.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+export SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+export CLIENT=$SCRIPT_DIR/cpp-client
+
+# First, we build all C++ dependencies
+# cd cpp-dependencies
+# chmod +x ./build-dependencies.sh
+# ./build-dependencies.sh
+# cd $SCRIPT_DIR
```
chipkent, line 10: Why commented out?

R/rdeephaven/lib/build-cpp.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# Next, we download, build, and install the Deephaven C++ client
+svn checkout https://github.com/deephaven/deephaven-core/trunk/cpp-client
```
chipkent, line 13: This looks like it is just pulling `main`.  Should this be versioned or locked to the SHA of the current branch or something?

R/rdeephaven/lib/build-cpp.sh
```diff
@@ -0,0 +1,28 @@
...
+
+export CLIENT=$SCRIPT_DIR/cpp-client
+export DEEPHAVEN_LOCAL=$SCRIPT_DIR/cpp-dependencies/local
+export CMAKE_PREFIX_PATH=${DEEPHAVEN_LOCAL}/abseil:${DEEPHAVEN_LOCAL}/boost:${DEEPHAVEN_LOCAL}/cares:${DEEPHAVEN_LOCAL}/flatbuffers:${DEEPHAVEN_LOCAL}/gflags:${DEEPHAVEN_LOCAL}/immer:${DEEPHAVEN_LOCAL}/protobuf:${DEEPHAVEN_LOCAL}/re2:${DEEPHAVEN_LOCAL}/zlib:${DEEPHAVEN_LOCAL}/grpc:${DEEPHAVEN_LOCAL}/arrow:${DEEPHAVEN_LOCAL}/deephaven
+export NCPUS=$(getconf _NPROCESSORS_ONLN)
+cmake -DCMAKE_INSTALL_PREFIX=${DEEPHAVEN_LOCAL}/deephaven .. && make -j$NCPUS install
+cd $SCRIPT_DIR
+
+# Finally, we download and build the examples
+svn checkout https://github.com/deephaven/deephaven-core/trunk/cpp-examples
```
chipkent, line 25: versioning

R/rdeephaven/lib/build-cpp.sh
chipkent: Is this really in subversion?!

R/rdeephaven/lib/build-cpp.sh
chipkent: Is this really in subversion?!

R/rdeephaven/lib/build-cpp.sh
chipkent, line 1: Build stuff eventually needs to be built into CRAN CI and/or DH CI.

chipkent (changes requested):

R/rdeephaven/NAMESPACE
```diff
@@ -0,0 +1,3 @@
+useDynLib(rdeephaven, .registration=TRUE)
+importFrom(Rcpp, evalCpp)
+exportPattern("^[[:alpha:]]+")
```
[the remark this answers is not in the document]
alexpeters1208: Yes, it is very easy to control permissions from here. The tricky bit is that anything hidden from the user in this way also gets hidden from any internal R stuff that might need to use what has been hidden. Primary example is .getStream().

R/rdeephaven/tests/test_connection.R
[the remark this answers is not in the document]
alexpeters1208: Definitely, this is just a file where I'm playing with stuff form the R side and iterating on the API

R/rdeephaven/tests/test_connection.R
```diff
@@ -0,0 +1,15 @@
...
+# load deephaven library
+library(rdeephaven)
+
+# connect to server using Client object
+client <- new(Client, target="localhost:10000")
+
+# open DH table and convert to arrow table
+static_table <- client$open_table("static_table1")
+arrow_table <- to_arrow_table(static_table)
+data_frame <- as.data.frame(arrow_table)
```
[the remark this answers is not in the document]
alexpeters1208: Implemented a to_data_frame function, will look at the other 2 points.

R/rdeephaven/tests/demo1.R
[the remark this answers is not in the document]
alexpeters1208: Yes.

R/rdeephaven/tests/demo1.R
```diff
@@ -0,0 +1,35 @@
+# load required libraries
+library(Rcpp)
+library(arrow)
+library(rdeephaven)
+
+# connect to server using Client object
+client <- new(Client, target="localhost:10000")
```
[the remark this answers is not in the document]
alexpeters1208: To me, the first feels more R. I do think `Client` could be renamed to something like `DeephavenClient`, but the `new` idiom is something that I'm used to seeing when creating objects from R6 land.

R/rdeephaven/src/Makevars
alexpeters1208: It belongs in git. I'm cleaning it here because having old versions of that file hanging around really screwed me up last week, and I want to make sure it gets deleted and redone every time I run clean. In the end it will not be in clean.

R/rdeephaven/src/client.cpp
alexpeters1208: Made the change, would like to know what difference `()` makes.

R/rdeephaven/src/client.cpp
alexpeters1208: Yes, for personal testing only.

R/rdeephaven/src/client.cpp
alexpeters1208: this is TBD

R/rdeephaven/src/client.cpp
alexpeters1208: It does appear in R, using the Rcpp client:
`class_<ClientWrapper>("Client")`
`.factory<const std::string&>(newClientWrapper)`
`.method("empty_table", &ClientWrapper::emptyTable)`
`.method("open_table", &ClientWrapper::openTable)`
`;`

uses that as a factory method without exposing its name to the user. This is the method that gets invoked in the call to `new`

R/rdeephaven/src/client.cpp
alexpeters1208: Depends on whether we break this file up into several, like `client.cpp`, `tableHandle.cpp`, etc. Then the Rcpp module at the bottom of each should be named appropriately.

R/rdeephaven/src/client.cpp
alexpeters1208: Yes

R/rdeephaven/lib/cpp-dependencies/build-dependencies.sh
alexpeters1208: There is a modification to this version of it that builds Cares as a static library, because that is what R wants. So, it's not identical to [this](https://github.com/deephaven/deephaven-base-images/blob/main/cpp-client/build-dependencies.sh), which also doesn't live in the same repo.

R/rdeephaven/lib/build-cpp.sh
alexpeters1208: Commented out so I could rebuild the client without rebuilding the dependencies. I should add that as a flag argument

R/rdeephaven/tests/test_connection.R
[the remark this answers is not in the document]
alexpeters1208: I briefly looked at Tidyverse style guide and all of their example functions use `_` in function names.

R/rdeephaven/lib/build-cpp.sh
alexpeters1208: See above

R/rdeephaven/lib/build-cpp.sh
chipkent: Try this:  https://git-scm.com/docs/git-sparse-checkout

https://medium.com/@marcoscannabrava/git-download-a-repositorys-specific-subfolder-ceeabc6023e2
https://github.blog/2020-01-17-bring-your-monorepo-down-to-size-with-sparse-checkout/

R/rdeephaven/R/helper_functions.R
```diff
@@ -0,0 +1,121 @@
...
+#' @description Converts an Arrow RecordBatchReader to a Deephaven Table, pushes that new table to the Deephaven server,
+#' and returns a TableHandle corresponding to the new table.
+#' 
+#' @param client A Deephaven Client object maintaining the connection to the server.
+#' @param rbr An Arrow RecordBatchReader containing the data to be converted to a Deephaven Table.
+#' @param name A string containing the name of the new Deephaven Table to be created on the server.
+#' @returns A Deephaven TableHandle, providing a reference to the new table on the server.
+rbr_to_dh_table <- function(client, rbr, name) {
+
+    if (class(client)[[1]] != "Rcpp_Client") {
```
chipkent, line 57: Why is `Rcpp_` in the type name?

R/rdeephaven/R/helper_functions.R
```diff
@@ -0,0 +1,121 @@
...
+#' @description Converts an Arrow Table to a Deephaven Table, pushes that new table to the Deephaven server,
+#' and returns a TableHandle corresponding to the new table.
+#' 
+#' @param client A Deephaven Client object maintaining the connection to the server.
+#' @param arrow_tbl An Arrow Table containing the data to be converted to a Deephaven Table.
+#' @param name A string containing the name of the new Deephaven Table to be created on the server.
+#' @returns A Deephaven TableHandle, providing a reference to the new table on the server.
+arrow_to_dh_table <- function(client, arrow_tbl, name) {
+
+    if (class(client)[[1]] != "Rcpp_Client") {
```
chipkent, line 83: Why is `Rcpp_` in the type name?

R/rdeephaven/R/helper_functions.R
```diff
@@ -0,0 +1,121 @@
...
+#' @description Converts an R Data Frame to a Deephaven Table, pushes that new table to the Deephaven server,
+#' and returns a TableHandle corresponding to the new table.
+#' 
+#' @param client A Deephaven Client object maintaining the connection to the server.
+#' @param data_frame An R Data Frame containing the data to be converted to a Deephaven Table.
+#' @param name A string containing the name of the new Deephaven Table to be created on the server.
+#' @returns A Deephaven TableHandle, providing a reference to the new table on the server.
+df_to_dh_table <- function(client, data_frame, name) {
+    
+    if (class(client)[[1]] != "Rcpp_Client") {
```
chipkent, line 108: Why is `Rcpp_` in the type name?

R/rdeephaven/R/helper_functions.R
```diff
@@ -0,0 +1,121 @@
...
+
+#' @title Push an R Data Frame to the server as a Deephaven Table.
+#' @description Converts an R Data Frame to a Deephaven Table, pushes that new table to the Deephaven server,
+#' and returns a TableHandle corresponding to the new table.
+#' 
+#' @param client A Deephaven Client object maintaining the connection to the server.
+#' @param data_frame An R Data Frame containing the data to be converted to a Deephaven Table.
+#' @param name A string containing the name of the new Deephaven Table to be created on the server.
+#' @returns A Deephaven TableHandle, providing a reference to the new table on the server.
+df_to_dh_table <- function(client, data_frame, name) {
```
chipkent, line 106: see other comments on this method.

chipkent (changes requested):

R/rdeephaven/prototyping/tmp.R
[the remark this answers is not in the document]
alexpeters1208: The name is because it's temporary. Just for me, will not make it into the final package. Stored in git so that I still have it if my laptop explodes.

R/rdeephaven/prototyping/tmp.R
```diff
@@ -0,0 +1,31 @@
+Rcpp::compileAttributes()
+roxygen2::roxygenise()
+install.packages("/home/user/rdeephaven", repos=NULL, type="source")
+
+# load deephaven library
+library(rdeephaven)
+
+# connect to server using Client object
+# I hate this api more than Rcpp hates default arguments
```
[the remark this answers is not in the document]
alexpeters1208: I lost a day trying to do this any other way. Rcpp does not do default or optional arguments and I have not yet figured out how to get into the R6 code to fix it by hand. Right now all of the args have to be provided, which I can't stand.

R/rdeephaven/prototyping/tmp.R
```diff
@@ -0,0 +1,31 @@
...
+client <- new(Client, target="192.168.1.241:10000", session_type="python",
+                      auth_type="default", key="", value="")
```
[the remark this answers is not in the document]
alexpeters1208: 1. Colin got me on this for running on servers outside of the docker container that this is all running in.
2. Noted

R/rdeephaven/prototyping/tmp.R
```diff
@@ -0,0 +1,31 @@
...
+static_table <- client$open_table("static_table1")
+data_frame1 <- dh_to_data_frame(static_table)
+data_frame1
+
+# modify dataframe with regular R stuff
+data_frame1$Name_Float_Col <- c(3.14, 2.71, 0.00)
+data_frame1
+
+# push new dataframe back to DH server and retrieve as DH table
+new_static_table <- df_to_dh_table(client, data_frame1, "static_table2")
```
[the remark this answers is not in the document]
alexpeters1208: `client$push()` is hard at the moment because we need a pointer from the C++ side and a function call from the R side to make it work. I can't put the R function call directly into the C++ code to wrap everything in `$push()`, and I can't yet get into the S4 to put it there by hand. If you notice, there's not a single S4 class to be found anywhere in this source code. That is still black magic to me.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,264 @@
...
+#include "deephaven/client/client.h"
+#include "deephaven/client/flight.h"
+
+#include <arrow/c/abi.h>
+#include <arrow/c/bridge.h>
+
+#include <Rcpp.h>
+
+
+// TODO: good error handling
```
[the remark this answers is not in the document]
alexpeters1208: Working on it.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,264 @@
...
+            runScript(delete_script);
+        }
+        else if (session_type == "groovy") {
+            std::string delete_script = tableName + " = null";
+            runScript(delete_script);
+        }
+        else {
+            std::cout << "Error: Cannot delete table without an active console.";
+        }
+    };
```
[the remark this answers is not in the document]
alexpeters1208: I guess I would have liked to have it, but it's not necessary if we don't want it.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,264 @@
...
+    bool checkForTable(std::string tableName) {
+        // we have to first fetchTable to check existence, fetchTable does not fail on its own, but .observe() will fail if table doesn't exist
+        deephaven::client::TableHandle table_handle = internal_tbl_hdl_mngr.fetchTable(tableName);
+        try {
+            table_handle.observe();
+        } catch(...) {
+            return false;
+        }
+        return true;
+    };
```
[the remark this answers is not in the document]
alexpeters1208: It's useful in testing to ensure that a thing that supposedly got pushed is actually there. There is probably a better way.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * Runs a script on the server in the console language if a console was created.
+     * @param code String of the code to be executed on the server.
+    */
+    void runScript(std::string code) {
+        if (!(session_type == "python" || session_type == "groovy")) {
+            std::cout << "Error: Cannot run a script without an active console.";
+        }
```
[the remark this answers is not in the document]
alexpeters1208: I don't yet know how to propagate errors up from C++ but present them cleanly. This is providing a kind message now with the hope that I can find a more robust solution after shipment #1.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,264 @@
...
+ * @param sessionType Type of console to start with this client connection, can be "none", "python", or "groovy". The ClientWrapper::runScript() method will only work
+ *                    with the language specified here.
+ * @param authType Type of authentication to use, can be "default", "basic", or "custom". "basic" uses username/password auth, "custom" uses general key/value auth.
+ * @param key Key credential for authentication, can be a username if authType is "basic", or a general key if authType is "custom". Set to "" if authType is "default".
+ * @param value Value credential for authentication, can be a password if authType is "basic", or a general value if authType is "custom". Set to "" if authType is "default".
+ */
+ClientWrapper* newClientWrapper(const std::string &target, const std::string &sessionType,
+                                const std::string &authType, const std::string &key, const std::string &value) {
+
+    deephaven::client::ClientOptions client_options = deephaven::client::ClientOptions();
```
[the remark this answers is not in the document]
alexpeters1208: I considered this but did not implement. Corey was not around and I couldn't make sense of the C++. This might be the way to go.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,264 @@
...
+RCPP_EXPOSED_CLASS(ArrowArrayStream)
+
+RCPP_MODULE(ClientModule) {
+
+    class_<TableHandleWrapper>("TableHandle")
+    .method("select", &TableHandleWrapper::select)
+    .method("view", &TableHandleWrapper::view)
+    .method("drop_columns", &TableHandleWrapper::dropColumns)
+    .method("update", &TableHandleWrapper::update)
+    .method(".get_arrowArrayStream_ptr", &TableHandleWrapper::getArrowArrayStreamPtr)
```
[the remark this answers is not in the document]
alexpeters1208: 1. To indicate internal use only (not yet documented) while we figure out how to hide that stuff.
2. To me, ArrowArrayStream is one thing, that's what I'm getting a pointer to. Snake case works too.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,264 @@
...
+    ;
+
+    class_<ClientWrapper>("Client")
+    .factory<const std::string&, const std::string&, const std::string&, const std::string&, const std::string&>(newClientWrapper)
+    .method("open_table", &ClientWrapper::openTable)
+    .method("delete_table", &ClientWrapper::deleteTable)
+    .method("check_for_table", &ClientWrapper::checkForTable)
+    .method("run_script", &ClientWrapper::runScript)
+    .method(".new_arrowArrayStream_ptr", &ClientWrapper::newArrowArrayStreamPtr)
+    .method(".new_table_from_arrowArrayStream_ptr", &ClientWrapper::newTableFromArrowArrayStreamPtr)
```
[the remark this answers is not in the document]
alexpeters1208: 1. To indicate internal use only (not yet documented) while we figure out how to hide that stuff.
2. To me, ArrowArrayStream is one thing, that's what I'm getting a pointer to. Snake case works too.

R/rdeephaven/R/helper_functions.R
```diff
@@ -0,0 +1,121 @@
+################ FUNCTIONS TO GET FROM DH TO R
```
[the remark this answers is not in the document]
alexpeters1208: I agree with you. This is where being unable to access the S4 internals in the source code is a pain. Since these require R functionality, they can't be walled up in the C++, and the proper methods cannot get added to the S4 without access.

R/rdeephaven/R/helper_functions.R
alexpeters1208: That's how Rcpp has put it together. I'm just now wondering if I can wrap all of the Rcpp-generated S4 stuff in an R6 class and tailor all the methods and names by hand. That would provide a solution to a lot of what's been discussed in this round of review. The problem with that would be that the underlying Rcpp stuff would likely still be visible at a user level, which sucks.

R/rdeephaven/prototyping/tmp.R
```diff
@@ -0,0 +1,31 @@
...
+# I hate this api more than Rcpp hates default arguments
+client <- new(Client, target="192.168.1.241:10000", session_type="python",
+                      auth_type="default", key="", value="")
+
+
+client$check_for_table("static_table2")
+
+# open DH table and convert to R dataframe
+static_table <- client$open_table("static_table1")
+data_frame1 <- dh_to_data_frame(static_table)
```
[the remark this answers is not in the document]
alexpeters1208: This is 100% what it should be.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,102 @@
...
+        },
+
+        # import an R object to a table to the server, and retrieve a TableHandle reference to that table
+        import_table = function(tableObject) {
+            if (!(class(tableObject)[[1]] == "data.frame" ||
+                  class(tableObject)[[1]] == "RecordBatchReader" ||
+                  (length(class(tableObject)) == 4 &&
+                   class(tableObject)[[1]] == "Table" &&
+                   class(tableObject)[[3]] == "ArrowObject"))) {
+                stop("'tableObject' must be either an R Data Frame, an Arrow Table, or an Arrow Record Batch Reader.")
```
chipkent, line 45: May be useful to print the object class.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,102 @@
...
+            }
+            if (class(tableObject)[[1]] == "data.frame") {
+                this_dh_table_handle = private$df_to_dh_table(tableObject)
```
chipkent, line 48:
```suggestion
                return(NewTableHandle$new(private$df_to_dh_table(tableObject)))
```
Could change these like this to reduce variable scope and clean up the code.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,102 @@
...
+            else if (class(tableObject)[[1]] == "RecordBatchReader") {
+                this_dh_table_handle = private$rbr_to_dh_table(tableObject)
+            }
+            else if ((length(class(tableObject)) == 4 &&
+                      class(tableObject)[[1]] == "Table" &&
+                      class(tableObject)[[3]] == "ArrowObject")) {
+                this_dh_table_handle = private$arrow_to_dh_table(tableObject)
+            }
+            else {
+                stop("An unexpected type was passed to 'push' and made it through our stop handling. 'tableObject' must be either an R Data Frame, an Arrow Table, or an Arrow Record Batch Reader.")
```
chipkent, line 59:
```suggestion
                stop("An unexpected type was passed to 'import_table' and made it through our stop handling. 'tableObject' must be either an R Data Frame, an Arrow Table, or an Arrow Record Batch Reader.")
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,102 @@
...
+            }
+            return(NewTableHandle$new(this_dh_table_handle))
+        },
+
+        run_script = function(script) {
+            private$verify_string(script, "script")
+            private$internal_client$run_script(script)
```
chipkent, line 66: This has me wondering about c++ error handling.  What happens if C++ throws an exception?  Do we behave ok?

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,102 @@
...
+
+        check_for_table = function(tableName) {
+            return(private$internal_client$check_for_table(tableName))
+        },
+
+        verify_string = function(stringCandidate, argName) {
+            if (class(stringCandidate) != "character") {
+                stop(paste0("'", argName, "' must be passed as a string."))
+            }
+        },
```
chipkent, line 81: This looks like a general purpose method that could be used in TableHandle and other things.  I suggest pulling it out into a `_verify_string` function.  Since this doesn't impact the API, it could happen post release.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,102 @@
...
+
+        rbr_to_dh_table = function(rbr) {
+            ptr = private$internal_client$new_arrowArrayStream_ptr()
```
chipkent, line 84: When I reviewed this before, I think it can be a function not associated with the client.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,102 @@
...
+            dh_table = private$rbr_to_dh_table(rbr)
+            return(dh_table)
+        },
+
+        df_to_dh_table = function(dataFrame) {
+            arrow_tbl = arrow_table(dataFrame)
+            dh_table = private$arrow_to_dh_table(arrow_tbl)
+            return(dh_table)
+        }
+    )
```
chipkent, line 101: If we support `to_tibble`, we should also support it in `import_table`.

chipkent (changes requested):

R/rdeephaven/R/helper_functions.R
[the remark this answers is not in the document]
alexpeters1208: It's out.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -0,0 +1,45 @@
...
+
+        to_arrow_table = function() {
+            rbsr = self$to_record_batch_stream_reader()
+            arrow_tbl = rbsr$read_table()
+            return(arrow_tbl)
+        },
+
+        to_data_frame = function() {
+            arrow_tbl = self$to_arrow_table()
+            data_frame = as.data.frame(as.data.frame(arrow_tbl))
```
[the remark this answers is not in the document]
alexpeters1208: as.data.frame(arrow_tbl) gives me a tibble, I have no idea why. to_tibble is easy to implement

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,102 @@
...
+                stop("'credentials' must be passed as a list of two strings representing a key/value pair for authentication.")
+            }
+
+            # now value checking
+            if (!(sessionType == "python" || sessionType == "groovy")) {
+                warning("The session type you provided is not currently supported. Defaulting to 'python'.")
+            }
+            if (!(authType == "default" || authType == "basic" || authType == "custom")) {
+                warning("The authentication type you provided is not currently supported. Defaulting to 'default'.")
+            }
```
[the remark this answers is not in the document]
alexpeters1208: When you say client creation, do you mean letting the underlying C++ call generate the error?

R/rdeephaven/R/client_wrapper.R
alexpeters1208: But in order to get the pointer, we ask the client to create it. Could take the client as an arg, but that seems to defeat the purpose.

R/rdeephaven/src/client.cpp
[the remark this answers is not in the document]
chipkent: I would lean to skipping the garbage check and just get a fix in soon.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,153 @@
...
+            if (class(stringCandidate) != "character") {
+                stop(paste0("'", arg_name, "' must be passed as a string."))
+            }
+        },
+
+        rbr_to_dh_table = function(rbr) {
+            ptr = private$internal_client$new_arrow_array_stream_ptr()
+            rbr$export_to_c(ptr)
+            dh_table = private$internal_client$new_table_from_arrow_array_stream_ptr(ptr)
+            return(dh_table)
```
chipkent, line 132:
```suggestion
            return( private$internal_client$new_table_from_arrow_array_stream_ptr(ptr) )
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,153 @@
...
+        },
+
+        arrow_to_dh_table = function(arrow_tbl) {
+            rbr = as_record_batch_reader(arrow_tbl)
+            dh_table = private$rbr_to_dh_table(rbr)
+            return(dh_table)
```
chipkent, line 138:
```suggestion
            return(private$rbr_to_dh_table(rbr))
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,153 @@
...
+        },
+
+        tibble_to_dh_table = function(tibbl) {
+            arrow_tbl = arrow_table(tibbl)
+            dh_table = private$arrow_to_dh_table(arrow_tbl)
+            return(dh_table)
```
chipkent, line 144:
```suggestion
            return(rivate$arrow_to_dh_table(arrow_tbl))
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,153 @@
...
+        },
+
+        df_to_dh_table = function(data_frame) {
+            arrow_tbl = arrow_table(data_frame)
+            dh_table = private$arrow_to_dh_table(arrow_tbl)
+            return(dh_table)
```
chipkent, line 150:
```suggestion
            return(private$arrow_to_dh_table(arrow_tbl))
```

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -0,0 +1,48 @@
...
+
+        to_tibble = function() {
+            rbsr = self$to_arrow_record_batch_stream_reader()
+            arrow_tbl = rbsr$read_table()
+            return(as_tibble(arrow_tbl))
+        },
+
+        to_data_frame = function() {
+            arrow_tbl = self$to_arrow_table()
+            return(as.data.frame(as.data.frame(arrow_tbl))) # for some reason as.data.frame on arrow table returns a tibble, not what I want
```
chipkent, line 41: This looks wrong, as I noted in the other comment that appears closed.

Also "I" , "we", "us" etc should not be in the comments.  You are developing in public, so all comments need to be "professional" for the project.

R/rdeephaven/Read-and-delete-me
chipkent, line 1: After we go through the ckecklist, we need to delete this.  Probably post release.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,153 @@
...
+#' - `$delete_table(name)`: Looks for a table named 'name' on the server, and deletes it if it exists.
+#'    Importantly, this is only effective 
+#' - `check_for_table`: 
+#' - `run_script`:
+
+#################################################################################################
+
+
+
+DeephavenClient <- R6Class("DeephavenClient",
```
chipkent, line 42: Do we have a reason to call this `DeephavenClient` instead of just `Client`?  We have package namespacing, so it isn't obvious that we need dh in the name.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,113 @@
...
+    };
+
+    SEXP getStream() {
+        ArrowArrayStream* stream = new ArrowArrayStream;
+
+        std::shared_ptr<arrow::flight::FlightStreamReader> fsr = internal_tbl_hdl.getFlightStreamReader();
+
+        // allocate memory for vector of RecordBatches and populate with fsr
+        std::vector<std::shared_ptr<arrow::RecordBatch>> empty_record_batches;
+        DEEPHAVEN_EXPR_MSG(fsr->ReadAll(&empty_record_batches)); // need to add OK or throw
```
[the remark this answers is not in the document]
chipkent: ERROR handling

R/rdeephaven/src/client.cpp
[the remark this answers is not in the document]
chipkent: CAn't you just rerequest the tablehandle?
CAn be fixed post merge.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,204 @@
...
+        while(true) {
+            std::shared_ptr<arrow::RecordBatch> this_batch;
+            DEEPHAVEN_EXPR_MSG(record_batch_reader->ReadNext(&this_batch)); // TODO: need to add ok or throw
+            if (this_batch == nullptr) {
+                break;
+            }
+            DEEPHAVEN_EXPR_MSG(fsw->WriteRecordBatch(*this_batch)); // TODO: need to add okOrThrow
+        }
+        DEEPHAVEN_EXPR_MSG(fsw->DoneWriting()); // TODO: need to add okOrThrow
+        DEEPHAVEN_EXPR_MSG(fsw->Close()); // TODO: need to add okOrThrow
```
chipkent, line 130: Bunch of error handling cases.

chipkent (changes requested):

R/rdeephaven/tests/testthat/client_infrastructure.R
```diff
@@ -0,0 +1,29 @@
...
+                      int_col = sample(0:100000, 2500000, TRUE))
+
+    return(list(df1, df2, df3, df4))
+}
+
+setup_client <- function(client_args, dataframes=NULL) {
+
+    client <- new(Client, target=client_args[1], session_type=client_args[2], auth_type=client_args[3], key=client_args[4], value=client_args[5])
+
+    if (!is.null(dataframes)) {
```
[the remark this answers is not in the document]
alexpeters1208: Argument name

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,153 @@
...
+                stop("'credentials' must be passed as a list of two strings representing a key/value pair for authentication.")
+            }
+
+            # now value checking
+            if (!(session_type == "python" || session_type == "groovy")) {
+                warning("The session type you provided is not currently supported. Defaulting to 'python'.")
+            }
+            if (!(auth_type == "default" || auth_type == "basic" || auth_type == "custom")) {
+                warning("The authentication type you provided is not currently supported. Defaulting to 'default'.")
+            }
```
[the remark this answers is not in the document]
alexpeters1208: I think I responded to one of them asking if the intention was to let C++ catch and throw the errors.

R/rdeephaven/R/table_handle_wrapper.R
alexpeters1208: I don't think I closed the comment, but I did comment and mention that this is here because as.data.frame on an arrow_table returns a tibble rather than a pure data frame.

`> static_table <- client$open_table("static_table1")`

`> data_frame <- static_table$to_data_frame()`
`> class(data_frame)`
`[1] "data.frame"`

`> arrow_table <- static_table$to_arrow_table()`
`> class(arrow_table)`
`[1] "Table"        "ArrowTabular" "ArrowObject"  "R6"`

`> tibbl <- static_table$to_tibble()`
`> class(tibbl)`
`[1] "tbl_df"     "tbl"        "data.frame"`

`> data_frame_from_arrow_table = as.data.frame(arrow_table)`
`> class(data_frame_from_arrow_table)`
`[1] "tbl_df"     "tbl"        "data.frame"`

`> new_data_frame_from_arrow_table = as.data.frame(as.data.frame(arrow_table))`
`> class(new_data_frame_from_arrow_table)`
`[1] "data.frame"`

I think this probably has more to do with R being weird about inheritance than someone being outright wrong. I can remove the second as.data.frame, but the result in that case is not what I expect.

R/rdeephaven/R/client_wrapper.R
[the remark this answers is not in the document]
chipkent: yes

R/rdeephaven/R/client_wrapper.R
chipkent: I don't think that creation has anything to do with the client.

R/rdeephaven/R/client_wrapper.R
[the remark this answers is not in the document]
chipkent: Yes.  No need to push the error handling to R if C++ is already doing it.

R/rdeephaven/R/table_handle_wrapper.R
chipkent: The big question is if it is allocating more memory or not.  If not, it doesn't matter.  If so, then it subpar but can be addressed in v2.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,126 @@
+######################################## INCOMPLETE DOCS ########################################
```
chipkent, line 1: INCOMPLETE?

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,126 @@
...
+
+#' @title The Deephaven Client
+#' @description The Deephaven Client class is responsible for establishing and maintaining
+#' a connection to a running Deephaven server and facilitating basic server requests. This is
+#' the primary interface for interacting with the Deephaven server from R.
+#' 
+#' @section Connecting to the Server via Client
+#' 
+#' To connect to a Deephaven server, you must create a new DeephavenClient object via `Client$new(...)`,
```
chipkent, line 10:
```suggestion
#' To connect to a Deephaven server, you must create a new Deephaven Client object via `Client$new(target,client_options)`,
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,126 @@
...
+#' with the following arguments:
+#' 
+#' - `target`: The URL that the Deephaven server is running on.
```
chipkent, line 13:
```suggestion
#' - `target`: The address of the Deephaven server.
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,126 @@
...
+#' - `client_options`: A `ClientOptions` instance that contains the relevant information for connecting to your server.
+#'   Check the documentation on client options via `?ClientOptions`
+#' 
+#' @section Methods
+#' 
+#' - `$open_table(name)`: Looks for a table named 'name' on the server, and returns a Deephaven TableHandle reference
```
chipkent, line 19:
```suggestion
#' - `$open_table(name)`: Looks for a table named 'name' on the server, and returns a TableHandle reference
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,126 @@
...
+#'    reference to that table if it exists. See the documentation on table handles via `?TableHandle` for more information.
+#' - `$import_table(table_object)`: Imports a new table to the Deephaven server and returns a Deephaven TableHandle
```
chipkent, line 21:
```suggestion
#' - `$import_table(table_object)`: Imports a new table to the Deephaven server and returns a TableHandle
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,126 @@
...
+#'    reference to the new table. Here, `table_object` can be an R Data Frame, an R Tibble, an Arrow Table, or an Arrow
+#'    RecordBatchReader. Note that this new table is not automatically bound to a variable name on the server.
+#'    See `?TableHandle` for more information.
+#' - `run_script(script)`: Runs a script on the server. The script must be in the language that the server console was
```
chipkent, line 25:
```suggestion
#' - `$run_script(script)`: Runs a script on the server. The script must be in the language that the server console was
```
Other methods had the $

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,126 @@
...
+#'    started with, and should be passed as a string to this method.
+#' 
+#' @examples
+#' 
+#' client_options <- ClientOptions$new()
+#' client_options$set_default_authentication()
+#' 
+#' # connect to the Deephaven server running on "localhost:10000" with anonymous 'default' authentication
+#' client <- Client$new(target="localhost:10000", client_options=client_options)
```
chipkent, line 34:
```suggestion
#' # connect to the Deephaven server running on "localhost:10000" with anonymous 'default' authentication
#'
#' client_options <- ClientOptions$new()
#' client_options$set_default_authentication()
#' 
#' client <- Client$new(target="localhost:10000", client_options=client_options)
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,126 @@
...
+#' 
+#' # open a table that already exists on the server
+#' new_table_handle1 <- client$open_table("table_on_the_server")
+#' 
+#' # create a new dataframe, import onto the server, and retrieve a reference
+#' new_data_frame <- data.frame(matrix(rnorm(10 * 1000), nrow = 10))
+#' new_table_handle2 <- client$import_table(new_data_frame)
+#' 
+#' # run a python script, since default client options specify a Python console
```
chipkent, line 43:
```suggestion
#' # run a python script on the server (default client options specify a Python console)
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,102 @@
...
+
+            private$internal_client <- new(INTERNAL_Client, target=target, sessionType=sessionType,
+                                           authType=authType, key=credentials[[1]], value=credentials[[2]])
+        },
+
+        # retrieve a TableHandle reference to a named table on the server
+        open_table = function(tableName) {
+            private$verify_string(tableName, "tableName")
+            if (!private$check_for_table(tableName)) {
+                stop("The table you're trying to pull does not exist on the server.")
```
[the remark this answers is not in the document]
chipkent: Not done.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -0,0 +1,126 @@
...
+            else if (table_object_class[[1]] == "RecordBatchReader") {
+                return(TableHandle$new(private$rbr_to_dh_table(table_object)))
+            }
+            else if ((length(table_object_class) == 4 &&
+                      table_object_class[[1]] == "Table" &&
+                      table_object_class[[3]] == "ArrowObject")) {
+                return(TableHandle$new(private$arrow_to_dh_table(table_object)))
+            }
+            else {
+                stop("'table_object' must be either an R Data Frame, an R Tibble, an Arrow Table, or an Arrow Record Batch Reader.")
```
chipkent, line 82: Error message should indicate the input class type.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -0,0 +1,47 @@
...
+
+        to_tibble = function() {
+            rbsr = self$to_arrow_record_batch_stream_reader()
+            arrow_tbl = rbsr$read_table()
+            return(as_tibble(arrow_tbl))
+        },
+
+        to_data_frame = function() {
+            arrow_tbl = self$to_arrow_table()
+            return(as.data.frame(as.data.frame(arrow_tbl))) # for some reason as.data.frame on arrow table returns a tibble, not a dataframe
```
chipkent, line 40:
```suggestion
            return(as.data.frame(as.data.frame(arrow_tbl))) #TODO: for some reason as.data.frame on arrow table returns a tibble, not a dataframe
```

R/rdeephaven/lib/build-cpp.sh
chipkent: Both of these problems need to be addressed in v2.

R/rdeephaven/man/Client.Rd
```diff
@@ -0,0 +1,92 @@
...
+a connection to a running Deephaven server and facilitating basic server requests. This is
+the primary interface for interacting with the Deephaven server from R.
+}
+\examples{
+
+client_options <- ClientOptions$new()
+client_options$set_default_authentication()
+
+# connect to the Deephaven server running on "localhost:10000" with anonymous 'default' authentication
+client <- Client$new(target="localhost:10000", client_options=client_options)
```
chipkent, line 17: Regenerate after my changes above.

R/rdeephaven/man/rdeephaven-package.Rd
```diff
@@ -0,0 +1,34 @@
+\name{rdeephaven-package}
```
chipkent, line 1: Docs need to be completed in v2.

R/rdeephaven/prototyping/new-api.R
```diff
@@ -0,0 +1,27 @@
+Rcpp::compileAttributes()
+roxygen2::roxygenise()
+install.packages("/home/user/rdeephaven", repos=NULL, type="source")
+
+# load deephaven library
+library(rdeephaven)
+
+# connect to server using Client object
+client <- DeephavenClient$new(target="192.168.1.241:10000")
```
chipkent, line 9: Does this work for you?  If so, it is more portable to anyone running a server on the same machine.  We can address later.
```suggestion
client <- DeephavenClient$new(target="localhost:10000")
```

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,233 @@
...
+class ClientOptionsWrapper;
+class ClientWrapper;
+
+// ######################### DH WRAPPERS #########################
+
+class TableHandleWrapper {
+public:
+    TableHandleWrapper(deephaven::client::TableHandle ref_table) : internal_tbl_hdl(std::move(ref_table)) {};
+
+    // DEEPHAVEN QUERY METHODS WILL GO HERE
```
chipkent, line 25:
```suggestion
    //TODO: DEEPHAVEN QUERY METHODS WILL GO HERE
```

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,233 @@
...
+    /**
+     * Creates and returns a pointer to an ArrowArrayStream C struct containing the data from the table referenced by internal_tbl_hdl.
+     * Intended to be used for creating an Arrow RecordBatchReader in R via RecordBatchReader$import_from_c(ptr).
+    */
+    SEXP getArrowArrayStreamPtr() {
+
+        std::shared_ptr<arrow::flight::FlightStreamReader> fsr = internal_tbl_hdl.getFlightStreamReader();
+
+        std::vector<std::shared_ptr<arrow::RecordBatch>> empty_record_batches;
+        DEEPHAVEN_EXPR_MSG(fsr->ReadAll(&empty_record_batches)); // need to add OK or throw
```
chipkent, line 45:
```suggestion
        DEEPHAVEN_EXPR_MSG(fsr->ReadAll(&empty_record_batches)); //TODO: need to add OK or throw
```

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+
+
+class ClientOptionsWrapper {
+public:
+
+    ClientOptionsWrapper() {
+        internal_options = new deephaven::client::ClientOptions();
+        bool authentication_set = false;
+        bool session_set = false;
```
chipkent, line 67: Is there a reason this isn't just left to the non-wrapper code?

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,233 @@
...
+    };
+
+    /**
+     * Allocates memory for an ArrowArrayStream C struct and returns a pointer to the new chunk of memory.
+     * Intended to be used to get a pointer to pass to Arrow's R library RecordBatchReader$export_to_c(ptr).
+    */
+    SEXP newArrowArrayStreamPtr() {
+        ArrowArrayStream* stream_ptr = new ArrowArrayStream();
+        return Rcpp::XPtr<ArrowArrayStream>(stream_ptr, true);
+    };
```
chipkent, line 144: As I noted elsewhere, this method has nothing to do with the client, so it logically does not belong here.  Can be fixed v2, or we can leave it like this.

R/rdeephaven/src/client.cpp
```diff
@@ -0,0 +1,233 @@
...
+// the current implementation of passing authentication args to C++ client is terrible and needs to be redone. Only this could make Rcpp happy in a days work
+
+/**
+ * Factory method for creating a new ClientWrapper, which is responsible for maintaining a connection to the client.
+ * @param target URL that the server is running on.
+ * @param client_options A ClientOptionsWrapper containing the server connection information. See deephaven::client::ClientOptions for more information.
+ */
+ClientWrapper* newClientWrapper(const std::string &target, const ClientOptionsWrapper &client_options) {
+    return new ClientWrapper(deephaven::client::Client::connect(target, *client_options.internal_options));
+};
```
chipkent, line 198: To simplify things, in v2, we could potentially get rid of the factory methods.  They were originally introduced to control the API the user sees.  Now there is the S6 classes to do that.  Overall, this is now probably extra complexity for no gain.

chipkent (changes requested):

R/rdeephaven/src/client.cpp
alexpeters1208: Nope. I was going to use them to check if ClientOptions had already been set, but I should do that in R anyway.

R/rdeephaven/src/client.cpp
alexpeters1208: Introduced for API reasons, but kept around because Rcpp _hates_ `const std::string&` types in the ".constructor", but is happy with those types in ".factory".

R/rdeephaven/prototyping/new-api.R
alexpeters1208: I don't really plan on this file getting shipped. Should I get rid of it now?

R/rdeephaven/src/client.cpp
chipkent: Why can't you just leave it to C++?

R/rdeephaven/src/client.cpp
chipkent: ok

R/rdeephaven/src/client.cpp
alexpeters1208: I thought better to catch earlier if possible?

R/rdeephaven/R/LICENSE
chipkent, line 1: empty

chipkent: Also, missing the file in the c++ dir.

chipkent (changes requested):

R/rdeephaven/lib/LICENSE
chipkent, line 1: Delete in v2.

chipkent (dismissed):

chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2725
Conceptual: https://github.com/deephaven/deephaven.io/issues/2726
Reference: https://github.com/deephaven/deephaven.io/issues/2724
Blog: https://github.com/deephaven/deephaven.io/issues/2727

PR #3884, opened 2023-05-30
Add meta into TableOperations

### LINKED ISSUE #3648: Move getMeta into TableOperations

devinrsmith:
Fixes #3648

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/partitioned/PartitionedTableProxyImpl.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/partitioned/PartitionedTableProxyImpl.java
@@ -586,5 +586,10 @@ class PartitionedTableProxyImpl extends LivenessArtifact implements PartitionedT
         return basicTransform(ct -> ct.dropColumns(columnNames));
     }
 
+    @Override
+    public PartitionedTable.Proxy meta() {
+        return basicTransform(Table::meta);
+    }
+
     // endregion TableOperations Implementation
 }
--- a/engine/api/src/main/java/io/deephaven/engine/table/Table.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/Table.java
@@ -42,15 +42,6 @@ public interface Table extends
     @ConcurrentMethod
     TableDefinition getDefinition();
 
-    /**
-     * Provides column metadata in Table form. Convenience method, behaves exactly the same as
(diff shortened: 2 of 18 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/partitioned/PartitionedTableProxyImpl.java
```diff
@@ -586,5 +586,10 @@ public PartitionedTable.Proxy dropColumns(String... columnNames) {
         return basicTransform(ct -> ct.dropColumns(columnNames));
     }
 
+    @Override
+    public PartitionedTable.Proxy meta() {
+        return basicTransform(Table::meta);
+    }
```
devinrsmith, line 592: This return type is a bit weird; that said, it's *consistent*. I'm not sure it's worth trying to re-work the return type in a generic way just to make this specific implementation "cleaner".

devinrsmith:
Here's an example from java client:

```
$ ./java-client/flight-examples/build/install/java-client-flight-examples/bin/get-meta-table --variable t

Schema<Name: Utf8, DataType: Utf8, ColumnType: Utf8, IsPartitioning: Bool, IsGrouping: Bool>
Name    DataType        ColumnType      IsPartitioning  IsGrouping
Timestamp       io.deephaven.time.DateTime      Normal  false   false

PT0.191009432S duration
```

py/server/deephaven/table.py
```diff
@@ -557,7 +557,7 @@ def columns(self) -> List[Column]:
     @property
     def meta_table(self) -> Table:
```
chipkent, line 558: @jmao-denver any reason to consider renaming the python method to `meta` or not?

rcaudy (commented):
No complaints, just philosophical concerns about return types and t-ops vs. concrete tables.

devinrsmith:
A simple rename was done as part of https://github.com/deephaven/deephaven-core/pull/3910

The tricky thing is the partition table; should we _allow_ it there? Either it's funky (all the constituent tables create a meta table, which is wasteful), or we need to do type trickery to make it cleaner. Another alternative is to not add the meta() op to TableOPerations, but instead have it in the TableCreator interface. This seems potentially promising, but needs further investigation by me.

PR #3898, opened 2023-05-31
Correct Javadoc warnings, prevent future ones

### LINKED ISSUE #1513: Remove non-LTS JDK 15 as part of javadocs process

niloc132:
Addresses several classes of issues, some in more than one way:
 * Ensure all javadoc runs use UTF-8 encoding, and remove non-ASCII chars which are intended to be interpreted as code
 * HTML-encode characters that must be encoded in javadoc, and correct wrongly-encoded chars.
 * Don't reference JVM internals that can't be linked, instead refer to them as `@code`.
 * Fixed mismatched html tags, broken attrs.
 * Corrected/removed classes/members that don't exist or can't be referenced in that way.

Fixes #1513


niloc132:
This is left as a draft for the moment to get some feedback from a few specific authors/owners. Only combined-javadoc will fail if there are warnings, allowing potentially "invalid" javadoc references that can be resolved by producing javadoc all at once.

DHProcess/src/main/java/io/deephaven/process/SystemCpuOshi.java
```diff
@@ -58,8 +58,8 @@ public abstract class SystemCpuOshi implements PropertySet {
     /**
      * Gets the Processor ID. This is a hexidecimal string representing an 8-byte value, normally obtained using the
      * CPUID opcode with the EAX register set to 1. The first four bytes are the resulting contents of the EAX register,
-     * which is the Processor signature, represented in human-readable form by {@link #getIdentifier()} . The remaining
-     * four bytes are the contents of the EDX register, containing feature flags.
+     * which is the Processor signature, represented in human-readable form by getIdentifier(). The remaining four bytes
```
niloc132, line 61: @devinrsmith can you offer another suggestion here? You're the original author of this code, and I can't find anywhere in the history of this class (even pre-DHC) where this method existed.

extensions/jdbc/src/main/java/io/deephaven/jdbc/JdbcToTableAdapter.java
```diff
@@ -37,7 +37,6 @@
 /**
  * The JdbcToTableAdapter class provides a simple interface to convert a Java Database Connectivity (JDBC)
  * {@link ResultSet} to a Deephaven {@link Table}.
- * <p/>
```
niloc132, line 40: this shouldnt be removed (but also "shouldnt" be self closing)

Util/src/main/java/io/deephaven/util/audit/AuditEventLoggerBasic.java
```diff
@@ -4,9 +4,10 @@
 package io.deephaven.util.audit;
 
 /**
- * An simple interface that hides the context necessary for creating and logging {@link AuditEvent}s. Useful so that
- * callers are only responsible for the core parts of the event, namely {@link AuditEvent#getEvent()} and
- * {@link AuditEvent#getDetails()}.
+ * TODO is this necessary to keep? The AuditEvent class doesn't exist and this can't be used anywhere. A simple
```
niloc132, line 7: Not sure who to ask about this, @rcaudy?

ModelFarm/src/main/java/io/deephaven/modelfarm/ModelFarmOnDemand.java
```diff
@@ -20,11 +20,13 @@
  * A ModelFarm implementation for evaluating a model upon request, retrieving a snapshot of data for all keys under a
  * single {@link UpdateGraphProcessor} lock.
  *
- * @param <KEYTYPE> The type of the keys (e.g. {@link io.deephaven.modelfarm.fitterfarm.FitScope}).
+ * TODO None of the following classes exist any more - should we remove this too?
```
niloc132, line 23: @chipkent I believe this is best directed at you, along with the other ModelFarm classes. Should these classes still exist, or should we just remove these references?

DHProcess/src/main/java/io/deephaven/process/SystemCpuOshi.java
devinrsmith: Looks like bad copy-paste from `oshi.hardware.CentralProcessor.ProcessorIdentifier#getProcessorID`

devinrsmith: I would remove it - oshi dependency is implementation not api atm.

Util/src/main/java/io/deephaven/util/audit/AuditEventLoggerBasic.java
rcaudy: Delete it.

ModelFarm/src/main/java/io/deephaven/modelfarm/RDMModelFarm.java
```diff
@@ -23,11 +23,11 @@
  * table}. Each row of this table should contain all of the data necessary to populate an instance of {@code DATATYPE},
  * which will then be passed to the {@link ModelFarmBase#model model}.
  *
- * @param <KEYTYPE> The type of the keys (e.g. {@link io.deephaven.modelfarm.fitterfarm.FitScope}).
+ * @param <KEYTYPE> The type of the keys (e.g. {@code io.deephaven.modelfarm.fitterfarm.FitScope}).
```
rcaudy, line 26: Remember to handle this the same was as the `ModelFarmOnDemand` decision.

buildSrc/src/main/groovy/io.deephaven.javadoc-conventions.gradle
```diff
@@ -0,0 +1,18 @@
...
+tasks.withType(Javadoc).configureEach {
+    javadocTool = javaToolchains.javadocToolFor{javadocSpec ->
+        // Javadoc version >=11 is needed for search
+        // Javadoc version >12 is needed to avoid javadoc bugs in linking with modules
+        javadocSpec.languageVersion = JavaLanguageVersion.of(17)
+    }
+
+    options.encoding = 'UTF-8'
+
+    options.links = ['https://docs.oracle.com/en/java/javase/11/docs/api/']
```
rcaudy, line 12: Why 11?

buildSrc/src/main/groovy/io.deephaven.javadoc-conventions.gradle
```diff
@@ -0,0 +1,18 @@
...
+    options.addBooleanOption('Xdoclint:none', true)
```
rcaudy, line 13: Dropping `-quiet` was intentional?

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/FillUnordered.java
```diff
@@ -19,7 +19,7 @@
      * 
      * <pre>
      * destination.setSize(keys.size());
-     * for (int ii = 0; ii < keys.size(); ++ii) {
+     * for (int ii = 0; ii &lt; keys.size(); ++ii) {
```
rcaudy, line 22: Is this needed inside of a `<pre>`?

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/emstd/BigDecimalEmStdOperator.java
```diff
@@ -21,7 +21,7 @@
  * Compute an exponential moving standard deviation for a BigDecimal column source. The output is expressed as a
  * BigDecimal value and is computed using the following formula:
  *
- * variance = alpha * (prevVariance + (1 − alpha) * (x − prevEma)^2)
+ * variance = alpha * (prevVariance + (1 - alpha) * (x - prevEma)^2)
```
rcaudy, line 24: I can't tell what changed here. Was it some special char?

engine/table/src/main/java/io/deephaven/libs/GroovyStaticImports.java
```diff
@@ -189,7 +189,7 @@ public class GroovyStaticImports {
     public static  double avg( io.deephaven.vector.LongVector values ) {return Numeric.avg( values );}
     /** @see io.deephaven.function.Numeric#avg(io.deephaven.vector.ShortVector) */
     public static  double avg( io.deephaven.vector.ShortVector values ) {return Numeric.avg( values );}
-    /** @see io.deephaven.function.BinSearch#binSearchIndex(T[],T,io.deephaven.function.BinSearchAlgo) */
+    /** @see io.deephaven.function.BinSearch#binSearchIndex(java.lang.Comparable[],java.lang.Comparable,io.deephaven.function.BinSearchAlgo) */
```
rcaudy, line 192: This file is code generated. Did you fix the generator?

rcaudy: Never mind, I see that you have.

buildSrc/src/main/groovy/io.deephaven.javadoc-conventions.gradle
devinrsmith: I'm guessing b/c that's the language level we target.

Generators/src/main/java/io/deephaven/libs/GroovyStaticImportGenerator.java
devinrsmith, line 1: This was the only file that stood out as having some sort of new code - I'm assuming you re-ran the generation?

devinrsmith (dismissed):
buildSrc changes look good. I did go fast through the actual javadoc changes.  I call this a big win, even if we feel like there are places we might want to follow-up on to make further improvements. I would not hold up this PR for javadoc content-type changes.

Base/src/main/java/io/deephaven/base/CompareUtils.java
```diff
@@ -159,7 +159,7 @@ public static boolean doubleNotEquals2(double x, double y, double tolerance) {
     }
 
     /**
-     * You can't do double1 > double2 because floating point numbers are not exact values. Just make sure that x-y is
+     * You can't do double1 &gt; double2 because floating point numbers are not exact values. Just make sure that x-y is
```
chipkent, line 162: It seems like this should be:
```suggestion
     * You can't do {@code double1 > double2} because floating point numbers are not exact values. Just make sure that x-y is
```
Same comment many places that `&gt` or `&lt` are used.  It seems like these symbols just indicate that we are in "code".  In my quick look, every time I see these changes, we are in code.

Base/src/main/java/io/deephaven/base/string/EncodingInfo.java
```diff
@@ -56,7 +56,8 @@ public String getEncodingName() {
     }
 
     /**
-     * Can this encoding info's charset be encoded or decoded by simple linear assignment of char->byte or byte->char.
+     * Can this encoding info's charset be encoded or decoded by simple linear assignment of char-&gt;byte or
+     * byte-&gt;char.
```
chipkent, line 60: Should these be using HTML arrows instead (e.g. `&larr;` or `&rarr`)?

Configuration/src/main/java/io/deephaven/configuration/PropertyFile.java
```diff
@@ -39,7 +39,7 @@ public PropertyFile() {
      *
      * @param filename name of file to load properties from
      * @param log Logger for error/warning messages, or null if not logging
-     * @param fatal If the file can't be read and fatal is true -> throw exception
+     * @param fatal If the file can't be read and fatal is true -&gt; throw exception
```
chipkent, line 42: arrow?

ModelFarm/src/main/java/io/deephaven/modelfarm/ModelFarmOnDemand.java
chipkent: For these, just go very generic.  This is very core code that some user is going to want, but we left the finance specific part in DHE.  This is why the refs are gone.

```
@param <KEYTYPE> The type of keys.
@param <DATATYPE> The type of data.
@param <ROWDATAMANAGERTYPE> The type of RowDataManager.
```

ModelFarm/src/main/java/io/deephaven/modelfarm/RDMModelFarm.java
chipkent: See above

engine/table/src/main/java/io/deephaven/engine/util/string/StringUtils.java
```diff
@@ -122,7 +122,7 @@ public String getKey(@NotNull final VALUE_TYPE value) {
      * Generic accessor for the singleton StringKey instance.
      * 
      * @param <VALUE_TYPE> The type of the value for this key implementation
-     * @return A String->StringKeyedObject key representation instance.
+     * @return A String-&gt;StringKeyedObject key representation instance.
```
chipkent, line 125: more arrows?

chipkent (changes requested):

Generators/src/main/java/io/deephaven/libs/GroovyStaticImportGenerator.java
niloc132: Codegen was updated, else the build would fail when the assert check ran.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/emstd/BigDecimalEmStdOperator.java
niloc132: Yes, an em dash to an en dash. The first isn't in ascii, the second is, and prevents copying this code to run it, or irritating cases where the machine doing the building doesnt do utf-8.

Base/src/main/java/io/deephaven/base/CompareUtils.java
niloc132: Unless I'm very much mistaken (will validate by changing the code to confirm), html is allowed inside of `@code` tags, so this will not work. It might separately be a good idea to wrap in `@code`, but I'm not reading content beyond "is it less wrong than it was before".

Base/src/main/java/io/deephaven/base/string/EncodingInfo.java
niloc132: That's really up to the author - I'm making it not-wrong, not necessarily reading the author's mind for what they originally intended.

ModelFarm/src/main/java/io/deephaven/modelfarm/ModelFarmOnDemand.java
chipkent: Use the doc strings in my comment above.  If we have a `RowDataManager` class, add a link.  If not, just change it to "row data manager".

buildSrc/src/main/groovy/io.deephaven.javadoc-conventions.gradle
niloc132: Yes, it did nothing, it was a workaround for a gradle javadoc limitation. Boolean is probably more accurate or at least less confusing, so no one thinks that `-quiet` was intended to be added as a real flag.

buildSrc/src/main/groovy/io.deephaven.javadoc-conventions.gradle
niloc132: Yes, just centralized the javadoc configuration to one place, so everything is built the same way, rather than 3-ish different setups.

Base/src/main/java/io/deephaven/base/CompareUtils.java
chipkent: I agree on not holding up for better.  Javadocs need real work at some point.

Util/src/main/java/io/deephaven/util/audit/AuditEventLoggerBasic.java
niloc132: Delete AuditEventLoggerBasic? or just the references to AuditEvent?

Base/src/main/java/io/deephaven/base/string/EncodingInfo.java
chipkent: I would not hold up for any of these arrow comments.  These can be handled later.

Base/src/main/java/io/deephaven/base/CompareUtils.java
niloc132: I am mistaken! The code tags do let this work, but I'm going to leave it as-is for now, we can revert/fix later once we're locked in with "this will build correctly" and have time for more specific review.

rcaudy (approved):

chipkent (approved):

PR #3909, opened 2023-06-01
Add and plumb AjRajTablesRequest

devinrsmith:


rcaudy (commented):
Proto looks good to me. Maybe we can just delete the old message?

devinrsmith:
> Proto looks good to me. Maybe we can just delete the old message?

Yeah - I mean, we can do an "in-place" breaking change; but then we'd need to update python, C++, and go clients in lockstep (or, be "ok" with them being broken).

rcaudy (dismissed):
I'm OK with this as-is, but I'd like @niloc132 to look, and I think we want to consider whether we're updating the clients in this PR or subsequent ones.

niloc132 (dismissed):
table.proto approved, with minor protest for not updating js client

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -694,6 +710,15 @@ message AsOfJoinTablesRequest {
   MatchRule as_of_match_rule = 7;
 }
 
+message AjRajTablesRequest {
+  Ticket result_id = 1;
+  TableReference left_id = 2;
+  TableReference right_id = 3;
+  repeated string exact_match_columns = 4;
+  string as_of_column = 5;
```
niloc132, line 718: as it seems this is a boolean column expr, either the name or docs should make this clear. i assume this is validated to ensure it is safe for arbitrary clients to pass it in (and that both of the other "column" fields are only names and not exprs)?

devinrsmith: Added docs.

jmao-denver (approved):
LGTM

rcaudy (approved):

PR #3928, opened 2023-06-05
Fix RowRedirectionLockFreeTest: Prevent Succeeding If Thread Fails

nbauernfeind:
This fixes one issue @devinrsmith pointed out in slack about the nightly failures.

Here is a nightly with this fix: https://github.com/nbauernfeind/deephaven-core/actions/runs/5178849528

### DIFF
--- a/engine/table/src/test/java/io/deephaven/engine/table/impl/util/RowRedirectionLockFreeTest.java
+++ b/engine/table/src/test/java/io/deephaven/engine/table/impl/util/RowRedirectionLockFreeTest.java
@@ -5,23 +5,32 @@ package io.deephaven.engine.table.impl.util;
 
 import io.deephaven.engine.context.ExecutionContext;
 import io.deephaven.engine.testutil.ControlledUpdateGraph;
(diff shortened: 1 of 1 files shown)


devinrsmith:
This is an example of the general issue I raised in https://github.com/deephaven/deephaven-core/issues/3930

> the test will only fail if it has specific logic to handle that scenario

I think improving this specific test in this way is good, but hoping we have a more robust solution in the future.

engine/table/src/test/java/io/deephaven/engine/table/impl/util/RowRedirectionLockFreeTest.java
```diff
@@ -81,6 +85,7 @@ public final void run() {
                 doOneIteration();
                 ++numIterations;
             }
+            cleanExit = true;
```
rcaudy, line 88: Shouldn't we just catch errors and fail the test from the threads?

rcaudy (approved):

PR #3929, opened 2023-06-05
Improve CI uploaded artifacts

devinrsmith:
The binary logs may provide useful debugging information when CI appears to fail for no other apparent reason.

* Upload binary test-results as artifact for check-ci and nightly-check-ci
* Combines test-results and reports for check-ci (matches nightly-check-ci behavior)
* Adds core as artifact for check-ci-jvm-err (matches nightly-check-ci behavior)

### DIFF
--- a/.github/workflows/check-ci.yml
+++ b/.github/workflows/check-ci.yml
@@ -51,19 +51,14 @@ jobs:
           name: check-ci-results
           path: |
             **/build/test-results/**
-            !**/build/test-results/**/binary/**
-
-      - name: Upload Test Reports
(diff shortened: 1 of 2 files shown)


devinrsmith:
Should be able to verify that nightly run fails and produces binary logs in artifact, https://github.com/devinrsmith/deephaven-core/actions/runs/5179041363. Will check back here when that is done.

niloc132 (approved):

rcaudy (approved):

devinrsmith:
Verified binary artifacts are properly uploaded

PR #3932, opened 2023-06-05
Fix UpdateGraph not always available to performance trackers

### LINKED ISSUE #3931: Performance Trackers may not be able to get a UpdateGraph instance
If the EngineTableLoggers.Factory has been replaced with something that does not support getting results as QueryTable, like we do in Enterprise, the drivers will crash the worker when they try to initialize with an NPE.

**Steps to reproduce**
The simplest reproduction uses an Enterprise worker.  Create one it will start,  then crash one the trackers try to start

**Expected results**
The worker does not crash

**Actual results**
The worker crashes

**Versions**
 - Deephaven: 0.25.0

abaranec:
This fixes #3931.

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/perf/UpdatePerformanceTracker.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/perf/UpdatePerformanceTracker.java
@@ -4,12 +4,18 @@
 package io.deephaven.engine.table.impl.perf;
 
 import io.deephaven.configuration.Configuration;
-import io.deephaven.engine.table.*;
+import io.deephaven.engine.context.ExecutionContext;
+import io.deephaven.engine.table.ShiftObliviousListener;
+import io.deephaven.engine.table.Table;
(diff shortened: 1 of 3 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/perf/UpdatePerformanceTracker.java
```diff
@@ -4,12 +4,18 @@
 package io.deephaven.engine.table.impl.perf;
 
 import io.deephaven.configuration.Configuration;
-import io.deephaven.engine.table.*;
+import io.deephaven.engine.context.ExecutionContext;
+import io.deephaven.engine.table.ShiftObliviousListener;
```
nbauernfeind, line 8: I think some of these might be unused.

nbauernfeind (changes requested):

rcaudy (commented):
LGTM, other than @nbauernfeind 's feedback. Deferring to him.

PR #3933, opened 2023-06-05
Fix TestTableTools: wrong junit version

nbauernfeind:
I believe this is the last test needed to get nightlies passing again.


rcaudy (approved):

rcaudy:
LGTM, but it might be better to just change to JUnit4.

PR #3940, opened 2023-06-05
Add blink time_table support

devinrsmith:


py/server/deephaven/table_factory.py
```diff
@@ -47,13 +47,14 @@ def empty_table(size: int) -> Table:
...
-def time_table(period: Union[str, int], start_time: str = None) -> Table:
+def time_table(period: Union[str, int], start_time: str = None, blink_table: bool = None) -> Table:
     """Creates a table that adds a new row on a regular interval.
 
     Args:
         period (Union[str, int]): time interval between new row additions, can be expressed as an integer in
             nanoseconds or a time interval string, e.g. "PT00:00:00.001"
-        start_time (str): start time for adding new rows
+        start_time (str, optional): start time for adding new rows, defaults to None
+        blink_table (bool, optional): if the time table should be a blink table, defaults to None
```
jmao-denver, line 57: 1. should we improve the start_time docstring, explicitly state what the default value 'None' means?
2. since blink_table is a bool, it might be better to default it to False, a bit more explicit and clear

py/server/tests/test_table_factory.py
```diff
@@ -64,6 +64,11 @@ def test_time_table(self):
         self.assertTrue(t.is_refreshing)
         self.assertEqual("2021-11-06T13:21:00.000000000 ET", format_datetime(t.j_table.getColumnSource("Timestamp").get(0), time_zone('ET')))
 
+    def test_time_table_blink(self):
+        t = time_table("PT1s", blink_table=True)
+        self.assertEqual(1, len(t.columns))
+        self.assertTrue(_JBlinkTableTools.isBlink(t.j_table))
+
```
jmao-denver, line 71: Would it be better to add a property 'is_blinking' to Table?

devinrsmith: I think we'd want to add `io.deephaven.engine.table.Table#isBlink` before we added it to the python side. Note: `io.deephaven.engine.table.impl.BaseTable#isBlink` exists, but is scoped to the implementations, not the interface. Should we elevate to engine Table interface @rcaudy?

devinrsmith: Regardless, I think that would be a new issue / PR.

rcaudy: Let's defer adding anything to the table interface. Users who really want to know can look at attributes for now.

rcaudy (commented):
LGTM, deferring to Jianfeng and/or Chip for Python approval.

jmao-denver (approved):
LGTM

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2757
Conceptual: https://github.com/deephaven/deephaven.io/issues/2755
Reference: https://github.com/deephaven/deephaven.io/issues/2754
Blog: https://github.com/deephaven/deephaven.io/issues/2756

PR #3942, opened 2023-06-06
Several JS API totals table fixes

### LINKED ISSUE #3921: JS API doesn't name Count columns correctly when creating totals tables
To reproduce, create a table with a string column
```python
t2 = empty_table(100).update('T = i / 10.0').update('Str=`` + T')
```
Then using the web UI, aggregate the table using Count and Unique

Expected: table contents render correctly, no browser console errors

Actual: browser console error, not all columns or totals are rendered

```
Uncaught TypeError: e.slice is not a function
    <anonymous> http://localhost:10000/ide/assets/AppRoot-a9081354.js:20
    m http://localhost:10000/ide/assets/vendor-aaea4a59.js:41
    tokensForCell http://localhost:10000/ide/assets/AppRoot-a9081354.js:20
    drawCellContent http://localhost:10000/ide/assets/AppRoot-a9081354.js:18
    drawCellContent http://localhost:10000/ide/assets/AppRoot-a9081354.js:20
    drawFloatingRows http://localhost:10000/ide/assets/AppRoot-a9081354.js:18
    drawGrid http://localhost:10000/ide/assets/AppRoot-a9081354.js:18
```

niloc132:
Updates totals table samples to work in DHC.
Permits `Distinct` aggregation on any column type instead of none.
Correctly names `Count` columns to match DHE's JS API.
Correctly drops unused columns like `Count` before returning table to the UI.

Fixes #3921

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
@@ -19,6 +19,7 @@ import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.table_pb.Aggr
 import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.table_pb.AsOfJoinTablesRequest;
 import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.table_pb.BatchTableRequest;
 import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.table_pb.CrossJoinTablesRequest;
+import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.table_pb.DropColumnsRequest;
 import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.table_pb.ExactJoinTablesRequest;
 import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.table_pb.ExportedTableCreationResponse;
 import io.deephaven.javascript.proto.dhinternal.io.deephaven.proto.table_pb.Literal;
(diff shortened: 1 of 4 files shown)


niloc132:
Tested by created a totals table with various types in the web UI, with one aggregation at a time, then all at once

web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
```diff
@@ -666,21 +667,32 @@ private Promise<JsTotalsTable> fetchTotals(Object config, JsProvider<ClientTable
...
+            if (updateViewExprs.length != 0) {
                 SelectOrUpdateRequest columnExpr = new SelectOrUpdateRequest();
                 columnExpr.setResultId(requestMessage.getResultId());
                 requestMessage.setResultId();
                 columnExpr.setColumnSpecsList(updateViewExprs);
-                TableReference prev = new TableReference();
-                prev.setBatchOffset(0);
-                columnExpr.setSourceId(prev);
+                columnExpr.setSourceId(new TableReference());
+                columnExpr.getSourceId().setBatchOffset(0);
```
mofojed, line 676: Was this causing a problem previous? Just seems odd to rewrite it this way, would expect to set the batchOffset before calling `setSourceId`.

mofojed (approved):

web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
niloc132: No, this is just a line shorter this way. The JS api for pb isn't immutable, so we can do it this way, and while it does offer a builder pattern (setter methods return `this`), the TS that is generated only offers plain setters, so they can't be chained (and the java is generated from the TS).

PR #3946, opened 2023-06-06
Add additional opts for pct, media, unique, ... in the Python client

jmao-denver:
To bring the same functionality of the server API to the client.


py/client/pydeephaven/agg.py
```diff
@@ -300,17 +313,52 @@ def std(cols: Union[str, List[str]] = None) -> Aggregation:
...
+            elif np.iinfo(np.int32).min <= non_unique_sentinel <= np.iinfo(np.int32).max:
+                agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(int_value=non_unique_sentinel)
+            else:
+                agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(long_value=non_unique_sentinel)
+        elif isinstance(non_unique_sentinel, str):
+            if len(non_unique_sentinel) == 1:
+                agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(
+                    char_value=ord(non_unique_sentinel))
+            else:
+                agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(string_value=non_unique_sentinel)
```
chipkent, line 353: I have a lot of concerns about this.  We are mapping to a type based on a value.  I'm always very concerned about code like this.

py/client/tests/test_table.py
```diff
@@ -294,6 +294,35 @@ def test_meta_table(self):
         test_table = self.session.import_table(pa_table).drop_columns(["e"])
         self.assertEqual(len(test_table.schema), len(test_table.meta_table.to_arrow()))
 
+    def test_agg_with_options(self):
+        pa_table = csv.read_csv(self.csv_file)
+        test_table = self.session.import_table(pa_table).update(["b = a % 10 > 5 ? null : b", "c = c % 10"])
+        aggs = [
+            median(cols=["ma = a", "mb = b"], average_evenly_divided=False),
+            pct(0.20, cols=["pa = a", "pb = b"], average_evenly_divided=True),
+            unique(cols=["ua = a", "ub = b"], include_nulls=True, non_unique_sentinel=-1),
```
chipkent, line 303: The test cases do not cover mismatched sentinel types or all of the weirdness that can come with inferring the type from the value.  Also, given the number of branches in the sentinel code, our coverage is very low.

chipkent (changes requested):

py/client/pydeephaven/agg.py
jmao-denver: @devinrsmith I was struggling with this, and I have to agree with @chipkent that it is ugly code. Without knowing the gRCP server implementation much, I am wondering if this type of mapping can be made easier for Python somehow?

devinrsmith: So, the python client can choose to go about this however it wants; at the RPC level though, you _will_ need some way to distinguish between all of these `<x>_value` cases though. Fundamentally, the java client does the same thing:

https://github.com/deephaven/deephaven-core/blob/v0.26.1/java-client/session/src/main/java/io/deephaven/client/impl/AggSpecBuilder.java#L206-L257

https://github.com/deephaven/deephaven-core/blob/v0.26.1/proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto#L868-L885

It uses a wrapper object though (as opposed to a plain `Object`/`object`): https://github.com/deephaven/deephaven-core/blob/v0.26.1/table-api/src/main/java/io/deephaven/api/object/UnionObject.java. The python client could do something similar:

```python
my_spec_1 = unique(..., non_unique_sentinal=int_(32))
my_spec_2 = unique(..., non_unique_sentinal=short_(32))
```

py/client/pydeephaven/agg.py
```diff
@@ -300,17 +313,55 @@ def std(cols: Union[str, List[str]] = None) -> Aggregation:
...
+            agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(long_value=non_unique_sentinel)
+        elif isinstance(non_unique_sentinel, np.float32):
+            agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(float_value=non_unique_sentinel)
+        elif isinstance(non_unique_sentinel, np.float64):
+            agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(double_value=non_unique_sentinel)
+        elif isinstance(non_unique_sentinel, np.uint16):
+            agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(char_value=non_unique_sentinel)
+        elif isinstance(non_unique_sentinel, str):
+            agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(string_value=non_unique_sentinel)
+        elif isinstance(non_unique_sentinel, bool):
```
chipkent, line 355: does `np.bool_` work?

py/client/pydeephaven/agg.py
```diff
@@ -300,17 +313,55 @@ def std(cols: Union[str, List[str]] = None) -> Aggregation:
...
+            agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(bool_value=non_unique_sentinel)
+        else:
```
chipkent, line 357: should we support `int` as `long_value` and `float` as `double_value`?  I lean to maybe, since they would be common use cases.  I can see that there could be reasons not to.

py/client/pydeephaven/agg.py
```diff
@@ -300,17 +313,55 @@ def std(cols: Union[str, List[str]] = None) -> Aggregation:
...
+        TypeError
 
     Returns:
         an aggregation
     """
-    agg_spec = _GrpcAggSpec(unique=_GrpcAggSpec.AggSpecUnique(include_nulls=False, non_unique_sentinel=None))
+    if non_unique_sentinel is not None:
+        if isinstance(non_unique_sentinel, np.byte):
+            agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(byte_value=non_unique_sentinel)
+        elif isinstance(non_unique_sentinel, np.short):
```
chipkent, line 341: is `np.short` identical to `np.int16`?

py/client/pydeephaven/agg.py
```diff
@@ -300,17 +313,55 @@ def std(cols: Union[str, List[str]] = None) -> Aggregation:
...
+
+    Raises:
+        TypeError
 
     Returns:
         an aggregation
     """
-    agg_spec = _GrpcAggSpec(unique=_GrpcAggSpec.AggSpecUnique(include_nulls=False, non_unique_sentinel=None))
+    if non_unique_sentinel is not None:
+        if isinstance(non_unique_sentinel, np.byte):
```
chipkent, line 339: is `np.byte` identical to `np.int8`?

py/client/pydeephaven/agg.py
```diff
@@ -300,17 +313,55 @@ def std(cols: Union[str, List[str]] = None) -> Aggregation:
...
+        elif isinstance(non_unique_sentinel, np.int32):
+            agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(int_value=non_unique_sentinel)
+        elif isinstance(non_unique_sentinel, np.int64):
+            agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(long_value=non_unique_sentinel)
+        elif isinstance(non_unique_sentinel, np.float32):
+            agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(float_value=non_unique_sentinel)
+        elif isinstance(non_unique_sentinel, np.float64):
+            agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(double_value=non_unique_sentinel)
+        elif isinstance(non_unique_sentinel, np.uint16):
+            agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(char_value=non_unique_sentinel)
```
chipkent, line 352: What is the thinking on supporting `uint16` as `char` instead of going to `int` or just not supporting it?  Similarly, `uint32` could be supported as `long`, but that is a size change that we may want to avoid.  Is it just the closest thing to a char in np?

py/client/pydeephaven/agg.py
```diff
@@ -300,17 +313,55 @@ def std(cols: Union[str, List[str]] = None) -> Aggregation:
...
+        elif isinstance(non_unique_sentinel, str):
+            agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(string_value=non_unique_sentinel)
+        elif isinstance(non_unique_sentinel, bool):
+            agg_spec_non_unique_sentinel = _GrpcAggSpec.AggSpecNonUniqueSentinel(bool_value=non_unique_sentinel)
+        else:
+            raise TypeError(f"invalid non-unique-sentinel value type {type(non_unique_sentinel)}")
```
chipkent, line 358: Should the error list the valid types?

py/client/pydeephaven/agg.py
```diff
@@ -300,17 +313,55 @@ def std(cols: Union[str, List[str]] = None) -> Aggregation:
...
 
     Args:
         cols (Union[str, List[str]]): the column(s) to aggregate on, can be renaming expressions, i.e. "new_col = col";
             default is None, only valid when used in Table agg_all_by operation
+        include_nulls (bool): whether null is treated as a value for the purpose of determining if the values in the
+            aggregation group are unique, default is False.
+        non_unique_sentinel (Union[np.number, str, bool]): the non-null sentinel value when no unique value exists,
+            default is None. Must be a non-None value when include_nulls is True. When passed in as a numpy scalar
+            number value, it must be of one of these types: np.int8, np.int16, np.uint16, np.int32, np.int64, np.float32
+            , np.float64.
```
chipkent, line 330: Does there need to be a note that `np.uint16` is interpreted as a char?

py/client/pydeephaven/agg.py
chipkent: I want to point out that there is a bit of funky usability to consider.  To actually do a char in py, you would do something like `np.uint16(ord("S"))`, which is on the complex side for a simple concept.  Not sure there is a better solution.

chipkent (changes requested):

py/client/pydeephaven/agg.py
jmao-denver: yes

py/client/pydeephaven/agg.py
jmao-denver: yes.

py/client/pydeephaven/agg.py
jmao-denver: Not sure if it helps much.

chipkent (approved):

chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2984
Reference: https://github.com/deephaven/deephaven.io/issues/2983

PR #3948, opened 2023-06-07
Minimal support for TLS connection and extra headers in python and C++ clients.

### LINKED ISSUE #3228: Add pydeephaven secure_channel support
The python `pydeephaven` client does not currently support secure connections.

We should be able to leverage https://grpc.github.io/grpc/python/grpc.html#grpc.secure_channel to connect securely when appropriate.

### LINKED ISSUE #3943: Inject Headers into pydeephaven Client
As a developer, I want to have a mechanism to inject arbitrary HTTP headers into the pydeephaven client grpc requests, so that I can integrate with proxies that expect them.

jcferretti:
Note the arrow flight client implementation itself is a very thin cython wrapper around the arrow flight C++ API, so the two changes in our clients in this PR, C++ and python, are functionally the same change being handled by the same underlying arrow flight C++ client API logic.

Resources:

* Python Arrow Flight cookbook for TLS
  https://arrow.apache.org/cookbook/py/flight.html#securing-connections-with-tls (Step 3)
* C++ Arrow Flight documentation for Connect
https://arrow.apache.org/docs/cpp/api/flight.html#_CPPv4N5arrow6flight12FlightClient7ConnectERK8LocationRK19FlightClientOptions
* C++ Arrow Flight documentation for Connect options:
https://arrow.apache.org/docs/cpp/api/flight.html#_CPPv4N5arrow6flight19FlightClientOptionsE
* Example of C++ TLS setup for client:
https://github.com/yasushi-saito/grpc-ssl-example/blob/master/cppclient/client.cc
* Python gRPC API
https://grpc.github.io/grpc/python/grpc.html#grpc.secure_channel
* DIscussion of target name override in the gRPC python repo:
https://github.com/grpc/grpc/issues/5697

Fixes #3228
Fixes #3943

### DIFF
--- a/cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
+++ b/cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
@@ -101,7 +101,8 @@ public:
       const std::string &target,
       const std::string &authorizationValue,
       bool use_tls = false,
-      const std::string &pem = "");
+      const std::string &pem = "",
+      const std::string &target_name_override = "");
   Server(const Server &other) = delete;
   Server &operator=(const Server &other) = delete;
   Server(Private,
--- a/cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
+++ b/cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
@@ -363,9 +363,6 @@ public:
     }
   }
 
-  std::pair<std::string, std::string> getAuthHeader() const;
-  const ClientOptions::extra_headers_t &getExtraHeaders() const;
-
   // TODO: make this private
   void setExpirationInterval(std::chrono::milliseconds interval);
 
--- a/cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
+++ b/cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
@@ -72,6 +72,7 @@ class ClientOptions {
 public:
   typedef std::vector<std::pair<std::string, int>> int_options_t;
   typedef std::vector<std::pair<std::string, std::string>> string_options_t;
+  typedef std::vector<std::pair<std::string, std::string>> extra_headers_t;
 public:
   static ClientOptions defaults() {
     return ClientOptions();
@@ -91,7 +92,7 @@ public:
   // a list of available options.
   // Example:
   //   copts.setStringOption("grpc.min_reconnect_backoff_ms", 2000)
-  ClientOptions &setIntOption(const std::string &opt, const int val) {
+  ClientOptions &addIntOption(const std::string &opt, const int val) {
     intOptions_.emplace_back(opt, val);
     return *this;
   }
@@ -100,8 +101,13 @@ public:
   // a list of available options.
   // Example:
   //   copts.setStringOption("grpc.target_name_override", "idonthaveadnsforthishost");
-  ClientOptions &setStringOption(const std::string &opt, const std::string &val) {
+  ClientOptions &addStringOption(const std::string &opt, const std::string &val) {
     stringOptions_.emplace_back(opt, val);
+    return *this;
+  }
+
+  ClientOptions &addExtraHeader(const std::string &header_name, const std::string &header_value) {
+    extraHeaders_.emplace_back(header_name, header_value);
     return *this;
   }
 
@@ -109,12 +115,14 @@ public:
   const std::string &pem() const { return pem_; }
   const int_options_t &intOptions() const { return intOptions_; }
   const string_options_t &stringOptions() const { return stringOptions_; }
+  const extra_headers_t &extraHeaders() const { return extraHeaders_; }
 private:
   ClientOptions()
     : useTls_(false)
     , pem_("")
     , intOptions_()
     , stringOptions_()
+    , extraHeaders_()
   {}
 
 private:
@@ -122,6 +130,7 @@ private:
   std::string pem_;
   int_options_t intOptions_;
   string_options_t stringOptions_;
+  extra_headers_t extraHeaders_;
 };
 
 class Server : public std::enable_shared_from_this<Server> {
@@ -170,6 +179,7 @@ public:
       std::unique_ptr<TableService::Stub> tableStub,
       std::unique_ptr<ConfigService::Stub> configStub,
       std::unique_ptr<arrow::flight::FlightClient> flightClient,
+      const ClientOptions::extra_headers_t &extraHeaders,
       std::string sessionToken,
       std::chrono::milliseconds expirationInterval,
       std::chrono::system_clock::time_point nextHandshakeTime);
@@ -284,6 +294,7 @@ public:
       TStub *stub, const TPtrToMember &pm);
 
   std::pair<std::string, std::string> getAuthHeader() const;
+  const ClientOptions::extra_headers_t &getExtraHeaders() const;
 
   // TODO: make this private
   void setExpirationInterval(std::chrono::milliseconds interval);
@@ -296,7 +307,7 @@ private:
   Ticket selectOrUpdateHelper(Ticket parentTicket, std::vector<std::string> columnSpecs,
       std::shared_ptr<EtcCallback> etcCallback, selectOrUpdateMethod_t method);
 
-  void addSessionToken(grpc::ClientContext *ctx);
+  void addMetadata(grpc::ClientContext *ctx);
 
   static void processCompletionQueueForever(const std::shared_ptr<Server> &self);
   bool processNextCompletionQueueItem();
@@ -310,6 +321,7 @@ private:
   std::unique_ptr<TableService::Stub> tableStub_;
   std::unique_ptr<ConfigService::Stub> configStub_;
   std::unique_ptr<arrow::flight::FlightClient> flightClient_;
+  const ClientOptions::extra_headers_t extraHeaders_;
   grpc::CompletionQueue completionQueue_;
 
   std::atomic<int32_t> nextFreeTicketId_;
@@ -328,7 +340,7 @@ void Server::sendRpc(const TReq &req, std::shared_ptr<SFCallback<TResp>> respons
   auto now = std::chrono::system_clock::now();
   // Keep this in a unique_ptr at first, for cleanup in case addAuthToken throws an exception.
   auto response = std::make_unique<ServerResponseHolder<TResp>>(now, std::move(responseCallback));
-  addSessionToken(&response->ctx_);
+  addMetadata(&response->ctx_);
   auto rpc = (stub->*pm)(&response->ctx_, req, &completionQueue_);
   // It is the responsibility of "processNextCompletionQueueItem" to deallocate the storage pointed
   // to by 'response'.
--- a/cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
+++ b/cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
@@ -97,7 +97,11 @@ class Server : public std::enable_shared_from_this<Server> {
   typedef SFCallback<ExportedTableCreationResponse> EtcCallback;
 
 public:
-  static std::shared_ptr<Server> createFromTarget(const std::string &target, const std::string &authorizationValue, const std::string pem = "");
+  static std::shared_ptr<Server> createFromTarget(
+      const std::string &target,
+      const std::string &authorizationValue,
+      bool use_ssl = false,
+      const std::string &pem = "");
   Server(const Server &other) = delete;
   Server &operator=(const Server &other) = delete;
   Server(Private,
--- a/cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
+++ b/cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
@@ -67,6 +67,7 @@ public:
   Response response_;
 };
 
+// Options for the creation of the underlying grpc channel.
 class ClientOptions {
 public:
   typedef std::vector<std::pair<std::string, int>> int_options_t;
@@ -81,16 +82,22 @@ public:
     return *this;
   }
 
-  ClientOptions &setPem(const std::string &pem) {
+  ClientOptions &setPem(const std::string &pem) {  // root certificate to use
     pem_ = pem;
     return *this;
   }
 
+  // See https://grpc.github.io/grpc/cpp/group__grpc__arg__keys.html for
+  // a list of available options.
   ClientOptions &setIntOption(const std::string &opt, const int val) {
     intOptions_.emplace_back(opt, val);
     return *this;
   }
 
+  // See https://grpc.github.io/grpc/cpp/group__grpc__arg__keys.html for
+  // a list of available options.
+  // Example:
+  //   copts.setStringOption("grpc.target_name_override", "idonthaveadnsforthishost");
   ClientOptions &setStringOption(const std::string &opt, const std::string &val) {
     stringOptions_.emplace_back(opt, val);
     return *this;
--- a/cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
+++ b/cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
@@ -12,6 +12,7 @@
 #include <cstdint>
 #include <arrow/flight/client.h>
 
+#include "deephaven/client/client_options.h"
 #include "deephaven/client/utility/executor.h"
 #include "deephaven/dhcore/utility/callbacks.h"
 #include "deephaven/proto/ticket.pb.h"
@@ -67,133 +68,6 @@ public:
   Response response_;
 };
 
-// Options for the creation of the underlying grpc channel.
-class ClientOptions {
-public:
-  typedef std::vector<std::pair<std::string, int>> int_options_t;
-  typedef std::vector<std::pair<std::string, std::string>> string_options_t;
-  typedef std::vector<std::pair<std::string, std::string>> extra_headers_t;
-
-  ~ClientOptions();
-
-  /**
-   * Get ClientOptions object initialized with defaults.
-   *
-   * @return A ClientOptions object initialized with defaults.
-   */
-  static ClientOptions defaults() {
-    return ClientOptions();
-  }
-
-  /**
-   * Configure whether to set server connections as TLS
-   *
-   * @param useTls true if server connections should be TLS/SSL, false for insecure.
-   * @return *this, to be used for chaining
-   */
-  ClientOptions &setUseTls(const bool useTls) {
-    useTls_ = useTls;
-    return *this;
-  }
-
-  /**
-   * Sets a PEM-encoded certificate root for server connections.  The empty string
-   * means use system defaults.
-   *
-   * @param pem a PEM encoded certificate chain.
-   * @return *this, to be used for chaining
-   */
-  ClientOptions &setPem(const std::string pem) {  // root certificate to use
-    pem_ = std::move(pem);
-    return *this;
-  }
-
-  /**
-   * Addss an int-valued option.
-   * See https://grpc.github.io/grpc/cpp/group__grpc__arg__keys.html for a list of available options.
-   *
-   * @example copt.setIntOption("grpc.min_reconnect_backoff_ms", 2000)
-   * @param opt The option key.
-   * @param val The option valiue.
-   * @return *this, to be used for chaining
-   */
-  ClientOptions &addIntOption(const std::string opt, const int val) {
-    intOptions_.emplace_back(std::move(opt), val);
-    return *this;
-  }
-
-  /**
-   * Adds a string-valued option.
-   * See https://grpc.github.io/grpc/cpp/group__grpc__arg__keys.html for a list of available options.
-   *
-   * @example copt.setStringOption("grpc.target_name_override", "idonthaveadnsforthishost")
-   * @param opt The option key.
-   * @param val The option valiue.
-   * @return *this, to be used for chaining
-   */
-  ClientOptions &addStringOption(const std::string opt, const std::string val) {
-    stringOptions_.emplace_back(std::move(opt), std::move(val));
-    return *this;
-  }
-
-  /**
-   * Adds an extra header with a constant name and value to be sent with every outgoing request.
-   *
-   * @param header_name The header name
-   * @param header_value The header value
-   * @return *this, to be used for chaining
-   */
-  ClientOptions &addExtraHeader(const std::string &header_name, const std::string &header_value) {
-    extraHeaders_.emplace_back(header_name, header_value);
-    return *this;
-  }
-
-  /**
-   * Returns true if server connections should be configured for TLS/SSL.
-   *
-   * @return true if this connection should be TLS/SSL, false for insecure.
-   */
-  bool useTls() const { return useTls_; }
-
-  /**
-   * The PEM-encoded certificate root for server connections, or the empty string
-   * if using system defaults.
-   *
-   * @return A PEM-encoded certificate chain
-   */
-  const std::string &pem() const { return pem_; }
-
-  /**
-   * Integer-valued channel options set for server connections.
-   *
-   * @return A vector of pairs of string option name and integer option value.
-   */
-  const int_options_t &intOptions() const { return intOptions_; }
-
-  /**
-   * String-valued channel options set for server connections.
-   *
-   * @return A vector of pairs of string option name and string option value.
-   */
-  const string_options_t &stringOptions() const { return stringOptions_; }
-
-  /**
-   * Extra headers that should be sent with each outgoing server request.
-   *
-   * @return A vector of pairs of string header name and string header value.
-   */
-  const extra_headers_t &extraHeaders() const { return extraHeaders_; }
-
-private:
-  ClientOptions();
-
-  bool useTls_ = false;
-  std::string pem_;
-  int_options_t intOptions_;
-  string_options_t stringOptions_;
-  extra_headers_t extraHeaders_;
-};
-
 class Server : public std::enable_shared_from_this<Server> {
   struct Private {
   };
@@ -226,11 +100,7 @@ class Server : public std::enable_shared_from_this<Server> {
 public:
   static std::shared_ptr<Server> createFromTarget(
       const std::string &target,
-      const std::string &authorizationValue);
-  static std::shared_ptr<Server> createFromTarget(
-      const std::string &target,
-      const std::string &authorizationValue,
-      const ClientOptions &client_options);
+      const deephaven::client::ClientOptions &client_options);
   Server(const Server &other) = delete;
   Server &operator=(const Server &other) = delete;
   Server(Private,
@@ -240,7 +110,7 @@ public:
       std::unique_ptr<TableService::Stub> tableStub,
       std::unique_ptr<ConfigService::Stub> configStub,
       std::unique_ptr<arrow::flight::FlightClient> flightClient,
-      const ClientOptions::extra_headers_t &extraHeaders,
+      const deephaven::client::ClientOptions::extra_headers_t &extraHeaders,
       std::string sessionToken,
       std::chrono::milliseconds expirationInterval,
       std::chrono::system_clock::time_point nextHandshakeTime);
@@ -389,7 +259,7 @@ private:
   std::unique_ptr<TableService::Stub> tableStub_;
   std::unique_ptr<ConfigService::Stub> configStub_;
   std::unique_ptr<arrow::flight::FlightClient> flightClient_;
-  const ClientOptions::extra_headers_t extraHeaders_;
+  const deephaven::client::ClientOptions::extra_headers_t extraHeaders_;
   grpc::CompletionQueue completionQueue_;
 
   std::atomic<int32_t> nextFreeTicketId_;
--- a/cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
+++ b/cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
@@ -74,47 +74,112 @@ public:
   typedef std::vector<std::pair<std::string, std::string>> string_options_t;
   typedef std::vector<std::pair<std::string, std::string>> extra_headers_t;
 
+  /**
+   * Get ClientOptions object initialized with defaults.
+   *
+   * @return A ClientOptions object initialized with defaults.
+   */
   static ClientOptions defaults() {
     return ClientOptions();
   }
 
+  /**
+   * Configure whether to set server connections as TLS
+   *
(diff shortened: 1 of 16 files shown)


py/client/pydeephaven/session.py
```diff
@@ -240,8 +244,17 @@ def _fetch_fields(self):
     def _connect(self):
         with self._r_lock:
             try:
-                self._flight_client = paflight.connect(location=(self.host, self.port), middleware=[
```
jcferretti, line 243: Note the `connect` call has been deprecated: https://github.com/apache/arrow/blob/c62ce6b179f67ff57fa1571b01f356739727c0c9/python/pyarrow/_flight.pyx#L1354

jmao-denver (changes requested):
We also need a test case for the Python client

py/client/pydeephaven/session.py
```diff
@@ -102,6 +102,9 @@ def __init__(self, host: str = None, port: int = None, auth_type: str = "Anonymo
                 authenticator, it must conform to the specific requirement of the authenticator
             never_timeout (bool, optional): never allow the session to timeout, default is True
             session_type (str, optional): the Deephaven session type. Defaults to 'python'
+            use_tls (bool, optional): if True, use a TLS connection.  Defaults to None
```
jmao-denver, line 105:
```suggestion
            use_tls (bool): if True, use a TLS connection.  Defaults to False
```

py/client/pydeephaven/_session_service.py
```diff
@@ -15,7 +15,16 @@ def __init__(self, session):
 
     def connect(self) -> grpc.Channel:
         """Connects to the server and returns a gRPC channel upon success."""
-        grpc_channel = grpc.insecure_channel(":".join([self.session.host, str(self.session.port)]))
+        target = ":".join([self.session.host, str(self.session.port)])
+        if self.session.target_name_override is None:
+            options = None
+        else:
+            options = (('grpc.ssl_target_name_override', self.session.target_name_override))
```
cpwright, line 22: I think you should let the user pass in the options as I would like, rather than trying to invent each and every one.  I also set my default authority, which is actually unnecessary (I took it out just now), while trying to debug this.  If we let you use the regular options, we'll not have problems with forgetting something and need another client release.

py/client/pydeephaven/session.py
```diff
@@ -240,8 +254,12 @@ def _fetch_fields(self):
     def _connect(self):
         with self._r_lock:
             try:
-                self._flight_client = paflight.connect(location=(self.host, self.port), middleware=[
-                    _DhClientAuthMiddlewareFactory(self)])
+                scheme = "grpc+tls" if use_tls else "grpc"
```
cpwright, line 257:
```suggestion
                scheme = "grpc+tls" if self.use_tls else "grpc"
```

py/client/pydeephaven/session.py
```diff
@@ -240,8 +254,12 @@ def _fetch_fields(self):
...
+                self._flight_client = paflight.FlightClient(
+                    location=f"{scheme}://{self.host}:{self.port}",
+                    middleware=[_DhClientAuthMiddlewareFactory(self)],
+                    tls_root_certs = pem
```
cpwright, line 261:
```suggestion
                    tls_root_certs = self.pem
```

cpwright (changes requested):
I know the use_tls one is right.  The self.pem is probably is also necessary; but I haven't been able to authenticate yet to prove it.

jcferretti:
> Because we use self-signed certificates, with names that don't match in addition to just the PEM data, we will need to specify grpc channel options. This is how we connect to our authentication server, for example using the regular python GRPC module:
> 
> ```
>     trustStore = "/Users/charleswright/iris/.programfiles/CV17D/resources/truststore-iris.pem"
> 
>     credentials = grpc.ssl_channel_credentials(root_certificates=open(trustStore, 'rb').read())
>     channel = grpc.secure_channel(server + ":" + str(auth_port), credentials, options=[('grpc.ssl_target_name_override', 'authserver'), ('grpc.default_authority', 'authserver')])
> ```
> 
> For things that have "nice" certificates, it is much easier and we can just use the defaults to create the channel, which looks something like. In this case we don't need to specify the PEM.
> 
> ```
>         channel = grpc.secure_channel(target, grpc.ssl_channel_credentials())
> ```

I added a `target_name_override` argument to the functions that create channels in both C++ and pyhon, with the associated logic for both channels.

py/client/pydeephaven/_session_service.py
jcferretti: Makes sense, done.

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
```diff
@@ -67,6 +67,63 @@ struct ServerResponseHolder final : public CompletionQueueCallback {
   Response response_;
 };
 
+// Options for the creation of the underlying grpc channel.
+class ClientOptions {
+public:
+  typedef std::vector<std::pair<std::string, int>> int_options_t;
+  typedef std::vector<std::pair<std::string, std::string>> string_options_t;
+public:
```
kosak, line 75: redundant `public:` ?

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
```diff
@@ -67,6 +67,63 @@ struct ServerResponseHolder final : public CompletionQueueCallback {
...
+
+  ClientOptions &setUseTls(const bool useTls) {
+    useTls_ = useTls;
+    return *this;
+  }
+
+  ClientOptions &setPem(const std::string &pem) {  // root certificate to use
+    pem_ = pem;
+    return *this;
+  }
```
kosak, line 88: Prefer having doxygen comments and using `std::move` as in

~~~
  /**
   * Sets the root certificate.
   * @param pem root certificate to use
   * @return *this, to be used for chaining.
   */
  ClientOptions &setPem(std::string pem) {
    pem_ = std::move(pem);
    return *this;
  }
~~~

cpp-client/deephaven/client/src/server/server.cc
```diff
@@ -75,26 +75,68 @@ const char *timeoutKey = "http.session.durationMs";
 const size_t handshakeResendIntervalMillis = 5 * 1000;
 }  // namespace
 
-std::shared_ptr<Server> Server::createFromTarget(const std::string &target, const std::string &authorizationValue) {
-  auto channel = grpc::CreateChannel(target, grpc::InsecureChannelCredentials());
+namespace {
+std::shared_ptr<grpc::ChannelCredentials> get_credentials(const bool useTls, const std::string &pem) {
```
kosak, line 79: can we name this `getCredentials`?  (camelCase vs snake_case)

cpp-client/deephaven/client/src/server/server.cc
```diff
@@ -75,26 +75,68 @@ const char *timeoutKey = "http.session.durationMs";
...
+      const std::string &target,
+      const std::string &authorizationValue) {
+  return createFromTarget(target, authorizationValue, ClientOptions::defaults());
+}
+
+std::shared_ptr<Server> Server::createFromTarget(
+      const std::string &target,
+      const std::string &authorizationValue,
+      const ClientOptions &copts) {
+  if (!copts.useTls() && copts.pem() != "") {
```
kosak, line 101: clang-tidy would prefer `!copts.pem.empty()`  (I'd slightly prefer it too)

cpp-client/deephaven/client/src/server/server.cc
```diff
@@ -75,26 +75,68 @@ const char *timeoutKey = "http.session.durationMs";
...
+    throw std::runtime_error(
+        "Server::createFromTarget: ClientOptions: useTls is false but pem provided");
+  }
+
+  grpc::ChannelArguments channel_args;
+  for (const std::pair<std::string, int>& opt : copts.intOptions()) {
```
kosak, line 107:
~~~
for (const auto &opt : copts.intOptions())
~~~

cpp-client/deephaven/client/src/server/server.cc
```diff
@@ -75,26 +75,68 @@ const char *timeoutKey = "http.session.durationMs";
...
+    channel_args.SetInt(opt.first, opt.second);
+  }
+  for (const std::pair<std::string, std::string>& opt : copts.stringOptions()) {
```
kosak, line 110:
~~~
for (const auto &opt : copts.stringOptions())
~~~

cpp-client/deephaven/client/src/server/server.cc
```diff
@@ -75,26 +75,68 @@ const char *timeoutKey = "http.session.durationMs";
...
 
   auto rc1 = arrow::flight::Location::Parse(flightTarget, &location);
   if (!rc1.ok()) {
-    auto message = stringf("Location::Parse(%o) failed, error = %o", flightTarget, rc1.ToString());
+    auto message = stringf("Location::Parse(%o) failed, error = %o",
+                           flightTarget, rc1.ToString());
     throw std::runtime_error(message);
   }
 
+  arrow::flight::FlightClientOptions options = arrow::flight::FlightClientOptions::Defaults();
```
kosak, line 133: I would do
~~~
auto options = ...
~~~

cpp-client/deephaven/client/src/server/server.cc
```diff
@@ -75,26 +75,68 @@ const char *timeoutKey = "http.session.durationMs";
...
-  auto flightTarget = "grpc://" + target;
+  auto flightTarget = ((copts.useTls()) ? "grpc+tls://" : "grpc://") + target;
   arrow::flight::Location location;
 
   auto rc1 = arrow::flight::Location::Parse(flightTarget, &location);
   if (!rc1.ok()) {
-    auto message = stringf("Location::Parse(%o) failed, error = %o", flightTarget, rc1.ToString());
+    auto message = stringf("Location::Parse(%o) failed, error = %o",
+                           flightTarget, rc1.ToString());
     throw std::runtime_error(message);
```
kosak, line 130:
```
throw std::runtime_error(DEEPHAVEN_DEBUG_MSG(message));
```

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
```diff
@@ -67,6 +67,63 @@ struct ServerResponseHolder final : public CompletionQueueCallback {
...
+private:
+  ClientOptions()
+    : useTls_(false)
+    , pem_("")
+    , intOptions_()
+    , stringOptions_()
+  {}
+
+private:
+  bool useTls_;
```
kosak, line 121: I'm slightly in favor of putting the defaults here, setting useTls_ specifically and leaving the others allone, as in:

```
  bool useTls_ = false;
  std::string pem_;
  int_options_t intOptions_;
  string_options_t stringOptions_;
```

And then making a constructor and destructor declaration here:

~~~
ClientOptions();
~ClientOptions();
~~~

and then in server.cc adding

```
ClientOptions::ClientOptions() = default;
ClientOptions::~ClientOptions() = default;
```

(the rationale for moving the definitions outline, into the cc file, is that the constructor code is "big enough" not to be inline)

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
```diff
@@ -67,6 +67,63 @@ struct ServerResponseHolder final : public CompletionQueueCallback {
...
+  }
+
+  // See https://grpc.github.io/grpc/cpp/group__grpc__arg__keys.html for
+  // a list of available options.
+  // Example:
+  //   copts.setStringOption("grpc.target_name_override", "idonthaveadnsforthishost");
+  ClientOptions &setStringOption(const std::string &opt, const std::string &val) {
+    stringOptions_.emplace_back(opt, val);
+    return *this;
+  }
```
kosak, line 106: Prefer having doxygen comments, plus std::move style, as in
```
  /**
   * Sets an int-valued option.
   * See https://grpc.github.io/grpc/cpp/group__grpc__arg__keys.html for a list of available options.
   *
   * @example copt.setIntOption("grpc.min_reconnect_backoff_ms", 2000)
   * @param opt The option key.
   * @param val The option valiue.
   * @return *this, to be used for chaining
   */
  ClientOptions &setIntOption(std::string opt, const int val) {
    intOptions_.emplace_back(std::move(opt), val);
    return *this;
  }

  /**
   * Sets a string-valued option.
   * See https://grpc.github.io/grpc/cpp/group__grpc__arg__keys.html for a list of available options.
   *
   * @example copt.setStringOption("grpc.target_name_override", "idonthaveadnsforthishost")
   * @param opt The option key.
   * @param val The option valiue.
   * @return *this, to be used for chaining
   */
  ClientOptions &setStringOption(std::string opt, std::string val) {
    stringOptions_.emplace_back(std::move(opt), std::move(val));
    return *this;
  }
```

cpp-client/deephaven/client/src/server/server.cc
```diff
@@ -75,26 +75,68 @@ const char *timeoutKey = "http.session.durationMs";
...
-std::shared_ptr<Server> Server::createFromTarget(const std::string &target, const std::string &authorizationValue) {
-  auto channel = grpc::CreateChannel(target, grpc::InsecureChannelCredentials());
+namespace {
+std::shared_ptr<grpc::ChannelCredentials> get_credentials(const bool useTls, const std::string &pem) {
+  if (!useTls) {
+    return grpc::InsecureChannelCredentials();
+  }
+  grpc::SslCredentialsOptions options;
+  if (pem != "") {
+    options.pem_root_certs = pem;
```
kosak, line 85:
~~~
if (!pem.empty()) {
    options.pem_root_certs = std::move(pem);
}
~~~

cpwright (dismissed):

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
jcferretti: I was used to doing that as a kind of "section separator" for typedefs.  I'll remove it.

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
jcferretti: This is a private header file.  There are no doxygen comments anywhere in this file.

cpp-client/deephaven/client/src/server/server.cc
jcferretti: We don't consistently use DEEPHAVEN_DEBUG_MSG everywhere else in this file (ie, we were throwing in other cases without it), I looked up the definition and tried to deduce a rationale for when to use it, I guess I deduced wrong?

kosak: Yeah. I would recommend using it on every single throw. It's definitely not done in a lot of places. But, anywhere it isn't being done is an oversight. :-)

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
jcferretti: We don't need to define a destructor, right?  The one that we get implicitly defined is right, and unlike the constructor, we don't need (actually it won't work if) the destructor is private.

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
kosak: That is a good point. However (and this is just a proposal) I think I would like to migrate to doxygen style everywhere, even for private header files. This is because my IDE renders them (when you hit control-Q on a symbol) which I think is great and looks nice.

So I think I might like to propose that, even for private header files, if there's going to be a comment provided, maybe it should be a doxygen-style comment.

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
kosak: If we don't define a destructor, the implicitly-defined one will be inlined. My personal style is to explicitly define a destructor (even if that definition is `ClientOptions::~ClientOptions() = default`) so that it is only expressed in one place in the code rather than sprinkled around many places as inlined code.

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
jcferretti: Ok, I am going to add doxygen comments to ClientOptions.  But I reiterate, there are no doxygen comments anywhere else in this file.  createForTarget seems to want them, and I guess this file being in the private/ part of the dir structure is maybe not right?

kosak: Part of what is happening is that my preferences are changing over time. I'm starting to believe in a lifestyle that says that that comments, when they exists, should be doxygen-style.

As to your public/private question. Now that I look at it, and assuming we are keeping the code in the same structure as it is now, then the only code that calls `Server::createFromTarget` is `Client::connect`.

Customers can only access `Client::connect` (because it is public). They don't even see Server::createFromTarget because it is private.

So this is starting to make me wonder whether your `ClientOptions` functionality (as defined in server.h) should be merged with my not-so-coincidentally-named `ClientOptions` class (defined in include/public/deephaven/client/client.h)

And then `Client::connect` can pull out the arguments it cares about (like it does now) and then pass the config object to `Server::createFromTarget`

I think to do this, `ClientOptions` should be moved to its own include file, in the public side of the directory tree, so that both Client and Server can see it.

Does this make sense? This comment is sort of complicated so we may need to discuss it offline.

kosak: Furthermore, it would be the style of the rest of the code for the "private" code to never reference stuff on the "public side", so I think one way I *might* do it is to have the existing public-side ClientOptions have its own methods plus a pointer to an opaque ServerOptions class, and that ServerOptions class is what has your TLS stuff and is what is passed to Server::createFromTarget.

This is another thing which is hard to type but perhaps easy to explain in person. Or if you want me to add to the branch, I can show you what I mean, in code.

cpp-client/deephaven/client/src/server/server.cc
jcferretti: Understood.

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
jcferretti: Ok, for this kind of object that makes a lot of sense.

cpwright (dismissed):
Python LGTM.

py/client/pydeephaven/session.py
```diff
@@ -100,8 +103,20 @@ def __init__(self, host: str = None, port: int = None, auth_type: str = "Anonymo
...
                 "user:password"; when auth_type is "Anonymous', it will be ignored; when auth_type is a custom-built
                 authenticator, it must conform to the specific requirement of the authenticator
-            never_timeout (bool, optional): never allow the session to timeout, default is True
-            session_type (str, optional): the Deephaven session type. Defaults to 'python'
+            never_timeout (bool): never allow the session to timeout, default is True
+            session_type (str): the Deephaven session type. Defaults to 'python'
+            use_tls (bool): if True, use a TLS connection.  Defaults to False
+            pem (bytes): PEM encoded certificate to use for TLS connection, or None to use system defaults.
+                 If not None implies use a TLS connection and the use_tls argument should have been passed
+                 as True. Defaults to None
```
jmao-denver, line 111: is this the root cert? or the certs of the client?

py/client/pydeephaven/session.py
```diff
@@ -162,7 +184,10 @@ def tables(self):
 
     @property
     def grpc_metadata(self):
-        return [(b'authorization', self._auth_token)]
+        l =[(b'authorization', self._auth_token)]
+        if self._extra_headers:
+            l.extend(list(tuple(self._extra_headers.items())))
```
jmao-denver, line 189:
```suggestion
            l.extend(list(self._extra_headers.items()))
```

py/client/pydeephaven/session.py
```diff
@@ -88,7 +88,10 @@ class Session:
     """
 
     def __init__(self, host: str = None, port: int = None, auth_type: str = "Anonymous", auth_token: str = "",
-                 never_timeout: bool = True, session_type: str = 'python'):
+                 never_timeout: bool = True, session_type: str = 'python',
+                 use_tls: bool = False, pem: bytes = None,
+                 client_opts: List[Tuple[str,Union[int,str]]] = None,
+                 extra_headers: Dict[bytes, bytes] = None):
```
jmao-denver, line 94: change style to one parameter per line?

jmao-denver (changes requested):
Some minor issues.

py/client/pydeephaven/session.py
jcferretti: root, I'll make the doc more specific

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
```diff
@@ -97,7 +98,9 @@ class Server : public std::enable_shared_from_this<Server> {
   typedef SFCallback<ExportedTableCreationResponse> EtcCallback;
 
 public:
-  static std::shared_ptr<Server> createFromTarget(const std::string &target, const std::string &authorizationValue);
+  static std::shared_ptr<Server> createFromTarget(
+      const std::string &target,
+      const deephaven::client::ClientOptions &client_options);
```
kosak, line 103: I might add
```
typedef deephaven::client::ClientOptions ClientOptions;
```

at line 94 or so, so that this (and below) can be shortened to `ClientOptions`

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
```diff
@@ -107,6 +110,7 @@ class Server : public std::enable_shared_from_this<Server> {
       std::unique_ptr<TableService::Stub> tableStub,
       std::unique_ptr<ConfigService::Stub> configStub,
       std::unique_ptr<arrow::flight::FlightClient> flightClient,
+      const deephaven::client::ClientOptions::extra_headers_t &extraHeaders,
```
kosak, line 113: Shorten name and then pass by value (we will change server.cc to use std::move)
```
ClientOptions::extra_headers_t extraHeaders,
```

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
```diff
@@ -220,21 +224,29 @@ class Server : public std::enable_shared_from_this<Server> {
   void sendRpc(const TReq &req, std::shared_ptr<SFCallback<TResp>> responseCallback,
       TStub *stub, const TPtrToMember &pm);
 
-  std::pair<std::string, std::string> getAuthHeader() const;
+  template<typename Fun>
+  void forEachHeaderNameAndValue(Fun fun) {
```
kosak, line 228: suggest making this not a template:

```
#include <functional>
```
at top of file and then change this declaration to
```
 void forEachHeaderNameAndValue(const std::function<void(const std::string &, const std::string &)> &fun);
```

and then move the body to server.cc.  Your callers should not need to change.

cpp-client/deephaven/client/src/server/server.cc
```diff
@@ -467,15 +513,6 @@ void Server::releaseAsync(Ticket ticket, std::shared_ptr<SFCallback<ReleaseRespo
...
 
-std::pair<std::string, std::string> Server::getAuthHeader() const {
-  return std::make_pair(authorizationKey, sessionToken_);
-}
-
-void Server::addSessionToken(grpc::ClientContext *ctx) {
-  std::lock_guard guard(mutex_);
-  ctx->AddMetadata(authorizationKey, sessionToken_);
-}
-
```
kosak, line 478: Somewhere in this file please add
```
  void Server::forEachHeaderNameAndValue(const std::function<void(const std::string &, const std::string &)> &fun) {
    {
      std::lock_guard guard(mutex_);
      fun(authorizationKey, sessionToken_);
    }
    for (const auto &header : extraHeaders_) {
      fun(header.first, header.second);
    }
  }
```

By the way ... it's just a bit of an anti-pattern in my coding style to invoke callbacks under lock. Because something horrible can happen in the callback and then you have a deadlock. You might want to consider this more defensive approach

```
mutex_.lock();
auto tokenCopy = sessionToken_;
mutex_.unlock();
fun(authorizationKey, tokenCopy);
```

kosak (commented):
Submitting some comments for now, pending a reload, so I can sync with your latest changes

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
jcferretti: Cool, I am liking this idiom, getting more assimilated in my brain as I use it more.

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
jcferretti: Much better.

cpp-client/deephaven/client/src/server/server.cc
jcferretti: Thanks, didn't think about that, and I consider that an anti pattern myself as well.

cpp-client/deephaven/client/include/private/deephaven/client/server/server.h
```diff
@@ -247,6 +259,7 @@ class Server : public std::enable_shared_from_this<Server> {
   std::unique_ptr<TableService::Stub> tableStub_;
   std::unique_ptr<ConfigService::Stub> configStub_;
   std::unique_ptr<arrow::flight::FlightClient> flightClient_;
+  const deephaven::client::ClientOptions::extra_headers_t extraHeaders_;
```
kosak: can shorten to
```
ClientOptions::extra_headers_t
```

cpp-client/deephaven/client/include/public/deephaven/client/client_options.h
```diff
@@ -0,0 +1,192 @@
...
+#include <vector>
+
+namespace deephaven::client {
+
+class Client;
+ 
+/**
+ * The ClientOptions object is intended to be passed to Client::connect(). For convenience, the mutating methods can be
+ * chained. For example:
+ * auto client = Client::connect("localhost:10000", ClientOptions().setBasicAuthentication("foo", "bar").setSessionType("groovy")
```
kosak, line 17: I think I'm the author of this comment, but "while you're here" can you delete "For example:" and use @example as in

```
 * The ClientOptions object is intended to be passed to Client::connect(). For convenience, the mutating methods can be
 * chained.
 * @example auto client = Client::connect("localhost:10000", ClientOptions().setBasicAuthentication("foo", "bar").setSessionType("groovy")
```

cpp-client/deephaven/client/src/client_options.cc
```diff
@@ -0,0 +1,77 @@
...
+  return *this;
+}
+
+ClientOptions &ClientOptions::setCustomAuthentication(const std::string &authenticationType,
+    const std::string &authenticationToken) {
+  authorizationValue_ = authenticationType + " " + authenticationToken;
+  return *this;
+}
+
+ClientOptions &ClientOptions::setSessionType(const std::string &sessionType) {
```
kosak, line 37: `std::string sessionType` and `std::move`

cpp-client/deephaven/client/src/client_options.cc
```diff
@@ -0,0 +1,77 @@
...
+  this->sessionType_ = sessionType;
+  return *this;
+}
+
+ClientOptions &ClientOptions::setUseTls(const bool useTls) {
+  useTls_ = useTls;
+  return *this;
+}
+
+ClientOptions &ClientOptions::setTlsRootCerts(const std::string tlsRootCerts) {
```
kosak, line 47: Please remove `const` here, because `const` is preventing std::move from happening and downgrading the operation to a copy.

cpp-client/deephaven/client/src/client_options.cc
```diff
@@ -0,0 +1,77 @@
...
+  tlsRootCerts_ = std::move(tlsRootCerts);
+  return *this;
+}
+
+ClientOptions &ClientOptions::setClientCertChain(const std::string clientCertChain) {
```
kosak, line 52: Please remove `const` here, because `const` is preventing std::move from happening and downgrading the operation to a copy.

py/client/pydeephaven/session.py
```diff
@@ -118,6 +145,15 @@ def __init__(self, host: str = None, port: int = None, auth_type: str = "Anonymo
         if not port:
             self.port = int(os.environ.get("DH_PORT", 10000))
 
+        self._use_tls = use_tls
+        self._tls_root_certs = tls_root_certs
+        self._client_cert_chain = client_cert_chain
+        self._client_private_key = client_private_key
+        if self._tls_root_certs is not None and not self._use_tls:
+            raise DHError("use_tls is false but tsl_root_certs is not None")
```
jmao-denver: There are some invalid argument combos not checked. Cristian and I discussed the best way to handle it and we both agreed that no check is probably the best, just let the lower level gRPC throw the exceptions which will be re-thrown as a DHError to the application.

cpp-client/deephaven/client/src/client_options.cc
```diff
@@ -0,0 +1,77 @@
...
+  clientCertChain_ = std::move(clientCertChain);
+  return *this;
+}
+
+ClientOptions &ClientOptions::setClientPrivateKey(const std::string clientPrivateKey) {
```
kosak, line 57: Please remove `const` here, because `const` is preventing std::move from happening and downgrading the operation to a copy.

cpp-client/deephaven/client/src/client_options.cc
```diff
@@ -0,0 +1,77 @@
...
+  clientPrivateKey_ = std::move(clientPrivateKey);
+  return *this;
+}
+
+ClientOptions &ClientOptions::addIntOption(const std::string opt, const int val) {
```
kosak, line 62: Please remove `const` here, because `const` is preventing std::move from happening and downgrading the operation to a copy.

cpp-client/deephaven/client/src/client_options.cc
```diff
@@ -0,0 +1,77 @@
...
+  intOptions_.emplace_back(std::move(opt), val);
+  return *this;
+}
+
+ClientOptions &ClientOptions::addStringOption(const std::string opt, const std::string val) {
```
kosak, line 67: Please remove `const` here, because `const` is preventing std::move from happening and downgrading the operation to a copy.

jmao-denver (approved):
The Python changes LGTM.

kosak (commented):
That's it! I think I'm done! Thank you for all this

jmao-denver (approved):

kosak (approved):

PR #3952, opened 2023-06-08
Ensure AuthContext is Available to Table Loggers

nbauernfeind:
This required making `:Stats` depend on `engine-context`. This introduced a cyclical dependency without moving `io.deephaven.stats.Driver` to engine-table.

I manually inspected all call sites of table logger `log` methods. These were the remaining without ExecutionContexts.


buildSrc/src/main/groovy/io.deephaven.java-classpath-conventions.gradle
```diff
@@ -32,12 +32,12 @@ configurations {
     fishNet.extendsFrom fishIo
     fishNumerics.extendsFrom fishBase
     fishUtil.extendsFrom fishConfig
-    fishStats.extendsFrom fishUtil
+    fishNet.extendsFrom fishUtil
```
rcaudy, line 35: This seems incorrect and/or duplicative (since you have `fishNet.extendsFrom` twice). I think you can delete this line and have `dhUtil` depend on `fishIo`. I think some usages of `:Net` in gradle build files might be overkill.

Not for the PR, but we should be deleting some of this code. Net and a lot of what's in IO should be replaced.

engine/context/build.gradle
```diff
@@ -11,11 +11,11 @@ configurations {
 
 dependencies {
     api project(':authentication')
+    api project(':Util')
```
rcaudy, line 14: Why is this an API dep now?

rcaudy (commented):
This remains incomplete, I think. There are other loggers that need this treatment.

buildSrc/src/main/groovy/io.deephaven.java-classpath-conventions.gradle
nbauernfeind: Opened #3962 for clean up task. Your gradle suggestions appear to work and are on their way.

buildSrc/src/main/groovy/io.deephaven.java-classpath-conventions.gradle
```diff
@@ -31,13 +31,12 @@ configurations {
     fishDataGenerator.extendsFrom jdom
     fishNet.extendsFrom fishIo
     fishNumerics.extendsFrom fishBase
-    fishUtil.extendsFrom fishConfig
-    fishStats.extendsFrom fishUtil
+    fishUtil.extendsFrom fishConfig, fishIo
```
rcaudy, line 34: I think this change should come out, and probably you should add fishIo to like 39 (for dhUtil).

buildSrc/src/main/groovy/io.deephaven.java-classpath-conventions.gradle
```diff
@@ -32,12 +32,11 @@ configurations {
     fishNet.extendsFrom fishIo
     fishNumerics.extendsFrom fishBase
     fishUtil.extendsFrom fishConfig
-    fishStats.extendsFrom fishUtil
     fishBaseTest.extendsFrom junit
     fishIoTest.extendsFrom fishBaseTest
 
     dhNumerics.extendsFrom fishNumerics, jama
-    dhUtil.extendsFrom commonsIo, commonsLang3, commonsText, fishUtil, fishStats, jdom
+    dhUtil.extendsFrom commonsIo, commonsLang3, commonsText, fishUtil, fishNet, fishIo, jdom
```
rcaudy, line 39: I think `fishIo` might be redundant now, but let's not worry about this.

rcaudy (approved):

PR #3955, opened 2023-06-08
Port Symbol table aggregation improvements from DHE

lbooker42:


engine/table/src/main/java/io/deephaven/engine/table/impl/by/StaticSymbolTableChunkedOperatorAggregationStateManager.java
```diff
@@ -0,0 +1,136 @@
...
+        return MAX_ARRAY_SIZE;
+    }
+
+    @Override
+    public SafeCloseable makeAggregationStateBuildContext(final ColumnSource<?>[] buildSources, long maxSize) {
+        return null;
+    }
+
+    @Override
+    public void add(final SafeCloseable bc, final RowSequence orderedKeys, final ColumnSource<?>[] sources, final MutableInt nextOutputPosition, final WritableIntChunk<RowKeys> outputPositions) {
```
rcaudy, line 57: `orderedKeys`->`rowSequence`

engine/table/src/main/java/io/deephaven/engine/table/impl/by/StaticSymbolTableChunkedOperatorAggregationStateManager.java
```diff
@@ -0,0 +1,136 @@
...
+        if (orderedKeys.isEmpty()) {
+            return;
+        }
+
+        outputPositions.setSize(orderedKeys.intSize());
+
+        try (final RowSequence.Iterator okIt = orderedKeys.getRowSequenceIterator();
+             final ChunkSource.FillContext fillContext = mappedKeySource.makeFillContext(CHUNK_SIZE);
```
rcaudy, line 65: Size these based on `min(CHUNK_SIZE, rowSequence.intSize())`

engine/table/src/main/java/io/deephaven/engine/table/impl/by/StaticSymbolTableChunkedOperatorAggregationStateManager.java
```diff
@@ -0,0 +1,136 @@
...
+        nextOutputPosition.setValue(nextPosition);
+    }
+
+    private void updateKeyHashTableSources(final WritableLongChunk<RowKeys> symbolTableValues, final int firstNewPosition) {
+        keyColumn.ensureCapacity(nextPosition);
+
+        final ColumnSource<?> symbolColumnSource = symbolTable.getColumnSource(SymbolTableSource.SYMBOL_COLUMN_NAME);
+
+        int symbolIterIdx = 0;
+        int nextPosition = firstNewPosition;
```
rcaudy, line 111: Please do something about this name. It's less clear than it could be by hiding the instance member.

engine/table/src/test/java/io/deephaven/engine/table/impl/QueryTableAggregationTest.java
```diff
@@ -3824,4 +3832,54 @@ public void testPreserveEmptyNoKey() {
...
+            testFunction.accept(table);
+
+            table.close();
+        } finally {
+            FileUtils.deleteRecursively(directory);
+        }
+    }
+
+    @NotNull
+    private Table makeDiskTable(File directory) throws IOException {
```
rcaudy, line 3867: I think we should be testing with more regions. Should build multiple Parquet files with partially-overlapping sets of unique symbol values, created in different order (meaning the relative order of the first encounter of each symbol should be different across files).

engine/table/src/test/java/io/deephaven/engine/table/impl/QueryTableAggregationTest.java
```diff
@@ -3824,4 +3834,97 @@ public void testPreserveEmptyNoKey() {
...
+        final File testRootFile = Files.createTempDirectory(QueryTableAggregationTest.class.getName()).toFile();
+        try {
+            final Table t1 = new InMemoryTable(
+                    new String[] {"StringKeys", "GroupedInts"},
+                    new Object[] {
+                            new String[] {"key1", "key2", "key1", "key2", "key1", "key2", "key1", "key2", "key1"},
+                            new short[] {1, 1, 2, 2, 2, 3, 3, 3, 3}
+                    });
+            final Table t1_asc = t1.sort("StringKeys");
+            final Table t1_desc = t1.sortDescending("StringKeys");
```
rcaudy, line 3849: I would like to see a case where the sets of defined symbols are not identical. I would leave the overall introduction order aligned with the names, just to keep things simple in a debugger. I would also like to see `null` symbols exercised.

Suggested:
{key1, key1,  key2, key2, key3, key2, key4, key2, key1}
{key4, null, key3, null, key1, null, null}
{key5, key4, key2, key5, key6}
{null, key6, key6, key6, key6, key7, null}

Be sure to run this test with coverage to make sure we're hitting the intended cases.

rcaudy (approved):

PR #3981, opened 2023-06-12
Remove nightly-benchmarks.yml

devinrsmith:
This is an old benchmarks workflow. It's been broken for a while, and causing our dependabot update script to error out because it is unable to parse this file.

image


rcaudy (approved):
OK to remove it, but seems like we might have to force merge.

PR #3982, opened 2023-06-13
Update versions of C++ dependent libraries.

jcferretti:
Since the build script is in the base images repo, this PR need will not pass checks until we merge the required changes in the matching PR there, at which point we can update the version of the base image in this PR to point to that new one and this PR check should pass (or at least, be meaningful).

Github actions checks are expected to fail before then.

Notes:
* All C++ protos were re-generated with the matching protoc version.


jcferretti:
That PR tried to update arrow to 12.0.0, which results in the cpp client hemorraging core dumps, which I don't have
time to debug right now, so closing for now and doing a more humble version here: https://github.com/deephaven/deephaven-core/pull/4000

PR #3984, opened 2023-06-13
Set jvm_maxmem on init_jvm

devinrsmith:
I suspect we are occasionally running into cases where GHA CI will hard-kill our jobs for going over memory limits.

In essence, the infrastructure we are currently running on has 7GB RAM and we are saying:

* Gradle can have 6GB
* `Integrations:test-py-deephavenRun` can have 80% of 7GB (~5.6GB)

Assuming nightlies pass, I think we can get away w/ Gradle at 5GB (for normal / local development purposes, we have `org.gradle.jvmargs=-Xms1g -Xmx3g`, see gradle.properties), and explicitly set the python integration tests at 1GB.

### DIFF
--- a/.github/env/Linux/gradle.properties
+++ b/.github/env/Linux/gradle.properties
@@ -8,7 +8,7 @@
 
 org.gradle.parallel=false
 org.gradle.workers.max=1
-org.gradle.jvmargs=-Xmx6g
+org.gradle.jvmargs=-Xmx5g
 
 # Our CI JDKs should be pre-provisioned and invoked correctly,
 # we shouldn't rely on gradle for any of this logic.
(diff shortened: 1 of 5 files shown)


.github/env/Linux/gradle.properties
```diff
@@ -8,7 +8,7 @@
 
 org.gradle.parallel=false
 org.gradle.workers.max=1
-org.gradle.jvmargs=-Xmx6g
+org.gradle.jvmargs=-Xmx5g
```
rcaudy, line 11: We're decreasing?

devinrsmith: Yes, added description to PR.

rcaudy (approved):

PR #3998, opened 2023-06-14
Improve the release process to better handle PyPi upload failures

devinrsmith:
This adds the deephaven-core and deephaven-server wheel as a GH artifact, and removes the netty artifacts (as we mainly just build netty to make sure it still builds, but we don't use the output otherwise).


stanbrub (approved):
Looks clearer.

rcaudy (approved):

chipkent (approved):

PR #4000, opened 2023-06-14
Update versions of C++ dependent libraries, take 2.

jcferretti:
Since the build script is in the base images repo, this PR need will not pass checks until we merge the required changes in the matching PR there, at which point we can update the version of the base image in this PR to point to that new one and this PR check should pass (or at least, be meaningful).

https://github.com/deephaven/deephaven-base-images/pull/61

Github actions checks are expected to fail before then.

Notes:

This is a reduced-scope version of the original PR at https://github.com/deephaven/deephaven-core/pull/3982
That PR tried to update arrow to 12.0.0, which results in the cpp client hemorraging core dumps, which I don't have
time to debug right now.


kosak (dismissed):

devinrsmith (approved):
Assuming CMakeLists.txt change is expected, LGTM assuming CI passes.

PR #4001, opened 2023-06-14
Update R Client to keep up with cpp client, add is_static and num_rows

alexpeters1208:


R/rdeephaven/R/client_wrapper.R
```diff
@@ -53,8 +53,10 @@ Client <- R6Class("Client",
         #' a variable name on the server. See `?TableHandle` for more information.
         #' @param table_object An R Data Frame, a dplyr Tibble, an Arrow Table, or an Arrow RecordBatchReader
         #' containing the data to import to the server.
+        #' @param num_rows Optional integer argument, only used if table_object is an Arrow RecordBatchReader.
+        #' Number of rows in the table represented by the RecordBatchReader.
```
chipkent, line 57: Should indicate that all rows are imported by default.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -53,8 +53,10 @@ Client <- R6Class("Client",
...
         #' @return TableHandle reference to the new table.
-        import_table = function(table_object) {
+        import_table = function(table_object, num_rows=NULL) {
```
chipkent, line 59: The change in this file makes no sense to me.  If a user wants fewer rows, they should do `head`, `tail`, or `where` to get less rows.  Providing this functionality for just one type of input feels like a bad API decision.  It should work everywhere or nowhere, and I think nowhere makes more sense.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -2,6 +2,9 @@
 #' @description Deephaven TableHandles are references to tables living on a Deephaven server. They provide an
 #' interface for interacting with tables on the server.
 #' 
+#' @field is_static Indicator variable for whether the table referenced by this TableHandle is static or not.
+#' @field num_rows Number of rows in the table referenced by this TableHandle. This value is only well-defined for static tables.
```
chipkent, line 6: I _think_ you are querying the number of rows at construction time and setting that as a constant.  That is never going to give the desired result.  It should be a function that gets the current value.

For API design, there is a principle of least surprise.  You want to minimize how many times you surprise the user -- ideally never.  This result will be very surprising in many cases.  On the other hand, `is_static` is an invariant, so it will never surprise the user.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -82,9 +87,29 @@ TableHandle <- R6Class("TableHandle",
...
 
+        #' @description
+        #' Number of rows in the table referenced by this TableHandle. This value is only well-defined for static tables.
+        #' @return The number of rows in the table, if the table is static.
+        num_rows = function() {
+            if(!private$is_static_field) {
+                stop("The table referenced by this TableHandle is not a static table, so the number of rows is not well-defined.")
+            }
+            return(private$num_rows_field)
+        }
```
chipkent, line 108: This should query the server for the current size.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -82,9 +87,29 @@ TableHandle <- R6Class("TableHandle",
...
         }
+    ),
+    active = list(
+
+        #' @description
+        #' Indicator variable for whether the table referenced by this TableHandle is static or not.
+        #' @return TRUE if the table is static, FALSE if the table is ticking.
+        is_static = function() {
+            return(private$is_static_field)
+        },
```
chipkent, line 98: I'm confused by this R -- or possibly just the documentation.  Is `is_static` a variable or a function?  From reading the code, it looks like a function, so these docs should not talk about indicator variables.  Also, it seems like the class docs for the fields should be removed to be consistent with the rest of the docs.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -35,6 +38,8 @@ TableHandle <- R6Class("TableHandle",
                 you are trying to call the constructor of TableHandle directly, which is not advised.")
             }
             private$internal_table_handle <- table_handle
+            private$is_static_field <- private$internal_table_handle$is_static()
+            private$num_rows_field <- private$internal_table_handle$num_rows()
```
chipkent, line 42: I think these variables are unnecessary, since it looks like the values are retrieved via function.

R/rdeephaven/src/client.cpp
```diff
@@ -24,14 +24,28 @@ class TableHandleWrapper {
...
+    bool isStatic() {
+        return internal_tbl_hdl.isStatic();
+    };
+
+    /**
+     * Number of rows in the table at the time internal_tbl_hdl was created.
+    */
+    int64_t numRows() {
+        return internal_tbl_hdl.numRows();
+    };
```
chipkent, line 39: See comments on least surprise and expected results.

R/rdeephaven/src/client.cpp
```diff
@@ -138,7 +152,7 @@ class ClientWrapper {
      * Uses a pointer to a populated ArrowArrayStream C struct to create a new table on the server from the data in the C struct.
      * @param stream_ptr Pointer to an existing and populated ArrayArrayStream, populated by a call to RecordBatchReader$export_to_c(ptr) from R.
     */
-    TableHandleWrapper* newTableFromArrowArrayStreamPtr(Rcpp::XPtr<ArrowArrayStream> stream_ptr) {
+    TableHandleWrapper* newTableFromArrowArrayStreamPtr(Rcpp::XPtr<ArrowArrayStream> stream_ptr, int64_t numRows) {
```
chipkent, line 155: See comments on API.

R/rdeephaven/src/client.cpp
```diff
@@ -200,25 +213,27 @@ RCPP_EXPOSED_CLASS(ArrowArrayStream)
...
-
-    class_<ClientWrapper>("INTERNAL_Client")
-    .factory<const std::string&, const ClientOptionsWrapper&>(newClientWrapper)
-    .method("open_table", &ClientWrapper::openTable)
-    .method("check_for_table", &ClientWrapper::checkForTable)
-    .method("run_script", &ClientWrapper::runScript)
-    .method("new_arrow_array_stream_ptr", &ClientWrapper::newArrowArrayStreamPtr)
-    .method("new_table_from_arrow_array_stream_ptr", &ClientWrapper::newTableFromArrowArrayStreamPtr)
-    ;
+        class_<TableHandleWrapper>("INTERNAL_TableHandle")
```
chipkent, line 216: This is hard to review because of formatting changes.  Use the autoformatter on the code and stick with whatever it spits out.

chipkent (changes requested):

R/rdeephaven/R/client_wrapper.R
```diff
@@ -63,7 +63,8 @@ Client <- R6Class("Client",
                 return(TableHandle$new(private$tibble_to_dh_table(table_object)))
             }
             else if (table_object_class[[1]] == "RecordBatchReader") {
-                return(TableHandle$new(private$rbr_to_dh_table(table_object)))
+                num_rows = dim(table_object$read_table())[[1]] # TODO: delete when Corey fixes c++ api
```
chipkent, line 66: Keep names out of the code.  Best is to put in a link to the ticket in github.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -35,6 +35,24 @@ TableHandle <- R6Class("TableHandle",
                 you are trying to call the constructor of TableHandle directly, which is not advised.")
             }
             private$internal_table_handle <- table_handle
+            private$is_static_field <- private$internal_table_handle$is_static()
+        },
+
+        #' @description
+        #' Whether the table referenced by this TableHandle is static or not.
+        #' @return TRUE if the table is static, FALSE if the table is ticking.
```
chipkent, line 43:
```suggestion
        #' @return TRUE if the table is static, or FALSE if the table is ticking.
```

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -35,6 +35,24 @@ TableHandle <- R6Class("TableHandle",
...
+
+        #' @description
+        #' Number of rows in the table referenced by this TableHandle. This value is only well-defined for static tables.
+        #' @return The number of rows in the table, if the table is static.
+        num_rows = function() {
+            if(!private$is_static_field) {
+                stop("The table referenced by this TableHandle is not a static table, so the number of rows is not well-defined.")
+            }
+            return(private$internal_table_handle$num_rows())
         },
```
chipkent, line 56: The documentation and what this is doing is not what we want.  Viable options are:
1) Rerequest the size when the method is called, and then return the size.
2) Raise an error that the size is not yet supported for dynamic tables.
The whole "not well defined" thing makes no sense.  For dynamic tables, the value may change, but it is well defined at the instant it is collected.
Eventually (1) needs to happen, but for the sake of shipping this, (2) is ok.  Just add a TODO.

R/rdeephaven/man/ClientOptions.Rd
chipkent, line 1: I assume these docs are regenerated properly.

R/rdeephaven/R/table_handle_wrapper.R
chipkent: Also, is it more idiomatic for this to be `num_rows` or `size`.  I think `num_rows` is more clear.  What does dataframe do?

chipkent (changes requested):

R/rdeephaven/R/table_handle_wrapper.R
alexpeters1208: data frame uses typical R styling of `nrow(dataframe)`. We can implement that later. Should this method be named `nrow` to be kept consistent with a future implementation that enables `nrow(TableHandle)`?

chipkent: If it is more idiomatc, I would do `nrow`... but it looks like `nrow` is a function that you would implement that would call whatever this function is called.

alexpeters1208: It is more idiomatic with typical R, but I do not think it's what a user would expect with an R6 class where everything else is done with `$`. In any case, when `f` is implemented for an R6 class, the corresponding method that gets called is also called `f`.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -35,6 +35,24 @@ TableHandle <- R6Class("TableHandle",
...
+        is_static = function() {
+            return(private$is_static_field)
+        },
+
+        #' @description
+        #' Number of rows in the table referenced by this TableHandle, currently only implemented for static tables.
+        #' @return The number of rows in the table.
+        num_rows = function() {
+            if(!private$is_static_field) {
+                stop("The number of rows is not yet supported for non-static tables.")
```
chipkent, line 53:
```suggestion
                stop("The number of rows is not yet supported for dynamic tables.")
```

R/rdeephaven/src/client.cpp
```diff
@@ -24,14 +24,28 @@ class TableHandleWrapper {
...
+
     /**
      * Binds the table referenced by this table handle to a variable on the server called tableName.
      * Without this call, new tables are not accessible from the client.
      * @param tableName Name for the new table on the server.
     */
     void bindToVariable(std::string tableName) {
         internal_tbl_hdl.bindToVariable(tableName);
-    }
+    };
```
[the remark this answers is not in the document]
chipkent:
```suggestion
    }
```

R/rdeephaven/src/client.cpp
```diff
@@ -24,14 +25,28 @@ class TableHandleWrapper {
 
     // TODO: DEEPHAVEN QUERY METHODS WILL GO HERE
 
+    /**
+     * Whether the table was static at the time internal_tbl_hdl was created.
+    */
+    bool isStatic() {
+        return internal_tbl_hdl.isStatic();
+    };
```
chipkent, line 33:
```suggestion
    }
```

R/rdeephaven/src/client.cpp
```diff
@@ -24,14 +25,28 @@ class TableHandleWrapper {
...
+
+    /**
+     * Number of rows in the table at the time internal_tbl_hdl was created.
+    */
+    int64_t numRows() {
+        return internal_tbl_hdl.numRows();
+    };
```
chipkent, line 40:
```suggestion
    }
```

chipkent (changes requested):

chipkent (approved):

PR #4008, opened 2023-06-15
Bump protobuf-java to 3.23.0

devinrsmith:
https://nvd.nist.gov/vuln/detail/CVE-2022-3171
https://nvd.nist.gov/vuln/detail/CVE-2022-3509
https://nvd.nist.gov/vuln/detail/CVE-2022-3510

This might not be sufficient to take care of the CVEs above - it is probably more important to update the compilation protoc versions, which comes from protoc-base image, https://github.com/deephaven/deephaven-base-images/pull/62


devinrsmith:
Actually, I think this is sufficient to fix the problem, as it seems to be the reflection-based API that is updated for CVEs mentioned above.

niloc132 (approved):

PR #4017, opened 2023-06-16
Clean up JS API samples to be easier to use on startup

niloc132:
Also removed unused samples.


bmingles (commented):
Changes look good to me

bmingles (dismissed):

bmingles (approved):

PR #4025, opened 2023-06-17
Change R client build to depend on a preexisting build of the C++ client.

jcferretti:
This depends on a base-images repo merge (for required changes to the build-dependencies.sh script).
This PR should merge first:
https://github.com/deephaven/deephaven-base-images/pull/69


R/rdeephaven/src/Makevars
```diff
@@ -1,64 +1,84 @@
...
+#
+# Note the name of the protobuf library changes to -lprotobuf
+# if the C++ client and its dependent libraries are built with
+#   CMAKE_BUILD_TYPE=Release
+# The value -lprotobufd matches the default build type `Debug`
+#
+DEPENDENCY_LIBS = \
+        -lclient \
+        -lprotobufd \
+        -larrow_flight -larrow -larrow_bundled_dependencies \
```
alexpeters1208, line 31: I had to include the arrow libs as explicit `.a` files because in the previous build, they were also being built as `.so` files, so I needed to point to the static builds explicitly. Are we 100% sure that those libs are _not_ being built as shared libraries in the new build? If they are, the linker will prefer those and our thing will not work.

R/rdeephaven/src/client.cpp
```diff
@@ -157,7 +157,7 @@ class ClientWrapper {
 
         auto wrapper = internal_tbl_hdl_mngr.createFlightWrapper();
         arrow::flight::FlightCallOptions options;
-        wrapper.addAuthHeaders(&options);
+        wrapper.addHeaders(&options);
```
alexpeters1208, line 160: Why this change? Are we positive this is not a breaking change?

chipkent (changes requested):

R/rdeephaven/src/client.cpp
jcferretti: This is not arrow's API, is our DH C++ client's API.  The version on head already changed, the `addAuthHeaders` method doesn't exist anymore, it was renamed `addHeaders` (because it does that, it adds headers, they don't need to be related to auth).

R/rdeephaven/src/Makevars
```diff
@@ -1,64 +1,84 @@
...
-	-labsl_synchronization -labsl_stacktrace -labsl_symbolize -labsl_debugging_internal -labsl_demangle_internal -labsl_graphcycles_internal \
-	-labsl_malloc_internal -labsl_time -labsl_strings -labsl_throw_delegate -labsl_int128 -labsl_strings_internal -labsl_base -labsl_raw_logging_internal \
-	-labsl_log_severity -labsl_spinlock_wait -labsl_civil_time -labsl_time_zone -lssl -lre2 -lcares
+
+#
+# Note the name of the protobuf library changes to -lprotobuf
+# if the C++ client and its dependent libraries are built with
+#   CMAKE_BUILD_TYPE=Release
+# The value -lprotobufd matches the default build type `Debug`
+#
```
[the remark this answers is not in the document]
jcferretti: This has not changed, I am just making you aware of this in a comment.  The default build type is Debug, and that's how things have been built with the `build-dependencies.sh` script.  When we do the CRAN packaging we can set environment variables to change that type and this would have to change.

chipkent (approved):

PR #4033, opened 2023-06-20
Use an ExecutionContext with an AuthContext during DeephavenApiServer#run

nbauernfeind:
This will need to be included in a point release for DHE.

### DIFF
--- a/server/src/main/java/io/deephaven/server/runner/DeephavenApiServer.java
+++ b/server/src/main/java/io/deephaven/server/runner/DeephavenApiServer.java
@@ -126,25 +126,23 @@ public class DeephavenApiServer {
         AbstractScriptSession.createScriptCache();
 
         log.info().append("Initializing Script Session...").endl();
-
-        final ScriptSession scriptSession = scriptSessionProvider.get();
+        scriptSessionProvider.get();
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/deephaven/server/runner/DeephavenApiServer.java
```diff
@@ -127,20 +127,21 @@ public DeephavenApiServer run() throws IOException, ClassNotFoundException, Time
...
-        scriptSessionProvider.get();
+        final ScriptSession scriptSession = scriptSessionProvider.get();
         pluginRegistration.registerAll();
+        final ExecutionContext executionContext = scriptSession.getExecutionContext();
 
         log.info().append("Starting UpdateGraph...").endl();
         ug.<PeriodicUpdateGraph>cast().start();
 
-        try (final SafeCloseable ignored = ExecutionContext.getContext().withUpdateGraph(ug).open()) {
+        try (final SafeCloseable ignored = executionContext.open()) {
```
rcaudy, line 137: Why don't we install that context for this thread and leave it that way?

rcaudy (approved):

PR #4038, opened 2023-06-21
Ignore parentheses when check vectorizability

### LINKED ISSUE #4036: Parentheses around a Python UDF call render the result as a String
**Description**

```from deephaven import empty_table
import numpy as np

def sinc(x) -> np.double:
    return np.sinc(x)

sinc_function = empty_table(100).update(["X = 0.1 * i", "SincX = sinc(X)", "SincXS=(sinc(X))"])```

![image](https://github.com/deephaven/deephaven-core/assets/29706529/b5aad5e9-a1af-4abe-92fe-9f03edaf3cdc)

**Expected results**

The parenthesis do not change the return type.

**Actual results**

The UI shows it as a string.

**Versions**
 - Deephaven: 0.25.2

### PR DESCRIPTION
**Pull request by jmao-denver**
Fixes #4036

### DIFF
--- a/py/server/tests/test_vectorization.py
+++ b/py/server/tests/test_vectorization.py
@@ -251,6 +251,13 @@ class VectorizationTestCase(BaseTestCase):
         result = source.update(f"X = my_sum({','.join(cols)})")
         self.assertEqual(len(cols) + 1, len(result.columns))
 
+    def test_enclosed_by_parentheses(self):
+        def sinc(x) -> np.double:
+            return np.sinc(x)
+
+        t = empty_table(100).update(["X = 0.1 * i", "SincXS=((sinc(X)))"])
+        self.assertEqual(t.columns[0].data_type, dtypes.double)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/py/server/tests/test_vectorization.py
+++ b/py/server/tests/test_vectorization.py
@@ -256,7 +256,14 @@ class VectorizationTestCase(BaseTestCase):
             return np.sinc(x)
 
         t = empty_table(100).update(["X = 0.1 * i", "SincXS=((sinc(X)))"])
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by cpwright — DISMISSED** [review]

**Review by rcaudy — APPROVED** [review]
LGTM. Why did the original version return a String?

File: py/server/tests/test_vectorization.py [hunk 1]
```diff
@@ -251,6 +251,13 @@ def my_sum(*args):
         result = source.update(f"X = my_sum({','.join(cols)})")
         self.assertEqual(len(cols) + 1, len(result.columns))
 
+    def test_enclosed_by_parentheses(self):
+        def sinc(x) -> np.double:
+            return np.sinc(x)
+
+        t = empty_table(100).update(["X = 0.1 * i", "SincXS=((sinc(X)))"])
+        self.assertEqual(t.columns[0].data_type, dtypes.double)
```
chipkent, line 259:
```suggestion
        def sinc(x) -> np.double:
            return np.sinc(x)

        t = empty_table(100).update(["X = 0.1 * i", "SincXS=((sinc(X)))"])
        self.assertEqual(t.columns[0].data_type, dtypes.double)
        
       def sinc2(x):
            return np.sinc(x)

        t = empty_table(100).update(["X = 0.1 * i", "SincXS=((sinc2(X)))"])
        self.assertEqual(t.columns[0].data_type, ***PyObject)

```

Should have a second test that non-annoted functions return PyObjects and not java Objects.

chipkent: Does there need to be any test to make sure that these two cases still vectorized?  I assume yes.

chipkent (changes requested):

rcaudy (approved):

rcaudy (approved):

chipkent (approved):

PR #4042, opened 2023-06-21
GRPC: Allow string duration parameters for UpdateBy ema/rolling windows

lbooker42:
This PR modifies the GRPC protobuf file to allow specification of duration strings for `update_by` operator time windows.  This adds functionality while preserving the ability to provide durations in nanoseconds directly.

Additional tests for the python client have been added.


proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -381,7 +381,10 @@ message UpdateByWindowScale {
   }
   message UpdateByWindowTime {
     string column = 1;
-    sint64 period_nanos = 2 [jstype=JS_STRING];
+    oneof window {
+      sint64 nanos = 2 [jstype = JS_STRING];
+      string duration_string = 3;
```
devinrsmith, line 386: Do we want / need to add documentation here about this string format? `DateTimeUtils.parseDurationNanos`?

lbooker42: I updated the function signatures to reflect that a time string is also allowed.  The current description is fairly vague however:
`a time interval string, e.g. "PT00:00:00.001"`

devinrsmith:
I've claimed this is _not_ a breaking change (old clients should be able to successfully talk to newer servers). @niloc132 should be able to theoretically confirm. That said, it might be nice to run a manual test.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
chipkent: The overall docs should cover the formatting of the string, so it shouldn't be as critical to have it in every location.  Also having it in every location may be more brittle, if we support more formats in the future.

py/client/tests/test_updateby.py
```diff
@@ -35,12 +35,15 @@ def test_ema(self):
...
-                              op_control=op_ctrl),
-               ]
+        ops = [
+            ema_tick_decay(decay_ticks=100, cols="ema_a = a"),
+            ema_tick_decay(decay_ticks=100, cols="ema_a = a", op_control=op_ctrl),
+            ema_time_decay(ts_col="Timestamp", decay_time=10, cols="ema_a = a"),
+            ema_time_decay(ts_col="Timestamp", decay_time=1000000, cols=["ema_c = c"], op_control=op_ctrl),
+            ema_time_decay(ts_col="Timestamp", decay_time="00:00:00.001", cols="ema_c = c", op_control=op_ctrl),
+            ema_time_decay(ts_col="Timestamp", decay_time="PT00:00:00.001", cols="ema_c = c", op_control=op_ctrl),
+            ema_time_decay(ts_col="Timestamp", decay_time="PT1M", cols="ema_c = c", op_control=op_ctrl),
```
chipkent, line 45: output col names are being reused

server/src/main/java/io/deephaven/server/table/ops/UpdateByGrpcImpl.java
```diff
@@ -349,7 +350,10 @@ private static WindowScale adaptWindowScale(UpdateByWindowScale timescale) {
             case TICKS:
                 return WindowScale.ofTicks(timescale.getTicks().getTicks());
             case TIME:
-                return WindowScale.ofTime(timescale.getTime().getColumn(), timescale.getTime().getPeriodNanos());
+                return timescale.getTime().hasDurationString()
```
chipkent, line 353: Do there need to be any checks to see if both the integer and string values have been set, or does setting one eliminate the other?

chipkent (changes requested):

server/src/main/java/io/deephaven/server/table/ops/UpdateByGrpcImpl.java
lbooker42: By specifying `oneof`, the `.proto` definition eliminates any chance of setting both.  Per the documentation:
```
Setting any member of the oneof automatically clears all the other members.
```

py/client/tests/test_updateby.py
lbooker42: Yes, but the python tests iterate over these operations (vs. combining into a single `update_by` call) so the names don't collide in the output.

Perhaps more combined operator testing is needed, will add an additional test composed of different types of operators.

jmao-denver (approved):
LGTM

nbauernfeind:
LGTM, too.

rcaudy (approved):

Issue #4005, opened 2023-06-15
Flight client doesn't publish Boolean column


ilyaberdnikov:
**Description**

A small modification of `AggInputBase` fails with an error.

**Steps to reproduce**
1. Apply patch
```
diff --git a/java-client/flight-examples/src/main/java/io/deephaven/client/examples/AggInputBase.java b/java-client/flight-examples/src/main/java/io/deephaven/client/examples/AggInputBase.java
index 6c882ef15..0684aa668 100644
--- a/java-client/flight-examples/src/main/java/io/deephaven/client/examples/AggInputBase.java
+++ b/java-client/flight-examples/src/main/java/io/deephaven/client/examples/AggInputBase.java
@@ -8,7 +8,7 @@ import io.deephaven.client.impl.TableHandle;
 import io.deephaven.qst.LabeledValue;
 import io.deephaven.qst.LabeledValues;
 import io.deephaven.qst.column.header.ColumnHeader;
-import io.deephaven.qst.column.header.ColumnHeaders8;
+import io.deephaven.qst.column.header.ColumnHeaders9;
 import io.deephaven.qst.table.InMemoryKeyBackedInputTable;
 import io.deephaven.qst.table.LabeledTable;
 import io.deephaven.qst.table.LabeledTables;
@@ -50,6 +50,7 @@ abstract class AggInputBase extends FlightExampleBase {
     protected void execute(FlightSession flight) throws Exception {
         final ColumnHeader<Integer> inputKey = INPUT_KEY;
         final ColumnHeader<Integer> groupKey = GROUP_KEY;
+        final ColumnHeader<Boolean> b = ColumnHeader.ofBoolean("B");
         final ColumnHeader<Byte> u = ColumnHeader.ofByte("U");
         final ColumnHeader<Short> v = ColumnHeader.ofShort("V");
         final ColumnHeader<Integer> w = ColumnHeader.ofInt("W");
@@ -59,7 +60,7 @@ abstract class AggInputBase extends FlightExampleBase {
 
         final Builder builder = LabeledTables.builder();
 
-        final TableSpec spec = InMemoryKeyBackedInputTable.of(TableHeader.of(inputKey, groupKey, u, v, w, x, y, z),
+        final TableSpec spec = InMemoryKeyBackedInputTable.of(TableHeader.of(inputKey, groupKey, b, u, v, w, x, y, z),
                 Collections.singletonList(inputKey.name()));
         builder.putMap("base", spec);
[...]

PR #4047, opened 2023-06-21
Address reported bug in Flight Boolean Columns

lbooker42:
Fixes reported bug #4005

### DIFF
--- a/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
+++ b/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
@@ -452,6 +452,8 @@ public class BarrageUtil {
                 // TODO (deephaven-core#3403) widen this check for better assurances
                 Class<?> defaultType = getDefaultType(getArrowType.apply(i), result, i);
                 Assert.eq(type.getValue(), "deephaven column type", defaultType, "arrow inferred type");
+                // force to boxed boolean to allow nullability in the column sources
+                type.setValue(Boolean.class);
             }
             columns[i] = ColumnDefinition.fromGenericType(name, type.getValue(), componentType.getValue());
         }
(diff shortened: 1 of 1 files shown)


extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
```diff
@@ -452,6 +452,8 @@ private static ConvertedArrowSchema convertArrowSchema(
                 // TODO (deephaven-core#3403) widen this check for better assurances
```
nbauernfeind, line 452: We need to change the `boolean.class` on line 450, too.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
```diff
@@ -452,6 +452,8 @@ private static ConvertedArrowSchema convertArrowSchema(
...
                 Class<?> defaultType = getDefaultType(getArrowType.apply(i), result, i);
                 Assert.eq(type.getValue(), "deephaven column type", defaultType, "arrow inferred type");
```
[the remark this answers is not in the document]
nbauernfeind: Colin and I would prefer to, eventually, create more assurances that the metadata types match the incoming arrow data type. Otherwise, there are either silent failures with bad parsed data (when the types are the same size) or we throw exceptions when parsing with the mismatches. So, let's keep the check.

I do agree we should change `getDefaultType` to return `Boolean.class` instead of `boolean.class`.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
nbauernfeind: This request is wrong; I now think we should check for both `boolean.class` and `Boolean.class`. This is validating that booleans are being sent using arrow's proper wire format. In the code I can see that both, as deephaven column types, use the same format, so we might as well.

nbauernfeind (approved):

PR #4048, opened 2023-06-21
Enable build parallelism by default.

devinrsmith:
This also paves the way for larger CI runner support.

### DIFF
--- a/.github/scripts/print-gradle-workers-max.sh
+++ b/.github/scripts/print-gradle-workers-max.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+# We know the gradle daemon takes 1GiB (see gradle.properties)
+DAEMON_BYTES=1073741824
+
+# We'll leave some buffer space for other system resources / processes
+OTHER_BYTES=2147483648
+
+TOTAL_SYSTEM_BYTES="$(free --bytes | grep Mem | awk -F " " '{print $2}')"
+
+# This is accounting for "worst case", assuming every single worker is using the theoretical maximum.
+# Currently, engine/table/build.gradle sets a heap size of 6GiB, so that's the maximum.
+PER_WORKER_BYTES=6442450944
+
+MAX_WORKERS="$(( (TOTAL_SYSTEM_BYTES - DAEMON_BYTES - OTHER_BYTES) / PER_WORKER_BYTES ))"
+MAX_WORKERS="$(( MAX_WORKERS > 0 ? MAX_WORKERS : 1 ))"
+
+echo "org.gradle.workers.max=${MAX_WORKERS}"
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -32,6 +32,8 @@ jobs:
         run: |
           cat .github/env/${{ runner.os }}/gradle.properties >> gradle.properties
           echo >> gradle.properties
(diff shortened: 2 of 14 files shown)


buildSrc/src/main/groovy/io.deephaven.java-test-conventions.gradle
```diff
@@ -45,10 +54,6 @@ project.tasks.withType(Test).all { Test t ->
             maxHeapSize = '3g'
         }
 
-        if (!maxParallelForks) {
-            maxParallelForks = 4
-        }
```
devinrsmith, line 50: This logic never worked; the inner code is never executed.

.github/scripts/print-gradle-workers-max.sh
```diff
@@ -0,0 +1,22 @@
...
+set -o pipefail
+set -o nounset
+
+# We know the gradle daemon takes 1GiB (see gradle.properties)
+DAEMON_BYTES=1073741824
+
+# We'll leave some buffer space for other system resources / processes
+OTHER_BYTES=2147483648
+
+TOTAL_SYSTEM_BYTES="$(free --bytes | grep Mem | awk -F " " '{print $2}')"
```
rcaudy, line 13: Where does `free` come from? I don't have it locally as a shell built-in or in my path.

devinrsmith: It's a very common Linux command line utility.

https://man7.org/linux/man-pages/man1/free.1.html

https://gitlab.com/procps-ng/procps/-/blob/master/src/free.c

.github/scripts/print-gradle-workers-max.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# This is accounting for "worst case", assuming every single worker is using the theoretical maximum.
+# Currently, engine/table/build.gradle sets a heap size of 6GiB, so that's the maximum.
+PER_WORKER_BYTES=6442450944
+
+MAX_WORKERS="$(( (TOTAL_SYSTEM_BYTES - DAEMON_BYTES - OTHER_BYTES) / PER_WORKER_BYTES ))"
```
jcferretti, line 19: This is memory right?
I get that we use this just below, to echo it as a property value.
If we are not going to call this MAX_WORKER_MEM_BYTES, can we at least call it WORKERS_MAX, like the property is blah.blah.blah.workers.max below?
I imagine we didn't pick that property name, so be it, but at least lets not add another permutation on a confusing name...

.github/workflows/build-ci.yml
```diff
@@ -32,6 +32,8 @@ jobs:
         run: |
           cat .github/env/${{ runner.os }}/gradle.properties >> gradle.properties
           echo >> gradle.properties
+          .github/scripts/print-gradle-workers-max.sh >> gradle.properties
+          echo >> gradle.properties
```
jcferretti, line 36: The block below is almost the same.  Too hard to factor the common parts out?

jcferretti (approved):

.github/scripts/print-gradle-workers-max.sh
devinrsmith: The unit is "units", not "memory". Unit analyses, the formula is:  (bytes) / (bytes / worker unit) -> worker unit

.github/workflows/build-ci.yml
devinrsmith: I've taken your advice and consolidated it down to a single script.

jcferretti (approved):

mofojed (approved):

PR #4060, opened 2023-06-23
Fix accidental double-freeing of log entries and fix LogEntry API

### LINKED ISSUE #4051: LogEntry NPE
```
java.lang.NullPointerException: Cannot invoke "java.nio.ByteBuffer.remaining()" because "this.buf" is null
   at io.deephaven.io.streams.ByteBufferOutputStream.appendBytes(ByteBufferOutputStream.java:365)
   at io.deephaven.io.streams.ByteBufferOutputStream.appendBytes(ByteBufferOutputStream.java:360)
   at io.deephaven.io.log.impl.LogOutputCsvImpl.append(LogOutputCsvImpl.java:125)
   at io.deephaven.io.log.impl.LogEntryImpl.append(LogEntryImpl.java:191)
   at io.deephaven.kafka.ingest.KafkaIngester.consumerLoop(KafkaIngester.java:300)
   at java.base/java.lang.Thread.run(Thread.java:833)
```

This seems like a rare race in the code. Investigating.

devinrsmith:
This logging code was accidentally misusing the LogEntry API, calling endl() multiple times. This resulted in the same entry being put into the pool multiple time. Later on, the entry could be used by multiple threads logging at the same time. Most of the time this would result in garbage logs, but occasionally it would result in an NPE.
  
This also breaks LogEntry binary and (potentially) source compatibility to ensure users can't string together fluent end()/endl().
    
Note: it's still possible for callers to double-free
    
```java
le = log.info();
le.append("hello");
le.append("world");
le.endl();
le.endl();
```

but this is a much rarer pattern. Adding additional runtime safety checks would be possible, but would not be free.

Fixes #4051

### DIFF
--- a/IO/src/main/java/io/deephaven/io/log/LogEntry.java
+++ b/IO/src/main/java/io/deephaven/io/log/LogEntry.java
@@ -20,8 +20,16 @@ public interface LogEntry extends LogOutput, LogSink.Element {
 
     LogEntry start(LogSink sink, LogLevel level, long currentTimeMicros, Throwable t);
 
+    /**
+     * Completes the log entry. Callers should not use {@code this} after completion. End or {@link #endl()} should be
+     * called exactly once.
+     */
     void end();
 
+    /**
+     * Completes the log entry with a newline. Callers should not use {@code this} after completion. Endl or
+     * {@link #end()} should be called exactly once.
+     */
     void endl();
 
     LogEntry append(boolean b);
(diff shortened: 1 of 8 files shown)


IO/src/main/java/io/deephaven/io/log/LogEntry.java
```diff
@@ -20,9 +20,9 @@ public interface LogEntry extends LogOutput, LogSink.Element {
 
     LogEntry start(LogSink sink, LogLevel level, long currentTimeMicros, Throwable t);
 
-    LogEntry end();
+    void end();
```
rcaudy, line 23: This is a very clever solution. You aren't adding anything to *prevent* double frees, if users really want, however. I think you should at least add JavaDoc to the effect that `end()` or `endl()` must be invoked exactly once.

rcaudy (commented):
Very clever

rcaudy (approved):

PR #4065, opened 2023-06-23
Cleanup and improve some logging code.

devinrsmith:
Improved as part of the investigation into #4051

### DIFF
--- a/IO/src/main/java/io/deephaven/io/streams/ByteBufferOutputStream.java
+++ b/IO/src/main/java/io/deephaven/io/streams/ByteBufferOutputStream.java
@@ -226,7 +226,6 @@ public class ByteBufferOutputStream extends java.io.OutputStream implements Data
     }
 
     private void putUtf8(int c) {
-        // todo: is this better if we have char c?
         if (c <= 0x7f) {
             if (c == 0) {
                 buf.put((byte) 0xC0);
--- a/IO/src/main/java/io/deephaven/io/streams/ByteBufferOutputStream.java
+++ b/IO/src/main/java/io/deephaven/io/streams/ByteBufferOutputStream.java
@@ -9,6 +9,7 @@ import java.nio.ByteBuffer;
 import java.io.IOException;
 import java.io.DataOutput;
 import java.io.UTFDataFormatException;
+import java.util.Objects;
 
 // TODO: this implementation is not suited for application which need to fill the underlying
 // TODO: buffers to their full capacity, because it calculates (or, for CSV appenders, estimates)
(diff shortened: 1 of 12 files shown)


IO/src/main/java/io/deephaven/io/streams/ByteBufferOutputStream.java
```diff
@@ -128,65 +124,49 @@ public void write(ByteBuffer b) throws IOException {
...
-        }
+        ensureRemaining(Long.BYTES);
         buf.putLong(v);
     }
 
     @Override
     public void writeFloat(float f) throws IOException {
-        if (buf.remaining() < 8) {
-            buf = sink.acceptBuffer(buf, 8);
-        }
```
devinrsmith, line 181: While this code works, it's a bit sloppier than using 4 (Float.BYTES).

IO/src/main/java/io/deephaven/io/streams/ByteBufferOutputStream.java
```diff
@@ -283,51 +249,45 @@ public void writeUTF(CharSequence str) throws IOException {
...
     public ByteBufferOutputStream appendByteBuffer(ByteBuffer bb) throws IOException {
-        if (buf.remaining() < bb.remaining()) {
-            buf = sink.acceptBuffer(buf, bb.remaining());
+        final int origPos = bb.position();
+        final int origLimit = bb.limit();
+        int remaining;
+        while ((remaining = buf.remaining()) < bb.remaining()) {
+            buf.put(bb.limit(bb.position() + remaining));
+            bb.limit(origLimit);
+            updateBuffer(buf.capacity());
```
rcaudy, line 258: Is this right? Or do we want `bb.remaining()`? I guess that could blow up if we ask for too much.

IO/src/main/java/io/deephaven/io/streams/ByteBufferOutputStream.java
```diff
@@ -228,53 +208,39 @@ public void writeUTF(CharSequence str) throws IOException {
...
-                }
-            } else if (c <= 0x7ff) {
-                buf.put((byte) (0xc0 | (0x1f & (c >> 6))));
-                buf.put((byte) (0x80 | (0x3f & c)));
+            putUtf8(str.charAt(position++));
+        }
+    }
+
+    private void putUtf8(int c) {
+        // todo: is this better if we have char c?
```
rcaudy, line 229: TODO....

IO/src/main/java/io/deephaven/io/streams/ByteBufferOutputStream.java
rcaudy: Why would we want to reserve 8 bytes for a 4 byte thing? Seems to me like your version is just plain better.

IO/src/main/java/io/deephaven/io/streams/ByteBufferOutputStream.java
devinrsmith: The whole idea is that we want to chunk it out instead of requesting huge buffers... the code I'm trying to update here asks for one big buffer, while the rest of this class takes the streaming approach (that I've updated this code path to here).

IO/src/main/java/io/deephaven/io/streams/ByteBufferOutputStream.java
devinrsmith: Yes, just noting that technically the logic has changed here.

rcaudy (approved):

Issue #4011, opened 2023-06-15
head_pct and tail_pct filters should validate parameter value


mattrunyon:
As a user, I want head_pct and tail_pct to error when given a value less than 0 or greater than 1 so that I get told when I'm using it incorrectly. 

Based on the name of the function, I would not expect it to be 0-1. I'm not sure what a better name would be, but because it has pct (percent) in the name, I expected `head_pct(10)` to be 10 percent while `head_pct(0.10)` is 0.10 percent

I also noticed you can provide a negative value and it just absolutes it. `head_pct(-0.5) == head_pct(0.5)`

PR #4066, opened 2023-06-23
Added validation for parameters in head_pct and tail_pct (#4011)

malhotrashivam:
Fixes the bug mentioned in #4011

**Change summary:** After this change, the engine will generate an IllegalArgumentException if arguments for head_pct or tail_pct have a value outside [0..1]. Added some unit tests for the change and manually tested it using Java and Groovy UI.

**Documentation Update:** We are requesting an update in the documentation for the [head_pct](https://deephaven.io/core/docs/reference/table-operations/filter/head-pct/) and [tail_pct](https://deephaven.io/core/docs/reference/table-operations/filter/tail-pct/) methods.
The documentation should clarify that the input parameter ("pct" for python and "percent" for Groovy) for the head_pct and tail_pct methods is a floating point number between [0,1]. Any values outside this range will generate an error.

### DIFF
--- a/engine/api/src/main/java/io/deephaven/engine/table/Table.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/Table.java
@@ -366,9 +366,9 @@ public interface Table extends
     /**
      * Provides a head that selects a dynamic number of rows based on a percent.
      *
-     * @param percent the fraction of the table to return (0..1), the number of rows will be rounded up. For example if
-     *        there are 3 rows, headPct(50) returns the first two rows. For percent values outside (0..1), the function
-     *        will throw an exception.
+     * @param percent the fraction of the table to return between [0, 1]. The number of rows will be rounded up. For
+     *        example if there are 3 rows, headPct(50) returns the first two rows. For percent values outside [0, 1],
+     *        the function will throw an exception.
      */
     @ConcurrentMethod
     Table headPct(double percent);
@@ -376,9 +376,9 @@ public interface Table extends
     /**
      * Provides a tail that selects a dynamic number of rows based on a percent.
      *
-     * @param percent the fraction of the table to return (0..1), the number of rows will be rounded up. For example if
-     *        there are 3 rows, tailPct(50) returns the last two rows. For percent values outside (0..1), the function
-     *        will throw an exception.
+     * @param percent the fraction of the table to return between [0, 1]. The number of rows will be rounded up. For
+     *        example if there are 3 rows, tailPct(50) returns the last two rows. For percent values outside [0, 1], the
+     *        function will throw an exception.
      */
     @ConcurrentMethod
     Table tailPct(double percent);
--- a/engine/api/src/main/java/io/deephaven/engine/table/Table.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/Table.java
@@ -367,11 +367,19 @@ public interface Table extends
      * Provides a head that selects a dynamic number of rows based on a percent.
      *
      * @param percent the fraction of the table to return (0..1), the number of rows will be rounded up. For example if
-     *        there are 3 rows, headPct(50) returns the first two rows.
+     *        there are 3 rows, headPct(50) returns the first two rows. For percent values outside (0..1), the function
+     *        will throw an exception.
      */
     @ConcurrentMethod
     Table headPct(double percent);
 
+    /**
+     * Provides a tail that selects a dynamic number of rows based on a percent.
+     *
+     * @param percent the fraction of the table to return (0..1), the number of rows will be rounded up. For example if
+     *        there are 3 rows, tailPct(50) returns the last two rows. For percent values outside (0..1), the function
+     *        will throw an exception.
+     */
     @ConcurrentMethod
     Table tailPct(double percent);
 
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
@@ -638,7 +638,7 @@ public class QueryTable extends BaseTable<QueryTable> {
     public Table headPct(final double percent) {
         if (percent < 0 || percent > 1) {
             throw new IllegalArgumentException(
-                    "headPct percentage of rows must be between [0,1]: : percent=" + percent);
+                    "percentage of rows must be between [0,1]: percent=" + percent);
         }
         final UpdateGraph updateGraph = getUpdateGraph();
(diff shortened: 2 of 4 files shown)


malhotrashivam:

engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -367,11 +367,19 @@ public interface Table extends
...
     Table headPct(double percent);
 
+    /**
+     * Provides a tail that selects a dynamic number of rows based on a percent.
+     *
+     * @param percent the fraction of the table to return (0..1), the number of rows will be rounded up. For example if
+     *        there are 3 rows, tailPct(50) returns the last two rows. For percent values outside (0..1), the function
+     *        will throw an exception.
+     */
     @ConcurrentMethod
```
malhotrashivam, line 383: Do I need to do anything else for updating the documentation?

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -636,6 +636,10 @@ public Table tail(final long size) {
 
     @Override
     public Table headPct(final double percent) {
+        if (percent < 0 || percent > 1) {
+            throw new IllegalArgumentException(
+                    "headPct percentage provided should be between [0,1]: percent=" + percent);
```
lbooker42, line 641: This function can be called from python and will have a different parameter name after mapping.  Use the core concept rather than the parameter name in the error.  I would also remove the word `provided` and go with something like this:

`percentage of rows must be between [0,1]: percent=`

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -644,6 +648,10 @@ public Table headPct(final double percent) {
 
     @Override
     public Table tailPct(final double percent) {
+        if (percent < 0 || percent > 1) {
+            throw new IllegalArgumentException(
+                    "tailPct percentage provided should be between [0,1]: percent=" + percent);
```
lbooker42, line 653: Same comment as above.

engine/table/src/main/java/io/deephaven/engine/table/impl/SliceLikeOperation.java
```diff
@@ -36,6 +36,7 @@ public static SliceLikeOperation headPct(final QueryTable parent, final double p
                 0, 0, true) {
             @Override
             protected long getLastPositionExclusive() {
+                // Assuming percent is not negative here
```
lbooker42, line 39: We aren't assuming, we know.  Suggest:
`// Already verified that percent is not negative`

engine/table/src/main/java/io/deephaven/engine/table/impl/SliceLikeOperation.java
```diff
@@ -46,6 +47,7 @@ public static SliceLikeOperation tailPct(final QueryTable parent, final double p
                 0, 0, false) {
             @Override
             protected long getFirstPositionInclusive() {
+                // Assuming percent is not negative here
```
lbooker42, line 50: repeat

engine/table/src/test/java/io/deephaven/engine/table/impl/QueryTableSliceTest.java
```diff
@@ -442,6 +443,28 @@ public void testHeadTailPct() {
                 TstUtils.testRefreshingTable(i(2).toTracking(), col("x", 1), col("y", 'a')));
         assertTableEquals(table.tailPct(0.1),
                 TstUtils.testRefreshingTable(i(6).toTracking(), col("x", 3), col("y", 'c')));
+
```
lbooker42, line 446: I would add tests at 0,1 as well.  If we decide inclusive, they should pass and have either 0 rows or be the same size as the input table.  If exclusive, should result in exceptions.

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
lbooker42: When we achieve clarity on the interval, make sure the interval in the exception matches (i.e. this might be `(0,1)`)

engine/table/src/main/java/io/deephaven/engine/table/impl/SliceLikeOperation.java
malhotrashivam: Done

malhotrashivam:
We are requesting an update in the documentation for the [head_pct](https://deephaven.io/core/docs/reference/table-operations/filter/head-pct/ ) and [tail_pct](https://deephaven.io/core/docs/reference/table-operations/filter/tail-pct/) methods.

The documentation should clarify that the input parameter ("pct" for python and "percent" for Groovy) for both head_pct and tail_pct is a floating point number between [0,1]. Any values outside this range will generate an error.

lbooker42:
Your comment about the requested change to the documentation should be moved (or copied) into the PR summary.  It would be easy to overlook as a tailing comment.

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -638,7 +638,7 @@ public Table tail(final long size) {
     public Table headPct(final double percent) {
         if (percent < 0 || percent > 1) {
             throw new IllegalArgumentException(
-                    "headPct percentage provided should be between [0,1]: percent=" + percent);
+                    "headPct percentage of rows must be between [0,1]: : percent=" + percent);
```
lbooker42, line 641: The Java parameter name is not needed and will be confusing for python users.
```suggestion
                    "percentage of rows must be between [0,1]: : percent=" + percent);
```

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -650,7 +650,7 @@ public Table headPct(final double percent) {
     public Table tailPct(final double percent) {
         if (percent < 0 || percent > 1) {
             throw new IllegalArgumentException(
-                    "tailPct percentage provided should be between [0,1]: percent=" + percent);
+                    "tailPct percentage of rows must be between [0,1]: percent=" + percent);
```
lbooker42, line 653: Repeat:
```suggestion
                    "percentage of rows must be between [0,1]: percent=" + percent);
```

lbooker42 (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2837
Reference: https://github.com/deephaven/deephaven.io/issues/2838

PR #4073, opened 2023-06-26
Add sortable column check

### LINKED ISSUE #1914: Error when sorting `StringSet` in the UI
`StringSet`s causes an error when it is sorted in the UI.  If we can't sort it, we shouldn't let users try to sort it.

This query can be used to illustrate the problem:
```python

ArrayStringSet = jpy.get_type("io.deephaven.stringset.ArrayStringSet")

values = ["A", "B", "C"]

def to_string_set() -> ArrayStringSet:
    """ Converts an iterable to a string set. """
    # print("HERE")
    return ArrayStringSet(values)


# t = emptyTable(10).update("SS = (io.deephaven.stringset.StringSet)to_string_set()")
# m = t.getMeta()

import threading
from time import sleep
import deephaven.Types as dht
from deephaven import DynamicTableWriter
from deephaven.conversion_utils import NULL_DOUBLE

dtw = DynamicTableWriter(["A","B"],[dht.int32, dht.stringset])

def populate():
    for i in range(1000):
        dtw.logRow([i, to_string_set()])

thread = threading.Thread(target=populate)
thread.start()

t1 = dtw.getTable()
t2 = t1.lastBy("A")
t3 = t1.naturalJoin(t2,"A", "C=B")
```

You get an error like:
```
heduler-Concurrent-4 | i.d.s.s.SessionState      | Internal Error '2fa7ff88-ec61-43d7-9588-ecf2a74af77b' io.deephaven.base.verify.RequirementFailure: Requirement failed: required Comparable.class.isAssignableFrom(sortColumns[ii].getType()) || sortColumns[ii].getType().isPrimitive(), instead sortColumnNames[ii] == "B", sortColumns[ii].getType() == class io.deephaven.engine.table.impl.preview.DisplayWrapper.
        at io.deephaven.base.verify.Require.requirement(Require.java:169)
        at io.deephaven.base.verify.Require.requirement(Require.java:175)
        at io.deephaven.engine.table.impl.SortOperation.<init>(SortOperation.java:54)
        at io.deephaven.engine.table.impl.QueryTable.sort(QueryTable.java:2369)
        at io.deephaven.server.table.ops.SortTableGrpcImpl.create(SortTableGrpcImpl.java:98)
        at io.deephaven.server.table.ops.SortTableGrpcImpl.create(SortTableGrpcImpl.java:21)
        at io.deephaven.server.table.TableServiceGrpcImpl$BatchExportBuilder.doExport(TableServiceGrpcImpl.java:461)
        at io.deephaven.server.table.TableServiceGrpcImpl.lambda$batch$5(TableServiceGrpcImpl.java:313)
        at io.deephaven.server.session.SessionState$ExportObject.doExport(SessionState.java:847)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at io.deephaven.server.runner.DeephavenApiServerModule$ThreadFactory.lambda$newThread$0(DeephavenApiServerModule.java:143)
```

ethanalvizo:
Closes #1914 

Implementing this same feature on the enterprise side has been separated into its own ticket:
https://deephaven.atlassian.net/browse/DH-9189

### DIFF
--- a/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
+++ b/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
@@ -158,20 +158,23 @@ public class BarrageUtil {
     }
 
     @NotNull
-    public static Map<String, String> attributesToMetadata(@NotNull final Map<String, Object> attributes, @NotNull final Collection<ColumnDefinition<?>> columnDefinitions) {
+    public static Map<String, String> attributesToMetadata(@NotNull final Map<String, Object> attributes,
+            @NotNull final Collection<ColumnDefinition<?>> columnDefinitions) {
         final Map<String, String> metadata = new HashMap<>();
 
         List<String> sortableColumns = new ArrayList<>();
         List<ColumnDefinition<?>> columnsToCheck = new ArrayList<>(columnDefinitions);
 
-        if(attributes.containsKey(GridAttributes.SORTABLE_COLUMNS_ATTRIBUTE)) {
-            String[] restrictedSortColumns = attributes.get(GridAttributes.SORTABLE_COLUMNS_ATTRIBUTE).toString().split(",");
-            columnsToCheck.removeIf(col -> Arrays.stream(restrictedSortColumns).noneMatch(name -> name.equals(col.getName())));
+        if (attributes.containsKey(GridAttributes.SORTABLE_COLUMNS_ATTRIBUTE)) {
+            String[] restrictedSortColumns =
+                    attributes.get(GridAttributes.SORTABLE_COLUMNS_ATTRIBUTE).toString().split(",");
+            columnsToCheck.removeIf(
+                    col -> Arrays.stream(restrictedSortColumns).noneMatch(name -> name.equals(col.getName())));
         }
 
         columnsToCheck.forEach(col -> {
             Class<?> dataType = col.getDataType();
-            if(dataType.isPrimitive() || Comparable.class.isAssignableFrom(dataType)) {
+            if (dataType.isPrimitive() || Comparable.class.isAssignableFrom(dataType)) {
                 sortableColumns.add(col.getName());
             }
         });
@@ -231,7 +234,6 @@ public class BarrageUtil {
             final Map<String, String> metadata = fieldMetadataFactory.apply(name);
 
             putMetadata(metadata, "isPartitioning", column.isPartitioning() + "");
-            putMetadata(metadata, "isSortable", (dataType.isPrimitive() || Comparable.class.isAssignableFrom(dataType)) + "");
 
             // Wire up style and format column references
             final String styleFormatName = ColumnFormatting.getStyleFormatColumn(name);
--- a/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
+++ b/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
@@ -207,6 +207,7 @@ public class BarrageUtil {
             final Map<String, String> metadata = fieldMetadataFactory.apply(name);
 
             putMetadata(metadata, "isPartitioning", column.isPartitioning() + "");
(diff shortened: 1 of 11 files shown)


mofojed (commented):
Just a minor cleanup, but looks good to me. Still would like someone from Core to approve first.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
```diff
@@ -192,13 +194,33 @@ public static Stream<Field> columnDefinitionsToFields(
...
 
+        // Find columns that are sortable
+        List<String> sortableColumns = new ArrayList<>();
+        List<ColumnDefinition<?>> columnsToCheck = new ArrayList<>(columnDefinitions);
+
+        // limits to restrictedSortTo columns if applicable
+        if (attributes.containsKey(GridAttributes.SORTABLE_COLUMNS_ATTRIBUTE)) {
+            String[] restrictedSortColumns =
+                    attributes.get(GridAttributes.SORTABLE_COLUMNS_ATTRIBUTE).toString().split(",");
+            columnsToCheck.removeIf(
```
nbauernfeind, line 213: This is an O(n^2) check. I am told sometimes users have 1k+ columns.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
```diff
@@ -192,13 +194,33 @@ public static Stream<Field> columnDefinitionsToFields(
...
+                    col -> Arrays.stream(restrictedSortColumns).noneMatch(name -> name.equals(col.getName())));
+        }
+
+        columnsToCheck.forEach(col -> {
+            Class<?> dataType = col.getDataType();
+            if (dataType.isPrimitive() || Comparable.class.isAssignableFrom(dataType)) {
+                sortableColumns.add(col.getName());
```
nbauernfeind, line 220: We shouldn't be adding columns if the sort is restricted, but rather further restricting them.

I'm thinking something like this:
```java
        Set<String> sortableColumns;
        if (attributes.containsKey(GridAttributes.SORTABLE_COLUMNS_ATTRIBUTE)) {
            final String[] restrictedColumns = attributes.get(GridAttributes.SORTABLE_COLUMNS_ATTRIBUTE).toString().split(",");
            sortableColumns = Arrays.stream(restrictedColumns)
                    .filter(columnName -> isDataTypeSortable(tableDefinition.getColumn(columnName).getDataType()))
                    .collect(Collectors.toSet());
        } else {
            sortableColumns = tableDefinition.getColumns().stream()
                    .filter(columnDefinition -> isDataTypeSortable(columnDefinition.getDataType()))
                    .map(ColumnDefinition::getName)
                    .collect(Collectors.toSet());
        }
```

Where:
```java
    private static boolean isDataTypeSortable(final Class<?> dataType) {
        return dataType.isPrimitive() || Comparable.class.isAssignableFrom(dataType);
    }
```

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/BarrageUtil.java
```diff
@@ -207,6 +229,7 @@ public static Stream<Field> columnDefinitionsToFields(
             final Map<String, String> metadata = fieldMetadataFactory.apply(name);
 
             putMetadata(metadata, "isPartitioning", column.isPartitioning() + "");
+            putMetadata(metadata, "isSortable", String.valueOf(sortableColumns.contains(name)));
```
nbauernfeind, line 232: This ends up being an additional `O(n^2)` check as sortableColumns is a List. I think I'd like a Set better.

ethanalvizo: I converted it into a set but I was wondering what the difference would be. For either data type wouldn't I still be using this contains method to check if that column name is present within it

nbauernfeind (approved):

PR #4074, opened 2023-06-26
SessionService: Inject ErrorTransformer for DHE instead of Forcing Obfuscation

nbauernfeind:
Per request, DHE would like to decide when to obfuscate errors and have the opportunity to do more with API errors.

### DIFF
--- a/server/src/main/java/io/deephaven/server/hierarchicaltable/HierarchicalTableViewSubscription.java
+++ b/server/src/main/java/io/deephaven/server/hierarchicaltable/HierarchicalTableViewSubscription.java
@@ -23,6 +23,7 @@ import io.deephaven.extensions.barrage.util.HierarchicalTableSchemaUtil;
 import io.deephaven.internal.log.LoggerFactory;
 import io.deephaven.io.logger.Logger;
 import io.deephaven.proto.util.Exceptions;
+import io.deephaven.server.session.SessionService;
 import io.deephaven.server.util.Scheduler;
 import io.deephaven.util.SafeCloseable;
 import io.grpc.stub.StreamObserver;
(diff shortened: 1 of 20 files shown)


devinrsmith:
Does this allow the ability to intercept and say, all superusers can receive unobfuscated errors, but everybody else gets obfuscated?

nbauernfeind:
> Does this allow the ability to intercept and say, all superusers can receive unobfuscated errors, but everybody else gets obfuscated?

Yes, exactly. The AuthContext should be properly installed at each call site. I mentally audited and no calls were strict counter examples. (I make mistakes.. but I am at 95% confidence that auth context is available.)

server/src/main/java/io/deephaven/server/hierarchicaltable/HierarchicalTableViewSubscription.java
```diff
@@ -272,14 +276,14 @@ private void process() {
                 }
             }
             if (sendError) {
-                GrpcUtil.safelyError(listener, GrpcUtil.securelyWrapError(log, upstreamFailure, Code.DATA_LOSS));
+                GrpcUtil.safelyError(listener, errorTransformer.transform(upstreamFailure));
```
devinrsmith, line 279: I'm not sure if this context was important, but we are losing `Code.DATA_LOSS`.

nbauernfeind: Yeah, that hurt me, too... -- But, it highlights the inability to differentiate different SRE status types when transforming non-SRE exceptions.

niloc132 (approved):

rcaudy (approved):
Code owner approval for EmbeddedServer

PR #4076, opened 2023-06-27
Added support for (start<0 & stop>0) in slice

### LINKED ISSUE #3130: Broaden the types of indices that `Table.slice` supports
`Table.slice()` returns subsets of a table's rows based upon the `i` index of the rows.  Slicing is allowed from the beginning or the end of the table.

As discussed in https://github.com/deephaven/deephaven-core/pull/3110#discussion_r1033839549, we should be supporting slices of the form `x[-7:8]`, where the first slice index is negative, and the second slice index is positive (`irstPositionInclusive < 0 && lastPositionExclusive > 0`).

The documentation for the following need to be updated.
- [ ] Java
- [ ] Python

malhotrashivam:
Feature request, closes #3130 

Change description: Earlier, an error was generated in this scenario. After this change, a subset of rows will be returned where start will be counted from the end of table (inclusively) and stop from the beginning. 

Documentation update: We need to update both the [java](https://deephaven.io/core/groovy/docs/reference/table-operations/filter/slice/#docusaurus_skipToContent_fallback) and [python](https://deephaven.io/core/docs/reference/table-operations/filter/slice/#docusaurus_skipToContent_fallback) documentation to reflect this change. 
Earlier, if (start<0 & stop>0), an error was generated. Now, a subset of rows will be returned. For example, slice(-3, 5)
will return all rows starting from the third-last row to the fifth row of the table. If there are no rows between these positions, the function will return an empty table. 
Note that index 5 corresponds to the sixth row in the table, so we are excluding the sixth row and ending the subset at the fifth row.

### DIFF
--- a/engine/api/src/main/java/io/deephaven/engine/table/Table.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/Table.java
@@ -356,8 +356,8 @@ public interface Table extends
      * If the firstPosition is negative and the lastPosition is negative, they are both counted from the end of the
      * table. For example, slice(-2, -1) returns the second to last row of the table.
      * <p>
-     * If firstPosition is negative and lastPosition is positive, then the firstRow is counted from the end of
-     * the table, inclusively. The lastPosition is counted from the beginning of the table. For example, slice(-3, 5)
+     * If firstPosition is negative and lastPosition is positive, then firstPosition is counted from the end of the
+     * table, inclusively. The lastPosition is counted from the beginning of the table. For example, slice(-3, 5)
      * returns all rows starting from the third-last row to the fourth row of the table. If there are no rows between
      * these positions, the function will return an empty table.
      *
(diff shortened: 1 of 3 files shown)


nbauernfeind (commented):
It's probably worth having a pair of ticking/refreshing tests that ensure correctness. One that shrinks the source table to empty, and another that grows the source table until the slice is empty.

engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -355,6 +355,11 @@ public interface Table extends
      * <p>
      * If the firstPosition is negative and the lastPosition is negative, they are both counted from the end of the
      * table. For example, slice(-2, -1) returns the second to last row of the table.
+     * <p>
+     * If firstPosition is negative and lastPosition is positive, then firstPosition is counted from the end of the
+     * table, inclusively. The lastPosition is counted from the beginning of the table. For example, slice(-3, 5)
+     * returns all rows starting from the third-last row to the fourth row of the table. If there are no rows between
```
lbooker42, line 361: Am I thinking wrong or is the 6th row of the table selected for index 5?

malhotrashivam: You are not wrong. Both firstPosition and lastPosition are indexes starting from 0. And the subset of rows returned are in the indexes [firstPosition, lastPosition).  So instead of fourth, I should have written fifth in the comment. The sixth row is excluded. I will update this.

nbauernfeind (dismissed):

nbauernfeind (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2902
Reference: https://github.com/deephaven/deephaven.io/issues/2903

PR #4081, opened 2023-06-27
Add more options to ClientOptions in the R Client

alexpeters1208:
Implements TLS authentication with optional PEM-encoded certificate root for server connections for the R client. This also enables the R client to send additional headers over the wire for authentication.


R/rdeephaven/R/client_options_wrapper.R
```diff
@@ -71,6 +71,33 @@ ClientOptions <- R6Class("ClientOptions",
...
         },
 
+        #' @description
+        #' Use the TLS protocol in authentication and subsequent communication.
+        #' @param root_certs Optional PEM-encoded certificate root for server connections. Defaults to system defaults.
+        use_tls = function(root_certs = "") {
+            self$internal_client_options$set_use_tls(TRUE, root_certs)
+        },
+
+        add_int_option = function(opt, val) {
```
chipkent, line 81: docs

R/rdeephaven/src/client.cpp
```diff
@@ -96,6 +96,23 @@ class ClientOptionsWrapper {
         internal_options->setSessionType(sessionType);
     }
 
+    void setUseTls(bool useTls, std::string tlsRootCerts) {
+        internal_options->setUseTls(useTls);
+        internal_options->setTlsRootCerts(tlsRootCerts);
+    }
```
chipkent, line 102: Here, two methods have been combined into one.  This may be ok if we know that these two methods will always be used together, and there will be no other complexity added in the future.  If we expect things to possibly change in this area, some consideration should be made to breaking this up.  I don't know the tradeoffs and expectations on this.  

I guess, looking at this API, I might go with splitting this or setting useTls to true if the method is ever called.  It seems weird if you made the call `setUseTls(false, "ABCD")`.

R/rdeephaven/R/client_options_wrapper.R
chipkent, line 1: Everything in this file needs to have input validation and testing consistent with the other PR.

chipkent (changes requested):

R/rdeephaven/src/client.cpp
alexpeters1208: Cristian and I talked about this a bit, C++ does them separately because of how they chain the operations together in a call. My reasoning here is that `useTls` can be sensibly used without rootCerts, but `tlsRootCerts` will never (my understanding of Cristian's words) be used without a `useTls` to go along with it. In that case, I don't want an R user to be able to think about these as orthogonal methods, which I feel an API like `options$setUseTls` and `options$setTlsRootCerts` allows you to think. However, the API I have `options$useTls(optional rootCerts)`  communicates that you can use TLS without root certs, but there are no root certs without TLS.

In the R wrapping, I have this:
```
use_tls = function(root_certs = "") {
    self$internal_client_options$set_use_tls(TRUE, root_certs)
},
```

This would be just as good, and maybe better, as the C++ wrapping would mirror the C++ client:
```
use_tls = function(root_certs="") {
    self$internal_client_options$set_use_tls(TRUE)
    self$internal_client_options$set_tls_root_certs(root_certs)
}
```

chipkent: If this is all in the C++ wrapper interface, I would definitely go with the final option.  It is much more clear, and it reflects the C++ API better, which is what DH devs expect..

chipkent (approved):

R/rdeephaven/R/helper_functions.R
```diff
@@ -1,7 +1,17 @@
...
     } else if (length(string_candidate) != 1) {
-        stop(paste0("'", arg_name, "' must be passed as a single string. Got character vector of length ", length(string_candidate), " instead."))
+        stop(paste0("'", arg_name, "' must be passed as a single string. Got a character vector of length ", length(string_candidate), " instead."))
+    }
+}
+
+.verify_int <- function(arg_name, int_candidate) {
+    if (class(int_candidate)[[1]] != "numeric") {
+        stop(paste0("'", arg_name, "' must be an integer. Got an object of class ", class(int_candidate)[[1]], " instead."))
+    } else if (all.equal(int_candidate, as.integer(int_candidate)) != TRUE) { # must use != TRUE as the result of all.equal() is not strictly boolean
```
chipkent, line 12: Comments after `{` are awful to read.  put it on the next line.

```suggestion
    } else if (all.equal(int_candidate, as.integer(int_candidate)) != TRUE) { 
        # must use != TRUE as the result of all.equal() is not strictly boolean
```

chipkent (changes requested):

chipkent (approved):

PR #4085, opened 2023-06-27
Allow string duration/period parameters for GRPC time table creation

### LINKED ISSUE #4050: GRPC: Allow string duration/period parameters for `TimeTable` creation
As a user of the GRPC clients, I want to optionally specify `TimeTable` parameters as ISO 8601 strings instead of being required to specify nanoseconds.

lbooker42:
Closes #4050


py/client/pydeephaven/_table_ops.py
```diff
@@ -62,7 +62,23 @@ def get_stub_func(cls, table_service_stub: table_pb2_grpc.TableServiceStub) -> A
...
+                                              start_time_string=self.start_time,
+                                              period_nanos=self.period)
+
+        if isinstance(self.period, str):
+            return table_pb2.TimeTableRequest(result_id=result_id,
+                                              start_time_nanos=self.start_time,
+                                              period_string=self.period)
+
+        return table_pb2.TimeTableRequest(result_id=result_id,
+                                          start_time_nanos=self.start_time,
```
devinrsmith, line 81: Probably a way we could do this w/ a single construction site:

```python
table_pb2.TimeTableRequest(
  result_id=result_id,
  start_time_nanos=self.start_time if not isinstance(self.start_time, str) else None,
  start_time_string=self.start_time if isinstance(self.start_time, str) else None,
  period_nanos=self.start_time if not isinstance(self.start_time, str) else None,
  period_string=self.start_time if isinstance(self.start_time, str) else None)
```

Would need to double check w/ the generated code to make sure `None` is treated as unset.

server/src/main/java/io/deephaven/server/table/ops/TimeTableGrpcImpl.java
```diff
@@ -48,9 +51,27 @@ public void validateRequest(final TimeTableRequest request) throws StatusRuntime
     public Table create(final TimeTableRequest request,
             final List<SessionState.ExportObject<Table>> sourceTables) {
         Assert.eq(sourceTables.size(), "sourceTables.size()", 0);
+        if (request.hasStartTimeString()) {
```
devinrsmith, line 54: Not a fan of these divergent construction sites. I'm assuming we should be able to adapt the types as necessary and have a single construction call.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -315,8 +315,14 @@ message EmptyTableRequest {
...
-  sint64 start_time_nanos = 2 [jstype=JS_STRING];
-  sint64 period_nanos = 3 [jstype=JS_STRING];
+  oneof start_time {
+    uint64 start_time_nanos = 2 [jstype = JS_STRING];
+    string start_time_string = 4;
+  }
+  oneof period {
+    sint64 period_nanos = 3 [jstype = JS_STRING];
+    string period_string = 5;
+  }
```
devinrsmith, line 325: This look structurally good. I'm wondering if we want to update these names so they don't have redundant prefixes?

Also, I wonder if we shouldn't change the name "period" to "duration" since this is _really_ a duration and not a period (wrt the time library terminology).

py/client/pydeephaven/_table_ops.py
lbooker42: This is clever, I was trying to think of how to simplify this code.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
lbooker42: The names of the `oneof` structures aren't propagated to the setter/getter functions of the Java/Python structures, so I think we need the specific names.

On the second point, I'm ambivalent.  `period` and `duration` seem equally accurate in this context, or even `interval` given the the nature of the time table being created.

lbooker42:
This PR requires some changes to the Go client before it can merge.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -315,8 +315,14 @@ message EmptyTableRequest {
 
 message TimeTableRequest {
   Ticket result_id = 1;
-  sint64 start_time_nanos = 2 [jstype=JS_STRING];
-  sint64 period_nanos = 3 [jstype=JS_STRING];
+  oneof start_time {
+    uint64 start_time_nanos = 2 [jstype = JS_STRING];
```
niloc132, line 319:
```suggestion
    sint64 start_time_nanos = 2 [jstype = JS_STRING];
```

this may be a backwards incompatible change?

devinrsmith: Agreed. I thought an earlier version of this RPC was "correct", but maybe I didn't review close enough.

niloc132: Confirmed, go tests will pass with the old generated code if you do this, once #4211 merges.

niloc132:
Filed https://github.com/deephaven/deephaven-core/issues/4219 to follow-up and correctly update the Go client (and revert #4211).

py/client/pydeephaven/session.py
```diff
@@ -414,12 +414,14 @@ def bind_table(self, name: str, table: Table) -> None:
...
-    def time_table(self, period: int, start_time: int = None) -> Table:
+    def time_table(self, period: Union[int, str], start_time: Union[int, str] = None) -> Table:
         """Creates a time table on the server.
 
         Args:
-            period (int): the interval (in nano seconds) at which the time table ticks (adds a row)
-            start_time (int): the start time for the time table in nano seconds, default is None (meaning now)
+            period (Union[int, str]): the interval at which the time table ticks (adds a row); units are nanoseconds
+                or a time interval string, e.g. "PT00:00:.001" or "PT1S"
+            start_time (Union[int, str]): the start time for the time table in nanoseconds or as a ISO 8601 date time
```
rcaudy, line 423:
```suggestion
            start_time (Union[int, str]): the start time for the time table in nanoseconds or as an ISO 8601 date time
```

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -315,8 +315,14 @@ message EmptyTableRequest {
...
-  sint64 start_time_nanos = 2 [jstype=JS_STRING];
-  sint64 period_nanos = 3 [jstype=JS_STRING];
+  oneof start_time {
+    sint64 start_time_nanos = 2 [jstype = JS_STRING];
+    string start_time_string = 5;
+  }
+  oneof period {
+    sint64 period_nanos = 3 [jstype = JS_STRING];
+    string period_string = 6;
+  }
```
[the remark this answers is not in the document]
rcaudy: I think "period" is more clear, from an english perspective, despite the potential for confusion w.r.t. the Java types Duration and Period. You could even call this "period_duration_XYZ" to make it more clear and more confusing at the same time. Unless other languages also have the same definitional weirdness that Java has, I might leave this alone.

jmao-denver (dismissed):

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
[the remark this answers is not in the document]
chipkent: I think "period" is most appropriate here because it is how often the table updates.

py/client/pydeephaven/_table_ops.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, table):
 
 
 class TimeTableOp(TableOp):
-    def __init__(self, start_time: int = 0, period: int = 1000000000):
+    def __init__(self, start_time: Union[int, str] = 0, period: Union[int, str] = 1000000000):
```
chipkent, line 56: I see a few problems here:
* In the python API, we put period first in the args list, because it should never have a default.  Here it is second AND has a default.
* The default zero start time argument is problematic, because 0 nanoseconds is a valid but unusual start time.  The default argument should be some form of "now"

Suggestions:
1) reverse the argument order.
2) remove defaults for period
3) pick a different null value for start time.

@jmao-denver should comment.

py/client/pydeephaven/session.py
```diff
@@ -414,12 +414,14 @@ def bind_table(self, name: str, table: Table) -> None:
...
+    def time_table(self, period: Union[int, str], start_time: Union[int, str] = None) -> Table:
         """Creates a time table on the server.
 
         Args:
-            period (int): the interval (in nano seconds) at which the time table ticks (adds a row)
-            start_time (int): the start time for the time table in nano seconds, default is None (meaning now)
+            period (Union[int, str]): the interval at which the time table ticks (adds a row); units are nanoseconds
+                or a time interval string, e.g. "PT00:00:.001" or "PT1S"
+            start_time (Union[int, str]): the start time for the time table in nanoseconds or as an ISO 8601 date time
+                formatted string; default is None (meaning now)
```
chipkent, line 424: Are we really restricted to 8601, or do we support any format that the time library on the server can handle?  I'm guessing the latter.

server/src/main/java/io/deephaven/server/table/ops/TimeTableGrpcImpl.java
```diff
@@ -48,9 +52,25 @@ public void validateRequest(final TimeTableRequest request) throws StatusRuntime
...
+                adaptPeriod(request),
+                false);
+    }
+
+    private static Instant adaptStartTime(@SuppressWarnings("unused") final TimeTableRequest request) {
+        if (request.hasStartTimeString()) {
+            return DateTimeUtils.parseInstant(request.getStartTimeString());
+        }
+        // Return null if start time nanos is zero or negative.
+        return request.getStartTimeNanos() <= 0 ? null : DateTimeUtils.epochNanosToInstant(request.getStartTimeNanos());
```
chipkent, line 67: I suggest using `Long.MIN_VALUE` instead of zero for the null value, since 0 is theoretically in the middle of the valid range of start times.

py/client/pydeephaven/session.py
chipkent: Looking later in the code, we do support anything that the time library can parse.

chipkent (changes requested):

server/src/main/java/io/deephaven/server/table/ops/TimeTableGrpcImpl.java
devinrsmith: This is a good point, but would break existing clients which expect 0 to mean "no start time"; which I think is the most common case. I wonder if we should create a new RPC for this? I think this may be somewhat an oversight on our part in originally developing this RPC. Or, we can say "use start_time_string if you need exactly the epoch, or before the epoch"... cc @niloc132 .

py/client/pydeephaven/_table_ops.py
lbooker42: I discussed this with @jmao-denver and this is the internal API.  We supply `None` as a default argument in the user-facing API in `session.py`. This default is never used.

py/client/pydeephaven/session.py
lbooker42: Removed the 8601-specific text.

py/client/pydeephaven/_table_ops.py
chipkent: In particular, the default period bothers me.  The default value there is completely arbitrary and users will always specify a value they want.

server/src/main/java/io/deephaven/server/table/ops/TimeTableGrpcImpl.java
lbooker42: Agreed to restore original behavior and not break existing clients.

py/client/pydeephaven/_table_ops.py
lbooker42: I have removed the internal defaults and now handle the `None` value by setting `start_time = 0`. 

The order of the parameters matches the GRPC member structure, it can be changed but unsure the benefit it will bring.

devinrsmith (approved):

chipkent (approved):

PR #4086, opened 2023-06-28
Plugin bidirectional communication

### LINKED ISSUE #3583: Allow two-way communication between UI and server for plugins
As a developer, I want to be able to create a plugin that can communicate between the backend/frontend so that we can have some more interactive plugins.

Currently, we can pass tables from the server to the client, and those tables can update. However, for sending data from the client back to the server/plugin, there's nothing we can do. I've tried some experiments with input tables, but they were both cumbersome and did not work reliably.

This is a requirement for programmatic layout support (#2774), and for other things like allowing interactive matplotlib plots or real-time matplotlib updates (#2078).

How the communication gets implemented is an open discussion. If we did something like allow websockets, it may allow integration with existing frameworks easily (such as [matplotlib's WebAgg interactive plotting](https://matplotlib.org/stable/gallery/user_interfaces/embedding_webagg_sgskip.html)), though there may be security implications and other concerns. Using something like input tables may be easy, but it may make integrations more difficult and/or have other limitations.

mattrunyon:
Fixes #3583 

Needs deephaven-plugin from this PR https://github.com/deephaven/deephaven-plugin/pull/11

If an object that a plugin handles extends the `BidiObjectBase` class in Python, then it will allow the object to use bidirectional communication. The Python object can handle messages by implementing `handle_message(self, msg, objects)` where `msg` is type `bytes`. The Python object can send messages by calling  `self.reference(obj)` to queue up export objects and then `self.sendMessage(msg: Union[str, bytes])` to send the message and export any new objects since the last `sendMessage` call.

JS plugins can `widget.addEventListener(dh.Widget.EVENT_MESSAGE, ...)` where the handler is a `CustomEvent` with the `JsWidgetMessageWrapper` type. JS plugins can also `widget.sendMessage(msg: string | ArrayBuffer)`


plugin/src/main/java/io/deephaven/plugin/type/ObjectType.java
```diff
@@ -99,4 +99,20 @@ interface Reference {
             Optional<String> type();
         }
     }
+
+    void addMessageSender(MessageSender channel);
+
+    void removeMessageSender();
+
+    void handleMessage(String message);
```
niloc132, line 107: To match the OutputStream, probably want to make this InputStream? Also, it should reference the tickets that the client sent to the server as a second arg (probably list of Object?).

mattrunyon: It's now just `byte[]` and also the objects that the tickets reference from the client. If you still think this should be a stream, can you give an example of how that might look? The main thing I'm concerned about is getting it into the Python side in a way that is at least somewhat user friendly. There is a layer (ObjectTypeAdapter in __init__.py) where we can do not user friendly things though if needed

py/server/deephaven/server/plugin/object/__init__.py
```diff
@@ -39,6 +40,16 @@ def __str__(self):
         return str(self._exporter)
 
 
+class MessageSenderAdapter(MessageSender, ExporterAdapter):
```
chipkent, line 43: @jmao-denver This plugin code never went through our review process and is not up to docs specs.

chipkent (approved):

py/server/deephaven/server/plugin/object/__init__.py
jmao-denver: Not much doc in this sub-package of deephaven-core. I believe when @devinrsmith wrote it, it was mainly for internal use. Now we are exposing it to the public, it is time to properly document it.  Not sure if this should apply to the dependency https://github.com/deephaven/deephaven-plugin/tree/main/src/deephaven/plugin.

mattrunyon: Plugin authors are intended to use https://github.com/deephaven/deephaven-plugin/ and not anything from this module

Still probably needs better documentation surrounding it

niloc132 (commented):
Some notes taken getting ready to fork and contribute back to this, mostly notes for myself as I continue.

plugin/src/main/java/io/deephaven/plugin/type/ObjectType.java
```diff
@@ -99,4 +93,21 @@ interface Reference {
...
+
+    void removeMessageSender();
+
+    void handleMessage(byte[] message, Object object, Object[] referenceObjects);
+
+    void sendMessage(byte[] message);
+
+    interface MessageSender extends Exporter {
+
+        Exporter getExporter();
```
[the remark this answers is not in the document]
mattrunyon: I think I added this before I just decided to make it extend `Exporter`, so just a remnant of an old iteration

py/server/deephaven/server/plugin/object/__init__.py
jmao-denver: @niloc132 ^^^

niloc132: ty, this is going to get another pass before its ready to go. presently leaning towards doing some api breaking changes, please see https://github.com/deephaven/deephaven-core/pull/4226. I'll be sure to go over these notes before marking it ready for review.

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidget.java
```diff
@@ -72,14 +121,76 @@ public Uint8Array getDataAsU8() {
...
     public JsWidgetExportedObject[] getExportedObjects() {
-        if (this.exportedObjects == null) {
-            this.exportedObjects = response.getTypedExportIdList().asList().stream()
-                    .map(ticket -> new JsWidgetExportedObject(connection, ticket))
-                    .toArray(JsWidgetExportedObject[]::new);
+        return Js.<JsWidgetExportedObject[]>uncheckedCast(exportedObjects);
+    }
+
+    @JsMethod
+    public JsWidgetExportedObject getExportedObject(int index) {
```
niloc132, line 135: @mattrunyon does this serve a specific purpose outside of just using `object.exportedObjects[index]`?

mattrunyon: Nothing I can think of or remember

mattrunyon:
Superseded by #4226

PR #4089, opened 2023-06-28
Use a blink table instead of DynamicTableWriter for UpdatePerformanceTracker

devinrsmith:


engine/table/src/main/java/io/deephaven/engine/table/impl/perf/UpdatePerformanceStreamPublisher.java
```diff
@@ -0,0 +1,115 @@
...
+    public void register(@NotNull StreamConsumer consumer) {
+        if (this.consumer != null) {
+            throw new IllegalStateException("Can not register multiple StreamConsumers.");
+        }
+        this.consumer = Objects.requireNonNull(consumer);
+    }
+
+    public synchronized void add(IntervalLevelDetails intervalLevelDetails, PerformanceEntry performanceEntry)
+            throws IOException {
+        chunks[0].<String>asWritableObjectChunk().add(EngineMetrics.getProcessInfo().getId().value());
```
devinrsmith, line 68: We can probably skip this column for DHC;  I don't think there is much value?

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/UpdatePerformanceTracker.java
```diff
@@ -251,6 +314,26 @@ public long getIntervalDurationNanos() {
 
     @NotNull
     public QueryTable getQueryTable() {
-        return MemoryTableLogger.maybeGetQueryTable(tableLogger);
+        // todo: should this interface still exist? Why QueryTable? If so, should we return blink, derived, or
```
nbauernfeind, line 317: Maybe we `@Deprecate` it and return the derived table for short-term backwards compatibility?

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/UpdatePerformanceStreamPublisher.java
devinrsmith: Ultimately, this field is in this column due to the nature of importing data into the centralized DIS for DHE - for DHC when we are querying it directly, we should prefer to present the data in its "normalized" form; if we need the process id on the table, we should make that easy for a user to do.

PR #4093, opened 2023-06-28
Use blink table for ServerStateTracker

devinrsmith:


engine/table/src/main/java/io/deephaven/engine/table/impl/util/ServerStateStreamPublisher.java
```diff
@@ -0,0 +1,94 @@
...
+import io.deephaven.time.DateTimeUtils;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Objects;
+
+class ServerStateStreamPublisher implements StreamPublisher {
+
+    private static final TableDefinition DEFINITION = TableDefinition.of(
+            ColumnDefinition.ofTime("IntervalStartTime"),
+            ColumnDefinition.ofInt("IntervalDurationMicros"),
```
devinrsmith, line 21: Should we change to Duration on these types of columns? Can always present a view for compatibility.

engine/table/src/main/java/io/deephaven/engine/table/impl/util/ServerStateTracker.java
```diff
@@ -176,25 +191,51 @@ private void logProcessMem(
...
                     (short) ugpCyclesOnBudget,
                     Arrays.copyOf(ugpCycleTimes, ugpCycles),
                     (short) ugpSafePoints,
-                    deltaMillisToMicros(ugpSafePointTimeMillis));
+                    intervalUGPCyclesSafePointTimeMicros);
         } catch (IOException e) {
             // Don't want to log this more than once in a report
             logger.error().append("Error sending ProcessMemoryLog data to memory").append(e).endl();
         }
+        publisher.add(
```
rcaudy, line 217: I'm surprised that we didn't just do this via a logger.

engine/table/src/main/java/io/deephaven/engine/table/impl/util/ServerStateTracker.java
```diff
@@ -176,25 +191,51 @@ private void logProcessMem(
...
+    /**
+     * Deprecated: use {@link #blinkTable()}.
+     */
+    @Deprecated(since = "0.26.0", forRemoval = true)
     public QueryTable getQueryTable() {
-        return MemoryTableLogger.maybeGetQueryTable(processMemLogger);
+        return (QueryTable) BlinkTableTools.blinkToAppendOnly(blink);
+    }
+
+    public Table blinkTable() {
```
rcaudy, line 238: Would be nice to use table types here.

engine/table/src/main/java/io/deephaven/engine/tablelogger/ServerStateLogLogger.java
```diff
@@ -1,6 +1,7 @@
 package io.deephaven.engine.tablelogger;
 
 import io.deephaven.tablelogger.Row;
+import io.deephaven.tablelogger.Row.Flags;
 
 import java.io.IOException;
```
rcaudy, line 7: Would be lovely to rename the default flags. Shouldn't mention "intraday logger".

engine/table/src/main/java/io/deephaven/engine/table/impl/util/ServerStateStreamPublisher.java
rcaudy: Unless we're doing it across the board, I don't think so. Also, we don't have great duration column sources.

engine/table/src/main/java/io/deephaven/engine/table/impl/util/ServerStateStreamPublisher.java
```diff
@@ -0,0 +1,94 @@
...
+            final int intervalDurationMicros,
+            final int totalMemoryMiB,
+            final int freeMemoryMiB,
+            final short intervalCollections,
+            final int intervalCollectionTimeMicros,
+            final short intervalUGPCyclesOnBudget,
+            final int[] intervalUGPCyclesTimeMicros,
+            final short intervalUGPCyclesSafePoints,
+            final int intervalUGPCyclesSafePointTimeMicros) {
+        chunks[0].asWritableLongChunk().add(DateTimeUtils.millisToNanos(intervalStartTimeMillis));
```
rcaudy, line 63: This could be more efficient if we did our casts once per buffer, and only set sizes in flushInternal.

Issue #3267, opened 2023-01-05
Query language parser tests should cover Python function calls


rbasralian:
The language parser tests (`TestQueryLanguageParser.java`) should cover Python function calls (just the formula conversion/vectorization/type handling parts, not evaluation), so that we can test the parser's behavior without spinning up a whole Python environment (and running the Python tests).

We should be able to do this by mocking PyCallableWrapper, but the PyCallableWrapper class cannot even be loaded because it has a bunch of static initialization that depends on Python.

```
java.lang.ExceptionInInitializerError
	at org.jpy.PyModule.importModule(PyModule.java:92)
	at io.deephaven.engine.util.PyCallableWrapper.getNumbaVectorizedFuncType(PyCallableWrapper.java:128)
	at io.deephaven.engine.util.PyCallableWrapper.<clinit>(PyCallableWrapper.java:22)
	< . . . >
```

There is already an example `testImplicitPythonCall()` test in `TestQueryLanguageParser.java`, which should pass once we figure out how to run it.

PR #4095, opened 2023-06-28
Replace some wrapper funcs in the deephaven.time module with Java wrapped method calls when used in formulas

### LINKED ISSUE #2303: lower_bin should return Instant type
Should be fixed by a solution to #2306 

**Description**

lower_bin returns a string and not a DateTime

**Steps to reproduce**

On edge using:

```python 
from deephaven.time import lower_bin


minute_in_nanos = 60 * 1000 * 1000 * 1000

source = time_table('00:00:01')\
    .update_view( ['received_at_minute = lower_bin(Timestamp, minute_in_nanos)'] )
```

**Expected results**

Expected new column to be a DateTime value without the extra casting
image

**Actual results**

Without casting it explicitly as a DateTime the column is a string.

image

**Versions**
 - Deephaven: ...
 - OS: ...
 - Browser: ...
 - Docker: ...

### LINKED ISSUE #2306: Python Wrapper Properties include Java static method

jmao-denver:
Fixes #2303
Partially fixes #2306

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
@@ -92,9 +92,12 @@ import io.deephaven.engine.context.ExecutionContext;
 import io.deephaven.engine.context.QueryScope;
 import io.deephaven.engine.table.impl.MatchPair;
 import io.deephaven.engine.table.impl.ShiftedColumnsFactory;
+import io.deephaven.engine.util.AbstractScriptSession;
 import io.deephaven.engine.util.PyCallableWrapper.ColumnChunkArgument;
 import io.deephaven.engine.util.PyCallableWrapper.ConstantChunkArgument;
 import io.deephaven.engine.util.PyCallableWrapper;
+import io.deephaven.engine.util.PythonDeephavenSession;
+import io.deephaven.engine.util.PythonScopeJpyImpl;
 import io.deephaven.internal.log.LoggerFactory;
 import io.deephaven.io.logger.Logger;
 import io.deephaven.util.type.TypeUtils;
@@ -109,6 +112,8 @@ import io.deephaven.vector.ShortVector;
 import io.deephaven.vector.Vector;
 import org.apache.commons.lang3.ClassUtils;
 import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.jpy.PyInputMode;
+import org.jpy.PyListWrapper;
 import org.jpy.PyObject;
 
 import java.lang.reflect.Array;
@@ -463,6 +468,66 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
                 .filter((cls) -> nestedClassName.equals(cls.getSimpleName()))
                 .collect(Collectors.toMap(Class::getSimpleName, Function.identity()));
         return m.get(nestedClassName);
+    }
+
+    private boolean pyToJavaReplaced(final Class<?> scope, final MethodCallExpr n) {
+        final String methodName = n.getNameAsString();
+        final QueryScope queryScope = ExecutionContext.getContext().getQueryScope();
+
+        if (scope == null) {
+            final Class<?> methodClass = variables.get(methodName);
+            if (methodClass == PyCallableWrapper.class) {
+                final Object paramValueRaw = queryScope.readParamValue(methodName, null);
+                final PyCallableWrapper pyCallableWrapper = (PyCallableWrapper) paramValueRaw;
+                try {
+                    String javaMethodName = pyCallableWrapper.getAttribute("_j_simple_name", String.class);
+                    n.setName(javaMethodName);
+                    int nargs = pyCallableWrapper.getAttribute("_nargs", int.class);
+                    PyListWrapper defArgList = (PyListWrapper) pyCallableWrapper.getAttribute("_def_args").asList();
+                    int defArgsToAdd = nargs - n.getArguments().size();
+                    if (defArgsToAdd > defArgList.size()) {
+                        throw new IllegalArgumentException("Missing args for " + methodName);
+                    }
+                    if (defArgsToAdd < 0) {
+                        // do nothing, could be valid if the java method has a vararg
+                    }
+                    // Since the func call follows the Java syntax, we don't do name matching for the keyword args
+                    for (int i = defArgList.size() - (nargs - n.getArguments().size()); i < defArgList.size(); i++) {
+                        Object v = PythonScopeJpyImpl.convert(defArgList.get(i));
+                        n.addArgument(v.getClass() == String.class ? "\"" + v.toString() + "\"" : v.toString());
+                    }
+                    return true;
+                } catch (IllegalArgumentException iae) {
+                    throw iae;
+                } catch (RuntimeException e) {
+                    // Not a Java replaceable callable, can safely ignore
+                }
+            }
+        } else if (scope == org.jpy.PyObject.class || scope == PyCallableWrapper.class) {
+            String pythonExpr = n.getScope().get().toString() + "." + methodName;
+            PythonDeephavenSession pds =
+                    (PythonDeephavenSession) ((AbstractScriptSession.UnsynchronizedScriptSessionQueryScope) queryScope)
+                            .scriptSession();
+            Map<String, PyObject> scopeVars = pds.getVariablesRaw();
+            PyObject pyobj;
+            try {
+                pyobj = PyObject.executeCode(pythonExpr, PyInputMode.EXPRESSION, scopeVars, null);
+            } catch (RuntimeException e) {
+                throw new RuntimeException("Cannot find Python callable: " + pythonExpr);
+            }
+            try {
+                Object obj = PythonScopeJpyImpl.convert(pyobj);
+                if (obj.getClass() == (PyCallableWrapper.class)) {
+                    String javaMethodName = ((PyCallableWrapper) obj).getAttribute("_j_simple_name", String.class);
+                    n.setScope(null);
+                    n.setName(javaMethodName);
+                    return true;
+                }
+            } catch (Exception e) {
+                // Not a Java replaceable callable, can safely ignore
+            }
+        }
+        return false;
     }
 
     private Method getMethod(final Class<?> scope, final String methodName, final Class<?>[] paramTypes,
@@ -1140,6 +1205,16 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
          * throw them to 'findClass()'. Many details are not relevant here. For example, field access is handled by a
          * different method: visit(FieldAccessExpr, StringBuilder).
          */
+        Map<String, String> pyConstantsMap = Map.of(
+                "True", "true",
+                "False", "false",
+                "None", "null");
+        final String name = n.getNameAsString();
+        String jConstant = pyConstantsMap.get(name);
+        if (jConstant != null) {
+            printer.append(jConstant);
+            return name.equals("None") ? NULL_CLASS : boolean.class;
+        }
         printer.append(n.getNameAsString());
 
         Class<?> ret = variables.get(n.getNameAsString());
@@ -1865,6 +1940,13 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
             return result;
         }).orElse(null);
 
+        if (pyToJavaReplaced(scope, n)) {
+            if (scope != null) {
+                scope = null;
+                innerPrinter.reset();
+            }
+        }
+
         Expression[] expressions = n.getArguments() == null ? new Expression[0]
                 : n.getArguments().toArray(new Expression[0]);
 
@@ -1873,7 +1955,6 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
         Class<?>[][] parameterizedTypes = getParameterizedTypes(expressions);
 
         Method method = getMethod(scope, n.getNameAsString(), expressionTypes, parameterizedTypes);
-
         Class<?>[] argumentTypes = method.getParameterTypes();
 
         // now do some parameter conversions...
@@ -1913,6 +1994,7 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
             }
         } else { // Groovy or Java method call
             printer.append(innerPrinter);
+            n.setName(method.getName());
             printer.append(n.getNameAsString());
         }
 
@@ -2536,6 +2618,12 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
 
         public boolean hasStringBuilder() {
             return builder != null;
+        }
+
+        public void reset() {
+            if (hasStringBuilder()) {
+                builder.setLength(0);
+            }
         }
 
         /**
--- a/engine/table/src/main/java/io/deephaven/engine/util/PythonDeephavenSession.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/PythonDeephavenSession.java
@@ -1,7 +1,7 @@
 /**
  * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
  */
-package io.deephaven.integrations.python;
+package io.deephaven.engine.util;
 
 import io.deephaven.base.FileUtils;
 import io.deephaven.base.verify.Assert;
@@ -10,13 +10,7 @@ import io.deephaven.engine.context.ExecutionContext;
 import io.deephaven.engine.exceptions.CancellationException;
 import io.deephaven.engine.context.QueryScope;
 import io.deephaven.engine.updategraph.UpdateGraph;
-import io.deephaven.engine.util.AbstractScriptSession;
-import io.deephaven.engine.util.PythonEvaluator;
-import io.deephaven.engine.util.PythonEvaluatorJpy;
-import io.deephaven.engine.util.PythonScope;
-import io.deephaven.engine.util.ScriptFinder;
-import io.deephaven.engine.util.ScriptSession;
-import io.deephaven.integrations.python.PythonDeephavenSession.PythonSnapshot;
+import io.deephaven.engine.util.PythonDeephavenSession.PythonSnapshot;
 import io.deephaven.engine.util.scripts.ScriptPathLoader;
 import io.deephaven.engine.util.scripts.ScriptPathLoaderState;
 import io.deephaven.internal.log.LoggerFactory;
@@ -40,13 +34,17 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.AbstractMap;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
+
+import static java.util.stream.Collectors.toMap;
 
 /**
  * A ScriptSession that uses a JPy cpython interpreter internally.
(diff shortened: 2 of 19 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -465,6 +471,49 @@ private Class<?> findNestedClass(Class<?> enclosingClass, String nestedClassName
...
+                    String javaMethodName = pyCallableWrapper.getAttribute("_j_simple_name", String.class);
+                    methodCallExpr.setName(javaMethodName);
+                    return getMethod(null, javaMethodName, paramTypes, parameterizedTypes);
+                } catch (RuntimeException e) {
+                    // Not, swappable function, can safely ignore
+                }
+            }
+        } else if (scope == org.jpy.PyObject.class || scope == PyCallableWrapper.class) {
+            String pythonExpr = methodCallExpr.getScope().get().toString() + "." + methodName;
+            PythonDeephavenSession pds = (PythonDeephavenSession) ((AbstractScriptSession.UnsynchronizedScriptSessionQueryScope) queryScope).scriptSession();
```
niloc132, line 494: This is gross - and at least going by other usages of this, we might need to hold a lock for it.

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -465,6 +471,49 @@ private Class<?> findNestedClass(Class<?> enclosingClass, String nestedClassName
...
+
+        if (scope == null) {
+            final Class<?> methodClass = variables.get(methodName);
+            if (methodClass == PyCallableWrapper.class) {
+                final Object paramValueRaw = queryScope.readParamValue(methodName, null);
+                final PyCallableWrapper pyCallableWrapper = (PyCallableWrapper) paramValueRaw;
+                try {
+                    String javaMethodName = pyCallableWrapper.getAttribute("_j_simple_name", String.class);
+                    methodCallExpr.setName(javaMethodName);
+                    return getMethod(null, javaMethodName, paramTypes, parameterizedTypes);
```
niloc132, line 487: it's pretty handy that we don't need to know what dispatch the python/jpy wiring would have done, that was a concern i had reading the python, but not a problem!

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -465,6 +471,49 @@ private Class<?> findNestedClass(Class<?> enclosingClass, String nestedClassName
...
+                PyObject pyobj = PyObject.executeCode(pythonExpr, PyInputMode.EXPRESSION, scopeVars, null);
+                Object obj = PythonScopeJpyImpl.convert(pyobj);
+                if (obj.getClass() == (PyCallableWrapper.class)) {
+                    String javaMethodName = ((PyCallableWrapper) obj).getAttribute("_j_simple_name", String.class);
+                    methodCallExpr.setScope(null);
+                    methodCallExpr.setName(javaMethodName);
+                    return getMethod(null, javaMethodName, paramTypes, parameterizedTypes);
+                }
+            } catch (Exception e) {
+                // to report a compilation error?
```
niloc132, line 510: perhaps a warn or some other logging, so we know we couldn't magically make it java?

niloc132 (commented):
This is pretty nice!

jmao-denver:
Some more progress, but still not totally satisfied, 
```
import deephaven.time
import deephaven.time as dhtime
from deephaven.time import now as dhnow
from deephaven.time import now

system_arg = True
resolution_arg = "ns"
nrow = 1000000
formulas = [
    "deephaven.time.now()", 
    "dhtime.now()", 
    "dhnow()", 
    "now()",
    "now(system_arg)",
    "now(system_arg, resolution_arg)",
    # Python built-in constants are not recognized.
    # "now(True, `ns`)",
    "now(true, `ns`)",
    # QLP convert to system==True/true
    #"now(system=True)",
    #"new(system=true)",
    ]
for f in formulas:
    t = empty_table(nrow).update(f"TS = {f}")
    

```

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -334,6 +334,26 @@ public static Instant now() {
         return currentClock().instantNanos();
     }
 
+    @ScriptApi
```
chipkent, line 337: no docs

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -2876,14 +2954,37 @@ public static Instant upperBin(@Nullable final Instant instant, long intervalNan
...
+     * @param intervalNanos size of the window in nanoseconds, either as an integer value or as a time interval string,
+     *        e.g. "PT00:00:00.001" or "PT1s"
+     * @param offset The window start offset in nanoseconds, either as an integer value or as a time interval string,
+     *        e.g. "PT00:00:00.001" or "PT1s". For example, a value of MINUTE would offset all windows by one minute.
+     * @return {@code null} if either input is {@code null}; otherwise, an {@link Instant} representing the end of the
+     *         window
+     */
+    @ScriptApi
+    @Nullable
+    public static Instant upperBin(@Nullable final Instant instant, Object intervalNanos, Object offset) {
```
chipkent, line 2972: and here

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -2899,6 +3000,28 @@ public static ZonedDateTime upperBin(@Nullable final ZonedDateTime dateTime, lon
...
+     * @param intervalNanos size of the window in nanoseconds, either as an integer value or as a time interval string,
+     *        e.g. "PT00:00:00.001" or "PT1s"
+     * @param offset The window start offset in nanoseconds, either as an integer value or as a time interval string,
+     *        e.g. "PT00:00:00.001" or "PT1s". For example, a value of MINUTE would offset all windows by one minute.
+     * @return {@code null} if either input is {@code null}; otherwise, a {@link ZonedDateTime} representing the end of
+     *         the window
+     */
+    @ScriptApi
+    @Nullable
+    public static ZonedDateTime upperBin(@Nullable final ZonedDateTime dateTime, Object intervalNanos, Object offset) {
```
chipkent, line 3018: and here

py/server/deephaven/time.py
```diff
@@ -54,18 +87,7 @@ def now(system: bool = False, resolution: str = 'ns') -> Instant:
...
-            else:
-                return _JDateTimeUtils.now()
-        elif resolution == "ms":
-            if system:
-                return _JDateTimeUtils.nowSystemMillisResolution()
-            else:
-                return _JDateTimeUtils.nowMillisResolution()
-        else:
-            raise ValueError("Unsupported time resolution: " + resolution)
+        return _JDateTimeUtils.now(system, resolution);
```
chipkent, line 90: I'm concerned that this rewrite everything as a single method call is going to be a fail as we get into more complex parts of the API.  I know that we have a lot of things that do branching, possibly in ways where this is not nice.

py/server/deephaven/time.py
```diff
@@ -1709,18 +1732,14 @@ def lower_bin(dt: Union[Instant, ZonedDateTime], interval: Union[int, str], offs
         DHError
     """
     try:
-        if isinstance(interval, str):
-            interval = parse_duration_nanos(interval)
-
-        if isinstance(offset, str):
-            offset = parse_duration_nanos(offset)
-
```
chipkent, line 1717: Here we've traded branching for rewriting the whole API with `Object`.  This is a problem.

py/server/tests/test_time.py
```diff
@@ -886,7 +890,53 @@ def test_parse_local_time(self):
...
+        resolution_arg = "ns"
+        nrow = 1000
+        formulas = [
+            "deephaven.time.now()",
+            "dhtime.now()",
+            "dhnow()",
+            "now()",
+            "now(system_arg)",
+            "now(system_arg, resolution_arg)",
+            # Python built-in constants are recognized.
```
chipkent, line 914: should test None, if that is the case.  The python keywords should probably also be added to the column name verifier so that we don't create column names that match special python words.

py/server/deephaven/time.py
chipkent: Thinking more about the branching problem, I think a big part of it is a dispatch problem.  We aren't using the type information from the columns or input variables to execute the correct function.  Instead, it is all getting forced into this object casting stuff, which I have big concerns about.

chipkent (changes requested):

py/server/deephaven/time.py
jmao-denver: It is really difficult to describe the branching logic in the Python code and then use it to translate to the correct Java code in the QLP, and that's why we want to move these branching logic to the Java API.
Replacing Python wrapper functions with Java ones should be for those wrapper functions that are expected to be used in query formulas.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -2794,6 +2814,39 @@ public static Instant lowerBin(@Nullable final Instant instant, long intervalNan
...
+     *         window
+     */
+    @ScriptApi
+    @Nullable
+    public static Instant lowerBin(@Nullable final Instant instant, Object intervalNanos, Object offset) {
+        long interval = getDurationNanos(intervalNanos);
+        long off = getDurationNanos(offset);
+
+        return lowerBin(instant, interval, off);
+    }
```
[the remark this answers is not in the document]
jmao-denver: This new overload is to support `Union` type in the Python API, it will not conflict with the original overload that takes primitive type parameters. The other approach would be to create multiple overload methods for all the possible combinations of parameter types.

rbasralian:
I very strongly feel that this **should not be merged without corresponding passing tests in TestQueryLanguageParser.java**.

niloc132:
> I very strongly feel that this should not be merged without corresponding passing tests in TestQueryLanguageParser.java.

Unfortunately, TestQueryLanguageParser cannot use jpy, so none of this functionality can be accessed. It should however be possible to add a test in `:python-engine-test` as a junit, so that we can run certain tests from an environment that has jpy and the deephaven wheel installed.

At this time however, all actual tests in that project are disabled (see #734), but the tests do build and run, successfully logging that everything was ignored, so new tests could be added in this way.

```bash
./gradlew :python-engine-test:check
```

niloc132:
https://github.com/niloc132/deephaven-core/pull/new/revive-python-engine-test fixes the broken test env, and adds a sample test that clearly works. Note that an exec context is at least required to do anything interesting, haven't experimented beyond that.

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -465,6 +470,66 @@ private Class<?> findNestedClass(Class<?> enclosingClass, String nestedClassName
...
+                    if (defArgsToAdd > defArgList.size()) {
+                        throw new IllegalArgumentException("Missing args for " + methodName);
+                    }
+                    if (defArgsToAdd < 0) {
+                        // do nothing, could be valid if the java method has a vararg
+                    }
+                    // Since the func call follows the Java syntax, we don't do name matching for the keyword args
+                    for (int i = defArgList.size() - (nargs - n.getArguments().size()); i < defArgList.size(); i++) {
+                        Object v = PythonScopeJpyImpl.convert(defArgList.get(i));
+                        n.addArgument(v.getClass() == String.class ? "\"" + v.toString() + "\"" : v.toString());
```
rcaudy, line 497: Any chance you unwrap `None` to `null` and then NPE on the `v.toString()`?

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -465,6 +470,66 @@ private Class<?> findNestedClass(Class<?> enclosingClass, String nestedClassName
...
+                    }
+                    return true;
+                } catch (IllegalArgumentException iae) {
+                    throw iae;
+                } catch (RuntimeException e) {
+                    // Not a Java replaceable callable, can safely ignore
```
rcaudy, line 503: Can we use `hasAttribute` to avoid using failed `getAttribute` calls for control flow?

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -465,6 +470,66 @@ private Class<?> findNestedClass(Class<?> enclosingClass, String nestedClassName
...
+                }
+            }
+        } else if (scope == org.jpy.PyObject.class || scope == PyCallableWrapper.class) {
+            String pythonExpr = n.getScope().get().toString() + "." + methodName;
+            PythonDeephavenSession pds =
+                    (PythonDeephavenSession) ((AbstractScriptSession.UnsynchronizedScriptSessionQueryScope) queryScope)
```
rcaudy, line 509:
```suggestion
                    (PythonDeephavenSession) ((AbstractScriptSession.ScriptSessionQueryScope) queryScope)
```

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -465,6 +470,66 @@ private Class<?> findNestedClass(Class<?> enclosingClass, String nestedClassName
...
+            } catch (RuntimeException e) {
+                throw new RuntimeException("Cannot find Python callable: " + pythonExpr);
+            }
+            try {
+                Object obj = PythonScopeJpyImpl.convert(pyobj);
+                if (obj.getClass() == (PyCallableWrapper.class)) {
+                    String javaMethodName = ((PyCallableWrapper) obj).getAttribute("_j_simple_name", String.class);
+                    n.setScope(null);
+                    n.setName(javaMethodName);
+                    return true;
```
rcaudy, line 524: Don't we want the same default handling? That's missing...

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -465,6 +470,66 @@ private Class<?> findNestedClass(Class<?> enclosingClass, String nestedClassName
...
+            String pythonExpr = n.getScope().get().toString() + "." + methodName;
+            PythonDeephavenSession pds =
+                    (PythonDeephavenSession) ((AbstractScriptSession.UnsynchronizedScriptSessionQueryScope) queryScope)
+                            .scriptSession();
+            Map<String, PyObject> scopeVars = pds.getVariablesRaw();
+            PyObject pyobj;
+            try {
+                pyobj = PyObject.executeCode(pythonExpr, PyInputMode.EXPRESSION, scopeVars, null);
+            } catch (RuntimeException e) {
+                throw new RuntimeException("Cannot find Python callable: " + pythonExpr);
```
rcaudy, line 516: Throw something else. `UncheckedDeephavenException` or a subclass.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -334,6 +334,26 @@ public static Instant now() {
         return currentClock().instantNanos();
     }
 
+    @ScriptApi
+    @NotNull
+    public static Instant now(boolean useSystem, String resolution) {
```
rcaudy, line 339: No `"us"` for microseconds? Feels pretty incomplete.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
rcaudy: Arguably we could take this out of the API in Python, instead of adding a new Java method. Presumably if we thought it was really necessary, we would already have a Java method like this. `TimeUnit` is a much better way to express this, rather than a String.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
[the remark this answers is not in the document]
rcaudy: An overload per type is the Java way. This current way is not usable from Java. If we want it, at least hide it in some other class that we never tell Java programmers about.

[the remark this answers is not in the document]
rcaudy: Maybe the answer looks something like applying a conversion to the input and calling the existing overload? That said, if the input is from a column, for example, that doesn't really work.

engine/table/src/main/java/io/deephaven/engine/util/PythonObjectWrapper.java
rcaudy, line 1: This class didn't need to move.

engine/table/src/main/java/io/deephaven/engine/util/PythonDeephavenSession.java
rcaudy, line 1: We should find a way to not move this. How about we have an interface in `engine.util` `PythonMethodExpressionResolver` or something more appropriate, that just takes a pair of strings?
```
public interface PythonMethodExpressionResolver {
     PyObject resolve(@NotNull final String scope, @NotNull final String name);
}
```
Then, just have the `PythonDeephavenScriptSession` implement that interface.

py/server/tests/test_time.py
```diff
@@ -225,89 +222,89 @@ def test_epoch_nanos(self):
...
         self.assertEqual(dt, epoch_nanos_to_zdt(nanos, tz))
-        self.assertEquals(None,epoch_nanos_to_zdt(NULL_LONG, tz))
+        self.assertEqual(None, epoch_nanos_to_zdt(NULL_LONG, tz))
 
     def test_epoch_micros_to_zdt(self):
         tz = time_zone("ET")
         nanos = 12345678987654321
-        micros = nanos // 10**3
-        dt = dtypes.Instant.j_type.ofEpochSecond(0, micros * 10**3).atZone(tz)
+        micros = nanos // 10 ** 3
```
rcaudy, line 277: When can we start using black or some other code beautifier?

table-api/src/main/java/io/deephaven/api/util/NameValidator.java
```diff
@@ -95,7 +95,7 @@ private boolean isValid(String name) {
     }
 
     private static final Set<String> QUERY_LANG_RESERVED_VARIABLE_NAMES =
-            Stream.of("in", "not", "i", "ii", "k").collect(
+            Stream.of("in", "not", "i", "ii", "k", "True", "False", "None").collect(
```
rcaudy, line 98: You're leaking your Python into my Java....

rbasralian:
> I very strongly feel that this **should not be merged without corresponding passing tests in TestQueryLanguageParser.java**.
> 
> #3267

Just to elaborate on this a bit further — the point of TestQueryLanguageParser is to test the _behavior of the parser_ without the added overhead and complexity of actually evaluating formulas. This should apply to Python as well — we should be able to verify the parser's handling of Python variables, transformation of Python-related formulas (including/especially vectorization and Java function substitution), and return types of formulas using Python methods/variables without actually running a Python environment.

Anything that actually requires crossing over to Python should be mocked or refactored to provide interfaces that support no-Python-required implementations we can use for testing from Java. We should be able to know from Java that, given a hypothetical Python function and hypothetical Python variables defined in a test case, the parser would transform a formula the way we expect and find the return type we expect.

The more we evolve the parser to improve Python support, the more important this becomes — otherwise, every Python change decreases the overall coverage we get from TestQueryLanguageParser, which increases the odds of introducing issues that can only be caught from Python tests. This gets expensive because the Python tests take way longer to run, are way harder to debug, etc.

Issue #3958, opened 2023-06-08
Our calendar files are very out of date, and we need a process for updating them


rcaudy:
Our calendar files are often years out of date w.r.t. holidays, meaning our business calendar support is severely compromised. We need to update them, and then design a process by which they become updated regularly.

rcaudy:
Re-opening. While we've done our damage control, we still need to consider:
1. Should we be in the "business calendar" business?
2. If so, should we maintain any calendars?
3. If so, what should our process be?

PR #4098, opened 2023-06-29
Remove calendars that have not been updated

chipkent:
Remove old calendars that have not been maintained so that users do not use an invalid calendar.  

Maintained calendars and an improved calendar interface will be added shortly.

Partially resolves https://github.com/deephaven/deephaven-core/issues/3958


rcaudy (approved):

rcaudy:
I'm replacing this PR with #4102

PR #4101, opened 2023-06-29
Create gRPC SessionService#publishFromTicket

### LINKED ISSUE #4075: bind_object
As a Developer, I want to be able to bind objects, like Figures, so that I can use them in a Jupyter system connected from a client instead of collocated in the Deephaven JVM.

nbauernfeind:
Fixes #4075.

Colin pointed out that DnD needs to be able to disable publishing to ScopeTickets for PQ's (which need to have frozen query scopes). This adds a flag that can be set. It highlights that we might need to add authorization hooks for `TicketResolver#publish`.

This is a breaking change for DnD shim as it changes the TicketResolver API to add a callback that can be invoked once the publish is complete.


server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
```diff
@@ -186,6 +176,37 @@ public void exportFromTicket(
...
+
+        final SessionState.ExportObject<Object> source = ticketRouter.resolve(
+                session, request.getSourceId(), "sourceId");
+        Ticket resultId = request.getResultId();
+
+        final SessionState.ExportBuilder<Object> publisher = ticketRouter.publish(
+                session, resultId, "resultId", () -> {
+                    // when publish is complete, complete the gRPC request
+                    GrpcUtil.safelyComplete(responseObserver, PublishResponse.getDefaultInstance());
+                });
```
devinrsmith, line 204: In the ExportTicketResolver case, this safelyComplete is happening on thread as opposed to as part of a submit - is that correct?

nbauernfeind: Yeah, but you are right that is kind of bogus; might as well submit a non-export.

server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
```diff
@@ -177,13 +167,45 @@ public void exportFromTicket(
...
+        Ticket resultId = request.getResultId();
+
+        final SessionState.ExportBuilder<Object> publisher = ticketRouter.publish(
+                session, resultId, "resultId", () -> {
+                    // when publish is complete, complete the gRPC request
+                    GrpcUtil.safelyComplete(responseObserver, PublishResponse.getDefaultInstance());
                 });
+        publisher.require(source)
+                .onError(responseObserver)
+                .submit(source::get);
```
devinrsmith, line 208: I'm still not completely in-tune w/ all the export builder constructions. Colin should approve

chipkent (approved):
I approved the py files.  I did not review other files.

niloc132 (dismissed):

server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
```diff
@@ -177,13 +167,45 @@ public void exportFromTicket(
...
                 .require(source)
                 .onError(responseObserver)
-                .submit(() -> {
-                    GrpcUtil.safelyComplete(responseObserver, ExportResponse.getDefaultInstance());
-                    return source.get();
+                .submit(source::get);
+        session.nonExport()
+                .require(destination)
+                .onError(responseObserver)
+                .submit(() -> GrpcUtil.safelyComplete(responseObserver, ExportResponse.getDefaultInstance()));
```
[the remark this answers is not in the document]
niloc132: I'm not sure I follow - source.get() is strictly before the safelyComplete call, since the source.get() will be written to `destination`, and the safelyComplete won't complete until `destination` is exported? If the source.get() fails, destination will be left in an error state, so safelyComplete won't be called - we would attempt to call the responseObserver.onError(...), but the second invocation would fail, so only the first error would be sent.

In the case of the publish method below, two distinct submit()s are required, one that depends on the source and one that depends on the destination. It is somewhat non-standard to require(destination) and not respond to the client until that happens though, since we know that destination must be an export ticket, so the client can already race this call. Getting a successful result here only depends on the source being a success - if we think this pattern is required here, it is likely required in many more places as well? Not wrong, just not necessary.

server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
niloc132: We must not complete the gRPC until publisher was complete, since it is possible that the client could get the response and start using the new ticket before it was actually written to.

publisher.require(source).submit(source::get) ensures that source exists and is ready for use, reads the underlying object out of source, passes it to the publisher. If we completed after calling source.get(), the client in theory could receive the message before the value was written into resultId. Then the runnable passed to ticketRouter.publish() will run after the publish iis complete, ensuring that the value is ready for use (but _not_ guaranteeing that someone else doesn't stomp on it in the meantime...).

So this one seems correct to me at least, and I don't think I would have any changes to make here.

niloc132 (commented):
Two replies to @devinrsmith - I'm still happy to merge it now, but we can discuss further before we merge.

niloc132:
To clarify, if we're not 100% on this, I think we should punt until Nate returns, and just guard the publish-to-scope-ticket call with a config, off by default.

server/src/main/java/io/deephaven/server/session/SessionServiceGrpcImpl.java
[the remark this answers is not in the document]
niloc132: After talking with Devin, I agree that the simpler fix would be to call source.get(), store in a temp variable, and then call safelyComplete after that source.get() was successful. I don't think this extra nonExport() buys us anything - if we think it is important, we must add it in many other places. I also don't think it is doing anything wrong, and cannot introduce an error we aren't already handling safely (from double-calling onError).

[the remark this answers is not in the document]
nbauernfeind: Yeah; Devin is right. This covers his original concern and the non-export bought us nothing.

devinrsmith (approved):

rcaudy (approved):

PR #4108, opened 2023-06-29
Use blink tables for query (op) performance tables

devinrsmith:


engine/table/src/main/java/io/deephaven/engine/table/impl/util/EngineMetrics.java
```diff
@@ -85,20 +88,36 @@ private EngineMetrics() {
...
+    /**
+     * Deprecated: see {@link #queryPerformanceTable()}.
+     */
+    @Deprecated(since = "0.26.0", forRemoval = true)
     public QueryTable getQplLoggerQueryTable() {
-        return MemoryTableLogger.maybeGetQueryTable(qplLogger);
+        return (QueryTable) BlinkTableTools.blinkToAppendOnly(qpImpl.blinkTable());
+    }
+
+    public Table queryPerformanceTable() {
```
rcaudy, line 99: Across these 3 or 4 PRs, all of them look "OK". That said, I think we should push ourselves to be more consistent in:
1. The pattern we follow when delivering data to loggers and publishers (sometimes we have a delegating+publishing logger, sometimes the source publishes to each separately, etc.
2. The pattern we expose to users for getting the tables to operate on.

I also think we want to be sure we future-proof (2) for when we might add table-type driven accessors/initializers.

PR #4115, opened 2023-06-30
Make AuthContext available to QPL, QOPL, and UPL

### LINKED ISSUE #4080: Performance Entry/Nugget should capture AuthContext
As a developer I want to have the Entry contain the authentication context of the user who initiated the operation, so that I can record it in logs.

rcaudy:
Fixes #4080

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
@@ -146,8 +146,10 @@ public abstract class AbstractScriptSession<S extends AbstractScriptSession.Snap
                 final SafeCloseable ignored = LivenessScopeStack.open(this, false)) {
 
             try {
-                // actually evaluate the script
-                executionContext.apply(() -> evaluate(script, scriptName));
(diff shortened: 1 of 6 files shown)


engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -146,8 +146,10 @@ public synchronized final Changes evaluateScript(final String script, final @Nul
                 final SafeCloseable ignored = LivenessScopeStack.open(this, false)) {
 
             try {
-                // actually evaluate the script
-                executionContext.apply(() -> evaluate(script, scriptName));
+                // Actually evaluate the script; use the enclosing auth context, since AbstractScriptSession's
+                // ExecutionContext never has a non-null AuthContext
```
devinrsmith, line 150: double negative.

devinrsmith (approved):

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
rcaudy: I'll update this in a subsequent PR, unless something goes wrong with unit testing.

PR #4116, opened 2023-06-30
Update Barrage performance tables to use blink tables

### LINKED ISSUE #4114: Replace Barrage performance tables with blink tables
Follow-up to https://github.com/deephaven/deephaven-core/pull/4110

devinrsmith:
Fixes #4114


engine/table/src/main/java/io/deephaven/engine/tablelogger/impl/memory/MemoryTableLogger.java
```diff
@@ -15,7 +15,11 @@
 
 /**
  * Base class for memory table loggers that create and initialize a {@link DynamicTableWriter}.
+ *
+ * <p>
+ * Deprecated: prefer constructions using blink tables, see {@link io.deephaven.stream.StreamToBlinkTableAdapter}.
  */
+@Deprecated(since = "0.26.0", forRemoval = true)
```
rcaudy, line 22: What still uses this?

rcaudy (approved):
Approved. Wouldn't mind deleting the deprecated class, though.

engine/table/src/main/java/io/deephaven/engine/tablelogger/impl/memory/MemoryTableLogger.java
devinrsmith: Nothing in DHC, but checking w/ DHE. Technically, users could too.

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2867
Reference: https://github.com/deephaven/deephaven.io/issues/2866

PR #4121, opened 2023-07-03
Include docstring of the constructors in html

### LINKED ISSUE #4120: The Python client docs doesn't include the docstring of __init__ method
Currently the generated doc is:
image

What we want is:
image

jmao-denver:
Fixes #4120

### DIFF
--- a/py/client/docs/source/conf.py
+++ b/py/client/docs/source/conf.py
@@ -68,6 +68,7 @@ html_theme_options = {
 add_module_names = False
 # if we allow sphinx to generate type hints for signatures (default), it would make the generated doc cluttered and hard to read
 autodoc_typehints = 'none'
+autoclass_content = 'both'
 
 #########################################################################################################################################################################
 
--- a/sphinx/source/conf.py
+++ b/sphinx/source/conf.py
@@ -67,6 +67,7 @@ html_theme_options = {
 add_module_names = False
 # if we allow sphinx to generate type hints for signatures (default), it would make the generated doc cluttered and hard to read
 autodoc_typehints = 'none'
+autoclass_content = 'both'
 
 #########################################################################################################################################################################
 


py/client/docs/source/conf.py
```diff
@@ -68,6 +68,7 @@
 add_module_names = False
 # if we allow sphinx to generate type hints for signatures (default), it would make the generated doc cluttered and hard to read
 autodoc_typehints = 'none'
+autoclass_content = 'both'
```
chipkent, line 71: I think this now includes doctrines for __init__ methods.  I don't think we have been as diligent about documenting them.  All of the sphinx docs need to be reviewed to see which methods now need good docstrings.

chipkent (changes requested):

py/client/docs/source/conf.py
jmao-denver: 1. Our convention is if the class shouldn't be instantiated by the user directly, meaning the instances can only be made through factory functions or other operations,  then we don't provide a doc string for the `__init__` method.
2. I actually checked both the server and the client code, and found that we stick to 1 pretty religiously.

chipkent: I think that was OK when we didn't publish the method, but maybe now, we should have an internal use only note.  Thoughts?

jmao-denver: 1. the new option isn't going to publish the __init__ method itself but instead combine the docstring of it with the one on the class.
2. I think in general we already have the note of similar purpose in the class docstrings, e.g. 

```
class Table(TableInterface):
    """A Table object represents a reference to a table on the server. It is the core data structure of
    Deephaven and supports a rich set of operations such as filtering, sorting, aggregating, joining, snapshotting etc.

    Note, an application should never instantiate a Table object directly. Table objects are always provided through
    factory methods such as Session.empty_table(), or import/export methods such as Session.import_table(),
    open_table(), or any of the Table operations.

```
and
```
class Table(JObjectWrapper):
    """A Table represents a Deephaven table. It allows applications to perform powerful Deephaven table operations.

    Note: It should not be instantiated directly by user code. Tables are mostly created by factory methods,
    data ingestion operations, queries, aggregations, joins, etc.

    """
```

chipkent (approved):

Issue #4122, opened 2023-07-03
Publish .module files for shadow projects


devinrsmith:
Most of our public projects publish a [.module format](https://docs.gradle.org/current/userguide/publishing_gradle_module_metadata.html) (in addition to the .pom format). This is most useful for downstream gradle consumers whose build processes can take advantage of this richer format.

Our shadow projects don't currently publish a .module file:

```
$ ./gradlew authentication:example-providers:sql-username-password:generateMetadataFileForMavenJavaPublication
...
> Task :authentication:example-providers:sql-username-password:generateMetadataFileForMavenJavaPublication SKIPPED
Maven publication 'mavenJava' isn't attached to a component. Gradle metadata only supports publications with software components (e.g. from component.java)
```

Discovered during investigation into https://github.com/deephaven/deephaven-core/issues/4119

PR #4129, opened 2023-07-05
Re-implement shadow's "component" method to correctly

### LINKED ISSUE #4119: Unable to build against DHC's psk-authentication-provider
**Description**

I'm building DnD locally in Vermilion against DHC 0.26.0, which i've published locally (since it's not released yet). I'm able to build fine when running jar, but on distTar, i'm getting errors like:
Execution failed for task ':startScripts'.
> Error while evaluating property 'relativeClasspath' of task ':startScripts'
   > Could not resolve all files for configuration ':runtimeClasspath'.
      > Could not find io.deephaven:authentication:0.26.0.
        Searched in the following locations:
          - file:/Users/afarhangi/.m2/repository/io/deephaven/authentication/0.26.0/authentication-0.26.0.pom
          - https://repo.maven.apache.org/maven2/io/deephaven/authentication/0.26.0/authentication-0.26.0.pom
          - https://illumon.jfrog.io/illumon/libs-local-jdk11/io/deephaven/authentication/0.26.0/authentication-0.26.0.pom
          - https://illumon.jfrog.io/illumon/libs-snapshot-local/io/deephaven/authentication/0.26.0/authentication-0.26.0.pom
        Required by:
            project : > project :Main > io.deephaven:deephaven-server:0.26.0 > io.deephaven:deephaven-psk-authentication-provider:0.26.0
...

Same thing for io.deephaven:Configuration:0.26.0 and io.deephaven:log-factory:0.26.0.

Note that they do exist locally w/ the "deephaven-" prefix, e.g. /Users/afarhangi/.m2/repository/io/deephaven/deephaven-authentication/0.26.0/deephaven-authentication-0.26.0.pom.

**Steps to reproduce**

**Expected results**

**Actual results**

**Versions**
 - Deephaven: main, 0.26.0, 709f2a493f4e00d7ee7e3db11c5ed8e0e7b042f1
 - OS: Mac
 - Browser: N/A
 - Docker: Desktop 4.16.2, Engine 20.10.22, Compose v2.15.1

niloc132:
Does not address #4122, which would require changing shadow to use components correctly.

Fixes #4119

### DIFF
--- a/buildSrc/src/main/groovy/io.deephaven.java-shadow-publishing-conventions.gradle
+++ b/buildSrc/src/main/groovy/io.deephaven.java-shadow-publishing-conventions.gradle
@@ -39,7 +39,6 @@ PublishingTools.setupPublications(project) { publication ->
           dependencyNode.appendNode('artifactId', base.archivesBaseName)
           dependencyNode.appendNode('version', it.version)
           dependencyNode.appendNode('scope', 'runtime')
-          println dependencyNode
         }
       }
     }
--- a/buildSrc/src/main/groovy/io.deephaven.java-shadow-publishing-conventions.gradle
+++ b/buildSrc/src/main/groovy/io.deephaven.java-shadow-publishing-conventions.gradle
@@ -19,8 +19,31 @@ tasks.withType(Javadoc) {
 }
 
 PublishingTools.setupPublications(project) { publication ->
-  // This assumes that the shadow plugin is enabled
-  project.shadow.component(publication)
+  // This assumes that the shadow plugin is enabled, rather than using the vanilla java component.
+
+  // Inlines the shadow.component call, which doesn't actually use components, and doesn't generate
+  // the pom late enough to pick up gradle's own artifactId wiring from our archivesBaseName changes
+//  project.shadow.component(publication)
+  publication.artifact(project.tasks.named("shadowJar"))
+  publication.pom {
+    withXml {
+      def root = asNode()
+      def dependenciesNode = root.appendNode('dependencies')
+
+      project.configurations.shadow.allDependencies.each {
+        if ((it instanceof ProjectDependency) || ! (it instanceof SelfResolvingDependency)) {
+          def dependencyNode = dependenciesNode.appendNode('dependency')
+          dependencyNode.appendNode('groupId', it.group)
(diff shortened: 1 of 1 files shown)


niloc132:
This is not an ideal fix, but it does the job. I wasn't able to get the shadow plugin to be delayed in its changes of the default pom instance so that the GenerateMavenPom task is able to write it correctly, but instead just reimplements the logic to more closely match what we expect.

A "better" fix would delay our call to project.shadow.component() until after gradle has processed the artifactId changes to all dependency projects, propagating them into their own poms, so that they can be read. 

A "good" fix would be to provide a variant or a component just for the shadow. It appears that the shadow plugin attempts to do this, but doesn't really commit to it (giving us #4122).

Note that we also cannot update to shadow 8.1.1 at this time, as it requires gradle 8.

devinrsmith:
Filed https://github.com/deephaven/deephaven-core/issues/4132 for follow-up.

buildSrc/src/main/groovy/io.deephaven.java-shadow-publishing-conventions.gradle
```diff
@@ -19,8 +19,31 @@ tasks.withType(Javadoc) {
...
+      project.configurations.shadow.allDependencies.each {
+        if ((it instanceof ProjectDependency) || ! (it instanceof SelfResolvingDependency)) {
+          def dependencyNode = dependenciesNode.appendNode('dependency')
+          dependencyNode.appendNode('groupId', it.group)
+          BasePluginConvention base = it.dependencyProject.convention.getPlugin(BasePluginConvention)
+
+          dependencyNode.appendNode('artifactId', base.archivesBaseName)
+          dependencyNode.appendNode('version', it.version)
+          dependencyNode.appendNode('scope', 'runtime')
+          println dependencyNode
```
devinrsmith, line 42: Remove.

devinrsmith (changes requested):

niloc132:
> Filed https://github.com/deephaven/deephaven-core/issues/4132 for follow-up.

Based on my read of the specific sources involved with this bug, this is unlikely to help.

devinrsmith (approved):

PR #4134, opened 2023-07-05
Add blink time table support for QST and RPC

devinrsmith:


qst/src/main/java/io/deephaven/qst/table/TimeTable.java
```diff
@@ -85,6 +91,8 @@ public interface Builder {
 
         Builder startTime(Instant startTime);
 
+        Builder blinkTable(boolean blinkTable);
+
         Builder id(UUID id);
```
rcaudy, line 96: I think you might need `of` overloads for `blinkTable`, or to randomize `id` by default. Otherwise, I worry users will mess up `id` when using the builder.

engine/table/src/main/java/io/deephaven/engine/table/impl/TableCreatorImpl.java
```diff
@@ -64,7 +65,12 @@ public final Table of(EmptyTable emptyTable) {
     public final Table of(TimeTable timeTable) {
         final io.deephaven.base.clock.Clock clock = ClockAdapter.of(timeTable.clock());
         final Instant firstTime = timeTable.startTime().orElse(null);
-        return TableTools.timeTable(clock, firstTime, timeTable.interval().toNanos());
+        return new io.deephaven.engine.table.impl.TimeTable(
+                ExecutionContext.getContext().getUpdateGraph(),
+                clock,
+                firstTime,
+                timeTable.interval().toNanos(),
+                timeTable.blinkTable());
```
rcaudy, line 73: Might be slightly preferable to use the TimeTable.Builder.

rcaudy (approved):

PR #4135, opened 2023-07-05
Added support to slice a table using percentages

### LINKED ISSUE #3203: Add `Table.slicePct`
The `Table.slice` operator is analogous to the `Table.head` and `Table.tail` operators.  The `Table.head` and `Table.tail` operators have `headPct` and `tailPct` methods.  `slice` should have `slicePct` to be complete and consistent with the other methods in the family.

malhotrashivam:
Feature request, closes #3203

**Change description:** Added a new slicePct() method that takes percentages, similar to how slice() works with row indexes.

**Documentation update:** We need to add both java and python documentations, similar to the slice() method.
For Java, the syntax is table.slicePct(double startPercentInclusive, double endPercentExclusive)
For Python, the syntax is table.slice_pct(start_pct:float, end_pct:float)

The function returns a subset of table in the range [startPercentInclusive * sizeOfTable , endPercentExclusive * sizeOfTable). For example, for a table of size 10, slicePct(0.1, 0.7) will return a subset from the second row to the seventh row. Similarly, slicePct(0, 1) would return the entire table (because row positions run from 0 to size-1).
 The percentage arguments must be in [0,1], otherwise the function returns an error.
Also, the method is not supported for blink tables and will give an error.


lbooker42 (commented):
Suggested test: Make sure that slices of a table when merged recreate exactly the original table.

Specifically, create a test where you create slices 0.0-0.25, 0.25-0.5, 0.5-0.75, 0.75-1.0 then verify all rows from the main table are in exactly one slice (no duplicates or missing rows).

lbooker42 (dismissed):

lbooker42 (dismissed):

py/server/tests/test_table.py
```diff
@@ -237,6 +237,13 @@ def test_head_tail_pct(self):
                 result_table = op(self.test_table, pct=0.1)
                 self.assertEqual(result_table.size, self.test_table.size * 0.1)
 
+    def test_slice_pct(self):
+        ops = [Table.slice_pct]
+        for op in ops:
```
jmao-denver, line 242: Not a big deal but are we going to have other similar ops that we might want to group test them? If not, then it is better to remove the iteration.

py/server/deephaven/table.py
```diff
@@ -2084,6 +2084,28 @@ def slice(self, start: int, stop: int) -> Table:
         except Exception as e:
             raise DHError(e, "table slice operation failed.") from e
 
+    def slice_pct(self, start_pct: float, end_pct: float) -> Table:
+        """Extracts a subset of a table by row percentages.
+
+        Returns a subset of table in the range [startPercentInclusive * sizeOfTable , endPercentExclusive * sizeOfTable).
```
jmao-denver, line 2090: These are Java names and don't match the parameter names.

engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -363,6 +363,20 @@ public interface Table extends
...
+     * Returns a subset of table in the range [startPercentInclusive * sizeOfTable , endPercentExclusive * sizeOfTable).
+     * For example, for a table of size 10, slicePct(0.1, 0.7) will return a subset from the second row to the seventh
+     * row. The percentage arguments must be in [0,1], otherwise the function returns an error.
+     *
+     * @param startPercentInclusive the starting percentage point for rows to include in the result, range [0, 1]
+     * @param endPercentExclusive the ending percentage point for rows to include in the result, range [0, 1]
+     * @return a new Table, which is the requested subset of rows from the original table
+     */
+    @ConcurrentMethod
+    Table slicePct(double startPercentInclusive, double endPercentExclusive);
```
chipkent, line 378: Looking at this spec, there is a weird case to consider.  `slicePct(0,1)` will return a table with `N-1` rows.  How would a user get all of the rows?  Seems impossible.  

... Actually, from reading the code, I _think_ that `slicePct(0,1)` will return the entire table, as desired.  This would indicate that the docs need to be reworked to reflect what the operation is actually doing.

engine/table/src/test/java/io/deephaven/engine/table/impl/QueryTableSliceTest.java
```diff
@@ -433,6 +433,36 @@ private void doSliceTest(QueryTable table, String expected, int firstPositionInc
         assertEquals(expected, result);
     }
 
+    public void testSlicePct() {
+        final QueryTable table = TstUtils.testRefreshingTable(
+                RowSetFactory.fromRange(10, 19).toTracking(),
+                TableTools.charCol("letter", "9876543210".toCharArray()));
+
+        doSlicePctTest(table, "9876543210", 0, 1);
```
chipkent, line 441: Here you actually test the `slicePct(0,1)` case, and it does return the whole range.  This indicates that the docs need a rework to be more precise.

py/server/deephaven/table.py
chipkent: Yep

chipkent, line 2090: Update the docs based on the javadoc comments to be precise about what the operation is doing.

engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -368,7 +368,8 @@ public interface Table extends
      * <p>
      * Returns a subset of table in the range [startPercentInclusive * sizeOfTable , endPercentExclusive * sizeOfTable).
```
chipkent, line 369: I suggest including the floors so that super quant nerds and number lawyers know what is going on.

py/server/deephaven/table.py
```diff
@@ -2087,9 +2087,10 @@ def slice(self, start: int, stop: int) -> Table:
     def slice_pct(self, start_pct: float, end_pct: float) -> Table:
         """Extracts a subset of a table by row percentages.
 
-        Returns a subset of table in the range [startPercentInclusive * sizeOfTable , endPercentExclusive * sizeOfTable).
+        Returns a subset of table in the range [start_pct * size_of_table , end_pct * size_of_table).
```
chipkent, line 2090: I suggest including the floors so that super quant nerds and number lawyers know what is going on.

py/server/deephaven/table.py
```diff
@@ -2087,9 +2087,10 @@ def slice(self, start: int, stop: int) -> Table:
...
         For example, for a table of size 10, slicePct(0.1, 0.7) will return a subset from the second row to the seventh
-        row. The percentage arguments must be in [0,1], otherwise the function returns an error.
+        row. Similarly, slicePct(0, 1) would return the entire table (because row positions run from 0 to size-1).
+        The percentage arguments must be in range [0,1], otherwise the function returns an error.
```
chipkent: The description is still using the java function name.  Check carefully for other places where java is used.

chipkent (changes requested):

py/server/deephaven/table.py
```diff
@@ -2084,6 +2084,29 @@ def slice(self, start: int, stop: int) -> Table:
         except Exception as e:
             raise DHError(e, "table slice operation failed.") from e
 
+    def slice_pct(self, start_pct: float, end_pct: float) -> Table:
+        """Extracts a subset of a table by row percentages.
+
+        Returns a subset of table in the range [floor(start_pct * size_of_table), floor(end_pct * size_of_table)).
+        For example, for a table of size 10, slice_pct(0.1, 0.7) will return a subset from the second row to the seventh
+        row. Similarly, slice_pct(0, 1) would return the entire table (because row positions run from 0 to size-1).
+        The percentage arguments must be in range [0,1], otherwise the function returns an error.
```
jmao-denver, line 2093:
```suggestion
        row. Similarly, slice_pct(0, 1) would return the entire table (because row positions run from 0 to size - 1).
        The percentage arguments must be in range [0, 1], otherwise the function returns an error.
```

py/server/deephaven/table.py
```diff
@@ -2084,6 +2084,29 @@ def slice(self, start: int, stop: int) -> Table:
...
+
+        Args:
+            start_pct (float): the starting percentage point for rows to include in the result, range [0, 1]
+            end_pct (float): the ending percentage point for rows to include in the result, range [0, 1]
+
```
jmao-denver, line 2098: is end_pct supposed to be exclusive?

jmao-denver (dismissed):

chipkent (dismissed):

jmao-denver (approved):
LGTM

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2910
Reference: https://github.com/deephaven/deephaven.io/issues/2909

PR #4136, opened 2023-07-05
Added support for head() and tail() for blink tables

### LINKED ISSUE #2411: Make head and tail on a blink table act as if they have full history
As a user, I'd like to be able to expect that `Table.head()` and `Table.tail()` on blink tables act the same as they would on an append-only (full history) table derived from the stream. This makes sense semantically if we think of these as being like aggregations.

Tail can be easily done via the ring table implementation (`io.deephaven.engine.table.impl.sources.ring.RingTableTools#of(io.deephaven.engine.table.Table, int, boolean)).

Head can be built by implementing a `limit` parameter to `io.deephaven.engine.table.impl.StreamTableTools#streamToAppendOnlyTable`. For a stretch goal, we could add `Table.removeParentReference` and allow the stream to be shutdown if the head was the only user.

malhotrashivam:
Feature request, closes #2411

**Change description:** Before this change, head/tail operations on blink table would return the head/tail of the rows added to the blink table in the latest update cycle. After this change, head/tail would act the same as they would on an append-only (full history) table derived from the stream. For example, head(10) on a blink table would store the first 10 rows received by the blink table after the operation and then won't change. Similarly, tail(10) would store the latest 10 rows received by the blink table.
Also, after this change, headPct(), taillPct() and slice() will not be supported for blink tables and will return an error.

**Documentation change:**  We need to provide the above description in the documentation pages of blink tables and these operations. These are some places I could find: [Blink table](https://deephaven.io/core/docs/conceptual/kafka-basic-terms/#blink), [head Python](https://deephaven.io/core/docs/reference/table-operations/filter/head/#docusaurus_skipToContent_fallback), [head Java](https://deephaven.io/core/groovy/docs/reference/table-operations/filter/head/#docusaurus_skipToContent_fallback), [tail Python](https://deephaven.io/core/docs/reference/table-operations/filter/tail/#docusaurus_skipToContent_fallback), [tail Java](https://deephaven.io/core/groovy/docs/reference/table-operations/filter/tail/#docusaurus_skipToContent_fallback)

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
@@ -58,7 +58,6 @@ public class BlinkTableTools {
         final UpdateGraph updateGraph = blinkTable.getUpdateGraph();
         try (final SafeCloseable ignored = ExecutionContext.getContext().withUpdateGraph(updateGraph).open()) {
             return internalBlinkToAppendOnly(blinkTable, true, rowLimit);
-            // How to do it in a better way in Java?
         }
     }
 
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
@@ -38,7 +38,8 @@ public class BlinkTableTools {
     public static Table blinkToAppendOnly(final Table blinkTable) {
         final UpdateGraph updateGraph = blinkTable.getUpdateGraph();
         try (final SafeCloseable ignored = ExecutionContext.getContext().withUpdateGraph(updateGraph).open()) {
-            return internalBlinkToAppendOnly(blinkTable, false, Long.MAX_VALUE);
+            // Setting the row limit as maximum allowed value
+            return internalBlinkToAppendOnly(blinkTable, Long.MAX_VALUE);
         }
     }
 
@@ -57,12 +58,11 @@ public class BlinkTableTools {
         }
         final UpdateGraph updateGraph = blinkTable.getUpdateGraph();
         try (final SafeCloseable ignored = ExecutionContext.getContext().withUpdateGraph(updateGraph).open()) {
-            return internalBlinkToAppendOnly(blinkTable, true, rowLimit);
+            return internalBlinkToAppendOnly(blinkTable, rowLimit);
         }
     }
 
-    // TODO How do we handle default arguments. Is there a better way to do this
-    private static Table internalBlinkToAppendOnly(final Table blinkTable, boolean hasMaxRowLimit, long maxRowLimit) {
+    private static Table internalBlinkToAppendOnly(final Table blinkTable, long maxRowLimit) {
         return QueryPerformanceRecorder.withNugget("blinkToAppendOnly", () -> {
             if (!isBlink(blinkTable)) {
                 throw new IllegalArgumentException("Input is not a blink table!");
@@ -99,29 +99,16 @@ public class BlinkTableTools {
                                     (WritableColumnSource<?>) ReinterpretUtils.maybeConvertToPrimitive(newColumn);
                         }
 
-                        // I am a bit confused between different row types
-                        // TODO Get feedback from Larry
-                        final TrackingWritableRowSet rowSet;
+                        final RowSet baseRowSet =
+                                (usePrev ? baseBlinkTable.getRowSet().prev() : baseBlinkTable.getRowSet());
                         final RowSet useRowSet;
-                        if (usePrev) {
-                            try (RowSet baseRowSet = baseBlinkTable.getRowSet().copyPrev()) {
-                                if (hasMaxRowLimit && baseRowSet.size() > maxRowLimit) {
-                                    useRowSet = baseRowSet.subSetByPositionRange(0, maxRowLimit);
-                                } else {
-                                    useRowSet = baseRowSet;
-                                }
-                                rowSet = RowSetFactory.flat(useRowSet.size()).toTracking();
-                                ChunkUtils.copyData(sourceColumns, useRowSet, destColumns, rowSet, usePrev);
-                            }
+                        if (baseRowSet.size() > maxRowLimit) {
+                            useRowSet = baseRowSet.subSetByPositionRange(0, maxRowLimit);
                         } else {
-                            if (hasMaxRowLimit && baseBlinkTable.getRowSet().size() > maxRowLimit) {
-                                useRowSet = baseBlinkTable.getRowSet().subSetByPositionRange(0, maxRowLimit);
-                            } else {
-                                useRowSet = baseBlinkTable.getRowSet();
-                            }
-                            rowSet = RowSetFactory.flat(useRowSet.size()).toTracking();
-                            ChunkUtils.copyData(sourceColumns, useRowSet, destColumns, rowSet, usePrev);
+                            useRowSet = baseRowSet;
                         }
+                        final TrackingWritableRowSet rowSet = RowSetFactory.flat(useRowSet.size()).toTracking();
+                        ChunkUtils.copyData(sourceColumns, useRowSet, destColumns, rowSet, usePrev);
 
                         final QueryTable result = new QueryTable(rowSet, columns);
                         result.setRefreshing(true);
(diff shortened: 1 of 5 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/SliceLikeOperation.java
```diff
@@ -92,31 +95,40 @@ protected long getLastPositionExclusive() {
...
-            if (parent.isAppendOnly()) {
-                resultTable.setAttribute(Table.APPEND_ONLY_TABLE_ATTRIBUTE, true);
+        } else {
+            if (operation.equals("head")) {
+                long size = getLastPositionExclusive();
+                resultTable = (QueryTable) BlinkTableTools.blinkToAppendOnly(parent, size);
+            } else if (operation.equals("tail")) {
+                long size = -1 * getFirstPositionInclusive();
+                resultTable = (QueryTable) RingTableTools.of(parent, (int) size); // TODO Do I need to do something for
+                // this type cast from long to int?
```
malhotrashivam, line 131: I am not sure how to handle this type case without making a big change in RingTableTools.

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -130,62 +117,49 @@ private static Table internalBlinkToAppendOnly(final Table blinkTable, boolean h
...
                                 }
-                            }, result);
-                        }
+                                RowSet newRowSet = upstream.added();
+                                long newRowsSize = newRowSet.size();
+                                if (newRowsSize == 0) {
+                                    return;
+                                }
+                                final long currentSize = rowSet.size();
+                                if (currentSize + newRowsSize >= maxRowLimit) {
```
lbooker42, line 135: I was going to suggest some code to protect against overflowing a Long but that seems pretty silly given the size of a table that would overflow.  So ignore this comment.

engine/table/src/main/java/io/deephaven/engine/table/impl/SliceLikeOperation.java
lbooker42: You should do a conversion that will cause exception on overflow.  `Math.toIntExact()` would be a reasonable choice.

lbooker42 (commented):
This is looking good.  Let's ask @rcaudy about disconnecting the listener when the table is full (assuming you haven't).

engine/table/src/main/java/io/deephaven/engine/table/impl/SliceLikeOperation.java
malhotrashivam: Exactly what I was looking for, thanks.

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -611,6 +611,9 @@ public TreeTable tree(String idColumn, String parentColumn) {
     public Table slice(final long firstPositionInclusive, final long lastPositionExclusive) {
         final UpdateGraph updateGraph = getUpdateGraph();
         try (final SafeCloseable ignored = ExecutionContext.getContext().withUpdateGraph(updateGraph).open()) {
+            if (isBlink()) {
+                throw unsupportedForBlinkTables("slice");
+            }
```
rcaudy, line 616: I would suggest moving this up, above the try.

engine/table/src/main/java/io/deephaven/engine/table/impl/SliceLikeOperation.java
```diff
@@ -92,31 +95,39 @@ protected long getLastPositionExclusive() {
...
-
-        if (operation.equals("headPct")) {
-            // headPct has a floating tail, so we can only propagate if append-only
-            if (parent.isAppendOnly()) {
-                resultTable.setAttribute(Table.ADD_ONLY_TABLE_ATTRIBUTE, true);
-                resultTable.setAttribute(Table.APPEND_ONLY_TABLE_ATTRIBUTE, true);
+        if (!parent.isBlink()) {
+            final TrackingRowSet parentRowSet = parent.getRowSet();
+            final TrackingRowSet resultRowSet =
+                    computeSliceIndex(usePrev ? parentRowSet.prev() : parentRowSet).toTracking();
```
rcaudy, line 101: Can you do me a favor and re-name `computeSliceIndex` to `computeSliceRowSet`.

engine/table/src/main/java/io/deephaven/engine/table/impl/SliceLikeOperation.java
```diff
@@ -134,6 +145,10 @@ public void onUpdate(TableUpdate upstream) {
     }
 
     private void onUpdate(final TableUpdate upstream) {
+        if (parent.isBlink()) {
```
rcaudy, line 148: We need to come up with a strategy to avoid creating this listener. We *could* create some kind of "no-op" listener. That said, it might be better to short-circuit up in `QueryTable` and never hit `SliceLikeOperation`.

engine/table/src/test/java/io/deephaven/engine/table/impl/BlinkTableAggregationTest.java
```diff
@@ -34,7 +35,7 @@
 import static io.deephaven.api.agg.Aggregation.*;
 
 /**
- * Unit tests that exercise special aggregation semantics for blink tables.
+ * Unit tests that exercise operations (like aggregations) which are specialized for blink tables.
```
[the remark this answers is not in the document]
rcaudy: Sure. `BlinkTableSemanticsTest` or similar.

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -41,11 +38,31 @@ public class BlinkTableTools {
...
+    /**
+     * Convert a Blink Table to an in-memory append only table with a limit on maximum row count. Any updates beyond
+     * that limit won't be appended to the table.
+     *
+     * @param blinkTable The input blink table
+     * @param rowLimit The maximum number of rows in the append-only table
+     * @return An append-only in-memory table representing all data encountered in the blink table across all cycles
+     *         till maximum row count
+     */
+    public static Table blinkToAppendOnly(final Table blinkTable, long rowLimit) {
```
rcaudy, line 55: You might consider renaming `rowLimit` to `sizeLimit` or similar; we want to be very clear that this isn't about the maximum row key to copy or anything like that.

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -102,24 +117,31 @@ private static Table internalBlinkToAppendOnly(final Table blinkTable) {
...
+                                if (newRowsSize == 0) {
                                     return;
                                 }
                                 final long currentSize = rowSet.size();
-                                columns.values().forEach(c -> c.ensureCapacity(currentSize + newRows));
-
-                                final RowSet newRange = RowSetFactory.fromRange(currentSize, currentSize + newRows - 1);
+                                if (currentSize + newRowsSize >= maxRowLimit) {
+                                    newRowsSize = (maxRowLimit - currentSize);
+                                    newRowSet = upstream.added().subSetByPositionRange(0, newRowsSize);
```
rcaudy, line 137: We should be sure to close `newRowSet` if we hit this branch.

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -128,6 +150,13 @@ public void onUpdate(TableUpdate upstream) {
                                 downstream.modifiedColumnSet = ModifiedColumnSet.EMPTY;
                                 downstream.shifted = RowSetShiftData.EMPTY;
                                 result.notifyListeners(downstream);
+
+                                Assert.leq(totalSize, "totalSize", maxRowLimit, "maxRowLimit");
```
rcaudy, line 154: Would be better to assert *before* notifying downstream listeners.

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -128,6 +150,13 @@ public void onUpdate(TableUpdate upstream) {
...
+                                if (totalSize == maxRowLimit) {
+                                    // No more rows can be appended, so remove the listener
+                                    blinkTable.removeUpdateListener(this);
+                                    return;
```
rcaudy, line 158: Unnecessary return

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -128,6 +150,13 @@ public void onUpdate(TableUpdate upstream) {
                                 downstream.modifiedColumnSet = ModifiedColumnSet.EMPTY;
                                 downstream.shifted = RowSetShiftData.EMPTY;
                                 result.notifyListeners(downstream);
+
+                                Assert.leq(totalSize, "totalSize", maxRowLimit, "maxRowLimit");
+                                if (totalSize == maxRowLimit) {
+                                    // No more rows can be appended, so remove the listener
+                                    blinkTable.removeUpdateListener(this);
```
rcaudy, line 157:
```suggestion
                                    forceReferenceCountToZero();
```

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2891
Reference: https://github.com/deephaven/deephaven.io/issues/2892

PR #4142, opened 2023-07-06
KafkaTools refactor to allow for alternative StreamConsumers

### LINKED ISSUE #3968: KakfaTools consumeToResult Enhancements
As a developer, I want to use Deephaven community Kafka ingestion, but bypass the blink tables accessing the StreamPublishers (ConsumerRecordToStreamPublisherAdapter) so that I can take advantage of the key and value parsing into Deephaven Community chunks to do non-table related things with the data (like write it to disk in a columnar format).

rcaudy:
* Refactor KafkaTools `consumeToX` to allow for `consume` to directly use a supplied `StreamConsumer`
* Fix some error handling and liveness management issues in the Kafka infrastructure

Fixes #3968


engine/table/src/main/java/io/deephaven/stream/StreamPublisher.java
```diff
@@ -30,4 +32,10 @@ public interface StreamPublisher {
      * {@link StreamConsumer#accept(WritableChunk[]) accept} method.
      */
     void flush();
+
+    /**
+     * Shutdown this StreamPublisher. Implementations should stop publishing new data and release any related resources
+     * as soon as practicable.
+     */
+    void shutdown();
```
devinrsmith, line 40: I like this functionality. That said, is there a reason to prefer "shutdown" over "close()" / AutoCloseable?

plugin/gc-app/src/main/java/io/deephaven/app/GcPoolsPublisher.java
```diff
@@ -139,18 +139,21 @@ private void maybeFlushAndGrow(int initialSize, int poolSize) {
...
+        chunks = StreamChunkUtils.makeChunksForDefinition(DEFINITION, chunkSize);
     }
 
     public void acceptFailure(Throwable e) {
         consumer.acceptFailure(e);
     }
 
+    @Override
+    public void shutdown() {}
+
```
devinrsmith, line 156: I'm wondering, should it be our process here to notify downstream `consumer` when we are `shutdown`? Either, we can extend `StreamConsumer` interface in the same way, or we can do `consumer.acceptFailure(new PublisherShutdown())` or something. (Ditto for all other StreamPublisher impls).

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -2189,8 +2299,8 @@ public long consume(List<? extends ConsumerRecord<?, ?>> consumerRecords) {
         }
 
         @Override
-        public void acceptFailure(@NotNull Throwable cause) {
-            streamToBlinkTableAdapter.acceptFailure(cause);
+        public void acceptFailure(@NotNull final Throwable cause) {
+            adapter.propagateFailure(cause);
```
devinrsmith, line 2303: Love it - makes this much cleaner. Really, `StreamPublisher` is an interface that is _meant_ to be extended with proper "here is new data" / "here is an error". We _might_ consider adding and documenting `propogateFailure` on `StreamPublisher` itself if it makes sense for all impls.

(In a similar way, see other comment about if shutdown() should "propogate" to consumer.)

engine/table/src/main/java/io/deephaven/stream/StreamToBlinkTableAdapter.java
```diff
@@ -277,50 +212,49 @@ public Table table() {
...
     @Override
     public void close() {
-        if (!alive) {
-            return;
-        }
-        synchronized (this) {
-            if (!alive) {
-                return;
-            }
-            alive = false;
```
devinrsmith, line 287: I'm assuming this old pattern works, you just wanted something a bit prettier?

engine/table/src/main/java/io/deephaven/stream/StreamToBlinkTableAdapter.java
```diff
@@ -68,11 +75,11 @@ public class StreamToBlinkTableAdapter extends ReferenceCountedLivenessNode
     private ChunkColumnSource<?>[] prevChunkSources;
 
     /** A list of failures that have occurred. */
-    private List<Throwable> enqueuedFailure;
+    private List<Throwable> enqueuedFailures;
```
devinrsmith, line 78: This could be final and allocated in the constructor, and instead the checkers could check isEmpty() instead of null.

engine/table/src/main/java/io/deephaven/stream/StreamToBlinkTableAdapter.java
```diff
@@ -277,50 +212,49 @@ public Table table() {
...
-
     @Override
     public void run() {
+        synchronized (this) {
+            // If we have an enqueued failure we want to process it first, before we allow the streamPublisher to flush
+            // itself.
+            if (enqueuedFailures != null) {
+                deliverFailure(MultiException.maybeWrapInMultiException(
+                        "Multiple errors encountered while ingesting stream",
+                        enqueuedFailures.toArray(new Throwable[0])));
```
devinrsmith, line 232: Looks like you can call this w/ the List, no need to toArray.

engine/table/src/main/java/io/deephaven/stream/StreamToBlinkTableAdapter.java
```diff
@@ -277,50 +212,49 @@ public Table table() {
...
-        close();
-    }
-
     @Override
     public void run() {
+        synchronized (this) {
+            // If we have an enqueued failure we want to process it first, before we allow the streamPublisher to flush
+            // itself.
+            if (enqueuedFailures != null) {
+                deliverFailure(MultiException.maybeWrapInMultiException(
```
devinrsmith, line 230: Do we need to concern ourselves w/ the potential for "double delivery" of these errors? I'm not sure the exact semantics in these layers, but we could easily guard against that if necessary.

engine/table/src/main/java/io/deephaven/stream/StreamToBlinkTableAdapter.java
```diff
@@ -277,50 +212,49 @@ public Table table() {
...
 
-    private void deliverUpdate(final TableUpdate downstream) {
+    private void deliverUpdate(@Nullable final TableUpdate downstream) {
         final QueryTable localTable = tableRef.get();
         if (localTable != null) {
             if (downstream != null) {
-                localTable.notifyListeners(downstream);
+                try {
+                    localTable.notifyListeners(downstream);
+                } catch (Exception e) {
```
devinrsmith, line 254: I notice here, and other places in this impl, we are specifically concerning ourselves w/ catching `Exception`, even though the downstream code is more general to accept `Throwable`. Practically, this means that means that `OutOfMemoryError`s won't get caught - I wonder, is this intentional, or just how it's been done before? Should we consider catching Throwable?

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -1312,80 +1331,183 @@ public static PartitionedTable consumeToPartitionedTable(
...
-            final Pair<StreamToBlinkTableAdapter, ConsumerRecordToStreamPublisherAdapter> singleAdapterPair =
-                    adapterFactory.get();
-            final Table blinkTable = singleAdapterPair.getFirst().table();
-            final Table result = tableType.walk(new BlinkTableOperation(blinkTable));
-            final IntFunction<KafkaStreamConsumer> consumerFactory = (final int partition) -> {
-                singleAdapterPair.getFirst().setShutdownCallback(() -> kafkaIngesterHolder.getValue().shutdown());
-                return new SimpleKafkaStreamConsumer(singleAdapterPair.getSecond(), singleAdapterPair.getFirst());
-            };
-            return new Pair<>(result, consumerFactory);
+        <T, V extends Visitor<T>> T walk(V visitor);
```
devinrsmith, line 1427: I don't think there is value in having generic `V` here:

```java
<T> T walk(Visitor<T> visitor);
```

is _good_ enough, I think?

devinrsmith (commented):
First pass - skipped over a large part of the KafkaTools guts b/c I feel like I add less value there. Going to do some testing with it.

engine/table/src/main/java/io/deephaven/stream/StreamToBlinkTableAdapter.java
rcaudy: Typically, the idea is that `Error` is uncaught and allowed to kill the process.

engine/table/src/main/java/io/deephaven/stream/StreamToBlinkTableAdapter.java
rcaudy: No, because we close immediately afterwards, ensuring that our `run()` method will not be called again. Otherwise, it would be a concern.

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
rcaudy: I was copying your pattern on `TableType`. Would you like me to change both, or am I missing some subtelty?

engine/table/src/main/java/io/deephaven/stream/StreamToBlinkTableAdapter.java
rcaudy: I thought it would be better to eliminate the lock call.

engine/table/src/main/java/io/deephaven/stream/StreamPublisher.java
rcaudy: The goal here was just to move shutdown ownership up a layer, so we don't have to have every consumer implement a `setShutdownCallback` method. I don't really want it called externally, and I prefer to save `AutoCloseable` or `SafeCloseable` implementations for resource cases.

plugin/gc-app/src/main/java/io/deephaven/app/GcPoolsPublisher.java
rcaudy: I could add documentation to that effect...
"This method should be invoked only by the eventual consumer, or by the constructing thread if no consumer will ever be registered".

cpwright (dismissed):

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
rcaudy: Generally, the `StreamPublisher` interface is an API for the `StreamConsumer` to interact with. I don't want to add propagation to that, it's an implementation detail of the publisher.

engine/table/src/main/java/io/deephaven/stream/StreamConsumer.java
```diff
@@ -11,14 +11,12 @@
  * Chunk-oriented consumer for streams of data.
  */
 public interface StreamConsumer extends StreamFailureConsumer {
+
     /**
-     * <p>
```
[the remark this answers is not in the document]
rcaudy: No, the leading <P> does nothing as far as I can tell.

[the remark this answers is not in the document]
rcaudy: Screenshot 2023-07-07 at 11 44 13 AM

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaStreamPublisher.java
```diff
@@ -74,63 +79,57 @@ private KafkaStreamPublisher(
     }
 
     public static ConsumerRecordToStreamPublisherAdapter make(
```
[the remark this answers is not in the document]
rcaudy: No need, regarding Python. This is currently only ever used from Java, internal to KafkaTools. Really, the parameters struct was just to avoid repeating a bunch of ugly code, and I had IntelliJ generate it automatically. I tried the builder on for size initially, but it was even more gratuitous code. Let me take one more crack at seeing if it makes things better.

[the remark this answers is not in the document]
rcaudy: OK, I added a bunch of code and a builder. It's cleaner, but I didn't clean up every bit of badness, either.

cpwright (approved):

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
```diff
@@ -355,11 +355,16 @@ private boolean pollOnce(final Duration timeout) {
             } catch (Throwable ex) {
                 ++messagesWithErr;
                 log.error().append(logPrefix).append("Exception while processing Kafka message:").append(ex).endl();
+                /*
```
devinrsmith, line 358: Afaict, this code block is never getting hit due to io.deephaven.kafka.KafkaTools.SimpleKafkaRecordConsumer#consume.

Example stack:
```
java.lang.IllegalArgumentException: argument "content" is null
        at com.fasterxml.jackson.databind.ObjectMapper._assertNotNull(ObjectMapper.java:4980)
        at com.fasterxml.jackson.databind.ObjectMapper.readTree(ObjectMapper.java:3217)
        at io.deephaven.kafka.ingest.JsonNodeUtil.makeJsonNode(JsonNodeUtil.java:29)
        at io.deephaven.kafka.KafkaTools.lambda$static$5(KafkaTools.java:2094)
        at io.deephaven.kafka.ingest.KafkaStreamPublisher.consumeRecords(KafkaStreamPublisher.java:270)
        at io.deephaven.kafka.KafkaTools$SimpleKafkaRecordConsumer.consume(KafkaTools.java:2302)
        at io.deephaven.kafka.ingest.KafkaIngester.pollOnce(KafkaIngester.java:354)
        at io.deephaven.kafka.ingest.KafkaIngester.consumerLoop(KafkaIngester.java:284)
```

rcaudy: Noted in the new issue.

devinrsmith (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2878
Reference: https://github.com/deephaven/deephaven.io/issues/2879

PR #4144, opened 2023-07-07
Proposed fix for scope snapshot diff gaps

### LINKED ISSUE #4145: ScriptSession API to ensure snapshots/diffs of bindings are always consistent.
Original title/topic: Generalize script session "deltas", probably by letting the script session entirely control snapshots internally, rather than expose this directly, such as by returning changes when running script code. This will also deal with some potential races, making it the script session implementation's responsibility to not lose track of changes.

Split off from #2453 to be its own issue.

This was found to be broken, where assigning/nulling a binding rapidly can break the application service's internal state, by missing changes and emitting inconsistent updates:
```python
from deephaven_server import Server
s = Server()
s.start()

import time

while True:
    t = time_table("PT1m")
    time.sleep(.04)
    t = None
    time.sleep(.04)
```
Usually takes 10s-1minute to fail.

rcaudy:
Fixes #4145

### DIFF
--- a/Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
+++ b/Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
@@ -280,24 +280,25 @@ public class PythonDeephavenSession extends AbstractScriptSession<PythonSnapshot
 
     @Override
     public synchronized void setVariable(String name, @Nullable Object newValue) {
-        try (PythonSnapshot fromSnapshot = takeSnapshot()) {
-            final PyDictWrapper globals = scope.mainGlobals();
-            if (newValue == null) {
-                try {
-                    globals.delItem(name);
-                } catch (KeyError key) {
-                    // ignore
-                }
-            } else {
-                if (!(newValue instanceof PyObject)) {
-                    newValue = PythonObjectWrapper.wrap(newValue);
-                }
-                globals.setItem(name, newValue);
+        // Observe any external changes that are not yet recorded
+        observeScopeChanges();
+
+        final PyDictWrapper globals = scope.mainGlobals();
+        if (newValue == null) {
+            try {
+                globals.delItem(name);
+            } catch (KeyError key) {
+                // ignore
             }
-            try (PythonSnapshot toSnapshot = takeSnapshot()) {
-                applyDiff(fromSnapshot, toSnapshot, null);
+        } else {
+            if (!(newValue instanceof PyObject)) {
+                newValue = PythonObjectWrapper.wrap(newValue);
             }
+            globals.setItem(name, newValue);
         }
+
+        // Observe changes from this "setVariable" (potentially capturing external changes from other threads)
+        observeScopeChanges();
     }
 
     @Override
--- a/Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
+++ b/Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
@@ -62,10 +62,10 @@ public class PythonDeephavenSession extends AbstractScriptSession<PythonSnapshot
 
     public static String SCRIPT_TYPE = "Python";
 
-    private final PythonScriptSessionModule module;
-    private final ScriptFinder scriptFinder;
     private final PythonEvaluator evaluator;
     private final PythonScope<PyObject> scope;
+    private final PythonScriptSessionModule module;
+    private final ScriptFinder scriptFinder;
 
     /**
      * Create a Python ScriptSession.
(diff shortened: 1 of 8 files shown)


py/embedded-server/java-runtime/src/main/java/io/deephaven/python/server/EmbeddedServer.java
```diff
@@ -115,24 +114,10 @@ public void start() throws Exception {
...
-                // We are limited in how we can track external changes, and the web IDE has made this change and
-                // already applied it.
-                // Take a fresh snapshot right away to continue polling
-                nextSnapshot = scriptSession.snapshot();
-            } else {
-                throw e;
-            }
-        }
-        ScriptSession.SnapshotScope s = nextSnapshot;
+    private void checkGlobals(ScriptSession scriptSession) {
```
niloc132, line 117: This should probably move into ScriptSession, and make the delay configurable?

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -96,31 +98,25 @@ public ExecutionContext getExecutionContext() {
...
-            applyDiff(beforeSnapshot, afterSnapshot, null);
+    private synchronized Changes observeScopeChangesInternal(@Nullable final RuntimeException evaluationError) {
+        final S beforeSnapshot = lastSnapshot;
+        lastSnapshot = takeSnapshot();
+        try (beforeSnapshot) {
+            return applyDiff(beforeSnapshot, lastSnapshot, evaluationError);
         }
     }
 
+    protected interface Snapshot extends SafeCloseable {
```
niloc132, line 118: Thought (that shouldn't preclude merging): If this is protected and we're removing the generics and other members, do we need generics anymore? As it is protected and not accessible from any non-protected methods, I don't think we need to consider it at all.

Consider: initialSnapshot only runs in the constructor anyway (obligatory "don't call abstract/non-final methods in a constructor), and takeSnapshot()/applyDiff() is only called by observeScopeChangesInternal.. I haven't gone through the exercise yet, but it might be possible to get rid of most of this, pushing it down to just part of the Snapshot interface/implementation, and let the script session not care/cast about what type it has (but the subclass would work out specifics).

niloc132 (dismissed):
Please link to https://github.com/deephaven/deephaven-core/issues/4145 for this.

Verified that the changed APIs won't affect DnD.

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
rcaudy: Let's defer this to a subsequent issue.

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -280,24 +280,25 @@ public boolean hasVariableName(String name) {
...
-                    globals.delItem(name);
-                } catch (KeyError key) {
-                    // ignore
-                }
-            } else {
-                if (!(newValue instanceof PyObject)) {
-                    newValue = PythonObjectWrapper.wrap(newValue);
-                }
-                globals.setItem(name, newValue);
+        // Observe any external changes that are not yet recorded
```
niloc132, line 283: Matching change needs to happen to the groovy script session, the notifyVariableChange method is called from there (and should be deleted), since it will notify the app service of a new variable... racing the current snapshot.

jmao-denver (approved):
LGTM

server/src/main/java/io/deephaven/server/runner/DeephavenApiServer.java
```diff
@@ -167,6 +171,13 @@ public DeephavenApiServer run() throws IOException, ClassNotFoundException, Time
         return this;
     }
 
+    private void checkGlobals(ScriptSession scriptSession) {
+        scriptSession.observeScopeChanges();
+        scheduler.runAfterDelay(100, () -> {
```
niloc132, line 176: should we parameterize this, so that in case 100ms breaks things, we can turn it down/off by configuration?

niloc132:
My testing seems to show that this does the job. First, running the python script in the linked issue (ran three times for 5+ mins each), then these two script commands:

## Python
```python
from deephaven.execution_context import get_exec_ctx
@get_exec_ctx()
def doStuff():
    global t
    t = time_table("PT1s").update("I=i")
import threading
thread = threading.Thread(target=doStuff)
```
before this, the table wouldn't show up, after the fix it correctly shows up in the table dropdown in the UI right away.

## Groovy
```groovy
ec = io.deephaven.engine.context.ExecutionContext.getContext()
Thread.start({ -> ec.open(); getBinding().setVariable("asdf", timeTable("PT1s"))  })
```
Before this fix sometimes this would cause an error, sometimes it would just not show the table in the dropdown, afterwards the table shows up in the UI right away.

niloc132 (approved):

jmao-denver (approved):

PR #4139, opened 2023-07-06
Kafka raw spec

devinrsmith:

PR #4149, opened 2023-07-07
Kafka raw spec python wrappers

devinrsmith:
Follow-up to https://github.com/deephaven/deephaven-core/pull/4139

### DIFF
--- a/py/server/deephaven/stream/kafka/consumer.py
+++ b/py/server/deephaven/stream/kafka/consumer.py
@@ -19,6 +19,7 @@ from deephaven.table import Table, PartitionedTable
 _JKafkaTools = jpy.get_type("io.deephaven.kafka.KafkaTools")
 _JKafkaTools_Consume = jpy.get_type("io.deephaven.kafka.KafkaTools$Consume")
 _JPythonTools = jpy.get_type("io.deephaven.integrations.python.PythonTools")
+_JColumnHeader = jpy.get_type("io.deephaven.qst.column.header.ColumnHeader")
 ALL_PARTITIONS = _JKafkaTools.ALL_PARTITIONS
 
 SEEK_TO_BEGINNING = _JKafkaTools.SEEK_TO_BEGINNING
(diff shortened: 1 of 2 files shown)


devinrsmith:
I'm not a fan of DType, and I think we might want to more closely match what QST does...

jmao-denver:
Is there a test that shows a typical use case?

rcaudy (commented):
.

devinrsmith:
No tests right now. There was some manual testing I did. Ultimately, this is to support custom kakfa deserializers. I've tested it with `io.confluent.kafka.serializers.protobuf.KafkaProtobufDeserializer`.

py/server/deephaven/stream/kafka/consumer.py
```diff
@@ -388,3 +389,28 @@ def simple_spec(col_name: str, data_type: DType = None) -> KeyValueSpec:
...
     except Exception as e:
         raise DHError(e, "failed to create a Kafka key/value spec") from e
+
+
+def raw_spec(col_name: str, data_type: str, deserializer: str) -> KeyValueSpec:
+    """Creates a raw spec with explicit expected data_type and kafka deserializer.
+
+    Args:
+        col_name (str): the Deephaven column name
+        data_type (str): the column data type class
```
chipkent, line 399: Is there a reason this is a string and not a `DType`?  I think I understand the reasoning, but I want to confirm.

py/server/deephaven/stream/kafka/consumer.py
```diff
@@ -388,3 +389,28 @@ def simple_spec(col_name: str, data_type: DType = None) -> KeyValueSpec:
...
+    Returns:
+        a KeyValueSpec
+
+    Raises:
+        DHError
+    """
+    try:
+        return KeyValueSpec(
+            j_spec=_JKafkaTools_Consume.rawSpec(
+                _JColumnHeader.of(col_name, jpy.get_type(data_type).jclass),
```
chipkent, line 411: Same comment on string vs `DType` here.

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -233,3 +233,24 @@ def simple_spec(col_name: str) -> KeyValueSpec:
...
+
+    Raises:
+        DHError
+    """
+    try:
+        return KeyValueSpec(
+            _JKafkaTools_Produce.simpleSpec(col_name, jpy.get_type(serializer).jclass)
+        )
+    except Exception as e:
+        raise DHError(e, "failed to create a Kafka key/value spec.") from e
```
chipkent, line 256: Unit tests are missing.

chipkent (changes requested):

PR #4151, opened 2023-07-07
Add optional debug logging to cpp client.

jcferretti:
Sample output follows.  Note below we get mixed lines from the gRPC DEBUG log and our own; for an example
of our own look for a line like

```
D0707 18:58:24.601635804 1824393 server.h:282]               0x564510c70690: Sending RPC io::deephaven::proto::backplane::grpc::FetchTableRequest at time 2023-07-07T18:58:24.601-0400.` 
```

below.

==

```
cfs@coipo 18:58:06 ~/dh/oss1/deephaven-core/cpp-client/deephaven/build
$ (cd ../../../cpp-examples/demos/build && make -j32 && GRPC_VERBOSITY=DEBUG ./chapter3)
Consolidate compiler generated dependencies of target dhcore
[ 33%] Built target dhcore
[ 91%] Built target client
[ 94%] Built target chapter3
[ 97%] Built target chapter2
[100%] Built target chapter1
I0707 18:58:13.414364967 1824393 ev_epoll1_linux.cc:121]     grpc epoll fd: 3
D0707 18:58:13.414447713 1824393 ev_posix.cc:171]            Using polling engine: epoll1
D0707 18:58:13.414510511 1824393 lb_policy_registry.cc:42]   registering LB policy factory for "grpclb"
D0707 18:58:13.414521892 1824393 lb_policy_registry.cc:42]   registering LB policy factory for "rls_experimental"
D0707 18:58:13.414529176 1824393 lb_policy_registry.cc:42]   registering LB policy factory for "priority_experimental"
D0707 18:58:13.414538714 1824393 lb_policy_registry.cc:42]   registering LB policy factory for "weighted_target_experimental"
D0707 18:58:13.414547350 1824393 lb_policy_registry.cc:42]   registering LB policy factory for "pick_first"
D0707 18:58:13.414554193 1824393 lb_policy_registry.cc:42]   registering LB policy factory for "round_robin"
D0707 18:58:13.414561006 1824393 lb_policy_registry.cc:42]   registering LB policy factory for "ring_hash_experimental"
D0707 18:58:13.414567438 1824393 dns_resolver_ares.cc:454]   Using ares dns resolver
D0707 18:58:13.414604808 1824393 certificate_provider_registry.cc:33] registering certificate provider factory for "file_watcher"
D0707 18:58:13.414612573 1824393 lb_policy_registry.cc:42]   registering LB policy factory for "cds_experimental"
D0707 18:58:13.414620448 1824393 lb_policy_registry.cc:42]   registering LB policy factory for "xds_cluster_impl_experimental"
D0707 18:58:13.414628613 1824393 lb_policy_registry.cc:42]   registering LB policy factory for "xds_cluster_resolver_experimental"
D0707 18:58:13.414635847 1824393 lb_policy_registry.cc:42]   registering LB policy factory for "xds_cluster_manager_experimental"
D0707 18:58:13.421390878 1824393 server.h:282]               0x564510c70690: Sending RPC io::deephaven::proto::backplane::script::grpc::StartConsoleRequest at time 2023-07-07T18:58:13.421-0400.
*** CHAPTER 3: MAIN MENU ***
1 - print full table
2 - print diffs with library
3 - process diffs

Please select 1-3: 1
Please enter the table name: demo1
D0707 18:58:24.601635804 1824393 server.h:282]               0x564510c70690: Sending RPC io::deephaven::proto::backplane::grpc::FetchTableRequest at time 2023-07-07T18:58:24.601-0400.
Press enter to unsubscribe: === The Full Table ===
[Row]   Timestamp
[0]     2023-07-07T22:58:20.000000000 UTC

=== The Full Table ===
[Row]   Timestamp
[0]     2023-07-07T22:58:20.000000000 UTC
[1]     2023-07-07T22:58:25.000000000 UTC

=== The Full Table ===
[Row]   Timestamp
[0]     2023-07-07T22:58:20.000000000 UTC
[1]     2023-07-07T22:58:25.000000000 UTC
[2]     2023-07-07T22:58:30.000000000 UTC

```

### DIFF
--- a/cpp-client/deephaven/dhcore/include/public/deephaven/dhcore/utility/utility.h
+++ b/cpp-client/deephaven/dhcore/include/public/deephaven/dhcore/utility/utility.h
@@ -5,9 +5,12 @@
 
 #include <chrono>
 #include <cstring>
+#include <cstdio>
 #include <iostream>
 #include <memory>
 #include <string>
+#include <sstream>
+#include <thread>
 #include <typeinfo>
 #include <vector>
 
@@ -263,6 +266,20 @@ TimePointToStr(
 template <class T> [[nodiscard]] std::string
 TypeName(const T& t) {
   return Demangle(typeid(t).name());
+}
+
+[[nodiscard]] inline std::string
+ObjectId(const std::string &classShortName, void* thisPtr) {
+  std::string id(classShortName + "[0x0000000000000000]");
+  snprintf(&id[classShortName.size() + 3], 16, "%p", thisPtr);
+  return id;
+}
+
+inline std::string
+ThreadIdToString(const std::thread::id tid) {
+  std::stringstream ss;
+  ss << tid;
+  return ss.str();
 }
 
 }  // namespace deephaven::dhcore::utility
--- a/cpp-client/deephaven/dhcore/include/public/deephaven/dhcore/utility/utility.h
+++ b/cpp-client/deephaven/dhcore/include/public/deephaven/dhcore/utility/utility.h
@@ -9,7 +9,6 @@
 #include <iostream>
 #include <memory>
 #include <string>
-#include <sstream>
 #include <thread>
 #include <typeinfo>
 #include <vector>
@@ -192,30 +191,7 @@ std::string formatDebugString(const char *func, const char *file, size_t line,
   ::deephaven::dhcore::utility::formatDebugString( \
     DEEPHAVEN_PRETTY_FUNCTION, __FILE__, __LINE__, MESSAGE)
 
-// https://stackoverflow.com/questions/281818/unmangling-the-result-of-stdtype-infoname
-template <typename T>
-constexpr std::string_view getTypeName() {
-#if defined(__clang__)
-  constexpr auto prefix = std::string_view{"[T = "};
-  constexpr auto suffix = "]";
-#elif defined(__GNUC__)
-  constexpr auto prefix = std::string_view{"with T = "};
-  constexpr auto suffix = "; ";
-#elif defined(__MSC_VER)
-  constexpr auto prefix = std::string_view{"get_type_name<"};
-  constexpr auto suffix = ">(void)";
-#else
-# error Unsupported compiler
-#endif
-
-  constexpr auto function = std::string_view{DEEPHAVEN_PRETTY_FUNCTION};
-
-  const auto start = function.find(prefix) + prefix.size();
-  const auto end = function.find(suffix);
-  const auto size = end - start;
-
-  return function.substr(start, size);
-}
+[[nodiscard]] std::string demangle(const char* name);
 
 template<typename DESTP, typename SRCP>
 DESTP verboseCast(const DebugInfo &debugInfo, SRCP ptr) {
@@ -227,7 +203,9 @@ DESTP verboseCast(const DebugInfo &debugInfo, SRCP ptr) {
   }
   typedef decltype(*std::declval<DESTP>()) destType_t;
   auto message = stringf("%o: Expected type %o. Got type %o",
-      debugInfo, getTypeName<destType_t>(), typeid(*ptr).name());
+      debugInfo,
+      demangle(typeid(destType_t).name()),
+      demangle(typeid(*ptr).name()));
   throw std::runtime_error(message);
 }
 
@@ -245,41 +223,25 @@ inline void trueOrThrow(const DebugInfo &debugInfo, bool value) {
 }
 
 [[nodiscard]] std::string
-EpochMillisToStr(std::int64_t epochMillis);
+epochMillisToStr(std::chrono::milliseconds::rep epochMillis);
 
-[[nodiscard]] inline std::int64_t
-TimePointToEpochMillis(
-    const std::chrono::time_point<std::chrono::system_clock> timePoint) {
-  using namespace std::chrono;
-  const milliseconds ms = duration_cast<milliseconds>(timePoint.time_since_epoch());
-  return ms.count();
-}
+[[nodiscard]] std::int64_t
+timePointToEpochMillis(
+    const std::chrono::time_point<std::chrono::system_clock> timePoint);
 
-[[nodiscard]] inline std::string
-TimePointToStr(
-    const std::chrono::time_point<std::chrono::system_clock> timePoint) {
-  return EpochMillisToStr(TimePointToEpochMillis(timePoint));
-}
-
-[[nodiscard]] std::string Demangle(const char* name);
+[[nodiscard]] std::string
+timePointToStr(
+    const std::chrono::time_point<std::chrono::system_clock> timePoint);
 
 template <class T> [[nodiscard]] std::string
-TypeName(const T& t) {
-  return Demangle(typeid(t).name());
+typeName(const T& t) {
+  return demangle(typeid(t).name());
 }
+
+[[nodiscard]] std::string
+objectId(const std::string &classShortName, void* thisPtr);
 
 [[nodiscard]] inline std::string
-ObjectId(const std::string &classShortName, void* thisPtr) {
-  std::string id(classShortName + "[0x0000000000000000]");
-  snprintf(&id[classShortName.size() + 3], 16, "%p", thisPtr);
-  return id;
-}
-
-inline std::string
-ThreadIdToString(const std::thread::id tid) {
-  std::stringstream ss;
-  ss << tid;
-  return ss.str();
-}
+threadIdToString(std::thread::id tid);
 
 }  // namespace deephaven::dhcore::utility
--- a/cpp-client/deephaven/dhcore/include/public/deephaven/dhcore/utility/utility.h
+++ b/cpp-client/deephaven/dhcore/include/public/deephaven/dhcore/utility/utility.h
@@ -3,10 +3,12 @@
  */
 #pragma once
 
+#include <chrono>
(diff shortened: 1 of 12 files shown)


cpp-client/deephaven/dhcore/include/public/deephaven/dhcore/utility/utility.h
```diff
@@ -238,4 +243,43 @@ inline void trueOrThrow(const DebugInfo &debugInfo, bool value) {
...
+TypeName(const T& t) {
+  return Demangle(typeid(t).name());
+}
+
+[[nodiscard]] inline std::string
+ObjectId(const std::string &classShortName, void* thisPtr) {
+  std::string id(classShortName + "[0x0000000000000000]");
+  snprintf(&id[classShortName.size() + 3], 16, "%p", thisPtr);
+  return id;
+}
```
kosak, line 276: This method has some bugs:
1. %p includes the prefix 0x on my computer, so the string will contain [0x0x...]
2. If the pointer does not have leading zeros it will be truncated because the last two digits won't fit
3. If the pointer does have 3 or more leading zeros it will be short, and a \0 character will be written into the middle of the string

I think this can be avoided with the stringstream approach. Suggest using the `SimpleOstringstream` in this same file, which has the advantage that its "str()" object can be moved away, thus saving a copy and improving Deehaven efficiency by 0.001%.  Using that, this code could be

```
std::string ObjectId(const std::string &classShortName, void *thisPtr) {
    SimpleOstringStream s;
    s << classShortName << '[' << thisPtr << ']';
    return std::move(s.str());
}
```

OR, if you don't hate the kosakian formatting routines, it can be this concise:

```
std::string ObjectId(const std::string &classShortName, void *thisPtr) {
    return stringf("%o[%o]", classShortName, thisPtr);
}
```

the latter is probably what I would do

cpp-client/deephaven/dhcore/include/public/deephaven/dhcore/utility/utility.h
```diff
@@ -238,4 +243,43 @@ inline void trueOrThrow(const DebugInfo &debugInfo, bool value) {
...
+
+[[nodiscard]] std::string Demangle(const char* name);
+
+template <class T> [[nodiscard]] std::string
+TypeName(const T& t) {
+  return Demangle(typeid(t).name());
+}
+
+[[nodiscard]] inline std::string
+ObjectId(const std::string &classShortName, void* thisPtr) {
```
kosak, line 272: here and elsewhere:
`void *thisPtr`
rather than
`void* thisPtr`

cpp-client/deephaven/dhcore/include/public/deephaven/dhcore/utility/utility.h
```diff
@@ -238,4 +243,43 @@ inline void trueOrThrow(const DebugInfo &debugInfo, bool value) {
...
+  snprintf(&id[classShortName.size() + 3], 16, "%p", thisPtr);
+  return id;
+}
+
+inline std::string
+ThreadIdToString(const std::thread::id tid) {
+  std::stringstream ss;
+  ss << tid;
+  return ss.str();
+}
```
kosak, line 283: Here I would again use `SimpleOstringstream` and `return std::move(ss.str())`

or even the kosakian

```
std::string ThreadIdToString(const std::thread::id tid) {
  return stringf("%o", tid);
}
```

cpp-client/deephaven/dhcore/src/utility/utility.cc
```diff
@@ -187,5 +196,59 @@ void dumpTillPercentOrEnd(ostream &result, const char **fmt) {
...
+  assert(resBuf[22] == '0');
+  resBuf[22] = DIGITS[singleMillisRest];
+  return result;
+}
+
+#ifdef __GNUG__
+std::string Demangle(const char* name) {
+  int status = -1;
+  std::unique_ptr<char, void(*)(void*)> res {
+    abi::__cxa_demangle(name, NULL, NULL, &status),
```
kosak, line 241: kindly use `nullptr` instead of `NULL`

cpp-client/deephaven/dhcore/src/utility/utility.cc
```diff
@@ -187,5 +196,59 @@ void dumpTillPercentOrEnd(ostream &result, const char **fmt) {
...
+    std::free
+  };
```
kosak, line 243: This is very subjective, but I think this is a little too "clever" for my tastes. I think I would rather go with a more straightforward

```
std::string Demangle(const char *name) {
  int status = -1;
  char *res = abi::__cxa_demangle(name, nullptr, nullptr, &status);
  std::string result = status == 0 ? res : name;
  free(res);
  return result;
}
```

cpp-client/deephaven/dhcore/src/utility/utility.cc
```diff
@@ -187,5 +196,59 @@ void dumpTillPercentOrEnd(ostream &result, const char **fmt) {
...
+  assert(resBuf[20] == '0');
+  resBuf[20] = DIGITS[hundredMillisRest];
+  assert(tensMillisRest < 10);
+  assert(resBuf[21] == '0');
+  resBuf[21] = DIGITS[tensMillisRest];
+  assert(singleMillisRest < 10);
+  assert(resBuf[22] == '0');
+  resBuf[22] = DIGITS[singleMillisRest];
+  return result;
+}
```
kosak, line 235: This is very subjective, but this is a little too much array surgery for my tastes. I have a method in `dhcore/src/types.cc` that does almost the same thing (but it prints nanos rather than millis, it prints in UTC rather than localtime, and it writes to a stream rather than returning a string). I think I would request that you use that approach (modifed of course for your purposes). You could use a SimpleOstringStream for example. That method looks like the below. Subjective, but I sort of prefer making little local C arrays and then stitching them together at the end, rather than surgery in-place.

```
void DateTime::streamIrisRepresentation(std::ostream &s) const {
  size_t oneBillion = 1000000000;
  time_t timeSecs = nanos_ / oneBillion;
  auto nanos = nanos_ % oneBillion;
  struct tm tm = {};
  gmtime_r(&timeSecs, &tm);
  char dateBuffer[32];  // ample
  char nanosBuffer[32];  // ample
  strftime(dateBuffer, sizeof(dateBuffer), "%FT%T", &tm);
  snprintf(nanosBuffer, sizeof(nanosBuffer), "%09zd", nanos);
  s << dateBuffer << '.' << nanosBuffer << " UTC";
}
```

cpp-client/deephaven/dhcore/src/utility/utility.cc
```diff
@@ -187,5 +196,59 @@ void dumpTillPercentOrEnd(ostream &result, const char **fmt) {
...
+  std::unique_ptr<char, void(*)(void*)> res {
+    abi::__cxa_demangle(name, NULL, NULL, &status),
+    std::free
+  };
+
+  return (status==0) ? res.get() : name;
+}
+#else
+// does nothing if not g++
+std::string Demangle(const char* name) {
```
kosak, line 249: this is going to be nice to have. For example, `verboseCast` in this same file can use your Demangle rather than the inferior `typeid(*ptr).name()`.  Feel free to change `verboseCast` if you feel like it.

We can delete my `getTypeName` and helpers altogether and we can change `verboseCast` to be more like

```
  auto message = stringf("%o: Expected type %o. Got type %o",
      debugInfo, Demangle(typeid(destType_t).name()), Demangle(typeid(*ptr).name());
```

cpp-client/deephaven/dhcore/src/utility/utility.cc
jcferretti: We discussed a bit over slack.  I am punting on doing something right now, in light of there being no correctness objection, and the amount of work associated with this one is relatively high (compared to addressing all the rest of the comments).  So my proposal is we move forward with this implementation for merge, and if we care enough we address this in a followup PR later.

kosak: I now have a correctness objection, because both epochMillisToStr() and objectId() return garbled strings. (I mentioned the problem with objectId() in a previous comment -- maybe you didn't see it?)

`objectId()` is incorrect because it makes output like
```
/home/kosak/git/deephaven-core/cpp-client/tests/utility_test.cc:29: Failure:
  CHECK(id == "hello[0xdeadbeef]")
with expansion:
  "hello[0x0xdeadbeef\x0000000]"
==
"hello[0xdeadbeef]"
```

with repeated "0x0x", an embedded \0, and other problems.

`epochMillisToStr()` is incorrect because it makes output like
```
/home/kosak/git/deephaven-core/cpp-client/tests/utility_test.cc:22: Failure:
  CHECK(d1 == "2023-07-12T00:40:24.000-0400")
with expansion:
  "2023-07-12T00:40:24.000-0400\x00###"
==
"2023-07-12T00:40:24.000-0400"
```

where there is an embedded \0 after the 0400 and trailing matter with the pound signs.

I've implemented both in the way I think makes sense and I have taken the liberty of pushing two commits plus simple unit tests to your branch, Kindly take a look. The second commit undoes a mistake I made in the first, where I accidentally trashed another unrelated unit test.

The unit test for epochMillisToStr() is not going to really work on the server because the server may have a different notion of local time. We may have to do something about that -- either not committing the unit test, forcing the server to temporarily believe in a certain local time zone, or some other idea.

kosak (approved):

PR #4164, opened 2023-07-11
Add ObjectMapper to KafkaTools.Consume.KeyOrValueSpec.Json Configuration

### LINKED ISSUE #4159: `JsonNodeUtil.objectMapper` should be public
As a user of Kafka Json processing, I would like to be able to enable disable specific jackson-core features.

One such feature:
```java
 mapper.enable(com.fasterxml.jackson.core.json.JsonReadFeature.ALLOW_NON_NUMERIC_NUMBERS.mappedFeature());
```

We should include some documentation instructing python users how to properly enable/disable jackson-core features. It is likely out-of-scope to make any subset of features "python-friendly", but instead should focus on the documentation for self-help and perhaps this use-case as a featured example.

It probably makes sense to also rename to `OBJECT_MAPPER`.

Issue is marked as a release blocker because 1) effort is low, and 2) was brought to our attention via a community user.

nbauernfeind:
Fixes #4159.

We'll still need some documentation with a REPL example. Note that our default factory has some configuration, too:
```java
    private static final ObjectMapper DEFAULT_OBJECT_MAPPER = new ObjectMapper()
            .setNodeFactory(JsonNodeFactory.withExactBigDecimals(true))
            .configure(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS, true);
```

### DIFF
--- a/server/src/main/java/io/deephaven/server/runner/DeephavenApiServer.java
+++ b/server/src/main/java/io/deephaven/server/runner/DeephavenApiServer.java
@@ -3,6 +3,7 @@
  */
 package io.deephaven.server.runner;
 
+import io.deephaven.UncheckedDeephavenException;
 import io.deephaven.auth.AuthenticationRequestHandler;
 import io.deephaven.configuration.Configuration;
 import io.deephaven.engine.context.ExecutionContext;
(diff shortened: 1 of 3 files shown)


server/src/main/java/io/deephaven/server/runner/DeephavenApiServer.java
```diff
@@ -3,6 +3,7 @@
  */
 package io.deephaven.server.runner;
 
+import io.deephaven.UncheckedDeephavenException;
```
rcaudy, line 6: Stray change

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2896
Reference: https://github.com/deephaven/deephaven.io/issues/2897

PR #4170, opened 2023-07-11
Add `Multijoin` feature to community

lbooker42:
Port the enterprise `multijoin` feature to allow more efficient computation of multiple table joins.

The multiJoin operation collects the set of distinct keys from the input tables, and then joins one row from each of the input tables onto the result. Input tables need not have a match row for each key, but they may not have multiple matching rows for a given key. The operation can be thought of as a merge of the key columns, followed by a selectDistinct and then a series of iterative naturalJoin operations as follows:

```
private Table doIterativeMultiJoin(String [] keyColumns, List<? extends Table> inputTables) {
    final List<Table> keyTables = inputTables.stream().map(t -> t.view(keyColumns)).collect(Collectors.toList());
    final Table base = TableTools.merge(keyTables).selectDistinct(keyColumns);

    Table result = base;
    for (int ii = 0; ii < inputTables.size(); ++ii) {
        result = result.naturalJoin(inputTables.get(ii), Arrays.asList(keyColumns));
    }

    return result;
}
```


engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTable.java
```diff
@@ -0,0 +1,778 @@
...
+ * </p>
+ *
+ * <pre>{@code
+ *     private Table doIterativeMultiJoin(String [] keyColumns, List<? extends Table> inputTables) {
+ *         final List<Table> keyTables = inputTables.stream().map(t -> t.view(keyColumns)).collect(Collectors.toList());
+ *         final Table base = TableTools.merge(keyTables).selectDistinct(keyColumns);
+ *
+ *         Table result = base;
+ *         for (int ii = 0; ii < inputTables.size(); ++ii) {
+ *             result = result.naturalJoin(inputTables.get(ii), Arrays.asList(keyColumns));
```
rcaudy, line 58: This probably needs a bit more fleshing out to document join additions being different by input, as well as join matches only needing a common LHS.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTable.java
```diff
@@ -0,0 +1,778 @@
...
+
+    /**
+     * Join tables that have common key column names.
+     * <p>
+     *
+     * @param keys the key column pairs in the format "Result=Source" or "ColumnInBoth"
+     * @param inputTables the tables to join together
+     * @return a MultiJoinTable with one row for each key and the corresponding row in each input table
+     */
+    public static MultiJoinTable of(@NotNull final String[] keys, @NotNull final Table... inputTables) {
```
rcaudy, line 168: I would move `of` to a static `MultiJoinFactory` class for now.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,643 @@
...
+    private final Table table;
+
+    @TestUseOnly
+    static MultiJoinTableImpl of(@NotNull final JoinControl joinControl,
+            @NotNull final MultiJoinInput... multiJoinInputs) {
+        return QueryPerformanceRecorder.withNugget("multiJoin",
+                () -> new MultiJoinTableImpl(joinControl, multiJoinInputs));
+    }
+
+    static MultiJoinTableImpl of(@NotNull final MultiJoinInput... joinDescriptors) {
```
rcaudy, line 48: Rename `joinDescriptors` to `multiJoinInputs` for consistency

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,643 @@
...
+     *
+     * @return The output {@link Table table}
+     */
+    public Table table() {
+        return table;
+    }
+
+    @Override
+    public Collection<String> keyColumns() {
+        return null;
```
rcaudy, line 64: Needs implementation.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,643 @@
...
+        table = multiJoin(joinControl, multiJoinInputs);
+    }
+
+    private Table multiJoin(@NotNull final JoinControl joinControl,
+            @NotNull final MultiJoinInput... multiJoinInputs) {
+        final TObjectIntHashMap<String> usedColumns =
+                new TObjectIntHashMap<>(multiJoinInputs[0].columnsToAdd().length, 0.5f, -1);
+        final List<String> expectedLeftMatches =
+                ColumnName.names(JoinMatch.lefts(Arrays.asList(multiJoinInputs[0].columnsToMatch())));
+        if (expectedLeftMatches.size() == 0) {
```
rcaudy, line 78: Move this up and base it on `multiJoinInputs[0].columnsToMatch().size()`

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,643 @@
...
+            return doMultiJoinZeroKey(multiJoinInputs);
+        }
+
+        final Set<String> keyColumnNames = new LinkedHashSet<>(expectedLeftMatches);
+        for (final String keyColumn : expectedLeftMatches) {
+            usedColumns.put(keyColumn, KEY_COLUMN_SENTINEL);
+        }
+        final ChunkType[] expectedChunkTypes = Arrays.stream(getKeySources(multiJoinInputs[0]))
```
rcaudy, line 86: You might want to do:
1. A single pass over the inputs to get *all* your match columns.
2. A single pass over the inputs to get *all* your addition columns.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,643 @@
...
+        for (int jj = 0; jj < multiJoinInputs.length; ++jj) {
+            MultiJoinInput joinDescriptor = useDescriptors[jj];
+
+            final List<String> currentLeftMatches =
+                    ColumnName.names(JoinMatch.lefts(Arrays.asList(joinDescriptor.columnsToMatch())));
+            if (!currentLeftMatches.equals(expectedLeftMatches)) {
+                throw new IllegalArgumentException("Key column mismatch for table " + jj + ", first key columns="
+                        + expectedLeftMatches + " table has " + currentLeftMatches);
+            }
+            final ChunkType[] currentChunkType = Arrays.stream(getKeySources(joinDescriptor))
```
rcaudy, line 100: We should check for matching data type and component type, not chunk type.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,643 @@
...
+
+        final MultiJoinInput[] useDescriptors = Arrays.copyOf(multiJoinInputs, multiJoinInputs.length);
+
+        for (int jj = 0; jj < multiJoinInputs.length; ++jj) {
+            MultiJoinInput joinDescriptor = useDescriptors[jj];
+
+            final List<String> currentLeftMatches =
+                    ColumnName.names(JoinMatch.lefts(Arrays.asList(joinDescriptor.columnsToMatch())));
+            if (!currentLeftMatches.equals(expectedLeftMatches)) {
+                throw new IllegalArgumentException("Key column mismatch for table " + jj + ", first key columns="
```
rcaudy, line 97: Prefer `String.format`, if you don't hate it.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,643 @@
...
+    private static MultiJoinInput maybePopulateColumnsToAdd(
+            @NotNull final Set<String> keyColumnNames,
+            @NotNull final MultiJoinInput[] useDescriptors, final int tableNumber) {
+        MultiJoinInput joinDescriptor = useDescriptors[tableNumber];
+        JoinAddition[] columnsToAdd = joinDescriptor.columnsToAdd();
+        if (columnsToAdd.length == 0) {
+            // create them on the fly from the table
+            final JoinAddition[] newColumnsToAdd = joinDescriptor.inputTable().getDefinition().getColumnNames().stream()
+                    .filter(cn -> !keyColumnNames.contains(cn)).map(MatchPairFactory::getExpression)
+                    .toArray(MatchPair[]::new);
```
rcaudy, line 237:
```suggestion
                    .filter(cn -> !keyColumnNames.contains(cn)).map(JoinAddition::parse)
                    .toArray(JoinAddition[]::new);
```

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,643 @@
...
+    }
+
+    @Override
+    public Collection<String> keyColumns() {
+        return null;
+    }
+
+    private MultiJoinTableImpl(@NotNull final JoinControl joinControl,
+            @NotNull final MultiJoinInput... multiJoinInputs) {
+        table = multiJoin(joinControl, multiJoinInputs);
```
rcaudy, line 69:
```suggestion
        final Table[] tables = Arrays.stream(multiJoinInputs).map(MultiJoinInput::inputTable).toArray(Table[]::new);
        final UpdateGraph updateGraph = tables[0].getUpdateGraph(tables);
        if (updateGraph != null) {
            updateGraph.checkInitiateSerialTableOperation();
        }
        try (final SafeCloseable ignored = ExecutionContext.getContext().withUpdateGraph(updateGraph).open()) {
            table = multiJoin(joinControl, multiJoinInputs);
        }
```

engine/api/src/main/java/io/deephaven/engine/table/MultiJoinInput.java
```diff
@@ -91,6 +93,19 @@ public static MultiJoinInput of(@NotNull final Table inputTable, String columnsT
                         : JoinAddition.from(columnsToAdd));
     }
 
+    /**
+     * Create an array of multiJoin table descriptors with common keys; includes all non-key columns as output columns.
```
rcaudy, line 97:
```suggestion
     * Create an array of MultiJoinInputs with common keys; includes all non-key columns as output columns.
```

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinStateManager.java
```diff
@@ -5,21 +5,21 @@
...
      * @param table the table to add
      * @param sources the column sources that contain the keys
-     * @param tableNumber the table number that we are adding rows for
+     * @param tableNumber the table number for which we are adding rows
      */
     void build(final Table table, ColumnSource<?>[] sources, int tableNumber);
 
     /**
-     * How many rows are in our result table?
+     * Get the number of rows in the result table
```
rcaudy, line 22: Is this overall or hash table size?

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -35,19 +35,28 @@
 
 public class MultiJoinTableImpl implements MultiJoinTable {
 
-    static final int KEY_COLUMN_SENTINEL = -2;
+    private static final int KEY_COLUMN_SENTINEL = -2;
     private final Table table;
 
+    private final Collection<String> keyColumns;
+
     @TestUseOnly
```
rcaudy, line 43:
```suggestion
    @VisibleForUnitTests
```

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -61,70 +70,90 @@ public Table table() {
...
+                .map(ColumnSource::getType).toArray(Class<?>[]::new);
+        final Class<?>[] expectedComponentTypes = Arrays.stream(firstKeySources)
+                .map(ColumnSource::getComponentType).toArray(Class<?>[]::new);
 
+        for (int ii = 1; ii < multiJoinInputs.length; ++ii) {
+            ColumnSource[] currentKeySources = getKeySources(multiJoinInputs[ii]);
+            // Verify matching column names.
             final List<String> currentLeftMatches =
-                    ColumnName.names(JoinMatch.lefts(Arrays.asList(joinDescriptor.columnsToMatch())));
+                    ColumnName.names(JoinMatch.lefts(Arrays.asList(multiJoinInputs[ii].columnsToMatch())));
```
rcaudy, line 111: Can we allow differing orders? Since we have a LHS name, we *could* allow for out-of-order matches.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -61,70 +70,90 @@ public Table table() {
...
             if (!currentLeftMatches.equals(expectedLeftMatches)) {
-                throw new IllegalArgumentException("Key column mismatch for table " + jj + ", first key columns="
-                        + expectedLeftMatches + " table has " + currentLeftMatches);
+                throw new IllegalArgumentException(
+                        String.format("Key column mismatch for table %d, first key columns=%s table has %s", ii,
```
rcaudy, line 114:
```suggestion
                        String.format("Key column mismatch for table %d, first table has key columns=%s, this table has %s", ii,
```

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -61,70 +70,90 @@ public Table table() {
...
+                                expectedLeftMatches, currentLeftMatches));
+            }
+            // Verify matching column and component types
+            final Class<?>[] currentColumnTypes = Arrays.stream(currentKeySources)
+                    .map(ColumnSource::getType).toArray(Class<?>[]::new);
+            final Class<?>[] currentComponentTypes = Arrays.stream(currentKeySources)
+                    .map(ColumnSource::getComponentType).toArray(Class<?>[]::new);
+            if (!Arrays.equals(currentColumnTypes, expectedColumnTypes)) {
+                throw new IllegalArgumentException(
+                        String.format("Key column type mismatch for table %d, first key columns types=%s table has %s",
```
rcaudy, line 124:
```suggestion
                        String.format("Key column type mismatch for table %d, first table has key column types=%s, this table has %s",
```

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -61,70 +70,90 @@ public Table table() {
...
             }
-            final ChunkType[] currentChunkType = Arrays.stream(getKeySources(joinDescriptor))
-                    .map(ColumnSource::getChunkType).toArray(ChunkType[]::new);
-            if (!Arrays.equals(currentChunkType, expectedChunkTypes)) {
-                throw new IllegalArgumentException("Key column type mismatch for table " + jj
-                        + ", first key columns types=" + Arrays.toString(expectedChunkTypes) + " table has "
-                        + Arrays.toString(currentChunkType));
+            if (!Arrays.equals(currentComponentTypes, expectedComponentTypes)) {
+                throw new IllegalArgumentException(String.format(
+                        "Key column component type mismatch for table %d, first key columns component types=%s table has %s",
```
rcaudy, line 129:
```suggestion
                        "Key column component type mismatch for table %d, first table has key column component types=%s, this table has %s",
```

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -61,70 +70,90 @@ public Table table() {
...
+                            columnName,
+                            previouslyUsed == KEY_COLUMN_SENTINEL ? "key columns" : Integer.toString(previouslyUsed),
+                            ii));
                 }
             }
         }
 
-        final ColumnSource<?>[] firstKeySources = getKeySources(useDescriptors[0]);
-        final ColumnSource<?>[] originalColumns = getOriginalKeyColumns(useDescriptors[0]);
+
```
rcaudy, line 151:
```suggestion
```

engine/table/src/main/java/io/deephaven/engine/table/impl/by/typed/HasherConfig.java
```diff
@@ -103,17 +101,30 @@ public ProbeSpec(String name, String stateValueName, boolean requiresRowKeyChunk
     }
 
     static class BuildSpec {
+        @FunctionalInterface
+        public interface TriConsumer<T, U, V> {
+            void accept(T t, U u, V v);
+        }
```
rcaudy, line 107: At this point, I would make specific interfaces, in lieu of BiConsumer and TriConsumer.

rcaudy (commented):
Review of just the changes since Tuesday.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinStateManager.java
lbooker42: This returns used hash table entry count (correct for static but incorrect for incremental).

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+                            result.newModifiedColumnSet(addNames);
+                    final ModifiedColumnSet.Transformer transformer =
+                            input.newModifiedColumnSetTransformer(result, addNames);
+
+                    final MultiJoinListenerRecorder listenerRecorder =
+                            new MultiJoinListenerRecorder("multiJoin(" + ii + ")", input, result, keySources,
+                                    sourceKeyModifiedColumnSet, transformer, ii);
+                    listenerRecorder.setMergedListener(mergedListener);
+                    input.addUpdateListener(listenerRecorder);
+                    listenerRecorders.add(listenerRecorder);
```
rcaudy, line 245: 1. The listener recorder does not need to be added as a parent reference for dependent. That relationship is typically enforced via the merged listener.
2. The merged listener should manage the listener recorders. Since it doesn't see them at construction time, you need to make that explicit here.

The best way to accomplish both is to pass the merged listener as dependent to listener recorder construction, instead of passing the result.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+    @NotNull
+    private static MultiJoinInput maybePopulateColumnsToAdd(
+            @NotNull final Collection<String> keyColumnNames,
+            @NotNull final MultiJoinInput[] multiJoinInputs, final int tableNumber) {
+        MultiJoinInput joinInput = multiJoinInputs[tableNumber];
+        JoinAddition[] columnsToAdd = joinInput.columnsToAdd();
+        if (columnsToAdd.length == 0) {
+            // create them on the fly from the table
+            final JoinAddition[] newColumnsToAdd = joinInput.inputTable().getDefinition().getColumnNames().stream()
+                    .filter(cn -> !keyColumnNames.contains(cn)).map(JoinAddition::parse)
```
rcaudy, line 264:
```suggestion
                    .filter(cn -> !keyColumnNames.contains(cn)).map(ColumnName::of)
```

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+                throw new IllegalArgumentException(String.format(
+                        "Key column component type mismatch for table %d, first key columns component types=%s table has %s",
+                        ii, Arrays.toString(expectedComponentTypes), Arrays.toString(currentComponentTypes)));
+            }
+        }
+
+        // Gather and verify the non-key output columns
+        final MultiJoinInput[] useJoinInputs = Arrays.copyOf(multiJoinInputs, multiJoinInputs.length);
+        for (int ii = 0; ii < multiJoinInputs.length; ++ii) {
+            final MultiJoinInput joinInput = maybePopulateColumnsToAdd(keyColumns, useJoinInputs, ii);
```
rcaudy, line 137: Wrong key columns. Needs to be in the namespace of the input table.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+        final MultiJoinInput[] useJoinInputs = Arrays.copyOf(multiJoinInputs, multiJoinInputs.length);
+        final TObjectIntHashMap<String> usedColumns =
+                new TObjectIntHashMap<>(multiJoinInputs[0].columnsToAdd().length, 0.5f, -1);
+
+        for (int jj = 0; jj < multiJoinInputs.length; ++jj) {
+            MultiJoinInput joinInput = useJoinInputs[jj];
+
+            if (joinInput.columnsToMatch().length != 0) {
+                Collection<String> matches = Arrays.stream(joinInput.columnsToMatch())
+                        .map(v -> v.left().name() + "=" + v.right().name()).collect(Collectors.toList());
```
rcaudy, line 284: Use `io.deephaven.api.Strings#ofJoinMatches`

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+                throw new IllegalArgumentException(
+                        "Key column mismatch for table " + jj + ", first table had no key columns this table has "
+                                + matches);
+            }
+
+            joinInput = maybePopulateColumnsToAdd(Collections.emptySet(), useJoinInputs, jj);
+            for (int cc = 0; cc < joinInput.columnsToAdd().length; ++cc) {
+                final String columnName = joinInput.columnsToAdd()[cc].newColumn().name();
```
rcaudy, line 292: This looks like a place to do foreach, rather than indexed iteration.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+        }
+        return joinInput;
+    }
+
+    private static Table doMultiJoinZeroKey(@NotNull final MultiJoinInput... multiJoinInputs) {
+        final MultiJoinInput[] useJoinInputs = Arrays.copyOf(multiJoinInputs, multiJoinInputs.length);
+        final TObjectIntHashMap<String> usedColumns =
+                new TObjectIntHashMap<>(multiJoinInputs[0].columnsToAdd().length, 0.5f, -1);
+
+        for (int jj = 0; jj < multiJoinInputs.length; ++jj) {
```
rcaudy, line 279: I wonder if we could implement this loop only once, shared between multi-key and no-key.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+
+            if (joinInput.columnsToMatch().length != 0) {
+                Collection<String> matches = Arrays.stream(joinInput.columnsToMatch())
+                        .map(v -> v.left().name() + "=" + v.right().name()).collect(Collectors.toList());
+                throw new IllegalArgumentException(
+                        "Key column mismatch for table " + jj + ", first table had no key columns this table has "
+                                + matches);
+            }
+
+            joinInput = maybePopulateColumnsToAdd(Collections.emptySet(), useJoinInputs, jj);
```
rcaudy, line 290: Maybe better to just mutate the array ourselves, rather than pass the array and an index down?

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+                hasResults = true;
+            } else {
+                throw new IllegalStateException("Duplicate rows for table " + tableNumber + " on zero-key multiJoin.");
+            }
+
+            final SingleValueRowRedirection rowRedirection =
+                    refreshing ? new WritableSingleValueRowRedirection(key) : new SingleValueRowRedirection(key);
+            if (refreshing) {
+                rowRedirection.writableSingleValueCast().startTrackingPrevValues();
+            }
```
rcaudy, line 324:
```suggestion
            final SingleValueRowRedirection rowRedirection;
            if (refreshing) {
                rowRedirection = new WritableSingleValueRowRedirection(key);
                rowRedirection.writableSingleValueCast().startTrackingPrevValues();
            } else {
                rowRedirection = new SingleValueRowRedirection(key);
            }
```

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+        }
+
+        final QueryTable result = new QueryTable(RowSetFactory.flat(hasResults ? 1 : 0).toTracking(), resultSources);
+
+        if (refreshing) {
+            final ModifiedColumnSet[] resultModifiedColumnSet = new ModifiedColumnSet[useJoinInputs.length];
+            final List<MultiJoinListenerRecorder> listenerRecorders = Collections.synchronizedList(new ArrayList<>());
+
+            final Logger log = LoggerFactory.getLogger(MultiJoinTableImpl.class);
```
rcaudy, line 339: Same comments as keyed.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+                    resultModifiedColumnSet[ii] =
+                            result.newModifiedColumnSet(addNames);
+                    final ModifiedColumnSet.Transformer transformer =
+                            input.newModifiedColumnSetTransformer(result, addNames);
+
+                    final MultiJoinListenerRecorder listenerRecorder = new MultiJoinListenerRecorder(
+                            "multiJoin(" + ii + ")", input, result, null, null, transformer, ii);
+                    listenerRecorder.setMergedListener(mergedListener);
+                    input.addUpdateListener(listenerRecorder);
+                    listenerRecorders.add(listenerRecorder);
```
rcaudy, line 365: Same comments as keyed.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+                        if (recorder.getParent().size() > 1) {
+                            throw new IllegalStateException(
+                                    "Multiple rows in " + recorder.tableNumber + " for zero-key multiJoin.");
+                        }
+                        resultModified = true;
+                        redirections[recorder.tableNumber].writableSingleValueCast()
+                                .setValue(recorder.getParent().getRowSet().firstRowKey());
+                        downstream.modifiedColumnSet.setAll(modifiedColumnSets[recorder.tableNumber]);
+                    } else if (recorder.getModified().isNonempty()) {
+                        resultModified = true;
```
rcaudy, line 634: I think this is only so if we actually modified any of the join addition columns.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+                        redirections[recorder.tableNumber].writableSingleValueCast()
+                                .setValue(recorder.getParent().getRowSet().firstRowKey());
+                    }
+                }
+            }
+
+            final boolean hasResults = Arrays.stream(redirections).anyMatch(rd -> rd.getValue() != NULL_ROW_KEY);
+            if (hasResults && result.size() == 0) {
+                result.getRowSet().writableCast().insert(0);
+                downstream.added = RowSetFactory.fromKeys(0);
```
rcaudy, line 648: `flat(1)` better, IMO.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+                downstream.modified = RowSetFactory.empty();
+                downstream.modifiedColumnSet = ModifiedColumnSet.EMPTY;
+            } else {
+                downstream.added = RowSetFactory.empty();
+                downstream.removed = RowSetFactory.empty();
+                if (resultModified) {
+                    downstream.modified = RowSetFactory.fromKeys(0);
+                } else {
+                    // this would be a useless update
+                    return;
```
rcaudy, line 665: I think we don't need to close the update, since we never set any fields on it, right?

rcaudy (commented):
WIP on review

rcaudy:
Review should continue with `MultiJoinMergedListener` and the actual state managers.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+                    }
+                }
+            }
+            downstream.removed = removedBuilder.build();
+            downstream.added.writableCast().insert(addedBuilder.build());
+
+            downstream.shifted = RowSetShiftData.EMPTY;
+
+            result.getRowSet().writableCast().remove(downstream.removed);
+            result.getRowSet().writableCast().insert(downstream.added);
```
rcaudy, line 591:
```suggestion
            result.getRowSet().writableCast().update(downstream.added, downstream.removed);
```

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+
+        @Override
+        protected void process() {
+            slotTracker.clear();
+            slotTracker.ensureTableCapacity(stateManager.getTableCount());
+
+            final long originalSize = stateManager.getResultSize();
+
+            for (MultiJoinListenerRecorder recorder : recorders) {
+                if (recorder.recordedVariablesAreValid()) {
```
rcaudy, line 445: I think we can ignore recorded updates if there are only modifies to columns that are not key columns and not join additions. That might be worthwhile. That does, however, create a possibility that we might have to *not* deliver a downstream notification in some cases, because it will be empty.

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/StaticMultiJoinStateManagerTypedBase.java
```diff
@@ -0,0 +1,218 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import static io.deephaven.engine.table.impl.JoinControl.CHUNK_SIZE;
+import static io.deephaven.engine.table.impl.JoinControl.MAX_TABLE_SIZE;
+import static io.deephaven.engine.table.impl.util.TypedHasherUtil.getKeyChunks;
+
+public abstract class StaticMultiJoinStateManagerTypedBase implements MultiJoinStateManager {
+    protected final ColumnSource<?>[] keySourcesForErrorMessages;
+    private final List<LongArraySource> indexSources = new ArrayList<>();
```
rcaudy, line 36: `rowKeySources`? `redirections`?

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/StaticMultiJoinStateManagerTypedBase.java
```diff
@@ -0,0 +1,218 @@
...
+
+    public static final long NO_RIGHT_STATE_VALUE = RowSet.NULL_ROW_KEY;
+    public static final long EMPTY_RIGHT_STATE = QueryConstants.NULL_LONG;
+    public static final long DUPLICATE_RIGHT_STATE = -2;
+
+    // The number of slots in our hash table.
```
rcaudy, line 42: Use JavaDoc format.

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/StaticMultiJoinStateManagerTypedBase.java
```diff
@@ -0,0 +1,218 @@
...
+
+    // The keys for our hash entries.
+    protected final ChunkType[] chunkTypes;
+    protected final WritableColumnSource[] mainKeySources;
+
+    // The output sources representing the keys of our joined table.
+    protected final WritableColumnSource[] outputKeySources;
+
+    // Store sentinel information and maps hash slots to output row keys.
+    protected ImmutableLongArraySource slotToOutputRow = new ImmutableLongArraySource();
```
rcaudy, line 60: We seem to have a conflict here. This is an immutable, so we cannot `ensureCapacity` on it more than once, but we're doing so in the constructor based on what the `JoinControl` says, rather than doing it at the end when building our output table.

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/StaticMultiJoinStateManagerTypedBase.java
```diff
@@ -0,0 +1,218 @@
...
+        Require.inRange(maximumLoadFactor, 0.0, 0.95, "maximumLoadFactor");
+
+        mainKeySources = new WritableColumnSource[tableKeySources.length];
+        chunkTypes = new ChunkType[tableKeySources.length];
+
+        outputKeySources = new WritableColumnSource[tableKeySources.length];
+
+        for (int ii = 0; ii < tableKeySources.length; ++ii) {
+            chunkTypes[ii] = tableKeySources[ii].getChunkType();
+            mainKeySources[ii] = InMemoryColumnSource.getImmutableMemoryColumnSource(tableSize,
```
rcaudy, line 81: These aren't immutable.

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/StaticMultiJoinStateManagerTypedBase.java
```diff
@@ -0,0 +1,218 @@
...
+                    tableKeySources[ii].getType(), tableKeySources[ii].getComponentType());
+            outputKeySources[ii] = ArrayBackedColumnSource.getMemoryColumnSource(tableSize,
```
rcaudy, line 83: These actually are immutable. But maybe capacity isn't being ensured?

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/StaticMultiJoinStateManagerTypedBase.java
```diff
@@ -0,0 +1,218 @@
...
+
+        private LeftBuildHandler(LongArraySource tableRedirSource, long tableNumber) {
+            this.tableRedirSource = tableRedirSource;
+            this.tableNumber = tableNumber;
+        }
+
+        @Override
+        public void doBuild(RowSequence chunkOk, Chunk<Values>[] sourceKeyChunks) {
+            final long maxSize = numEntries + chunkOk.intSize();
+            slotToOutputRow.ensureCapacity(maxSize);
```
rcaudy, line 126: I find it confusing that this is happening here, and in `ensureCapacity`. I think it's either like `outputKeySources` *or* like `mainKeySources`. Which is correct?

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/StaticMultiJoinStateManagerTypedBase.java
```diff
@@ -0,0 +1,218 @@
...
+        if (table.isEmpty()) {
+            return;
+        }
+        final LongArraySource tableRedirSource = indexSources.get(tableNumber);
+        try (final BuildContext bc = makeBuildContext(keySources, table.size())) {
+            buildTable(bc, table.getRowSet(), keySources, new LeftBuildHandler(tableRedirSource, tableNumber));
+        }
+    }
+
+    private class LeftBuildHandler implements TypedHasherUtil.BuildHandler {
```
rcaudy, line 114: Do we need a `Left`? Or ` buildFromLeftSide`?

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/StaticMultiJoinStateManagerTypedBase.java
```diff
@@ -0,0 +1,218 @@
...
+            final BuildContext bc,
+            final RowSequence buildRows,
+            final ColumnSource<?>[] buildSources,
+            final TypedHasherUtil.BuildHandler buildHandler) {
+        try (final RowSequence.Iterator rsIt = buildRows.getRowSequenceIterator()) {
+            // noinspection unchecked
+            final Chunk<Values>[] sourceKeyChunks = new Chunk[buildSources.length];
+
+            while (rsIt.hasMore()) {
+                final RowSequence chunkOk = rsIt.getNextRowSequenceWithLength(bc.chunkSize);
```
rcaudy, line 148: `rows`?

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/StaticMultiJoinStateManagerTypedBase.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    @Override
+    public ColumnSource<?>[] getKeyHashTableSources() {
+        return outputKeySources;
+    }
+
+    @Override
+    public WritableRowRedirection getRowRedirectionForTable(int tableNumber) {
+        return new LongColumnSourceWritableRowRedirection(indexSources.get(tableNumber));
```
rcaudy, line 201: I think this need not be writable. You just need to mark the redirections in the incremental case to start tracking previous values after the initial build.

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/StaticMultiJoinStateManagerTypedBase.java
lbooker42: This has to be done initially to set the size of the immutable columns.  After this, the columns are resized during the rehash.

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/StaticMultiJoinStateManagerTypedBase.java
lbooker42: The pattern for OA is to have these Immutable, then adjust size and contents during rehash.

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/StaticMultiJoinStateManagerTypedBase.java
lbooker42: Constructor sets initial size, rehash will size as needed.

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/StaticMultiJoinStateManagerTypedBase.java
lbooker42: This is incorrect here, has been removed.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinModifiedSlotTracker.java
```diff
@@ -0,0 +1,264 @@
...
+    private long doModify(final long slot, final int tableNumber, byte flags) {
+        maybeAllocateChunk();
+        initializeNextTrackerSlot(slot, tableNumber, flags);
+        for (final LongArraySource originalRedirectionForTable : this.originalRedirection) {
+            originalRedirectionForTable.set(pointer, SENTINEL_UNINITIALIZED_KEY);
+        }
+        return getCookieFromPointer(pointer++);
+    }
+
+    private long updateFlags(final long cookie, final int tableNumber, final long originalRedirection, byte flags) {
```
rcaudy, line 211: This name is misleading. Doesn't cover full usage.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinModifiedSlotTracker.java
```diff
@@ -0,0 +1,264 @@
...
+            final long flagLocation = flagLocationForSlotAndTable(pointer, 0);
+            final long flagDestination = flagLocationForSlotAndTable(newTableLocation, 0);
+
+            // Move all the flags for the tables to the new location.
+            for (int ii = 0; ii < flagLocationsPerSlot; ++ii) {
+                flagSource.set(flagLocation + ii, flagDestination + ii);
+            }
+            modifiedSlots.set(pointer, modifiedSlots.getLong(oldTableLocation));
+        }
+    }
```
rcaudy, line 151:
```suggestion
    /**
     * Move a main table location.
     *
     * @param cookie the cookie for the hash slot
     * @param oldTableLocation the old hash slot
     * @param newTableLocation the new hash slot
     */
    public void moveTableLocation(
            long cookie,
            @SuppressWarnings("unused") int oldTableLocation,
            int newTableLocation) {
        if (isValidCookie(cookie)) {
            final long pointer = getPointerFromCookie(cookie);
            modifiedSlots.set(pointer, newTableLocation);
        }
    }
```

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinModifiedSlotTracker.java
```diff
@@ -0,0 +1,264 @@
...
+    private int flagShiftForTable(int tableNumber) {
+        return (tableNumber % FLAGS_PER_LOCATION) * FLAG_BITS;
+    }
+
+    private long setFlagInLong(int tableNumber, byte flag) {
+        return ((long) flag) << flagShiftForTable(tableNumber);
+    }
+
+    private long setFlagInLong(long existingLong, int tableNumber, byte flag) {
+        return existingLong | ((long) flag) << flagShiftForTable(tableNumber);
```
rcaudy, line 166:
```suggestion
        return existingLong | setFlagInLong(tableNumber, flag);
```

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/StaticMultiJoinStateManagerTypedBase.java
rcaudy: I think I understand. We should just be sure to avoid re-ensuring capacity.

rcaudy (commented):
WIP. Only incremental state managers left.

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/typed/incopen/gen/IncrementalMultiJoinHasherChar.java
```diff
@@ -0,0 +1,345 @@
...
+                            Assert.neq(alternateTableLocation, "alternateTableLocation", firstAlternateTableLocation, "firstAlternateTableLocation");
+                        }
+                    }
+                    numEntries++;
+                    mainKeySource0.set(tableLocation, k0);
+                    final int outputKey = numEntries - 1;
+                    slotToOutputRow.set(tableLocation, outputKey);
+                    tableRedirSource.set(outputKey, rowKeyChunk.get(chunkPosition));
+                    outputKeySources[0].set((long)outputKey, k0);
+                    mainModifiedTrackerCookieSource.set(tableLocation, EMPTY_COOKIE_SLOT);
```
rcaudy, line 86: We should note that even if this output row has subsequent adds, we will ignore its modification state on its creation step.

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/typed/incopen/gen/IncrementalMultiJoinHasherChar.java
```diff
@@ -0,0 +1,345 @@
...
+                final int firstAlternateTableLocation = hashToTableLocationAlternate(hash);
+                boolean alternateFound = false;
+                if (firstAlternateTableLocation < rehashPointer) {
+                    int alternateTableLocation = firstAlternateTableLocation;
+                    while ((slotValue = alternateSlotToOutputRow.getUnsafe(alternateTableLocation)) != EMPTY_OUTPUT_ROW) {
+                        if (eq(alternateKeySource0.getUnsafe(alternateTableLocation), k0)) {
+                            final long mappedRowKey = tableRedirSource.getUnsafe(slotValue);
+                            Assert.eq(rowKeyChunk.get(chunkPosition), "rowKey", mappedRowKey, "mappedRowKey");
+                            tableRedirSource.set(slotValue, mappedRowKey + shiftDelta);
+                            final int cookie = mainModifiedTrackerCookieSource.getUnsafe(tableLocation);
```
rcaudy, line 194: alternate

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/typed/incopen/gen/IncrementalMultiJoinHasherChar.java
```diff
@@ -0,0 +1,345 @@
...
+        }
+    }
+
+    private static int hash(char k0) {
+        int hash = CharChunkHasher.hashInitialSingle(k0);
+        return hash;
+    }
+
+    private boolean migrateOneLocation(int locationToMigrate,
+            MultiJoinModifiedSlotTracker modifiedSlotTracker) {
```
rcaudy, line 261: Can drop the tracker.

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/typed/incopen/gen/IncrementalMultiJoinHasherChar.java
```diff
@@ -0,0 +1,345 @@
...
+    protected void rehashInternalFull(final int oldSize) {
+        final char[] destKeyArray0 = new char[tableSize];
+        final int[] destState = new int[tableSize];
+        Arrays.fill(destState, EMPTY_OUTPUT_ROW);
+        final char [] originalKeyArray0 = mainKeySource0.getArray();
+        mainKeySource0.setArray(destKeyArray0);
+        final int [] originalStateArray = slotToOutputRow.getArray();
+        slotToOutputRow.setArray(destState);
+        final int [] oldModifiedCookie = mainModifiedTrackerCookieSource.getArray();
+        final int [] destModifiedCookie = new int[tableSize];
```
rcaudy, line 322: I think this suggests cookie generation should be initialized to 1, at construction or at wraparound.

rcaudy (commented):
Reviewed current state of incremental state managers.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -35,19 +34,139 @@
...
+        }
+    }
+
+    @VisibleForTesting
     static MultiJoinTableImpl of(@NotNull final JoinControl joinControl,
             @NotNull final MultiJoinInput... multiJoinInputs) {
-        return QueryPerformanceRecorder.withNugget("multiJoin",
-                () -> new MultiJoinTableImpl(joinControl, multiJoinInputs));
+
+        final Table[] tables = Arrays.stream(multiJoinInputs).map(MultiJoinInput::inputTable).toArray(Table[]::new);
```
rcaudy, line 157: We could probably memoize this, using the memoization cache on the first table.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -35,19 +34,139 @@
...
+        /** The input non-key columns in in order provided by the MutiJoinInput */
+        final String[] originalAddColumnNames;
+
+        final Map<String, ColumnSource<?>> keySourceMap;
+        final Map<String, ColumnSource<?>> originalKeySourceMap;
+
+        final JoinAddition[] columnsToAdd;
+
+        MultiJoinInputHelper(@NotNull MultiJoinInput input) {
+            table = input.inputTable();
```
rcaudy, line 60: `.coalesce()`

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -61,223 +180,166 @@ public Table table() {
...
                     final ModifiedColumnSet.Transformer transformer =
-                            input.newModifiedColumnSetTransformer(result, addNames);
+                            input.newModifiedColumnSetTransformer(result, pairs);
 
                     final MultiJoinListenerRecorder listenerRecorder =
-                            new MultiJoinListenerRecorder("multiJoin(" + ii + ")", input, result, keySources,
-                                    sourceKeyModifiedColumnSet, transformer, ii);
+                            new MultiJoinListenerRecorder("multiJoin(" + ii + ")", input, mergedListener, keySources,
+                                    sourceKeyModifiedColumnSet, sourceAdditionModifiedColumnSet, transformer, ii);
                     listenerRecorder.setMergedListener(mergedListener);
```
rcaudy, line 323: Could do this in the constructor now.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -318,66 +380,53 @@ private static Table doMultiJoinZeroKey(@NotNull final MultiJoinInput... joinDes
                     resultModifiedColumnSet,
                     redirections);
 
-            for (int ii = 0; ii < useDescriptors.length; ++ii) {
-                final MultiJoinInput jd = useDescriptors[ii];
-                if (jd.inputTable().isRefreshing()) {
-                    final QueryTable input = (QueryTable) jd.inputTable();
+            for (int ii = 0; ii < joinInputHelpers.length; ++ii) {
```
rcaudy, line 383: There might be some code to re-use here with keyed.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -466,21 +522,21 @@ protected void process() {
...
+            final long[] currentRedirections = new long[tableCount];
+            final boolean[] modifiedTables = new boolean[tableCount];
+            final boolean[] modifiedTablesOnThisRow = new boolean[tableCount];
 
-            final RowSetBuilderRandom modifiedBuilder = RowSetFactory.builderRandom();
-            final RowSetBuilderRandom removedBuilder = RowSetFactory.builderRandom();
-            final RowSetBuilderRandom addedBuilder = RowSetFactory.builderRandom();
+            final RowSetBuilderRandom modifiedBuilder = new BitmapRandomBuilder((int) newSize);
+            final RowSetBuilderRandom emptiedBuilder = new BitmapRandomBuilder((int) newSize);
+            final RowSetBuilderRandom reincarnatedBuilder = new BitmapRandomBuilder((int) newSize);
```
rcaudy, line 531: Maybe use regular random builders for these? Might want to run by Charles as a straw-man.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -489,33 +545,39 @@ protected void process() {
...
+                            rowModified |= (flagValues[tableNumber] & notShift) != 0;
                         }
-                        rowModified |= (flagValues[ii] & notShift) != 0;
                     } else {
                         // If the redirection has changed and we have done anything other than a shift, we must light
                         // up all the columns for the table as a modification. Similarly, if the row was added and
                         // deleted from the original table, then we must also light up all the columns as modified.
-                        if ((flagValues[ii] & addOrRemove) != 0 || currentRedirections[ii] != originalValues[ii]) {
-                            rowModified |= (modifiedTablesOnThisRow[ii] = (flagValues[ii] & notShift) != 0);
+                        if ((flagValues[tableNumber] & addOrRemove) != 0
```
rcaudy, line 568: I'm curious as to why the proposed modification didn't work.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
```diff
@@ -0,0 +1,671 @@
...
+                        rowModified |= (flagValues[ii] & notShift) != 0;
+                    } else {
+                        // If the redirection has changed and we have done anything other than a shift, we must light
+                        // up all the columns for the table as a modification. Similarly, if the row was added and
+                        // deleted from the original table, then we must also light up all the columns as modified.
+                        if ((flagValues[ii] & addOrRemove) != 0 || currentRedirections[ii] != originalValues[ii]) {
+                            rowModified |= (modifiedTablesOnThisRow[ii] = (flagValues[ii] & notShift) != 0);
+                        } else {
+                            rowModified |= (flagValues[ii] & notShift) != 0;
+                        }
```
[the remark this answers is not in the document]
lbooker42: Implemented this and it works.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
lbooker42: I re-implemented and retested and it does work correctly.  I inadvertently reverted this trying to resolve another error.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
lbooker42: I'm not convinced it's worth the effort.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinModifiedSlotTracker.java
```diff
@@ -25,11 +25,14 @@ public class MultiJoinModifiedSlotTracker {
      * The location that we must write to in modified slots; also if we have a pointer that falls outside the range [0,
      * pointer); then we know it is invalid
      */
-    private int pointer;
-    /** Tow many slots we have allocated */
+    private long pointer;
+    /** How many slots we have allocated */
     private int allocated;
```
rcaudy, line 30: Check your `flagSource.ensureCapacity` calls. I think there's risk of integer overflow. You need to cast one of the inputs to `long`.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinModifiedSlotTracker.java
```diff
@@ -8,8 +8,8 @@
 import java.util.List;
 
 public class MultiJoinModifiedSlotTracker {
-    /** The slot that was modified. */
-    private final IntegerArraySource modifiedSlots = new IntegerArraySource();
+    /** The main table row that was modified. */
```
rcaudy, line 11: Talking about main table in this class is misleading. It is only ever concerned with output row position. There are a few comments that need updating.

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/IncrementalMultiJoinStateManagerTypedBase.java
```diff
@@ -442,19 +439,14 @@ public boolean doRehash(boolean fullRehash, MutableInt rehashCredits, int nextCh
 
         Assert.eqZero(rehashPointer, "rehashPointer");
 
-        for (int ii = 0; ii < mainKeySources.length; ++ii) {
-            alternateKeySources[ii] = mainKeySources[ii];
-            mainKeySources[ii] = InMemoryColumnSource.getImmutableMemoryColumnSource(tableSize,
-                    alternateKeySources[ii].getType(), alternateKeySources[ii].getComponentType());
-            mainKeySources[ii].ensureCapacity(tableSize);
-        }
         alternateTableSize = oldTableSize;
```
rcaudy, line 442: Does this assignment effect anything? I think it's immaterial, since we'll ignore any hashed values that are greater than the rehashPointer (which is zero). Some part of me wants to assign it to oldTableSize in newAlternate, and back to 1 in clearAlternate, just for cleanliness.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
lbooker42: Agreed to use BMBuilder for modifications, Random for emptied/reincarnated similar to Agg usage.

engine/table/src/main/java/io/deephaven/engine/table/impl/multijoin/IncrementalMultiJoinStateManagerTypedBase.java
rcaudy: I think what you have done is good.

rcaudy: Although I don't love that you're clearing it in clearAlternate, and assigning it *after* newAlternate, I think this is good enough for now.

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
rcaudy: This can hold for another PR.

rcaudy (approved):

engine/table/src/main/java/io/deephaven/engine/table/impl/MultiJoinTableImpl.java
lbooker42: Created issue https://github.com/deephaven/deephaven-core/issues/4273 to track this.

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2995
Reference: https://github.com/deephaven/deephaven.io/issues/2996

PR #4180, opened 2023-07-12
Remove ClientWrapper factory, reduce definitely/indirectly lost bytes under Valgrind to 0

alexpeters1208:


jcferretti (dismissed):

kosak (dismissed):

R/rdeephaven/src/client.cpp
```diff
@@ -116,17 +116,17 @@ class ClientOptionsWrapper {
         internal_options->addExtraHeader(header_name, header_value);
     }
 
-private:
-
-    deephaven::client::ClientOptions* internal_options;
-    friend ClientWrapper* newClientWrapper(const std::string &target, const ClientOptionsWrapper &client_options);
+    std::shared_ptr<deephaven::client::ClientOptions> internal_options;
```
chipkent, line 119: Is this now public for a reason?

alexpeters1208: `friend ClientWrapper;` would also work. To me it's simpler to keep it public, since none of those attributes get touched by the R level. I am indifferent, unless `friend ClientWrapper;` is better practice.

chipkent: I'm biased to friend, in case RCpp decides to make stuff public now or in the future.

chipkent (approved):

PR #4181, opened 2023-07-12
Add JS API Databar Reading Support

### LINKED ISSUE #3546: Add API support for `databar` column formatting

### LINKED ISSUE #3547: JS API support for `databar` column formatting

ethanalvizo:
Moving this feature to deephaven.UI

Closes #3547 
Closes #3546 

Works in tandem with https://github.com/deephaven/web-client-ui/pull/1415 on the UI side

### Demo Code:
```
from deephaven.column import int_col, double_col, string_col

students = new_table(
    [
        string_col("Name", ["Andy", "Claire", "Jane", "Steven"]),
        int_col("StudentID", [1, 2, 3, 4]),
        int_col("Threshold", [3, 3, 3, 3]),
        double_col("GPA", [-3.0, 4.0, 3.7, 2.8]),
        double_col("DirectionalAxis", [-3.0, 4.0, 3.7, 2.8]),
        double_col("MiddleAxis", [-3.0, 4.0, 3.7, 2.8]),
        double_col("OverlapPlacement", [-3.0, 4.0, 3.7, 2.8]),
        double_col("HidePlacement", [-3.0, 4.0, 3.7, 2.8]),
        double_col("RTLDirection", [-3.0, 4.0, 3.7, 2.8]),
        double_col("Markers", [-3.0, 4.0, 3.7, 2.8]),
        double_col("Markers2", [-3.0, 4.0, 3.7, 2.8]),
        double_col("Gradient", [-3.0, 4.0, 3.7, 2.8]),
    ]
)

data_bars = (
    students.format_data_bar(col="GPA", value_col="GPA")
    .format_data_bar(
        col="DirectionalAxis", value_col="DirectionalAxis", axis="directional"
    )
    .format_data_bar(col="MiddleAxis", value_col="MiddleAxis", axis="middle")
    .format_data_bar(
        col="OverlapPlacement",
        value_col="OverlapPlacement",
        value_placement="overlap",
    )
    .format_data_bar(
        col="HidePlacement", value_col="HidePlacement", value_placement="hide"
    )
    .format_data_bar(
        col="RTLDirection", value_col="RTLDirection", direction="RTL"
    )
    .format_data_bar(
        col="Markers", value_col="Markers", marker_col="Threshold"
    )
    .format_data_bar(
        col="Markers2",
        value_col="Markers2",
        marker_col="StudentID",
        marker_color="#0000FF",
    )
    .format_data_bar(
        col="Gradient",
        value_col="Gradient",
        negative_color=["#FF0000", "#FFFF00"],
        positive_color=["#FFFF00", "#00FF00"],
    )
)
```

### Expected Output:
image


web/client-api/src/main/java/io/deephaven/web/client/api/subscription/ViewportRow.java
```diff
@@ -65,6 +70,37 @@ public Format getFormat(Column column) {
...
             formatString = formatStrings.getAtAsAny(offsetInSnapshot).asString();
         }
-        return new Format(cellColors, rowColors, numberFormat, formatString);
+        if(column.getFormatDatabarColumnIndexRange() != null) {
+            Integer[] range = column.getFormatDatabarColumnIndexRange();
+            Integer start = range[0];
+            Integer end = range[1];
+
+            JsArray<Any> formatDatabarStrings = Js.uncheckedCast(dataColumns[start]);
+            formatDatabarString = formatDatabarStrings.getAtAsAny(offsetInSnapshot).asString();
```
ethanalvizo, line 79: I think this line is causing the error below since disabling that line makes the error go away

image

web/client-api/src/main/java/io/deephaven/web/client/api/Column.java
```diff
@@ -75,6 +78,11 @@ public Format getFormat(TableData.Row row) {
         return row.getFormat(this);
     }
 
+//    @JsMethod
+//    public Format getDatabar(TableData.Row row) {
+//        return row.getFormat(this);
+//    }
+
```
mofojed, line 85: Delete unused

web/client-api/src/main/java/io/deephaven/web/client/api/Column.java
```diff
@@ -217,18 +234,19 @@ public int hashCode() {
...
+        result = 31 * result + (formatDatabarColumnIndexRange != null ? formatDatabarColumnIndexRange.hashCode() : 0);
         result = 31 * result + type.hashCode();
         result = 31 * result + name.hashCode();
         return result;
     }
 
     public Column withFormatStringColumnIndex(int formatStringColumnIndex) {
         return new Column(jsIndex, index, formatColumnIndex, styleColumnIndex, type, name, isPartitionColumn,
-                formatStringColumnIndex, description, isInputTableKeyColumn);
+                formatStringColumnIndex, formatDatabarColumnIndexRange, description, isInputTableKeyColumn);
```
mofojed, line 245: I think you're already refactoring, but yea use a map instead.

web/client-api/src/main/java/io/deephaven/web/client/api/barrage/def/ColumnDefinition.java
```diff
@@ -117,6 +128,19 @@ public void setFormatColumnName(String formatColumn) {
...
+
+    public Collection<String> getDatabarColumnNames() {
+        return databarColumn.values();
+    }
+
+//    public void setDatabarColumnName(Map<String, String> databarColumn) {
+//    public void setDatabarColumnName(String databarColumn) {
+//        this.databarColumn = databarColumn;
+//    }
+
```
mofojed, line 143: Remove unused

web/client-api/src/main/java/io/deephaven/web/client/api/barrage/def/ColumnDefinition.java
```diff
@@ -143,27 +167,46 @@ public String getDescription() {
...
+                databarFormatIndexRange[0] = Integer.MAX_VALUE;
+            }
+            if(databarFormatIndexRange[1] == null) {
+                databarFormatIndexRange[1] = Integer.MIN_VALUE;
+            }
+
+            ColumnDefinition column = byNameMap.get(colName);
+            databarFormatIndexRange[0] = Math.min(databarFormatIndexRange[0], column.getColumnIndex());
+            databarFormatIndexRange[1] = Math.max(databarFormatIndexRange[1], column.getColumnIndex());
+        }
```
mofojed, line 192: Use a map instead

web/client-api/src/main/java/io/deephaven/web/public/table_viewport.html
```diff
@@ -107,6 +107,7 @@ <h3>Table data</h3>
 
         var viewportData = event.detail;
         var rows = viewportData.rows;
+        console.log(JSON.stringify(rows));
```
mofojed, line 110: Remove before merging.

mofojed (changes requested):

engine/table/src/main/java/io/deephaven/engine/table/impl/TableDefaults.java
```diff
@@ -223,6 +226,72 @@ default Table formatColumns(String... columnFormats) {
...
+        if (!axis.equals("proportional") && !axis.equals("middle") && !axis.equals("directional")) {
+            throw new IllegalArgumentException("invalid axis option!");
+        }
+        if (min != null && max != null && min > max) {
+            throw new IllegalArgumentException("min cannot be greater than max!");
+        }
+        if (positiveColor != null) {
+            try {
+                Color c = new Color(positiveColor);
+            } catch (IllegalArgumentException e) {
```
ethanalvizo, line 241: probably a better way to test if the string is a valid color?

web/client-api/src/main/java/io/deephaven/web/client/api/subscription/SubscriptionTableData.java
```diff
@@ -590,7 +646,20 @@ public Format getFormat(long index, Column column) {
                 JsArray<Any> formatStrings = Js.uncheckedCast(data[column.getFormatStringColumnIndex()]);
                 formatString = formatStrings.getAtAsAny(redirectedIndex).asString();
             }
-            return new Format(cellColors, rowColors, numberFormat, formatString);
+            // if (column.getFormatDatabarColumnIndexRange() != null) {
+            // JsArray<Any> formatDatabarStrings = Js.uncheckedCast(data[column.getFormatDatabarColumnIndexRange()]);
+            // formatDatabarString = formatDatabarStrings.getAtAsAny(redirectedIndex).asString();
+            // }
```
mofojed, line 652: Delete unused code

web/client-api/src/main/java/io/deephaven/web/client/api/subscription/ViewportData.java
```diff
@@ -86,6 +86,10 @@ public ViewportData(RangeSet includedRows, Object[] dataColumns, JsArray<Column>
             if (c.getFormatStringColumnIndex() != null) {
                 data[c.getFormatStringColumnIndex()] = dataColumns[c.getFormatStringColumnIndex()];
             }
+            // TODO: Figure out why I don't need this section here
+            // if(c.getFormatDataBarColumnIndices() != null) {
+            // c.getFormatDataBarColumnIndices().values().forEach(idx -> data[idx] = dataColumns[idx]);
+            // }
```
mofojed, line 92: You probably do still need this.
Make a table with like 100 columns, and add databar formatting to the first column (so you can only see the first few columns). We still want to have these columns returned.

web/client-api/src/main/java/io/deephaven/web/client/api/subscription/ViewportRow.java
```diff
@@ -65,6 +66,67 @@ public Format getFormat(Column column) {
...
+        if (column.getFormatDataBarColumnIndices() != null) {
+            formatDataBar = getDataBarFormat(column);
+        }
+        return new Format(cellColors, rowColors, numberFormat, formatString, formatDataBar);
+    }
+
+    public static native void console(Object text)
+    /*-{
+        console.log(text, typeof text);
+    }-*/;
```
mofojed, line 78: Not used, delete

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -165,6 +165,11 @@ private TreeViewportData(double offset, long viewportSize, double treeSize, Colu
                     data[c.getFormatStringColumnIndex()] = dataColumns[c.getFormatStringColumnIndex()].getData();
                 }
 
+                // TODO: Related to the todo in ViewportData.java. Not sure if I need this
+                // if(c.getFormatDataBarColumnIndices() != null) {
+                // c.getFormatDataBarColumnIndices().values().forEach(idx -> data[idx] = dataColumns[idx]);
+                // }
```
mofojed, line 171: Same, I think you need, try with hundreds of columns

mofojed (changes requested):

engine/table/src/main/java/io/deephaven/engine/table/impl/TableDefaults.java
```diff
@@ -223,6 +226,72 @@ default Table formatColumns(String... columnFormats) {
...
+                        TableTools.col(ColumnFormatting.getDataBarFormatColumnName(column,
+                                ColumnFormatting.DataBarFormatColumnType.POSITIVE_COLOR), positiveColor),
+                        TableTools.col(ColumnFormatting.getDataBarFormatColumnName(column,
+                                ColumnFormatting.DataBarFormatColumnType.NEGATIVE_COLOR), negativeColor),
+                        TableTools.col(ColumnFormatting.getDataBarFormatColumnName(column,
+                                ColumnFormatting.DataBarFormatColumnType.VALUE_PLACEMENT), valuePlacement),
+                        TableTools.col(ColumnFormatting.getDataBarFormatColumnName(column,
+                                ColumnFormatting.DataBarFormatColumnType.DIRECTION), direction),
+                        TableTools.col(ColumnFormatting.getDataBarFormatColumnName(column,
+                                ColumnFormatting.DataBarFormatColumnType.OPACITY), opacity)),
```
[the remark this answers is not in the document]
ethanalvizo: Yes it does

ethanalvizo:
Type hints are working for the general `format_data_bar` function but not for each of the params (axis, direction , etc.). They show as a NoneType() rather than their enum type

mofojed (changes requested):

engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -308,6 +308,11 @@ public interface Table extends
     @ConcurrentMethod
     Table formatColumnWhere(String columnName, String condition, String formula);
 
+    @ConcurrentMethod
+    Table formatDataBar(String column, String valueColumn, String axis, Double min, Double max,
```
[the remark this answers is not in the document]
ethanalvizo: I could actually just move it since the enums in ColumnFormatting were meant for the type hints but as shown in the comment above the param specific hints aren't working properly anyways

py/server/deephaven/table.py
```diff
@@ -1949,6 +1980,70 @@ def format_row_where(self, cond: str, formula: str) -> Table:
         except Exception as e:
             raise DHError(e, "failed to color format rows conditionally.") from e
 
+    def format_data_bar(self, column: str, value_column: str = None, min: float = None, max: float = None,
+                        axis: Union[AxisOption, str] = None, positive_color: Union[str, List[str]] = None,
+                        negative_color: Union[str, List[str]] = None, value_placement: Union[ValuePlacementOption, str] = None,
+                        direction: Union[DirectionOption, str] = None, opacity: float = None, marker_column: str = None,
+                        marker_color: str = None) -> Table:
```
chipkent, line 1987: There is a color type that should be supported and included in the type hints.
https://deephaven.io/core/pydoc/code/deephaven.plot.color.html

py/server/deephaven/table.py
```diff
@@ -1949,6 +1980,70 @@ def format_row_where(self, cond: str, formula: str) -> Table:
...
+            column (str): where to place the data bars
+            value_column (str): where to get the values to form the data bars from
+            min (float): minimum value for data bar scaling
+            max (float): maximum value for data bar scaling
+            axis (AxisOption): orientation of data bar relative to cell
+            positive_color (Union[str, List[str]]): color or list of colors for positive bar
+            negative_color (Union[str, List[str]]): color or list of colors for negative bar
+            value_placement (ValuePlacementOption): orientation of values relative to data bar
+            direction (DirectionOption): orientation of data bar relative to horizontal axis
+            opacity (float): opacity of data bars
```
chipkent, line 2000: Should indicate the valid range

py/server/deephaven/table.py
```diff
@@ -1949,6 +1980,70 @@ def format_row_where(self, cond: str, formula: str) -> Table:
...
 
+    def format_data_bar(self, column: str, value_column: str = None, min: float = None, max: float = None,
+                        axis: Union[AxisOption, str] = None, positive_color: Union[str, List[str]] = None,
+                        negative_color: Union[str, List[str]] = None, value_placement: Union[ValuePlacementOption, str] = None,
+                        direction: Union[DirectionOption, str] = None, opacity: float = None, marker_column: str = None,
+                        marker_color: str = None) -> Table:
+        """ Applies data bar formatting to the column specified.
+
+        Args:
+            column (str): where to place the data bars
```
chipkent, line 1991: Arg strings somehow feel stylisticly inconsistent with the rest of the docs.  I can't say why.  Give the existing docs a quick scan and see if there are obvious changes.  @jmao-denver may also have comments.
e.g. "where to get the values to form data bars from" -> "column containing the values to form data bars from" may make the docs more consistent.

chipkent (changes requested):

engine/table/src/main/java/io/deephaven/engine/table/impl/TableDefaults.java
```diff
@@ -223,6 +229,89 @@ default Table formatColumns(String... columnFormats) {
...
+        if (markerColor != null) {
+            validateColor(markerColor);
+        }
+        if (opacity != null && (opacity > 1 || opacity < 0)) {
+            throw new IllegalArgumentException("Opacity must be between 0 and 1.");
+        }
+    }
+
+    @Override
+    default Table formatDataBar(String column, String valueColumn, AxisOptions axis, Double min, Double max,
```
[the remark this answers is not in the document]
ethanalvizo: @chipkent just wanted to double check I understood this comment. The min and max fields should be able to take both a constant or a column name as an input. And in the latter case, the min and max would be obtained from that specified column?

engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -308,6 +308,52 @@ public interface Table extends
     @ConcurrentMethod
     Table formatColumnWhere(String columnName, String condition, String formula);
 
+    enum AxisOptions {
```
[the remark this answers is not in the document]
ethanalvizo: I appended "DataBar" as a prefix to all of the enum options. Don't know if it's a huge concern but the names become a little long:
DataBarAxisOptions, DataBarValuePlacementOptions, DataBarDirectionOptions

py/server/deephaven/table.py
```diff
@@ -97,6 +100,34 @@ class SearchDisplayMode(Enum):
...
+    HIDE = _JValuePlacementOption.Hide
+    """Hide the value."""
+
+
+class DirectionOption(Enum):
+    """An enum of for horizontal axis orientation options for data bar"""
+    LTR = _JDirectionOption.LTR
+    """Left to right"""
+    RTL = _JDirectionOption.RTL
+    """Right to left"""
```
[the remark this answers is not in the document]
jmao-denver: I am OK with its being in the deephaven packageI but I'd like to propose that we create a webui module and move both format_data_bar() and the layout_hints() into it, which means both will not be methods on Table but functions that takes a Table as the 1st arg. This is a breaking change but will result in a clear separation between data ops and UI ops and table.py being less cluttered.  @mattrunyon @mofojed @chipkent your thoughts?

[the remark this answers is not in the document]
mofojed: I'm not sure I'd like the syntax when chaining them all together. For example, in Ethan's code above, he formats a bunch of data bars:
```python
data_bars = (
    students.format_data_bar(column="GPA", value_column="GPA")
    .format_data_bar(
        column="DirectionalAxis", value_column="DirectionalAxis", axis="directional"
    )
    .format_data_bar(column="MiddleAxis", value_column="MiddleAxis", axis="middle")
    .format_data_bar(
        column="OverlapPlacement",
        value_column="OverlapPlacement",
        value_placement="overlap",
    )
#    ... etc
)
```
With what you're proposing, you'd either have to nest all the tables, e.g.:
```python
data_bars = format_data_bars(format_data_bars(format_data_bars(students, ...), ...))
```
or use an intermediate variable (could be the same name), e.g.:
```python
data_bars = format_data_bars(students, ...)
data_bars = format_data_bars(students, ...)
... etc
```

Also would make it less discoverable (as you can see these functions currenty just from typing `t.` and seeing the autocomplete suggestions).

Could separate the format/UI functions as a separate interface but still implement it in the base table?

py/server/deephaven/table.py
ethanalvizo: Was getting a circular import error when I tried using the Color type from deephaven.plot. I do still verify in TableDefaults.java that the str passed in is a valid color. I noticed that the color in layout hints seems to use str as well so I'm not if this is a change we still want and if so I assume I should change it for that one as well?

`Type: <class 'ImportError'>
Value: cannot import name 'Table' from partially initialized module 'deephaven.table' (most likely due to a circular import) (/Users/ethanalvizo/Deephaven/deephaven-core/py/venv/lib/python3.11/site-packages/deephaven/table.py)
`

engine/table/src/main/java/io/deephaven/engine/table/impl/TableDefaults.java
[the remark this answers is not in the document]
mattrunyon: I think the suggestion is there should be 3 options. The min/max value might not be constant for all rows in a table. For example, a table of stock quotes of several symbols. The min/max is the min/max for that stock symbol. User already has min/max for that symbol (at the time of the quote) as columns in the table

```java
// Create a min/max col based on the min/max of valueColumn
Table formatDataBar(String column, String valueColumn, AxisOptions axis)

// Use the names of minCol and maxCol as the min/max cols
Table formatDataBar(String column, String valueColumn, AxisOptions axis, String minCol, String maxCol)

// Use the values of minVal and maxVal for a constant min/max
Table formatDataBar(String column, String valueColumn, AxisOptions axis, Double minVal, Double maxVal)
```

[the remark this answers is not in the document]
chipkent: Yes.

py/server/deephaven/table.py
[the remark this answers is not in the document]
chipkent: I agree with @mofojed 's concerns.  As we have implemented things, formatting really is part of the table.  It isn't independent.

py/server/deephaven/table.py
chipkent: This may be from type hints.  See https://peps.python.org/pep-0484/#forward-references

ethanalvizo: Still same error log when I tried 'Color' so that it would use forward references

web/client-api/src/main/java/io/deephaven/web/client/api/DatabarFormatBuilder.java
```diff
@@ -0,0 +1,75 @@
+package io.deephaven.web.client.api;
+
+public class DatabarFormatBuilder {
```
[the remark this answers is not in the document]
ethanalvizo: I can change it back. Some of the props that use double like min, max, and opacity already have default values so null shouldn't be possible. Since value and marker prop are obtained from another column it could be null

`public DataBarFormat(double min, double max, double value, String axis, String positiveColor, String negativeColor,
            String valuePlacement, String direction, double opacity, double marker, String markerColor)`

py/server/deephaven/table.py
chipkent: The syntax is correct.  The grammar doesn't feel consistent with the other docs.  For example `where to place the data bars` feels inconsistent with what we have in `deephaven.table.Table`.  As an example https://deephaven.io/core/pydoc/code/deephaven.table.html#deephaven.table.PartitionedTable.sort has `the columns(s) to be sorted on.  Can't include the constituent column.`  If similar grammar is applied to `column` in this case, it would end up as something like `the column data bars are placed into`.

engine/table/src/main/java/io/deephaven/engine/table/impl/TableDefaults.java
```diff
@@ -223,6 +229,89 @@ default Table formatColumns(String... columnFormats) {
         return updateView(selectColumns);
     }
 
+    default void validateColor(String color) throws IllegalArgumentException {
```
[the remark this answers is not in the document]
ethanalvizo: Changed this to be `private void validate...` instead so that it isn't exposed to API but still within TableDefaults. Not sure if the BaseTable approach is preferred in this scenario

py/server/deephaven/table.py
chipkent: Then I think you are missing a change somewhere.

chipkent: Or is it circular in imports not typehints?

engine/table/src/main/java/io/deephaven/engine/table/impl/TableDefaults.java
```diff
@@ -240,59 +238,101 @@ default Table formatColumns(String... columnFormats) {
...
     }
 
-    default void validateColor(String color) throws IllegalArgumentException {
-        new Color(color); // Throws if it can't create it
-    }
-
-    default void validateDataBarOptions(String column, String valueColumn, AxisOptions axis, Double min, Double max,
-            String positiveColor, String negativeColor, ValuePlacementOptions valuePlacement,
-            DirectionOptions direction, Double opacity, String markerColumn, String markerColor)
+    private void validateDataBarOptions(String column, String valueColumn, DataBarAxisOptions axis, double min,
```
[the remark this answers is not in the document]
niloc132: Yeah, definitely should check how this behaves with other column types, like BigInteger, BigDecimal, int, long, etc.

engine/table/src/main/java/io/deephaven/engine/table/impl/TableDefaults.java
```diff
@@ -240,59 +238,101 @@ default Table formatColumns(String... columnFormats) {
...
+        if (opacity != QueryConstants.NULL_DOUBLE && (opacity > 1 || opacity < 0)) {
+            throw new IllegalArgumentException("Opacity: " + opacity + ", must be between 0 and 1.");
         }
     }
 
     @Override
-    default Table formatDataBar(String column, String valueColumn, AxisOptions axis, Double min, Double max,
-            String positiveColor, String negativeColor, ValuePlacementOptions valuePlacement,
-            DirectionOptions direction, Double opacity, String markerColumn, String markerColor) {
+    default Table formatDataBar(String column, String valueColumn, DataBarAxisOptions axis, double min, double max,
```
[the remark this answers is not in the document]
ethanalvizo: No it's just below this function on line 324. I overloaded the function and changed min and max to be string

`default Table formatDataBar(String column, String valueColumn, DataBarAxisOptions axis, String min, String max,
            String positiveColor, String negativeColor, DataBarValuePlacementOptions valuePlacement,
            DataBarDirectionOptions direction, double opacity, String markerColumn, String markerColor)`

py/server/deephaven/table.py
```diff
@@ -2049,25 +2050,31 @@ def format_data_bar(self, column: str, value_column: str = None, min: float = No
             if value_column is None:
                 value_column = column
 
+            if isinstance(min, str) and isinstance(max, float):
+                max = value_column
+
+            if isinstance(max, str) and isinstance(min, float):
+                min = value_column
```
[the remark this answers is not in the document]
ethanalvizo: This was for the scenario that the user just passes in one column for either min or max and leaving the other undefined. The default value is NULL_DOUBLE so I was just over-writing that and assigning the value column so that both min and max will be strings if one is defined as a string.

I'm not sure if this is the desired behaviour though if one of them is defined as a column. If we wanted to keep the other param as a constant I can create 4 overloaded methods for formatDataBar

```
formatDataBar(double min, double max)
formatDataBar(str min, double max)
formatDataBar(double min, str max)
formatDataBar(str min, str max)
```

engine/api/src/main/java/io/deephaven/engine/table/Table.java
```diff
@@ -363,10 +373,33 @@ enum DirectionOptions {
         RTL
     }
 
+    /**
+     * Produce a new table with data bars generated at the specified column
```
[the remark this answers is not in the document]
ethanalvizo: The first and only necessary param is column and that dictates which column will get the data bar format. It doesn't always mean the data bars will use the values in that column, that is defined by `value_column`. But it will be generated in the column passed in by the user which is what I meant by "specified column".

py/server/deephaven/table.py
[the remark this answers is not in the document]
ethanalvizo: Just spoke with web team and I think this is the intended behaviour so I will add those additional overloaded methods

py/server/deephaven/table.py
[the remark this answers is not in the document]
mofojed: To elaborate further on my final point from the previous comment - what would the implications be if we added the function to the `Table` prototype/metaclass when the Web UI module was loaded? e.g.
```python
def format_data_bar(self: Table, options: .... ):
    ...

def layout_hints(self: Table, options: ....):
    ...

Table.format_data_bar = format_data_bar
Table.layout_hints = layout_hints
```

This would separate the UI functions into their own module outside of the core table functionality, but the functions would still be available on the table itself (so could be written like my first snippet in previous comment) and available in autocomplete (for discoverability). Not sure where they would end up in our docs though.

Would something like that make sense? Or are there other cons I'm not thinking about?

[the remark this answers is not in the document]
jmao-denver: @mofojed I think what you proposed would work functionally but would mess up auto-complete and docs. Looking at the Java Table interface, as @chipkent mentioned, it is clear that we are treating UI ops as part of Table ops. So unless we are starting to separate the interfaces on the Java side, it doesn't make much sense to do that in Python. All things considered I am happy with the change as it stands now.

py/server/deephaven/table.py
```diff
@@ -1981,6 +2013,76 @@ def format_row_where(self, cond: str, formula: str) -> Table:
         except Exception as e:
             raise DHError(e, "failed to color format rows conditionally.") from e
 
+    def format_data_bar(self, column: str, value_column: str = None, min: Union[float, str] = NULL_DOUBLE, max: Union[float, str] = NULL_DOUBLE,
+                        axis: Union[DataBarAxisOption, str] = None, positive_color: Union[str, List[str]] = None,
+                        negative_color: Union[str, List[str]] = None, value_placement: Union[DataBarValuePlacementOption, str] = None,
+                        direction: Union[DataBarDirectionOption, str] = None, opacity: float = NULL_DOUBLE, marker_column: str = None,
```
jmao-denver, line 2019: The current naming convention for 'column' in parameters is 'col'.

jmao-denver (changes requested):
Minor naming issue.

py/server/deephaven/table.py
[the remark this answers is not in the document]
mofojed: Just to be clear, I checked doing the above and autocomplete seemed to pick it up correctly. We may want to discuss this further when we flesh out programmatic layouts and callbacks.

py/server/deephaven/table.py
ethanalvizo: Touched up the grammar for these. Let me know if it's still not quite up to par

engine/table/src/main/java/io/deephaven/engine/table/impl/TableDefaults.java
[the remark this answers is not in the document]
ethanalvizo: Verified that these work. I converted them into the double type using the coerceToDouble function we have. We loss some precision in the conversion but I don't think it will be noticeable on the UI side for differences that small

py/server/deephaven/table.py
ethanalvizo: Sorry my mistake on this one I needed to remove the import at the top once I changed it to use forward references. I forgot to do that before so the circular import was still technically there

engine/table/src/main/java/io/deephaven/engine/table/impl/TableDefaults.java
```diff
@@ -234,6 +238,157 @@ default Table formatColumns(String... columnFormats) {
...
+
+        String maxColumn =
+                ColumnFormatting.getDataBarFormatColumnName(column, ColumnFormatting.DataBarFormatColumnType.MAX);
+        newTable = newTable.updateView(maxColumn + "=" + max);
+
+        return formatDataBarHelper(newTable, column, valueColumn, axis, positiveColor, negativeColor, valuePlacement,
+                direction, opacity, markerColumn, markerColor);
+    }
+
+    private Table formatDataBarHelper(Table newTable, String column, String valueColumn, DataBarAxisOptions axis,
```
ethanalvizo, line 359: probably a better name I could use here just couldn't think of a better one

py/server/deephaven/table.py
```diff
@@ -544,11 +570,6 @@ def is_refreshing(self) -> bool:
             self._is_refreshing = self.j_table.isRefreshing()
         return self._is_refreshing
 
-    @property
-    def is_blink(self) -> bool:
-        """Whether this table is a blink table."""
-        return _JBlinkTableTools.isBlink(self.j_table)
-
```
niloc132, line 551: from our slack conv, this probably shouldn't have been removed.

engine/table/src/main/java/io/deephaven/engine/table/impl/TableDefaults.java
```diff
@@ -234,6 +238,157 @@ default Table formatColumns(String... columnFormats) {
         return updateView(selectColumns);
     }
 
+    private void validateDataBarOptions(String positiveColor, String negativeColor, double opacity, String markerColor)
+            throws IllegalArgumentException {
+        if (positiveColor != null) {
+            for (String color : positiveColor.split(",")) {
+                new Color(color); // Throws if it can't create it
```
mofojed, line 245: Instead of doing this same patter in multiple places (`new Color(color); // Throws if it can't create it`), put it in it's own method
```
private void validateColor(String color) {
  // Throws if it can't create it
  new Color(color);
}
```

Then it's clear what you're doing in these cases.

engine/table/src/main/java/io/deephaven/engine/util/ColumnFormatting.java
```diff
@@ -74,6 +78,14 @@ static String getNumberFormatColumn(String baseColumn) {
         return baseColumn + Constants.TABLE_NUMBER_FORMAT_SUFFIX;
     }
 
+    enum DataBarFormatColumnType {
+        VALUE, AXIS, MAX, MIN, POSITIVE_COLOR, NEGATIVE_COLOR, VALUE_PLACEMENT, DIRECTION, OPACITY, MARKER, MARKER_COLOR
+    }
+
+    static String getDataBarFormatColumnName(String baseColumn, DataBarFormatColumnType prop) {
+        return baseColumn + "__" + prop.name() + Constants.TABLE_DATA_BAR_FORMAT_SUFFIX;
```
mofojed, line 86: Should lead with two `__` as well

mofojed (changes requested):
I'm seeing the formatting columns in the table, those should be hidden:
image

mofojed:
@ethanalvizo update the PR description to include the newest/correct syntax for your example:
```python
from deephaven.column import int_col, double_col, string_col

students = new_table(
    [
        string_col("Name", ["Andy", "Claire", "Jane", "Steven"]),
        int_col("StudentID", [1, 2, 3, 4]),
        int_col("Threshold", [3, 3, 3, 3]),
        double_col("GPA", [-3.0, 4.0, 3.7, 2.8]),
        double_col("DirectionalAxis", [-3.0, 4.0, 3.7, 2.8]),
        double_col("MiddleAxis", [-3.0, 4.0, 3.7, 2.8]),
        double_col("OverlapPlacement", [-3.0, 4.0, 3.7, 2.8]),
        double_col("HidePlacement", [-3.0, 4.0, 3.7, 2.8]),
        double_col("RTLDirection", [-3.0, 4.0, 3.7, 2.8]),
        double_col("Markers", [-3.0, 4.0, 3.7, 2.8]),
        double_col("Markers2", [-3.0, 4.0, 3.7, 2.8]),
        double_col("Gradient", [-3.0, 4.0, 3.7, 2.8]),
    ]
)

data_bars = (
    students.format_data_bar(col="GPA", value_col="GPA")
    .format_data_bar(
        col="DirectionalAxis", value_col="DirectionalAxis", axis="directional"
    )
    .format_data_bar(col="MiddleAxis", value_col="MiddleAxis", axis="middle")
    .format_data_bar(
        col="OverlapPlacement",
        value_col="OverlapPlacement",
        value_placement="overlap",
    )
    .format_data_bar(
        col="HidePlacement", value_col="HidePlacement", value_placement="hide"
    )
    .format_data_bar(
        col="RTLDirection", value_col="RTLDirection", direction="RTL"
    )
    .format_data_bar(
        col="Markers", value_col="Markers", marker_col="Threshold"
    )
    .format_data_bar(
        col="Markers2",
        value_col="Markers2",
        marker_col="StudentID",
        marker_color="#0000FF",
    )
    .format_data_bar(
        col="Gradient",
        value_col="Gradient",
        negative_color=["#FF0000", "#FFFF00"],
        positive_color=["#FFFF00", "#00FF00"],
    )
)
```

engine/table/src/main/java/io/deephaven/engine/table/impl/TableDefaults.java
ethanalvizo: https://github.com/deephaven/deephaven-core/pull/4181#discussion_r1293732128

Colin had a different perspective. He didn't think it needed to exist since it is just a longer way to write new Color

engine/table/src/main/java/io/deephaven/engine/util/ColumnFormatting.java
ethanalvizo: Do you mean like "__" + baseColumn + "__" +... ? The other formatting columns don't seem to have these two underscores at the start

image

mofojed:
We'll add this to `ui.table` in `deephaven.ui`.

PR #4187, opened 2023-07-14
Stop using temp dir for GWT compile

niloc132:
The file `/tmp/gwtWork/io.deephaven.web.DeephavenApi/compiler/permutation-0.js` is in fact left behind after running `./gradlew prepareCompose`.  Since the file has always the same path regarding of user or user session, this mean two users on the same machine won't be able to build without clashing with each other (because the second user to run won't be able to overwrite the first user's file in /tmp), and for the same user, two parallel executions could clash.

### DIFF
--- a/buildSrc/src/main/groovy/GwtTools.groovy
+++ b/buildSrc/src/main/groovy/GwtTools.groovy
@@ -85,12 +85,10 @@ class GwtTools {
                 // See https://github.com/esoco/gwt-gradle-plugin for all options
                 /** The level of logging detail (ERROR, WARN, INFO, TRACE, DEBUG, SPAM, ALL) */
                 logLevel = "INFO"
-                /** The workDir, where we'll save gwt unitcache.  Use /tmp to avoid cluttering jenkins caches*/
-                workDir = Files.createTempDirectory("gradleGwtWork").toFile()
                 /** Where to write output files */
                 war = warPath
                 /** Compile a report that tells the "Story of Your Compile". */
-                compileReport = true
+                compileReport = false
                 /** Compile quickly with minimal optimizations. */
                 draftCompile = false
                 /** Include assert statements in compiled output. */
--- a/buildSrc/src/main/groovy/GwtTools.groovy
+++ b/buildSrc/src/main/groovy/GwtTools.groovy
@@ -9,6 +9,8 @@ import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.plugins.JavaPlugin
 import org.gradle.api.tasks.compile.JavaCompile
 
+import java.nio.file.Files
+
 /**
  * Helper to simplify / centralize configuring gwt plugins in build files
  */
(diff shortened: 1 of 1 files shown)


jcferretti (approved):

niloc132:
As discussed in slack, i'm going to read a bit more on seeing about preventing some of this from being written to disk, and adding a deleteOnExit for that path.

buildSrc/src/main/groovy/GwtTools.groovy
```diff
@@ -84,7 +86,7 @@ class GwtTools {
                 /** The level of logging detail (ERROR, WARN, INFO, TRACE, DEBUG, SPAM, ALL) */
                 logLevel = "INFO"
                 /** The workDir, where we'll save gwt unitcache.  Use /tmp to avoid cluttering jenkins caches*/
-                workDir = new File(System.getProperty("java.io.tmpdir"), "gwtWork")
+                workDir = Files.createTempDirectory("gradleGwtWork").toFile()
```
JamesXNelson, line 89: Doesn't a new directory defeat the purpose of a unit cache?

You should probably just put it into `build/gwtWork` and then it would go away on a `clean`

niloc132:
The unit cache is a different argument, this is the work dir which is
mostly, but not entirely, cleaned up at the end of each build.

buildSrc/src/main/groovy/GwtTools.groovy
niloc132: The unit cache is a different argument, this is the work dir which is mostly, but not entirely, cleaned up at the end of each build. Work dir holds stuff like incremental output (disabled by default for prod builds), intermediate linker output... some serialized precompile output to pass to out-of-process workers. It looks like there is a deleteOnExit for nearly everything, but apparently not for the dir itself, or the the permutation-N.js file (intermediate linker output). For these reasons, I think it probably makes more sense to remove the workDir param entirely - after a successful build, the full contents is less than 9kb for DHC, 14kb for DHE for just one file each.

The unit cache is disabled by default, and has to be explicitly given a directory to work in via system property `gwt.persistentunitcachedir`. If not provided, it will default to a sibling directory to the war dir.

The gradle plugin selected for GWT in our projects unconditionally puts the unit cache dir in the build dir:
https://github.com/esoco/gwt-gradle-plugin/blame/dca725f60a7e75dc9de11ea98d5c07d7f566a842/src/main/java/de/esoco/gwt/gradle/command/CompileCommand.java#L53

The unit cache might make sense to override, a dozen or so files, 120kb, but it isn't immediately obvious how we would do that.

niloc132:
I've updated the code changes to remove the work dir setting itself (since the output is tiny, and let's us sidestep these concerns entirely), and turn off the SOYC html report (so as to not write unnecessary/unused details to disk that would need to be cached).

devinrsmith (approved):

PR #4189, opened 2023-07-14
Optimize allocations and copies for SortedRanges.insert when it is effectively an append

jcferretti:
Related to DHE https://github.com/deephaven-ent/iris/pull/651 and https://deephaven.atlassian.net/browse/DH-15310

### DIFF
--- a/engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/sortedranges/SortedRanges.java
+++ b/engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/sortedranges/SortedRanges.java
@@ -2091,7 +2091,7 @@ public abstract class SortedRanges extends RefCountedCow<SortedRanges> implement
         // there are two cases, depending on whether we need to merge our last range with
         // other's first range.
         final SortedRanges result;
-        final int otherFirstPosToRead;  // always the beginning of a range or a singleton (eg, > 0)
+        final int otherFirstPosToRead; // always the beginning of a range or a singleton (eg, > 0)
         final int ourFirstPosToWrite;
         long unpackedLast = unpackedGet(count - 1);
         if (Math.abs(unpackedLast) + 1 == other.first()) {
--- a/engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/sortedranges/SortedRanges.java
+++ b/engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/sortedranges/SortedRanges.java
@@ -2087,8 +2087,82 @@ public abstract class SortedRanges extends RefCountedCow<SortedRanges> implement
         return insertImpl(other, true);
     }
 
+    public final SortedRanges mergeAppend(final SortedRanges other, final boolean writeCheck) {
+        // there are two cases, depending on whether we need to merge our last range with
+        // other's first range.
+        final SortedRanges result;
+        final int otherFirstPosToRead;  // always the beginning of a range or a singleton (eg, > 0)
+        final int ourFirstPosToWrite;
+        long unpackedLast = unpackedGet(count - 1);
+        if (Math.abs(unpackedLast) + 1 == other.first()) {
+            final boolean weEndInRange = unpackedLast < 0;
+            long otherKeyAtPos1 = -1;
+            if (other.cardinality > 1 && (otherKeyAtPos1 = other.unpackedGet(1)) < 0) {
+                // we are merging a full range from other's front at our end.
+                if (weEndInRange) {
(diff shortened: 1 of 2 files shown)


engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/sortedranges/SortedRanges.java
```diff
@@ -2087,9 +2087,83 @@ public final OrderedLongSet insertImpl(final SortedRanges other) {
         return insertImpl(other, true);
     }
 
+    public final SortedRanges mergeAppend(final SortedRanges other, final boolean writeCheck) {
+        // there are two cases, depending on whether we need to merge our last range with
+        // other's first range.
+        final SortedRanges result;
+        final int otherFirstPosToRead; // always the beginning of a range or a singleton (eg, > 0)
+        final int ourFirstPosToWrite;
+        long unpackedLast = unpackedGet(count - 1);
```
rcaudy, line 2096:
```suggestion
        final long unpackedLast = unpackedGet(count - 1);
```

engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/sortedranges/SortedRanges.java
```diff
@@ -2087,9 +2087,83 @@ public final OrderedLongSet insertImpl(final SortedRanges other) {
...
+                return null;
+            }
+            otherFirstPosToRead = 0;
+            ourFirstPosToWrite = count;
+        }
+        // copy over the rest.
+        int ourPosToWrite = ourFirstPosToWrite;
+        for (int otherPosToRead = otherFirstPosToRead; otherPosToRead < other.count; ++otherPosToRead) {
+            result.unpackedSet(ourPosToWrite++, other.unpackedGet(otherPosToRead));
+        }
```
rcaudy, line 2150: This looks like a candidate for an array copy, but I assume it's written this way to address different offsets and different internal widths?

rcaudy (commented):
Some pedantic suggestions, an optimization question, and a suggestion for further optimization. Change looks correct as-is, however.

engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/sortedranges/SortedRanges.java
jcferretti: Yeah, because essentially we are not guaranteed that our two sortedRanges are of the same underlying "type" (eg if they are SortedRangesInt or SortedRangesShort, the unpack part may apply an offset and other fun).
We "could" [TM] optimize to array copy if both are SortedRangesLong (no offset to apply in that case).

engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/sortedranges/SortedRanges.java
```diff
@@ -2087,9 +2087,83 @@ public final OrderedLongSet insertImpl(final SortedRanges other) {
...
+        return result;
+    }
+
     public final OrderedLongSet insertImpl(final SortedRanges other, final boolean writeCheck) {
+        if (isEmpty()) {
+            return other.cowRef();
+        }
         // we know other can't be empty.
-        if (!USE_RANGES_ARRAY || isEmpty() || last() < other.first()) {
+        if (last() < other.first()) {
```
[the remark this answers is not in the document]
jcferretti: I thought about it but punted due to complexity; I'd like the delta to go to Bard and Genesis so on the enterprise PR I think that is a good tradeoff.   We could just do more here in the community PR, and also port that once the code rolls forward to vermillion to vermillion itself.  But the truth is I shouldn't be working on any of this due to priorities, so I don't want to put another couple of hours just now.  Is not obvious to me how common that additional optimization would be hit either, so potentially questionable also.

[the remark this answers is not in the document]
rcaudy: I think this is a reasonable answer. I think the case you've optimized is common. I'm certain we *could* also handle the overlapping case at the cost of a bit more complexity, but I'm not sure that a single overlapping range is much more common than general overlapping.

engine/rowset/src/main/java/io/deephaven/engine/rowset/impl/sortedranges/SortedRanges.java
rcaudy: Ack. This is exactly what I expected.

rcaudy (dismissed):

PR #4190, opened 2023-07-14
Simplify ScriptSession dependencies

niloc132:
This removes the only python init script, and removes an unused groovy init script. The config variables are left in place for now, though they will default to blank.

One slightly unexpected change that this results in is that the `jpy`, `sys`, and `os` modules are no longer imported by default into a python script session - this probably should be considered a feature, but does need to be documented.

In a future patch I'll propose unifying how init scripts are run, so that there is no potential loss in functionality. If reviewers would prefer, I can just go ahead and add those changes here too, though my aim was to let this be a smaller step by itself.


devinrsmith (approved):
Definitely in favor of these types of changes.

jmao-denver (approved):
LGTM

py/server/deephaven_internal/jvm/__init__.py
```diff
@@ -66,3 +66,25 @@ def init_jvm(*args, **kwargs):
             "Unable to initialize JVM, try setting the environment variable JAVA_HOME (JDK 11+ required)") from e
     ready()
     return result
+
+def init_py():
+    """Finishes starting Python to be usable from inside of a Java process. Not intended to be called in cases
+    where there process was start as Python, and Java was started from inside Python.
```
chipkent, line 72:
```suggestion
    where there process was started as Python, and Java was started from inside Python.
```

py/server/deephaven_internal/jvm/__init__.py
```diff
@@ -66,3 +66,25 @@ def init_jvm(*args, **kwargs):
...
+    import sys
+    from deephaven_internal.stream import TeeStream
+
+    # Set stdin to /dev/null to prevent functions (like help()) that attempt to read from stdin from hanging python
+    # execution from within Java.
+    os.dup2(os.open("/dev/null", os.O_RDONLY), 0)
+
+    jpy.VerboseExceptions.enabled = True
+    # If you want jpy to tell you about all that it is doing, change this
+    # jpy.diag.flags = jpy.diag.F_ALL
```
chipkent, line 86: Should this be an ENV controlled option to help us debug users?

chipkent (changes requested):

engine/table/build.gradle
```diff
@@ -10,15 +10,6 @@ sourceSets {
     main {
         resources {
```
rcaudy, line 11: Any chance we want to move these resources to src/main/resources to standardize?

rcaudy (approved):

py/server/deephaven_internal/jvm/__init__.py
niloc132: also fixed there->the

py/server/deephaven_internal/jvm/__init__.py
niloc132: That seems like a decent idea, but in a different PR (i'm just moving the contents of this file so we have one fewer way to screw up startup).

chipkent (approved):

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3211
Reference: https://github.com/deephaven/deephaven.io/issues/3210

PR #4191, opened 2023-07-14
[WIP] Improved handling of writing NULLs in parquet

### LINKED ISSUE #1548: Parquet: Null-encoding for primitives is broken/disabled
To make our Parquet data properly intelligible to other systems, we should be recording `null`s in the standard fashion. Many layers of the integration code already intend to support this capability, but the lower level writing code disables it for primitives. We should fix this immediately.

malhotrashivam:
Bug fix, closes #1548 

**Original State:**
Before this change, nulls were handled incorrectly for primitive types as well as for arrays and vectors, such that null values were written as Deephaven null representations in the parquet file. For example, a null in an integer column would be written as QueryConstants.NULL_INT in the parquet file. Therefore, any reader other than deephaven's parquet reader would not be able to process it correctly.

**Change description:**
This PR fixes multiple small bugs in both java and python code for processing all nulls in a standard manner such that instead of writing different Deephaven null representations to the parquet file, we encode them by storing offsets of null values. #4186 was also discovered and fixed along with this PR for proper representation of floats in the parquet file.

Along with that, we have done refactoring in the ParquetFileWriter class to make the code more readable and add more documentation.


extensions/parquet/base/src/main/java/io/deephaven/parquet/base/AbstractBulkValuesWriter.java
```diff
@@ -41,6 +41,18 @@ public int writeBulkVector(@NotNull final BUFFER_TYPE bulkValues,
...
 
+    /**
+     * This method is used to provide a specialized definition of NULL for PlainIntChunkedWriter. For other writers,
+     * this method does nothing
+     *
+     * @param nullDefinition The specialized value of NULL to consider while writing
+     */
+    @Override
+    public void setNull(int nullDefinition) {
+        // TODO Should I raise an error here since this should not be called, should only be called for the
```
lbooker42, line 52: It's a pretty common pattern to throw an UnsupportedOperationException for functions that should never be called.

jmao-denver (dismissed):
The Python changes look good.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/PlainIntChunkedWriter.java
```diff
@@ -21,6 +21,12 @@
  */
 public class PlainIntChunkedWriter extends AbstractBulkValuesWriter<IntBuffer> {
     private static final int MAXIMUM_TOTAL_CAPACITY = Integer.MAX_VALUE / Integer.BYTES;
+
+    // This value will be considered as a NULL while writing. For example, if NULL_DEF is set as 65535, then 65535 will
+    // be written as NULL by this writer. By default, we use the definition from QueryConstants but can be overridden.
+    // TODO Can there be a better name for this?
+    private int NULL_DEF = QueryConstants.NULL_INT;
```
rcaudy, line 28: Convention is to use ALL_CAPS_WITH_UNDERSCORES for `static final` constants only.
```suggestion
    private int nullValue = QueryConstants.NULL_INT;
```

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/PlainIntChunkedWriter.java
```diff
@@ -125,6 +132,17 @@ public WriteResult writeBulkFilterNulls(@NotNull final IntBuffer bulkValues,
...
 
+    /**
+     * This method is used to provide a specialized definition of NULL while writing. The method is useful for Byte,
+     * Char, and Short data type which also used the int writer but have a different definition of NULL than default.
+     *
+     * @param nullDefinition The specialized value of NULL to consider while writing
+     */
+    @Override
+    public void setNull(int nullDefinition) {
+        NULL_DEF = nullDefinition;
```
rcaudy, line 143:
```suggestion
     * This method is used to provide a type-specific {@code null} representation for writing. The method is useful for Byte,
     * Char, and Short data type which are written as primitive ints but have a different definition of {@code null} than default.
     *
     * @param nullValue The value that represents {@code null} for this writer
     */
    @Override
    public void setNullValue(int nullValue) {
        this.nullValue = nullValue;
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -600,10 +599,8 @@ private static <DATA_TYPE> void encodePlain(@NotNull final ParquetInstructions w
                         repeatCount.limit(lenChunk.size());
                         columnWriter.addVectorPage(bufferToWrite, repeatCount, transferObject.rowCount());
                         repeatCount.clear();
-                    } else if (supportNulls) {
-                        columnWriter.addPage(bufferToWrite, transferObject.rowCount());
                     } else {
-                        columnWriter.addPageNoNulls(bufferToWrite, transferObject.rowCount());
+                        columnWriter.addPage(bufferToWrite, transferObject.rowCount(), columnType);
```
rcaudy, line 603: Does this mean we can delete `addPageNoNulls`?

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/PlainIntChunkedWriter.java
```diff
@@ -105,6 +111,7 @@ public WriteResult writeBulkFilterNulls(@NotNull final IntBuffer bulkValues,
         return new WriteResult(rowCount);
     }
 
+    // TODO Test this method
```
rcaudy, line 114: Be sure that TODOs are deleted, or reference a GitHub issue.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnWriterImpl.java
```diff
@@ -205,6 +217,7 @@ public void addVectorPage(
         }
         initWriter();
         // noinspection unchecked
+        // TODO Should I also change this path? I don't fully understand this path. Talk to Ryan for this.
```
rcaudy, line 220: Almost certainly. This path (I think) is used for writing vectors or arrays, which effectively have one (nested) column of values, and one (outer) column of offsets into the values specifying the various vectors/arrays.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/BulkWriter.java
```diff
@@ -97,4 +97,12 @@ int writeBulkVector(@NotNull final BUFFER_TYPE bulkValues,
      */
     ByteBuffer getByteBufferView() throws IOException;
 
+    /**
+     * This function is used to provide a specialized definition of NULL for this writer. For example, if setNull(65535)
+     * is called, then 65535 will be written as NULL by this writer
+     *
+     * * @param nullValue The value which will be interpreted as NULL for this writer
+     */
+    void setNull(int nullValue);
```
rcaudy, line 106: See proposed changes to naming and JavaDoc in `PlainIntChunkedWriter`. This should get most of those changes, possibly absent the specific commentary about other types.

If we don't make changes to just figure out `nullValue` in the constructor, I think we should consider making this take `Object`, so we can blindly call it for every type of `ColumnWriter`. The non-int implementations can just verify that they got the expected value. The int implementation can unbox the value to an `int`, probably by casting to `Number` and calling `intValue()`.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/AbstractBulkValuesWriter.java
```diff
@@ -41,6 +41,17 @@ public int writeBulkVector(@NotNull final BUFFER_TYPE bulkValues,
...
 
+    /**
+     * This method is used to provide a specialized definition of NULL for PlainIntChunkedWriter. For other writers,
+     * this method does nothing.
+     *
+     * @param nullDefinition The value which will be interpreted as NULL for this writer
+     */
+    @Override
+    public void setNull(int nullDefinition) {
+        throw new UnsupportedOperationException("This method should not be called");
```
rcaudy, line 52: This doesn't look like "doing nothing". Maybe you'd prefer to leave this method abstract? Also, no need to replicate the JavaDoc at every layer, unless you want to be more specific.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnWriter.java
```diff
@@ -13,7 +13,7 @@ public interface ColumnWriter extends SafeCloseable {
 
     void addDictionaryPage(Object dictionaryValues, int valuesCount) throws IOException;
 
-    void addPage(Object pageData, int valuesCount) throws IOException;
+    void addPage(Object pageData, int valuesCount, final Class columnType) throws IOException;
```
rcaudy, line 16: I wonder if we can add the type information to the `ColumnWriter's` constructor, rather than doing this every page?

rcaudy (commented):
Change makes sense to me, good find. I suggested a bit of a refactoring for how we implement it, and I think we need to extend it to vectors and arrays.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
malhotrashivam: We cannot do that because the method is called for string columns if it the column has no nulls

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/AbstractBulkValuesWriter.java
malhotrashivam: I was a bit confused here. So all writers inherit from AbstractBulkValuesWriter class which inherits from the abstract class BulkWriter. 
I wanted to give a specialized definition of this method for PlainIntChunkedWriter and raise an exception for all other chunkedWriter types. So I provided a default implementation in this class, so that the other chunkedWriter types don't have to implement it. Do you think there is a better way to implement this?

rcaudy: I like the solution you've arrived at. This is better done at construction time.

In a hypothetical world where we didn't do that, see my other comment about making this take `Object` and thus being generic.

At that point, we'd have a decision to make:
1. add the null-value override functionality to all writers, or
1. throw a UOE as you've done for writers that don't support that, or
1. have `setNull` be a no-op for other writers.

The JavaDoc implies (to me) that you've taken approach 3, but you've actually taken approach 2.

malhotrashivam: The inconsistency between the JavaDoc and the implementation was my mistake, thanks for pointing that out.
Even though we are not doing this now, I like the approach 1 where the functionality is more generic.

malhotrashivam:
> I am testing it right now. The writing logic works but the reading logic seems to be having some problems in Python. So debugging that.

rcaudy (commented):
Ready to approve this, holding off since I understand there are still test issues.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -426,6 +426,16 @@ static <DATA_TYPE> void writeColumnSource(
...
 
+        // These two lists are parallel, where each element represents a single page. The rawItemCountPerPage
+        // contains the number of items (the sum of the array sizes for each containing row) contained within the
+        // page.
+        final TIntArrayList rawItemCountPerPage = new TIntArrayList();
+
+        // The originalRowsPerPage contains the number of arrays (rows) from the original table contained within
+        // the page.
+        final TIntArrayList originalRowsPerPage = new TIntArrayList();
+        boolean case1 = false;
```
malhotrashivam, line 437: This is the just a hack to make the code work.
Mainly the idea is that the variables rowStepGetter and valuesStepGetter are LongSupplier variables.
In case 1 they are initialized with            
            rowStepGetter = originalRowsPerPage.iterator()::next;
            valuesStepGetter = rawItemCountPerPage.iterator()::next;
            (check line 498-499) 
And in other case, they are initialized with a constant 
           rowStepGetter = valuesStepGetter = () -> finalTargetSize;
            (check line 510)
In case 1, they are used to step through the entries in case we have arrays, In step 2, they don't do much.

Now for strings, we have two code paths, we first try to encode them with dictionary and then plain. In case the dictionary encoding fails, we have already stepped through the valuesStepGetter. So we have to reset it before trying to encode plain, otherwise it gives an exception.
So in this hack, I just remembered the initial state and which case it was. But I would really appreciate suggestions on how to implement this better.

py/server/tests/test_parquet.py
```diff
@@ -183,7 +179,40 @@ def test_round_trip_data(self):
             # "nullBigDecColumn = (java.math.BigDecimal)null",
             # "nullBigIntColumn = (java.math.BigInteger)null"
         ])
+        return dh_table
+
+    def get_table_with_array_data(self):
+        # create a table with columns to test different types and edge cases
+        dh_table = empty_table(20).update(formulas=[
+            "someStringArrayColumn = new String[] {i % 10 == 0?null:(`` + (i % 101))}",
```
malhotrashivam, line 187: Is there a way to convert vector columns into arrays instead of doing the way I have done?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -537,6 +540,13 @@ static <DATA_TYPE> void writeColumnSource(
                         maxValuesPerPage,
                         maxOriginalRowsPerPage,
                         pageCount);
+                if (!usedDictionary) {
+                    // Need to reset the state so that all values can again be read again for plain encoding
+                    if (case1) {
+                        rowStepGetter = originalRowsPerPage.iterator()::next;
+                        valuesStepGetter = rawItemCountPerPage.iterator()::next;
+                    }
+                }
```
rcaudy, line 549:
```suggestion
                if (!usedDictionary && case1) { /* Maybe rawItemCountPerPage != null instead of case1 */
                    // Need to reset the state so that all values can again be read again for plain encoding
                    rowStepGetter = originalRowsPerPage.iterator()::next;
                    valuesStepGetter = rawItemCountPerPage.iterator()::next;
                }
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -426,6 +426,16 @@ static <DATA_TYPE> void writeColumnSource(
...
         LongSupplier valuesStepGetter;
 
+        // These two lists are parallel, where each element represents a single page. The rawItemCountPerPage
+        // contains the number of items (the sum of the array sizes for each containing row) contained within the
+        // page.
+        final TIntArrayList rawItemCountPerPage = new TIntArrayList();
+
+        // The originalRowsPerPage contains the number of arrays (rows) from the original table contained within
+        // the page.
+        final TIntArrayList originalRowsPerPage = new TIntArrayList();
```
rcaudy, line 436: Your comment is explanatory and helpful. The lists themselves could be allocated only in the case where they are used, and assigned to `null` in the other case, which also lets you eliminate the `case1` boolean.

I think the names `rowStepGetter`, `valuesStepGetter`, `rawItemCountPerPage`, and `originalRowsPerPage` could be improved; your comment goes a long way towards improving the internal documentation in this code, however.

One approach we could take would be to clean up the whole mess a bit further by introducing some kind of structure around the two cases. It would be nice to have a way to say that having `lengthSource`, a `rowSet` that isn't the `tableRowSet`/`originalRowSet`, the two lists, and the length supplier could be all or none. That would also give us a place to centralize the documentation, instead of having it splayed across three methods, with variable naming differences between the caller and callees, etc.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -537,6 +540,13 @@ static <DATA_TYPE> void writeColumnSource(
                         maxValuesPerPage,
                         maxOriginalRowsPerPage,
                         pageCount);
+                if (!usedDictionary) {
```
rcaudy, line 543: Just hanging this here, since I can't comment on code you didn't edit. In `tryEncodeDictionary`, there's an early return for "all nulls".
```
            if (keyCount == 0 && hasNulls) {
                return false;
            }
```
I think that could be moved up to above the block that starts with 
```
            List<IntBuffer> arraySizeBuffers = null;
            if (lengthSource != null) {
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
rcaudy: This was a good find, and the fix is both surgical and suitable. About the only thing I would do differently is try to batch all of this up in a single helper class for "array-encoded" data.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -437,30 +490,27 @@ static <DATA_TYPE> void writeColumnSource(
...
-            final TIntArrayList rawItemCountPerPage = new TIntArrayList();
-
-            // The originalRowsPerPage contains the number of arrays (rows) from the original table contained within
-            // the page.
-            final TIntArrayList originalRowsPerPage = new TIntArrayList();
+            final Table ungroupedArrays = lengthsTable.ungroup("array");
+            columnWritingData.initVectorProcessing(lengthSource, ungroupedArrays.getRowSet());
+            columnSource = ungroupedArrays.getColumnSource("array");
+            // TODO I want to mark the variables lengthsTable, lengthSource and ungroupedArrays as unusable here. How
+            // should I do that? Should I use some form of scoping?
```
malhotrashivam, line 501: TODO Need suggestions for this

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -411,6 +411,63 @@ private static ParquetFileWriter getParquetFileWriter(
                 writeInstructions.getCompressionCodecName(), extraMetaData);
     }
 
+    /**
+     * This is a helper struct storing all the useful data required to write column source in the parquet file
+     */
+    static class ColumnWritingData {
```
malhotrashivam, line 417: Any suggestions for the name

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -475,17 +475,17 @@ private VectorColumnWriterHelper(
             lengthPageSizes = new TIntArrayList();
         }
 
+        public boolean isVectorFormat() {
+            return true;
+        }
```
malhotrashivam, line 480: I added this just to rerun the jobs, nothing significant.

jmao-denver (approved):
The Python changes LGTM

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -411,6 +410,84 @@ private static ParquetFileWriter getParquetFileWriter(
                 writeInstructions.getCompressionCodecName(), extraMetaData);
     }
 
+    private interface ColumnWriteHelper {
+
+        IntSupplier valuePageSizeSupplier();
+
+        boolean isVectorFormat();
```
rcaudy, line 417:
```suggestion
        boolean isVectorFormat();

        IntSupplier valuePageSizeSupplier();
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -411,6 +410,84 @@ private static ParquetFileWriter getParquetFileWriter(
...
+            this.maxValuePageSize = maxValuePageSize;
+        }
+
+        public IntSupplier valuePageSizeSupplier() {
+            return () -> maxValuePageSize;
+        }
+
+        public boolean isVectorFormat() {
+            return false;
+        }
```
rcaudy, line 440:
```suggestion
        public boolean isVectorFormat() {
            return false;
        }

        public IntSupplier valuePageSizeSupplier() {
            return () -> maxValuePageSize;
        }
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -420,215 +497,211 @@ static <DATA_TYPE> void writeColumnSource(
...
-            final int finalTargetSize = getTargetRowsPerPage(columnSource.getType(),
-                    writeInstructions.getTargetPageSize());
-            rowStepGetter = valuesStepGetter = () -> finalTargetSize;
-            maxValuesPerPage = maxOriginalRowsPerPage = (int) Math.min(rowSet.size(), finalTargetSize);
-            pageCount = (int) (rowSet.size() / finalTargetSize + ((rowSet.size() % finalTargetSize) == 0 ? 0 : 1));
+            final long tableSize = tableRowSet.size();
+            final int targetPageSize = getTargetRowsPerPage(
+                    valueSource.getType(), writeInstructions.getTargetPageSize());
+            helper = new FlatColumnWriterHelper(targetPageSize);
+            maxValuesPerPage = maxRowsPerPage = (int) Math.min(tableSize, targetPageSize);
```
rcaudy, line 575:
```suggestion
            maxValuesPerPage = maxRowsPerPage = (int) Math.min(tableSize, targetPageSize);
            helper = new FlatColumnWriterHelper(maxValuesPerPage);
```

rcaudy (commented):
Slight suggestions

rcaudy (approved):

PR #4199, opened 2023-07-18
Don't use ExecutionContext to read constants

niloc132:
Instead of reading a constant value from the ExecutionContext, get it from the same root source. This isn't actually used for threading, so much as deciding how much to split up work, so this value is more of a hint than a requirement to have.

Also changes the Math.max to instead server as a default, so that the user-specified value can be either higher _or_ lower than the number of processors.


rcaudy (approved):
I guess we ported it like this, and then propagated when UG moved into EC. Good fix.

PR #4201, opened 2023-07-18
Calendar API cleanup

chipkent:
Clean up for the calendar and business calendar.


alexpeters1208 (commented):
Partially done.

py/server/deephaven/calendar.py
```diff
@@ -1,509 +1,136 @@
...
+        _JCalendars.removeCalendar(name)
     except Exception as e:
-        raise DHError(e, "failed to obtain the available calendar names.") from e
+        raise DHError(e, f"failed to remove calendar '{name}'") from e
 
 
-def default_calendar_name():
-    """ Returns the default calendar name which is set by the 'Calendar.default' property in the configuration file
-    that the Deephaven server is started with.
+def add_calendar(cal: Optional[BusinessCalendar] = None, file: Optional[str] = None) -> None:
```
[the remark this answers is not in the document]
chipkent: There is intentionally no setter so that the calendar is immutable.  The name comes from the configuration.

py/server/deephaven/calendar.py
```diff
@@ -1,509 +1,136 @@
...
 
-        Raises:
-            DHError
-        """
-        try:
-            j_days = self.j_calendar.nonBusinessDaysInRange(start, end)
-            return list(j_days)
-        except Exception as e:
-            raise DHError(e, "failed in non_business_days_in_range.") from e
+def calendar_names() -> List[str]:
```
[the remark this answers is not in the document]
chipkent: way too long for no benefit.

py/server/deephaven/calendar.py
```diff
@@ -1,509 +1,136 @@
...
-            DHError
-        """
-        try:
-            return self.j_calendar.numberOfNonBusinessDays(start, end, end_inclusive)
-        except Exception as e:
-            raise DHError(e, "failed in number_of_non_business_days.") from e
 
-    def is_last_business_day_of_month(self, date: str) -> bool:
-        """ Returns if the specified date is the last business day of the month.
+def calendar(name: Optional[str] = None) -> BusinessCalendar:
```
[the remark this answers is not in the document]
chipkent: The extra length doesn't buy anything and makes users type more.

alexpeters1208 (commented):
Another partial review, got some smaller stuff out of the way.

py/server/deephaven/calendar.py
jmao-denver, line 33:
```suggestion
def add_calendar(cal: Union[BusinessCalendar，str]) -> None:
```
I'd prefer this signature design which saves us from having to do all the check on the input below.

py/server/deephaven/calendar.py
```diff
@@ -1,509 +1,136 @@
...
-            bool
 
-        Raises:
-            DHError
-        """
-        try:
-            return self.j_calendar.isLastBusinessDayOfWeek(date)
-        except Exception as e:
-            raise DHError(e, "failed in is_last_business_day_of_week.") from e
+#TODO: to_numpy_calendar method
```
jmao-denver, line 136: remove?

jmao-denver:
Maybe this is an ignorant comment, but I'd like to see a Python test case that demos whether and how a calendar can be used in a query. I am guessing that the lack of a Python wrapper indicates that a business calendar isn't to be used directly in Python, but more likely through setting the default calendar and referencing the calendar entirely in Java in a formula.

py/server/deephaven/calendar.py
```diff
@@ -1,509 +1,136 @@
...
-    @property
-    def is_business_day(self) -> bool:
-        """ If today is a business day. """
-        return self.j_calendar.isBusinessDay()
 
-    @property
-    def default_business_periods(self) -> List[str]:
-        """ The default business periods for the business days. Returns a list of strings with a comma separating
-        open and close times. """
+def set_calendar(name: str) -> None:
```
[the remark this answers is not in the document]
jjbrosnan: If the default calendar is saved across sessions (e.g. the default stays the same when Deephaven is brought down and back up), then I agree with Alex's name suggestion.

py/server/deephaven/calendar.py
```diff
@@ -1,509 +1,136 @@
...
-            int
 
-        Raises:
-            DHError
-        """
-        try:
-            return self.j_calendar.numberOfBusinessDays(start, end, end_inclusive)
-        except Exception as e:
-            raise DHError(e, "failed in number_of_business_days.") from e
+def calendar_name() -> str:
```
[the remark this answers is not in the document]
jjbrosnan: It seems like this is for consistency with the Java API. But I agree that the lack of difference in the method names `calendar_name` and `calendar_names` could cause confusion.

jjbrosnan (commented):
First glance review. Will take a more in-depth look another time. Python code LGTM so far.

alexpeters1208 (commented):
A little more review

engine/time/src/main/java/io/deephaven/time/calendar/BusinessPeriod.java
```diff
@@ -5,56 +5,70 @@
...
+        if (startTime.compareTo(endTime) > 0) {
             throw new IllegalArgumentException("Start is after end: startTime=" + startTime + " endTime=" + endTime);
         }
+
+        this.nanos = start.until(end, ChronoUnit.NANOS);
     }
 
     /**
-     * Returns the start of the period.
+     * Start of the period.
```
alexpeters1208, line 48: Can you comment on the type here? "Start of the period as an ISO-formatted..." or "... in nanoseconds since the epoch". Does it depend on how this class is instantiated, since it's templated with `T`?

alexpeters1208 (commented):
A little more review

engine/time/src/main/java/io/deephaven/time/calendar/Calendars.java
```diff
@@ -152,126 +92,143 @@ private void load(final String businessCalendarConfig)
...
-    }
+    // region Methods
 
-    @Override
-    public BusinessCalendar put(String key, BusinessCalendar value) {
-        Require.neqNull(key, "key");
-        key = key.toUpperCase();
-        return calendars.put(key, value);
+    /**
+     * Sets the default calendar name.
```
alexpeters1208, line 175: What does this mean in the context of a "collection of calendars"?

engine/time/src/main/java/io/deephaven/time/calendar/Calendars.java
```diff
@@ -152,126 +92,143 @@ private void load(final String businessCalendarConfig)
...
+     *
+     * @param name calendar name
```
alexpeters1208, line 177: Does this set the name of the collection? Or the name of a specific calendar? The name of the default?

alexpeters1208 (commented):
Almost done with review

engine/time/src/test/java/io/deephaven/time/calendar/TestBusinessCalendar.java
```diff
@@ -0,0 +1,1506 @@
+/**
+ * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
```
alexpeters1208, line 2:
```suggestion
 * Copyright (c) 2016-2023 Deephaven Data Labs and Patent Pending
```

alexpeters1208 (commented):
Someone with more Java experience should look over the Java code, especially the code generation stuff.

py/server/deephaven/calendar.py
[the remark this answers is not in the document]
chipkent: The default does not persist over bounces.

py/server/deephaven/calendar.py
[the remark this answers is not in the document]
chipkent: Changes to discuss:
* `set_calendar` vs `set_default_calendar`
* `calendar_name` vs `default_calendar_name`

Not interested in changing because the name is more typing and/or not-pythonic for no benefit:
* `calendar_names` vs `list_calendar_names` or `get_calendar_names`
* `calendar` vs `get_calendar`

Generators/src/main/java/io/deephaven/pythonPreambles/plotV2.py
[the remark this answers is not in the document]
chipkent: This is a preamble to generated code.  It is not the generated code.  Just a header.

engine/table/src/test/java/io/deephaven/engine/table/impl/select/FilterKernelArraySample.java
```diff
@@ -66,6 +66,7 @@
 import static io.deephaven.gui.color.Color.*;
 
 import static io.deephaven.time.DateTimeUtils.*;
+import static io.deephaven.time.calendar.Calendars.*;
```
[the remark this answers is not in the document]
chipkent: This is used to validate that generated code matches.  The generator includes the line.

engine/table/src/test/java/io/deephaven/engine/table/impl/select/FilterKernelSample.java
```diff
@@ -66,6 +66,7 @@
 import static io.deephaven.gui.color.Color.*;
 
 import static io.deephaven.time.DateTimeUtils.*;
+import static io.deephaven.time.calendar.Calendars.*;
```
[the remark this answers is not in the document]
chipkent: This is used to validate that generated code matches. The generator includes the line.

engine/table/src/test/java/io/deephaven/engine/table/impl/select/FormulaKernelSample.java
```diff
@@ -66,6 +66,7 @@
 import static io.deephaven.gui.color.Color.*;
 
 import static io.deephaven.time.DateTimeUtils.*;
+import static io.deephaven.time.calendar.Calendars.*;
```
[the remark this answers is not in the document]
chipkent: This is used to validate that generated code matches. The generator includes the line.

engine/table/src/test/java/io/deephaven/engine/table/impl/select/FormulaSample.java
```diff
@@ -66,6 +66,7 @@
 import static io.deephaven.gui.color.Color.*;
 
 import static io.deephaven.time.DateTimeUtils.*;
+import static io.deephaven.time.calendar.Calendars.*;
```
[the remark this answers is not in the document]
chipkent: This is used to validate that generated code matches. The generator includes the line.

engine/time/build.gradle
```diff
@@ -26,3 +26,11 @@ dependencies {
             project(path: ':test-configs')
     Classpaths.inheritSlf4j(project, 'slf4j-simple', 'testRuntimeOnly')
 }
+
+spotless {
```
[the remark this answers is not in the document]
chipkent: Correct.

py/server/deephaven/numpy.py
```diff
@@ -159,3 +161,44 @@ def to_table(np_array: np.ndarray, cols: List[str]) -> Table:
         raise
     except Exception as e:
         raise DHError(e, "failed to create a Deephaven Table from a Pandas DataFrame.") from e
+
+
+def to_busdaycalendar(cal: BusinessCalendar, include_partial: bool = True) -> np.busdaycalendar:
```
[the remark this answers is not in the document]
chipkent: The prefix is the `numpy` module.

py/server/tests/test_calendar.py
```diff
@@ -1,141 +1,81 @@
...
+            add_calendar(cal=cal, file=path1)
+
+        add_calendar(file=path1)
+        self.assertIn("TEST1", calendar_names())
+
+        add_calendar(cal=cal)
+        self.assertIn("TEST2", calendar_names())
 
-        self.assertIn("RuntimeError", cm.exception.root_cause)
+        remove_calendar("TEST1")
```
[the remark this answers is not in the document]
chipkent: This is included here because if it was not the global state would be polluted for other tests.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -39,6 +39,7 @@
 import io.deephaven.libs.GroovyStaticImports;
 import io.deephaven.plugin.type.ObjectTypeLookup;
 import io.deephaven.time.DateTimeUtils;
+import io.deephaven.time.calendar.StaticCalendarMethods;
```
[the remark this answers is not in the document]
chipkent: Good catch.  The wrong class was being imported.

py/server/deephaven/calendar.py
chipkent: I'm hesitant to make this change without discussion.  What if we want to add the ability to add a calendar from a json string or something like that?  Then things get funky.

py/server/tests/test_numpy.py
```diff
@@ -146,6 +148,54 @@ def test_to_table(self):
                 tmp_table3 = to_table(np_array[:, [0, 1, 3]], [col.name for col in tmp_table.columns])
             self.assertIn("doesn't match", cm.exception.root_cause)
 
+    def get_resource_path(self, resource_path) -> str:
+        obj = jpy.get_type("io.deephaven.integrations.python.PythonTimeComponentsTest")()
+        Paths = jpy.get_type("java.nio.file.Paths")
+        Objects = jpy.get_type("java.util.Objects")
+        return Paths.get(Objects.requireNonNull(obj.getClass().getResource(resource_path)).toURI()).toString()
+
+    def test_to_busdaycalendar(self):
```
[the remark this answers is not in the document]
chipkent: I'm not following.

Generators/src/main/java/io/deephaven/plot/util/GenerateFigureImmutable.java
```diff
@@ -91,15 +93,14 @@ private boolean skip(final JavaFunction f) {
...
-                outputClassNameShort,
-                functionNamer.apply(f),
-                f.getTypeParameters(),
-                f.getReturnType(),
-                f.getParameterTypes(),
-                f.getParameterNames(),
-                f.isVarArgs());
+        final Type returnType = new Type() {
+            @Override
+            public String getTypeName() {
```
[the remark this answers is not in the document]
chipkent: `Type` is an interface.  This is creating a concrete implementation that makes everything downstream work.

engine/time/src/main/java/io/deephaven/time/calendar/BusinessPeriod.java
chipkent: It depends on how it is instantiated.  The comment indicates this:
```java
@param <T> time type
```

engine/time/src/main/java/io/deephaven/time/calendar/Calendars.java
chipkent: Updated docs

engine/time/src/main/java/io/deephaven/time/calendar/Calendars.java
chipkent: Updated docs

py/server/deephaven/numpy.py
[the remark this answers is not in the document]
alexpeters1208: Good point, I did not consider this.

py/server/deephaven/calendar.py
jmao-denver: Not sure if we need to worry about the specific case of JSON, since`json.load` returns a dict by default, so an additional dict type in the union would work nicely. IMO what I suggested is more Pythonic but it is just a suggestion.

py/server/tests/test_numpy.py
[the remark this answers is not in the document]
alexpeters1208: Test for cases where `to_busdaycalendar` fails. Looks like the code block right under this comment may be that.

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendarParser.java
```diff
@@ -0,0 +1,187 @@
+/**
+ * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
```
alexpeters1208, line 2:
```suggestion
 * Copyright (c) 2016-2023 Deephaven Data Labs and Patent Pending
```

engine/time/src/main/java/io/deephaven/time/calendar/BusinessDay.java
```diff
@@ -0,0 +1,224 @@
+/**
+ * Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
```
alexpeters1208, line 2:
```suggestion
 * Copyright (c) 2016-2023 Deephaven Data Labs and Patent Pending
```

engine/time/src/main/java/io/deephaven/time/calendar/BusinessDay.java
```diff
@@ -0,0 +1,224 @@
...
+            if (p.contains(time)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
alexpeters1208, line 186: I think this could use a short comment on how equality is being defined for this object.

alexpeters1208 (changes requested):
Very minor comments.

engine/time/src/main/java/io/deephaven/time/calendar/Calendars.java
```diff
@@ -3,109 +3,49 @@
...
-        return instance.keySet().toArray(String[]::new);
-    }
-
-
-
-    private final Map<String, BusinessCalendar> calendars = new HashMap<>();
-
-
-    private Calendars() {
+    static {
```
cpwright, line 38: Deferring complex work that can result in errors after class initialization is helpful for error reporting.  This is going to end up with hard to diagnose class not found errors.

py/server/deephaven/calendar.py
```diff
@@ -1,509 +1,129 @@
...
-             date (str): the date str, format must be "yyyy-MM-dd"
-
-        Returns:
-            a BusinessSchedule instance
-
-        Raises:
-            DHError
-        """
+    if cal is None:
+        raise DHError("cal must be specified")
```
jmao-denver, line 44:
```suggestion
        raise DHError(message="cal must be specified")
```

py/server/deephaven/calendar.py
```diff
@@ -1,509 +1,129 @@
...
-        """ Returns the number of business days between the start and end dates.
+def calendar_names() -> List[str]:
+    """ Returns the names of all available calendar names.
 
-        Args:
-            start (str): the start day of the range
-            end (str): the end day of the range
-            end_inclusive (bool): whether to include the end date, default is False
+    Returns:
+        a list of names of all available calendar names
```
jmao-denver, line 76:
```suggestion
        a list of all available calendar names
```

jmao-denver (commented):
Only minor things

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -418,7 +430,8 @@ private CachedCurrentDate(@NotNull final ZoneId timeZone) {
 
         @Override
         void update(final long currentTimeMillis) {
-            value = formatDate(epochMillisToInstant(currentTimeMillis), timeZone);
+            date = toLocalDate(epochMillisToInstant(currentTimeMillis), timeZone);
+            str = formatDate(date);
             valueExpirationTimeMillis =
                     epochMillis(atMidnight(epochNanosToInstant(millisToNanos(currentTimeMillis) + DAY), timeZone));
```
devinrsmith, line 436: I know this is existing code, but it doesn't look correct to me - adding 24 hours of nanos might not account for leap occurrences. Think this needs to be done wrt `date.plusDays(1)`.

props/configs/src/main/resources/calendar/UTC.calendar
```diff
@@ -1,11 +1,14 @@
...
   -->
 
 <calendar>
     <name>UTC</name>
     <timeZone>UTC</timeZone>
+    <description>UTC</description>
+    <firstValidDate>1500-01-01</firstValidDate>
+    <lastValidDate>2500-12-31</lastValidDate>
     <default>
-        <businessPeriod>00:00,24:00</businessPeriod>
```
devinrsmith, line 9: Was this change necessitated from a parsing issue? IMO, the exact close time should be excluded (in which case, 24:00 is correct).

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
```diff
@@ -1,774 +1,1898 @@
...
-     * @param time time; if null, return null
-     * @param days number of days
-     * @return the non-business date {@code days} non-business days before input {@code time}
+     * @param date date
+     * @return ratio of the business day length and the standard business day length for the date
+     * @throws RequirementFailure if the input is null
+     * @throws InvalidDateException if the date is not in the valid range
      */
-    String previousNonBusinessDay(final Instant time, int days);
+    public double fractionStandardBusinessDay(final LocalDate date) {
```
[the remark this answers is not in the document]
chipkent: Yes.  Very common quant operation.  Wouldn't have added it otherwise.

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendarParser.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * A parser for reading business calendar files.
+ *
+ * Business calendar files should be formatted as:
```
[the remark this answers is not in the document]
chipkent: It doesn't match a CSV format.

engine/time/src/main/java/io/deephaven/time/calendar/BusinessPeriod.java
```diff
@@ -5,56 +5,75 @@
...
+ *
+ * @param <T> time type
  */
-public class BusinessPeriod implements Serializable {
-    private static final long serialVersionUID = 8196837269495115196L;
-    private final Instant startTime;
-    private final Instant endTime;
+public class BusinessPeriod<T extends Comparable<T> & Temporal> {
+    private final T start;
+    private final T end;
```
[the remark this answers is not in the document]
chipkent: It does not always have a time zone.  This is created with `LocalTime` to set the default business day.

engine/table/src/main/java/io/deephaven/libs/GroovyStaticImports.java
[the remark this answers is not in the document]
chipkent: I went through this line by line because I had the same problem.  It resulted from refactoring the code generation.

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendarParser.java
```diff
@@ -0,0 +1,187 @@
...
+ *
+ * <calendar> <name>USNYSE</name> <description>New York Stock Exchange Calendar</description>
+ * <timeZone>America/New_York</timeZone> <default>
+ * <businessPeriod><open>09:30</open><close>16:00</close></businessPeriod> <weekend>Saturday</weekend>
+ * <weekend>Sunday</weekend> </default> <firstValidDate>1999-01-01</firstValidDate>
+ * <lastValidDate>2003-12-31</lastValidDate> <holiday> <date>19990101</date> </holiday> <holiday> <date>20020705</date>
+ * <businessPeriod><open>09:30</open><close>13:00</close></businessPeriod> </holiday> </calendar>
```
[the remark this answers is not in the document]
chipkent: Ugh... Should be fixed.  Forgot the special syntax and then spotless left a mess.

engine/time/src/main/java/io/deephaven/time/calendar/BusinessDay.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    /**
+     * Determines if the specified time is a business time for the day.
+     *
+     * @param time time.
+     * @return true if the time is a business time for the day; otherwise, false.
+     */
+    public boolean isBusinessTime(final T time) {
+        for (BusinessPeriod<T> p : openPeriods) {
```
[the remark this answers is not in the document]
chipkent: Almost all markets have exactly 1 business period.  Weirdo markets like Japan have two.  I don't know of any that have 3 or more.

engine/time/src/main/java/io/deephaven/time/calendar/StaticCalendarMethods.java
```diff
@@ -1,414 +1,435 @@
...
+import io.deephaven.time.calendar.Calendar;
+import java.lang.String;
 import java.time.DayOfWeek;
 import java.time.Instant;
 import java.time.LocalDate;
+import java.time.LocalTime;
 import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.util.Map;
+import java.util.Set;
```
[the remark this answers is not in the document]
chipkent: These are from generated code.  The generator is making certain that input/output types are included in case they are used in the function bodies.  If you find it particularly objectionable, I can remove it, but it isn't hand written code.

engine/time/src/main/java/io/deephaven/time/calendar/BusinessDay.java
chipkent: This is equality so that the object works in java collections.  The current code is more clear, but it is just checking for identical contents.  Very standard, so it is using the default javadoc via `@Override`.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -1403,6 +1449,51 @@ public static ZonedDateTime excelToZonedDateTime(final double excel, @Nullable f
...
+     *
+     * @param date starting date
+     * @param days number of days to add
+     * @return {@code null} if either input is {@code null} or {@link QueryConstants#NULL_LONG}; otherwise the starting
+     *         date plus the specified number of days
+     * @throws DateTimeOverflowException if the resultant date time exceeds the supported range
+     */
+    @ScriptApi
+    @Nullable
+    public static LocalDate plus(@Nullable final LocalDate date, final long days) {
```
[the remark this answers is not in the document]
chipkent: The generic `plus` method allows the query language to support things like `Date+1`, which seems desirable.  Do you still object?

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
```diff
@@ -1,774 +1,1898 @@
...
- * A business calendar. Calendar is extended with the concept of business and non-business time.
- *
- * To comply with the ISO-8601 standard for dates, Strings should be of the form "yyyy-MM-dd",
+ * A business calendar, with the concept of business and non-business time.
+ * <p>
+ * Date strings must be in a format that can be parsed by {@code DateTimeUtils#parseDate}. Methods that accept strings
+ * can be slower than methods written explicitly for {@code Instant}, {@code ZonedDateTime}, or {@code LocalDate}.
  */
-public interface BusinessCalendar extends Calendar {
+public class BusinessCalendar extends Calendar {
```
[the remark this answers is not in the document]
chipkent: It was originally an interface, but it seemed Rube Goldberg for this case since I don't think there is an expectation that others create business calendars that are not based on `BusinessCalendar`.  Do you still want an interface?

engine/time/src/main/java/io/deephaven/time/calendar/Calendars.java
chipkent: Would you prefer all methods to have lazy init class as opposed to the static init?

props/configs/src/main/resources/calendar/UTC.calendar
chipkent: The exact problem is a bit complex, and I keep waffling about the right solution.  I'm not yet convinced that the existing solution isn't right.  This is related to the open vs closed interval question.

Consider the case of the NYSE closing auction.  These trades occur and post at exactly 4PM.  If the interval is open, these trades are considered to be after business time.  If the interval is closed, they occur during business time.  I'm inclined to think that the closed interval is the correct interpretation, but I'm discussing this with @pete-petey and @rbasralian.  

Additionally:
1) The max value of `LocalTime` is `23:59:59.999999999`.  https://docs.oracle.com/javase/8/docs/api/java/time/LocalTime.html
2) Our `DateTimeUtils.parseLocalTime` is permissive and parses `24:00` as `00:00`.  Since it is a local time, the fact that `24:00` intends to be the next day is lost.
3) Because of the above, `00:00,24:00` was being parsed as a standard business day of `00:00,00:00`, which is clearly not what is desired.

It probably makes sense to reject `24:00` in the XML parser if the closed interval is the winner.  It may also make sense to reject `24:00` in  `DateTimeUtils.parseLocalTime`.

engine/time/src/main/java/io/deephaven/time/calendar/BusinessPeriod.java
```diff
@@ -5,56 +5,75 @@
...
     /**
-     * Returns the length of the period in nanoseconds.
+     * Length of the period in nanoseconds.
      *
      * @return length of the period in nanoseconds
      */
-    public long getLength() {
-        return DateTimeUtils.minus(endTime, startTime);
+    public long nanos() {
+        return nanos;
```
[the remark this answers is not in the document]
chipkent: I've added a `duration()` method which can return more value types and is idiomatic to the new Java time world.  Also now compute nanos on demand.

props/configs/src/main/resources/calendar/UTC.calendar
chipkent: Both @pete-petey and @rbasralian think the interval should be closed.  Changes have been made to make it easier to support open intervals in the future if anyone cares.  Also a proper error is raised on `24:00` since it is on the next day.

engine/time/src/main/java/io/deephaven/time/calendar/BusinessDay.java
```diff
@@ -0,0 +1,224 @@
...
+        return businessNanos;
+    }
+
+    /**
+     * Amount of business time in nanoseconds that has elapsed on the given day by the specified time.
+     *
+     * @param time time
+     * @return business time in nanoseconds that has elapsed on the given day by the specified time
+     */
+    public long businessNanosElapsed(final T time) {
```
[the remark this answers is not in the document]
chipkent: They now return `Duration`s, which should give the same result.

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
```diff
@@ -1,774 +1,1898 @@
...
-     * @param date date; if null, return null
-     * @return the next business day after {@code date}
+     * @param time time
+     * @return the fraction of the business day complete, or 0.0 if the day is not a business day
+     * @throws RequirementFailure if the input is null
+     * @throws InvalidDateException if the date is not in the valid range
      */
-    String nextBusinessDay(String date);
+    public double fractionBusinessDayRemaining(final Instant time) {
+        Require.neqNull(time, "time");
```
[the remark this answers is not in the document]
chipkent: Fixed

engine/time/src/main/java/io/deephaven/time/calendar/Calendars.java
chipkent: Deferring.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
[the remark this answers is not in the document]
chipkent: At a very minimum, `plus(LocalDate, Period)` should exist.

py/server/deephaven/calendar.py
```diff
@@ -1,509 +1,129 @@
...
-        """ Returns if the specified date is the last business day of the month.
 
-        Args:
-            date (str): the date
+def calendar(name: Optional[str] = None) -> BusinessCalendar:
+    """ Returns the calendar with the given name.
 
-        Returns:
-            bool
+    The returned calendar is a 'io.deephaven.time.calendar.BusinessCalendar' Java object that can be used in Python.
```
jjbrosnan, line 107:
```suggestion
    The returned calendar is an 'io.deephaven.time.calendar.BusinessCalendar' Java object that can be used in Python.
```

jjbrosnan (commented):
One extremely minor comment. LGTM.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -1541,6 +1647,51 @@ public static ZonedDateTime plus(@Nullable final ZonedDateTime dateTime, @Nullab
...
+     *
+     * @param date starting date
+     * @param days number of days to subtract
+     * @return {@code null} if either input is {@code null} or {@link QueryConstants#NULL_LONG}; otherwise the starting
+     *         date plus the specified number of days
+     * @throws DateTimeOverflowException if the resultant date time exceeds the supported range
+     */
+    @ScriptApi
+    @Nullable
+    public static LocalDate minus(@Nullable final LocalDate date, final long days) {
```
devinrsmith, line 1661: Did we want this to be `minusDays`?

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -2600,11 +2751,27 @@ public static int hourOfDay(@Nullable final ZonedDateTime dateTime) {
...
 
+    /**
+     * Returns a 1-based int value of the day of the week for a {@link ZonedDateTime} in the specified time zone, with 1
+     * being Monday and 7 being Sunday.
+     *
+     * @param date date to find the day of the week of
+     * @return {@link QueryConstants#NULL_INT} if either input is {@code null}; otherwise, the day of the week
+     */
+    @ScriptApi
+    public static int dayOfWeek(@Nullable final LocalDate date) {
```
devinrsmith, line 2762: Should we have `DayOfWeek dayOfWeek(...)` and `int dayOfWeekValue(...)`?

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -2699,11 +2898,27 @@ public static int dayOfYear(@Nullable final ZonedDateTime dateTime) {
...
 
+    /**
+     * Returns a 1-based int value of the month of the year (Julian date) for a {@link LocalDate}. January is 1,
+     * February is 2, etc.
+     *
+     * @param date date to find the month of the year of
+     * @return {@link QueryConstants#NULL_INT} if either input is {@code null}; otherwise, the month of the year
+     */
+    @ScriptApi
+    public static int monthOfYear(@Nullable final LocalDate date) {
```
devinrsmith, line 2909: The other methods have lined up well w/ the java time API - should this be `monthValue` to also line up?

engine/time/src/main/java/io/deephaven/time/calendar/CalendarDay.java
```diff
@@ -0,0 +1,271 @@
...
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.Temporal;
+import java.util.*;
+
+import static io.deephaven.util.QueryConstants.NULL_LONG;
+
+/**
+ * Schedule for a single calendar day. The schedule contains a list of business time ranges, which are the ranges of
+ * time during which businesses are open.
+ *
```
devinrsmith, line 20: `<p>`

props/configs/src/main/resources/calendar/UTC.calendar
```diff
@@ -1,11 +1,14 @@
...
-  ~ Copyright (c) 2016-2021 Deephaven Data Labs and Patent Pending
+  ~ Copyright (c) 2016-2023 Deephaven Data Labs and Patent Pending
   -->
 
 <calendar>
     <name>UTC</name>
     <timeZone>UTC</timeZone>
+    <description>UTC</description>
+    <firstValidDate>1500-01-01</firstValidDate>
+    <lastValidDate>2500-12-31</lastValidDate>
```
devinrsmith, line 10: Is this loaded by default? Assuming it will be map w/ ~365,000 elements?

props/configs/src/main/resources/calendar/USNYSE.calendar
devinrsmith, line 1: How was this sourced? What's our process for maintaining? Is this something we want deephaven-core to be responsible for?

props/configs/src/main/resources/calendar/USBANK.calendar
devinrsmith, line 1: How was this sourced? What's our process for maintaining? Is this something we want deephaven-core to be responsible for?

engine/time/src/main/java/io/deephaven/time/calendar/CalendarDay.java
devinrsmith: And lots of other places in this file.

engine/time/src/main/java/io/deephaven/time/calendar/Calendar.java
```diff
@@ -1,280 +1,643 @@
...
- * a different time zone is considered a different calendar.
- *
- *
- * Frequently, the default implementation for methods on Instants is to call the corresponding method on a String with
- * {@code DateTimeUtils.formatDate}. This can be slower than methods written explicitly for DateTimes. If performance is
- * an issue, consider overriding these methods with other behavior.
+ * <p>
+ * A calendar is associated with a specific time zone.
+ * <p>
+ * Date strings must be in a format that can be parsed by {@code DateTimeUtils#parseDate}. Methods that accept strings
```
devinrsmith, line 25: `{@link}` instead of `{@code}`?

engine/time/src/main/java/io/deephaven/time/calendar/Calendar.java
```diff
@@ -1,280 +1,643 @@
...
+
+    // region Now
+
+    /**
+     * The current date for the calendar.
+     *
+     * @return current date
+     */
+    public LocalDate calendarDate() {
+        return DateTimeUtils.todayLocalDate(timeZone());
```
devinrsmith, line 100: It's important context, but we are depending on the `clock` state in `DateTimeUtils`. I'm not the biggest fan of the static state in `DateTimeUtils` in-and-of itself, but I wonder if we should be persisting that stateful dependency here? Should `clock` be a member of `Calendar`?

engine/time/src/main/java/io/deephaven/time/calendar/Calendar.java
```diff
@@ -1,280 +1,643 @@
...
+        return plusDays(DateTimeUtils.parseLocalDate(date), days);
     }
 
     /**
-     * Gets the previous date.
+     * Adds a specified number of days to an input time. Adding negative days is equivalent to subtracting days.
      *
-     * @param time time; if null, return null
-     * @return the day before {@code time}
+     * Day additions are not always 24 hours. The resultant time will have the same local time as the input time, as
```
devinrsmith, line 213: Missing `<p>` here and other places in this file.

engine/time/src/main/java/io/deephaven/time/calendar/Calendar.java
```diff
@@ -1,280 +1,643 @@
...
+
+    /**
+     * Creates a new calendar.
+     *
+     * @param name calendar name.
+     * @param description calendar description.
+     * @param timeZone calendar time zone.
+     * @throws RequirementFailure if any parameter is {@code null}
+     */
+    public Calendar(final String name, final String description, final ZoneId timeZone) {
```
devinrsmith, line 44: Is this meant to be user-visible? I see it's only used by test and BusinessCalendar right now. If we need it to be user visible, we should move it to a static method; regardless, we should make this package-private (to support theoretical interface refactor in the future if necessary).

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
```diff
@@ -1,774 +1,2164 @@
...
+     * @param timeZone calendar time zone.
+     * @param firstValidDate first valid date for the business calendar.
+     * @param lastValidDate last valid date for the business calendar.
+     * @param standardBusinessDay business day schedule for a standard business day
+     * @param weekendDays weekend days
+     * @param holidays holidays. Business day schedules for all holidays. A holiday is a date that has a schedule that
+     *        is different from the schedule for a standard business day or weekend.
+     * @throws RequirementFailure if any argument is null.
+     */
+    public BusinessCalendar(final String name, final String description, final ZoneId timeZone,
```
devinrsmith, line 150: Is this meant to be user-visible? If we need it to be user visible, we should move it to a static method; regardless, we should make the constructor package-private (to support theoretical interface refactor in the future if necessary).

props/configs/src/main/resources/calendar/USNYSE.calendar
chipkent: Caroline and Elijah got the values from NYSE.  In earlier discussions, we decided to support exactly two calendars: USNYSE and USBANK.  Both are the most basic standards, and the values are very unambiguous.  We do need an EOY event to update the two calendars every year.

props/configs/src/main/resources/calendar/USBANK.calendar
chipkent: Caroline and Elijah got the values from the FED.  In earlier discussions, we decided to support exactly two calendars: USNYSE and USBANK.  Both are the most basic standards, and the values are very unambiguous.  We do need an EOY event to update the two calendars every year.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
chipkent: I do like this change, but it is breaking.  I'll poll for concerns.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
chipkent: I'm not inclined to make this change.  The current naming is intuitive and consistent with other functions like `dayOfMonth`.

alexpeters1208 (approved):

props/configs/src/main/resources/calendar/UTC.calendar
chipkent: It will get loaded if someone uses any calendars.  Good point on the 365k elements.  I have reduced the date range to 73k. 
```
    <firstValidDate>1900-01-01</firstValidDate>
    <lastValidDate>2100-12-31</lastValidDate>
```
There are tradeoffs if we go to a fully lazy load, so let's see if this change is sufficient.

engine/time/src/main/java/io/deephaven/time/calendar/Calendar.java
chipkent: I'm not fully following this set of comments.

`DateTimeUtils` is the foundational layer.  `Calendar` is built on top of `DateTimeUtils`.  `BusinessCalendar` is built on top of `Calendar`.  If `Calendar` has its own clock, separate from `DateTimeUtils`, it is just going to lead to user nightmare scenarios since we can't tell the user to just change the clock in a single place.  Instead, they will end up with an inconsistent view of the world where they need to know multiple places they need to update to achieve consistency.

engine/time/src/main/java/io/deephaven/time/calendar/Calendar.java
```diff
@@ -22,8 +19,9 @@
  * <p>
  * A calendar is associated with a specific time zone.
  * <p>
- * Date strings must be in a format that can be parsed by {@code DateTimeUtils#parseDate}. Methods that accept strings
- * can be slower than methods written explicitly for {@code Instant}, {@code ZonedDateTime}, or {@code LocalDate}.
+ * Date strings must be in a format that can be parsed by {@link DateTimeUtils#parseLocalDate(String)}. Methods that
+ * accept strings can be slower than methods written explicitly for {@code Instant}, {@code ZonedDateTime}, or
+ * {@code LocalDate}.
```
devinrsmith, line 24: Can be `{@link}` too.

props/configs/src/main/resources/calendar/USNYSE.calendar
chipkent: Changed to USNYSE_EXAMPLE and USBANK_EXAMPLE.

props/configs/src/main/resources/calendar/USBANK.calendar
chipkent: Changed to USNYSE_EXAMPLE and USBANK_EXAMPLE.

alexpeters1208 (approved):

jjbrosnan (approved):

devinrsmith (approved):

jmao-denver (approved):
The Python changes LGTM.

deephaven-internal:

Community: https://github.com/deephaven/deephaven.io/issues/3610

PR #4221, opened 2023-07-23
Add sort_descending in the Python client Table class

### LINKED ISSUE #4083: sort_descending is not available from Python Client
**Description**

There is no sort_descending method on Table.

**Steps to reproduce**

session.run_script("from deephaven import empty_table\nx=empty_table(10).update([\"A=1\"])")
sorted = session.open_table("x").sort("A")

**Expected results**

The table is sorted descending by A.

**Actual results**

I get an error message saying sort_descending is not found.

**Versions**
 - Deephaven: 0.25.3

jmao-denver:
Fixes #4083

### DIFF
--- a/py/client/pydeephaven/table.py
+++ b/py/client/pydeephaven/table.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 from typing import List, Union
 
 import pyarrow as pa
+from pydeephaven._utils import to_list
 
 from pydeephaven._table_ops import MetaTableOp, SortDirection
 from pydeephaven.agg import Aggregation
@@ -217,6 +218,23 @@ class Table(TableInterface):
         Raises:
             DHError
         """
+        return super().sort(order_by, order)
+
+    def sort_descending(self, order_by: Union[str, List[str]]) -> Table:
+        """The sort_descending method creates a new table where rows in a table are sorted in a largest to smallest
+        order based on the order_by column(s).
+
+        Args:
+            order_by (Union[str, List[str]]): the column name(s)
+
+        Returns:
+            a Table object
+
+        Raises:
+            DHError
+        """
+        order_by = to_list(order_by)
+        order = [SortDirection.DESCENDING] * len(order_by)
         return super().sort(order_by, order)
 
     def where(self, filters: Union[str, List[str]]) -> Table:
@@ -729,7 +747,7 @@ class Table(TableInterface):
         Raises:
             DHError
         """
-        return  super(Table, self).where_not_in(filter_table, cols)
+        return super(Table, self).where_not_in(filter_table, cols)
 
 
 class InputTable(Table):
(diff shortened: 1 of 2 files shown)


py/client/pydeephaven/table.py
```diff
@@ -219,6 +220,23 @@ def sort(self, order_by: Union[str, List[str]], order: Union[SortDirection, List
         """
         return super().sort(order_by, order)
 
+    def sort_descending(self, order_by: Union[str, List[str]]) -> Table:
+        """The sort_descending method creates a new table where rows in a table are sorted in a largest to smallest
```
rcaudy, line 224: Is largest to smallest the right way to phrase this? Descending order, or "last to first" or something might be better.

chipkent:
```suggestion
        """The sort_descending method creates a new table where rows in a table are sorted in descending
```
I think this is most consistent with our `sort` docs.

py/client/pydeephaven/table.py
```diff
@@ -219,6 +220,23 @@ def sort(self, order_by: Union[str, List[str]], order: Union[SortDirection, List
...
+            order_by (Union[str, List[str]]): the column name(s)
+
+        Returns:
+            a Table object
+
+        Raises:
+            DHError
+        """
+        order_by = to_list(order_by)
+        order = [SortDirection.DESCENDING] * len(order_by)
```
chipkent, line 237:
```suggestion
        order = SortDirection.DESCENDING
```
I think this should be sufficient and is more simple.

py/client/pydeephaven/table.py
```diff
@@ -219,6 +220,23 @@ def sort(self, order_by: Union[str, List[str]], order: Union[SortDirection, List
         """
         return super().sort(order_by, order)
 
+    def sort_descending(self, order_by: Union[str, List[str]]) -> Table:
```
chipkent, line 223: * Should there be a `sort_ascending`?
* Is this needed if a user can simply do `sort(order_by, DESCENDING)`?
* If this method is here, should it also be on the proxy?

chipkent (changes requested):

py/client/pydeephaven/table.py
jmao-denver: - It does seem odd to have it, but I think only someone who remembers its history can explain its existence and the need to keep it for backwards compatibility.
- see above
- not sure if we want to pollute the API when we have doubts about its very existence.

chipkent: I see https://github.com/deephaven/deephaven-core/issues/4083.  The user wants the client API to match the server API, which is reasonable.  Looking at our server API, we only have `sort_descending`.  `sort_ascending` is accomplished via `sort` with default arguments.

Looking some more, the server has this method for:
* PartitionedTableProxy
* RollupNodeOperationsRecorder
* Table
* TreeNodeOperationsRecorder

Looking at the client, we just have Table, so it appears that this is the only needed change.

All that to say that I think the needed work is done.

chipkent (approved):

PR #4240, opened 2023-07-28
R Client 2.0 - Table Ops

### LINKED ISSUE #256: Re-implement R Deephaven client over Open API
irisDB.r has been deleted, as it relied heavily on IrisIntegrationGroovySession which in turn relied on the legacy Java API. A new deephaven R connector should be created, ideally as a high-level R Deephaven API.

### LINKED ISSUE #4090: Implement generic methods in the R Client
While the R client currently has methods for TableHandles like `tableHandle$nrow()` and `tableHandle$to_data_frame()`, we should also support idiomatic R generics like `nrow(tableHandle)` and `as.data.frame(tableHandle)`. These should be easily to implement as outlined [here](https://stackoverflow.com/questions/59636438/how-to-dispatch-summary-method-in-r6-object), but they've proven to be not-so-easy.

alexpeters1208:
This PR introduces new table operations to the R client, presented in a `dplyr` like syntax that will feel familiar to R users. The best places to see the comparison between the new table operations and what "native R" would look and feel like is in the `inst/tests/testthat/test_table_ops.R` and `inst/tests/testthat/test_agg_by.R` files. These unit tests are approached as solving the same problem in `dplyr` with `tibble`s and the new R client table operations, ensuring correctness while also providing a kind of translation manual.

The documentation of the `table_ops.R` is not hooked in yet, and there is discussion to be had over the names of some of our operations, as they are consistent with the other Deephaven clients but conflict with function names that already exist in R. This is an easily solvable problem, but it is worth discussing whether we want to mask common functions.

This PR also adds a `close` method to the client, and it is encouraged to call this method when the client connection is no longer needed.

Resolves #256 
Resolves #4090


chipkent:
The ticket description is mentioning `dplyr` syntax, which I think we discussed is not a target.
Additionally, all of these changes should have been separate PRs for the sake of review.

R/rdeephaven/R/table_ops.R
chipkent, line 1: 1. The current overall design has some significant inconsistencies for table handles.  For example, `select` can be used via the pipe notation but not via the `$` notation.  Conversely, `is_static` can be used via the `$` notation but not via the pipe notation.

To cleanly resolve these inconsistencies and provide the most flexibility to the user, I think that all methods should be implemented on the table handle _AND_ all table handle methods should be implemented functionally for use in pipes.  This bulk of the logic in this class will move to the table handle, leaving a bunch of one-line functions.

2. We should remove all reference to `dplyr-like` or `dplyr` syntax.  This is all just pipe syntax that can be implemented using various pipe tooling by the user.  It is not locked into the dplyr `%>%` syntax.

chipkent (changes requested):

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -25,15 +25,17 @@
 #' # push new data frame to the server and name it "new_table"
 #' new_table_handle2 <- client$import_table(new_data_frame)
 #' new_table_handle2$bind_to_variable("new_table")
+
```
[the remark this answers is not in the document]
alexpeters1208: Yes.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -95,10 +97,53 @@ TableHandle <- R6Class("TableHandle",
     to_data_frame = function() {
       arrow_tbl <- self$to_arrow_table()
       return(as.data.frame(as.data.frame(arrow_tbl))) # TODO: for some reason as.data.frame on arrow table returns a tibble, not a data frame
-    }
+    },
+    internal_table_handle = NULL
```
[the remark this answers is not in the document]
alexpeters1208: It must be public if any function (including our own) needs to access the internal thing. Otherwise it can be private. This is an upside to your proposed refactoring of the table ops.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -95,10 +97,53 @@ TableHandle <- R6Class("TableHandle",
...
   ),
   private = list(
-    internal_table_handle = NULL,
     is_static_field = NULL
   )
 )
+
+# EXISTING GENERICS THAT WE CAN SUPPORT
```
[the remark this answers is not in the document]
alexpeters1208: Because these are generics, they already have docstrings that tell you what they should do. There may be a `?as.data.frame.TableHandle` to get the docstrings specifically for this implemented generic, but I'm not sure.

R/rdeephaven/inst/tests/testthat/test_table_handle_wrapper.R
```diff
@@ -52,8 +55,9 @@ test_that("is_static returns the correct value", {
   expect_true(data$th2$is_static())
   expect_true(data$th3$is_static())
   expect_true(data$th4$is_static())
+  expect_false(data$th5$is_static())
```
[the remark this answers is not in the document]
alexpeters1208: I do not understand this, lines 54-57 verify that `is_static()` confirms static tables are static

R/rdeephaven/src/client.cpp
```diff
@@ -14,17 +15,257 @@
...
+        return new TableHandleWrapper(internal_tbl_hdl.updateView(columnSpecs));
+    };
+
+    TableHandleWrapper* dropColumns(std::vector<std::string> columnSpecs) {
+        return new TableHandleWrapper(internal_tbl_hdl.dropColumns(columnSpecs));
+    };
+
+    TableHandleWrapper* where(std::string condition) {
+        return new TableHandleWrapper(internal_tbl_hdl.where(condition));
+    };
```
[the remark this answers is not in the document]
alexpeters1208: Underlying C++ currently does not support multiple conditions. At least not as of ~1 week ago.

R/rdeephaven/src/client.cpp
```diff
@@ -14,17 +15,257 @@
...
+
+    TableHandleWrapper* groupBy(std::vector<std::string> columnSpecs) {
+        return new TableHandleWrapper(internal_tbl_hdl.by(columnSpecs));
```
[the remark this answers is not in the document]
alexpeters1208: Should I communicate this to Corey?

R/rdeephaven/src/client.cpp
```diff
@@ -14,17 +15,257 @@
...
+    };
+
+    TableHandleWrapper* ungroup(std::vector<std::string> groupByColumns) {
+        return new TableHandleWrapper(internal_tbl_hdl.ungroup(false, groupByColumns));
+    };
+
+    // AGGREGATION OPERATIONS
+
+    TableHandleWrapper* aggBy(Rcpp::List aggregations, std::vector<std::string> groupByColumns) {
+        std::vector<deephaven::client::Aggregate> converted_aggregations = convertRcppListToVectorOfTypeAggregate(aggregations);
```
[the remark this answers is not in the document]
alexpeters1208: Not a fan, I like to see everything that's happening in maximum detail because I don't understand this well enough for hiding explicit types to be helpful

R/rdeephaven/src/client.cpp
```diff
@@ -14,17 +15,257 @@
...
+        sort_pairs.reserve(columnSpecs.size());
+        if (descending.size() == 1) {
+            if (descending[0] == false) {
+                for(int i = 0; i < columnSpecs.size(); i++) {
+                    sort_pairs.push_back(deephaven::client::SortPair::ascending(columnSpecs[i], false));
+                }
+            } else {
+                for(int i = 0; i < columnSpecs.size(); i++) {
+                    sort_pairs.push_back(deephaven::client::SortPair::descending(columnSpecs[i], false));
+                }
```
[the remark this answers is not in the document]
alexpeters1208: I would expect an R user to accomplish this by applying abs() to a column and then sorting. Would not hurt to implement this, but I don't think it's the natural way to do this as far as R people are concerned.

R/rdeephaven/src/client.cpp
```diff
@@ -208,8 +481,20 @@ class ClientWrapper {
...
+    * after Client's destructor is invoked. If the caller tries to do so, the behavior is
+    * unspecified.
+    */
+    void close() {
+        internal_client.close();
+    }
+
 private:
-    const deephaven::client::Client internal_client;
+    deephaven::client::Client internal_client;
```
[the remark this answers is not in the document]
alexpeters1208: Because of this:
```
client.cpp: In member function ‘void ClientWrapper::close()’:
client.cpp:454:31: error: passing ‘const deephaven::client::Client’ as ‘this’ argument discards qualifiers [-fpermissive]
  454 |         internal_client.close();
      |                               ^
In file included from client.cpp:13:
/home/alexpeters/deephaven/dhcpp/local/deephaven/include/deephaven/client/client.h:208:8: note:   in call to ‘void deephaven::client::Client::close()’
  208 |   void close();
      |        ^~~~~
```

Probably I just don't know C++.

R/rdeephaven/R/helper_functions.R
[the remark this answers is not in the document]
alexpeters1208: No docstrings yet, will add docstrings when the rest of the documentation happens.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+      } else if (auth_type == "custom") {
+        if (auth_token_pair != "") {
+          verify_string("auth_token_pair", auth_token_pair, TRUE)
+          key_value <- strsplit(auth_token_pair, ":", fixed = TRUE)
+          options$set_custom_authentication(key_value[1], key_value[2])
+        } else {
+          stop("Custom authentication was requested, but no 'auth_token_pair' was provided.")
+        }
+      } else {
+        stop(paste0("'auth_type' must be 'anonymous', 'basic', or 'custom', but got ", auth_type, " instead."))
```
chipkent, line 51: This is awkward because it is a level deeper than it should be.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+    # set session type if a valid session type is provided
+    if ((session_type == "python") || (session_type == "groovy")) {
+      options$set_session_type(session_type)
+    } else {
+      stop(paste0("'session_type' must be 'python' or 'groovy', but got ", session_type, " instead."))
+    }
+
+    # if tls is requested, set it and set the root_certs if provided
+    if (use_tls != FALSE) {
+      if (use_tls == TRUE) {
```
chipkent, line 64: ???.  

If you are concerned about non-boolean stuff, you should have an assertion at the start of the function to avoid this.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+  function(target,
+           auth_type = "anonymous",
+           auth_token_pair = "",
+           session_type = "python",
+           use_tls = FALSE,
+           tls_root_certs = "",
+           int_option = "",
+           string_option = "",
+           extra_header = "") {
+    options <- new(INTERNAL_ClientOptions)
```
chipkent, line 30: assert inputs?

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
-    },
-    df_to_dh_table = function(data_frame) {
-      arrow_tbl <- arrow_table(data_frame)
-      return(private$arrow_to_dh_table(arrow_tbl))
+
+    if (int_option != "") {
+      verify_string("int_option", int_option, TRUE)
+      new_int_option <- strsplit(int_option, ":", fixed = TRUE)
+      options$add_int_option(new_int_option[1], as.numeric(new_int_option[2]))
     }
```
chipkent, line 79: This looks wacky:
1) doing weird string handling stuff again
2) this is handling integer options as strings
3) this looks like multiple options can be provided, but this R API is not supporting it.  Why not a map containing the options?

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
-  )
+
+    if (string_option != "") {
+      verify_string("string_option", string_option, TRUE)
+      new_string_option <- strsplit(string_option, ":", fixed = TRUE)
+      options$add_string_option(new_string_option[1], new_string_option[2])
+    }
```
chipkent, line 85: See comments on int options.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+
+    if (extra_header != "") {
+      verify_string("extra_header", extra_header, TRUE)
+      new_extra_header <- strsplit(extra_header, ":", fixed = TRUE)
+      options$add_extra_header(new_extra_header[1], new_extra_header[2])
+    }
```
chipkent, line 91: See comments on int options.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+  "connect",
+  function(target, ...) {
+    return(standardGeneric("connect"))
+  },
+  signature = c("target")
+)
+
+#' @export
+setMethod(
+  "connect",
```
chipkent, line 19: Thoughts on such a vague name.  Would we be better with `dhconnect` since the signature is not that insightful?

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+  function(client_instance, name) {
+    return(standardGeneric("open_table"))
+  },
+  signature = c("client_instance", "name")
+)
+
+#' @export
+setMethod(
+  "open_table",
+  signature = c(client_instance = "Client"),
```
chipkent, line 142: This comment applies everywhere.

The signature provided in the methods should be as specific as possible to avoid crappy dispatch.  In this case, the function takes a `Client` and a string.  As such, the signature should indicate both.  Just defaulting to single dispatch is bad practice and is the whole reason S4 was created.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+  function(client_instance, ...) {
+    return(standardGeneric("time_table"))
+  },
+  signature = "client_instance"
+)
+
+#' @export
+setMethod(
+  "time_table",
+  signature = c(client_instance = "Client"),
```
chipkent, line 181: Not sure what the dispatch signature would be in this optional arg case. 
... how will this evolve as strings are possible.  Another dispatch question.

chipkent (changes requested):

R/rdeephaven/inst/tests/testthat/test_agg_by_magrittr_pipe.R
chipkent, line 1: 1) This looks like a duplicate of the base pipe test with the pipes changed out.  Can we not have all of the tests together so that it is obvious where things need changing?  I'm concerned that splitting things this way makes it very hard to maintain.

2) Duplicating every test for both kinds of pipes is probably overkill.  If we have a single test of sufficient complexity that does it both ways, that is probably sufficient.  This is basically a syntactic extension to the language for calling functions, so I would expect everything to work or everything to fail.

chipkent (changes requested):

R/rdeephaven/R/helper_functions.R
```diff
@@ -1,18 +1,86 @@
...
-  if (class(string_candidate)[[1]] != "character") {
-    stop(paste0("'", arg_name, "' must be passed as a single string. Got an object of class ", class(string_candidate)[[1]], " instead."))
-  } else if (length(string_candidate) != 1) {
-    stop(paste0("'", arg_name, "' must be passed as a single string. Got a character vector of length ", length(string_candidate), " instead."))
+first_class <- function(arg) {
+  return(class(arg)[[1]])
+}
+
+verify_type <- function(arg_name, candidate, required_type, message_type_name, is_scalar) {
+  if (first_class(candidate) == "list") {
```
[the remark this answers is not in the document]
alexpeters1208: I moved the `if(is_scalar)` on line 20 up to the top, so we check for scalar first, and do not try to evaluate the lists if that fails. I think this is what you mean, not totally sure.

R/rdeephaven/R/helper_functions.R
```diff
@@ -1,18 +1,86 @@
...
+
+  if (is_scalar) {
+    if (length(candidate) != 1) {
+      stop(paste0("'", arg_name, "' must be passed as a single ", message_type_name, ". Got a ", message_type_name, " vector of length ", length(candidate), " instead."))
+    }
+  }
+}
+
+# does not attempt to verify that candidate is numeric
+verify_in_range <- function(arg_name, candidate, message, a = NULL, b = NULL, a_open = TRUE, b_open = TRUE) {
```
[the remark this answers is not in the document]
alexpeters1208: I had that in here in a previous commit, and got rid of it because I thought the error messages were cleaner with `nonnegative` and `positive` instead of `[0, inf)` and `(0, inf)`, since this function is currently only used in `verify_nonnegative_int` and `verify_positive_int`.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+  )
+)
+
+setGeneric(
+  "connect",
+  function(target, ...) {
+    return(standardGeneric("connect"))
+  },
+  signature = c("target")
+)
```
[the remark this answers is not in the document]
alexpeters1208: I believe this only works if `connect` is an _existing_ generic that we want to implement:
```
Error in setGeneric("connect") : 
  must supply a function skeleton for ‘connect’, explicitly or via an existing function
Error: unable to load R code in package ‘rdeephaven’
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+           extra_header = "") {
+    options <- new(INTERNAL_ClientOptions)
+
+    # check if auth_type needs to be changed and set credentials accordingly
+    if (auth_type != "anonymous") {
+      if (auth_type == "basic") {
+        if (auth_token_pair != "") {
+          verify_string("auth_token_pair", auth_token_pair, TRUE)
+          username_password <- strsplit(auth_token_pair, ":", fixed = TRUE)
+          options$set_basic_authentication(username_password[1], username_password[2])
```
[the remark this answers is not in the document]
alexpeters1208: 1. Yes (https://github.com/deephaven/deephaven-core/blob/f59b7246ee30729d34300cd9f31d53294f1452b5/cpp-client/deephaven/client/src/client_options.cc#L27)
2. Fair concern. I'm following Python (https://github.com/deephaven/deephaven-core/blob/f59b7246ee30729d34300cd9f31d53294f1452b5/py/client/pydeephaven/session.py#L112C11-L112C11), but am happy to not do so.
3. Not a helpful error

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+        } else {
+          stop("Basic authentication was requested, but no 'auth_token_pair' was provided.")
+        }
+      } else if (auth_type == "custom") {
+        if (auth_token_pair != "") {
+          verify_string("auth_token_pair", auth_token_pair, TRUE)
+          key_value <- strsplit(auth_token_pair, ":", fixed = TRUE)
+          options$set_custom_authentication(key_value[1], key_value[2])
```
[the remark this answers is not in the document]
alexpeters1208: Probably yes. Should also do this in Python.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+           session_type = "python",
+           use_tls = FALSE,
+           tls_root_certs = "",
+           int_option = "",
+           string_option = "",
+           extra_header = "") {
+    options <- new(INTERNAL_ClientOptions)
+
+    # check if auth_type needs to be changed and set credentials accordingly
+    if (auth_type != "anonymous") {
```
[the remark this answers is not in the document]
alexpeters1208: I have it this way because using "anonymous" doesn't do anything to the internal object, so the "anonymous" branch would be empty.

R/rdeephaven/R/client_wrapper.R
alexpeters1208: Added assertions for `target` and `use_tls` at function beginning. Everything else is asserted if it's non-default.

R/rdeephaven/R/client_wrapper.R
alexpeters1208: 1. Following Python
2. Yes, weird
3. Good catch, I forgot about that

R/rdeephaven/R/client_wrapper.R
alexpeters1208: I like `dhConnect`.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+  },
+  signature = c("target")
+)
+
+#' @export
+setMethod(
+  "connect",
+  signature = c(target = "character"),
+  function(target,
+           auth_type = "anonymous",
```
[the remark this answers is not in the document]
alexpeters1208: My understanding of `signature` is that it gives the argument names and types of the arguments to be dispatched on. Since `auth_type` must be a string, adding `signature = c(target = "character", auth_type = "character")` does not help dispatch, if we were to use it.

If you mean that `auth_type` should not have a default and must be user-specified, that is a different question. I think anon default makes sense.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+  },
+  signature = c("client_instance", "size")
+)
+
+#' @export
+setMethod(
+  "empty_table",
+  signature = c(client_instance = "Client"),
+  function(client_instance, size) {
+    verify_positive_int("size", size, TRUE)
```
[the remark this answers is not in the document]
alexpeters1208: It totally freaks out when I pass `0` here. I don't know what expected behavior is with 0 rows.

R/rdeephaven/R/client_wrapper.R
alexpeters1208: My only argument against this is that the error messages produced by R being unable to find a method to dispatch are not very helpful. So, if you use S4 dispatch to insist that `name` is a string, you have no control over the error message it gives off, and the one it gives off is not as informative as the one produced by `verify_string`.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+
+tibble_to_dh_table <- function(client, tibbl) {
+  arrow_tbl <- arrow_table(tibbl)
+  return(arrow_to_dh_table(client, arrow_tbl))
+}
+
+df_to_dh_table <- function(client, data_frame) {
+  arrow_tbl <- arrow_table(data_frame)
+  return(arrow_to_dh_table(client, arrow_tbl))
+}
```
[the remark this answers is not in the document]
alexpeters1208: Similar to my comment below, my problem with this is that if a user passes something other than one of the four acceptable types, the error message produced is not helpful.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,249 @@
...
+    } else if ((length(table_object_class) == 4 &&
+      table_object_class[[1]] == "Table" &&
+      table_object_class[[3]] == "ArrowObject")) {
+      rcpp_dh_table <- arrow_to_dh_table(client_instance, table_object)
+    } else {
+      stop(paste0("'table_object' must be either an R Data Frame, a dplyr Tibble, an Arrow Table, or an Arrow Record Batch Reader. Got an object of class ", table_object_class[[1]], " instead."))
+    }
+    return(new("TableHandle", .internal_rcpp_object = rcpp_dh_table))
+  }
+)
```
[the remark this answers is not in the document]
alexpeters1208: Similar to my comment below, my problem with this is that if a user passes something other than one of the four acceptable types, the error message produced is not helpful.

R/rdeephaven/R/helper_functions.R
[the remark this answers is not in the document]
chipkent: You could create those messages automatically.

R/rdeephaven/R/client_wrapper.R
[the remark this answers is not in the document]
chipkent: I don't know.  That's why it was a question.  Try in python.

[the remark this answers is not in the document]
alexpeters1208: Python doesn't complain.

R/rdeephaven/R/client_wrapper.R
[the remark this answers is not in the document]
alexpeters1208: First point resolved, second point not.

R/rdeephaven/R/client_wrapper.R
alexpeters1208: I think this can evolve with more generics to add more types without impacting the API at all.

R/rdeephaven/R/helper_functions.R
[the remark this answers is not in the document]
chipkent: I meant to change this line to something like:
```
if (!is_scalar && first_class(candidate) == "list") {
```
so that the grammar for all of your checks is correct for scalar vs vector as well as having all of the checks be correct.  That is the fundamental thing I was getting at.  Correctness of check and grammar that makes sense to the user.  e.g. if the check expects a scalar, why can our message talk about vectors and plural?

R/rdeephaven/R/client_wrapper.R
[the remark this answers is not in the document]
chipkent: Then you might want to understand that before deciding on the right assertion.  Check the python server and client.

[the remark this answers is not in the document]
alexpeters1208: It looks like the Python client does not impose any constraints on the size and lets the server complain if it's going to complain. I did not want to do this because if I remember correctly, the way that it failed with size=0 in R was pretty spectacular. I need to ask Corey to verify what C++ behavior should be, because R itself isn't doing anything with the size other than sending it to C++.

R/rdeephaven/R/helper_functions.R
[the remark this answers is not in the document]
alexpeters1208: I feel we have a lot more flexibility just passing the messages directly. If you I create stuff internally I can do that.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -1,104 +1,123 @@
...
-    to_data_frame = function() {
-      arrow_tbl <- self$to_arrow_table()
-      return(as.data.frame(as.data.frame(arrow_tbl))) # TODO: for some reason as.data.frame on arrow table returns a tibble, not a data frame
-    }
-  ),
-  private = list(
-    internal_table_handle = NULL,
-    is_static_field = NULL
-  )
+# TODO: Implement ncol, dim
```
chipkent, line 57: todo

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -1,104 +1,123 @@
...
+  function(x, ...) {
+    rbsr <- as_record_batch_reader(x)
+    arrow_tbl <- rbsr$read_table()
+    return(as_tibble(arrow_tbl))
+  }
+)
+
+#' @export
+setMethod(
+  "as.data.frame",
```
chipkent, line 96: for consistency in the API, should we also provide an `as_data_frame` method?

chipkent (changes requested):

R/rdeephaven/R/table_handle_wrapper.R
alexpeters1208: My understanding of this is that it's deprecated, seen by `?as_data_frame`. Would not hurt to provide it.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,256 @@
...
+  signature = c(client_instance = "Client", period = "numeric"),
+  function(client_instance, period, start_time = 0) {
+    verify_any_int("period", period, TRUE)
+    verify_any_int("start_time", start_time, TRUE)
+    return(new("TableHandle", .internal_rcpp_object = client_instance@.internal_rcpp_object$time_table(start_time, period)))
+  }
+)
+
+setGeneric(
+  "as_dh_table",
```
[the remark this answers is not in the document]
alexpeters1208: I think the initial name of `import_table` was clear after I got used to it. My very first suggestion months ago was `push_table`, which I still think is clearer than `import_table` on what that function is doing. If we're between `to_dh_table` and `as_dh_table`, `as` is more typical of R things, and I would expect the user to read the documentation to know what's going on.

R/rdeephaven/R/table_handle_wrapper.R
alexpeters1208: At the moment I don't believe C++ can support this directly

R/rdeephaven/R/table_ops.R
```diff
@@ -0,0 +1,642 @@
...
+
+#' @export
+setMethod(
+  "merge",
+  signature = c(x = "list", y = "NULL"),
+  function(x, y, ...) {
+    verify_type("x", x, "TableHandle", "Deephaven TableHandle", FALSE)
+    return(base_merge(x, y, c(...)))
+  }
+)
```
chipkent, line 113: It seems like most of our `merge` complexity is from us not limiting support to `list` inputs.  Can you remind me why the various kinds of `TableHandle` inputs seemed like a good idea when we talked?

R/rdeephaven/inst/tests/testthat/test_table_ops.R
```diff
@@ -0,0 +1,879 @@
...
+  expect_equal(as.data.frame(new_th9), new_df3)
+  
+  new_th10 <- merge(c(data$th5, data$th6, data$th6), data$th5)
+  expect_equal(as.data.frame(new_th10), new_df3)
+  
+  new_th11 <- merge(c(data$th5, data$th6), c(data$th6, data$th5))
+  expect_equal(as.data.frame(new_th11), new_df3)
+  
+  new_th12 <- merge(c(data$th5, data$th6, data$th6, data$th5))
+  expect_equal(as.data.frame(new_th12), new_df3)
```
chipkent, line 100: Should have a few tests that contain `NULL` within a `c()`.

R/rdeephaven/inst/tests/testthat/test_table_ops.R
```diff
@@ -0,0 +1,879 @@
...
+    rename(
+      Number1 = Number1.x, Number2 = Number2.x,
+      Number3 = Number1.y, Number4 = Number2.y
+    )
+  expect_equal(as.data.frame(new_th1), as.data.frame(new_tb1))
+
+  close(data$client)
+})
+
+# TODO: Verify that inner_join is the analog of exact_join
```
chipkent, line 856: todo

check the code for todos

R/rdeephaven/src/client.cpp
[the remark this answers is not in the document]
chipkent: Looks like other methods in libs would need const.  It is a good feature, but it frequently does not get used.

I this case, it might be correct to not have const, since Close does change the object.

R/rdeephaven/inst/tests/testthat/test_client_wrapper.R
```diff
@@ -0,0 +1,289 @@
...
+  )
+
+  return(list("df1" = df1, "df2" = df2, "df3" = df3, "df4" = df4))
+}
+
+##### TESTING GOOD INPUTS #####
+
+test_that("client dhConnection works in the simple case of anonymous authentication", {
+
+  # TODO: assumes server is actually running on localhost:10000, this is probably bad for CI
```
chipkent, line 32: Should this be a todo or just a comment?  Should this attempt to pull the target from an env variable and then just default to the localhost string to make CI config easier?

If these changes are made, they may need to happen in all test files, so I suggest a utility function.

Optional

R/rdeephaven/R/table_handle_wrapper.R
chipkent: I was suggesting adding it for consistency with the other methods we provide.

chipkent (changes requested):

R/rdeephaven/R/table_ops.R
```diff
@@ -0,0 +1,642 @@
...
+
+# We need many implementations of 'merge' because of the kind of args it should support
+
+base_merge <- function(x, y, ...) {
+  arg_list <- c(unlist(x), unlist(y), unlist(c(...)))
+  if (length(arg_list) == 1) {
+    return(arg_list[[1]])
+  }
```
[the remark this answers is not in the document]
alexpeters1208: No. Type gets verified by R's dispatch and by `verify_type("x", x, "TableHandle", "Deephaven TableHandle", FALSE)` before anything gets to this function. This comment made me go look and I realized I had left out the above call in one of the methods, so now it should all be airtight.

R/rdeephaven/R/client_wrapper.R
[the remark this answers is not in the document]
alexpeters1208: `import_table` is being used in Python and Go. I think `import_table` does not really tell you what the function does, and I definitely had to get used to the naming. That said, this may be another case of not wanting to spend time worrying about function names. I would advocate changing all of the API's to `push_to_table` if it wouldn't break anyone, but it probably would.

R/rdeephaven/inst/tests/testthat/test_client_wrapper.R
alexpeters1208: I think this whole thing should be revisited when CI gets hooked up. I don't see the point in spending time guessing what that is going to require now.

R/rdeephaven/R/table_ops.R
chipkent: Do we work with multiple list arguments?

R/rdeephaven/R/table_ops.R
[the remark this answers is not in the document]
chipkent: I don't think this statement is true.  Verification happens on the `x` and `y` args, but it does not happen on the varargs.

R/rdeephaven/R/table_ops.R
```diff
@@ -0,0 +1,643 @@
...
+  unwrapped_arg_list <- lapply(arg_list, strip_s4_wrapping)
+  return(new("TableHandle", .internal_rcpp_object = unwrapped_arg_list[[1]]$merge(unwrapped_arg_list[2:length(unwrapped_arg_list)])))
```
chipkent, line 16: I don't think that nulls in the varargs get stripped out before calling merge.

R/rdeephaven/R/table_ops.R
chipkent: Now that we see the implementation, would we be better off only supporting a single list arg -- that could contain null, table handles, and lists.  I'm concerned that, given how funky R is, we might miss something.

If we ship just supporting lists, it is easy to add more variants in the future.  If we ship with stuff we don't like, it is harder to take it away.

R/rdeephaven/R/client_wrapper.R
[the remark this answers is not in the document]
chipkent: Reading the Go code, it looks like this is a name created in Go and not an underlying gRPC API name.
https://github.com/deephaven/deephaven-core/blob/7ae3a2341522810af7cdc80f9fdcdc43772c80f0/go/pkg/client/flight_stub.go#L86
Same in python:
https://github.com/deephaven/deephaven-core/blob/7ae3a2341522810af7cdc80f9fdcdc43772c80f0/py/client/pydeephaven/session.py#L466
https://github.com/deephaven/deephaven-core/blob/7ae3a2341522810af7cdc80f9fdcdc43772c80f0/py/client/pydeephaven/_arrow_flight_service.py#L18

That gives some more freedom in naming.  Neither `importTable` nor `import_table` appear in our docs.  

Maybe lean to `push_to_table` here and create tickets for the other APIs to consider renaming and deprecating the old names.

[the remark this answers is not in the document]
chipkent: @jmao-denver should comment.

chipkent (changes requested):

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,262 @@
...
+  signature = c(target = "character"),
+  function(target,
+           auth_type = "anonymous",
+           auth_token = "",
+           session_type = "python",
+           use_tls = FALSE,
+           tls_root_certs = "",
+           int_options = list(),
+           string_options = list(),
+           extra_headers = list()) {
```
chipkent, line 29: There is some design debate on if there should be `username` and `password` arguments for basic auth.  The answer will probably be yes.  For maximum flexibility, for basic auth, consider continuing to support `auth_token` as it is as well as `username` and `password`.  Just throw an error if they are both specified.

chipkent (changes requested):

R/rdeephaven/R/table_ops.R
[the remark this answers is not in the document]
alexpeters1208: You are correct, the varargs are not type checked. But, our merge API has at least one required argument and we don't support a generic for `x` and `y` to be `NULL`, so the only case where a varargs entry ends up being the single element of a list and that check is true is `merge(c(NULL), c(NULL), other_thing)`. If I want to type check the varargs as well, I have to add checking for both `TableHandle` and `list` varargs, so my list of `merge` methods will get twice as long.

R/rdeephaven/R/table_ops.R
alexpeters1208: If we're going to restrict support, it should be to both of `merge(t1, t2, t3)` and `merge(c(t1, t2, t3))`. I don't think forcing coercion to a list is desirable if we can easily support `TableHandle` varargs (also easy to type check), and only supporting these two variants reduces the combinatorial mess.

R/rdeephaven/R/table_ops.R
alexpeters1208:
```
> unlist(c(NULL, 1, 2, 3, NULL, 4, NULL, NULL, 5))
[1] 1 2 3 4 5
```

The call to `unlist(c(...))` in `arg_list <- c(unlist(x), unlist(y), unlist(c(...)))` takes care of it by default.

R/rdeephaven/R/client_wrapper.R
alexpeters1208: Are we getting into a world where this signature is getting big and ugly, and could easily get bigger and uglier in the future? Is there still a reason that some R-idiomatic variant of `ClientOptions` is not more robust and flexible than this?

R/rdeephaven/R/table_ops.R
chipkent: `merge(t1,t2,t3)` is not easy to support properly.  It is the source of most of this problem.  Here, any of these can be null.  Then you get a mess.

R/rdeephaven/inst/tests/testthat/test_table_handle_wrapper.R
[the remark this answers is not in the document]
chipkent: You are not verifying that the dynamic handles return true.

R/rdeephaven/src/client.cpp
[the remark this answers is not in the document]
chipkent: Is there a ticket?

R/rdeephaven/src/client.cpp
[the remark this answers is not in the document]
chipkent: This should be in the API.  Expecting an R user to do DH calls to accomplish this is not how it is done in DH.  That would require creating a new table column, dropping it if you don't want it, etc.

R/rdeephaven/R/table_ops.R
alexpeters1208: `merge(t1, t2, t3)` is (as far as I understand) easy to support directly, because you can type check varargs as long as they are all of the same type. The source of this problem is trying to support many combinations of TableHandles, lists, and NULLs. I think we can support vararg TableHandles and single list with two methods.

R/rdeephaven/inst/tests/testthat/test_table_handle_wrapper.R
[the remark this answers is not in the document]
alexpeters1208: but... is_static on dynamic handles should return false?

R/rdeephaven/src/client.cpp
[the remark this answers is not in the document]
alexpeters1208: https://github.com/deephaven/deephaven-core/issues/4319

R/rdeephaven/R/client_wrapper.R
```diff
@@ -40,11 +42,12 @@ setMethod(
     if (auth_type == "anonymous") {
       options$set_default_authentication()
     } else if (auth_type == "basic") {
-      if (auth_token != "") {
-        verify_string("auth_token", auth_token, TRUE)
-        options$set_basic_authentication(auth_token)
+      if ((username != "") && (password != "")) {
+        verify_string("username", username, TRUE)
+        verify_string("password", password, TRUE)
+        options$set_basic_authentication(username, password)
```
chipkent, line 48: This is not doing what my review comment asked for.  We should be supporting both passwords AND auth_token for basic auth.  This is just doing one.  auth_token should be exactly like the code that got deleted here.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -102,6 +105,10 @@ setMethod(
     if ((auth_token != "") && (auth_type == "anonymous")) {
       warning("'auth_token' was set but it will not be used, as 'auth_type' is 'anonymous'.")
     }
+    
+    if (((username != "") || (password != "")) && auth_type != "basic") {
+      warning("At least one of 'username' and 'password' were set but they will not be used, as 'auth_type' is not 'basic'.")
+    }
```
chipkent, line 111: This check is redundant with the test in the "basic" block above.  The test should be in the basic block because then it localizes the logic for "basic" stuff.

R/rdeephaven/R/table_ops.R
```diff
@@ -622,7 +634,7 @@ setMethod(
 #' @export
 setGeneric(
   "sort",
-  function(table_handle, by = character(), descending = FALSE, ...) {
+  function(table_handle, by = character(), descending = FALSE, abs_col = FALSE, ...) {
```
chipkent, line 637: can we use `abs`?

R/rdeephaven/R/table_ops.R
```diff
@@ -632,12 +644,16 @@ setGeneric(
 setMethod(
   "sort",
   signature = c(table_handle = "TableHandle"),
-  function(table_handle, by, descending = FALSE) {
+  function(table_handle, by, descending = FALSE, abs_col = FALSE) {
```
chipkent, line 647: can we use `abs`?

R/rdeephaven/src/client.cpp
```diff
@@ -335,9 +329,8 @@ class ClientOptionsWrapper {
         internal_options->SetDefaultAuthentication();
     }
 
-    void SetBasicAuthentication(const std::string &authentication_token) {
-        const std::string authentication_token_base64 = Base64Encode(authentication_token);
-        internal_options->SetCustomAuthentication("Basic", authentication_token_base64);
+    void SetBasicAuthentication(const std::string &username, const std::string &password) {
+        internal_options->SetBasicAuthentication(username, password);
```
chipkent, line 333: See the prior note about needing to support both ways.

R/rdeephaven/R/table_ops.R
```diff
@@ -0,0 +1,659 @@
...
+  signature = c(x = "TableHandle", y = "TableHandle"),
+  function(x, y, ...) {
+    if (length(c(...)) != 0) {
+      verify_type("...", c(...), "TableHandle", "Deephaven TableHandle", FALSE)
+    }
+    return(base_merge(x, y, c(...)))
+  }
+)
+
+#' #' @export
```
chipkent, line 50: commented out code

R/rdeephaven/R/table_ops.R
```diff
@@ -0,0 +1,659 @@
...
+setMethod(
+  "merge",
+  signature = c(x = "TableHandle", y = "missing"),
+  function(x, y = NULL) {
+    return(base_merge(x, y))
+  }
+)
+
+#' @export
+setMethod(
```
chipkent, line 39: I do not think these merge methods are doing what is desired.  There is a big contortion to make `merge(t1,t2,t3)` work, but it never works right.  For example `merge(NULL, NULL, t1)` is within scope of what we want, but it will not work.  It isn't obvious how to even make it work.  On the other hand it is easy to make `merge(c(NULL,NULL,t1))` work.  If we don't have a compelling reason to support `merge(t1,t2,t3)` we should remove it and get a clean implementation that works as desired.

R/rdeephaven/src/client.cpp
```diff
@@ -241,29 +241,23 @@ class TableHandleWrapper {
         return new TableHandleWrapper(internal_tbl_hdl.Merge(converted_sources));
     };
 
-    TableHandleWrapper* Sort(std::vector<std::string> columnSpecs, std::vector<bool> descending) {
+    TableHandleWrapper* Sort(std::vector<std::string> columnSpecs, std::vector<bool> descending, std::vector<bool> absCol) {
```
chipkent, line 244: rename to `abs`.  Less confusing.

chipkent (changes requested):

R/rdeephaven/R/table_ops.R
chipkent: Can we support:
1. list
2. varargs of Union[TableHandle,None] (python syntax)?  I think I've seen arrays of types before.

R/rdeephaven/R/table_ops.R
alexpeters1208: I can make `merge(NULL, NULL, t1)` work, but I don't know why we should. I think that `merge(t1, t2, t3)`, `merge(c(t1, t2, t3))` and `merge(t1, c(t2, t3))` should work, the first and last cases coming from piping use cases. I can support NULL if needed. The class unions you sent may be the way to go. This is terrible.

R/rdeephaven/R/table_ops.R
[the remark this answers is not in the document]
alexpeters1208: I've now made that not the case.

R/rdeephaven/R/table_ops.R
alexpeters1208: Will create a conflict with base R's `abs()` in the scope of the function. Probably not a killer, but should be aware of it.

chipkent: The problem with `abs_col` is that it mixes with our existing nomenclature where `col` is a column name.  Here it is not.  `abs` makes it clear that it is how sorting would happen, as would something like `sort_abs`.

R/rdeephaven/DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: rdeephaven
 Type: Package
 Title: R Client for Deephaven Core
-Version: 1.0
+Version: 0.26.0
```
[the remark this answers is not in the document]
chipkent: Bump to `0.27.1` and create a ticket to follow up with @devinrsmith when he is back.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,110 +1,275 @@
...
+    if ((auth_token != "") && (auth_type == "anonymous")) {
+      warning("'auth_token' was set but it will not be used, as 'auth_type' is 'anonymous'.")
+    }
+    
+    if (((username != "") || (password != "")) && auth_type != "basic") {
+      warning("At least one of 'username' and 'password' were set but they will not be used, as 'auth_type' is not 'basic'.")
+    }
+
+    if ((tls_root_certs != "") && (use_tls == FALSE)) {
+      warning("'tls_root_certs' was set but it will not be used, as 'use_tls is FALSE.")
```
chipkent, line 120:
```suggestion
      warning("'tls_root_certs' was set but it will not be used, as 'use_tls' is FALSE.")
```

R/rdeephaven/R/table_ops.R
```diff
@@ -0,0 +1,595 @@
...
+  return(new("TableHandle", .internal_rcpp_object = unwrapped_arg_list[[1]]$merge(unwrapped_arg_list[2:length(unwrapped_arg_list)])))
+}
+
+# supports merge(c(t1, t2, t3))
+#' @export
+setMethod(
+  "merge",
+  signature = c(x = "list", y = "missing"),
+  function(x, y = NULL) {
+    verify_type("x", x, "TableHandle", "Deephaven TableHandle", FALSE)
```
chipkent, line 25: What is the behavior of this on null elements?

R/rdeephaven/R/table_ops.R
```diff
@@ -0,0 +1,595 @@
...
+#' @export
+setMethod(
+  "merge",
+  signature = c(x = "TableHandle", y = "list"),
+  function(x, y) {
+    verify_type("y", y, "TableHandle", "Deephaven TableHandle", FALSE)
+    return(base_merge(x, y))
+  }
+)
+
```
chipkent, line 63: I still have some questions about this `merge`.

1. How does R handle dispatch for NULL?  Is NULL any type or is NULL a specific type?  If it is any type, how is disambiguation handled?  If it is a specific type, then it is a type we need to handle.

2. We have a nasty product of types that need to be handled.
Types are {TableHandle, list, NULL(maybe), missing}.  This would mean we would need to support:
* TH, missing
* list, missing
* NULL, missing (?)
* TH, TH
* list, list
* TH, list
* list, TH 
* NULL, TH (?)
* TH, NULL (?)
* NULL, list (?)
* list, NULL (?)
* NULL, NULL (?)

3. Can we simply handle the above with a union type so that we don't need to implement the cartesian product?

chipkent (changes requested):

R/rdeephaven/DESCRIPTION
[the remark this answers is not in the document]
alexpeters1208: https://github.com/deephaven/deephaven-core/issues/4323

R/rdeephaven/R/table_ops.R
```diff
@@ -1,65 +1,13 @@
...
-#' @export
-setMethod(
-  "merge",
-  signature = c(x = "TableHandle", y = "missing"),
-  function(x, y = NULL) {
-    return(base_merge(x, y))
+merge_tables <- function(...) {
+  table_list <- unlist(c(...))
+  verify_type("table_list", table_list, "TableHandle", "Deephaven TableHandle", FALSE)
+  if (length(table_list) == 1) {
```
chipkent, line 5: I think `length(table_list) == 0` should return NULL.  This should happen if someone calls `merge_tables(NULL)`.  Test for this also.

chipkent (changes requested):

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,3 +1,76 @@
...
+#' following list of methods for facilitating server requests.
+#'
+#' - `open_table(client, name)`: Opens a table named `"name"` on the server, if
+#' it exists. Returns a `TableHandle`.
+#' - `empty_table(client, size)`: Creates an empty table on the server with no
+#' columns and `size` rows. Thus, `size` must be a non-negative integer. Returns a `TableHandle`.
+#' - `time_table(client, period, start_time)`:
+#' - `as_dh_table(client, table_object)`:
+#' - `run_script(client, script)`:
+#' - `close(client)`:
```
chipkent, line 47: missing docs

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,3 +1,76 @@
...
+#' na_array$null_count
+#'
+#' # zero-copy slicing; the offset of the new Array will be the same as the index passed to $Slice
+#' new_array <- na_array$Slice(5)
+#' new_array$offset
+#'
+#' # Compare 2 arrays
+#' na_array2 <- na_array
+#' na_array2 == na_array # element-wise comparison
+#' na_array2$Equals(na_array) # overall comparison
```
chipkent, line 72: Seems totally out of context.

chipkent (approved):

PR #4242, opened 2023-07-29
Add callbacks to KafkaIngester consumer loop.

### LINKED ISSUE #4239: Provide KafkaConsumer Callback in Poll Loop
As a systems integrator, I want to have access to the KafkaConsumer inside of the poll loop so that I can control when messages are committed to the broker.

cpwright:
Closes #4239.


extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
```diff
@@ -176,6 +200,55 @@ public KafkaIngester(
...
+        this(log, props, topic, partitionFilter, partitionToStreamConsumer,
+                new IntToLongFunctionAdapter(partitionToInitialSeekOffset));
+    }
+
+    /**
+     * Determines the initial offset to seek to for a given TopicPartition.
+     */
+    @FunctionalInterface
+    public interface PartitionToInitialOffsetFunction {
+        long partitionToInitialOffset(KafkaConsumer<?, ?> consumer, TopicPartition topicPartition);
```
rcaudy, line 233: `InitialOffsetLookup`? Name is too precise and insufficiently accurate.

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
```diff
@@ -176,6 +200,55 @@ public KafkaIngester(
...
+            @NotNull final String topic,
+            @NotNull final IntPredicate partitionFilter,
+            @NotNull final Function<TopicPartition, KafkaRecordConsumer> partitionToStreamConsumer,
+            @NotNull final IntToLongFunction partitionToInitialSeekOffset) {
+        this(log, props, topic, partitionFilter, partitionToStreamConsumer,
+                new IntToLongFunctionAdapter(partitionToInitialSeekOffset));
+    }
+
+    /**
+     * Determines the initial offset to seek to for a given TopicPartition.
```
rcaudy, line 229:
```suggestion
     * Determines the initial offset to seek to for a given KafkaConsumer and TopicPartition.
```

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
```diff
@@ -176,6 +200,55 @@ public KafkaIngester(
...
+    @FunctionalInterface
+    public interface PartitionToInitialOffsetFunction {
+        long partitionToInitialOffset(KafkaConsumer<?, ?> consumer, TopicPartition topicPartition);
+    }
+
+    /**
+     * Adapts an IntToLongFunction to a PartitionToInitialOffsetFunction by ignoring the topic and consumer parameters.
+     */
+    public static class IntToLongFunctionAdapter implements PartitionToInitialOffsetFunction {
+        final IntToLongFunction function;
```
rcaudy, line 240:
```suggestion

        private final IntToLongFunction function;
```

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
```diff
@@ -281,7 +355,9 @@ private void consumerLoop() {
             }
             final long beforePoll = System.nanoTime();
             final long remainingNanos = beforePoll > nextReport ? 0 : (nextReport - beforePoll);
-            boolean more = pollOnce(Duration.ofNanos(remainingNanos));
+            consumerLoopCallbacks.forEach(x -> x.beforePoll(kafkaConsumer));
+            final boolean more = pollOnce(Duration.ofNanos(remainingNanos));
+            consumerLoopCallbacks.forEach(x -> x.afterPoll(kafkaConsumer, more));
```
rcaudy, line 360: I think you need some error handling here. Do we trust `beforePoll` and `afterPoll` to never throw exceptions? If they do throw, we should probably advise the downstream consumers. We should also handle this and not just hit the thread's UncaughtExceptionHandler.

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -1603,15 +1603,50 @@ public Function<TopicPartition, KafkaRecordConsumer> visit(@NotNull final PerPar
...
             @NotNull final Consume.KeyOrValueSpec keySpec,
             @NotNull final Consume.KeyOrValueSpec valueSpec,
             @NotNull final KafkaTools.StreamConsumerRegistrarProvider streamConsumerRegistrarProvider) {
+        return consume(kafkaProperties, topic, partitionFilter,
+                new KafkaIngester.IntToLongFunctionAdapter(partitionToInitialOffset), keySpec, valueSpec,
+                streamConsumerRegistrarProvider);
+    }
+
+    /**
+     * Consume from Kafka to a {@link StreamConsumer} supplied by {@code streamConsumerRegistrar}.
```
rcaudy, line 1623: Can you correct the one you copied from, too?
```suggestion
     * Consume from Kafka to {@link StreamConsumer stream consumers} supplied by {@code streamConsumerRegistrar}.
```

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -1603,15 +1603,50 @@ public Function<TopicPartition, KafkaRecordConsumer> visit(@NotNull final PerPar
      *        {@link TableDefinition}. See {@link StreamConsumerRegistrarProvider#single(SingleConsumerRegistrar)
      *        single} and {@link StreamConsumerRegistrarProvider#perPartition(PerPartitionConsumerRegistrar)
      *        per-partition}.
+     * @return the newly created KafkaIngester, the caller must call start on the returned object to begin processing
+     *         messages
```
rcaudy, line 1607:
```suggestion
     * @return the newly created KafkaIngester; the caller must call {@link KafkaIngester#start() start} on the returned object to begin processing
     *         messages
```

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -1603,15 +1603,50 @@ public Function<TopicPartition, KafkaRecordConsumer> visit(@NotNull final PerPar
...
      */
-    public static void consume(
+    public static KafkaIngester consume(
```
rcaudy, line 1609: I don't think that this is an innocuous change. It makes `KafkaIngester` part of the API. We should at least consider restricting the shutdown methods to package-private, or putting a simpler interface in front of the implementation before returning it.

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
```diff
@@ -50,6 +52,8 @@ public int getIntKey(@NotNull final TopicPartition topicPartition) {
                 }
             });
 
+    private final Collection<ConsumerLoopCallback> consumerLoopCallbacks = new CopyOnWriteArrayList<>();
```
rcaudy, line 55: Should we consider whether this should be a `WeakReferenceManager`?

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
devinrsmith, line 1607: This is a breaking change to an existing public API; I'd suggest returning void and internally starting the consume (same as old behavior). I think your new `KafkaIngester consume(...PartitionToInitialOffsetFunction...)` method should suffice for any new callers who want more explicit control?

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
```diff
@@ -404,4 +480,22 @@ public void shutdownPartition(final int partition) {
...
+    }
+
+    /**
+     * Remove a callback from the consumer loop.
+     * 
+     * @param callback a previously added callback.
+     */
+    public void removeConsumerLoopCallback(@NotNull final ConsumerLoopCallback callback) {
+        consumerLoopCallbacks.remove(callback);
+    }
```
devinrsmith, line 500: I'm curious, does DHE need dynamic add/remove callback support? Would `Set<ConsumerLoopCallback>` in constructor suffice?

devinrsmith (changes requested):

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
rcaudy: Worth pointing out, this API call was just added last release, specifically for DnD use cases.

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
rcaudy: Good point!

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -1603,24 +1603,25 @@ public Function<TopicPartition, KafkaRecordConsumer> visit(@NotNull final PerPar
      *        {@link TableDefinition}. See {@link StreamConsumerRegistrarProvider#single(SingleConsumerRegistrar)
      *        single} and {@link StreamConsumerRegistrarProvider#perPartition(PerPartitionConsumerRegistrar)
      *        per-partition}.
-     * @return the newly created KafkaIngester, the caller must call start on the returned object to begin processing
-     *         messages
+     * @return the newly created KafkaIngester; the caller must call {@link KafkaIngester#start() start} on the returned
```
rcaudy, line 1606: This bit about start is not true anymore....

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
```diff
@@ -355,9 +360,29 @@ private void consumerLoop() {
...
+                if (consumerLoopCallback != null) {
+                    try {
+                        consumerLoopCallback.afterPoll(kafkaConsumer, more);
+                    } catch (Exception e) {
+                        log.error().append(logPrefix).append("Exception while executing afterPoll callback:").append(e)
+                                .append(", aborting.").endl();
+                        more = false;
+                    }
+                }
+            }
```
rcaudy, line 385: We need to deliver errors to the consumers...

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -1611,7 +1613,42 @@ public static void consume(
...
+     * @param keySpec Conversion specification for Kafka record keys
+     * @param valueSpec Conversion specification for Kafka record values
+     * @param streamConsumerRegistrarProvider A provider for a function to
+     *        {@link StreamPublisher#register(StreamConsumer) register} {@link StreamConsumer} instances. The registered
+     *        stream consumers must accept {@link ChunkType chunk types} that correspond to
+     *        {@link StreamChunkUtils#chunkTypeForColumnIndex(TableDefinition, int)} for the supplied
+     *        {@link TableDefinition}. See {@link StreamConsumerRegistrarProvider#single(SingleConsumerRegistrar)
+     *        single} and {@link StreamConsumerRegistrarProvider#perPartition(PerPartitionConsumerRegistrar)
+     *        per-partition}.
+     * @return the newly created KafkaIngester, the caller must call start on the returned object to begin processing
```
rcaudy, line 1640: Same as above

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
cpwright: I followed the existing pattern in pollOnce where the poll() method could return an error:
        final ConsumerRecords<?, ?> records;
        try {
            ++pollCalls;
            records = kafkaConsumer.poll(timeout);
        } catch (WakeupException we) {
            // we interpret a wakeup as a signal to stop /this/ poll.
            return true;
        } catch (Exception ex) {
            log.error().append(logPrefix).append("Exception while polling for Kafka messages:").append(ex)
                    .append(", aborting.").endl();
            return false;
        }

rcaudy (dismissed):

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaIngester.java
rcaudy: We discussed offline that I fixed that bug yesterday, and you have generalized a solution from that fix. Marking resolved.

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2978
Reference: https://github.com/deephaven/deephaven.io/issues/2979

PR #4249, opened 2023-07-31
Apply Tidyverse styleguide

alexpeters1208:
This applies the Tidyverse style guide to the R package, which will make the diff in #4240 more meaningful. Confirmed that all unit tests still pass.


R/rdeephaven/R/client_options_wrapper.R
chipkent, line 1: PR is just missing one thing... documentation on how to run the tidyverse code formatting so that it can be done on future changes.

chipkent (changes requested):

chipkent (approved):

PR #4252, opened 2023-08-01
Ensure the output of a docker run (eg, tests) is copied even if the run (tests) fail.

jcferretti:


niloc132 (approved):
Is this sufficient, does it copy into the test-results/ dir or whatnot, or just into the temp dir?

niloc132:
This likely needs to be revisited, as it does not copy into test-results/, and so failures like this still won't provide good feedback when they happen in CI.

PR #4254, opened 2023-08-01
For docker runs (tests) ensure we get results even when the run (test) fails.

jcferretti:


buildSrc/src/main/groovy/Docker.groovy
```diff
@@ -334,18 +345,14 @@ class Docker {
...
-            return project.tasks.register(taskName, Sync) { sync ->
-                sync.with {
-                    dependsOn buildAndRun
-
-                    // run the provided closure first
-                    cfg.copyOut.execute(sync)
 
-                    // then set the from location
-                    from dockerCopyLocation
+            return project.tasks.register(taskName) { task ->
```
niloc132, line 352: this doesn't make sense to me, it doesn't do anything, but it still claims it has outputs - why not return syncAfterBuildAndRun, which already has the outputs declared (by being of type Sync)?

jcferretti: We discussed over slack and you said you will followup.  The gist of the discussion was:
```
Because syncAfterBuildAndRun does not depend on buildAndRun
so buildAndRun wouldn't run.  We can't just return buildAndRun because
the rest of the code wants both ${task}Run and ${task} to exist.
```
You said you wanted to look at the rest of the code to see if we could remove the need for having both `${task}Run` and `$task` exist.

jmao-denver (approved):
LGTM

buildSrc/src/main/groovy/Docker.groovy
```diff
@@ -334,20 +345,11 @@ class Docker {
...
-                    dependsOn buildAndRun
-
-                    // run the provided closure first
-                    cfg.copyOut.execute(sync)
 
-                    // then set the from location
-                    from dockerCopyLocation
-                }
-            }
+            return buildAndRun
```
niloc132, line 352: inline this, so you don't need a local variable?

niloc132 (approved):
Verified that the reports are copied out both when tests are successful and when they fail.

Verified that the web-client-ui:ui task and others correctly are still marked as up-to-date, and don't need to re-run when there are no changes to their inputs.

PR #4268, opened 2023-08-03
(WIP) Invalidate old file handles when overwriting a parquet file

malhotrashivam:
This will prevent users from reading incorrect data from old handles


extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -185,20 +186,28 @@ public static void write(
...
                     tableInfoBuilder.addGroupingColumns(GroupingColumnInfo.of(parquetColumnName,
                             destDirPath.relativize(Paths.get(groupingPath)).toString()));
                     write(auxiliaryTables[gci], auxiliaryTables[gci].getDefinition(), writeInstructions, groupingPath,
                             Collections.emptyMap());
                 }
             }
             write(t, definition, writeInstructions, destPathName, incomingMeta, tableInfoBuilder);
+
+            // Invalidate old file handles so that we don't read from files which have been updated
+            TrackedFileHandleFactoryWithLookup.getInstance().invalidateHandles(new File(destPathName));
```
rcaudy, line 199: This should probably happen up in `ParquetTools`, based on the other branch. We should invalidate all dest files *before* installing shadow files.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/util/TrackedSeekableChannelsProvider.java
```diff
@@ -34,9 +35,20 @@ public static SeekableChannelsProvider getInstance() {
         return instance;
     }
 
+    public static SeekableChannelsProvider getInstance(@NotNull final TrackedFileHandleFactory fileHandleFactory) {
```
rcaudy, line 38: We need to think about this pattern. There's potentially a race here.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/util/TrackedSeekableChannelsProvider.java
```diff
@@ -34,9 +35,20 @@ public static SeekableChannelsProvider getInstance() {
...
+                }
+            }
+        }
+        return instance;
+    }
+
     private final TrackedFileHandleFactory fileHandleFactory;
 
-    public TrackedSeekableChannelsProvider(@NotNull final TrackedFileHandleFactory fileHandleFactory) {
+    private TrackedSeekableChannelsProvider(@NotNull final TrackedFileHandleFactory fileHandleFactory) {
```
rcaudy, line 51: I wonder if we should be integration the invalidation into this class, instead of the factory,

engine/table/src/main/java/io/deephaven/engine/util/file/TrackedFileHandleFactoryWithLookup.java
```diff
@@ -0,0 +1,127 @@
...
+ * invalidate old file handles when overwriting an existing file.
+ */
+public class TrackedFileHandleFactoryWithLookup extends TrackedFileHandleFactory {
+
+    private static volatile TrackedFileHandleFactoryWithLookup instance;
+
+    /**
+     * Mapping from absolute file path to list of handles
+     */
+    private final Map<String, LinkedList<WeakReference<FileHandle>>> fileToHandleMap = new HashMap<>();
```
rcaudy, line 36: We need this map and its lists to be accessed in a thread-safe manner. Either `synchronized`, or using concurrent data structures.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/util/TrackedSeekableChannelsProvider.java
rcaudy: The `CachedChannelProvider` instances also need to be invalidated so that they cannot allocate new channels. I think we want to start using a map from File (canonicalized) to WeakReference<CachedChannelProvider>. The provider could then start keeping track of all outstanding handles, rather than just the pooled ones.

extensions/parquet/base/build.gradle
```diff
@@ -12,6 +12,7 @@ dependencies {
     implementation project(':Base')
     implementation project(':Util')
     implementation depCommonsIo
+    implementation project(path: ':engine-table') // TODO Look for a better way to do this
```
lbooker42, line 15: We need help from more experienced `gradle`ers to decide if this is acceptable or there is a better solution.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/util/CachedChannelProvider.java
```diff
@@ -28,6 +30,12 @@ public class CachedChannelProvider implements SeekableChannelsProvider {
     private long logicalClock;
     private long pooledCount;
 
+    /**
+     * An invalid {@link CachedChannelProvider} will invalidate all the channels it has produced and will not create any
+     * more channels. Used to prevent creating channels to files which have been overwritten.
```
lbooker42, line 35:
```suggestion
     * Invalidating a {@link CachedChannelProvider} will invalidate all the channels it has produced and prevent creation of 
     * new channels. This prevents creating channels to files which have been overwritten.
```

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/util/CachedChannelProvider.java
```diff
@@ -45,20 +53,32 @@ enum ChannelType {
     private final RAPriQueue<PerPathPool> releasePriority =
             new RAPriQueue<>(8, PerPathPool.RAPQ_ADAPTER, PerPathPool.class);
 
+    /**
+     * Following stores all the channels (and not just the pooled ones) created by this provider for any path and is
+     * useful for invalidating all file handles associated with this provider.
```
lbooker42, line 58:
```suggestion
     * Stores all channels (not just the pooled ones) created by this provider for all paths. Used
     * for invalidating file handles associated with this provider.
```

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/util/CachedChannelProvider.java
```diff
@@ -67,10 +87,50 @@ public SeekableByteChannel getWriteChannel(@NotNull final Path path, final boole
...
+        }
+        final SeekableByteChannel newWriteChannel = wrappedProvider.getWriteChannel(path, append);
+        channelCreatorHelper(path, newWriteChannel);
+        return new CachedChannel(newWriteChannel, channelType, pathKey);
+    }
+
+    private void channelCreatorHelper(@NotNull final Path path, @NotNull final SeekableByteChannel newChannel)
+            throws IOException {
+        // If channel creator already marked invalid, mark the new channels invalid.
+        // This is needed because we cannot return a null channel. So we return invalidated channels.
```
lbooker42, line 102:
```suggestion
        // Required because we cannot return a null channel so we return invalidated channels.
```

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/util/LocalFSChannelProvider.java
```diff
@@ -32,3 +32,4 @@ public SeekableByteChannel getWriteChannel(@NotNull final Path filePath, final b
         return result;
     }
 }
+// TODO Can we remove this class since its never used?
```
lbooker42, line 35: IMO, yes. @nbauernfeind or @rcaudy might need to concur though.

lbooker42 (commented):
Except for the gradle question, this looks good to merge.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/util/CachedChannelProvider.java
```diff
@@ -67,10 +88,51 @@ public SeekableByteChannel getWriteChannel(@NotNull final Path path, final boole
...
+        final SeekableByteChannel newWriteChannel = wrappedProvider.getWriteChannel(path, append);
+        channelCreatorHelper(path, newWriteChannel);
+        return new CachedChannel(newWriteChannel, channelType, pathKey);
+    }
+
+    private void channelCreatorHelper(@NotNull final Path path, @NotNull final SeekableByteChannel newChannel)
+            throws IOException {
+        // If channel creator is already marked invalid, mark the new channels invalid.
+        // Required because CachedChannelProvider cannot return a null channel, so it returns invalid channels.
+        // TODO Should we just throw an exception here?
```
malhotrashivam, line 104: Need opinion on this behavior.

extensions/parquet/base/src/test/java/io/deephaven/parquet/base/util/CachedChannelProviderTest.java
```diff
@@ -154,6 +170,60 @@ public void testReuse10() throws IOException {
...
+        final SeekableChannelsProvider wrappedProvider = new TestChannelProvider();
+        // Register cached channel providers with different files
+        for (int i = 0; i < CachedChannelProviderTracker.PROVIDER_MAP_CLEANUP_LIMIT - 1; i++) {
+            SeekableChannelsProvider ccp = new CachedChannelProvider(wrappedProvider, 100);
+            ccp.getReadChannel("rc" + i);
+        }
+        Assert.assertEquals(CachedChannelProviderTracker.getInstance().size(),
+                CachedChannelProviderTracker.PROVIDER_MAP_CLEANUP_LIMIT - 1);
+
+        // Trigger garbage collection to clear any old providers
```
malhotrashivam, line 218: Can I add a test that relies on gc?
I saw some other test files doing this.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/util/CachedChannelProvider.java
lbooker42: An exception with a clear explanation is better, I think.

extensions/parquet/base/src/test/java/io/deephaven/parquet/base/util/CachedChannelProviderTest.java
lbooker42: I think this is legit, but I would get a more qualified second opinion.

lbooker42: Maybe can force a reset on the first usage, but GC looks to be required to test this code.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/util/CachedChannelProvider.java
```diff
@@ -37,6 +38,11 @@ public class CachedChannelProvider implements SeekableChannelsProvider {
      */
     private boolean invalid;
 
+    /**
+     * The path to file that caused this provider to be invalidated.
```
lbooker42, line 42:
```suggestion
     * The path to the file that caused this provider to be invalidated.
```

lbooker42 (dismissed):

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/util/CachedChannelProvider.java
```diff
@@ -45,6 +61,14 @@ enum ChannelType {
     private final RAPriQueue<PerPathPool> releasePriority =
             new RAPriQueue<>(8, PerPathPool.RAPQ_ADAPTER, PerPathPool.class);
 
+    /**
+     * Stores all channels (not just the pooled ones) created by this provider for all paths. Used for invalidating file
+     * handles associated with this provider.
+     */
+    private final Collection<WeakReference<SeekableByteChannel>> channelList = new ArrayList<>();
```
rcaudy, line 68: Having this use weak refs is probably unnecessary. In general, it's an error (resource leak) to make a channel and not close it. The code we have using these channels is very careful about that.

Maybe we want to:
1. Integrate "outstanding channel tracking" into `PerPathPool`.
2. Make `CachedChannel` instances aware of their pool.
3. The pool can then be aware of available and outstanding channels, and handle invalidating both (and itself).

I'm trying to think of a way to get rid of more of the data structures here, but not at a useful point yet.

PR #4269, opened 2023-08-03
Add python wrapper for TablePublisher

### LINKED ISSUE #4259: Wrap TablePublisher in the Python server API

devinrsmith:
Fixes #4259

### DIFF
--- a/Integrations/src/main/java/io/deephaven/integrations/python/PythonRunnable.java
+++ b/Integrations/src/main/java/io/deephaven/integrations/python/PythonRunnable.java
@@ -1,0 +1,21 @@
+package io.deephaven.integrations.python;
+
+import org.jpy.PyObject;
+
+import java.util.Objects;
+
+/**
+ * A {@link Runnable} implementation which calls a Python callable.
+ */
+public class PythonRunnable implements Runnable {
+    private final PyObject pyCallable;
+
+    public PythonRunnable(PyObject pyCallable) {
+        this.pyCallable = Objects.requireNonNull(pyCallable);
+    }
+
+    @Override
+    public void run() {
+        pyCallable.call("__call__");
+    }
+}
--- a/py/server/deephaven/stream/table_publisher.py
+++ b/py/server/deephaven/stream/table_publisher.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
+# Copyright (c) 2016-2023 Deephaven Data Labs and Patent Pending
 #
 """The table_publisher module supports publishing Deephaven Tables into blink Tables."""
 
@@ -22,7 +22,7 @@ _JRuntimeException = jpy.get_type("java.lang.RuntimeException")
 
 
 class TablePublisher(JObjectWrapper):
-    """Supports publishing Table data."""
+    """The interface for publishing table data into a blink table."""
 
     j_object_type = _JTablePublisher
 
@@ -35,19 +35,21 @@ class TablePublisher(JObjectWrapper):
 
     @property
     def is_alive(self) -> bool:
-        """Checks whether this publisher is alive; if False, the publisher should stop publishing new data and release
-        any related resources as soon as practicable since publishing won't have any downstream effects.
+        """Checks whether this is alive; if False, the caller should stop adding new data and release any related
+        resources as soon as practicable since adding new data won't have any downstream effects.
 
-        Once this is False, it will always remain False. For more prompt notifications, publishers may
-        prefer to use on_shutdown_callback during construction.
+        Once this is False, it will always remain False. For more prompt notifications, callers may prefer to use
+        on_shutdown_callback during construction.
 
         Returns:
             if this publisher is alive
         """
         return self.j_table_publisher.isAlive()
 
-    def add(self, table: Table):
-        """Adds a snapshot of the data from table into the blink table according to the blink table's definition.
+    def add(self, table: Table) -> None:
+        """Adds a snapshot of the data from table into the blink table. The table must contain a superset of the columns
+        from the blink table's definition; the columns may be in any order. Columns from table that are not in the blink
+        table's definition are ignored.
 
         All of the data from table will be:
 
@@ -60,17 +62,15 @@ class TablePublisher(JObjectWrapper):
         """
         self.j_table_publisher.add(table.j_table)
 
-    def publish_failure(self, failure: Exception):
-        """Publish a failure for notification to the listeners of the blink table. Future calls to add will silently
-        return. Will cause the on-shutdown callback to be invoked if it hasn't already been invoked.
+    def publish_failure(self, failure: Exception) -> None:
+        """Indicate that data publication has failed. Blink table listeners will be notified of the failure, the
+        on-shutdown callback will be invoked if it hasn't already been, this publisher will no longer be alive, and
+        future calls to add will silently return without publishing. These effects may resolve asynchronously.
 
         Args:
             failure (Exception): the failure
         """
         self.j_table_publisher.publishFailure(_JRuntimeException(str(failure)))
-
-
-# Note: we may want to expand the column_definitions typing here and elsewhere in the future to also accept List[Column]
 
 
 def table_publisher(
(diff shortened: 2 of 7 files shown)


py/server/deephaven/stream/table_publisher.py
```diff
@@ -0,0 +1,121 @@
...
+from deephaven.jcompat import j_runnable
+from deephaven.table import Table
+from deephaven.update_graph import UpdateGraph
+
+_JTableDefinition = jpy.get_type("io.deephaven.engine.table.TableDefinition")
+_JTablePublisher = jpy.get_type("io.deephaven.stream.TablePublisher")
+_JRuntimeException = jpy.get_type("java.lang.RuntimeException")
+
+
+class TablePublisher(JObjectWrapper):
```
jmao-denver, line 24: docstring missing

py/server/deephaven/stream/table_publisher.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        self.j_table_publisher.publishFailure(_JRuntimeException(str(failure)))
+
+
+# Note: we may want to expand the column_definitions typing here and elsewhere in the future to also accept List[Column]
+
+
+def table_publisher(
+    name: str,
+    column_definitions: Dict[str, DType],
```
jmao-denver, line 76:
```suggestion
    col_defs: Dict[str, DType],
```

jmao-denver (commented):
Very minor things otherwise LGTM

py/server/deephaven/stream/table_publisher.py
```diff
@@ -0,0 +1,123 @@
...
+    def __init__(self, j_table_publisher: jpy.JType):
+        self.j_table_publisher = j_table_publisher
+
+    @property
+    def j_object(self) -> jpy.JType:
+        return self.j_table_publisher
+
+    @property
+    def is_alive(self) -> bool:
+        """Checks whether this publisher is alive; if False, the publisher should stop publishing new data and release
```
chipkent, line 38: The overloading of the word "publisher" will make this sentence difficult for a user to follow.  Same for later in this docstring.

py/server/deephaven/stream/table_publisher.py
```diff
@@ -0,0 +1,123 @@
...
+
+        Once this is False, it will always remain False. For more prompt notifications, publishers may
+        prefer to use on_shutdown_callback during construction.
+
+        Returns:
+            if this publisher is alive
+        """
+        return self.j_table_publisher.isAlive()
+
+    def add(self, table: Table):
```
chipkent, line 49:
```suggestion
    def add(self, table: Table) -> None:
```
I think we have been indicating no return values.

py/server/deephaven/stream/table_publisher.py
```diff
@@ -0,0 +1,123 @@
...
+        1. consistent with a point in time
+        2. fully contained in a single blink table's update cycle
+        3. non-interleaved with any other calls to add (concurrent, or not)
+
+        Args:
+            table (Table): the table to add
+        """
+        self.j_table_publisher.add(table.j_table)
+
+    def publish_failure(self, failure: Exception):
```
chipkent, line 63:
```suggestion
    def publish_failure(self, failure: Exception) -> None:
```
I think we have been indicating no return values.

py/server/deephaven/stream/table_publisher.py
```diff
@@ -0,0 +1,123 @@
...
+        Once this is False, it will always remain False. For more prompt notifications, publishers may
+        prefer to use on_shutdown_callback during construction.
+
+        Returns:
+            if this publisher is alive
+        """
+        return self.j_table_publisher.isAlive()
+
+    def add(self, table: Table):
+        """Adds a snapshot of the data from table into the blink table according to the blink table's definition.
```
chipkent, line 50: Will not be clear to a new user.

py/server/deephaven/stream/table_publisher.py
```diff
@@ -0,0 +1,123 @@
...
+        3. non-interleaved with any other calls to add (concurrent, or not)
+
+        Args:
+            table (Table): the table to add
+        """
+        self.j_table_publisher.add(table.j_table)
+
+    def publish_failure(self, failure: Exception):
+        """Publish a failure for notification to the listeners of the blink table. Future calls to add will silently
+        return. Will cause the on-shutdown callback to be invoked if it hasn't already been invoked.
```
chipkent, line 65: I don't think this will be clear to a user.  How about:
```suggestion
        """Indicate that data publication has failed.  Blink table listeners will be notified of the failure, the on-shutdown callback will be invoked if it hasn't already been, the listener will no longer be alive, and future calls to add will silently return without publishing.
```

py/server/deephaven/stream/table_publisher.py
```diff
@@ -0,0 +1,123 @@
...
+
+        Args:
+            failure (Exception): the failure
+        """
+        self.j_table_publisher.publishFailure(_JRuntimeException(str(failure)))
+
+
+# Note: we may want to expand the column_definitions typing here and elsewhere in the future to also accept List[Column]
```
chipkent, line 73: If we need to do something, I'd prefer a ticket assigned to Jianfeng than a comment.

py/server/deephaven/stream/table_publisher.py
```diff
@@ -0,0 +1,123 @@
...
+
+    The on_shutdown_callback, if present, is called one time when the publisher should stop publishing new data
+    and release any related resources as soon as practicable since publishing won't have any downstream effects.
+
+    The update_graph is the update graph that the resulting table will belong to. If unset, the update graph of
+    the current execution context will be used.
+
+    The chunk_size is the size at which chunks will be filled from the source table during an add.
+
+    The caller is responsible for keeping a reference to the resulting blink (or derived) Table.
```
chipkent, line 93: This belongs in the args and not here.

chipkent, line 93: I would drop this line about the caller being responsible since that is generally expected in python.

py/server/deephaven/stream/table_publisher.py
chipkent: What are the requirements on table?  Does it need to identically match the schema of the table publisher?  If there are too many columns, are they ignored?  If there are too few columns, are they null filled or exception?

py/server/deephaven/stream/table_publisher.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def table_publisher(
+    name: str,
+    col_defs: Dict[str, DType],
+    on_shutdown_callback: Optional[Callable[[], None]] = None,
+    update_graph: Optional[UpdateGraph] = None,
+    chunk_size: int = 2048,
+) -> tuple[Table, TablePublisher]:
+    """Constructs a blink Table and TablePublisher.
```
chipkent, line 83:
```suggestion
    """Constructs a blink Table and TablePublisher to populate it.
```

py/server/tests/test_table_publisher.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class TablePublisherTestCase(BaseTestCase):
+    """
+    Test cases for the deephaven.stream.table_publisher module
+    """
+
+    def test_add(self):
+        my_table, my_publisher = table_publisher("test_add", table_definition())
+        self.assertTrue(isinstance(my_table, Table))
```
chipkent, line 49: could have an equivalent assertion for the publisher

chipkent (changes requested):

py/server/deephaven/stream/table_publisher.py
devinrsmith: Added https://github.com/deephaven/deephaven-core/issues/4274, will remove comment.

Integrations/src/main/java/io/deephaven/integrations/python/PythonRunnable.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.Objects;
+
+/**
+ * A {@link Runnable} implementation which calls a Python callable.
+ */
+public class PythonRunnable implements Runnable {
+    private final PyObject pyCallable;
+
+    public PythonRunnable(PyObject pyCallable) {
+        this.pyCallable = Objects.requireNonNull(pyCallable);
```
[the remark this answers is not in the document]
devinrsmith: This is somewhat a stylistic and pragmatic approach IMO - to prefer the native JDK functions whenever possible; but also to avoid further entrenching the reliance on "Base" (which I feel has grown beyond what one might call "Base").

py/server/deephaven/stream/table_publisher.py
```diff
@@ -0,0 +1,123 @@
...
+from deephaven.table import Table
+from deephaven.update_graph import UpdateGraph
+
+_JTableDefinition = jpy.get_type("io.deephaven.engine.table.TableDefinition")
+_JTablePublisher = jpy.get_type("io.deephaven.stream.TablePublisher")
+_JRuntimeException = jpy.get_type("java.lang.RuntimeException")
+
+
+class TablePublisher(JObjectWrapper):
+    """Supports publishing Table data."""
```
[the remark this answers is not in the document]
devinrsmith: Let me know what you think of the new one.

Integrations/src/main/java/io/deephaven/integrations/python/PythonRunnable.java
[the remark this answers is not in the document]
chipkent: I don't have any objections in this case.  If it was user-facing, I'd prefer them to know which param was bad.

chipkent (approved):

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/2998
Reference: https://github.com/deephaven/deephaven.io/issues/2997

Issue #4276, opened 2023-08-04
Parsing mistake in MultiJoinInput.of


rcaudy:
See `io.deephaven.engine.table.MultiJoinInput#of(io.deephaven.engine.table.Table, java.lang.String, java.lang.String)`:
```
    public static MultiJoinInput of(@NotNull final Table inputTable, String columnsToMatch, String columnsToAdd) {
        return of(inputTable,
                columnsToMatch == null || columnsToMatch.isEmpty()
                        ? Collections.emptyList()
                        : JoinMatch.from(columnsToMatch),
                columnsToAdd == null || columnsToAdd.isEmpty()
                        ? Collections.emptyList()
                        : JoinAddition.from(columnsToAdd));
    }
```

We need to split the input strings on `","` before delegating to `JoinMatch.from(String...)` and `JoinAddition.from(String...)`.

Until we fix this bug, using more than one join match or join addition column will result in an invalid column name, rather than correct operation.

PR #4279, opened 2023-08-07
MultiJoin - correct bug when multiple columns are supplied

lbooker42:
Adds missing support for parsing multiple columns in the `MultijoinInput` creation for a `MultiJoin` table.

Addresses #4276.


engine/table/src/test/java/io/deephaven/engine/table/impl/QueryTableMultiJoinTest.java
```diff
@@ -763,6 +760,31 @@ public void testColumnConflicts() {
...
+    @Test
+    public void testRenamingAndMultiColumnStrings() {
+        final Table t1 = TableTools.newTable(col("A", "a", "b"), intCol("B", 1, 2), doubleCol("C", 3.0, 4.0),
+                doubleCol("D", 10.0, 20.0));
+        final Table t2 = TableTools.newTable(col("A", "a", "b"), intCol("B", 1, 3), doubleCol("C", 5.0, 6.0),
+                doubleCol("D", 30.0, 40.0));
+
+        MultiJoinTable mj = MultiJoinTableImpl.of(
+                MultiJoinInput.of(t1, "Key1=A,Key2=B", "C1=C,D1=D"),
+                MultiJoinInput.of(t2, "Key1=A,Key2=B", "C2=C,D2=D"));
```
rcaudy, line 772: Please add cases with spaces around the column names, and "==". I believe JoinAddition.parse, JoinMatch.parse, and ColumnName.parse address these, but it would be best to verify.

rcaudy: Also, we can pretty much just verify that we got the expected multi join input. No need to run an actual MJ for all test cases of MJI.

lbooker42: I did test these manually because I was also curious but I will formalize into a test. Good point about the MJI, can be much simpler.

rcaudy (approved):

Issue #4277, opened 2023-08-04
Python Server does not reliably print output to Jupyter


chipkent:
On the demo system, I created a new Jupyter notebook that was populated with these cells:

```
import socket, errno

def is_socket_bound(address, port) -> bool:
    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)

    try:
        s.bind((address, port))
        s.close()
    except socket.error as e:
        if e.errno == errno.EADDRINUSE:
            return True

    return False
```
```
from deephaven_server import Server

def start_deephaven():
    if Server.instance is None:
        if is_socket_bound("127.0.0.1", 8080):
            raise Exception("Deephaven server is already running on this host.  Shut down all Jupyter kernels before executing this notebook.")

        _server = Server(port=8080, jvm_args=["-Xmx10g", "-DAuthHandlers=io.deephaven.auth.AnonymousAuthenticationHandler"]).start()
        print("Started Deephaven server.")
    else:
        print("Deephaven server is already running.")
```

```
start_deephaven()
```
```
print("123")
```

If I kill the kernel and run all cells, I get the following.  Note that "123" appears in the server log string instead of being printed with the appropriate line.

![image](https://github.com/deephaven/deephaven-core/assets/5250374/0e55db96-64ce-46f0-81ac-689120ff312c)

Now, if I rerun the `start_deephaven()` cell and the following cells, I get no output.  I assume it is going to the server log.

![image](https://github.com/deephaven/deephaven-core/assets/5250374/066c0de0-d909-4ef2-ba85-63370cc4cc2e)

PR #4324, opened 2023-08-15
Fix print outputs in Jupyter notebook 

jnumainville:
Fixes https://github.com/deephaven/deephaven-core/issues/4277

Essentially, the parent needs to be updated so that Jupyter's `OutputStream` can redirect the output properly
More info here: https://github.com/StanfordLegion/legion/issues/1393

Tested with code at https://github.com/deephaven/deephaven-core/issues/4277
Initial run:
Screenshot 2023-08-15 at 11 07 37 AM
rerun: 
Screenshot 2023-08-15 at 11 08 02 AM

### DIFF
--- a/py/server/deephaven_internal/stream.py
+++ b/py/server/deephaven_internal/stream.py
@@ -93,3 +93,7 @@ class TeeStream(io.TextIOBase):
     def __del__(self):
         # Do nothing, override superclass which would call close()
         pass
+
+    def set_parent(self, parent):
+        # Parent needs to be set on the original stream so that output shows in proper cell
+        self._stream.set_parent(parent)


py/server/deephaven_internal/stream.py
```diff
@@ -93,3 +93,7 @@ def close(self):
     def __del__(self):
         # Do nothing, override superclass which would call close()
         pass
+
+    def set_parent(self, parent):
+        # Parent needs to be set on the original stream so that output shows in proper cell
+        self._stream.set_parent(parent)
```
niloc132, line 99: At a glance, I'm not seeing set_parent defined in https://docs.python.org/3/library/io.html#io.TextIOBase or in the [io packages for cpython](https://github.com/python/cpython/tree/main/Modules/_io) - what is the definition for this, should we explicitly not call super (like the other overrides we provide)?

jnumainville: This is something defined by Jupyter's `OutStream`
https://github.com/ipython/ipykernel/blob/main/ipykernel/iostream.py#L508
The sys.stdout that is pulled at the time that the TeeStream is created is a Jupyter `OutStream`, so that's where the issue is.

niloc132: Sorry, I didn't read deeply enough through the linked code, see the definition in https://github.com/ipython/ipykernel/blob/6bf3fe9e44f1caf4bc371f700ac0c2e9c9c3bd84/ipykernel/iostream.py#L508-L510

I'm wondering if there are circumstances where we don't want this defined, for example to make sure we don't get something else calling `set_parent` on us if the stream that we wrap itself doesn't define `set_parent`. I'm not sure what that case would be, but some kind of intermediate "we've replaced sys.stdout too early for jupyter to have replaced it before us" situation?

I do notice that the containing type, `OutStream`, doesn't have any other supertypes to test if it is an instance, and the linked issue reports that jupyter is "graceful" in its handling if there is no set_parent method defined. Which isn't very graceful (since it leads to this issue) and seems like it could silently fail in other ways if we miss some other poorly documented bit of expected magic...

---

Backing up for a minute, the purpose of defining our own TeeStream type to begin with is so that a running server can make the full stdout available to any connecting client. @chipkent do we think that's a reasonable use case here, or if something else is mucking with stdout, should the creation of a `deephaven_server.Server` instead have a flag to be hands-off, and assume that jupyter will handle all the logs?

chipkent: I think that having the logs everywhere is the least confusing option for a user.  I think we leave it like this until we have a compelling reason to make a change.

chipkent (approved):

jnumainville:
@chipkent merged, thanks

Issue #3328, opened 2023-01-19
Table cannot be written to Parquet


rbasralian:
The following file contains a CSV of a one-column, 519k-row table that cannot be written to Parquet:

[2022-12-04_PullRequests_truncated_bad.csv.7z.zip](https://github.com/deephaven/deephaven-core/files/10453437/2022-12-04_PullRequests_truncated_bad.csv.7z.zip)

The file is 7zipped to make it sufficiently small to upload (from 2.3gb to 18mb), and then zipped again so that GitHub is happy with the file type. So to import it, you have to decompress it twice:

```shell
unzip 2022-12-04_PullRequests_truncated_bad.csv.7z.zip
7za x 2022-12-04_PullRequests_truncated_bad.csv.7z
```

(You can't read the 7zipped version directly; there is some error in `SevenZipInputStream`. It was compressed with `7za a -md=64m -mx=9` FWIW.)

Then you can read it:

```groovy
pr_body_col_broken_fromcsv = io.deephaven.csv.CsvTools.readCsv("/dh/data/shared/github-data/2022-12-04_PullRequests_truncated_bad.csv")
```

And if you try writing it to Parquet, it will fail. (Note that if you don't set it to UNCOMPRESSED, it winds up taking forever — from a cursory glance at a thread dump, it looked like it wound up calling `org.apache.parquet.hadoop.codec.SnappyCompressor#setInput` in a way that leads to a ludicrous amount of copying.)

```groovy
import io.deephaven.parquet.table.ParquetInstructions
import io.deephaven.parquet.table.ParquetTools
ParquetTools.writeTable(
    pr_body_col_broken_fromcsv,
    new File("/dh/data/shared/github-data/2022-12-04_PullRequests_bad.parquet"),
    ParquetInstructions.builder()
        .useDictionary("body", false)
        .setCompressionCodecName("UNCOMPRESSED")
        .build()
)
```

and it fails with this exception (using [this branch](https://github.com/deephaven/deephaven-core/compare/main...rbasralian:deephaven-core:raffi_parquet_write_overflow) to trigger the exception — without that change, integer overflow causes it to get stuck in a loop):
```
io.deephaven.UncheckedDeephavenException: Error writing table to /dh/data/shared/github-data/2022-12-04_PullRequests_bad.parquet
        at io.deephaven.parquet.table.ParquetTools.writeParquetTableImpl(ParquetTools.java:673)
[...]

Issue #946, opened 2021-08-01
Explore ways to accelerate Parquet table writing


rcaudy:
Currently, there's nothing at all parallel about our Parquet table writing, barring using multiple tables/files. We should investigate options here if it becomes a performance bottleneck for users.

malhotrashivam:
Potential optimization idea:
Currently, while using dictionary encoding, if we hit the limit on dictionary size, we discard all the work done so far and fall back to plain encoding for all the pages. A more optimized way would be to add a dictionary page first with the data collected so far and then use plain encoding for all the following pages.

malhotrashivam:
Some more optimization opportunities:
- In the current writing code, we copy the contents of the table into a buffer (inside TransferObject class) and then use that buffer for writing to the parquet file (inside ColumnWriter class). We can skip this intermediate step of creating the buffer and directly write to the parquet file. An example for long type column is here: #4587.
- As found during #4541, when writing identical content, pyarrow generally writes fewer number of pages per file than our code. This can lead to performance benefits since writing each page comes with additional metadata. We can look closely on how pyarrow decides on page boundaries and mimic that in our core code.

malhotrashivam:
The first and third suggestion above were tried and didn't show much improvement. More details can be found [on this doc](https://docs.google.com/document/d/1Y8iQzI6yIND348NuX6RqowSO5XzI4p-Hk3mAHAShCKA/edit#heading=h.fyueu9l6iux8).

PR #4334, opened 2023-08-17
Introduce a limit on maximum dictionary size for a parquet file

malhotrashivam:
Partially fixes issue #3328 

Our parquet writing code uses dictionary encoding for string columns.
Quoting from the [Parquet docs](https://parquet.apache.org/docs/file-format/data-pages/encodings/): "If the dictionary grows too big, whether in size or number of distinct values, the encoding will fall back to the plain encoding."

In our parquet writing code, there already exists a limit on number of keys in the dictionary for a single row group. This change introduces a limit on number of bytes. If we hit either of these limits, the writer will fall back to plain encoding.


extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -683,16 +685,18 @@ private static <DATA_TYPE> boolean tryEncodeDictionary(
...
-
+                                    // Copy into an array of double the size with upper limit at maxKeys
                                     encodedKeys = Arrays.copyOf(encodedKeys, (int) Math.min(keyCount * 2L, maxKeys));
                                 }
                                 encodedKeys[keyCount] = Binary.fromString(key);
                                 dictionaryPos = keyCount;
+                                dictSize += encodedKeys[keyCount].length();
                                 keyCount++;
+                                if ((keyCount >= maxKeys) || (dictSize >= maxDictSize)) {
+                                    throw new DictionarySizeExceededException(
```
lbooker42, line 696: We are discarding a lot of work assembling a dictionary and encoded chunks when we error and exit this function.  Our own reader can read hybrid columns (both RLE_DICTIONARY and PLAIN). If this does not violate an acceptable standard, I think we should try to leverage this work.

chipkent (approved):

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
malhotrashivam: Thanks for pointing this out, I have noted it as a comment on #946 and added a comment in the code.

chipkent (approved):

PR #4339, opened 2023-08-18
Added statistics tracking to DH parquet table writing.

### LINKED ISSUE #949: Parquet: Write encoding stats
`io.deephaven.parquet.ColumnChunkReaderImpl#usesDictionaryOnEveryPage` returns `false` for all columns in Deephaven-written Parquet data, since we don't write any encoding stats:
```
        final ColumnMetaData columnMeta = columnChunk.getMeta_data();
        if (columnMeta.encoding_stats == null) {
            // We don't know, so we bail out to "false"
            return false;
        }
```
We should fix this, so that our files can take advantage of the same optimizations we allow for other sources.

lbooker42:
This PR adds tracking of min/max and num_nulls for columns in row groups written by the Deephaven parquet writer.

Currently, vector columns are ignored and will not produce statistics.

Will close #949 when merged.


extensions/parquet/base/src/main/java/io/deephaven/parquet/base/BulkWriter.java
```diff
@@ -81,7 +84,8 @@ int writeBulkVector(@NotNull final BUFFER_TYPE bulkValues,
      * @return a {@link WriteResult} containing the statistics of the result.
      */
     @NotNull
-    WriteResult writeBulkFilterNulls(@NotNull BUFFER_TYPE bulkValues, int rowCount);
+    WriteResult writeBulkFilterNulls(@NotNull BUFFER_TYPE bulkValues,
```
malhotrashivam, line 87: Just for clarity, do you think we can rename this method
```suggestion
    WriteResult writeBulkVectorFilterNulls(@NotNull BUFFER_TYPE bulkValues,
```

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnWriter.java
```diff
@@ -4,16 +4,58 @@
...
+import io.deephaven.util.annotations.FinalDefault;
+import org.apache.parquet.column.statistics.Statistics;
 
 import java.io.IOException;
 import java.nio.IntBuffer;
 
 public interface ColumnWriter extends SafeCloseable {
-    void addPageNoNulls(Object pageData, int valuesCount) throws IOException;
+    /**
+     * Add a page with no nulls to the file. Does not track statistics, so the caller must track statistics in another
```
malhotrashivam, line 15: Thanks for adding these comments

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnWriterImpl.java
```diff
@@ -84,13 +86,15 @@ public class ColumnWriterImpl implements ColumnWriter {
                         getWidthFromMaxInt(column.getMaxRepetitionLevel()), MIN_SLAB_SIZE, targetPageSize, allocator);
         this.owner = owner;
         offsetIndexBuilder = OffsetIndexBuilder.getBuilder();
+        statistics = Statistics.createStats(column.getPrimitiveType());
```
malhotrashivam, line 89: Make sure to test it out with all the types, and compare the statistics with pyarrow. There are examples in ParquetTableReadWriteTest.java and test_parquet.py which go over different data types.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/RleIntChunkedWriter.java
```diff
@@ -95,22 +97,32 @@ public String memUsageString(String prefix) {
...
-
+    public void writeBulk(@NotNull IntBuffer bulkValues,
+                          final int rowCount,
+                          @Nullable final Statistics<?> statistics) {
+        // Track statistics while we write the values.
         for (int i = 0; i < rowCount; i++) {
-            writeInteger(bulkValues.get());
+            final int v = bulkValues.get();
+            writeInteger(v);
+            statistics.updateStats(v);
```
malhotrashivam, line 107: From what I understood, if we use dictionary encoding, the statistics will be collected while building the dictionary in ParquetTableWriter. If we use plain encoding, they will be populated here. 

Is there any reason why we don't want to keep statistics logic just in the ChunkedWriters? Is it because the data that reaches RleIntChunkedWriter is actually indexes into the dictionary and not actual values? If that is the reason, do we need statistics counting logic in RleIntChunkedWriter?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -727,18 +742,24 @@ private static <DATA_TYPE> boolean tryEncodeDictionary(
 
             columnWriter.addDictionaryPage(encodedKeys, keyCount);
             final Iterator<IntBuffer> arraySizeIt = arraySizeBuffers == null ? null : arraySizeBuffers.iterator();
-            for (final IntBuffer pageBuffer : pageBuffers) {
+            for (int i = 0; i < pageBuffers.size(); ++i) {
+                final IntBuffer pageBuffer = pageBuffers.get(i);
+                final boolean pageHasNulls = pageBufferHasNull.get(i);
                 pageBuffer.flip();
                 if (vectorHelper != null) {
```
malhotrashivam, line 749: So we are not tracking stats in any of these cases, right?
If that is the case, the comments make it seem that we are tracking in the vector case.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/RleIntChunkedWriter.java
lbooker42: I may misunderstand, but RleIntChunkedWriter is currently used by DHC only for dictionary values, but it can also be used for any integer column as well.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/BulkWriter.java
malhotrashivam: Thanks

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/RleIntChunkedWriter.java
malhotrashivam: Yea makes sense. Thanks for explaining.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -742,20 +740,22 @@ private static <DATA_TYPE> boolean tryEncodeDictionary(
 
             columnWriter.addDictionaryPage(encodedKeys, keyCount);
             final Iterator<IntBuffer> arraySizeIt = arraySizeBuffers == null ? null : arraySizeBuffers.iterator();
+            // Accumulate integer statistics in a temporary object.
```
malhotrashivam, line 743: I am assuming this is needed because we are generating stats over the keys here, mainly to get stats of the nulls, which we then add to overall stats.
Is that right?
Can you please add a more descriptive comment to explain that what we mean by integer statistics here?

malhotrashivam (approved):

PR #4343, opened 2023-08-18
Make cummax consistent with cummin

chipkent:
Make `cummax` consistent with `cummin`.  `cummax` did not handle nulls properly.

### DIFF
--- a/engine/function/src/templates/Numeric.ftl
+++ b/engine/function/src/templates/Numeric.ftl
@@ -1571,15 +1571,30 @@ public class Numeric {
             return null;
         }
 
-        if (values.length == 0) {
+        if (values.isEmpty()) {
             return new ${pt.primitive}[0];
         }
 
-        ${pt.primitive}[] result = new ${pt.primitive}[values.length];
-        result[0] = values[0];
+        final int n = values.intSize("cummax");
+        ${pt.primitive}[] result = new ${pt.primitive}[n];
 
-        for (int i = 1; i < values.length; i++) {
-            result[i] = compare(result[i-1], values[i]) > 0 ? result[i-1] : values[i];
+        try ( final ${pt.vectorIterator} vi = values.iterator() ) {
+            result[0] = vi.${pt.iteratorNext}();
+            int i = 1;
+
+            while (vi.hasNext()) {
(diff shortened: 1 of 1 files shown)


engine/function/src/templates/Numeric.ftl
```diff
@@ -1579,7 +1579,13 @@ public class Numeric {
         result[0] = values[0];
 
         for (int i = 1; i < values.length; i++) {
-            result[i] = compare(result[i-1], values[i]) > 0 ? result[i-1] : values[i];
+            if (isNull(result[i - 1])) {
+                result[i] = values[i];
+            } else if (isNull(values[i])) {
+                result[i] = result[i - 1];
+            } else {
+                result[i] = (${pt.primitive})Math.max(result[i - 1],  values[i]);
```
kosak, line 1587: Can we remove the extra space before values[i] at the end

engine/function/src/templates/Numeric.ftl
```diff
@@ -1606,10 +1612,18 @@ public class Numeric {
...
             while (vi.hasNext()) {
                 final ${pt.primitive} v = vi.${pt.iteratorNext}();
-                result[i] = compare(result[i-1], v) > 0 ? result[i-1] : v;
+
+                if (isNull(result[i - 1])) {
+                    result[i] = v;
+                } else if (isNull(v)) {
+                    result[i] = result[i - 1];
+                } else {
+                    result[i] = (${pt.primitive})Math.max(result[i - 1],  v);
```
kosak, line 1624: Can we remove the extra space before v at the end

engine/function/src/templates/Numeric.ftl
chipkent:
```suggestion
                result[i] = (${pt.primitive})Math.max(result[i - 1], values[i]);
```

engine/function/src/templates/Numeric.ftl
chipkent:
```suggestion
                    result[i] = (${pt.primitive})Math.max(result[i - 1], v);
```

kosak (approved):

PR #4344, opened 2023-08-18
Enforce page size limit in Parquet 

malhotrashivam:
For string columns and columns which use codecs, we don't enfoce page size limits properly. This was found while debugging #3328. The PR fixes the this issue. Also, this PR has some code refactoring to remove code duplication and make the code more readable.


extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -89,13 +89,32 @@ enum CacheTags {
      * @param <B>
      */
     interface TransferObject<B> extends SafeCloseable {
-        void propagateChunkData();
-
-        B getBuffer();
+        /**
```
lbooker42, line 92: Love the comments, thank you!

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -842,18 +868,30 @@ static class PrimitiveTransfer<C extends WritableChunk<Values>, B extends Buffer
...
         private final ColumnSource<?> columnSource;
         private final ChunkSource.FillContext context;
+        private boolean hasMoreDataToBuffer;
 
         PrimitiveTransfer(ColumnSource<?> columnSource, C chunk, B buffer, int targetSize) {
             this.columnSource = columnSource;
             this.chunk = chunk;
             this.buffer = buffer;
             context = columnSource.makeFillContext(targetSize);
+            hasMoreDataToBuffer = false;
```
lbooker42, line 878: We typically only initialize non-default values.  `hasMoreDataToBuffer` is `false` by default.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -1016,22 +1091,66 @@ static class StringTransfer implements TransferObject<Binary[]> {
...
+        /**
+         * Stores the maximum size of data in a single page. For strings, we don't know in advance what the length of
+         * each entry would be, therefore we need to keep track of both maximum size and number of values.
+         */
+        private final int targetPageSize;
+
+        /**
+         * Total number of chunks buffered so far
+         */
+        private int numChunksBuffered;
```
lbooker42, line 1111: This name doesn't seem correct.  Isn't this a chunk index?  maybe `currentChunkIdx` is more clear?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -1016,22 +1091,66 @@ static class StringTransfer implements TransferObject<Binary[]> {
...
-                String value = chunk.get(i);
-                buffer[i] = value == null ? null : Binary.fromString(value);
+        public boolean hasMoreDataToBuffer() {
+            return ((chunk != null) && (numChunksBuffered < chunk.size()));
+        }
+
+        public int bufferFetchedData() {
+            if (numChunksBuffered != 0) {
+                // Clear any old data in the buffer from previous iterations
+                buffer = new Binary[maxValuesPerPage];
```
lbooker42, line 1131: Do you keep around a reference to the old `buffer`? If not, can reuse this array and just reset the value.
```suggestion
                Arrays.fill(buffer, null);
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -1016,22 +1091,66 @@ static class StringTransfer implements TransferObject<Binary[]> {
...
+            int bufferedDataSize = 0;
+            int bufferIdx = 0; // Stores the number of chunks buffered in this iteration
+            while (numChunksBuffered < chunk.size()) {
+                final String value = chunk.get(numChunksBuffered);
+                if (value == null) {
+                    buffer[bufferIdx++] = null;
+                } else {
+                    Binary binaryVal = Binary.fromString(value);
+                    if (binaryVal.length() > targetPageSize) {
+                        throw new UncheckedDeephavenException(String.format("Cannot fit string data of size %d in a " +
```
lbooker42, line 1142: Do we need to throw an exception or can we write the page and continue? Also, consider the case where we have a single string >= targetPageSize, shouldn't we write an abnormal page size and continue writing?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -1063,21 +1183,32 @@ static class CodecTransfer<T> implements TransferObject<Binary[]> {
...
 
         @Override
-        public void propagateChunkData() {
+        public boolean hasMoreDataToBuffer() {
+            return hasMoreDataToBuffer;
+        }
+
+        @Override
+        public int bufferFetchedData() {
+            if (!hasMoreDataToBuffer) {
```
lbooker42, line 1203: The binaries generated from the codec also are variable length and would benefit from the page size tests (similar to String).

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -1016,22 +1091,66 @@ static class StringTransfer implements TransferObject<Binary[]> {
...
+                if (value == null) {
+                    buffer[bufferIdx++] = null;
+                } else {
+                    Binary binaryVal = Binary.fromString(value);
+                    if (binaryVal.length() > targetPageSize) {
+                        throw new UncheckedDeephavenException(String.format("Cannot fit string data of size %d in a " +
+                                "single page of maximum size %d", value.length(), targetPageSize));
+                    }
+                    if (bufferedDataSize + binaryVal.length() > targetPageSize) {
+                        break;
```
lbooker42, line 1146: As you noted in our call, this termination  may cause truncation when a vector of strings is to be written  that exceeds the target page size.  We need to handle vectors/arrays of strings differently and be accept potential OOM and oversize pages.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -642,16 +646,14 @@ private static <DATA_TYPE> void encodePlain(@NotNull final ParquetInstructions w
                                 .asIntChunk();
                         lenChunk.copyToTypedBuffer(0, repeatCount, 0, lenChunk.size());
                         repeatCount.limit(lenChunk.size());
-                        transferObject.bufferFetchedData();
-                        // Return value ignored, this can lead to errors in case we are not able to fit all the data in
-                        // a single page, which is possible for vector<string> columns.
-                        columnWriter.addVectorPage(transferObject.getBuffer(), repeatCount, transferObject.rowCount());
+                        int numValuesBuffered = transferObject.bufferAllFetchedData();
```
lbooker42, line 649: I think I would leave a comment here explaining why you must grab the entire vector (because it must be written in a single page, etc.) and there is a risk of OOM exceptions.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -1087,150 +1089,118 @@ public void close() {
...
-
         /**
          * Stores the maximum size of data in a single page. For strings, we don't know in advance what the length of
          * each entry would be, therefore we need to keep track of both maximum size and number of values.
          */
         private final int targetPageSize;
 
         /**
-         * Total number of chunks buffered so far
+         * Index of next string to be buffered from the string chunk
```
lbooker42, line 1109: This is a generic class now, so `String` references should go away in favor of something generic.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -1087,150 +1089,118 @@ public void close() {
         }
     }
 
-    static class StringTransfer implements TransferObject<Binary[]> {
-
+    /**
+     * Used as a base class for transfer object types like strings or big integers that need specialized encoding or
+     * codecs.
+     */
+    abstract static class EncodedTransfer<T> implements TransferObject<Binary[]> {
```
lbooker42, line 1096: This abstraction is excellent.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -875,14 +878,18 @@ static class PrimitiveTransfer<C extends WritableChunk<Values>, B extends Buffer
...
             context = columnSource.makeFillContext(targetSize);
-            hasMoreDataToBuffer = false;
         }
 
         @Override
         public boolean hasMoreDataToBuffer() {
             return hasMoreDataToBuffer;
         }
 
+        @Override
```
lbooker42, line 888: Instead of implementing this for each class, you **could** make it a `default` method on the `TransferObject` 
class. `EncodedTransfer` is the only one that actually needs a custom implementation.

What you have is fine, however, and might be better.  default methods can get you in trouble sometimes and we probably need to think carefully about this function with any new implementation of this class.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
malhotrashivam: Nice catch, thanks!

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
malhotrashivam: I wanted to do that, but like you explained, defaults were making me uncomfortable because part of the class needs to be specialized for each type and some parts can be default. So I kept it like this to make it easier to read instead of splitting across base and derived types.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
malhotrashivam: Thanks!

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -920,20 +919,22 @@ public void close() {
         }
     }
 
-    static class ShortTransfer implements TransferObject<IntBuffer> {
+    /**
+     * Used as a base class of transfer objects for types where we don't enforce any page size limits
+     */
+    abstract static class BaseTransfer<T extends ChunkBase<Values>> implements TransferObject<IntBuffer> {
```
lbooker42, line 925: I don't think we need TransferObject and BaseTransfer.  You can make TransferObject an abstract class instead of an interface and move this common functionality to it.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -952,14 +953,16 @@ public int bufferFetchedData() {
                 return 0;
             }
             buffer.clear();
-            for (int i = 0; i < chunk.size(); i++) {
-                buffer.put(chunk.get(i));
+            for (int chunkIdx = 0; chunkIdx < chunk.size(); chunkIdx++) {
```
lbooker42, line 956: I think you'll get better performance if you make this a bulk transfer function.  Instead of making `chunk.size()` virtual calls (with the virtualization overhead), you can make a single call and pass in the size.

lbooker42 (dismissed):

lbooker42 (approved):

chipkent (approved):

lbooker42 (dismissed):

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -790,7 +821,9 @@ private static int getTargetRowsPerPage(@NotNull final Class<?> columnType,
         }
 
         if (columnType == String.class) {
-            return targetPageSize / Integer.BYTES;
+            // We don't know the length of strings till we read the actual data. Therefore, we take a relaxed estimate
```
chipkent, line 824:
```suggestion
            // We don't know the length of strings untill we read the actual data. Therefore, we take a relaxed estimate
```

chipkent (changes requested):

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
malhotrashivam: Thanks for the comment.

chipkent (dismissed):
LGTM

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -90,13 +90,37 @@ enum CacheTags {
...
-        int rowCount();
+        /**
+         * Copy the fetched data into an internal buffer, which can then be accessed using
+         * {@link TransferObject#getBuffer()}. The method should only be called after
+         * {@link TransferObject#fetchData(RowSequence)}}
+         *
+         * @return The number of fetched data entries copied into the buffer. This can be different from the number of
+         *         entries fetched for types like strings where we enforce additional page size limits while copying.
+         */
+        int bufferFetchedData();
```
rcaudy, line 116: I recommend a clearer name:
`transferOnePageToBuffer`
And a re-write to the JavaDoc to reflect that this is meant to be used when writing paginated data, and not meant to be interleaved with calls to the other method.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -90,13 +90,37 @@ enum CacheTags {
...
-        B getBuffer();
+        /**
+         * Copy all the fetched data into an internal buffer, which can then be accessed using
+         * {@link TransferObject#getBuffer()}. This method should only be called after
+         * {@link TransferObject#fetchData(RowSequence)}}. Note that this method can lead to Out-of-Memory error for
+         * types like strings if the fetched data is too big.
+         *
+         * @return The number of fetched data entries copied into the buffer.
+         */
+        int bufferAllFetchedData();
```
rcaudy, line 106: I recommend a clearer name:
transferAllToBuffer
And a re-write to the JavaDoc to reflect that this is meant to be used when writing unpaginated data, and not meant to be interleaved with calls to the other method.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
malhotrashivam: I didn't paginate vector data because the implementation was becoming unnecessarily complicated in cases where we might run out of our page limit in the middle of the vector. 

- One approach is to pull out all elements of the vector, find the length in binary and then add it up to check if we are hitting the limit and only then transfer to buffer.
- Other way is to consider the limit in a relaxed manner and always put all elements of a single vector, and if we have hit or exceeded the limit, then stop.

Let me know if you want me to add this logic.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -903,225 +949,223 @@ public void close() {
...
+                if (value == null) {
+                    buffer[bufferIdx++] = null;
+                    continue;
+                }
+                Binary binaryVal = encodeToBinary(value);
+                buffer[bufferIdx++] = binaryVal;
+                size += binaryVal.length();
+                if (size > targetPageSize) {
+                    break;
+                }
```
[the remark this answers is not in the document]
malhotrashivam: I agree with the >= comment.

For the optimisitic encoding, I tried that approach (it might be in one of the old commits) but then I was looking at how pyarrow does it and they are even more relaxed with the limit. They put elements in chunks of 100 in a page and then check if they have exceeded the limit. So I thought of keeping the implementation simpler by checking after buffering.
I can revert to the checking before buffering approach if you think that would be more appropriate.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -90,13 +90,37 @@ enum CacheTags {
      * @param <B>
      */
     interface TransferObject<B> extends SafeCloseable {
```
malhotrashivam, line 92: Once we complete the review and before merging, I am thinking of reorganizing this file by moving `TransferObject` closer to its child classes and reordering all methods in child classes to be in same order as this interface.

rcaudy: Sounds great. Do it as a single, isolated commit, please.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -903,225 +949,223 @@ public void close() {
...
+        @Override
+        final public int bufferAllFetchedData() {
+            if (!hasMoreDataToBuffer()) {
+                return 0;
+            }
+            while (currentChunkIdx < chunk.size()) {
+                final T value = chunk.get(currentChunkIdx);
+                buffer[currentChunkIdx++] = value == null ? null : encodeToBinary(value);
             }
+            return chunk.size();
```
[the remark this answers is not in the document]
rcaudy: I'm happy enough if the assumption is just declared and documented.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
rcaudy: There's complexity at the layer above, too.
See `io.deephaven.parquet.table.pagestore.topage.ToVectorPage`
and `io.deephaven.parquet.table.pagestore.topage.ToArrayPage`

rcaudy: I think all we really want here is a comment that explains the problem (our code expecting all elements from an array or vector being on the same page, and a reference to that code) with a TODO referencing a new ticket that goes on the backlog.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
rcaudy: Try not to do anything that will introduce even more conflicts with https://github.com/deephaven/deephaven-core/pull/4479

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -949,6 +958,59 @@ public void close() {
         }
     }
 
+    final static class IntTransfer extends PrimitiveTransfer<WritableIntChunk<Values>, IntBuffer> {
+        /**
+         * Check docs in {@link PrimitiveTransfer} for more details.
+         */
+        IntTransfer(ColumnSource<?> columnSource, WritableIntChunk<Values> chunk, IntBuffer buffer, int targetSize) {
+            super(columnSource, chunk, buffer, targetSize);
+            assert (Arrays.equals(chunk.array(), buffer.array()));
```
rcaudy, line 967: Prefer `io.deephaven.base.verify.Assert#eq(java.lang.Object, java.lang.String, java.lang.Object, java.lang.String)`

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -949,6 +958,59 @@ public void close() {
         }
     }
 
+    final static class IntTransfer extends PrimitiveTransfer<WritableIntChunk<Values>, IntBuffer> {
```
rcaudy, line 961:
```suggestion
    private static final class IntTransfer extends PrimitiveTransfer<WritableIntChunk<Values>, IntBuffer> {
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -1083,45 +1154,45 @@ final public boolean hasMoreDataToBuffer() {
...
-        final public int bufferAllFetchedData() {
-            if (!hasMoreDataToBuffer()) {
-                return 0;
-            }
-            while (currentChunkIdx < chunk.size()) {
-                final T value = chunk.get(currentChunkIdx);
-                buffer[currentChunkIdx++] = value == null ? null : encodeToBinary(value);
+        final public int transferAllToBuffer() {
+            // Assuming this method is called after fetchData() and that the buffer is empty.
+            assert (chunk != null && currentChunkIdx == 0 && bufferCount == 0);
```
rcaudy, line 1159: `Assert.neqNull` and `Assert.eqZero` or `Assert.assertion`

py/server/tests/test_parquet.py
```diff
@@ -286,7 +286,7 @@ def round_trip_with_compression(self, compression_codec_name, dh_table, vector_c
 
         # Write the pandas dataframe back to parquet (via pyarraow) and read it back using deephaven.parquet to compare
         dataframe.to_parquet('data_from_pandas.parquet',
-                             compression=None if compression_codec_name is 'UNCOMPRESSED' else compression_codec_name)
+                             compression=None if compression_codec_name == 'UNCOMPRESSED' else compression_codec_name)
```
rcaudy, line 289: Ask @jmao-denver why this doesn't work and to approve as a Python code owner.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -903,225 +949,223 @@ public void close() {
...
+                }
+                Binary binaryVal = encodeToBinary(value);
+                buffer[bufferIdx++] = binaryVal;
+                size += binaryVal.length();
+                if (size > targetPageSize) {
+                    break;
+                }
+            }
+            if (currentChunkIdx == chunk.size()) {
+                chunk = null;
```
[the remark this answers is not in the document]
rcaudy: I was just suggesting to set `chunk == null` in the other method on this class.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetInstructions.java
```diff
@@ -445,7 +445,8 @@ public static class Builder {
         private int maximumDictionaryKeys = defaultMaximumDictionaryKeys;
         private int maximumDictionarySize = defaultMaximumDictionarySize;
         private boolean isLegacyParquet;
-        private int targetPageSize = defaultTargetPageSize;
+        @VisibleForTesting
+        int targetPageSize = defaultTargetPageSize;
```
rcaudy, line 449: I see why you did this, but it's a little unfortunate. Is the default inappropriately high? Should we make it configurable?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -101,7 +102,7 @@ interface TransferObject<B> extends SafeCloseable {
          * {@link TransferObject#fetchData(RowSequence)}}. This method should be used when writing unpaginated data, and
          * should not be interleaved with calls to {@link TransferObject#transferOnePageToBuffer()}. Note that this
          * method can lead to out-of-memory error for variable-width types (e.g. strings) if the fetched data is too
-         * big.
+         * long.
```
rcaudy, line 105: "long" -> "too big to fit in the available heap"?

malhotrashivam: Thank you, this is much better :)

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -87,6 +88,9 @@ public void setUp() {
         }
         // noinspection ResultOfMethodCallIgnored
         rootFile.mkdirs();
+
+        // Allow a small page size for testing
+        Configuration.getInstance().setProperty("Parquet.minTargetPageSize", "64");
```
rcaudy, line 93: This isn't going to work reliably. If `ParquetInstructions` has been statically initialized, this will be ineffective. Why not just amend the tests below to set `MIN_TARGET_PAGE_SIZE` as the default, and try to write pages bigger than that?

malhotrashivam: Will do!

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -856,38 +853,40 @@ public void dictionaryEncodingTest() {
     @Test
     public void overflowingStringsTest() {
         // Test the behavior of writing parquet files if entries exceed the page size limit
+        final int pageSize = 2 << 10;
```
rcaudy, line 856: I might have said to explicitly reference the constant, but this is fine.

rcaudy (approved):

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
malhotrashivam: Will make the change in another PR

PR #4352, opened 2023-08-21
Option to ingest Kafka String columns in Avro Utf8 format

devonpallison:


.github/workflows/cla.yml
devinrsmith, line 1: you should not be modifying this file.

extensions/kafka/src/main/java/io/deephaven/kafka/AvroImpl.java
```diff
@@ -68,6 +69,8 @@ static final class AvroConsume extends Consume.KeyOrValueSpec {
         /** fields mapped to null are skipped. */
         private final Function<String, String> fieldPathToColumnName;
 
+        boolean useUTF8Strings = false;
```
devinrsmith, line 72: should be private, final.

devonpallison:

extensions/kafka/src/main/java/io/deephaven/kafka/AvroImpl.java
```diff
@@ -453,7 +474,11 @@ private static void pushColumnTypesFromAvroField(
                 break;
             case ENUM:
             case STRING:
-                columnsOut.add(ColumnDefinition.ofString(mappedNameForColumn));
+                if (useUTF8Strings) {
+                    columnsOut.add(ColumnDefinition.of(mappedNameForColumn, Type.find(CharSequence.class)));
```
devinrsmith, line 478: We should probably use the more specific type, `org.apache.avro.util.Utf8`. Also, can store the type as a constant.

extensions/kafka/src/main/java/io/deephaven/kafka/AvroImpl.java
```diff
@@ -84,6 +87,17 @@ static final class AvroConsume extends Consume.KeyOrValueSpec {
             this.fieldPathToColumnName = fieldPathToColumnName;
         }
 
+        private AvroConsume(final String schemaName,
```
devinrsmith, line 90: The other constructor can delegate to this one.

extensions/kafka/src/main/java/io/deephaven/kafka/AvroImpl.java
devinrsmith: We'll also want to be very careful that the avro deserializer doesn't pool / keep hold of the `Utf8` objects; the javadoc explicitly states it's mutable:

```java
/**
 * A Utf8 string. Unlike {@link String}, instances are mutable. This is more
 * efficient than {@link String} when reading or writing a sequence of values,
 * as a single instance may be reused.
 */
```

devinrsmith (changes requested):
In favor of this change, seems good. Couple of finishing points need to be addressed first.

devinrsmith:
recheck

devinrsmith (commented):
Does this provide DHE w/ enough hooks to actually use utf8?

I see `KafkaTools.avroSchemaToColumnDefinitions`, but no tie-ins w/  `KafkaTools.avroSpec`. Even though you've added `io.deephaven.kafka.AvroImpl.AvroConsume#useUTF8Strings`, that's a package private class, not publicly visible.

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -483,6 +493,95 @@ public static KeyOrValueSpec avroSpec(final String schemaName) {
...
+         * @param fieldNameToColumnName A mapping specifying which Avro fields to include and what column name to use
+         *        for them; fields mapped to null are excluded.
+         * @return A spec corresponding to the schema provided.
+         */
+        @SuppressWarnings("unused")
+        public static KeyOrValueSpec avroSpecWithUTF8Strings(final String schemaName,
+                final String schemaVersion,
+                final Function<String, String> fieldNameToColumnName) {
+            return new AvroConsume(schemaName, schemaVersion, fieldNameToColumnName, true);
+        }
```
devinrsmith, line 538: Not a big fan of the proliferation of `WithUTF8Strings` messages; arguably, we may want to follow more of a builder pattern for cases like this in the future.

What about creating only having a single new method `avroSpec(String, String, Function, boolean)` that mirrors the new AvroConsume constructor?

devonpallison: Sure. Should be sufficient.

devinrsmith (approved):

PR #4353, opened 2023-08-21
R6 rewrite

alexpeters1208:
This PR implements the consensus of the discussion around the R API, where we decided to move back to an R6 API with S3 implementations for base-R generics where sensible. Documentation coming in the next PR.


R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,275 +1,208 @@
...
+    #' a variable name on the server. See `?TableHandle` for more information.
+    #' @param table_object An R Data Frame, a dplyr Tibble, an Arrow Table, or an Arrow RecordBatchReader
+    #' containing the data to import to the server.
+    #' @return TableHandle reference to the new table.
+    import_table = function(table_object) {
+      table_object_class <- class(table_object)
+      if (table_object_class[[1]] == "data.frame") {
+        return(TableHandle$new(private$df_to_dh_table(table_object)))
+      }
+      if (table_object_class[[1]] == "tbl_df") {
```
chipkent, line 155:
```suggestion
      } else if (table_object_class[[1]] == "tbl_df") {
```

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -1,147 +1,322 @@
...
+    #' Whether the table referenced by this TableHandle is static or not.
+    #' @return TRUE if the table is static, or FALSE if the table is ticking.
+    is_static = function() {
+      return(self$.internal_rcpp_object$is_static())
+    },
+    
+    #' @description
+    #' Binds the table referenced by this TableHandle to a variable on the server,
+    #' enabling it to be accessed by that name from any Deephaven API.
+    #' @param name Name for this table on the server.
```
chipkent, line 26: I'm not reviewing the doc strings, as you suggested.  I would remove them so that they all actually get reviewed in the docs pr.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -1,147 +1,322 @@
...
+    #' Imports the table referenced by this TableHandle into an R Data Frame.
+    #' @return A Data Frame containing the data from the table referenced by this TableHandle.
+    as_data_frame = function() {
+      arrow_tbl <- self$as_arrow_table()
+      return(as.data.frame(as.data.frame(arrow_tbl))) # TODO: for some reason as.data.frame on arrow table returns a tibble, not a data frame
+    },
+    
+    ### DEEPHAVEN TABLE OPERATIONS
+    
+    select = function(by = character()) {
```
chipkent, line 98: All of these functions need to get their arg names checked against the python names.  For example, `select` should not have `by` args.  Should check all functions, in case I just didn't notice elsewhere.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -1,147 +1,322 @@
...
+    merge = function(...) {
+      table_list <- unlist(c(...))
+      if (length(table_list) == 0) {
+        return(TableHandle$new(self$.internal_rcpp_object$merge(NULL)))
+      }
+      verify_type("table_list", table_list, "TableHandle", "Deephaven TableHandle", FALSE)
+      unwrapped_table_list <- lapply(table_list, strip_r6_wrapping)
+      return(TableHandle$new(self$.internal_rcpp_object$merge(unwrapped_table_list)))
+    }
+  )
```
chipkent, line 275: I can go either way about having `merge` attached to the table handle.  We must have the function `merge` not associated with a table handle.  If this is kept in, confirm with works with null inputs.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -1,147 +1,322 @@
...
+      if ((length(abs_sort) > 1) && length(abs_sort) != length(by)) {
+        stop(paste0("'abs_sort' must be the same length as 'by' if more than one entry is supplied. Got 'by' with length ", length(by), " and 'abs_sort' with length ", length(abs_sort), "."))
+      }
+      return(TableHandle$new(self$.internal_rcpp_object$sort(by, descending, abs_sort)))
+    },
+    
+    merge = function(...) {
+      table_list <- unlist(c(...))
+      if (length(table_list) == 0) {
+        return(TableHandle$new(self$.internal_rcpp_object$merge(NULL)))
```
chipkent, line 269: COuldn't this just return self without having to do a round trip to the server?

chipkent (changes requested):

chipkent (changes requested):

R/rdeephaven/R/table_handle_wrapper.R
alexpeters1208: Ehhh.. Right now these R6 objects are set to not be clonable, and I'm not sure I want to return a reference to the same object, if that's possible in R. I can make R6 clonable if the C++ TableHandle has a copy constructor, which I'm sure it does. Not sure what the implications on efficiency there would be, in terms of cloning the whole thing vs a server call.

R/rdeephaven/R/table_handle_wrapper.R
alexpeters1208: I handled them all except for the argument names in the join functions, which I find to be far more informative than their Python analogs. `right_side`, `columns_to_match`, and `columns_to_add` in R, versus `table`, `on`, and `joins` in Python. I can change them if needed.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -117,22 +112,20 @@ Client <- R6Class("Client",
...
       return(TableHandle$new(self$.internal_rcpp_object$empty_table(size)))
     },
-    
+
     time_table = function(period, start_time = "now") {
       verify_string("period", period, TRUE)
-      verify_string("start_time", start_time, TRUE)
+      if(any(start_time != "now") && any(start_time != "")) {
+        verify_string("start_time", start_time, TRUE)
+      }
```
chipkent, line 125: I don't understand this change.  Shouldn't you be verifying that it is a string no matter what?  Then the whole "now" or "" thing is not making sense to me.

R/rdeephaven/R/helper_functions.R
```diff
@@ -3,7 +3,11 @@ first_class <- function(arg) {
 }
 
 verify_type <- function(arg_name, candidate, required_type, message_type_name, is_scalar) {
-  if (!is_scalar && (first_class(candidate) == "list")) {
+  if (required_type == "list") {
+    if (first_class(candidate) != "list") {
+      stop(paste0("'", arg_name, "' must be a list or a vector of lists. Got an object of class ", first_class(candidate), "."))
```
chipkent, line 8:
```suggestion
      stop(paste0("'", arg_name, "' must be a list. Got an object of class ", first_class(candidate), "."))
```
Is the removed text always appropriate?

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -56,6 +56,16 @@ TableHandle <- R6Class("TableHandle",
       return(c(self$nrow(), self$ncol()))
     },
     
+    merge = function(...) {
+      table_list <- unlist(c(...))
+      if (length(table_list) == 0) {
+        return(TableHandle$new(self$.internal_rcpp_object$merge(NULL)))
```
chipkent, line 62: Seems like this should just return self.

chipkent (changes requested):

R/rdeephaven/R/helper_functions.R
alexpeters1208: I believe so. As far as R is concerned, there is no difference between `list(a = "val", b = "val")` and `c(list(a = "val"), list(b = "val"))`. so we cannot programmatically distinguish between them. So, the "or a vector of lists" is technically what we will accept for any list-valued argument.

R/rdeephaven/R/table_handle_wrapper.R
alexpeters1208: When I do this, the C++ object being pointed to is the same, so we now have two references to the same object, and modification to either reference will change the underlying thing. This is pretty un-Deephaven.

alexpeters1208: Same issue as the above proposal with merge.

R/rdeephaven/R/table_handle_wrapper.R
chipkent: Make them the same.

R/rdeephaven/R/table_handle_wrapper.R
chipkent: I don't think anything needs to be cloned.  just `return(self)`

R/rdeephaven/R/table_handle_wrapper.R
chipkent: `return(self)`.  No reason to touch the c++.

R/rdeephaven/R/helper_functions.R
chipkent:
```suggestion
      stop(paste0("'", arg_name, "' must be a list or a vector. Got an object of class ", first_class(candidate), "."))
```

R/rdeephaven/R/helper_functions.R
```diff
@@ -3,10 +3,8 @@ first_class <- function(arg) {
 }
 
 verify_type <- function(arg_name, candidate, required_type, message_type_name, is_scalar) {
-  if (required_type == "list") {
-    if (first_class(candidate) != "list") {
-      stop(paste0("'", arg_name, "' must be a list or a vector of lists. Got an object of class ", first_class(candidate), "."))
-    }
+  if ((required_type == "list") && (first_class(candidate) != "list")) {
```
chipkent: Does this check work?  What if the list is `c(NULL, c(a,b,c))`?

chipkent (changes requested):

chipkent (approved):

PR #4356, opened 2023-08-22
Python wrapper for MultiJoin feature

### LINKED ISSUE #4280: Python support for `MultiJoin` feature
As a user, I want to use the `MultiJoin` feature from Python instead of being limited to Java/Groovy.

lbooker42:
The call syntax is the following:
```
from deephaven.table import MultiJoinInput, MultiJoinTable, multi_join

# complex join
mj_input = [
    MultiJoinInput(table=t1, on="key"), # all columns added
    MultiJoinInput(table=t2, on="key=otherKey", joins=["col1", "col2"]), # specific columns added
]
multitable = multi_join(input=mj_input)

#simple joins
multitable = multi_join(tables=[t1,t2], on="common_key") # all columns from t1,t2 included in output
multitable = multi_join(tables=[t1,t2], on=["common_key1", "common_key2"]) # all columns from t1,t2 included
```

Closes #4280


py/server/deephaven/jcompat.py
```diff
@@ -157,15 +157,21 @@ def j_lambda(func: Callable, lambda_jtype:jpy.JType, return_dtype: DType = None)
...
-def to_sequence(v: Union[T, Sequence[T]] = None) -> Sequence[Union[T, jpy.JType]]:
-    """A convenience function to create a sequence of unwrapped object from either one or a sequence of input values to
-    help JPY find the matching Java overloaded method to call.
+def to_sequence(v: Union[T, Sequence[T]] = None, wrapped : bool = False) -> Sequence[Union[T, jpy.JType]]:
+    """A convenience function to create a sequence of wrapped or unwrapped object from either one or a sequence of
+    input values to help JPY find the matching Java overloaded method to call.
 
     This also enables a function to provide parameters that can accept both singular and plural values of the same type
     for the convenience of the users, e.g. both x= "abc" and x = ["abc"] are valid arguments.
     """
```
chipkent, line 166: This  is a user-facing method, so the args should be documented.

py/server/deephaven/multijoin.py
```diff
@@ -0,0 +1,126 @@
...
+MultiJoinInput objects or Table objects should be provided as the `input` parameter. 
+
+Args:
+    input (Union[Table, Sequence[Table], MultiJoinInput, Sequence[MultiJoinInput]]): the input objects specifying the
+        tables and columns to include in the join.
+    on (Union[str, Sequence[str]], optional): the column(s) to match if Table objects are provided as the input 
+        parameter, must be a common name or an equality expression that matches every input table, i.e. "col_a = col_b"
+        to rename output column names.  When using MultiJoinInput objects, this parameter is ignored.
+"""
+def multi_join(input: Union[Table, Sequence[Table], MultiJoinInput, Sequence[MultiJoinInput]],
```
chipkent, line 124: Should this be included in `deephaven`?

py/server/deephaven/multijoin.py
```diff
@@ -0,0 +1,126 @@
...
+    MultiJoinInput(table=t1, on="key"), # all columns added
+    MultiJoinInput(table=t2, on="key=otherKey", joins=["col1", "col2"]), #specific columns added
+]
+multitable = multi_join(input=mj_input)
+
+#simple joins
+multitable = multi_join(tables=[t1,t2], on="common_key") # all columns from t1,t2 included in output
+multitable = multi_join(tables=[t1,t2], on=["common_key1", "common_key2"]) # all columns from t1,t2 included
+
+"""
```
chipkent, line 32: I don't think this pydoc will make it to sphinx.  It needs to be in the pydoc at the start of the file.  Also, check if sphinx has markup for code blocks.

chipkent: Actually, I would put this code block in the PR so that DevRel can document it.  We don't have anything similar for other joins in these docs.  It is not obvious to me that this is sufficiently complex or difficult to include that doc here, instead of in the web docs.

py/server/deephaven/multijoin.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+class MultiJoinInput(JObjectWrapper):
```
chipkent, line 35: Should this be included in `deephaven`?

py/server/deephaven/multijoin.py
```diff
@@ -0,0 +1,126 @@
+#
```
chipkent, line 1: @jmao-denver What are your thoughts on where this stuff should live?  I can make arguments that it too should be in `deephaven.table`, since that is where the other joins are.  Not sure.

py/server/deephaven/multijoin.py
```diff
@@ -0,0 +1,126 @@
...
+
+#simple joins
+multitable = multi_join(tables=[t1,t2], on="common_key") # all columns from t1,t2 included in output
+multitable = multi_join(tables=[t1,t2], on=["common_key1", "common_key2"]) # all columns from t1,t2 included
+
+"""
+
+
+class MultiJoinInput(JObjectWrapper):
+    """A MultiJoinInput represents the input tables, key columns and additional columns to be used in the multiJoin. """
```
chipkent, line 36: `multiJoin` is Java.  We need common syntax on this.  `multijoin`?  `multi-join`?  `multi-table natural join`?

py/server/deephaven/multijoin.py
```diff
@@ -0,0 +1,126 @@
...
+
+    @property
+    def j_object(self) -> jpy.JType:
+        return self.j_multijoininput
+
+    def __init__(self, table: Table, on: Union[str, Sequence[str]], joins: Union[str, Sequence[str]] = None):
+        """Creates a new MultiJoinInput containing the table to include for the join, the key columns from the table to
+        match with other table keys plus additional columns containing data from the table. Rows containing unique keys
+        will be added to the output table, otherwise the data from these columns will be added to the existing output
+        rows.
```
chipkent, line 47: I'm not exactly sure what this means.  Consider reworking.
See https://deephaven.io/core/pydoc/code/deephaven.table.html#deephaven.table.Table.natural_join

py/server/deephaven/multijoin.py
```diff
@@ -0,0 +1,126 @@
...
+
+        Args:
+            table (Table): the table to include in the join
```
chipkent, line 50:
```suggestion
            table (Table): a right table to include in the join
```

py/server/deephaven/multijoin.py
```diff
@@ -0,0 +1,126 @@
...
+            self.table = table
+            on = to_sequence(on)
+            joins = to_sequence(joins)
+            self.j_multijoininput = _JMultiJoinInput.of(table.j_table, on, joins)
+        except Exception as e:
+            raise DHError(e, "failed to build a MultiJoinInput object.") from e
+
+
+class MultiJoinTable(JObjectWrapper):
+    """A MultiJoinTable represents the result of a multiJoin. """
```
chipkent, line 69: `multiJoin`: see above

py/server/deephaven/multijoin.py
```diff
@@ -0,0 +1,126 @@
...
+        """Creates a new MultiJoinTable from MultiJoinInput or directly from Tables and key columns. Either
+        MultiJoinInput objects or Table objects should be provided as the `input` parameter.
+
+        Args:
+            input (Union[Table, Sequence[Table], MultiJoinInput, Sequence[MultiJoinInput]]): the input objects
+                specifying the tables and columns to include in the join.
+            on (Union[str, Sequence[str]], optional): the column(s) to match if Table objects are provided as the input
+                parameter, must be a common name or an equality expression that matches every input table, i.e.
+                "col_a = col_b" to rename output column names.  When using MultiJoinInput objects, this parameter
+                is ignored.
```
chipkent, line 90:
```suggestion
            on (Union[str, Sequence[str]], optional): the column(s) to match, can be a common name or an equality expression that matches every input table, i.e.
                "col_a = col_b" to rename output column names.  When using MultiJoinInput objects, this parameter
                is ignored.
```
The extra phrase in the original made it hard to follow.

py/server/deephaven/multijoin.py
```diff
@@ -0,0 +1,126 @@
...
+    def j_object(self) -> jpy.JType:
+        return self.j_multijointable
+
+    def table(self) -> Table:
+        return Table(j_table=self.j_multijointable.table())
+
+    def __init__(self, input: Union[Table, Sequence[Table], MultiJoinInput, Sequence[MultiJoinInput]],
+                 on: Union[str, Sequence[str]] = None):
+        """Creates a new MultiJoinTable from MultiJoinInput or directly from Tables and key columns. Either
+        MultiJoinInput objects or Table objects should be provided as the `input` parameter.
```
chipkent, line 82:
```suggestion
        """Creates a new MultiJoinTable.  The join can be specified in terms of either tables or MultiJoinInputs.
```

py/server/deephaven/multijoin.py
```diff
@@ -0,0 +1,126 @@
...
+                    j_tables = to_sequence(input)
+                    self.j_multijointable = _JMultiJoinFactory.of(on, *j_tables)
+            elif isinstance(input, MultiJoinInput) or (isinstance(input, Sequence) and all(isinstance(ji, MultiJoinInput) for ji in input)):
+                wrapped_input = to_sequence(input, wrapped=True)
+                tables = [ji.table for ji in wrapped_input]
+                with auto_locking_ctx(*tables):
+                    input = to_sequence(input)
+                    self.j_multijointable = _JMultiJoinFactory.of(*input)
+            else:
+                raise DHError(message="input must be Table or MultiJoinInput.")
```
chipkent, line 108:
```suggestion
                raise DHError(message="input must be a Table, a sequence of Tables, a MultiJoinInput, or a sequence of MultiJoinInputs.")
```

py/server/deephaven/multijoin.py
```diff
@@ -0,0 +1,126 @@
...
+
+"""Creates a new MultiJoinTable from MultiJoinInput or directly from tables and key columns. Either
+MultiJoinInput objects or Table objects should be provided as the `input` parameter. 
+
+Args:
+    input (Union[Table, Sequence[Table], MultiJoinInput, Sequence[MultiJoinInput]]): the input objects specifying the
+        tables and columns to include in the join.
+    on (Union[str, Sequence[str]], optional): the column(s) to match if Table objects are provided as the input 
+        parameter, must be a common name or an equality expression that matches every input table, i.e. "col_a = col_b"
+        to rename output column names.  When using MultiJoinInput objects, this parameter is ignored.
```
chipkent, line 122: See change to similar wording above.

py/server/deephaven/multijoin.py
```diff
@@ -0,0 +1,126 @@
...
+                    self.j_multijointable = _JMultiJoinFactory.of(*input)
+            else:
+                raise DHError(message="input must be Table or MultiJoinInput.")
+
+        except Exception as e:
+            raise DHError(e, "failed to build a MultiJoinTable object.") from e
+
+
+"""Creates a new MultiJoinTable from MultiJoinInput or directly from tables and key columns. Either
+MultiJoinInput objects or Table objects should be provided as the `input` parameter.
```
chipkent, line 115: Model the docstring off of https://deephaven.io/core/pydoc/code/deephaven.table.html#deephaven.table.Table.natural_join

py/server/tests/test_multijoin.py
```diff
@@ -0,0 +1,81 @@
+#
+# Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
```
chipkent, line 2:
```suggestion
# Copyright (c) 2016-2023 Deephaven Data Labs and Patent Pending
```

py/server/deephaven/multijoin.py
chipkent, line 124: Does this or any of the other methods here need any of our locking or other annotations used on Table?
e.g. 
```
with auto_locking_ctx(self, table)
```

py/server/tests/test_multijoin.py
```diff
@@ -0,0 +1,81 @@
...
+        mj_table = multi_join(mj_input)
+
+        # Output table is refreshing
+        self.assertTrue(mj_table.table().is_refreshing)
+        # Output table has same # rows as sources
+        with update_graph.exclusive_lock(self.test_update_graph):
+            self.assertEqual(mj_table.table().size, self.ticking_tableA.size)
+            self.assertEqual(mj_table.table().size, self.ticking_tableB.size)
+
+if __name__ == '__main__':
```
chipkent, line 80: I don't see any tests that take a single Table or a single MultiJoinInput.

chipkent (changes requested):

py/server/deephaven/multijoin.py
lbooker42: Yes, currently using `auto_locking_ctx()` in the `MultiJoinTable` `__init__` method.

py/server/tests/test_multijoin.py
lbooker42: Added single table instantiation for all tests.

py/server/tests/test_multijoin.py
```diff
@@ -0,0 +1,120 @@
...
+
+import unittest
+
+from deephaven import read_csv, time_table, update_graph
+from deephaven.multijoin import MultiJoinInput, MultiJoinTable, multi_join
+from tests.testbase import BaseTestCase
+from deephaven.execution_context import get_exec_ctx
+
+
+class UpdateByTestCase(BaseTestCase):
```
jmao-denver, line 13: rename?

py/server/deephaven/multijoin.py
jmao-denver: Good question!
1. Based on the precedents of `PartitionedTable`, `TreeTable`, `RollupTable`, it does seem to make sense to place MultiJoinTable in table.py
2. I am not sure the level of 'table-ness' MultiJoinTable is. My guess is that it is at the same level as the above mentioned ones. The only difference is that those are created by methods on Table, not unbound functions. In a way, it is more similar to `table_factory.merge()`. 
3. One of the reasons we put `PartitionedTable` etc. in table.py is to avoid cyclical imports. MultiJoinTable doesn't seem to have this problem.

Although a bit torn, my personal preference is for consistency by having MultiJoinTable in table.py.  Maybe we can move the function `multi_join' to `table_factory`?

py/server/deephaven/multijoin.py
jmao-denver: I'd say no.

py/server/deephaven/multijoin.py
jmao-denver: I'd say yes, especially if we move it to `table_factory`.

jmao-denver (changes requested):

py/server/deephaven/multijoin.py
lbooker42: I have followed these suggestions from @jmao-denver:

- `multi_join()` is moved to `table_factory.py`
- `MultiJoinTable`/`MultiJoinInput` are moved into `table.py`
- `multijoin.py` is removed

lbooker42: - Also added `multi_join()` to `deephaven/__init__.py`

py/server/deephaven/table_factory.py
```diff
@@ -320,3 +320,23 @@ def ring_table(parent: Table, capacity: int, initialize: bool = True) -> Table:
...
+            tables and columns to include in the join.
+        on (Union[str, Sequence[str]], optional): the column(s) to match, can be a common name or an equality expression
+            that matches every input table, i.e. "col_a = col_b" to rename output column names.  When using MultiJoinInput
+            objects, this parameter is ignored.
+
+    Returns:
+        MultiJoinTable: the result of the multi-table natural join operation. To access the underlying Table, use the
+            table() method.
+    """
+    return MultiJoinTable(input, on)
```
jmao-denver, line 342: Looking at it now I think it was a mistake to suggest that we move it to table_factory.py. I believe table.py is still a better place for it.

jmao-denver (changes requested):
Please move `multi_join` to table.py

py/server/deephaven/table_factory.py
```diff
@@ -13,7 +13,7 @@
 from deephaven.column import InputColumn, Column
 from deephaven.dtypes import DType
 from deephaven.jcompat import to_sequence
-from deephaven.table import Table
+from deephaven.table import Table, MultiJoinTable, MultiJoinInput
```
jmao-denver, line 16: delete?

jmao-denver, line 16: delete?

jmao-denver, line 16:
```suggestion
from deephaven.table import Table
```

jmao-denver (approved):
LGTM

py/server/deephaven/table.py
```diff
@@ -3555,3 +3560,101 @@ def update_by(self, ops: Union[UpdateByOperation, List[UpdateByOperation]],
...
+            tables and columns to include in the join.
+        on (Union[str, Sequence[str]], optional): the column(s) to match, can be a common name or an equality expression
+            that matches every input table, i.e. "col_a = col_b" to rename output column names.  When using MultiJoinInput
+            objects, this parameter is ignored.
+
+    Returns:
+        MultiJoinTable: the result of the multi-table natural join operation. To access the underlying Table, use the
+            table() method.
+    """
+    return MultiJoinTable(input, on)
```
chipkent, line 3660: General API question.  I don't know the answer, but I'm asking it....

Is there any reason the user would need to know about `MultiJoinInputTable`?  Right now, the query syntax looks like:
```
t = multi_join([t1,t2,t3]).table()
```
Any reason it shouldn't just be
```
t = multi_join([t1,t2,t3])
```

chipkent (changes requested):

py/server/deephaven/table.py
lbooker42: Yes, but only because in the future we intend to allow addition (and potentially removal) of tables from this join. 
```
mt = multi_join([t1,t2,t3]);
mt_new = mt.add_table(t4) # returns Immutable MultiJoinTable
```

chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3127
Reference: https://github.com/deephaven/deephaven.io/issues/3128

PR #4360, opened 2023-08-23
4338 live pyobject

### LINKED ISSUE #4338: Unable to export a custom object using `connection.on_data` from Plugin framework
**Description**

If you try and export a custom object in the `connection.on_data` method in the Plugin framework, you get an error. You can export `Table` objects using the same method successfully, and also export a custom object if it's assigned to it's own variable name, but trying to export the object in a `MessageStream` throws an error:

**Steps to reproduce**
1. Install custom plugin that has a custom object that exports another custom object type in a MessageStream

**Expected results**

1. Custom object should export correctly

**Actual results**

1. Exception is thrown. Object is not exported successfully.

```sh
2023-08-18T15:08:40.392Z | heduler-Concurrent-1 | ERROR | i.d.s.s.SessionService    | Internal Error '032d618f-9f48-4127-9fb5-7250c41d48ca' java.lang.RuntimeException: Error in Python interpreter:
Type: <class 'ValueError'>
Value: cannot convert a Python 'TextInput' to a Java 'java.lang.Object'
Line: 51
Namespace: on_data
File: /home/bender/dev/deephaven/iris-oss/core/main/.venv/lib/python3.10/site-packages/deephaven_internal/plugin/object/__init__.py
  File "/home/bender/dev/deephaven/iris-oss/core/main/.venv/lib/python3.10/site-packages/deephaven_internal/plugin/object/__init__.py", line 87, in create_client_connection
  File "/home/bender/dev/deephaven/iris-oss/core/main/.venv/lib/python3.10/site-packages/deephaven/ui/__init__.py", line 64, in create_client_connection
  File "/home/bender/dev/deephaven/iris-oss/core/main/.venv/lib/python3.10/site-packages/deephaven/ui/__init__.py", line 32, in start
  File "/home/bender/dev/deephaven/iris-oss/core/main/.venv/lib/python3.10/site-packages/deephaven/ui/__init__.py", line 41, in send_result
  File "/home/bender/dev/deephaven/iris-oss/core/main/.venv/lib/python3.10/site-packages/deephaven_internal/plugin/object/__init__.py", line 51, in on_data

        at org.jpy.PyLib.callAndReturnValue(Native Method)
        at org.jpy.PyObject.call(PyObject.java:474)
        at org.jpy.PyObject.call(PyObject.java:482)
        at io.deephaven.server.plugin.python.ObjectTypeAdapter.compatibleClientConnection(ObjectTypeAdapter.java:58)
        at io.deephaven.plugin.type.ObjectTypeBase.clientConnection(ObjectTypeBase.java:23)
        at io.deephaven.server.object.ObjectServiceGrpcImpl$SendMessageObserver.lambda$onNext$0(ObjectServiceGrpcImpl.java:145)
        at io.deephaven.server.object.ObjectServiceGrpcImpl$SendMessageObserver$EnqueuedStreamOperation.lambda$run$0(ObjectServiceGrpcImpl.java:97)
        at io.deephaven.server.session.SessionState$ExportBuilder.lambda$submit$2(SessionState.java:1380)
        at io.deephaven.server.session.SessionState$ExportObject.doExport(SessionState.java:921)
        at io.deephaven.server.runner.scheduler.SchedulerModule$ThreadFactory.lambda$newThread$0(SchedulerModule.java:78)
        at org.jpy.PyLib.callAndReturnObject(Native Method)
        at org.jpy.PyObject.call(PyObject.java:449)
        at io.deephaven.server.console.python.DebuggingInitializer.lambda$createInitializer$0(DebuggingInitializer.java:30)
```
**Versions**
Engine Version: 0.28.0
Web UI Version: 0.45.1
Java Version: 11.0.20
Barrage Version: 0.6.0

niloc132:
This attempts to sidestep #1775 by offering a consistent convention to pass Python and Java objects between their runtimes - instead of exclusively "wrapping" for Python and "unwrapping" to return to Java, this encourages "javaify" to wrap or unwrap as necessary to to pass to Java, and "pythonify" to unwrap or wrap as necessary to pass to Python. This way, PyObjects can be avoided being passed to Java directly. 

This new wrapper is a liveness node, to ensure that the surrounding liveness scope always takes ownership on creation, even if the calling Java code isn't aware that it has received a resource that needs to be managed.

Fixes #4338


Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -308,8 +308,10 @@ public String scriptType() {
     @Override
     public Object unwrapObject(Object object) {
         if (object instanceof PyObject) {
-            final PyObject pyObject = (PyObject) object;
-            final Object unwrapped = module.unwrap_to_java_type(pyObject);
+            final Object unwrapped;
+            try (PyObject pyObject = (PyObject) object) {
```
niloc132, line 312: I'm not quite sure about this fix, as it risks the caller not wanting the unwrapped object to be closed. However, this is used in only a few places, and it seems safe to do this, and it prevents a leak. This is added in its own commit, and would be easy to remove, and doesn't affect the rest of the patch directly (except for its blatent wrongness).

There are many more remaining reference leaks in our code - I spent time auditing how often we create/close PyObjects, and the results are not great. Some of these leaks arent terrible, like references to boolean values, but we quickly leak lots of other values in terrible ways, and should find a better way to address this. 

We could consider adding a finalize() to jpy's PyObject, or drop Java8 support and look into Cnealer/etc instead, but probably only want to do that as a logging mechanism rather than actually managing resources this way.

server/src/main/java/io/deephaven/server/plugin/python/PyObjectRefCountedNode.java
```diff
@@ -0,0 +1,40 @@
...
+import io.deephaven.util.annotations.ScriptApi;
+import org.jpy.PyObject;
+
+/**
+ * Provides a mapping between Python refcount and Deephaven's liveness mechanism, allowing liveness scopes to manage the
+ * single <a href='https://docs.python.org/3/glossary.html#term-strong-reference'>strong reference</a> that the PyObject
+ * instance represents. This way, as long as PyObjectRefCountedNode instances are managed correctly by their parent
+ * scope, the PyObject strong reference will be correctly released when the object is no longer used.
+ */
+public class PyObjectRefCountedNode extends LivenessArtifact {
```
devinrsmith, line 14: So, I think `PyObject`s _should_ be correctly released already. I don't know about the tie-ins to the liveness mechanism and if they are necessary for correctness or not...

devinrsmith (commented):
I think `pythonify` may be extraneous, as we should be able to handle all of that in our calling code to save an extra boundary crossing (may make sense to have `pythonify` as a java method).

I'm guessing the liveness issues are real and that's why PyObjectRefCountedNode is necessary, but I don't think it's bc they won't be decRefd...

Need to spend some more time thinking about ref counting and how all of this plays together...

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
devinrsmith: I don't think we should be closing PyObject here.

`finalize()` is a very critical time, and I don't think we want to risk blocking on the GIL (which is necessary when we decRef).

What is `Cnealer/etc`?

py/server/deephaven/_wrapper.py
```diff
@@ -117,6 +119,36 @@ def _lookup_wrapped_class(j_obj: jpy.JType) -> Optional[type]:
...
+
+def pythonify(j_obj: Union[jpy.JType, Any]) -> Optional[Union[JObjectWrapper, jpy.JType, Any]]:
+    """
+    Reciprocal of javaify, returns an object that is safe to be used in Python after being passed
+    from Java.
+    """
+    if not isinstance(j_obj, jpy.JType):
+        return j_obj
+    # Definitely a JType, check if it is a PyObjectRefCountedNode
+    if hasattr(j_obj,'getPythonObject'):
```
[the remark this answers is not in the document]
niloc132: Nice, ty, will do that and make the class final

server/src/main/java/io/deephaven/server/plugin/python/PyObjectRefCountedNode.java
```diff
@@ -0,0 +1,40 @@
...
+    public PyObjectRefCountedNode(PyObject pythonObject) {
+        super(true);
+        this.pythonObject = pythonObject;
+    }
+
+    @Override
+    protected void destroy() {
+        super.destroy();
+        pythonObject.close();
+    }
```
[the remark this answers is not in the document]
devinrsmith: `PyObject#close` is an explicit way for us to say "we aren't going to use this PyObject instance from java anymore" (ie, we don't plan on keeping any hard references to it); there is definitely reason for it to exist IMO. User / implementation code should proactively `PyObject#close` when appropriate (and there are definitely cases where we could be better about calling close), even though yes, it may block on the GIL... we should never do this in destroy() though...

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
niloc132: Sorry, apparently typo'd [Cleaner](https://docs.oracle.com/javase/9/docs/api/java/lang/ref/Cleaner.html) in a weird way.

I'm not sure how finalize enters into this discussion - liveness should not depend on finalize at all, and LivenessArtifact and its hierarchy don't even have a finalize() override to ensure we unmanage when gc'd. Instead, these will almost all be called (with the exception of this one place I've noted might not belong in the patch) when the client calls SessionService.Release, when a session ends, or at the end of a submit() call's own liveness scope, etc.

server/src/main/java/io/deephaven/server/plugin/python/PyObjectRefCountedNode.java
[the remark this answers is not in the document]
niloc132: Does `destroy()` not have the same same semantics, the object will go out of scope, we don't expect any remaining hard references? This isn't related to `finalize()`, and should happen well before finalize goes off (or not at all, I don't see any Cleaner/PhantomReference/finalize implementations that call `destroy()`). 

The alternative is a patch that "fixes" 1775 by wrapping PyObject in some other generic Object and just leaking the references, and hoping that making Python GC dependent on Java GC won't ever go wrong.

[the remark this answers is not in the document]
devinrsmith: Ohhhhhh I had a brain fart and was thinking this was `finalize()` the whole time.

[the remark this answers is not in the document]
niloc132: Haha yeah I'd be pretty unhappy with this patch then too, 100%. Lmk if you'd want a call or some time to debug to make sure about this.

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
devinrsmith: In this case, you mentioned

> We could consider adding a finalize() to jpy's PyObject

I'm not in favor of that. I think we do a lot of what Cleaner already does, see `org.jpy.PyObjectReferences#cleanupOnlyUseFromGIL(long[])`

niloc132: Agreed, I also don't want to add finalize to PyObject, but I do want to decref objects that are dropped by java. Example calling code that is fixed by this, in ScopeTicketResolver:
```
        T export = gss.getExecutionContext().getUpdateGraph().sharedLock().computeLocked(() -> {
            T scopeVar = null;
            try {
                // noinspection unchecked
                scopeVar = (T) gss.unwrapObject(gss.getVariable(scopeName));
            } catch (QueryScope.MissingVariableException ignored) {
            }
            return scopeVar;
        });
```
The result of getVariable immediately goes out of scope and isn't accessible again. We could store its result in a local, and test if it is Closable, and if so use try-with-resources? It seems a shame to need to do that.

Another option is to actually change the semantics of getVariable to return this new PyObjectRefCountedNode instead, and assume that a parent liveness scope exists. There is no explicit liveness scope specified here, but _most_ (didn't confirm all) calls that get here have an implicit/automatic liveness scope. However, changing getVariable seemed like too big of a step, though it would fix this specific case.

Other thoughts? Back this one piece out and plan on a bigger change later to update getVariable to always wrap the same way?

server/src/main/java/io/deephaven/server/plugin/python/PyObjectRefCountedNode.java
jmao-denver: Right now PyObject isn't instantiate-able, so this could fill that hole. But would making PyObject constructors public achieve almost the same goal besides the liveness benefits? Just a question.

niloc132: I don't think that would help, even ignoring the liveness issue - the problem of #1775 is that we can't freely pass both java and python objects from python to java - we need some "real" java object that contains the PyObject.

server/src/main/java/io/deephaven/server/plugin/python/PyObjectRefCountedNode.java
```diff
@@ -0,0 +1,40 @@
+package io.deephaven.server.plugin.python;
```
niloc132, line 1: A point from code review with Jianfeng over slack - should this go in a different package/module? It needs to be in a project that can see both jpy's PyObject and engine-updategraph's LivenessArtifact, and I wasn't immediately happy with any of the choices I saw...

jmao-denver (dismissed):
LGTM

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
devinrsmith: Yes - I suggest backing this change out and adding a ticket. I think there are a few different ways we could tackle it in the future. Another option is new method, `getUnwrappedVariable`.

devinrsmith (approved):

jmao-denver (dismissed):

server/src/main/java/io/deephaven/server/plugin/python/PyObjectRefCountedNode.java
chipkent: I think this is a valid question.  I would have looked for it in Integrations, but I think this is more of a @jmao-denver question.

py/server/deephaven/_wrapper.py
```diff
@@ -117,6 +119,39 @@ def _lookup_wrapped_class(j_obj: jpy.JType) -> Optional[type]:
     return None
 
 
+def javaify(obj: Union[JObjectWrapper, jpy.JType, Any]) -> Optional[jpy.JType]:
```
chipkent, line 122:
```suggestion
def javaify(obj: Any) -> Optional[jpy.JType]:
```
Are the other type hints redundant?

py/server/deephaven/_wrapper.py
```diff
@@ -117,6 +119,39 @@ def _lookup_wrapped_class(j_obj: jpy.JType) -> Optional[type]:
...
+    if isinstance(obj, JObjectWrapper):
+        return obj.j_object
+    if isinstance(obj, jpy.JType):
+        return obj
+    # We must return a java object, so wrap in a PyObjectLivenessNode so that the server's liveness tracking
+    # will correctly notify python that the object was released
+    return JPyObjectRefCountedNode(obj)
+
+
+def pythonify(j_obj: Union[jpy.JType, Any]) -> Optional[Union[JObjectWrapper, jpy.JType, Any]]:
```
chipkent, line 138:
```suggestion
def pythonify(j_obj: Any) -> Optional[Union[JObjectWrapper, jpy.JType, Any]]:
```
redundant type hints?

chipkent (changes requested):

py/server/deephaven/_wrapper.py
niloc132: Yes, they technically are redundant - in Java (esp lacking unions...) I would certainly just say "object", but given that type hints in python are more of "documentation plus", I was trying to be explicit that there are three "categories" of objects that could be passed in, and will be handled separately. In a different universe/language, this might be done with an overloaded method.

It seemed more descriptive to me this way, but I can see this point too, let me know what you'd prefer.

py/server/deephaven/_wrapper.py
niloc132: Ditto, typehints have no actual effect beyond "documentation and your IDE can look disapprovingly at you".

server/src/main/java/io/deephaven/server/plugin/python/ObjectTypeAdapter.java
```diff
@@ -88,12 +105,21 @@ private PythonServerMessageStream(PyObject instance) {
...
             byte[] bytes = new byte[payload.limit()];
             payload.get(bytes);
+
+            for (int i = 0; i < references.length; i++) {
+                Object reference = references[i];
+                if (reference instanceof LivePyObjectWrapper) {
+                    reference = ((LivePyObjectWrapper) reference).getPythonObject();
+                }
+                references[i] = reference;
+            }
```
devinrsmith, line 115: I'm a _little_ bit worried about modifying the existing array, even if it probably doesn't matter in practice:

1. Caller _may_ want to do something w/ `references` after.
2. We are potentially wiping away the only reference to a `LivePyObjectWrapper`; if we get unlucky, would GC cause `LivePyObjectWrapper#destroy`, ie `PyObject#close`?

I think it's safer to create a new array.

devinrsmith (changes requested):

py/server/deephaven/_wrapper.py
```diff
@@ -21,7 +21,7 @@
 _di_wrapper_classes: Set[JObjectWrapper] = set()
 _has_all_wrappers_imported = False
 
-JPyObjectRefCountedNode = jpy.get_type('io.deephaven.server.plugin.python.PyObjectRefCountedNode')
+JLivePyObjectWrapper = jpy.get_type('io.deephaven.server.plugin.python.LivePyObjectWrapper')
```
chipkent, line 24: It does not appear like this type is exposed to the user.  Can it be made private? (`_` prefix)

chipkent (changes requested):

mofojed (commented):
Looks good to me, I agree with Devin's comment about creating a new array.

py/server/deephaven/_wrapper.py
niloc132: The whole file is already private, `_wrapper.py`, is that still advised?

Changed to both methods to use `Any` as suggested after discussing with Jianfeng.

devinrsmith (dismissed):

server/src/main/java/io/deephaven/server/plugin/python/ObjectTypeAdapter.java
```diff
@@ -106,14 +106,18 @@ public void onData(ByteBuffer payload, Object[] references) {
             byte[] bytes = new byte[payload.limit()];
             payload.get(bytes);
 
+            // Make a new array holding objects to pass to python, to avoid mutating the passed array. Note that this
+            // will have no impact on GC/liveness, as the JVM is free to GC the original references array after the loop
+            // byt before the call() invocation.
```
mofojed, line 111:
```suggestion
            // but before the call() invocation.
```

mofojed (dismissed):

mofojed (approved):

rcaudy (approved):

chipkent (approved):

PR #4361, opened 2023-08-23
Update R README.md

alexpeters1208:
Update R readme to account for new C++ build process

### DIFF
--- a/R/rdeephaven/README.md
+++ b/R/rdeephaven/README.md
@@ -104,24 +104,47 @@ Currently, the R client is only supported on Ubuntu 20.04 or 22.04 and must be b
    **NOTE**
 
    If using RStudio for this step, the environment variables that were set in step 3 may not persist into the RStudio environment,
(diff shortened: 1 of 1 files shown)


R/rdeephaven/README.md
```diff
@@ -109,7 +162,7 @@ compilation terminated.
 ```
 this means that the C++ compiler does not know where to find the relevant header files for the Deephaven C++ client. This can happen for a handul of reasons:
 1. Step 1 was skipped, and the Deephaven C++ client was not installed. In this case, please ensure that the client is installed before attempting to build the R client.
-2. The Deephaven C++ client is installed, but the `DHCPP` environment variable is not set. To test this, run
+2. The Deephaven C++ client is installed, but step 3 was skilled and the `DHCPP` environment variable is not set. To test this, run
```
chipkent, line 165: skilled?

chipkent (changes requested):

chipkent (approved):

PR #4362, opened 2023-08-23
Add additional methods to the `MultiJoinFactory` and `MultiJoinInput` Java/Groovy interface for simple construction.

lbooker42:
Support comma-separated string arguments for the `MultiJoinFactory` construction methods similar to the other join methods.


devinrsmith (approved):
We should probably and try to work up these interfaces for `table-api` / RPCs.

PR #4363, opened 2023-08-23
Add docstrings to R Client

alexpeters1208:
Once these are reviewed, I will use them to create custom .Rd pages that are accessible via ?. Cross-join is not documented, as it is not documented in the Python API, not available in the Python API, and our cross-join does not behave how I would expect, so I don't know what to do about that.


R/rdeephaven/R/aggregate_wrapper.R
```diff
@@ -14,80 +14,135 @@ Aggregation <- R6Class("Aggregation",
 
 ### All of the functions below return an instance of the above class
 
+#' @description
+#' Aggregation function that computes the first value of each column in `cols`, for each aggregation group.
+#' Intended to be used in a call to `TableHandle$agg_by()`.
+#' @param cols String or list of strings denoting the names of the columns to aggregate.
+#' @return Result of the aggregation.
```
chipkent, line 21: This documentation is inaccurate and should largely be ripped off from the python equivalent.  The python equivalent is `Creates a First aggregation`, which indicates that the method is creating an aggregation.  What is here seems to indicate that the function actually performs an aggreation, which is inaccurate.

I will review the rest of the file after changes are made.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -239,6 +446,10 @@ dim.TableHandle <- function(x) {
   return(x$dim())
 }
 
+#' @description
```
chipkent, line 449: All of the documentation should largely be ripped off from Python where it has gone through extensive review as to what it should say.  `Merge one or more TableHandles` is very unclear/imprecise compared to `Merges several tables into one table on the server.`
https://deephaven.io/core/client-api/python/code/pydeephaven.html#pydeephaven.Session.merge_tables

chipkent (changes requested):

R/rdeephaven/R/aggregate_wrapper.R
```diff
@@ -1,95 +1,176 @@
...
   )
 )
 
 ### All of the functions below return an instance of the above class
 
+#' @description
+#' Create a First aggregation that computes the group-wise first value
+#' of each column in `cols`, for each aggregation group.
+#' @param cols String or list of strings denoting the column(s) to aggregate. Can be renaming expressions, i.e. “new_col = col”.
+#' Default is to use all non-grouping columns, which is only valid in the `agg_all_by()` operation.
```
chipkent, line 22: Did not rip off.  Can someone with broken English as a 3rd language parse this?

chipkent (changes requested):

R/rdeephaven/R/aggregate_wrapper.R
alexpeters1208: Pydoc is
`the column(s) to aggregate on, can be renaming expressions, i.e. “new_col = col”; default is None, only valid when used in Table agg_all_by operation`
I feel this is as close as I can accurately get to that, as now `default is None` has no meaning in R.

R/rdeephaven/R/aggregate_wrapper.R
```diff
@@ -1,95 +1,176 @@
...
+      self$.internal_rcpp_object <- do.call(aggregation, args)
     }
   )
 )
 
 ### All of the functions below return an instance of the above class
 
+#' @description
+#' Create a First aggregation that computes the group-wise first value
+#' of each column in `cols`, for each aggregation group.
```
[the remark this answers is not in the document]
alexpeters1208: Pydoc is
`Creates a First aggregation.`
I don't think this is the most helpful documentation on this function. What do you think of
`Creates a First aggregation that computes the first value of each column in `cols` for each aggregation group.`

R/rdeephaven/R/aggregate_wrapper.R
```diff
@@ -1,95 +1,176 @@
-#' @export
```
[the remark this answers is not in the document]
alexpeters1208: See comment above.
I can still include a docstring for Aggregation although it's not exported. The end-stage RStudio `?Aggregation` etc docs will not ultimately be tied to these docstrings. Although it is R convention to have it done this way, I've played with it a lot and determined that is not the best solution for our package.

R/rdeephaven/R/aggregate_wrapper.R
```diff
@@ -17,69 +20,150 @@ Aggregation <- R6Class("Aggregation",
...
+#' Default is to aggregate all non-grouping columns, which is only valid in the `agg_all_by()` operation.
+#' @return Aggregation function to be used in `agg_by()` or `agg_all_by()`.
+#' @export
 agg_sum <- function(cols = character()) {
   verify_string("cols", cols, FALSE)
   return(Aggregation$new(INTERNAL_agg_sum, "agg_sum", cols=cols))
 }
 
+#' @description
+#' Creates an Absolute Sum aggregation that computes the sum of each column in `cols` for each aggregation group.
```
chipkent, line 79: inaccurate

R/rdeephaven/R/aggregate_wrapper.R
```diff
@@ -17,69 +20,150 @@ Aggregation <- R6Class("Aggregation",
...
   verify_string("cols", cols, FALSE)
   return(Aggregation$new(INTERNAL_agg_percentile, "agg_percentile", percentile=percentile, cols=cols))
 }
 
-agg_count <- function(col) {
+#' @description
+#' Creates a Count aggregation that counts the number of rows in each aggregation group.
+#' Note that this operation is not supported in `agg_all_by()`.
+#' @param col String denoting the name of the new column to hold the counts of each aggregation group.
+#' Defaults to "n".
```
chipkent, line 163: I think this was changed

R/rdeephaven/R/aggregate_wrapper.R
```diff
@@ -17,69 +20,150 @@ Aggregation <- R6Class("Aggregation",
...
+#' @return Aggregation function to be used in `agg_by()`.
+#' @export
+agg_count <- function(col = "n") {
```
chipkent, line 166: I think this was changed.  Look through for the few places from the other PR that were modified.

chipkent (changes requested):

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,8 +1,18 @@
+#' @description
+#' A Client is the entry point for interacting with the Deephaven server. It is used to create new tables,
+#' import data to and from the server, and run scripts on the server.
```
chipkent, line 3:
```suggestion
#' import data to and export data from the server, and run queries on the server.
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -170,14 +235,27 @@ Client <- R6Class("Client",
         stop(paste0("'table_object' must be a single data frame, tibble, arrow table, or record batch reader. Got an object of class ", table_object_class[[1]], "."))
       }
     },
+
+    #' @description
+    #' Uses an existing ticket to create a new table on the server.
+    #' @param ticket String denoting the ticket to use to create the new table.
+    #' @return TableHandle reference to the new table.
```
chipkent, line 242: This is not creating a new table on the server.  The ticket is a key used to retrieve a table from the server.  At least, that is my impression of what it does.

R/rdeephaven/R/client_wrapper.R
```diff
@@ -138,22 +179,46 @@ Client <- R6Class("Client",
...
       if (!private$check_for_table(name)) {
         stop(paste0("The table '", name, "' does not exist on the server."))
       }
       return(TableHandle$new(self$.internal_rcpp_object$open_table(name)))
     },
+
+    #' @description
+    #' Imports a new table to the Deephaven server. Note that this new table is not automatically bound to
+    #' a variable name on the server. See `?TableHandle` for more information.
+    #' @param table_object R Data Frame, a dplyr Tibble, an Arrow Table, or an Arrow RecordBatchReader
```
chipkent, line 219:
```suggestion
    #' @param table_object R Data Frame, dplyr Tibble, Arrow Table, Arrow RecordBatchReader, or other supported table
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -138,22 +179,46 @@ Client <- R6Class("Client",
...
+    #' Defaults to now.
+    #' @return TableHandle reference to the new table.
     time_table = function(period, start_time = "now") {
       verify_string("period", period, TRUE)
       verify_string("start_time", start_time, TRUE)
       return(TableHandle$new(self$.internal_rcpp_object$time_table(period, start_time)))
     },
+
+    #' @description
+    #' Opens a table named `name` from the server if it exists.
```
chipkent, line 205:
```suggestion
    #' Gets a reference to a table on the server using a variable name in the server's scope.
```
More or less clear?

R/rdeephaven/R/client_wrapper.R
```diff
@@ -19,10 +29,41 @@ Client <- R6Class("Client",
...
+    #' @description
+    #' Initialize a Client object using a pointer to an existing client connection.
+    #' @param xptr External pointer to an existing client connection.
     initialize_for_xptr = function(xptr) {
       verify_type("xptr", xptr, "externalptr", "XPtr", TRUE)
       self$.internal_rcpp_object = new(INTERNAL_Client, xptr)
     },
+
+    #' @description
+    #' Initializes a Client object that connects to a Deephaven server
```
chipkent, line 42:
```suggestion
    #' Initializes a Client object and connects to a Deephaven server.
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -19,10 +29,41 @@ Client <- R6Class("Client",
       }
       return(do.call(self$initialize_for_target, args))
     },
+    
+    #' @description
+    #' Initialize a Client object using a pointer to an existing client connection.
```
chipkent, line 34:
```suggestion
    #' Initializes a Client object using a pointer to an existing client connection.
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,8 +1,18 @@
...
+#' A Client is the entry point for interacting with the Deephaven server. It is used to create new tables,
+#' import data to and from the server, and run scripts on the server.
 #' @export
 Client <- R6Class("Client",
   cloneable = FALSE,
   public = list(
     .internal_rcpp_object = NULL,
+
+    #' @description
+    #' Initializes a Client object that connects to the Deephaven server.
```
chipkent, line 11:
```suggestion
    #' Initializes a Client object connected to a Deephaven server.
```

R/rdeephaven/R/client_wrapper.R
```diff
@@ -1,8 +1,18 @@
...
+    #' This method can accept either a string or an Rcpp::XPtr object, and will call
+    #' `initialize_for_target` or `initialize_for_xptr` accordingly.
+    #' @param ... String denoting the address of a running Deephaven server, formatted as `"ip:port"`,
+    #' or an Rcpp::XPtr object that points to an existing client connection.
```
chipkent, line 15: This is confusing and unclear.  The "string" path can take a ton more args than just a string.  This needs to get reworked to be very clear and precise.

chipkent (changes requested):

R/rdeephaven/R/client_wrapper.R
alexpeters1208: I previously sent this to Cristian and he was happy with this doc. I don't totally understand the ticket idea, so I can't say for sure. I will look again.

R/rdeephaven/R/client_wrapper.R
alexpeters1208: I don't find either as clear as it could be. Maybe "Gets a reference to a table named 'name' on the server." ?

R/rdeephaven/R/client_wrapper.R
alexpeters1208: You're right, but those args are documented in the "initialize_from_target" doc. I would rather not have this method exposed at all and wish users could think of it as an overloaded constructor. I don't really know how to document this properly.

R/rdeephaven/R/client_wrapper.R
chipkent: I'm concerned that they view 'name' as the variable name.

R/rdeephaven/R/client_wrapper.R
chipkent: Ask Colin.

rcaudy (approved):
Approved, let's merge what we have and follow up next week if further revisions are required.

PR #4367, opened 2023-08-24
Binary search kernel for pushdown predicates

nbauernfeind:
This includes the binary search kernel from the pushdown predicate port, a fix to make timsort Context's as AutoCloseable. (Closing was missed in the original PR, and we usually use that pattern anyway).

I've written a handful of replicated tests as well; these tests give us good coverage on the implementation.

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/kernel/CharRegionBinarySearchKernel.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/kernel/CharRegionBinarySearchKernel.java
@@ -1,0 +1,114 @@
+/**
+ * Copyright (c) 2016-2023 Deephaven Data Labs and Patent Pending
+ */
+package io.deephaven.engine.table.impl.sources.regioned.kernel;
+
+import io.deephaven.api.SortColumn;
+import io.deephaven.chunk.WritableCharChunk;
+import io.deephaven.chunk.attributes.Any;
+import io.deephaven.engine.rowset.RowSet;
+import io.deephaven.engine.rowset.RowSetBuilderSequential;
+import io.deephaven.engine.rowset.RowSetFactory;
+import io.deephaven.engine.table.impl.sort.timsort.CharTimsortDescendingKernel;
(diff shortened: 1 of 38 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/kernel/CharRegionBinarySearchKernel.java
```diff
@@ -0,0 +1,114 @@
...
+import io.deephaven.util.type.ArrayTypeUtils;
+import org.jetbrains.annotations.NotNull;
+
+public class CharRegionBinarySearchKernel {
+    public static RowSet binSearchMatch(
+            ColumnRegionChar<?> region,
+            long firstKey,
+            final long lastKey,
+            @NotNull final SortColumn sortColumn,
+            @NotNull final Object[] sortedKeys) {
```
lbooker42, line 25: Is `sortedKeys` the correct name? I believe these are the values to search for and they may not be sorted at this point. Would `searchValues` or `toFind` be better?

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/kernel/CharRegionBinarySearchKernel.java
```diff
@@ -0,0 +1,114 @@
...
+import io.deephaven.engine.rowset.RowSetFactory;
+import io.deephaven.engine.table.impl.sort.timsort.CharTimsortDescendingKernel;
+import io.deephaven.engine.table.impl.sort.timsort.CharTimsortKernel;
+import io.deephaven.engine.table.impl.sources.regioned.ColumnRegionChar;
+import io.deephaven.util.compare.CharComparisons;
+import io.deephaven.util.type.ArrayTypeUtils;
+import org.jetbrains.annotations.NotNull;
+
+public class CharRegionBinarySearchKernel {
+    public static RowSet binSearchMatch(
```
lbooker42, line 20: Can you add comments to this function?  It's the sole user-facing function and some of parameters have misleading names.

lbooker42 (dismissed):
Thanks, much clearer.

Issue #3077, opened 2022-11-11
AbstractTableLocationProvider should support removal of dead (deleted) locations


abaranec:
In Enterprise we added support for notifying providers of removed / truncated locations.  We should port this support to community

PR #4373, opened 2023-08-24
Add Location removal support and update SourcePartitionedTable

### LINKED ISSUE #4372: Update SourcePartitionedTable to handle location removal
WIth #3077 SourcePartitionedTable should be able to support removal of locations as well.

abaranec:
This fixes #4372 and #3077.


engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/AbstractTableLocationProvider.java
```diff
@@ -214,6 +245,31 @@ private void verifyPartitionKeys(@NotNull TableLocationKey locationKey) {
         }
     }
 
+    // region Mark / Sweep support
+    protected final void markLocationKeys() {
```
cpwright, line 249: It isn't  yet clear to me why we can't do this at our own (higher) level.

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/FilteredTableDataService.java
```diff
@@ -136,13 +136,22 @@ public boolean hasTableLocationKey(@NotNull final TableLocationKey tableLocation
...
             }
             return inputProvider.getTableLocationIfPresent(tableLocationKey);
         }
 
+        @Override
+        public void removeTableLocationKey(@NotNull final TableLocationKey locationKey) {
+            if (!locationKeyFilter.accept(locationKey)) {
+                return;
+            }
+
```
cpwright, line 151: Not obvious we should filter instead of removing it unconditionally.  Enterprise does not check before removing.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
```diff
@@ -209,6 +211,20 @@ private final class ConstituentListenerRecorder extends LinkedListenerRecorder {
...
+            // We will just record the error here for now.  If the table was removed, then we don't actually care about it
+            // but if the error is real, and the table is not removed, it will be propagated in processExisting()
+            this.setNotificationStep(getUpdateGraph().clock().currentStep());
+            this.error = originalException;
+            mergedListener.notifyChanges();
+        }
+
+        @Override
+        public boolean recordedVariablesAreValid() {
+            return error == null && super.recordedVariablesAreValid();
```
cpwright, line 225: Don't you need to fix this check in MergedListener to use the recordedVariablesAreValid instead?

                for (ListenerRecorder recorder : recorders) {
                    if (recorder.getNotificationStep() == currentStep) {
                        added += recorder.getAdded().size();
                        removed += recorder.getRemoved().size();
                        modified += recorder.getModified().size();
                        shifted += recorder.getShifted().getEffectiveSize();
                    }
                }

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
```diff
@@ -491,6 +507,12 @@ private void processExisting(@NotNull final Table constituent) {
             if (constituent.isRefreshing()) {
                 assert nextListener != null;
                 Assert.eq(nextListener.getParent(), "listener parent", constituent, "existing constituent");
+
+                // Make sure we propagate any actual error on to the listeners
+                if(nextListener.error != null) {
+                    throw new UncheckedDeephavenException(nextListener.error);
```
cpwright, line 513: Probably worth propagating this in a more managed way (like a special exception type) to handle the in process() explicitly.

engine/table/src/main/java/io/deephaven/engine/table/impl/SourcePartitionedTable.java
```diff
@@ -186,20 +183,43 @@ private RowSet sortAndAddLocations(@NotNull final Stream<TableLocation> location
...
+            final PartitionAwareSourceTable constituent = new PartitionAwareSourceTable(
                     constituentDefinition,
                     "SingleLocationSourceTable-" + tableLocation,
                     RegionedTableComponentFactoryImpl.INSTANCE,
                     new SingleTableLocationProvider(tableLocation),
-                    refreshSizes ? refreshCombiner : null));
+                    refreshSizes ? refreshCombiner : null);
+
+            // These can't be systemic or when they get notified on error,  they will crash the worker.
+            constituent.setAttribute(Table.SYSTEMIC_TABLE_ATTRIBUTE, false);
```
cpwright, line 199: We need to be able to determine this based on the systematic attribute of the whole table, or our context not just force it to false.  A systemic SourcePartitionedTable should propagate that attribute to the constituents; one that is not systemic should not.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
abaranec: No,  the recorder will return empty everytthing when recorded values are not valid.  That's why I did it that way

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/AbstractTableLocationProvider.java
abaranec: The problem was the visitation.  The locations need to be visited by handleTableLocationKey,  which is final in AbstractTableLocationProvider.   I don't want to make it not final and override because then we'd have to copy-pasta the implementation. I could add a do-nothing protected method `onLocationAdded()` with some javadoc explaining how it is meant to be used.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
```diff
@@ -427,13 +434,23 @@ else if (nextCurrentKey == nextModifiedKey) {
...
+                    }
                     advanceCurrent();
                     ++nextCurrentSlot;
                     ++nextPreviousSlot;
                 }
             }
 
+            if(constituentExceptions != null) {
+                throw new UncheckedDeephavenException(
+                        new MultiException("Errors occurred processing constituent tables:",
```
cpwright, line 450: io.deephaven.util.MultiException#maybeWrapInMultiException(java.lang.String, java.lang.Throwable...) so we don't wrap a multiexception for one thing

cpwright:
I am satisfied with this code.  @rcaudy can you please assign a Community team member to review it?

engine/table/src/main/java/io/deephaven/engine/table/impl/SourcePartitionedTable.java
```diff
@@ -219,22 +241,66 @@ private void processPendingLocations(final boolean notifyListeners) {
...
+            if (relevantRemovedLocations.isEmpty()) {
+                return RowSetFactory.empty();
+            }
+
+            final RowSetBuilderSequential deleteBuilder = RowSetFactory.builderSequential();
+
+            // We don't have a map of location to row key, so we have to iterate them. If we decide this is too slow, we
+            // could add a TObjectIntMap
+            // as we process pending added locations and then we can just make an index of rows to remove by looking up
+            // in that map.
```
rcaudy, line 266:
```suggestion
            // We don't have a map from location key to row key, so we have to iterate the location keys. If we decide
            // this is too slow, we could add a TObjectIntMap as we process pending added locations and then just make
            // a RowSet of rows to remove by looking up in that map.
```

engine/table/src/main/java/io/deephaven/engine/table/impl/SourcePartitionedTable.java
```diff
@@ -219,22 +241,66 @@ private void processPendingLocations(final boolean notifyListeners) {
...
+                        if (relevantRemovedLocations.contains(removedKeys.get(chunkPos))) {
+                            deleteBuilder.appendKey(keyChunk.get(chunkPos));
+
+                            final Table deletedTable = tableChunk.get(chunkPos);
+                            result.unmanage(deletedTable);
+                            deletedTable.close();
+                        }
+                    }
+                }
+            }
```
rcaudy, line 292: This code would be prettier (although maybe a little less efficient) if you wrote it with column iterators. Actually, now that I see you using key chunks, I suspect it might just be better the other way.

Regardless of whether you take my suggestion, `resultTableLocationKeys` should be changed to hold `ImmutableTableLocationKey`.

```suggestion
            // @formatter:off
            try (final CloseableIterator<ImmutableTableLocationKey> keysIterator =
                         ChunkedObjectColumnIterator.make(resultTableLocationKeys, resultRows);
                 final CloseableIterator<Table> constituentsIterator =
                         ChunkedObjectColumnIterator.make(resultLocationTables, resultRows);
                 final RowSet.Iterator rowsIterator = resultRows.iterator()) {
                // @formatter:on

                while (keysIterator.hasNext()) {
                    final TableLocationKey key = keysIterator.next();
                    final Table constituent = constituentsIterator.next();
                    final long rowKey = rowsIterator.nextLong();
                    if (relevantRemovedLocations.contains(key)) {
                        deleteBuilder.appendKey(rowKey);
                        result.unmanage(constituent);
                        constituent.close();
                    }
                }
            }
```

engine/table/src/main/java/io/deephaven/engine/table/impl/SourcePartitionedTable.java
```diff
@@ -219,22 +241,66 @@ private void processPendingLocations(final boolean notifyListeners) {
...
+                                    subSeq)
+                            .asObjectChunk();
+
+                    for (int chunkPos = 0; chunkPos < keyChunk.size(); chunkPos++) {
+                        if (relevantRemovedLocations.contains(removedKeys.get(chunkPos))) {
+                            deleteBuilder.appendKey(keyChunk.get(chunkPos));
+
+                            final Table deletedTable = tableChunk.get(chunkPos);
+                            result.unmanage(deletedTable);
+                            deletedTable.close();
```
rcaudy, line 288: 1. This `unmanage` should happen after downstream tables have processed the removes. Use an `UpdateCommitter`.
2. I'm not sure what you're on about with a `close()`. It's not really something we do. If you're going to do it, at least do it after cycle close.

Crap. This actually raises doubts about the entire PR. A removed location cannot be accessed safely (only cached data can be read), but we're telling downstream ops about the removed table here and kind of hoping they don't look at any of its data, which is almost certainly what every useful operation will want to do.

engine/table/src/main/java/io/deephaven/engine/table/impl/SourcePartitionedTable.java
```diff
@@ -187,19 +184,42 @@ private RowSet sortAndAddLocations(@NotNull final Stream<TableLocation> location
...
 
         private Table makeConstituentTable(@NotNull final TableLocation tableLocation) {
-            return applyTablePermissions.apply(new PartitionAwareSourceTable(
+            final PartitionAwareSourceTable constituent = new PartitionAwareSourceTable(
                     constituentDefinition,
                     "SingleLocationSourceTable-" + tableLocation,
                     RegionedTableComponentFactoryImpl.INSTANCE,
                     new SingleTableLocationProvider(tableLocation),
-                    refreshSizes ? refreshCombiner : null));
+                    refreshSizes ? refreshCombiner : null);
```
rcaudy, line 196: I *think* we could stop using a `refreshCombiner` for the constituents if we added installed a `ConstituentDependency` on the result table. Might be more efficient.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
```diff
@@ -209,6 +212,22 @@ private final class ConstituentListenerRecorder extends LinkedListenerRecorder {
             setMergedListener(mergedListener);
         }
 
+        @Override
+        protected void onFailureInternal(@NotNull final Throwable originalException,
+                @Nullable final Entry sourceEntry) {
+            // We will just record the error here for now. If the table was removed, then we don't actually care about
+            // it
+            // but if the error is real, and the table is not removed, it will be propagated in processExisting()
```
rcaudy, line 220: The suggestion is just cleanup, but I think the comment and code could use a re-write.

```suggestion
        protected void onFailureInternal(
                @NotNull final Throwable originalException,
                @Nullable final Entry sourceEntry) {
            // We will just record the error here for now. If the table was removed, then we don't actually care about
            // it but if the error is real, and the table is not removed, it will be propagated in processExisting()
```

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
```diff
@@ -411,13 +436,23 @@ else if (nextCurrentKey == nextModifiedKey) {
...
+                    }
                     advanceCurrent();
                     ++nextCurrentSlot;
                     ++nextPreviousSlot;
                 }
             }
 
+            if (constituentExceptions != null) {
+                throw new UncheckedDeephavenException(
+                        MultiException.maybeWrapInMultiException("Errors occurred processing constituent tables:",
```
rcaudy, line 452:
```suggestion
                        MultiException.maybeWrapInMultiException("Constituent tables reported failures",
```

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
```diff
@@ -411,13 +436,23 @@ else if (nextCurrentKey == nextModifiedKey) {
...
+                        constituentExceptions = collectConstituentException(constituentExceptions, ex);
+                    }
                     advanceCurrent();
                     ++nextCurrentSlot;
                     ++nextPreviousSlot;
                 }
             }
 
+            if (constituentExceptions != null) {
+                throw new UncheckedDeephavenException(
```
rcaudy, line 451: This is bad. You're throwing to L350 in `MergedListener`, which is going to add some nonsense about uncaught exceptions.
You want to make `MergedListener.propagateError` `protected` and call that. Don't wrap in an `UncheckedDeephavenException` either.

rcaudy (changes requested):
I think this is fundamentally flawed. You will not be able to properly handle previous data from the removed table location when processing a remove, which makes this entire effort kind of moot.

I think there's utility in adding proper removal propagation, if it's corrected to not be part of the interface.

rcaudy:
Since the implementations of removal are out of scope for this PR (and for core at this time), I think we can rehabilitate the PR if we ensure that the `removeTableLocation` JavaDoc clearly states that implementations must provide certain guarantees:
```Reads should only succeed against removed locations if they will return complete, correct, consistent data. Otherwise, they should fail with a meaningful error message.```

File: engine/table/src/main/java/io/deephaven/engine/table/impl/SourcePartitionedTable.java (hunk 10 above)
Comment (abaranec) [reply engine/table/src/main/java/io/deephaven/engine/table/impl/SourcePartitionedTable.java:196]: I'm going to skip this for now -- I am worried that i don't understand it well enough to not make a surprising mistake.

File: engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java (hunk 3 above)
Comment (cpwright) [reply engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java:228]: The source table may not double notify.  In the case that the source table of the ListenerRecorder notifies us of an error, we should correctly set the notification step and record the error - the listener did have a notification on that step.  Given that there is an error there will never again be a valid notification from that source table.

The update in this case is going to be null; and if we ask the question "did this get updated on the current step" and then fetch and dereference the null update as is done on line 319-326 of MergedListener we have an error.  The change to the recordedVariablesAreValid (which is just asking about the step) means that we get the appropriate empty update in this section of the MergedListener.

File: engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/TableLocationSubscriptionBuffer.java [hunk 14]
```diff
@@ -4,32 +4,55 @@
 package io.deephaven.engine.table.impl.locations.impl;
 
 import io.deephaven.base.verify.Require;
-import io.deephaven.engine.table.impl.locations.ImmutableTableLocationKey;
-import io.deephaven.engine.table.impl.locations.TableDataException;
-import io.deephaven.engine.table.impl.locations.TableLocationProvider;
+import io.deephaven.engine.table.impl.locations.*;
```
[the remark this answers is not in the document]
rcaudy: Just consistency. I like wildcards, some other devs like expanded, but unnecessary change is perturbing.

engine/table/src/test/java/io/deephaven/engine/table/impl/SourcePartitionedTableTest.java
```diff
@@ -33,9 +32,13 @@ public class SourcePartitionedTableTest extends RefreshingTableTestCase {
     private ExtraControlledUpdateGraph updateGraph;
     private SafeCloseable myContext;
 
-    private static final class ExtraControlledUpdateGraph extends ControlledUpdateGraph {
+    private static final class ExtraControlledUpdateGraph extends PeriodicUpdateGraph {
```
rcaudy, line 35: This is still not what I was looking for; I don't want to make "real" update graphs if we can help it. The following patch is what I think we should do, although it's a lot of code:
```
Index: engine/table/src/test/java/io/deephaven/engine/table/impl/SourcePartitionedTableTest.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/engine/table/src/test/java/io/deephaven/engine/table/impl/SourcePartitionedTableTest.java b/engine/table/src/test/java/io/deephaven/engine/table/impl/SourcePartitionedTableTest.java
--- a/engine/table/src/test/java/io/deephaven/engine/table/impl/SourcePartitionedTableTest.java	(revision c1cf1bd7999ecb032b23afe4fbbe1fb86407c523)
+++ b/engine/table/src/test/java/io/deephaven/engine/table/impl/SourcePartitionedTableTest.java	(date 1694127898688)
@@ -1,27 +1,35 @@
 package io.deephaven.engine.table.impl;
 
+import io.deephaven.base.log.LogOutput;
 import io.deephaven.engine.table.Table;
 import io.deephaven.engine.table.impl.locations.ImmutableTableLocationKey;
 import io.deephaven.engine.table.impl.locations.TableLocationRemovedException;
 import io.deephaven.engine.table.impl.sources.ConstituentTableException;
+import io.deephaven.engine.testutil.ControlledUpdateGraph;
 import io.deephaven.engine.testutil.locations.DependentRegistrar;
 import io.deephaven.engine.testutil.locations.TableBackedTableLocationProvider;
 import io.deephaven.engine.testutil.testcase.RefreshingTableTestCase;
-import io.deephaven.engine.updategraph.impl.PeriodicUpdateGraph;
+import io.deephaven.engine.updategraph.LogicalClock;
+import io.deephaven.engine.updategraph.NotificationAdapter;
+import io.deephaven.engine.updategraph.UpdateGraph;
 import io.deephaven.engine.util.TableTools;
+import io.deephaven.io.log.LogEntry;
 import io.deephaven.io.logger.StreamLoggerImpl;
 import io.deephaven.test.types.OutOfBandTest;
 import io.deephaven.util.FindExceptionCause;
 import io.deephaven.util.SafeCloseable;
+import io.deephaven.util.locks.AwareFunctionalLock;
 import io.deephaven.util.process.ProcessEnvironment;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 
 import static io.deephaven.engine.testutil.TstUtils.*;
 import static io.deephaven.engine.util.TableTools.col;
@@ -29,25 +37,133 @@
 @Category(OutOfBandTest.class)
 public class SourcePartitionedTableTest extends RefreshingTableTestCase {
 
-    private ExtraControlledUpdateGraph updateGraph;
-    private SafeCloseable myContext;
+    private CapturingUpdateGraph updateGraph;
+    private SafeCloseable contextCloseable;
+
+    private static final class CapturingUpdateGraph implements UpdateGraph {
 
-    private static final class ExtraControlledUpdateGraph extends PeriodicUpdateGraph {
-        final List<Runnable> sources = new ArrayList<>();
+        private final ControlledUpdateGraph delegate;
 
-        private ExtraControlledUpdateGraph() {
-            super("TEST", true, 1000, 25, -1);
+        private final ExecutionContext context;
+
+        private final List<Runnable> sources = new ArrayList<>();
+
+        private CapturingUpdateGraph(@NotNull final ControlledUpdateGraph delegate) {
+            this.delegate = delegate;
+            context = ExecutionContext.getContext().withUpdateGraph(this);
         }
 
         @Override
         public void addSource(@NotNull Runnable updateSource) {
-            super.addSource(updateSource);
+            delegate.addSource(updateSource);
             sources.add(updateSource);
         }
 
-        public void refreshSources() {
+        private void refreshSources() {
             sources.forEach(Runnable::run);
         }
+
+        @Override
+        public LogOutput append(LogOutput logOutput) {
+            return logOutput.append("CapturingUpdateGraph of ").append(delegate);
+        }
+
+        @Override
+        public boolean satisfied(final long step) {
+            return delegate.satisfied(step);
+        }
+
+        @Override
+        public UpdateGraph getUpdateGraph() {
+            return this;
+        }
+
+        private Notification wrap(@NotNull final Notification notification) {
+            return new NotificationAdapter(notification) {
+                @Override
+                public void run() {
+                    try (final SafeCloseable ignored = context.open()) {
+                        super.run();
+                    }
+                }
+            };
+        }
+
+        @Override
+        public void addNotification(@NotNull final Notification notification) {
+            delegate.addNotification(wrap(notification));
+        }
+
+        @Override
+        public void addNotifications(@NotNull final Collection<? extends Notification> notifications) {
+            delegate.addNotifications(notifications.stream().map(this::wrap).collect(Collectors.toList()));
+        }
+
+        @Override
+        public boolean maybeAddNotification(@NotNull final Notification notification, final long deliveryStep) {
+            return delegate.maybeAddNotification(wrap(notification), deliveryStep);
+        }
+
+        @Override
+        public String getName() {
+            return "CapturingUpdateGraph";
+        }
+
+        @Override
+        public AwareFunctionalLock sharedLock() {
+            return delegate.sharedLock();
+        }
+
+        @Override
+        public AwareFunctionalLock exclusiveLock() {
+            return delegate.exclusiveLock();
+        }
+
+        @Override
+        public LogicalClock clock() {
+            return delegate.clock();
+        }
+
+        @Override
+        public int parallelismFactor() {
+            return delegate.parallelismFactor();
+        }
+
+        @Override
+        public LogEntry logDependencies() {
+            return delegate.logDependencies();
+        }
+
+        @Override
+        public boolean currentThreadProcessesUpdates() {
+            return delegate.currentThreadProcessesUpdates();
+        }
+
+        @Override
+        public boolean serialTableOperationsSafe() {
+            return delegate.serialTableOperationsSafe();
+        }
+
+        @Override
+        public boolean setSerialTableOperationsSafe(final boolean newValue) {
+            return delegate.serialTableOperationsSafe();
+        }
+
+        @Override
+        public boolean supportsRefreshing() {
+            return delegate.supportsRefreshing();
+        }
+
+        @Override
+        public void requestRefresh() {
+            delegate.requestRefresh();
+        }
+
+        @Override
+        public void removeSource(@NotNull Runnable updateSource) {
+            sources.remove(updateSource);
+            delegate.removeSource(updateSource);
+        }
     }
 
     @Override
@@ -59,16 +175,13 @@
         super.setUp();
         setExpectError(false);
 
-        updateGraph = new ExtraControlledUpdateGraph();
-        updateGraph.enableUnitTestMode();
-        updateGraph.resetForUnitTests(false);
-        updateGraph.setSerialTableOperationsSafe(true);
-        myContext = ExecutionContext.getContext().withUpdateGraph(updateGraph).open();
+        updateGraph = new CapturingUpdateGraph(ExecutionContext.getContext().getUpdateGraph().cast());
+        contextCloseable = updateGraph.context.open();
     }
 
     @Override
     public void tearDown() throws Exception {
-        myContext.close();
+        contextCloseable.close();
         super.tearDown();
     }
 
@@ -122,7 +235,7 @@
         tlp.removeTableLocationKey(tlks[0]);
         tlp.refresh();
 
-        allowingError(() -> updateGraph.runWithinUnitTestCycle(() -> {
+        allowingError(() -> updateGraph.delegate.runWithinUnitTestCycle(() -> {
             updateGraph.refreshSources();
             registrar.run();
         }), errors -> errors.size() == 1 &&
@@ -135,7 +248,7 @@
 
         tlp.addPending(p3);
         tlp.refresh();
-        updateGraph.runWithinUnitTestCycle(() -> {
+        updateGraph.delegate.runWithinUnitTestCycle(() -> {
             updateGraph.refreshSources();
             registrar.run();
         });
@@ -148,7 +261,7 @@
         tlp.removeTableLocationKey(tlks[0]);
         tlp.refresh();
 
-        allowingError(() -> updateGraph.runWithinUnitTestCycle(() -> {
+        allowingError(() -> updateGraph.delegate.runWithinUnitTestCycle(() -> {
             updateGraph.refreshSources();
             registrar.run();
         }), errors -> errors.size() == 1 &&
@@ -163,8 +276,8 @@
         // The TableBackedTableLocation has a copy() of the p3 table which is itself a leaf. Erroring P3 will
         // cause one error to come from the copied table, and one from the merged() table. We just need to validate
         // that the exceptions we see are a ConstituentTableException and an ISE
-        allowingError(() -> updateGraph.runWithinUnitTestCycle(
-                () -> p3.notifyListenersOnError(new IllegalStateException("This is a test error"), null)),
+        allowingError(() -> updateGraph.delegate.runWithinUnitTestCycle(
+                        () -> p3.notifyListenersOnError(new IllegalStateException("This is a test error"), null)),
                 errors -> errors.stream()
                         .anyMatch(e -> FindExceptionCause.findCause(e,
                                 IllegalStateException.class) instanceof IllegalStateException)
```

engine/table/src/main/java/io/deephaven/engine/table/impl/SourceTable.java
```diff
@@ -8,6 +8,7 @@
 import io.deephaven.engine.rowset.TrackingWritableRowSet;
 import io.deephaven.engine.table.Table;
 import io.deephaven.engine.table.TableDefinition;
+import io.deephaven.engine.table.impl.locations.TableLocationKey;
```
rcaudy, line 11: Unused

engine/table/src/main/java/io/deephaven/engine/table/impl/SourceTable.java
```diff
@@ -139,8 +140,7 @@ protected final void initializeAvailableLocations() {
                             new TableLocationSubscriptionBuffer(locationProvider);
                     final TableLocationSubscriptionBuffer.LocationUpdate locationUpdate =
                             locationBuffer.processPending();
-                    throwIfLocationsRemoved(locationUpdate);
-
+                    maybeRemoveLocations(locationUpdate.getPendingRemovedLocationKeys());
```
rcaudy, line 143: This is confusing. We just made the subscription buffer. It could have removals, I suppose, but none of those locations would have been added to the RCSM yet.

There could be cases where we have the same location as a pending add and a pending remove. I think the correct thing to do is process adds before removes, ironically. Don't forget to apply the same change in the refresh case.

In the long run, we'll need version stamps to tell if we're removing something we actually care about, or not.

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/AbstractTableLocationProvider.java
```diff
@@ -211,30 +211,37 @@ public TableLocation getTableLocationIfPresent(@NotNull final TableLocationKey t
...
+     *          causing them to stop updating. Tables backed by those notifications will end up 'failed'. This location
+     *          provider will continue to update other locations and will no longer provide or request information about
+     *          the deleted location.
+     *
+     * @param locationKey the TableLocation to remove
+     */
     public void removeTableLocationKey(@NotNull final TableLocationKey locationKey) {
         final Object removedLocation = tableLocations.remove(locationKey);
 
+        handleTableLocationKeyRemoved(locationKey.makeImmutable());
```
rcaudy, line 227: Should we validate that we have a non-null `removedLocation` here? Or just NPE?

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/AbstractTableLocationProvider.java
```diff
@@ -211,30 +211,37 @@ public TableLocation getTableLocationIfPresent(@NotNull final TableLocationKey t
...
         // need to notify subscribers of removed location, and of "removed" size
```
rcaudy, line 228: This comment should be refined. It now partially applies to the line above, and partially to the line below. Maybe just delete it, it's kind of low value.

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/AbstractTableLocationProvider.java
```diff
@@ -211,30 +211,37 @@ public TableLocation getTableLocationIfPresent(@NotNull final TableLocationKey t
...
+            abstractLocation.handleUpdate(null, System.currentTimeMillis());
+            abstractLocation.clearColumnLocations();
         }
     }
 
     /**
-     * Remove the location, and notify subscribers that it is gone
-     *
-     * @param location the TableLocation to be removed
+     * Notify subscribers that {@code location} was removed.
```
rcaudy, line 237:
```suggestion
     * Notify subscribers that {@code locationKey} was removed.
```

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/impl/AbstractTableLocationProvider.java
```diff
@@ -211,30 +211,37 @@ public TableLocation getTableLocationIfPresent(@NotNull final TableLocationKey t
...
-    @Override
+    /**
+     * Remove the given table location.
+     *
+     * @apiNote Use with caution: When this is called all subscribers to the specified location will be notified,
+     *          causing them to stop updating. Tables backed by those notifications will end up 'failed'. This location
+     *          provider will continue to update other locations and will no longer provide or request information about
+     *          the deleted location.
+     *
+     * @param locationKey the TableLocation to remove
```
rcaudy, line 222: This is missing our discussed warnings, and otherwise is a bit off-message for this layer.
```suggestion
     * Remove a {@link TableLocationKey} and its corresponding {@link TableLocation} (if it was created). All
     * subscribers to this TableLocationProvider will be
     * {@link TableLocationProvider.Listener#handleTableLocationKeyRemoved(ImmutableTableLocationKey) notified}. If the
     * TableLocation was created, all of its subscribers will additionally be
     * {@link TableLocation.Listener#handleUpdate() notified} that it no longer exists. This TableLocationProvider will
     * continue to update other locations and will no longer provide or request information about the removed location.
     * 
     * <p>
     * In practice, there are three downstream patterns in use.
     * <ol>
     * <li>Intermediate TableLocationProviders, which will simply pass the removal on</li>
     * <li>{@link io.deephaven.engine.table.impl.SourceTable SourceTables}, which will propagate a failure notification
     * to all downstream listeners, and become {@link Table#isFailed() failed}.</li>
     * <li>{@link io.deephaven.engine.table.impl.SourcePartitionedTable SourcePartitionedTables}, which will notify
     * their downstream consumers of the removed constituent.</li>
     * </ol>
     * 
     * @apiNote <em>Use with caution!</em> Implementations that call this method must provide certain guarantees: Reads
     *          should only succeed against removed locations if they will return complete, correct, consistent data.
     *          Otherwise, they should fail with a meaningful error message.
     * 
     * @param locationKey The {@link TableLocationKey} to remove
```

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/TableLocationRemovedException.java
```diff
@@ -2,6 +2,9 @@
 
 import org.jetbrains.annotations.NotNull;
 
+/**
+ * This exception is thrown when a {@link TableLocation} has been removed from a provider.
```
rcaudy, line 6:
```suggestion
 * This exception is thrown when a {@link TableLocation} that was in use by a {@link Table} is removed.
```

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/TableLocationProvider.java
```diff
@@ -38,13 +36,13 @@ interface Listener extends BasicTableDataListener {
         void handleTableLocationKey(@NotNull ImmutableTableLocationKey tableLocationKey);
 
         /**
-         * Notify the listener of a table location that has been removed encountered while initiating or maintaining the
-         * location subscription. This should occur at most once per location, but the order of delivery is <i>not</i>
-         * guaranteed.
+         * Notify the listener of a {@link TableLocationKey} that has been removed encountered while initiating or
```
rcaudy, line 39: I think this has some copy-paste errors.
```
         * Notify the listener of a {@link TableLocationKey} that has been removed.
         *
         * @param tableLocationKey The table location key that was removed
```

engine/table/src/main/java/io/deephaven/engine/table/impl/SourcePartitionedTable.java
```diff
@@ -248,45 +248,46 @@ private RowSet processAdditions(final TableLocationSubscriptionBuffer.LocationUp
...
 
+            // At the end of the cycle we need to make sure we unmanage any removed constituents.
+            this.removedConstituents = new ArrayList<>(relevantRemovedLocations.size());
+            this.removedLocationsComitter = new UpdateCommitter<>(
+                    this,
+                    result.getUpdateGraph(),
+                    ignored -> {
+                        removedConstituents.forEach(result::unmanage);
+                        removedConstituents = null;
+                    });
```
rcaudy, line 269: The pattern is to create this exactly once, and then call `.maybeActivate()` on any cycle you need it to run for.

engine/table/src/main/java/io/deephaven/engine/table/impl/SourcePartitionedTable.java
```diff
@@ -248,45 +248,46 @@ private RowSet processAdditions(final TableLocationSubscriptionBuffer.LocationUp
...
+            this.removedLocationsComitter.maybeActivate();
```
rcaudy, line 270: I would rather have you activate this once you've finished populating `removedConstituents` if its non-empty.

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/ColumnRegion.java
```diff
@@ -18,6 +18,13 @@ default long firstRowOffset() {
         return 0;
     }
 
+    //TODO: EXPERIMENT:
```
rcaudy: Clean up

engine/table/src/main/java/io/deephaven/engine/table/impl/SourceTable.java
```diff
@@ -160,12 +159,14 @@ private void maybeAddLocations(@NotNull final Collection<ImmutableTableLocationK
                 .forEach(lk -> columnSourceManager.addLocation(locationProvider.getTableLocation(lk)));
     }
 
-    private void maybeRemoveLocations(@NotNull final Collection<ImmutableTableLocationKey> removedKeys) {
+    private ImmutableTableLocationKey[] maybeRemoveLocations(@NotNull final Collection<ImmutableTableLocationKey> removedKeys) {
         if (removedKeys.isEmpty()) {
-            return;
+            return new ImmutableTableLocationKey[0];
```
rcaudy, line 165: Constant

engine/table/src/main/java/io/deephaven/engine/table/impl/SourceTable.java
```diff
@@ -140,7 +137,9 @@ protected final void initializeAvailableLocations() {
                             new TableLocationSubscriptionBuffer(locationProvider);
                     final TableLocationSubscriptionBuffer.LocationUpdate locationUpdate =
                             locationBuffer.processPending();
-                    maybeRemoveLocations(locationUpdate.getPendingRemovedLocationKeys());
+
+                    // We do not need to consider removed locations here -- we are initializing, so we haven't
```
rcaudy, line 141: Disagree

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/ColumnRegion.java
```diff
@@ -18,6 +18,13 @@ default long firstRowOffset() {
...
+
+    /**
+     * Poison the region -- any further reads will fail.
+     */
+    void poison();
```
rcaudy, line 24: "invalidate" - this is user-facing, better naming
This implementation will work, but it's kind of aggressive. We only need to prevent new reads; reads from cache are totally cool.

Util/src/main/java/io/deephaven/util/FindExceptionCause.java
```diff
@@ -27,36 +27,45 @@ public static Exception findCause(Exception original, Class<? extends Exception>
     }
 
     /**
-     * Given an exception and a list of expected exception types, traverse the cause tree and return the first exception
+     * Given a throwable and a list of expected throwable types, traverse the cause tree and return the first exception
      * that matches the list of expected cause types.
      */
-    public static <E extends Exception> Optional<E> findCause(@NotNull final Exception original, @NotNull final Class<E> expectedType) {
+    @SafeVarargs
+    public static Throwable findCause(Throwable original, Class<? extends Throwable>... expectedTypes) {
```
rcaudy, line 34: You say you don't need this, so revert it.

Util/src/main/java/io/deephaven/util/FindExceptionCause.java
```diff
@@ -27,36 +27,45 @@ public static Exception findCause(Exception original, Class<? extends Exception>
...
-    public static Throwable findCause(Throwable original, Class<? extends Throwable>... expectedTypes) {
+    public static <E extends Throwable> Optional<E> isOrCausedBy(@NotNull final Throwable original,
+            @NotNull final Class<E> expectedType) {
+        if (expectedType.isAssignableFrom(original.getClass())) {
+            return Optional.of((E) original);
+        }
+
         Throwable cause = original.getCause();
         while (cause != null) {
             final Throwable checkCause = cause;
```
rcaudy, line 62: Why do I need `checkCause`?

engine/table/src/main/java/io/deephaven/engine/table/impl/locations/TableLocationRemovedException.java
```diff
@@ -9,7 +9,8 @@
 public class TableLocationRemovedException extends TableDataException {
     private final ImmutableTableLocationKey[] locationKeys;
 
-    public TableLocationRemovedException(@NotNull final String message, @NotNull final ImmutableTableLocationKey... locationKeys) {
+    public TableLocationRemovedException(@NotNull final String message,
+            @NotNull final ImmutableTableLocationKey... locationKeys) {
```
rcaudy, line 13:
```suggestion
    public TableLocationRemovedException(
            @NotNull final String message,
            @NotNull final ImmutableTableLocationKey... locationKeys) {
```

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/RegionedColumnSourceManager.java
```diff
@@ -332,68 +332,69 @@ private void pollUpdates(final RowSetBuilderSequential addedRowSetBuilder) {
...
+                    rowSetAtLastUpdate.close();
+                    rowSetAtLastUpdate = updateRowSet;
                 }
-            } finally {
-                rowSetAtLastUpdate.close();
-                rowSetAtLastUpdate = updateRowSet;
+            } catch (Exception ex) {
+                poison();
+                throw ex;
             }
```
rcaudy, line 393: I think this is heavy-handed and unnecessary.

rcaudy (commented):
* In the buffer:
** If you see a remove: if there’s a pending add, delete it, and ignore the remove
* In consumers of the buffer:
** Always process removes before adds

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/RegionedColumnSourceManager.java
rcaudy: In the words of the immortal bard Sheryl Crow: If it makes you happy, it can't be that bad.

engine/table/src/main/java/io/deephaven/engine/table/impl/SourcePartitionedTable.java
```diff
@@ -92,7 +93,7 @@ private static final class UnderlyingTableMaintainer {
         @SuppressWarnings("FieldCanBeLocal") // We need to hold onto this reference for reachability purposes.
         private final Runnable processNewLocationsUpdateRoot;
 
-        private UpdateCommitter<UnderlyingTableMaintainer> removedLocationsComitter = null;
+        final private UpdateCommitter<UnderlyingTableMaintainer> removedLocationsComitter;
```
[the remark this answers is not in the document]
abaranec: It was a full moon and I forgot to bring an umbrella

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/region/ParquetColumnRegionBase.java
```diff
@@ -55,11 +59,13 @@ public final void fillChunkAppend(
             @NotNull final FillContext context,
             @NotNull final WritableChunk<? super ATTR> destination,
             @NotNull final RowSequence.Iterator rowSequenceIterator) {
+        throwIfPoisioned();
         columnChunkPageStore.fillChunkAppend(context, destination, rowSequenceIterator);
     }
 
     @Override
     public final ChunkPage<ATTR> getChunkPageContaining(final long elementIndex) {
+        throwIfPoisioned();
```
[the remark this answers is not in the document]
abaranec: I think that we agreed to stay aggressive between @rcaudy @cpwright and @abaranec

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/ColumnRegion.java
abaranec: I think that we agreed to stay aggressive between @rcaudy @cpwright and @abaranec

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/regioned/GenericColumnRegionBase.java
```diff
@@ -24,13 +24,13 @@ public final long mask() {
...
+    public void invalidate() {
+        this.invalidated = true;
     }
 
-    protected final void throwIfPoisoned() {
-        if (poisoned) {
-            throw new PoisonedRegionException("Regioned is poisoned");
+    protected final void throwIfInvalidated() {
+        if (invalidated) {
+            throw new InvalidatedRegionException("Regioned has been invalidated");
```
rcaudy, line 33:
```suggestion
            throw new InvalidatedRegionException("Column region has been invalidated due to data removal");
```

PR #4387, opened 2023-08-28
Run Python tests under liveness scopes

### LINKED ISSUE #4386: Seemingly random check CI failure due to OOM from the Kafka test cases
@alexpeters1208 made a R client PR but got an unrelated check CI failure.
```
   test_vectorized_const_arg (tests.test_vectorization.VectorizationTestCase) ... Exception in thread "KafkaIngestershare_price:ALL"
  Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "KafkaIngestershare_price:ALL"
  java.lang.OutOfMemoryError: Java heap space
  	at org.apache.kafka.clients.consumer.internals.Fetcher.prepareFetchRequests(Fetcher.java:1184)
  	at org.apache.kafka.clients.consumer.internals.Fetcher.sendFetches(Fetcher.java:252)
  	at org.apache.kafka.clients.consumer.KafkaConsumer.pollForFetches(KafkaConsumer.java:1297)
  	at org.apache.kafka.clients.consumer.KafkaConsumer.poll(KafkaConsumer.java:1247)
  	at 
```

The trace tells us that there is probably some delayed cleaning up of the kafka ingester in the Kafka tests. Explicit liveness management could be the solution.

jmao-denver:
Fixes #4386

### DIFF
--- a/py/server/tests/test_liveness.py
+++ b/py/server/tests/test_liveness.py
@@ -13,6 +13,7 @@ from deephaven.liveness_scope import liveness_scope
 from deephaven.update_graph import exclusive_lock
 from tests.testbase import BaseTestCase
 
+
 class LivenessTestCase(BaseTestCase):
 
     def setUp(self) -> None:
@@ -52,6 +53,8 @@ class LivenessTestCase(BaseTestCase):
             l_scope.preserve(must_keep)
             l_scope.close()
             l_scope_2.close()
+        l_scope_2.close()
+        l_scope.close()
 
     def test_liveness_nested(self):
         with liveness_scope() as l_scope:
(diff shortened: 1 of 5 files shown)


py/server/tests/test_liveness.py
```diff
@@ -52,6 +53,8 @@ def test_liveness(self):
             l_scope.preserve(must_keep)
             l_scope.close()
             l_scope_2.close()
+        l_scope_2.close()
+        l_scope.close()
```
niloc132, line 57: to confirm, the preceding `with` should throw, so neither close() call in that block can be invoked? If so, why even have the close() calls there, if they are effectively "dead code"?

jmao-denver: It is these calls (out of order) that will throw. The added lines are just to correctly pop the liveness scopes so that the tearDown won't complain.

rcaudy (approved):

PR #4388, opened 2023-08-28
Transition to python using native python time types instead of Java types

### LINKED ISSUE #4214: Decommission time.py from the server Python API
Reasons:
1. After switching from home-grown DH DateTime type to Java's own Instant and siblings to store date/time data in DH, the utility provided by time.py could be achieved by 3rd party/DH built conversion support between Java Instant and Python DateTime and Python's own DateTIme manipulation libraries. 
2. A number of the wrapper functions in time.py, including lower_bin(), upper_bin(), now(), when used in queries, make the execution 20x+ slower because of the double-crossing between JVM/Python. The effort to automatically substitute them with their Java camelCase counterparts has proven to be messy and unsatisfactory. 
3. Since these utility functions should not be considered as typical Python UDFs, but rather part of the built-ins of DH query language, it is arguably better/more acceptable to clearly document them and educate the user on how to use them than for purist reasons to wrap them in Python but making it confusing for the users and trying to solve the performance problem that could be avoided otherwise.

This would also render following #2777, #2306, #2303 non-issues.

chipkent:
Transition from using Java time types in python to using python time types in python.

Fixes #4214


py/server/deephaven/time.py
```diff
@@ -155,1885 +135,485 @@ def time_zone_alias_rm(alias: str) -> bool:
...
-        return _JDateTimeUtils.microsToMillis(micros)
+        if dt is None:
+            return None
+        elif isinstance(dt, str):
+            return _JDateTimeUtils.parseLocalTime(dt)
+        elif isinstance(dt, datetime.time) or isinstance(dt, datetime.datetime):
+            return _JLocalTime.of(dt.hour, dt.minute, dt.second, dt.microsecond * 1000)
+        elif isinstance(dt, np.datetime64):
+            # Conversion only supports micros resolution
+            return to_j_local_time(dt.astype(datetime.time))
```
jmao-denver, line 239: Not sure about the lost precision here. How about converting to pandas.Timestamp first? Also should we consider adding pandas.Timestamp to the supported types in those to_j_xxx functions?

py/server/deephaven/time.py
```diff
@@ -70,13 +63,16 @@ def now(system: bool = False, resolution: str = 'ns') -> Instant:
         raise DHError(e) from e
 
 
-def today(tz: TimeZone) -> str:
-    """ Provides the current date string according to the current clock.
+# TODO: rename to system_today or dh_now?  I lean to today.
+# TODO: return string or a python date type?  I lean to string.
```
jmao-denver, line 67: My vote is to remove. Its name clashes with Python built-in `datetime.date.today`.

py/server/deephaven/time.py
```diff
@@ -155,1885 +135,485 @@ def time_zone_alias_rm(alias: str) -> bool:
...
     """
-    if not dt:
-        return NULL_LONG
-
     try:
-        return _JDateTimeUtils.epochMicros(dt)
+        if dt is None:
+            return None
+        elif isinstance(dt, Instant.j_type):
+            ts = dt.getEpochSecond() + (dt.getNano() / 1000000000)
```
devinrsmith, line 504: Worth storing 1000000000 as constant and using in all the places?

py/server/tests/test_time.py
```diff
@@ -56,825 +35,441 @@ def test_today(self):
...
-        timezone_str = "ET"
-        tz = time_zone(timezone_str)
-        dt = parse_instant(f"{datetime_str} {timezone_str}")
-        dt_str = format_datetime(dt, tz)
-        self.assertEqual(f"{datetime_str} {timezone_str}", dt_str)
+        try:
+            to_j_local_date(123)
+            self.fail("Expected DHError")
+        except DHError:
+            pass
```
jmao-denver, line 121: `with self.assertRaises(DHError):` is more idiomatic.

jmao-denver (changes requested):

py/server/deephaven/time.py
```diff
@@ -1,43 +1,36 @@
...
+_JDuration = jpy.get_type("java.time.Duration")
+_JPeriod = jpy.get_type("java.time.Period")
+_epoch64 = np.datetime64('1970-01-01T00:00:00Z')
 
 
 # region Clock
 
-
+# TODO: rename to system_now or dh_now?  I lean to now.
+# TODO: return Instant or a python type?  I lean to Instant.
```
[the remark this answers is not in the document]
chipkent: We need to have it available.  Without it, replay is not very useful.  When replaying, a script needs to be able to get the current date and time to function properly.  We can rename or whatever, but I don't think we can remove it.

py/server/deephaven/time.py
chipkent: We need to have it available. Without it, replay is not very useful. When replaying, a script needs to be able to get the current date and time to function properly. We can rename or whatever, but I don't think we can remove it.

py/server/deephaven/time.py
```diff
@@ -155,1885 +135,485 @@ def time_zone_alias_rm(alias: str) -> bool:
...
+        Instant
 
     Raises:
         DHError
     """
     try:
-        return _JDateTimeUtils.nanosToSeconds(nanos)
+        if dt is None:
+            return None
+        elif isinstance(dt, str):
```
cpwright, line 269: If you add a j_isntant check here; you can simplify many of the call sites.

py/server/deephaven/time.py
```diff
@@ -155,1885 +135,485 @@ def time_zone_alias_rm(alias: str) -> bool:
 
 # endregion
 
-# region Conversions: Time Units
-
+#TODO: Review all function names: to_j_<xyz>?  to_db_<xyz>?  to_dh_<xyz>?  to_<xyz>? as_j_<xyz>?  as_db_<xyz>?  as_dh_<xyz>?
```
alexpeters1208, line 138: I would vote `to_dh_<xyz>`.

py/server/deephaven/time.py
```diff
@@ -155,1885 +135,485 @@ def time_zone_alias_rm(alias: str) -> bool:
...
+#TODO: Should the doc strings refer to "Deephaven" or "Java" types?
```
alexpeters1208, line 139: I would think we want the docstrings to refer to Deephaven types. I think a user either understands that Deephaven types are implicitly Java types under the hood and doesn't need the docstring to remind them, or they don't care at all and will do just fine being shielded from all the Java stuff. This assumes that the API does not drag in anything that doesn't feel like a Deephaven type.

py/server/deephaven/time.py
```diff
@@ -155,1885 +135,485 @@ def time_zone_alias_rm(alias: str) -> bool:
...
-    Raises:
-        DHError
+def to_j_local_date(dt: Union[None, str, datetime.date, datetime.time, datetime.datetime, np.datetime64]) -> \
+        Optional[LocalDate]:
     """
-    try:
-        return _JDateTimeUtils.secondsToNanos(seconds)
-    except Exception as e:
-        raise DHError(e) from e
+    Converts a Python string, date, time, datetime, or numpy datetime64 to a Deephaven local date,
```
alexpeters1208, line 181: Name is `to_j_local_date`, but docstring says "converts ... to a Deephaven local date". These are not consistent, and I vote for `to_dh_local_date`

py/server/deephaven/time.py
```diff
@@ -155,1885 +135,485 @@ def time_zone_alias_rm(alias: str) -> bool:
...
-    Raises:
-        DHError
+def to_j_local_time(dt: Union[None, str, datetime.time, datetime.datetime, np.datetime64]) -> Optional[LocalTime]:
     """
-    try:
-        return _JDateTimeUtils.secondsToMicros(seconds)
-    except Exception as e:
-        raise DHError(e) from e
-
+    Converts a Python string, time, datetime, or numpy datetime64 to a Deephaven local time,
```
alexpeters1208, line 214: Same inconsistency comment as above

py/server/deephaven/time.py
alexpeters1208: I do think Pandas timestamps should be supported

py/server/deephaven/time.py
```diff
@@ -155,1885 +135,485 @@ def time_zone_alias_rm(alias: str) -> bool:
...
-    Raises:
-        DHError
+def to_j_instant(dt: Union[None, str, datetime.datetime, np.datetime64]) -> Optional[Instant]:
     """
-    try:
-        return _JDateTimeUtils.secondsToMillis(seconds)
-    except Exception as e:
-        raise DHError(e) from e
+    Converts a Python string, datetime, or numpy datetime64 to a Deephaven instant,
+    which is a point in time on the time-line.
```
alexpeters1208, line 249: Is "on the time-line" needed?

py/server/deephaven/time.py
```diff
@@ -155,1885 +135,485 @@ def time_zone_alias_rm(alias: str) -> bool:
...
-
-    Args:
-        dt (Union[Instant,ZonedDateTime]): Date time.
-
-    Returns:
-        milliseconds since Epoch, or a NULL_LONG value if the date time is null.
-
-    Raises:
-        DHError
+def to_datetime64(dt: Union[None, Instant, ZonedDateTime]) -> Optional[np.datetime64]:
```
alexpeters1208, line 515: `to_np_datetime64`?

alexpeters1208: If there are other datetime64-like things that are also in here. To me `to_datetime64` does not automatically tell me it will be numpy, but maybe I don't know Python well enough.

py/server/deephaven/time.py
```diff
@@ -155,1885 +135,485 @@ def time_zone_alias_rm(alias: str) -> bool:
...
+                return np.timedelta64(d, 'D')
+            else:
+                raise Exception("Unsupported conversion: " + str(type(dt)) + " -> datetime.timedelta64: (" + dt + ")")
+        else:
+            raise Exception("Unsupported conversion: " + str(type(dt)) + " -> datetime.timedelta64")
     except Exception as e:
         raise DHError(e) from e
 
 
-def epoch_seconds_to_instant(seconds: int) -> Instant:
```
alexpeters1208, line 662: Why is all this stuff going away? Does the new API somehow capture all of it? Is it no longer useful?

py/server/deephaven/time.py
```diff
@@ -155,1884 +148,594 @@ def time_zone_alias_rm(alias: str) -> bool:
 
 # endregion
 
-# region Conversions: Time Units
+# TODO: Review all function names: to_j_<xyz>?  to_db_<xyz>?  to_dh_<xyz>?  to_<xyz>? as_j_<xyz>?  as_db_<xyz>?  as_dh_<xyz>?
```
chipkent, line 151: todo

py/server/deephaven/time.py
```diff
@@ -155,1884 +148,594 @@ def time_zone_alias_rm(alias: str) -> bool:
...
+# TODO: Should the doc strings refer to "Deephaven" or "Java" types?
```
chipkent, line 152: todo

py/server/deephaven/time.py
```diff
@@ -85,21 +81,16 @@ def today(tz: TimeZone) -> str:
...
-
-# region Time Zone
-
-
-def time_zone(tz: Optional[str]) -> TimeZone:
-    """ Gets the time zone for a time zone name.
-
-    Args:
-        tz (Optional[str]): Time zone name.  If None is provided, the system default time zone is returned.
+def dh_time_zone() -> TimeZone:
```
jjbrosnan, line 92: Are we providing any way for users to specify a different Deephaven time zone?

jjbrosnan: Never mind. I see it down below.

py/server/deephaven/time.py
```diff
@@ -155,1884 +148,594 @@ def time_zone_alias_rm(alias: str) -> bool:
...
-    Args:
-        dt (Instant): Date time to convert.
-        tz (TimeZone): Time zone.
-
-    Returns:
-        ZonedDateTime or None if any input is None.
-
-    Raises:
-        DHError
+def to_j_period(dt: Union[None, Period, str, datetime.timedelta, numpy.timedelta64]) -> Optional[Period]:
```
jjbrosnan, line 433: How come [Pandas Timedelta](https://pandas.pydata.org/docs/reference/api/pandas.Timedelta.html) isn't included?

py/server/deephaven/time.py
```diff
@@ -155,1884 +148,594 @@ def time_zone_alias_rm(alias: str) -> bool:
...
-        dt (Union[Instant,ZonedDateTime]): Date time.
+        dt (Union[None, Instant, ZonedDateTime]): A Java date time.
+            If None is provided, None is returned.
 
     Returns:
-        seconds since Epoch, or a NULL_LONG value if the date time is null.
-
-    Raises:
-        DHError
+        pandaw.Timestamp
```
jjbrosnan, line 593: typo?
```suggestion
        pandas.Timestamp
```

jjbrosnan (commented):
Mostly LGTM, some small stuff.

py/server/deephaven/time.py
jmao-denver, line 92: We have a `get_server_timezone()` in `config/_init__.py`, should we remove it?

chipkent: I'm not sure.

py/server/deephaven/time.py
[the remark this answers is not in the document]
chipkent: Documented

py/server/deephaven/time.py
chipkent: I'm probably not opposed to removing it.

jmao-denver: let's do nothing for now.

jmao-denver (approved):
LGTM

jmao-denver (dismissed):

py/server/deephaven/table_factory.py
```diff
@@ -65,11 +72,19 @@ def time_table(period: Union[str, int], start_time: str = None, blink_table: boo
...
     try:
         builder = _JTableTools.timeTableBuilder()
+
+        if not isinstance(period, str) and not isinstance(period, int):
+            period = str(time.to_j_duration(period))
+
         builder.period(period)
+
         if start_time:
+            start_time = str(time.to_j_instant(start_time))
```
cpwright, line 82: The builder can accept a java Instant; it seems odd to wash that through a string, what am I missing?

py/server/deephaven/table_factory.py
```diff
@@ -65,11 +72,19 @@ def time_table(period: Union[str, int], start_time: str = None, blink_table: boo
     """
     try:
         builder = _JTableTools.timeTableBuilder()
+
+        if not isinstance(period, str) and not isinstance(period, int):
+            period = str(time.to_j_duration(period))
```
cpwright, line 77: The builder can accept a java Duration; it seems odd to wash that through a string, what am I missing?

py/server/deephaven/time.py
```diff
@@ -155,1884 +160,595 @@ def time_zone_alias_rm(alias: str) -> bool:
...
+            return _JDateTimeUtils.parseTimeZone(tz)
+        elif isinstance(tz, datetime.tzinfo):
+            return _JDateTimeUtils.parseTimeZone(str(tz))
+        elif isinstance(tz, datetime.datetime):
+            if not tz.tzname():
+                return _JDateTimeUtils.parseTimeZone(tz.astimezone().tzname())
+
+            return _JDateTimeUtils.parseTimeZone(tz.tzname())
+        else:
+            raise Exception("Unsupported conversion: " + str(type(tz)) + " -> TimeZone")
```
cpwright, line 200: Should we use DHError directly to avoid one level of wrapping for the unsupported conversions in these methods?

py/server/tests/test_arrow.py
```diff
@@ -6,6 +6,7 @@
 from datetime import datetime
 from typing import List, Any
 
+import jpy
```
cpwright, line 9: Why the need for import jpy?  I don't see any additions to the code; and if we are requiring it, shouldn't the module we are importing from do that for us?

py/server/tests/test_numpy.py
```diff
@@ -5,6 +5,7 @@
 import unittest
 from dataclasses import dataclass
 
+import jpy
```
cpwright, line 8: Same comment about the new jpy import as before.

py/server/tests/test_pandas.py
```diff
@@ -4,23 +4,22 @@
 
 import unittest
 from dataclasses import dataclass
+import jpy
```
cpwright, line 7: new jpy dep

py/server/deephaven/time.py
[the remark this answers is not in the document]
chipkent: The functions are currently there as `dh_xxx` with a documentation warning.  @rcaudy thinks they shouldn't have the `dh_prefix`.

py/server/deephaven/time.py
chipkent: It was determined that `to_j_` was more consistent with the rest of the API.

py/server/deephaven/time.py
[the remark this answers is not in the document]
chipkent: Discussion is to leave this as is.

py/server/deephaven/time.py
jmao-denver: I'd prefer the built-in `TypeError` which is more precise although I could live with `DHError`.

py/server/deephaven/time.py
```diff
@@ -64,7 +64,9 @@ def dh_now(system: bool = False, resolution: str = 'ns') -> Instant:
             else:
                 return _JDateTimeUtils.nowMillisResolution()
         else:
-            raise ValueError("Unsupported time resolution: " + resolution)
+            raise TypeError("Unsupported time resolution: " + resolution)
+    except TypeError as e:
+        raise e
```
jmao-denver, line 69: Just a comment, could be a bit simpler: 
```
except TypeError:
    raise
```

jmao-denver (approved):
LGTM

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3123
Reference: https://github.com/deephaven/deephaven.io/issues/3124

PR #4399, opened 2023-08-29
Correct initialization for Float/Double RollingMinMax feature.

### LINKED ISSUE #4396: `rolling_max_tick` behaves incorrectly on `float` columns
**Description**

When `rolling_max_tick` gets applied to a column of type `float`, if all of the observations in a given window are negative, the maximum is calculated to be `0`.

**Steps to reproduce**
Run the following code in the Deephaven IDE:
```
from deephaven.updateby import rolling_max_tick

test1 = empty_table(100).update("X = -1")
result1 = test1.update_by(ops = rolling_max_tick("X", rev_ticks = 3))

test2 = empty_table(100).update("X = -10.12345")
result2 = test2.update_by(ops = rolling_max_tick("X", rev_ticks = 3))
```

**Expected results**

I expect `result1` to be a table with a single column, whose entries are all `-1`, and I expect `result2` to be a table with a single column, whose entries are all `-10.12345`.

**Actual results**

`result1` is as it should be. However, all of the entries in `result2` are `0.0000`.

**Versions**
 - Deephaven: 0.28.0.
 - OS: MacOS 13.2.1
 - Browser: Safari 16.3

lbooker42:
Float and Double initial values were incorrectly set to Float.MIN_VALUE and Double.MIN_VALUE which are small positive numbers instead of the highest magnitude negative number.

Will close #4396

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingminmax/DoubleRollingMinMaxOperator.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingminmax/DoubleRollingMinMaxOperator.java
@@ -33,7 +33,7 @@ public class DoubleRollingMinMaxOperator extends BaseDoubleUpdateByOperator {
         protected Context(final int affectedChunkSize, final int influencerChunkSize) {
             super(affectedChunkSize);
             if (isMax) {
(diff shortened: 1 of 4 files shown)


alexpeters1208 (approved):
Tested from the R Client, LGTM

nbauernfeind (approved):

PR #4412, opened 2023-08-30
R Client - UpdateBy

alexpeters1208:
Introduces the UpdateBy operation and unit tests to the R client.


R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -0,0 +1,615 @@
...
+  verify_string("rev_time", rev_time, TRUE)
+  verify_string("fwd_time", fwd_time, TRUE)
+  return(UpdateByOp$new(INTERNAL_rolling_sum_time(ts_col, cols, rev_time, fwd_time)))
+}
+
+#' @description
+#' Creates a rolling group UpdateByOp for each column in `cols`, using ticks as the windowing unit.
+#' Ticks are row counts, and you may specify the reverse and forward window in number of rows to include.
+#' The current row is considered to belong to the reverse window but not the forward window.
+#' Also, negative values are allowed and can be used to generate completely forward or completely reverse windows.
```
chipkent, line 321: In python, all of our windowed operations have detailed examples for defining the windows.  I think we should have something similar in R, unless there is a major, compelling reason not to.  We did this because it was too confusing without examples.

e.g.
https://deephaven.io/core/pydoc/code/deephaven.updateby.html#deephaven.updateby.rolling_avg_tick
https://deephaven.io/core/pydoc/code/deephaven.updateby.html#deephaven.updateby.rolling_avg_time

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -0,0 +1,615 @@
...
+#' Creates a rolling sum UpdateByOp for each column in `cols`, using ticks as the windowing unit.
+#' Ticks are row counts, and you may specify the reverse and forward window in number of rows to include.
+#' The current row is considered to belong to the reverse window but not the forward window.
+#' Also, negative values are allowed and can be used to generate completely forward or completely reverse windows.
+#' @param cols String or list of strings denoting the column(s) to operate on. Can be renaming expressions, i.e. “new_col = col”.
+#' Default is to compute the rolling sum for all non-grouping columns.
+#' @param rev_ticks Integer scalar denoting the look-behind window size in number of rows.
+#' @param fwd_ticks Integer scalar denoting the look-ahead window size in number of rows. Default is 0.
+#' @export
+udb_roll_sum_tick <- function(cols, rev_ticks, fwd_ticks = 0) {
```
chipkent, line 290: In all of these, `rolling` was shortened to `roll`.  I'm not sure what I think.  It will make it harder to find similar docs from other langs, so I might bias to `rolling`, even though it is slightly longer.

R/rdeephaven/R/utility_functions.R
chipkent, line 1: The review tool is not showing a nice diff vs the old `helper_functions.R`.  If anything material has changed in the logic, let me know to review in detail.  My review of this file is superficial.

chipkent (changes requested):

R/rdeephaven/R/utility_classes.R
```diff
@@ -0,0 +1,48 @@
...
+      self$.internal_rcpp_object <- op_control
+    }
+  )
+)
+
+#' @description
+#' Creates an OperationControl that controls behavior on special values in `udb_em*` operations. The arguments `on_null`
+#' and `on_nan` can take the following values:
+#'   'poison': Allow bad data to poison the result. This is only valid for use with NaN.
+#'   'reset': Reset the state for the bucket to None when invalid data is encountered.
```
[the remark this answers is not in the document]
alexpeters1208: R supports `NULL`, `NA`, and `NaN`. I figured we pick the one that makes everything consistent.

R/rdeephaven/R/update_by_ops_wrapper.R
alexpeters1208: I think there is a good reason not to do this. These docstrings are not the literal characters that are going to get turned into R HTML docs. I am going to split the within-R-studio docs into reference-style pages and tutorial-style pages. Most within-R-studio documentation sort of shoves both of these things into one doc, but I think our stuff is sufficiently complicated to warrant not doing that. In that case, the tutorial-style doc will have the windowing examples you speak of. All R docs should also have actual examples demonstrating usage of the thing they're documenting, so we will have that as well. I think there is not a need to clutter these docstrings with examples, as they are really only intended to be read from source.

alexpeters1208: It's worth mentioning that the doc split I'm proposing breaks R convention and breaks the typical roxygen workflow. I'm ok with this, as both of those things are bad.

alexpeters1208: This also inclines me to shorten these window docs even further, to something like 
```
Creates a rolling group UpdateByOp for each column in 'cols', using ticks as the windowing unit.
Ticks are row counts, and you may specify the reverse and forward window in number of rows to include.
```

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -0,0 +1,615 @@
...
+    stop(paste0("'delta_control' must be one of 'null_dominates', 'value_dominates', or 'zero_dominates'. Got '", delta_control, "'."))
+  }
+  return(UpdateByOp$new(INTERNAL_delta(cols, delta_control)))
+}
+
+#' @description
+#' Creates an exponential moving average UpdateByOp for each column in `cols`, using ticks as the decay unit.
+#' The formula used is
+#'   `a = e^(-1 / decay_ticks)`
+#'   `ema_next = a * ema_last + (1 - a) * value`
```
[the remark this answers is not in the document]
alexpeters1208: I do not want to clutter these with the Latex syntax that the web docs are getting, because these aren't going to be seen in that format. I do think they could use the additional detail that the new docs provide.

R/rdeephaven/R/utility_classes.R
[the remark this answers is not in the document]
chipkent: But shouldn't it then be called `null` and not `None`?

R/rdeephaven/R/update_by_ops_wrapper.R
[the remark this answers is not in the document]
chipkent: It does look like R supports latex formatting as well.  https://stackoverflow.com/questions/37397607/can-r-help-manuals-have-latex-math-in-them

R/rdeephaven/R/update_by_ops_wrapper.R
chipkent: No part of this plan makes sense to me.

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -380,7 +392,7 @@ udb_roll_avg_tick <- function(cols, rev_ticks, fwd_ticks = 0) {
 #' @param rev_time ISO-8601-formatted string specifying the look-behind window size.
 #' @param fwd_time ISO-8601-formatted string specifying the look-ahead window size. Default is 0 seconds.
 #' @export
-udb_roll_avg_time <- function(ts_col, cols, rev_time, fwd_time = "PT0s") {
+uby_rolling_avg_time <- function(ts_col, cols, rev_time, fwd_time = "PT0s") {
```
chipkent, line 395: If this is a simple moving average, that and SMA should be indicated in the docstring.

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -361,7 +373,7 @@ udb_roll_group_time <- function(ts_col, cols, rev_time, fwd_time = "PT0s") {
 #' @param rev_ticks Integer scalar denoting the look-behind window size in number of rows.
 #' @param fwd_ticks Integer scalar denoting the look-ahead window size in number of rows. Default is 0.
 #' @export
-udb_roll_avg_tick <- function(cols, rev_ticks, fwd_ticks = 0) {
+uby_rolling_avg_tick <- function(cols, rev_ticks, fwd_ticks = 0) {
```
chipkent, line 376: If this is a simple moving average, that and SMA should be indicated in the docstring.

R/rdeephaven/R/update_by_ops_wrapper.R
chipkent: In particular, users almost never read all of the documentation.  They need what they want at the point they ask for it.  By removing the documentation from `?`, they will never see what they need to.

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -312,6 +323,17 @@ uby_rolling_sum_tick <- function(cols, rev_ticks, fwd_ticks = 0) {
...
+#'   `rev_time = 0, fwd_time = 0` - contains rows that exactly match the current row timestamp
+#'   `rev_time = "PT00:10:00", fwd_time = "0"` - contains rows from 10m before through the current row timestamp (inclusive)
+#'   `rev_time = 0, fwd_time = 600_000_000_000` - contains rows from the current row through 10m following the
+#'     current row timestamp (inclusive)
+#'   `rev_time = "PT00:10:00", fwd_time = "PT00:10:00"` - contains rows from 10m before through 10m following
+#'     the current row timestamp (inclusive)
+#'   `rev_time = "PT00:10:00", fwd_time = "-PT00:05:00"` - contains rows from 10m before through 5m before the
+#'     current row timestamp (inclusive), this is a purely backwards looking window
+#'   `rev_time = "-PT00:05:00", fwd_time = "PT00:10:00"` - contains rows from 5m following through 10m
+#'     following the current row timestamp (inclusive), this is a purely forwards looking window
```
chipkent, line 336: Do the non-string versions work with R?  Do the quoted integers work with R?

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -349,6 +382,17 @@ uby_rolling_group_tick <- function(cols, rev_ticks, fwd_ticks = 0) {
...
```
chipkent, line 395: Do the non-string versions work with R? Do the quoted integers work with R?

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -460,6 +559,17 @@ uby_rolling_max_tick <- function(cols, rev_ticks, fwd_ticks = 0) {
...
```
chipkent, line 572: Do the non-string versions work with R? Do the quoted integers work with R?

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -534,6 +677,17 @@ uby_rolling_count_tick <- function(cols, rev_ticks, fwd_ticks = 0) {
...
```
chipkent, line 690: Do the non-string versions work with R? Do the quoted integers work with R?

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -571,6 +736,17 @@ uby_rolling_std_tick <- function(cols, rev_ticks, fwd_ticks = 0) {
...
```
chipkent, line 749: Do the non-string versions work with R? Do the quoted integers work with R?

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -610,6 +797,17 @@ uby_rolling_wavg_tick <- function(wcol, cols, rev_ticks, fwd_ticks = 0) {
...
```
chipkent, line 810: Do the non-string versions work with R? Do the quoted integers work with R?

alexpeters1208:
Here is a ticket for continued discussion of R docs: https://github.com/deephaven/deephaven-core/issues/4573

chipkent (approved):

PR #4414, opened 2023-08-30
Fix kafka json byte and char types

### LINKED ISSUE #4384: Kafka Ingester gets Byte / Char columns wrong
**Description**

While testing an Enterprise feature. I discovered that when I write Byte and Char columns using the pk.produce method,  and consume with KafkaTools to an In Memory table. both Byte and Char columns are incorrect,  I don't know if the issue is on the production to kafka, or the Ingestion from Kafka side.

**Steps to reproduce**

Here is the Producer query
```python
from deephaven import kafka_producer as pk
from deephaven.stream.kafka.producer import KeyValueSpec
from deephaven.time import parse_instant, parse_duration
from deephaven.column import *
from deephaven.time import time_zone, today

import jpy

dc = []
bc = []
cc = []
sc = []
ic = []
lc = []
fc = []
doc = []
stc = []


start_time = parse_instant(today(time_zone("ET")) + "T00:00:00.000000000 ET")
for i in range(0, 86399):
    dc.append(start_time.plusSeconds(i))
    bc.append(i % 256)
    cc.append(33 + (i % (127-33)))
    sc.append((i +100) % 65536)
    ic.append(i + 200)
    lc.append(i + 300)
    fc.append(i + (i / 100))
    doc.append(i + (i / 1000))
    stc.append("GLORIOUS DAY IN THE CORPS #"+str(i))


source = new_table([
    datetime_col("Timestamp", dc),
    byte_col("ByteCol", bc),
    char_col("CharCol", cc),
    short_col("ShortCol", sc),
    int_col("IntCol", ic),
    long_col("LongCol", lc),
    float_col("FloatCol", fc),
    double_col("DoubleCol", doc),
    string_col("StringCol", stc)
    ]).update_view("Date=today()")

SystemClock = jpy.get_type("io.deephaven.base.clock.SystemClock")
clock = SystemClock.of()
UnsortedClockFilter = jpy.get_type("io.deephaven.engine.table.impl.select.UnsortedClockFilter")
ticking_source = source.where(filters=[UnsortedClockFilter("Timestamp", clock, True)])

write_topic = pk.produce(
    ticking_source,
    {"bootstrap.servers": "http://abaranec-vermilion-jdk17.int.illumon.com:9092"},
    "dh14982",
    pk.json_spec(["LongCol"]),
    pk.json_spec(
        [
            "Timestamp",
            "ByteCol",
            "CharCol",
            "ShortCol",
            "IntCol",
            "LongCol",
            "FloatCol",
            "DoubleCol",
            "StringCol"
        ]
    ),
    False,
)
```

Here is the consumer query
```groovy
import io.deephaven.kafka.KafkaTools
import java.util.stream.Collectors
import io.deephaven.engine.table.ColumnDefinition

source = newTable(
    instantCol("Timestamp"),
    byteCol("ByteCol"),
    charCol("CharCol"),
    shortCol("ShortCol"),
    intCol("IntCol"),
    longCol("LongCol"),
    floatCol("FloatCol"),
    doubleCol("DoubleCol"),
    stringCol("StringCol"),
    stringCol("Date"))

cols = source.getDefinition().getColumnsArray()

props = new Properties()
props.put('bootstrap.servers', 'http://abaranec-vermilion-jdk17.int.illumon.com:9092')
props.put('schema.registry.url', 'http://abaranec-vermilion-jdk17.int.illumon.com:8081')
props.put("fetch.min.bytes", "65000")
props.put("fetch.max.wait.ms", "200")
props.put("enable.auto.commit", "false")
props.put("group.id", "dhdis")

import io.deephaven.kafka.KafkaTools

robert = KafkaTools.consumeToTable(props,
    "dh14982",
    {x -> true},
    {x -> 0l},
    io.deephaven.kafka.KafkaTools.Consume.IGNORE,
    io.deephaven.kafka.KafkaTools.Consume.jsonSpec(cols),
    KafkaTools.TableType.append())
```

**Expected results**

The ByteCol and CharCol columns have the same values as in the source table from the producer query

**Actual results**

Expected (from producer table):
Screenshot 2023-08-28 at 10 21 31 AM

Actual:
Screenshot 2023-08-28 at 10 23 01 AM

The ByteCol and CharCol columns have wrong values.

**Versions**
 - Deephaven: 0.27.1
 - OS: Centos 7
 - Browser: Chrome
 - Docker: Not relevant

devinrsmith:
byte consumption was incorrect; it was being deserialized as a text node with the first byte of that text being the result. Instead, it should be treated as a numeric node, and cast to byte. This matches how other numeric types are handled.

char production was incorrect; it was being serialized as an int node. Instead, it should be treated as a text node which matches the expectations of the existing deserialization logic. Additionally, we can extend char deserialization to work in the case where the node actually _is_ a number.

Fixes #4384

### DIFF
--- a/extensions/kafka/src/main/java/io/deephaven/kafka/ingest/JsonNodeUtil.java
+++ b/extensions/kafka/src/main/java/io/deephaven/kafka/ingest/JsonNodeUtil.java
@@ -12,6 +12,7 @@ import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import io.deephaven.UncheckedDeephavenException;
 import io.deephaven.time.DateTimeUtils;
 import io.deephaven.util.QueryConstants;
+import io.deephaven.util.type.TypeUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -75,8 +76,7 @@ public class JsonNodeUtil {
      */
     public static int getInt(@NotNull final JsonNode node, @NotNull final String key,
             final boolean allowMissingKeys, final boolean allowNullValues) {
-        final JsonNode tmpNode = checkAllowMissingOrNull(node, key, allowMissingKeys, allowNullValues);
-        return getInt(tmpNode);
+        return getInt(checkAllowMissingOrNull(node, key, allowMissingKeys, allowNullValues));
     }
 
     /**
@@ -88,8 +88,7 @@ public class JsonNodeUtil {
      */
     public static int getInt(@NotNull final JsonNode node, @NotNull JsonPointer ptr,
             final boolean allowMissingKeys, final boolean allowNullValues) {
-        final JsonNode tmpNode = checkAllowMissingOrNull(node, ptr, allowMissingKeys, allowNullValues);
-        return getInt(tmpNode);
+        return getInt(checkAllowMissingOrNull(node, ptr, allowMissingKeys, allowNullValues));
     }
 
     /**
@@ -112,8 +111,7 @@ public class JsonNodeUtil {
     @Nullable
     public static Integer getBoxedInt(@NotNull final JsonNode node, @NotNull final String key,
             final boolean allowMissingKeys, final boolean allowNullValues) {
-        final JsonNode tmpNode = checkAllowMissingOrNull(node, key, allowMissingKeys, allowNullValues);
-        return getBoxedInt(tmpNode);
+        return TypeUtils.box(getInt(node, key, allowMissingKeys, allowNullValues));
     }
 
     /**
@@ -124,7 +122,7 @@ public class JsonNodeUtil {
      */
     @Nullable
     public static Integer getBoxedInt(final JsonNode node) {
-        return isNullOrMissingField(node) ? null : node.asInt();
+        return TypeUtils.box(getInt(node));
     }
 
     /**
@@ -136,8 +134,7 @@ public class JsonNodeUtil {
      */
     public static short getShort(@NotNull final JsonNode node, @NotNull final String key,
             final boolean allowMissingKeys, final boolean allowNullValues) {
-        final JsonNode tmpNode = checkAllowMissingOrNull(node, key, allowMissingKeys, allowNullValues);
-        return getShort(tmpNode);
+        return getShort(checkAllowMissingOrNull(node, key, allowMissingKeys, allowNullValues));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


devinrsmith:
We don't have any kafka / json parsing tests. Unfortunately, they are very tightly coupled. My hope in the near future is to extract the json parsing logic from kafka so we can have appropriately scoped unit tests.

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/JsonNodeUtil.java
```diff
@@ -476,7 +468,15 @@ public static char getChar(final JsonNode node) {
         if (isNullOrMissingField(node)) {
             return QueryConstants.NULL_CHAR;
         }
+        if (node.isNumber()) {
+            // We don't expect this to be a common case; but if the node happens to be a number, we'll assume it's meant
```
rcaudy, line 472: This is a little on the weird side.

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/JsonNodeUtil.java
```diff
@@ -498,11 +498,19 @@ public static Character getBoxedChar(@NotNull final JsonNode node, @NotNull fina
     }
 
     @Nullable
-    public static Character getBoxedChar(final JsonNode tmpNode) {
-        if (isNullOrMissingField(tmpNode)) {
+    public static Character getBoxedChar(final JsonNode node) {
```
rcaudy, line 501: Why don't our boxed cases delegate to the unboxed and then box the result?

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/JsonNodeUtil.java
devinrsmith: So, consider:

```json
{
  "my_field": 97
}
```

Let's say we say we want to consume this as a char. This PR chooses to treat it as `(char)97` (`a`). The existing code treats it as `"97".charAt(0)` (`9`). We could consider being stricter and throwing an exception, but none of the other code paths are strict. I'm happy to leave the existing code in place if we prefer that behavior.

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/JsonNodeUtil.java
devinrsmith: Ostensibly to be more "efficient" and save a NULL_CHAR comparison. I can update these code paths to use `TypeUtils.box`?

devinrsmith: I don't see any callers of the boxed methods, but they are part of the public API...

rcaudy (approved):

PR #4417, opened 2023-08-31
Support on close callbacks for C++ client

jcferretti:


jcferretti:
The new test passes:
```
cfs@caicai 20:58:13 ~/dh/oss1/deephaven-core/cpp-client/deephaven/build/tests
$ ./tests "On Close Callbacks can be added and removed and are executed"
Filters: On Close Callbacks can be added and removed and are executed
Connecting to localhost:10000
===============================================================================
All tests passed (3 assertions in 1 test case)
```

cpp-client/deephaven/dhclient/include/public/deephaven/client/utility/misc_types.h
```diff
@@ -20,4 +22,28 @@ using TimePointSpecifier = std::variant<std::chrono::system_clock::time_point, i
  * string.
  */
 using DurationSpecifier = std::variant<std::chrono::nanoseconds, int64_t, std::string>;
+
+/**
+ * Used to identify OnClose callbacks, eg, allowing their removal after addition.
+ */
+struct OnCloseCbId {
+  std::uint32_t id;
```
kosak, line 30:
```
  friend bool operator<(OnCloseCbId lhs, OnCloseCbId rhs) {
    return lhs.id < rhs.id;
  }
```

cpp-client/deephaven/dhclient/include/public/deephaven/client/utility/misc_types.h
```diff
@@ -20,4 +22,28 @@ using TimePointSpecifier = std::variant<std::chrono::system_clock::time_point, i
...
+// specialized std::less for deephaven::client::utility::OnCloseCbId.
+namespace std {
+using OnCloseCbId = deephaven::client::utility::OnCloseCbId;
+template<> struct less<OnCloseCbId>
+{
+  bool operator() (const OnCloseCbId &lhs, const OnCloseCbId &rhs) const {
+    return lhs.id < rhs.id;
+  }
+};
+} // namespace std
```
kosak, line 49: FYI, except in one specialized circumstance involving `std::hash`, it's very often an anti-pattern to open up `namespace std` and add stuff to it.

There's no need to specialize `std::less` just to get comparisons working. Instead, you can delete this `std::less` specialization and add an `operator<` to your struct as indicated above.

This is more general anyway, because it means that `id1 < id2` will also work as expected.

cpp-client/deephaven/dhclient/src/impl/client_impl.cc
```diff
@@ -25,25 +24,55 @@ std::shared_ptr<ClientImpl> ClientImpl::Create(
...
+  std::unique_lock lock(on_close_mux_);
+  manager_impl_->Shutdown();
+  if (on_close_cb_ctx_ == nullptr) {
+    return;
+  }
+  for (const auto &entry : on_close_cb_ctx_->map_) {
+    entry.second();
+  }
+  on_close_cb_ctx_.reset();
+}
```
kosak, line 76: Holding a lock while invoking a user callback is kind of an antipattern. I prefer a style where you make the collection "local" and then you work through it. (I also don't see why we're calling `manager_impl_->Shutdown` under lock, as we have no protection against calling it twice anyway)

I'd recommend something like
```
void ClientImpl::Shutdown() {
  manager_impl_->Shutdown();

  on_close_mux_.lock();
  auto local_ctx = std::move(on_close_cb_ctx_);
  on_close_mux_.unlock();

  if (local_ctx == nullptr) {
    return;
  }
  for (const auto &entry : local_ctx->map_) {
    entry.second();
  }
}
```

Or if we want to protect against multiple calls to `Shutdown`, add a boolean flag, tested and set under lock.

cpp-client/deephaven/dhclient/include/private/deephaven/client/impl/client_impl.h
```diff
@@ -27,14 +29,24 @@ class ClientImpl {
...
 
 private:
-  std::shared_ptr<TableHandleManagerImpl> managerImpl_;
+  std::shared_ptr<TableHandleManagerImpl> manager_impl_;
+  mutable std::mutex on_close_mux_;
+  struct OnCloseCbContext {
+    std::uint32_t next_id_;
+    std::map<OnCloseCbId, OnCloseCb> map_;
+  };
+  std::unique_ptr<OnCloseCbContext> on_close_cb_ctx_;
```
kosak, line 50: Why do we have a unique_ptr here rather than
```
OnCloseCbContext on_close_cb_ctx;
```

Or just promote the two fields into the enclosing class and get rid of the `struct OnCloseCbContext` altogether.

I don't see any advantage to the unique_ptr, and the disadvantage is extra code to do null tests. The fields (a uint32_t and a map) are pretty small so I don't think we should be worried about making ClientImpl slightly larger.

kosak (approved):

kosak (approved):

Issue #976, opened 2021-08-05
Parquet: Support non-UTC timestamps


rcaudy:
Running (in groovy):
```
t = io.deephaven.db.tables.utils.ParquetTools.readTable("/data/parquetFiles/eth_v2_p2_cBROTLI.parquet")
```
Would result in:
```
io.deephaven.UncheckedDeephavenException: Unable to read column [tpep_pickup_datetime]: TimestampLogicalType, isAdjustedToUTC=false, unit=MICROS not supported
```
If we didn't have a fortunate schema conversion error that effectively treats the timestamps as UTC. We would like to reverse the error and properly support zoned timestamps. Presumably we could specify a timezone in read instructions.

We *do* have limited support in: `io.deephaven.db.v2.locations.parquet.topage.ToDBDateTimePageFromInt96`
Likely the reference time zone used there should be moved out to `ParquetInstructions`. Then we just need to integrate those instructions into the Int96 support and into the more standard datetime fields.

rcaudy:
So, we have a fairly buggy bit of schema reader code in use right now. One side effect is that we totally ignore that a timestamp might not be UTC, except for the unfortunate hack for int96 mentioned in the description.

malhotrashivam:
**Parquet handling of timestamps:**

TIMESTAMP type has two extra fields: isAdjustedToUTC (= true/false), unit (= MILLIS/MICROS/NANOS).

- TIMESTAMP with `isAdjustedToUTC=true` is defined as the time elapsed since the Unix epoch. So we can translate such a instant to UTC time and then adjust it to local timezone. For example, TIMESTAMP(isAdjustedToUTC=true, unit=MILLIS) value of 172800000 corresponds to 1970-01-03 00:00:00 UTC and the corresponding local time will change according to timezone.
- TIMESTAMP with `isAdjustedToUTC=false` represents time fields in a local timezone, regardless of what local timezone is. For example, TIMESTAMP(isAdjustedToUTC=false, unit=MILLIS) value of 172800000 corresponds to 1970-01-03 00:00:00 in any timezone.

More details in [official docs here.](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md)

devinrsmith:
[...]

PR #4421, opened 2023-08-31
Raise exception when reading non-UTC adjusted timestamps in parquet

### LINKED ISSUE #3588: Unable to read parquet TimestampLogicalTypeAnnotation that is not adjusted to UTC
In some situations, we are unable to read parquet files that have timezones (even if it's just reading into DateTime after applying offset).

The general structure for generating these parquet files was a python virtual env (`pip install pandas pyarrow fastparquet`) with

```python
import pandas as pd
import numpy as np

df = pd.DataFrame(
    {
        "a": list("abc"),
        "b": list(range(1, 4)),
        "c": np.arange(3, 6).astype("u1"),
        "d": np.arange(4.0, 7.0, dtype="float64"),
        "e": [True, False, True],
        "f": pd.date_range("20130101", periods=3),
        "g": pd.date_range("20130101", periods=3, tz="US/Eastern"),
        "h": pd.Categorical(list("abc")),
        "i": pd.Categorical(list("abc"), ordered=True),
    }
)

df.to_parquet("pyarrow.parquet", engine='pyarrow', compression=None)
df.to_parquet("fastparquet.parquet", engine='fastparquet', compression=None)
```

Deephaven is able to read `pyarrow.parquet`, but not `fastparquet.parquet`:

```
io.deephaven.UncheckedDeephavenException: Unable to read column [f]: TimestampLogicalType, isAdjustedToUTC=false, unit=NANOS not supported
	at io.deephaven.parquet.table.ParquetSchemaReader.lambda$readParquetSchema$2(ParquetSchemaReader.java:275)
	at java.base/java.util.Optional.orElseThrow(Optional.java:408)
	at io.deephaven.parquet.table.ParquetSchemaReader.readParquetSchema(ParquetSchemaReader.java:269)
	at io.deephaven.parquet.table.ParquetTools.convertSchema(ParquetTools.java:647)
	at io.deephaven.parquet.table.ParquetTools.readTableInternal(ParquetTools.java:384)
	at io.deephaven.parquet.table.ParquetTools.readTable(ParquetTools.java:77)
```

Pandas is able to successfully read both.

Here is some of the schema information: 

```
D SELECT * FROM PARQUET_SCHEMA('pyarrow.parquet');
┌─────────────────┬─────────┬────────────┬─────────────┬─────────────────┬──────────────┬──────────────────┬───────┬───────────┬──────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────┐
│    file_name    │  name   │    type    │ type_length │ repetition_type │ num_children │  converted_type  │ scale │ precision │ field_id │                                            logical_type                                             │
│     varchar     │ varchar │  varchar   │   varchar   │     varchar     │    int64     │     varchar      │ int64 │   int64   │  int64   │                                               varchar                                               │
├─────────────────┼─────────┼────────────┼─────────────┼─────────────────┼──────────────┼──────────────────┼───────┼───────────┼──────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ pyarrow.parquet │ schema  │ BOOLEAN    │ 0           │ REQUIRED        │            9 │ UTF8             │     0 │         0 │        0 │                                                                                                     │
│ pyarrow.parquet │ a       │ BYTE_ARRAY │ 0           │ OPTIONAL        │            0 │ UTF8             │     0 │         0 │        0 │ StringType()                                                                                        │
│ pyarrow.parquet │ b       │ INT64      │ 0           │ OPTIONAL        │            0 │ UTF8             │     0 │         0 │        0 │                                                                                                     │
│ pyarrow.parquet │ c       │ INT32      │ 0           │ OPTIONAL        │            0 │ UINT_8           │     0 │         0 │        0 │ IntType(bitWidth, isSigned=0)                                                                      │
│ pyarrow.parquet │ d       │ DOUBLE     │ 0           │ OPTIONAL        │            0 │ UTF8             │     0 │         0 │        0 │                                                                                                     │
│ pyarrow.parquet │ e       │ BOOLEAN    │ 0           │ OPTIONAL        │            0 │ UTF8             │     0 │         0 │        0 │                                                                                                     │
│ pyarrow.parquet │ f       │ INT64      │ 0           │ OPTIONAL        │            0 │ TIMESTAMP_MICROS │     0 │         0 │        0 │ TimestampType(isAdjustedToUTC=0, unit=TimeUnit(MILLIS=<null>, MICROS=MicroSeconds(), NANOS=<null>)) │
│ pyarrow.parquet │ g       │ INT64      │ 0           │ OPTIONAL        │            0 │ TIMESTAMP_MICROS │     0 │         0 │        0 │ TimestampType(isAdjustedToUTC=1, unit=TimeUnit(MILLIS=<null>, MICROS=MicroSeconds(), NANOS=<null>)) │
│ pyarrow.parquet │ h       │ BYTE_ARRAY │ 0           │ OPTIONAL        │            0 │ UTF8             │     0 │         0 │        0 │ StringType()                                                                                        │
│ pyarrow.parquet │ i       │ BYTE_ARRAY │ 0           │ OPTIONAL        │            0 │ UTF8             │     0 │         0 │        0 │ StringType()                                                                                        │
├─────────────────┴─────────┴────────────┴─────────────┴─────────────────┴──────────────┴──────────────────┴───────┴───────────┴──────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 10 rows                                                                                                                                                                                                                            11 columns │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
D SELECT * FROM PARQUET_SCHEMA('fastparquet.parquet');
┌─────────────────────┬─────────┬────────────┬─────────────┬─────────────────┬──────────────┬────────────────┬───────┬───────────┬──────────┬────────────────────────────────────────────────────────────────────────────────────────────────────┐
│      file_name      │  name   │    type    │ type_length │ repetition_type │ num_children │ converted_type │ scale │ precision │ field_id │                                            logical_type                                            │
│       varchar       │ varchar │  varchar   │   varchar   │     varchar     │    int64     │    varchar     │ int64 │   int64   │  int64   │                                              varchar                                               │
├─────────────────────┼─────────┼────────────┼─────────────┼─────────────────┼──────────────┼────────────────┼───────┼───────────┼──────────┼────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ fastparquet.parquet │ schema  │ BOOLEAN    │ 0           │ REQUIRED        │            9 │ UTF8           │     0 │         0 │        0 │                                                                                                    │
│ fastparquet.parquet │ a       │ BYTE_ARRAY │ 0           │ OPTIONAL        │            0 │ UTF8           │     0 │         0 │        0 │                                                                                                    │
│ fastparquet.parquet │ b       │ INT64      │ 64          │ OPTIONAL        │            0 │ UTF8           │     0 │         0 │        0 │                                                                                                    │
│ fastparquet.parquet │ c       │ INT32      │ 8           │ OPTIONAL        │            0 │ UINT_8         │     0 │         0 │        0 │                                                                                                    │
│ fastparquet.parquet │ d       │ DOUBLE     │ 64          │ OPTIONAL        │            0 │ UTF8           │     0 │         0 │        0 │                                                                                                    │
│ fastparquet.parquet │ e       │ BOOLEAN    │ 1           │ OPTIONAL        │            0 │ UTF8           │     0 │         0 │        0 │                                                                                                    │
│ fastparquet.parquet │ f       │ INT64      │ 0           │ OPTIONAL        │            0 │ UTF8           │     0 │         0 │        0 │ TimestampType(isAdjustedToUTC=0, unit=TimeUnit(MILLIS=<null>, MICROS=<null>, NANOS=NanoSeconds())) │
│ fastparquet.parquet │ g       │ INT64      │ 0           │ OPTIONAL        │            0 │ UTF8           │     0 │         0 │        0 │ TimestampType(isAdjustedToUTC=1, unit=TimeUnit(MILLIS=<null>, MICROS=<null>, NANOS=NanoSeconds())) │
│ fastparquet.parquet │ h       │ BYTE_ARRAY │ 0           │ OPTIONAL        │            0 │ UTF8           │     0 │         0 │        0 │                                                                                                    │
│ fastparquet.parquet │ i       │ BYTE_ARRAY │ 0           │ OPTIONAL        │            0 │ UTF8           │     0 │         0 │        0 │                                                                                                    │
├─────────────────────┴─────────┴────────────┴─────────────┴─────────────────┴──────────────┴────────────────┴───────┴───────────┴──────────┴────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 10 rows                                                                                                                                                                                                                             11 columns │
└────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

malhotrashivam:
Closes #3588 
Related to #976 

Also added all parquet files in the repo to Git LFS.


chipkent (approved):

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/TestParquetTools.java
chipkent, line 1: confirming that all of these binary files are in LFS

chipkent (approved):

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileReader.java
```diff
@@ -236,8 +236,16 @@ private static void buildChildren(Types.GroupBuilder builder, Iterator<SchemaEle
             }
 
             if (schemaElement.isSetLogicalType()) {
-                ((Types.Builder) childBuilder)
-                        .as(getLogicalTypeAnnotation(schemaElement.logicalType));
+                LogicalType logicalType = schemaElement.logicalType;
+                if (logicalType.isSetTIMESTAMP()) {
+                    TimestampType timestamp = logicalType.getTIMESTAMP();
+                    if (!timestamp.isAdjustedToUTC) {
```
rcaudy, line 242: Is it possible to not be adjusted, but to still be in UTC?

malhotrashivam: **Short answer:** A timestamp that is not adjusted is assumed to be in local timezone, irrespective of what the timezone is. So yes, if local timezone is UTC, then this would be in UTC.

**Long answer:** Let me try to summarize what the [official doc](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md) says:
-  A TIMESTAMP with `isAdjustedToUTC=true` is defined as the time elapsed since the Unix epoch. So we can translate such a instant to UTC time and then  adjust it to local timezone. For example, TIMESTAMP(isAdjustedToUTC=true, unit=MILLIS) value of 172800000 corresponds to 1970-01-03 00:00:00 UTC and will change according to timezone.

- A TIMESTAMP with `isAdjustedToUTC=false` represents time fields in a local timezone, regardless of what local timezone is. For example, TIMESTAMP(isAdjustedToUTC=false, unit=MILLIS) value of 172800000 corresponds to 1970-01-03 00:00:00 in any timezone.

Coming back to our issue, when `isAdjustedToUTC=false`, our code crashes with a random exception in some cases and assumes it to be in UTC in other cases. So I added a clear exception for both cases. There is another issue #976 for properly handling non-UTC adjusted timestamps.

rcaudy: The point of my question was:
If we see `isAdjustedToUTC=false` and the timezone is specified as `UTC`, can we just treat that as an `Instant`? Is there a timezone in the type metadata?

malhotrashivam: My bad. So no, there is no timezone in the TIMESTAMP type, we just have two fields : isAdjustedToUTC (= true/false)
unit (= MILLIS/MICROS/NANOS). So that information combined with the local timezone gives us how to interpret the timestamp value.

rcaudy: Oh. That's horrifying. "This data means something different if you read it in Hong Kong than if you read it in New York."

rcaudy (dismissed):

devinrsmith (commented):
Workflow updates look fine to me. I'm happy to take an `lfs: true` approach for all workflow files as you have done here - the advantage is it's consistent across all workflows. The other approach is to just use `lfs: true` for the tests - the advantage is that non-test workflows won't need to download the LFS files. I'm indifferent, at least until a point in time where we have a significant LFS download cost. I'll defer to other reviewers.

extensions/parquet/table/src/test/resources/e0/.gitattributes
```diff
@@ -0,0 +1 @@
+*.parquet filter=lfs diff=lfs merge=lfs -text
```
rcaudy, line 1: Should we be using binary instead of text?

rcaudy:
> Workflow updates look fine to me. I'm happy to take an `lfs: true` approach for all workflow files as you have done here - the advantage is it's consistent across all workflows. The other approach is to just use `lfs: true` for the tests - the advantage is that non-test workflows won't need to download the LFS files. I'm indifferent, at least until a point in time where we have a significant LFS download cost. I'll defer to other reviewers.

Which approach do you think is best, @devinrsmith ?

devinrsmith:
I'm fine proceeding w/ a principal of "good hygiene"

extensions/parquet/table/src/test/resources/e0/.gitattributes
malhotrashivam: This command is automatically generated on running `git lfs track "*.parquet"`

The `-text` part specifies that Git should treat these files as binary files rather than text files. This is important to avoid any end-of-line conversion issues.
So we do want to keep it like this.

Reference: 
https://earthly.dev/blog/a-developer-guide-to-git-lfs/
https://stackoverflow.com/a/70075078

malhotrashivam:
> I'm fine proceeding w/ a principal of "good hygiene"

Cool, I have reverted back the change to just Check CI and Nightly Check CI jobs.

rcaudy (approved):

PR #4422, opened 2023-08-31
Make use of bulk conversion when creating datetime cols

### LINKED ISSUE #4349: Efficient creation of datetime_col via datetime objects
Right now, `datetime_col` expects a list of already-created java Instants (dtypes.Instant).

```python
from deephaven.time import parse_instant
from deephaven.column import datetime_col

my_strings = ["2016-06-29T19:36:29.3453Z"]
my_dt_col = datetime_col("Timestamp", [parse_instant(s) for s in my_strings])
```

This is clunky, as every item in `my_strings` causes a boundary-crossing event.

It would be nice if `datetime_col` could accept `list[datetime]` (https://docs.python.org/3/library/datetime.html#datetime.datetime). Then, the wrapping layer could adapt that to long nanoseconds without any more boundary crossings than something like `long_col` can achieve.

```python
import dateutil.parser

from deephaven.column import datetime_col

my_strings = ["2016-06-29T19:36:29.3453Z"]
my_dt_col = datetime_col("Timestamp", [dateutil.parser.parse(s) for s in my_strings])
```

jmao-denver:
Fixes #4349

### DIFF
--- a/py/server/deephaven/column.py
+++ b/py/server/deephaven/column.py
@@ -215,8 +215,13 @@ def datetime_col(name: str, data: Sequence) -> InputColumn:
         except Exception as e:
             ...
 
-    if isinstance(data, np.ndarray) and data.dtype.kind == 'M':
-        longs = jpy.array('long', data.astype('datetime64[ns]').astype('int64'))
+    if isinstance(data, np.ndarray) and data.dtype.kind in ('M', 'i', 'U'):
+        if data.dtype.kind == 'M':
+            longs = jpy.array('long', data.astype('datetime64[ns]').astype('int64'))
+        elif data.dtype.kind == 'i':
+            longs = jpy.array('long', data.astype('int64'))
+        else:  # data.dtype.kind == 'U'
+            longs = jpy.array('long', [pd.Timestamp(str(dt)).to_numpy().astype('int64') for dt in data])
         data = _JPrimitiveArrayConversionUtility.translateArrayLongToInstant(longs)
 
     if not isinstance(data, dtypes.instant_array.j_type):
--- a/py/server/deephaven/column.py
+++ b/py/server/deephaven/column.py
@@ -9,15 +9,19 @@ from enum import Enum
 from typing import Sequence, Any
 
 import jpy
+import numpy as np
+import pandas as pd
 
 import deephaven.dtypes as dtypes
 from deephaven import DHError, time
 from deephaven.dtypes import DType
+from deephaven.time import to_j_instant
 
 _JColumnHeader = jpy.get_type("io.deephaven.qst.column.header.ColumnHeader")
 _JColumn = jpy.get_type("io.deephaven.qst.column.Column")
 _JColumnDefinition = jpy.get_type("io.deephaven.engine.table.ColumnDefinition")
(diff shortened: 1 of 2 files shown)


devinrsmith (commented):
A think we should have some tests to verify it's doing what we think it should do?

py/server/deephaven/column.py
```diff
@@ -202,15 +206,30 @@ def datetime_col(name: str, data: Sequence) -> InputColumn:
...
+    # it to reduce the number of round trips to the JVM
+    if not isinstance(data, np.ndarray):
+        try:
+            data = np.array([pd.Timestamp(dt).to_numpy() for dt in data], dtype=np.datetime64)
+        except Exception as e:
+            ...
+
+    if isinstance(data, np.ndarray) and data.dtype.kind == 'M':
+        longs = jpy.array('long', data.astype('datetime64[ns]').astype('int64'))
+        data = _JPrimitiveArrayConversionUtility.translateArrayLongToInstant(longs)
```
chipkent, line 220: Are there concerns about this pattern doubling the memory allocations?

py/server/deephaven/column.py
```diff
@@ -202,15 +206,30 @@ def datetime_col(name: str, data: Sequence) -> InputColumn:
...
+
+    if not isinstance(data, dtypes.instant_array.j_type):
+        data = [to_j_instant(d) for d in data]
```
chipkent, line 223: I think the logic here doesn't quite work.  What if the user passed in a numpy array of strings?  That should be valid to convert, but I think it fails.  Either way, it should be tested.

chipkent (changes requested):

py/server/deephaven/column.py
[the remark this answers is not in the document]
jmao-denver: I don't think so. It is more confusing than helpful IMO.

jmao-denver: I think it is not so bad. IMO the performance gain outweighs the additional memory consumption.

py/server/deephaven/column.py
jmao-denver: It should work. I have added a test case.

py/server/deephaven/column.py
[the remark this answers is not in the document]
chipkent: I don't follow this logic.  If `data.dtype.kind == 'i'`, you could have an additional block that looks like:
```
if isinstance(data, np.ndarray) and data.dtype.kind == 'i':
        longs = jpy.array('long', data.astype('int64'))
        data = _JPrimitiveArrayConversionUtility.translateArrayLongToInstant(longs)
```
This would just yield one boundary crossing for an array of ints vs what is happening now with a boundary crossing pervalue.

Similarly, there could be another block like this that coerces to a long array instead of a datetime array ... unless the timestamp coercion is already handling this case.  Either way, we need a unit test for epoch nanos ints to make sure they work. 
```
    if not isinstance(data, np.ndarray):
        try:
            data = np.array([pd.Timestamp(dt).to_numpy() for dt in data], dtype=np.datetime64)
        except Exception as e:
            ...
```

chipkent (approved):

PR #4427, opened 2023-09-01
Wrap FunctionGeneratedTableFactory for Python

rbasralian:
- Wrap `FunctionGeneratedTableFactory` for Python
- Fix a potential double-notify when multiple sources are specified


py/server/deephaven/util/function_generated_table.py
```diff
@@ -0,0 +1,84 @@
...
+from deephaven.execution_context import ExecutionContext
+from deephaven.jcompat import j_lambda
+from deephaven.table import Table
+
+_J_Table = jpy.get_type("io.deephaven.engine.table.Table")
+_J_Supplier = jpy.get_type('java.util.function.Supplier')
+_J_FunctionGeneratedTableFactory = jpy.get_type("io.deephaven.engine.table.impl.util.FunctionGeneratedTableFactory")
+
+
+def create(table_generator: Callable[[], Table],
```
jmao-denver, line 15: Not sure if it should stand on its own. Maybe `table_factory.py` is a more fitting landing place.

py/server/deephaven/util/function_generated_table.py
```diff
@@ -0,0 +1,84 @@
...
+
+    The table definition must not change between invocations of the 'table_generator' function.
+
+    Note that any tables used by the 'table_generator' function *MUST* be specified in the 'source_tables'.
+
+    :param table_generator: The table generator function. This function must return a Table.
+    :param source_tables: Source tables used by the 'table_generator' function. The 'table_generator' is rerun when any of these tables tick.
+    :param refresh_interval_ms: Interval (in milliseconds) at which the 'table_generator' function is rerun.
+    :param exec_context: Optionally, a custom execution context. If 'None', the current execution context is used.
+    :return: A Table that is automatically regenerated when any of the source_tables tick. This table will be regenerated *after* any changes to the source_tables have been processed.
```
jmao-denver, line 33: 1. Change to Google style to be consistent with the rest of the code base.
2. `source_tables` could be renamed to `trigger_tables`?
3. `exec_context` could be made more concise to `exec_ctx`; the current execution context might be absent, what happens then?
4. Would be nice to mention a typical use case

py/server/deephaven/util/function_generated_table.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+
+    if refresh_interval_ms is None and source_tables is None:
```
jmao-denver, line 36: `source_tables` could be empty?

py/server/deephaven/util/function_generated_table.py
```diff
@@ -0,0 +1,84 @@
...
+        raise DHError("Either refresh_interval_ms or source_tables must be defined!")
+
+    if refresh_interval_ms is not None and source_tables is not None:
+        raise DHError("Only one of refresh_interval_ms and source_tables must be defined!")
+
+    # If no execution context is provided, assume we want to use the current one.
+    if exec_context is None:
+        exec_context = execution_context.get_exec_ctx()
```
jmao-denver, line 44: check if return value is not None?

py/server/deephaven/util/function_generated_table.py
chipkent: I haven't done a proper review, but I am very much against `util` packages.  They just become dumping grounds for garbage over time.  Also, this giant module name is not pythonic.  It is way too long.

rcaudy (commented):
Only reviewing FunctionBackedTableFactory.

py/server/deephaven/util/function_generated_table.py
rbasralian: 1. done (I think)
2. It is `sourceTables` in the Java signature. I think `source` is an important distinction because any 'sources tables' used by the function must be specified here (to ensure that the engine has updated them before rerunning the function).
3. I'm not sure — whatever `with exec_context` does. Do you think I should manually throw an error here?
4. added some examples

py/server/deephaven/util/function_generated_table.py
rbasralian: Yup, in that case the function only runs once. I thought about making it an error, but I could imagine a case where you use this dynamically and reasonably wind up with zero source tables and no requirement to refresh on an interval. (Like with a function that iterates over a collection of other lambdas that may or may not depend on DH sources.)

py/server/deephaven/util/function_generated_table.py
rbasralian: Moved it to `table_factory.py`

py/server/deephaven/table_factory.py
```diff
@@ -335,3 +340,88 @@ def ring_table(parent: Table, capacity: int, initialize: bool = True) -> Table:
...
+    'source_tables' ticks or (b) after refresh_interval_ms have elapsed. Either 'refresh_interval_ms' or
+    'source_tables' must be set (but not both).
+
+    Function-generated tables can be used to produce dynamic tables from sources outside Deephaven. For example,
+    function-generated tables can create tables that are produced by arbitrary Python logic (including using Pandas or
+    numpy). They can also be used to retrieve data from external sources (such as files or websites).
+
+    The table definition must not change between invocations of the 'table_generator' function.
+
+    Note that any tables used by the 'table_generator' function *MUST* be specified in the 'source_tables'.
```
chipkent, line 361:
```suggestion
    Note: any tables used by the 'table_generator' function *MUST* be specified in the 'source_tables'.
```
I think this makes it consistent with the other docs.

py/server/deephaven/table_factory.py
```diff
@@ -335,3 +340,88 @@ def ring_table(parent: Table, capacity: int, initialize: bool = True) -> Table:
...
+    Creates an abstract table that is generated by running the table_generator() function. The function will first be
+    run to generate the table when this method is called, then subsequently either (a) whenever one of the
+    'source_tables' ticks or (b) after refresh_interval_ms have elapsed. Either 'refresh_interval_ms' or
+    'source_tables' must be set (but not both).
+
+    Function-generated tables can be used to produce dynamic tables from sources outside Deephaven. For example,
+    function-generated tables can create tables that are produced by arbitrary Python logic (including using Pandas or
+    numpy). They can also be used to retrieve data from external sources (such as files or websites).
+
+    The table definition must not change between invocations of the 'table_generator' function.
```
chipkent, line 359:
```suggestion
    The table definition must not change between invocations of the 'table_generator' function, or an exception will be raised.
```
does this cause an exception?  If so, indicate in the docstring.

py/server/deephaven/table_factory.py
```diff
@@ -335,3 +340,88 @@ def ring_table(parent: Table, capacity: int, initialize: bool = True) -> Table:
...
+
+    Note that any tables used by the 'table_generator' function *MUST* be specified in the 'source_tables'.
+
+    Args:
+        table_generator (Callable[[], Table]): The table generator function. This function must return a Table.
+        source_tables (Union[Table, List[Table]]): Source tables used by the 'table_generator' function. The
+            'table_generator' is rerun when any of these tables tick.
+        refresh_interval_ms (int): Interval (in milliseconds) at which the 'table_generator' function is rerun.
+        exec_ctx (ExecutionContext) -> Table): Optionally, a custom execution context. If 'None', the current
+            execution context is used. If there is no current execution context, a ValueError is raised.
```
chipkent, line 369:
```suggestion
        exec_ctx (ExecutionContext) -> Table): A custom execution context. If 'None', the current
            execution context is used. If there is no current execution context, a ValueError is raised.
```
I think this is more consistent with the other docs.

py/server/deephaven/table_factory.py
```diff
@@ -335,3 +340,88 @@ def ring_table(parent: Table, capacity: int, initialize: bool = True) -> Table:
...
+
+    Returns:
+        A Table that is automatically regenerated when any of the source_tables tick. This table will be regenerated
+        *after* any changes to the source_tables have been processed.
```
chipkent, line 373:
```suggestion
    Returns:
        a new table
        
    Raises:
        DHError
```
This is more consistent with the existing docs.  The first sentence is covered already.  If we need second sentence, it should be moved to the body.

py/server/deephaven/table_factory.py
```diff
@@ -335,3 +340,88 @@ def ring_table(parent: Table, capacity: int, initialize: bool = True) -> Table:
...
+    """
+
+    if refresh_interval_ms is None and source_tables is None:
+        raise DHError("Either refresh_interval_ms or source_tables must be defined!")
+
+    if refresh_interval_ms is not None and source_tables is not None:
+        raise DHError("Only one of refresh_interval_ms and source_tables must be defined!")
```
chipkent, line 380:
```suggestion
        raise DHError("Either refresh_interval_ms or source_tables must be provided!")

    if refresh_interval_ms is not None and source_tables is not None:
        raise DHError("Only one of refresh_interval_ms and source_tables must be provided!")
```
Make consistent with existing messages.

py/server/deephaven/table_factory.py
```diff
@@ -335,3 +340,88 @@ def ring_table(parent: Table, capacity: int, initialize: bool = True) -> Table:
...
+
+    # If no execution context is provided, assume we want to use the current one.
+    if exec_ctx is None:
+        exec_ctx = execution_context.get_exec_ctx()
+        if exec_ctx is None:
+            raise ValueError("No execution context is available and exec_ctx was not set! ")
```
chipkent, line 386:
```suggestion
            raise ValueError("No execution context is available and exec_ctx was not provided! ")
```

chipkent (changes requested):

chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3159
Reference: https://github.com/deephaven/deephaven.io/issues/3158

engine/table/src/main/java/io/deephaven/engine/table/impl/util/FunctionGeneratedTableFactory.java
```diff
@@ -4,14 +4,18 @@
...
 import io.deephaven.engine.table.impl.QueryTable;
-import io.deephaven.engine.table.impl.BaseTable;
-import io.deephaven.engine.table.impl.sources.*;
+import io.deephaven.engine.table.impl.sources.ArrayBackedColumnSource;
 import io.deephaven.engine.updategraph.UpdateGraph;
+import org.jetbrains.annotations.NotNull;
 
-import java.util.LinkedHashMap;
-import java.util.Map;
+import java.util.*;
```
mofojed, line 18: I thought we avoid wildcard imports...

PR #4433, opened 2023-09-02
Fix Client::dtor calling close after addition of copy ctor.

jcferretti:


cpp-client/deephaven/dhclient/src/impl/client_impl.cc
```diff
@@ -58,13 +65,25 @@ bool ClientImpl::RemoveOnCloseCallback(OnCloseCbId cb_id) {
...
+  gpr_log(GPR_DEBUG, "ClientImpl(%p): Shutdown starting...", v_this);
+  temp->Shutdown();
   std::unique_lock lock(on_close_.mux);
   auto map = std::move(on_close_.map);
+  gpr_log(GPR_DEBUG, "ClientImpl(%p): Executing on close callbacks...", v_this);
   lock.unlock();
+  std::size_t cb_num = 0;
   for (const auto &entry : map) {
+    ++cb_num;
+    gpr_log(GPR_DEBUG, "ClientImpl(%p): Executing on close call %lu.", v_this, cb_num);
```
kosak, line 83: I think the format for size_t is %zu

jcferretti:
We discussed with Corey over slack and agreed that this is the path to madness.  We are going to close this PR without merging and will revert the part of the copy operations change to add copy operations to Client.

PR #4443, opened 2023-09-05
Update RELEASE.md to explicitly list the files that have version numbers.

jcferretti:


stanbrub (commented):
There are two places where these files are modified.  One is during a patch release like you have here.  The other is after a minor release where we do the Post Release Bump to ...  Should we make a section for the list and then reference it in both places?  Or is this good enough to just have it here?

devinrsmith (approved):

PR #4446, opened 2023-09-05
Added support for LZ4_RAW compression codec for Parquet

### LINKED ISSUE #3148: LZ4_RAW parquet support
**Description**

It appears that the "LZ4" constant in pandas/pyarrow results in a compression value that can't be read by our Java code.

**Steps to reproduce**
Create a dataframe and write it to a parquet file with LZ4 compression:
```
dh_table = empty_table(10).update("I=i")
dataframe = to_pandas(dh_table)
# this line will fail
dataframe.to_parquet('data_from_pandas.parquet', compression="LZ4")
result_table = read('data_from_pandas.parquet')
# next verify contents...
```

**Expected results**

The parquet file can be decompressed and read

**Actual results**

```
java.lang.RuntimeException: Error in Python interpreter:

Value: failed to read parquet data. : RuntimeError: java.lang.IllegalArgumentException: No enum constant org.apache.parquet.hadoop.metadata.CompressionCodecName.LZ4_RAW
  File "/home/colin/.pyenv/versions/3.7.12/lib/python3.7/site-packages/deephaven/parquet.py", line 88, in read
    return Table(j_table=_JParquetTools.readTable(path))
RuntimeError: java.lang.IllegalArgumentException: No enum constant org.apache.parquet.hadoop.metadata.CompressionCodecName.LZ4_RAW
	at java.base/java.lang.Enum.valueOf(Enum.java:240)
	at org.apache.parquet.hadoop.metadata.CompressionCodecName.valueOf(CompressionCodecName.java:26)
	at io.deephaven.parquet.base.tempfix.ParquetMetadataConverter.fromFormatCodec(ParquetMetadataConverter.java:535)
	at io.deephaven.parquet.base.tempfix.ParquetMetadataConverter.fromParquetMetadata(ParquetMetadataConverter.java:1195)
	at io.deephaven.parquet.table.location.ParquetTableLocationKey.getMetadata(ParquetTableLocationKey.java:99)
	at io.deephaven.parquet.table.ParquetTools.readTableInternal(ParquetTools.java:386)
	at io.deephaven.parquet.table.ParquetTools.readTable(ParquetTools.java:77)
	at org.jpy.PyLib.executeCode(Native Method)
	at org.jpy.PyObject.executeCode(PyObject.java:138)
	at io.deephaven.engine.util.PythonEvaluatorJpy.evalScript(PythonEvaluatorJpy.java:73)
	at io.deephaven.integrations.python.PythonDeephavenSession.lambda$evaluate$1(PythonDeephavenSession.java:183)
	at io.deephaven.util.locks.FunctionalLock.doLockedInterruptibly(FunctionalLock.java:50)
	at io.deephaven.integrations.python.PythonDeephavenSession.evaluate(PythonDeephavenSession.java:182)
	at io.deephaven.engine.util.AbstractScriptSession.lambda$evaluateScript$1(AbstractScriptSession.java:145)
	at io.deephaven.engine.context.ExecutionContext.lambda$apply$0(ExecutionContext.java:123)
	at io.deephaven.engine.context.ExecutionContext.apply(ExecutionContext.java:134)
	at io.deephaven.engine.context.ExecutionContext.apply(ExecutionContext.java:122)
	at io.deephaven.engine.util.AbstractScriptSession.evaluateScript(AbstractScriptSession.java:145)
	at io.deephaven.engine.util.DelegatingScriptSession.evaluateScript(DelegatingScriptSession.java:87)
	at io.deephaven.engine.util.ScriptSession.evaluateScript(ScriptSession.java:113)
	at io.deephaven.server.console.ConsoleServiceGrpcImpl.lambda$executeCommand$8(ConsoleServiceGrpcImpl.java:169)
	at io.deephaven.server.session.SessionState$ExportBuilder.lambda$submit$2(SessionState.java:1343)
	at io.deephaven.server.session.SessionState$ExportObject.doExport(SessionState.java:880)
	at io.deephaven.server.runner.DeephavenApiServerModule$ThreadFactory.lambda$newThread$0(DeephavenApiServerModule.java:166)


Line: 90
Namespace: read
File: /home/colin/.pyenv/versions/3.7.12/lib/python3.7/site-packages/deephaven/parquet.py
  File "<string>", line 4, in <module>
  File "/home/colin/.pyenv/versions/3.7.12/lib/python3.7/site-packages/deephaven/parquet.py", line 90, in read

        at org.jpy.PyLib.executeCode(PyLib.java:-2)
        at org.jpy.PyObject.executeCode(PyObject.java:138)
        at io.deephaven.engine.util.PythonEvaluatorJpy.evalScript(PythonEvaluatorJpy.java:73)
```

Likely we could resolve this by treating the value "LZ4_RAW" as if it was "LZ4" in our codec factory.

**Versions**
- Deephaven: 0.18+
- OS: Linux
- Docker: N/A

malhotrashivam:
Closes #3148 

Also, added a new adapter for LZ4 which tries to decompress a file with LZ4_RAW decompressors in case it fails with LZ4 decompressors. The retry mechanism is particularly useful for decompressing parquet files that are compressed with LZ4_RAW but tagged as LZ4 in the metadata. 

These are some interesting links to read about the [LZ4 format](https://docs.fileformat.com/compression/lz4/#:~:text=A%20file%20with%20.,be%20decompressed%20using%20the%20same) and [LZ4_RAW or LZ4 Block format](https://android.googlesource.com/platform/external/lz4/+/HEAD/doc/lz4_Block_format.md)

**Documentation Update:**
- For writeTable [Java Page](https://deephaven.io/core/groovy/docs/reference/data-import-export/Parquet/writeTable/), do the following additions for the `ParquetInstructions` argument:
	- `ParquetTools.UNCOMPRESSED` - The output will not be compressed.
	- `ParquetTools.SNAPPY` - Aims for high speed, and a reasonable amount of compression, based on Snappy compression format by [Google](https://github.com/google/snappy/blob/main/format_description.txt).
	- `ParquetTools.LZ4_RAW` - A codec based on the [LZ4 block format ](https://github.com/lz4/lz4/blob/dev/doc/lz4_Block_format.md), recommended to use instead of `ParquetTools.LZ4`. 
	- For `ParquetTools.LZ4`, mention that this is deprecated and users should use `ParquetTools.LZ4_RAW`.
	- Also, mention that if `ParquetInstructions` not specified will default to `ParquetTools.SNAPPY` compression format.

Similar changes to be done for the [python page](https://deephaven.io/core/docs/reference/data-import-export/Parquet/writeTable/).

Overall, there should be 7 compression formats now `UNCOMPRESSED`, `LZ4`, `LZ4_RAW`, `LZO`, `GZIP`, `ZSTD`, `SNAPPY` with default being `SNAPPY` and we need to update the documentation everywhere to reflect this. For Java pages, they are referred as `ParquetTools.<format>` (example `ParquetTools.LZ4`) and for python, we refer them as `"<format>"` (example `"LZ4"`).
- Make similar changes to the instructions provided at the cheat sheet for [Java ](https://deephaven.io/core/groovy/docs/reference/cheat-sheets/parquet/#docusaurus_skipToContent_fallback) and [Python](https://deephaven.io/core/docs/reference/cheat-sheets/parquet/#docusaurus_skipToContent_fallback)
- Make similar changes to "How to write and read single Parquet files" page in [Java ](https://deephaven.io/core/docs/how-to-guides/data-import-export/parquet-single/#docusaurus_skipToContent_fallback) and [Python](https://deephaven.io/core/groovy/docs/how-to-guides/parquet-single/)
- Make similar changes to "How to write and read multiple Parquet files" page in [Java](https://deephaven.io/core/docs/how-to-guides/data-import-export/parquet-directory/#docusaurus_skipToContent_fallback) and [Python](https://deephaven.io/core/groovy/docs/how-to-guides/parquet-directory/#docusaurus_skipToContent_fallback)


extensions/parquet/table/src/brotliTest/java/io/deephaven/parquet/table/BrotliParquetTableReadWriteTest.java
```diff
@@ -94,24 +94,17 @@ public boolean equals(Object obj) {
         }
     }
 
-    private void compressionCodecTestHelper(final String codec) {
-        final String currentCodec = ParquetInstructions.getDefaultCompressionCodecName();
```
niloc132, line 98: The risk with removing this is that we might end up one day with a brotli codec that can be used cross-platform, and this might test method might run in the same jvm as other tests, contaminating them by changing state like this. Unless there is a good reason to remove this?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -842,11 +842,17 @@ public static Pair<List<ColumnDefinition<?>>, ParquetInstructions> convertSchema
     }
 
     public static final ParquetInstructions LZ4 = ParquetInstructions.builder().setCompressionCodecName("LZ4").build();
+    public static final ParquetInstructions LZ4_RAW =
+            ParquetInstructions.builder().setCompressionCodecName("LZ4_RAW").build();
```
niloc132, line 846: should we also test for `LZ4RAW`, since that is apparently the actual name for it?

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -317,52 +317,51 @@ public void groupingByBigInt() {
...
-        try {
-            ParquetInstructions.setDefaultCompressionCodecName(codec);
-            String path = rootFile + File.separator + "Table1.parquet";
-            final Table table1 = getTableFlat(10000, false, true);
-            ParquetTools.writeTable(table1, path);
-            assertTrue(new File(path).length() > 0);
-            final Table table2 = ParquetTools.readTable(path);
-            TstUtils.assertTableEquals(maybeFixBigDecimal(table1), table2);
-        } finally {
-            ParquetInstructions.setDefaultCompressionCodecName(currentCodec);
```
niloc132, line 331: (edited before sending)

Is it possible that we should junk ParquetInstructions.setDefaultCompressionCodecName entirely (or at least deprecate for now), and always insist that the writeTable() method include a codec?

extensions/parquet/table/src/brotliTest/java/io/deephaven/parquet/table/BrotliParquetTableReadWriteTest.java
malhotrashivam: I modified the test to keep it similar to how we expect users to specify compression codec. In this original implementation, we modify the default compression codec and then write the table with the default. This seems weird because users won't have the option to change the default.

Instead I tested it like we recommend here `https://deephaven.io/core/groovy/docs/how-to-guides/parquet-single/` by just passing the codec enum with the API call.

This should not contaminate any state in case this test is run on the same JVM as other tests.

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
malhotrashivam: Yup, I can do that.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
malhotrashivam: I think the actual name is `LZ4_RAW` only like mentioned [here](https://arrow.apache.org/docs/python/generated/pyarrow.Codec.html) and [here](https://github.com/apache/parquet-format/blob/master/Compression.md). Do you still think we should add tests for `LZ4RAW`?

chipkent (dismissed):
Python LGTM.  Did not review java.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
niloc132: If I'm reading this right, the only reason that LZ4_RAW works at all is the `if` block in `DeephavenCompressorAdapterFactory.getByName` - the underlying hadoop `CompressionCodecFactory.getCodecByName` expects LZ4RAW (well, .toLowerCase())... So I was guessing that something downstream might actually want to pass the string "lz4raw" at some point, so we should validate that it works? 

I've forgotten a lot of the specifics here though, so I'm guessing. I'm not suggesting we have an extra constant here, just something to test that the string Lz4raw still works.

niloc132 (dismissed):

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
malhotrashivam: Makes sense, thanks for explaining. I have added a test in the python code using string `LZ4RAW`.

extensions/parquet/compression/src/main/java/io/deephaven/parquet/compress/LZ4WithLZ4RawBackupCompressorAdapter.java
```diff
@@ -0,0 +1,60 @@
...
+class LZ4WithLZ4RawBackupCompressorAdapter extends DeephavenCompressorAdapterFactory.CodecWrappingCompressorAdapter {
+    private CompressorAdapter lz4RawAdapter = null; // Lazily initialized
+
+    LZ4WithLZ4RawBackupCompressorAdapter(CompressionCodec compressionCodec,
+            CompressionCodecName compressionCodecName) {
+        super(compressionCodec, compressionCodecName);
+    }
+
+    @Override
+    public BytesInput decompress(final InputStream inputStream, final int compressedSize,
```
rcaudy, line 28: Should this remember that we failed and fell back once before? And then always fall back without trying to fail in the future?

extensions/parquet/compression/src/main/java/io/deephaven/parquet/compress/LZ4WithLZ4RawBackupCompressorAdapter.java
```diff
@@ -0,0 +1,79 @@
...
+            final int uncompressedSize) throws IOException {
+        if (mode == DecompressionMode.LZ4) {
+            return super.decompress(inputStream, compressedSize, uncompressedSize);
+        }
+        if (mode == DecompressionMode.LZ4_RAW) {
+            assert lz4RawAdapter != null;
```
malhotrashivam, line 44: I did not want to add a new gradle dependency on `Deephaven_Community_Core.Base.main`, that is why using this `assert` instead of `Assert.neqNull`.
Let me know if I should add that dependency.

extensions/parquet/compression/src/main/java/io/deephaven/parquet/compress/LZ4WithLZ4RawBackupCompressorAdapter.java
```diff
@@ -12,12 +12,22 @@
 import java.io.InputStream;
 
 /**
- * This is the default adapter for LZ4 data. It attempts to decompress with LZ4, and falls back to LZ4_RAW on failure.
- * The fallback mechanism is particularly useful for decompressing parquet files that are compressed with LZ4_RAW but
- * tagged as LZ4 in the metadata.
+ * This is the default adapter for LZ4 files. It attempts to decompress with LZ4 and falls back to LZ4_RAW on failure.
+ * After that, it always uses LZ4_RAW for decompression. This fallback mechanism is particularly useful for
+ * decompressing parquet files that are compressed with LZ4_RAW but tagged as LZ4 in the metadata. This adapter is
+ * internally stateful in some cases and therefore a single instance should not be re-used across files.
```
rcaudy, line 18: Good comment, but I would like a note on the code that builds one of these, too.

rcaudy (approved):

extensions/parquet/compression/src/main/java/io/deephaven/parquet/compress/LZ4WithLZ4RawBackupCompressorAdapter.java
niloc132: Should probably throw IllegalStateException/IllegalArgumentException then - `assert` is turned off by default at runtime, so this will not throw in production.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -841,14 +841,33 @@ public static Pair<List<ColumnDefinition<?>>, ParquetInstructions> convertSchema
...
+            ParquetInstructions.builder().setCompressionCodecName("BROTLI").build();
     public static final ParquetInstructions LEGACY = ParquetInstructions.builder().setIsLegacyParquet(true).build();
 
+    /**
+     * Deprecated: Do not use this method directly, instead pass the above codecs as arguments to
+     * {@link #writeTable(Table, File, ParquetInstructions)} method
+     */
+    @Deprecated
     public static void setDefaultCompressionCodecName(final String compressionCodecName) {
         ParquetInstructions.setDefaultCompressionCodecName(compressionCodecName);
```
niloc132, line 872: Should we deprecate the ParquetInstructions method too?

malhotrashivam: Done

niloc132 (approved):

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3199
Reference: https://github.com/deephaven/deephaven.io/issues/3200

PR #4451, opened 2023-09-06
DeferredViewTable: Only Manage When Refreshing

### LINKED ISSUE #4448: DeferredViewTable Manage's Filters and TableReference but is not Refreshing
To reproduce:

```groovy
import io.deephaven.parquet.table.ParquetTools

// write a partition table with an int partition column
def part = TableTools.emptyTable(4).update("II = ii")
ParquetTools.writeTable(part, "/tmp/pt-test/intCol=first/part.parquet")
ParquetTools.writeTable(part, "/tmp/pt-test/intCol=second/part.parquet")

// load the partition table
partition_table = ParquetTools.readTable("/tmp/pt-test")
```

On the partition_table widget on the UI, create a quick filter on the II column. Then select the `first` partition. It fails to load as the DVT is destroyed before being coalesced. The DVT is an export object, but since this check fails, the gRPC session state framework does not manage it:
```java
if (result instanceof LivenessReferent && DynamicNode.notDynamicOrIsRefreshing(result)) {
  manage((LivenessReferent) result);
}
```

This ultimately throws this exception:
```
2023-09-05T22:08:43.299Z | heduler-Concurrent-4 | ERROR | i.d.s.s.SessionState      | Internal Error '4c485e1a-57e6-41d6-9ce5-a3bcd3ca8974' io.deephaven.engine.liveness.LivenessStateException: QueryTable (1) failed to add ConjunctiveFilter([intCol in ["first"], II in [2]]) (0), because either this manager or referent is no longer live
	at io.deephaven.engine.liveness.LivenessManager.manage(LivenessManager.java:24)
	at io.deephaven.engine.table.impl.QueryTable$FilteredTable.<init>(QueryTable.java:965)
	at io.deephaven.engine.table.impl.QueryTable.lambda$whereInternal$27(QueryTable.java:1264)
	at io.deephaven.engine.table.impl.BaseTable.initializeWithSnapshot(BaseTable.java:1261)
	at io.deephaven.engine.table.impl.QueryTable.lambda$whereInternal$28(QueryTable.java:1202)
	at io.deephaven.engine.table.impl.QueryTable$MemoizedResult.getOrCompute(QueryTable.java:3504)
	at io.deephaven.engine.table.impl.QueryTable.memoizeResult(QueryTable.java:3473)
	at io.deephaven.engine.table.impl.QueryTable.lambda$whereInternal$29(QueryTable.java:1197)
	at io.deephaven.engine.table.impl.perf.QueryPerformanceRecorder.withNugget(QueryPerformanceRecorder.java:519)
	at io.deephaven.engine.table.impl.QueryTable.whereInternal(QueryTable.java:1152)
	at io.deephaven.engine.table.impl.QueryTable.where(QueryTable.java:1140)
	at io.deephaven.engine.table.impl.QueryTable.where(QueryTable.java:111)
	at io.deephaven.engine.table.impl.DeferredViewTable.doCoalesce(DeferredViewTable.java:238)
	at io.deephaven.engine.table.impl.UncoalescedTable.coalesce(UncoalescedTable.java:78)
	at io.deephaven.engine.table.impl.UncoalescedTable.flatten(UncoalescedTable.java:441)
	at io.deephaven.server.table.ops.FlattenTableGrpcImpl.create(FlattenTableGrpcImpl.java:29)
	at io.deephaven.server.table.ops.FlattenTableGrpcImpl.create(FlattenTableGrpcImpl.java:16)
	at io.deephaven.server.table.ops.TableServiceGrpcImpl$BatchExportBuilder.doExport(TableServiceGrpcImpl.java:692)
	at io.deephaven.server.table.ops.TableServiceGrpcImpl.lambda$batch$5(TableServiceGrpcImpl.java:541)
	at io.deephaven.server.session.SessionState$ExportObject.doExport(SessionState.java:981)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at io.deephaven.server.runner.scheduler.SchedulerModule$ThreadFactory.lambda$newThread$0(SchedulerModule.java:78)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

Ryan and I discussed how this should be fixed.
1. TableReference to record whether its input was refreshing
1. DVT to become refreshing in the constructor, rather than externally, based on TableReference and maybe based on whether WhereFilters are refreshing
1. DVT to only manage filters or table reference if it is refreshing
1. Apply same filter management logic to `QueryTable.FilteredTable`

nbauernfeind:
Fixes #4448.

`WhereFilterLivenessArtifactImpl` is not also a `DynamicNode`, which leads to an unusual relationship where a `DeferredViewTable` might manage a `WhereFilter`, but then the DVT is not managed by the `SessionState.ExportObject` infrastructure, causing liveness issues later when the where filter is finally applied to the coalesced table when resolving downstream gRPC table operations.

Instead of unconditionally managing WhereFilters, now we'll only manage them if they are `WhereFilter#isRefreshing`.

Release Notes: Fixed liveness issues related to DeferredViewTable and non-refreshing where filters

Nightlies: https://github.com/nbauernfeind/deephaven-core/actions/runs/6092770745

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/DeferredViewTable.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/DeferredViewTable.java
@@ -300,8 +300,11 @@ public class DeferredViewTable extends RedefinableTable<DeferredViewTable> {
      */
     public static abstract class TableReference extends LivenessArtifact implements SimpleReference<Table> {
 
+        private final boolean isRefreshing;
+
         TableReference(Table t) {
-            if (t.isRefreshing()) {
+            isRefreshing = t.isRefreshing();
+            if (isRefreshing) {
                 manage(t);
             }
         }
@@ -311,7 +314,9 @@ public class DeferredViewTable extends RedefinableTable<DeferredViewTable> {
          *
          * @return true if the node is updating; false otherwise.
          */
-        public abstract boolean isRefreshing();
+        public final boolean isRefreshing() {
+            return isRefreshing;
+        }
 
         /**
          * Returns the table in a form that the user can run queries on it. This may be as simple as returning a
@@ -382,11 +387,6 @@ public class DeferredViewTable extends RedefinableTable<DeferredViewTable> {
         public SimpleTableReference(Table table) {
             super(table);
             this.table = table;
-        }
-
-        @Override
-        public boolean isRefreshing() {
-            return table.isRefreshing();
         }
 
         @Override
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/DeferredViewTable.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/DeferredViewTable.java
@@ -51,8 +51,9 @@ public class DeferredViewTable extends RedefinableTable<DeferredViewTable> {
         this.deferredFilters = deferredFilters == null ? WhereFilter.ZERO_LENGTH_SELECT_FILTER_ARRAY : deferredFilters;
         for (final WhereFilter sf : this.deferredFilters) {
             sf.init(parentDefinition);
-            if (sf instanceof LivenessReferent) {
+            if (sf instanceof LivenessReferent && sf.isRefreshing()) {
                 manage((LivenessReferent) sf);
+                setRefreshing(true);
             }
         }
 
@@ -61,7 +62,10 @@ public class DeferredViewTable extends RedefinableTable<DeferredViewTable> {
         final boolean haveView = this.deferredViewColumns.length > 0;
         final boolean haveFilter = this.deferredFilters.length > 0;
 
-        manage(tableReference);
+        if (tableReference.isRefreshing()) {
+            manage(tableReference);
+            setRefreshing(true);
+        }
 
         if (haveDrop && haveFilter) {
             throw new IllegalStateException("Why do we have a drop and a filter all at the same time?");
@@ -269,7 +273,6 @@ public class DeferredViewTable extends RedefinableTable<DeferredViewTable> {
     protected DeferredViewTable copy() {
         final DeferredViewTable result = new DeferredViewTable(definition, description, new SimpleTableReference(this),
                 null, null, null);
-        result.setRefreshing(isRefreshing());
         LiveAttributeMap.copyAttributes(this, result, ak -> true);
         return result;
     }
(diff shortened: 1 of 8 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/DeferredViewTable.java
```diff
@@ -307,6 +306,13 @@ public static abstract class TableReference extends LivenessArtifact implements
             }
         }
 
+        /**
+         * Is the node updating?
+         *
+         * @return true if the node is updating; false otherwise.
+         */
+        public abstract boolean isRefreshing();
```
rcaudy, line 314: It seems to me that this need not be abstract; we take a `table` in the constructor, don't we? I imagine we could write down at that time if it is refreshing or not.

engine/table/src/main/java/io/deephaven/engine/table/impl/PartitionAwareSourceTable.java
```diff
@@ -202,25 +202,22 @@ protected final BaseTable redefine(@NotNull final TableDefinition newDefinition)
                 description + "-retainColumns",
                 componentFactory, locationProvider, updateSourceRegistrar, partitioningColumnDefinitions,
                 partitioningColumnFilters);
-        final DeferredViewTable deferredViewTable = new DeferredViewTable(newDefinition, description + "-retainColumns",
+        Assert.equals(redefined.isRefreshing(), "redefined.isRefreshing()", isRefreshing(), "isRefreshing()");
```
rcaudy, line 205: Why did/do we need this assertion?

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -1314,22 +1316,6 @@ void handleUncaughtException(Exception throwable) {
                 });
     }
 
-    @SuppressWarnings("WeakerAccess")
-    protected WritableRowSet filterRows(RowSet currentMapping, RowSet fullSet, boolean usePrev,
```
rcaudy, line 1318: This was unused?

engine/table/src/main/java/io/deephaven/engine/table/impl/select/WhereFilter.java
```diff
@@ -20,7 +20,6 @@
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
-import java.util.stream.Collectors;
```
rcaudy, line 23: Since we're cleaning up, want to hit the unused import of:
`import io.deephaven.engine.table.impl.BlinkTableTools;`
 in `QueryTable`?

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
nbauernfeind: Yes, QueryTable#filterRows exists as ConjunctiveFilter#andImpl now. I came across the snippet when looking for for-loops that manage WhereFilters.

engine/table/src/main/java/io/deephaven/engine/table/impl/PartitionAwareSourceTable.java
nbauernfeind: We don't need it, but technically speaking we were setting refreshing from this and not from the redefined table. We pass the redefined table in as the source table, not this; and that adds a layer of indirection. I was just being carefully lazy.

The change passes nightlies, so I'm happy to drop the assertion if you would rather it be gone.

rcaudy (approved):

PR #4453, opened 2023-09-06
Rename make_table_from_ticket to ticket_to_table in R client API

jcferretti:


py/client/pydeephaven/session.py
```diff
@@ -413,12 +400,44 @@ def open_table(self, name: str) -> Table:
...
-                        raise DHError(f"no table by the name {name}") from None
+                        raise DHError(not_found_err) from None
                 raise e
             finally:
                 # Explicitly close the table without releasing it (because it isn't ours)
                 faketable.ticket = None
                 faketable.schema = None
+        
+    def ticket_to_table(self, ticket: str) -> Table:
+        """Opens a table in the global scope for the given ticket on the server.
```
chipkent, line 411: Does it have to be in the global scope?  I have no idea how tickets work.

jcferretti: I think so, I am not an expert on the arrow/ticket stuff, but given how `open_table` was implemented based on the same code I think that statement has to be true.

alexpeters1208 (commented):
R changes look good to me.

py/client/pydeephaven/session.py
niloc132: This is nothing to do with global scope, it just lets you take a string (should be bytes?). The existing method `open_table` is the "open a table in the global scope", and is documented as such. It looks like the comment was copied directly... and from the tests, it looks like the use case actually wants to call `open_table`?

py/client/tests/test_console.py
```diff
@@ -49,6 +49,9 @@ def vectorized_func(x, y):
         pa_table = self.session.open_table('demo_table').to_arrow()
         df = pa_table.to_pandas()
         self.assertEquals(1000, len(df.index))
+        pa_table = self.session.ticket_to_table('s/demo_table'.encode(encoding='ascii')).to_arrow()
```
niloc132, line 52: this is "wrong" - the ticket_to_table method says that it takes a string, but you're passing a `bytes`

```python
python
Python 3.11.3 (main, Jun  5 2023, 09:32:32) [GCC 13.1.1 20230429] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> type('s/demo_table'.encode(encoding='ascii'))
<class 'bytes'>
```

py/client/pydeephaven/session.py
jmao-denver: I am not crazy with the name `ticket_to_table`, would `open_ticket` be better and more consistent with the naming of other methods?

jmao-denver: Also @niloc132 found that in _console_services, we have an `open_table` that is so out of date it is wrong and not used anymore, @jcferretti Could you remove it?

jmao-denver (approved):
The Python changes LGTM

py/client/pydeephaven/_console_service.py
```diff
@@ -42,28 +42,6 @@ def run_script(self, server_script: str) -> Any:
...
-
-            if response.success:
-                return Table(self.session, ticket=response.result_id.ticket,
-                             schema_header=response.schema_header,
-                             size=response.size,
-                             is_static=response.is_static)
-            else:
-                raise DHError("error open a table: " + response.error_info)
-        except Exception as e:
-            raise DHError("failed to open a table.") from e
```
chipkent, line 65: I don't understand why this is being removed.  It seems reasonable for the client to ask the server for a table by name.

chipkent (changes requested):

py/client/pydeephaven/_console_service.py
jcferretti: That one method is not the public api method.
The public API method open_table is on session.py
That method is a method that Colin realized (1) is wrong and (2) is unused
So Jianfeng asked me to remove it.




[4:01](https://illumon.slack.com/archives/C053QMX2LBW/p1694030468643529)
The public API method open_table is on session.py
[4:01](https://illumon.slack.com/archives/C053QMX2LBW/p1694030498138679)
That method that is being removed in the PR latest version as of now is a method that Colin realized (1) is wrong and (2) is unused
[4:01](https://illumon.slack.com/archives/C053QMX2LBW/p1694030504057739)
So Jianfeng asked me to remove it.




[4:01](https://illumon.slack.com/archives/C053QMX2LBW/p1694030468643529)
The public API method open_table is on session.py
[4:01](https://illumon.slack.com/archives/C053QMX2LBW/p1694030498138679)
That method that is being removed in the PR latest version as of now is a method that Colin realized (1) is wrong and (2) is unused
[4:01](https://illumon.slack.com/archives/C053QMX2LBW/p1694030504057739)
So Jianfeng asked me to remove it.

PR #4456, opened 2023-09-07
Add skip initial table publish with kafka

### LINKED ISSUE #4442: Expose ability to skip initial table publish with kafka.
As a engineer working on deephaven, I would like to have the ability to configure whether or not the initial table state sent to a kafka publisher is published. 

The use case for this is that I have a local state file that I have configured Deephaven to leverage for storing intermediate aggregation states, and persisting them through restarts. This works flawlessly, however on startup currently emits every row from the saved state, this is unnecessary and just adds considerable load to both Kafka and our downstream services to reprocess already emitted data. Ideally we would have a toggle that we could pass the publisher to skip the initial table state sync.

devinrsmith:
Adds builder object KafkaPublishOptions

Fixes #4442

### DIFF
--- a/py/server/deephaven/stream/kafka/producer.py
+++ b/py/server/deephaven/stream/kafka/producer.py
@@ -11,6 +11,7 @@ from deephaven import DHError
 from deephaven.jcompat import j_hashmap, j_hashset, j_properties
 from deephaven._wrapper import JObjectWrapper
 from deephaven.table import Table
+from deephaven.update_graph import auto_locking_ctx
 
 _JKafkaTools = jpy.get_type("io.deephaven.kafka.KafkaTools")
 _JAvroSchema = jpy.get_type("org.apache.avro.Schema")
@@ -86,7 +87,8 @@ def produce(
             .build()
         )
 
-        runnable = _JKafkaTools.produceFromTable(options)
+        with auto_locking_ctx(table):
+            runnable = _JKafkaTools.produceFromTable(options)
 
         def cleanup():
             try:
--- a/py/server/deephaven/stream/kafka/producer.py
+++ b/py/server/deephaven/stream/kafka/producer.py
@@ -15,6 +15,7 @@ from deephaven.table import Table
 _JKafkaTools = jpy.get_type("io.deephaven.kafka.KafkaTools")
 _JAvroSchema = jpy.get_type("org.apache.avro.Schema")
 _JKafkaTools_Produce = jpy.get_type("io.deephaven.kafka.KafkaTools$Produce")
+_JKafkaPublishOptions = jpy.get_type("io.deephaven.kafka.KafkaPublishOptions")
 
 
 class KeyValueSpec(JObjectWrapper):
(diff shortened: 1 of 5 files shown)


rcaudy (approved):

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -32,12 +34,13 @@ def j_object(self) -> jpy.JType:
...
-    key_spec: KeyValueSpec,
-    value_spec: KeyValueSpec,
-    last_by_key_columns: bool = False,
+        table: Table,
+        kafka_config: Dict,
+        topic: str,
+        key_spec: KeyValueSpec,
+        value_spec: KeyValueSpec,
+        last_by_key_columns: bool = False,
+        publish_initial: bool = True,
```
jmao-denver, line 43: inconsistent indentation?

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -70,22 +75,28 @@ def produce(
...
+            runnable = _JKafkaTools.produceFromTable(options)
+
         def cleanup():
             try:
                 runnable.run()
             except Exception as ex:
-                raise DHError(ex, "failed to stop publishing to Kafka and the clean-up.") from ex
+                raise DHError(
+                    ex, "failed to stop publishing to Kafka and the clean-up."
+                ) from ex
```
jmao-denver, line 99: not sure why it needs to be multiple lines.

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -56,6 +59,8 @@ def produce(
         last_by_key_columns (bool): whether to publish only the last record for each unique key, Ignored if key_spec is
             KeyValueSpec.IGNORE. Otherwise, if last_by_key_columns is true this method will internally perform a last_by
             aggregation on table grouped by the input columns of key_spec and publish to Kafka from the result.
+        publish_initial (bool): whether the initial data in table should be published. When False, table.is_refreshing
+            must be True. By default, is True.
```
jmao-denver, line 63: Add a check for the case of  `publish_initial = False and not table.is_refreshing`?

jmao-denver (changes requested):

py/server/deephaven/stream/kafka/producer.py
devinrsmith: I used black formatter; but until we apply it everywhere, I'll revert this part and make a single line.

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -70,17 +75,23 @@ def produce(
...
-        kafka_config = j_properties(kafka_config)
-        runnable = _JKafkaTools.produceFromTable(
-            table.j_table,
-            kafka_config,
-            topic,
-            key_spec.j_object,
-            value_spec.j_object,
-            last_by_key_columns,
+        if not publish_initial and not table.is_refreshing:
+            raise ValueError("publishInitial == False and table.is_refreshing == False")
```
jmao-denver, line 79:
```suggestion
            raise ValueError("publish_initial == False and table.is_refreshing == False")
```

jmao-denver (approved):
LGTM

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3151
Reference: https://github.com/deephaven/deephaven.io/issues/3152

PR #4461, opened 2023-09-07
Setup ExecutionContext in Barrage Client Examples

nbauernfeind:
It was recently discovered that the barrage client examples were no longer working. This was a side effect of adding multiple update graph support to the engine. 

Additionally, there is server-side configuration that is accessible to the client that requires a class that is only in the server package. This server-side configuration is likely to be reworked in the near future and even possibly the need for this server-side class that can't be loaded. For now, I've modified the thread factory initialization to ignore class not found exceptions within the server-side package.

To work around this while still using a previous release one either needs to provide their own `dh-defaults.prop` file or add the following jvm flag `-Dthread.initialization=` to remove the reference to the server-side class.

From a documentation perspective, we should add setup instructions for the ticking java-client use case (as otherwise demonstrated by the example).

### DIFF
--- a/Util/src/main/java/io/deephaven/util/thread/ThreadInitializationFactory.java
+++ b/Util/src/main/java/io/deephaven/util/thread/ThreadInitializationFactory.java
@@ -5,6 +5,7 @@ import io.deephaven.configuration.Configuration;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 import java.util.stream.Collectors;
 
 /**
@@ -23,10 +24,19 @@ public interface ThreadInitializationFactory {
                     return clazz.getDeclaredConstructor().newInstance();
                 } catch (ClassNotFoundException | NoSuchMethodException | InvocationTargetException
                         | InstantiationException | IllegalAccessException e) {
+
+                    // Currently the default property file is shared between both the java client and the server. This
+                    // means that client-side usage will attempt to load the thread.initialization property intended for
+                    // the server which is not available on the class path.
+                    if (e instanceof ClassNotFoundException && type.startsWith("io.deephaven.server.")) {
+                        return null;
+                    }
+
                     throw new IllegalArgumentException(
                             "Error instantiating initializer " + type + ", please check configuration", e);
                 }
             })
+            .filter(Objects::nonNull)
             .collect(Collectors.toUnmodifiableList());
 
     /**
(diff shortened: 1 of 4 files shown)


Util/src/main/java/io/deephaven/util/thread/ThreadInitializationFactory.java
```diff
@@ -23,10 +24,19 @@ public interface ThreadInitializationFactory {
                     return clazz.getDeclaredConstructor().newInstance();
                 } catch (ClassNotFoundException | NoSuchMethodException | InvocationTargetException
                         | InstantiationException | IllegalAccessException e) {
+
+                    // Currently the default property file is shared between both the java client and the server. This
+                    // means that client-side usage will attempt to load the thread.initialization property intended for
+                    // the server which is not available on the class path.
+                    if (e instanceof ClassNotFoundException && type.startsWith("io.deephaven.server.")) {
+                        return null;
+                    }
```
devinrsmith, line 33: Let's create a ticket and link here; I think this is an "ok" workaround for now, but we should have a better / easier configuration answer for this in the future.

Util/src/main/java/io/deephaven/util/thread/ThreadInitializationFactory.java
```diff
@@ -23,10 +24,20 @@ public interface ThreadInitializationFactory {
                     return clazz.getDeclaredConstructor().newInstance();
                 } catch (ClassNotFoundException | NoSuchMethodException | InvocationTargetException
                         | InstantiationException | IllegalAccessException e) {
+
+                    // TODO (https://github.com/deephaven/deephaven-core/issues/4040):
```
devinrsmith, line 28: Is this the right ticket?

nbauernfeind: I believe so; I acquired it from Colin. It is the ticket related to python debug initialization. This snippet can be reverted once Colin takes an alternative approach.

devinrsmith (approved):

Util/src/main/java/io/deephaven/util/thread/ThreadInitializationFactory.java
niloc132: Yes, this is the ticket (in several parts...) that will address this.

niloc132 (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3162
Reference: https://github.com/deephaven/deephaven.io/issues/3161

PR #4465, opened 2023-09-08
Add `agg_all_by()` to the R client and add error handling around empty aggregations in `agg_by()`

alexpeters1208:
This PR _includes_ changes from #4464, so that should be merged first to minimize the diff. This PR is a _blocker_ on #4363, because the correctness of the docstrings for `agg_by()` depend on whether it will accept empty aggregations.

This PR introduces `agg_all_by()` to the R wrapping, and adds C++ binding to support it. It also removes the ability of `agg_by()` to accept empty aggregations like `th$agg_by(agg_min(), by = "grouping")` by adding additional checks at the R level.

The current C++ level implementation of `agg_all_by()` and `agg_by()` will change based on recent conversations over the deprecated `ComboAggRequest` that they both utilize, but this PR aims to provide a stable API that can be distributed and documented independently of any underlying C++ changes.


chipkent (changes requested):

R/rdeephaven/R/aggregate_wrapper.R
```diff
@@ -1,95 +1,85 @@
...
 agg_percentile <- function(percentile, cols = character()) {
   verify_in_unit_interval("percentile", percentile, TRUE)
   verify_string("cols", cols, FALSE)
-  return(Aggregation$new(INTERNAL_agg_percentile(percentile, cols)))
+  return(Aggregation$new(INTERNAL_agg_percentile, "agg_percentile", percentile=percentile, cols=cols))
 }
 
-#' @export
-agg_count <- function(col) {
+agg_count <- function(col = "n") {
```
chipkent, line 82: 1) Do we have this default in other clients?
2) We endorse camel case with starting caps for column names.  This does not match that convention.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -107,9 +107,18 @@ TableHandle <- R6Class("TableHandle",
...
+        if (!is.null(aggs[[i]]$.internal_num_cols) && aggs[[i]]$.internal_num_cols == 0) {
+          stop(paste0("Aggregations with no columns cannot be used in 'agg_by'. Got '", aggs[[i]]$.internal_agg_name, "' at index ", i, " with an empty 'cols' argument."))
+        }
+      }
       unwrapped_aggs <- lapply(aggs, strip_r6_wrapping)
       return(TableHandle$new(self$.internal_rcpp_object$agg_by(unwrapped_aggs, by)))
     },
+    agg_all_by = function(agg, by = character()) {
+      verify_type("agg", agg, "Aggregation", "Deephaven Aggregation", TRUE)
+      return(TableHandle$new(self$.internal_rcpp_object$agg_all_by(agg$.internal_rcpp_object, by)))
```
chipkent, line 120:
```suggestion
      verify_type("agg", agg, "Aggregation", "Deephaven Aggregation", TRUE)
      verify_string("by", by, FALSE)
      return(TableHandle$new(self$.internal_rcpp_object$agg_all_by(agg$.internal_rcpp_object, by)))
```

chipkent (changes requested):

R/rdeephaven/R/aggregate_wrapper.R
alexpeters1208: This default does not exist in other clients. I will remove it here.

chipkent (approved):

PR #4479, opened 2023-09-11
ParquetTableWriter Extract/Replicate Transfer Kernels

nbauernfeind:
This is a portion of DH-12664 pushdown/predicates in this [commit](https://github.com/deephaven-ent/iris/commit/7901a878f49b72461249739871fb5e3603f1b01d).

### DIFF
--- a/extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/EncodedTransfer.java
+++ b/extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/EncodedTransfer.java
@@ -1,0 +1,147 @@
+/**
+ * Copyright (c) 2016-2023 Deephaven Data Labs and Patent Pending
+ */
+package io.deephaven.parquet.table.transfer;
+
+import io.deephaven.base.verify.Assert;
+import io.deephaven.chunk.ObjectChunk;
+import io.deephaven.chunk.attributes.Values;
+import io.deephaven.engine.rowset.RowSequence;
+import io.deephaven.engine.table.ChunkSource;
+import io.deephaven.engine.table.ColumnSource;
+import org.apache.parquet.column.statistics.Statistics;
+import org.apache.parquet.io.api.Binary;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Arrays;
+
+/**
+ * Used as a base class of transfer objects for types like strings or big integers that need specialized encoding, and
+ * thus we need to enforce page size limits while writing.
+ */
+public abstract class EncodedTransfer<T> implements TransferObject<Binary[]> {
+    private final ColumnSource<?> columnSource;
+    private final ChunkSource.GetContext context;
+    private final Binary[] buffer;
+
+    private ObjectChunk<T, Values> chunk;
+
+    /**
+     * Number of objects buffered
+     */
+    private int bufferedDataCount;
+
+    /**
+     * The target size of data to be stored in a single page. This is not a strictly enforced "maximum" page size.
+     */
+    private final int targetPageSize;
+
+    /**
+     * Index of next object from the chunk to be buffered
+     */
+    private int currentChunkIdx;
+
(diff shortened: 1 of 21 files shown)


extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/TransferObject.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Map;
+
+/**
+ * Classes that implement this interface are responsible for converting data from individual DH columns into buffers
+ * to be written out to the Parquet file.
+ *
+ * @param <B>
+ */
+public interface TransferObject<B> extends SafeCloseable {
+    public static <DATA_TYPE> TransferObject<?> create(
```
rcaudy, line 26:
```suggestion

    static <DATA_TYPE> TransferObject<?> create(
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/TransferObject.java
```diff
@@ -0,0 +1,122 @@
...
+     *
+     * @return the buffer
+     */
+    B getBuffer();
+
+    /**
+     * After a sequence of calls to {@link #fetchData(RowSequence)} this may be called to append statistic information
+     * into the Parquet footer for readers.
+     */
+    <T extends Comparable<T>> void updateStatistics(@NotNull Statistics<T> stats);
```
rcaudy, line 121: This is done at a lower level, internal to the `io.deephaven.parquet.base.BulkWriter` implementations. Replicating statistics generation at this level adds negative value unless we clean up the lower level as well, but the lower level is doing a better job (tracking null count, for example).

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/IntTransfer.java
```diff
@@ -0,0 +1,102 @@
...
+    private IntChunk<? extends Values> chunk;
+    private int minValue = QueryConstants.NULL_INT;
+    private int maxValue = QueryConstants.NULL_INT;
+
+    public IntTransfer(
+            @NotNull final ColumnSource<?> columnSource,
+            final int targetSize) {
+        this.columnSource = columnSource;
+        this.buffer = IntBuffer.allocate(targetSize);
+        context = columnSource.makeGetContext(targetSize);
```
rcaudy, line 33: This is strictly inferior, in my opinion, for int, long, float, double, and Boolean. We seem to have standardized on the `getChunk` and copy implementation, instead of just directly filling the backing array for our chunk and our buffer at the same time. What's the upside?

I *do* really like that we're allocating the buffer and chunks inside the constructor now.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/EncodedTransfer.java
```diff
@@ -0,0 +1,147 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Arrays;
+
+/**
+ * Used as a base class of transfer objects for types like strings or big integers that need specialized encoding, and
+ * thus we need to enforce page size limits while writing.
+ */
+public abstract class EncodedTransfer<T> implements TransferObject<Binary[]> {
```
rcaudy, line 23: Why public instead of package-private?

rcaudy (changes requested):
I think this PR adds some value, by making `ParquetTableWriter` less cumbersome and enabling replication. That said, we should bring back to bifurcation between the good primitive transfers and the ones that have to copy to ints first. I also think we should consider keeping the class hierarchy we had, and just replicating sub-classes.

Likely the redundant statistics gathering should be ripped out, or removed from the lower level if we're certain it's better this way. If we're going to do that, null counting needs to be added at the higher level.

rcaudy (approved):

PR #4480, opened 2023-09-12
Retry decompressing with LZ4_RAW if LZ4 fails

malhotrashivam:
PR combined with #4446 

The retry mechanism is particularly useful for decompressing parquet files that are compressed with LZ4_RAW but tagged as LZ4 in the metadata.

Interesting reads:
- [LZ4 format ](https://docs.fileformat.com/compression/lz4/#:~:text=A%20file%20with%20.,be%20decompressed%20using%20the%20same)
- [LZ4_RAW or LZ4 Block format](https://android.googlesource.com/platform/external/lz4/+/HEAD/doc/lz4_Block_format.md)


extensions/parquet/compression/src/main/java/io/deephaven/parquet/compress/DeephavenCompressorAdapterFactory.java
```diff
@@ -166,6 +167,72 @@ public void close() {
...
+    /**
+     * Following adapter attempts to decompress with LZ4 and falls back to LZ4_RAW on failure. This is the default
+     * adapter for decompressing LZ4 data. The fallback mechanism is particularly useful for decompressing parquet files
+     * that are compressed with LZ4_RAW but tagged as LZ4 in the metadata.
+     */
+    private static class LZ4WithLZ4RawBackupCompressorAdapter implements CompressorAdapter {
+        /**
+         * Set to true only when decompressing for the first time
+         */
+        boolean tryFallback;
```
niloc132, line 179: fields should both be private

extensions/parquet/compression/src/main/java/io/deephaven/parquet/compress/DeephavenCompressorAdapterFactory.java
```diff
@@ -161,6 +167,54 @@ public void close() {
         }
     }
 
+    /**
+     * This is the default adapter for LZ4 data. It attempts to decompress with LZ4, and falls back to LZ4_RAW on
+     * failure. The fallback mechanism is particularly useful for decompressing parquet files that are compressed with
+     * LZ4_RAW but tagged as LZ4 in the metadata.
+     */
+    private static class LZ4WithLZ4RawBackupCompressorAdapter extends CodecWrappingCompressorAdapter {
```
niloc132, line 175: move this to its own file?

extensions/parquet/compression/src/main/java/io/deephaven/parquet/compress/DeephavenCompressorAdapterFactory.java
```diff
@@ -183,21 +237,30 @@ private DeephavenCompressorAdapterFactory(CompressionCodecFactory compressionCod
...
+                        String.format("Failed to find CompressionCodec for codecName=%s", codecName));
+            }
         }
         final CompressionCodecName ccn = codecClassnameToCodecName.get(codec.getClass().getName());
         if (ccn == null) {
             throw new IllegalArgumentException(String.format(
                     "Failed to find CompressionCodecName for codecName=%s, codec=%s, codec.getDefaultExtension()=%s",
                     codecName, codec, codec.getDefaultExtension()));
         }
+        if (ccn == CompressionCodecName.LZ4) {
```
niloc132, line 261: can't we use the configuration in createInstance() to replace this?

extensions/parquet/compression/src/main/java/io/deephaven/parquet/compress/DeephavenCompressorAdapterFactory.java
```diff
@@ -161,6 +167,54 @@ public void close() {
...
+                    lz4RawAdapter = DeephavenCompressorAdapterFactory.getInstance().getByName("LZ4_RAW");
+                }
+                return lz4RawAdapter.decompress(bufferedInputStream, compressedSize, uncompressedSize);
+            }
+        }
+
+        @Override
+        public void reset() {
+            super.reset();
+            if (lz4RawAdapter != null) {
```
niloc132, line 204: to confirm, for both this and close(), we don't want to null out the lz4raw codec?

extensions/parquet/compression/src/main/java/io/deephaven/parquet/compress/DeephavenCompressorAdapterFactory.java
niloc132: Oh I see - this is specifically to say "take the expected LZ4, but wrap it in the backup impl (which will in turn look up a lz4raw instance from config if needed". Otherwise, we have to hardcode the lz4 impl (since we don't want it to look itself up).

malhotrashivam: Yup, exactly.

extensions/parquet/compression/src/main/java/io/deephaven/parquet/compress/DeephavenCompressorAdapterFactory.java
malhotrashivam: The docs for reset: `Reset the internal state of this {@link CompressorAdapter} so more rows can be read or written.` 
So doesn't make sense to deallocate lz4raw.

For close, I can null out the codec

niloc132: I think no need to close it out if the instance cant be used again anyway - it looks like we get close() by way of SafeCloseable, so no one should attempt to use it after close().

malhotrashivam: Cool

niloc132:
Might be simpler to see this folded into #4446, but I can see it either way.

malhotrashivam:
Combining this PR with #4446

Issue #4476, opened 2023-09-08
test_function_generated_table flakiness


devinrsmith:
test_function_generated_table python tests is causing failures in unrelated PRs. Seen a few so far, there more be more:

```
2023-09-08T22:02:38.2501245Z ======================================================================
2023-09-08T22:02:38.2501890Z FAIL [0.021s]: test_generated_table_2triggers (tests.test_function_generated_table.TableTestCase)
2023-09-08T22:02:38.2502793Z ----------------------------------------------------------------------
2023-09-08T22:02:38.2503320Z Traceback (most recent call last):
2023-09-08T22:02:38.2503935Z   File "/python/tests/test_function_generated_table.py", line 107, in test_generated_table_2triggers
2023-09-08T22:02:38.2504545Z     self.assertEqual(result_int, 54321)
2023-09-08T22:02:38.2505034Z AssertionError: 12345 != 54321
```

```
2023-09-08T20:48:57.1451774Z ======================================================================
2023-09-08T20:48:57.1452760Z FAIL [0.032s]: test_generated_table_1trigger (tests.test_function_generated_table.TableTestCase)
2023-09-08T20:48:57.1453750Z ----------------------------------------------------------------------
2023-09-08T20:48:57.1454456Z Traceback (most recent call last):
2023-09-08T20:48:57.1455282Z   File "/python/tests/test_function_generated_table.py", line 53, in test_generated_table_1trigger
2023-09-08T20:48:57.1456158Z     self.assertEqual(result_str, 'test string')
2023-09-08T20:48:57.1456947Z AssertionError: None != 'test string'
2023-09-08T20:48:57.1457340Z 
2023-09-08T20:48:57.1457681Z ======================================================================
2023-09-08T20:48:57.1458531Z FAIL [0.011s]: test_generated_table_2triggers (tests.test_function_generated_table.TableTestCase)
2023-09-08T20:48:57.1459442Z ----------------------------------------------------------------------
2023-09-08T20:48:57.1460143Z Traceback (most recent call last):
2023-09-08T20:48:57.1460961Z   File "/python/tests/test_function_generated_table.py", line 84, in test_generated_table_2triggers
2023-09-08T20:48:57.1461849Z     self.assertEqual(result_str, 'test string')
2023-09-08T20:48:57.1462625Z AssertionError: None != 'test string'
2023-09-08T20:48:57.1463025Z 
2023-09-08T20:48:57.1463434Z ----------------------------------------------------------------------
[...]

PR #4483, opened 2023-09-12
Fix for inconsistent tests

rbasralian:
Improvements to the intermittently-failing tests in `test_function_genreated_table.py` (#4476)

(2/3 fixed — `test_generated_table_1trigger` still a mystery)

### DIFF
--- a/py/server/tests/test_function_generated_table.py
+++ b/py/server/tests/test_function_generated_table.py
@@ -1,9 +1,11 @@
 from datetime import datetime
+from typing import Any
 
 import deephaven.dtypes as dht
 from deephaven import empty_table, input_table, new_table, update_graph, function_generated_table
 from deephaven.column import string_col, int_col
 from deephaven.execution_context import get_exec_ctx
+from deephaven.table import Table
 from tests.testbase import BaseTestCase
 
 
@@ -23,12 +25,14 @@ class TableTestCase(BaseTestCase):
         with update_graph.exclusive_lock(self.test_update_graph):
             result_table = function_generated_table(table_generator_function, refresh_interval_ms=2000)
             self.assertEqual(result_table.size, 1)
-            initial_time = result_table.j_table.getColumnSource("Timestamp").get(0)
+            first_row_key = get_row_key(0, result_table)
+            initial_time = result_table.j_table.getColumnSource("Timestamp").get(first_row_key)
 
             if not result_table.await_update(5_000):
                 raise RuntimeError("Result table did not update within 5 seconds")
 
-            later_time = result_table.j_table.getColumnSource("Timestamp").get(0)
+            first_row_key = get_row_key(0, result_table)
+            later_time = result_table.j_table.getColumnSource("Timestamp").get(first_row_key)
 
         # Make sure it ticked at least once within 5 seconds. It should have ticked twice,
         # but leaving a wider margin to ensure the test passes -- as long as it ticks at all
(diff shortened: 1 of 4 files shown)


py/server/tests/test_function_generated_table.py
```diff
@@ -39,20 +42,26 @@ def test_generated_table_1trigger(self):
...
         result_table = function_generated_table(table_generator_function, source_tables=append_only_input_table)
 
         self.assertEqual(result_table.size, 0)
 
+        print("Adding row at time: " + str(datetime.now()))
         append_only_input_table.add(new_table([string_col(name='MyStr', data=['test string'])]))
+        print("add() returned at time: " + str(datetime.now()))
         self.wait_ticking_table_update(result_table, row_count=1, timeout=30)
-        self.assertEqual(result_table.size, 1)
+        print("result_table has 1 row at time: " + str(datetime.now()))
```
jmao-denver, line 56: So the row somehow vanished. Is the table a blink table?

py/server/tests/test_function_generated_table.py
```diff
@@ -39,20 +42,26 @@ def test_generated_table_1trigger(self):
...
 
         result_str = result_table.j_table.getColumnSource("ResultStr").get(0)
         self.assertEqual(result_str, 'test string')
```
rcaudy, line 59: This is where we were failing, right @rbasralian ? @jmao-denver or @chipkent, is it possible that this is the wrong way to compare strings? I see some stuff in the unittest.TestCase documentation about type-specific equals that was added to assertEquals in 3.1. Is that Python 3.1, or v3.1 of the testcase package, or?

py/server/tests/test_function_generated_table.py
```diff
@@ -39,20 +42,26 @@ def test_generated_table_1trigger(self):
...
 
     def test_generated_table_2triggers(self):
+        # NOTE: This tests that both trigger tables cause the refresh function to run.
+        # It does not test updating two source tables in the same cycle (that is covered by
+        # io.deephaven.engine.table.impl.util.TestFunctionGeneratedTableFactory.testMultipleSources).
```
rcaudy, line 64: I don't see where this is tested at all. There's no "wait".

py/server/tests/test_function_generated_table.py
rbasralian: nah — it's an input table, which is backed by an AppendOnlyArrayBackedMutableTable

py/server/tests/test_function_generated_table.py
```diff
@@ -39,20 +42,26 @@ def test_generated_table_1trigger(self):
...
         self.assertEqual(result_table.size, 0)
 
+        print("Adding row at time: " + str(datetime.now()))
         append_only_input_table.add(new_table([string_col(name='MyStr', data=['test string'])]))
+        print("add() returned at time: " + str(datetime.now()))
         self.wait_ticking_table_update(result_table, row_count=1, timeout=30)
-        self.assertEqual(result_table.size, 1)
+        print("result_table has 1 row at time: " + str(datetime.now()))
 
         result_str = result_table.j_table.getColumnSource("ResultStr").get(0)
```
rcaudy, line 58: Don't do `get(0)`, use the row set.

py/server/tests/test_function_generated_table.py
jmao-denver: `assertEqual` is the right way. The logged error is "None not equal to 'test string' which indicates `get(0)` returning a null value.

chipkent (approved):

rcaudy (approved):

PR #4490, opened 2023-09-13
Add messageStream support for java client

### LINKED ISSUE #4489: Support messageStream via java client

devinrsmith:
Fixes #4489


java-client/session/src/main/java/io/deephaven/client/impl/FetchedObject.java
```diff
@@ -45,16 +52,42 @@ public void writeTo(OutputStream out) throws IOException {
...
         return "FetchedObject{" +
                 "type='" + type + '\'' +
                 ", bytes=" + bytes +
-                ", exportIds=" + exportIds +
+                ", exports=" + exports +
                 '}';
     }
+
+    /**
+     * Closes all of {@link #exports}.
```
niloc132, line 85: Why does close() release the exports, but not the object itself? How do you release the object itself - shouldn't FetchableObject.close() release the object's own ticket? 

Releasing the exports as the default operation doesn't really make sense to me - there's no reason that they shouldn't be able to outlive the object that first provided them.

java-client/session/src/main/java/io/deephaven/client/impl/FetchedObject.java
```diff
@@ -5,28 +5,35 @@
...
 import java.util.Objects;
+import java.util.stream.Collectors;
 
-public final class FetchedObject {
+/**
+ * The results from fetching an object.
+ *
+ * @see Session#fetchObject(HasTypedTicket)
+ */
+public final class FetchedObject implements Closeable {
```
niloc132, line 21: In light of the other new types, I'm not sure that this class makes sense by itself - shouldn't this implement ServerObject or at least HasTypedTicket/etc? This type functionally is the result of fetching a ServerObject using the legacy/deprecated FetchObject call - perhaps we deprecate it and avoid adding new  functionality to it, encourage users to move to CustomObject, which should fully replace it?

java-client/session/src/main/java/io/deephaven/client/impl/TableObject.java
```diff
@@ -0,0 +1,39 @@
...
+    TableObject(Session session, ExportId exportId) {
+        super(session, exportId);
+        checkType(TYPE, exportId);
+    }
+
+    /**
+     * Creates a table handle.
+     *
+     * <p>
+     * Note: the table handle lifecycle is managed separately from {@code this}; the caller is still responsible for
```
niloc132, line 26: Why is that true? Won't either one release the ticket that references the table?

niloc132 (commented):
Would like at least some simple tests that export a sample custom/table/unknown object and confirms that the wiring built here behaves as expected.

java-client/session/src/main/java/io/deephaven/client/impl/FetchedObject.java
devinrsmith: B/c there may or may _not_ be ownership of the original object. The interface exposes fetching via any arbitrary ticket; for example, when you fetch from the global scope, you don't close that object

```protobuf
message FetchObjectRequest {
    io.deephaven.proto.backplane.grpc.TypedTicket source_id = 1;
}
```

In the cases where you _are_ fetching from an export id, the caller _does_ have the control to manage the life of the object itself and the new typed_export_ids separately.

java-client/session/src/main/java/io/deephaven/client/impl/TableObject.java
devinrsmith: `io.deephaven.client.impl.TableHandleManager#execute(io.deephaven.qst.table.TableSpec)` under the covers does batch-able ops (TableSpec); in this respect, it does a `FetchTable`, and isn't even wired up for `GetExportedTableCreationResponse`.

It's possible that we could extend the implementation to support creating `TableHandle` via exports you already have, but that would take some amount of extra work.

Plot/Plot.gradle
```diff
@@ -17,7 +17,7 @@ dependencies {
     implementation project(':engine-tuple')
     implementation project(':log-factory')
 
-    Classpaths.inheritGroovy(project, 'groovy', 'implementation')
+    Classpaths.inheritGroovy(project, 'groovy', 'api')
```
niloc132, line 20: technically true, but I am not sure those are actually usable?

java-client/session/src/main/java/io/deephaven/client/impl/TableObject.java
niloc132: Ok, so fixing this oversight will be a breaking change later, since clients today may close the TableObject after the TableHandle is a real Table, but after the fix closing the TableObject will shut down the TableHandle? Seems like it might be worth it to fix now, rather than break later?

java-client/session/src/main/java/io/deephaven/client/impl/FetchedObject.java
niloc132: To confirm, it is definitely deprecated for removal, and we want to remove the extra junk in the impl that exists only to bridge the gap for the unary impl. Ryan reviewed and allowed some of the ugliness with the understanding that it was going away "Soon".

What is the difference between FetchedObject and a ServerObject that is only fetchable? With FetchObject removed, is it possible to create a FetchedObject, either now or in the future? The plugin can't specify that one vs the other is the only way to fetch an object either.

Plot/Plot.gradle
devinrsmith: My test did not compile when this was listed as `implementation`; I think it has to do w/ the FigureFactory / interface.

niloc132: Huh. Worth tracking down, but ... not today.

java-client/session/src/main/java/io/deephaven/client/impl/FetchedObject.java
niloc132: If you fetch from the global scope without a re-export you're probably asking for trouble - consider having the API transparently (race a call to) reexport with the messagestream/fetch. Technically neither actually stops you from handing over a non-export ticket, but its a bad idea to use non-export tickets (and we document that at least).

I'm more scared of the "ServerObject.close() sometimes closes the object and leaves the children, and sometimes closes the children and leaves the object" - it doesn't look like it is going to encourage good code, but instead require that otherwise "generic" code typechecks what it is talking to before each close() call. Might as well make them different methods...

Plot/Plot.gradle
devinrsmith:
```
> Task :java-client-session-dagger:compileTestJava FAILED
/home/devin/dev/deephaven/deephaven-core/java-client/session-dagger/src/test/java/io/deephaven/client/ObjectServiceTest.java:101: error: cannot access Closure
        return FigureFactory.figure().plot("Test", t, "X", "Y").show();
                                     ^
  class file for groovy.lang.Closure not found
1 error
```

java-client/session/src/main/java/io/deephaven/client/impl/ExportId.java
```diff
@@ -17,6 +16,9 @@ public final class ExportId implements HasExportId {
     private final int exportId;
 
     ExportId(String type, int exportId) {
```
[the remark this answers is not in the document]
niloc132: On the one hand, I really hate exposing these to users at all - it should be "here's an object, as far as you know it is local" (like we do with tables). On the other hand, this is a package-protected ctor, and a user shouldn't care about this. Hiding this away in `impl` seems reasonable?

java-client/session/src/main/java/io/deephaven/client/impl/TableObject.java
devinrsmith: I don't expect we'd break this in the future; we can just add a new method.

java-client/session/src/main/java/io/deephaven/client/impl/FetchedObject.java
devinrsmith: ServerObject.close() never closes the children; it only ever closes the export associated w/ the ServerObject.

devinrsmith:
The latest push removes `FetchedObject` and provides more general `DataAndExports` object (which can be used in both the fetch and messageStream cases), as well as providing export functionality. Also, it provides a distinction between the high-level and low-level interfaces, while still sharing some of the common objects. The testing also explicitly executes the same logic via the high-level and low-level interfaces.

java-client/session/src/main/java/io/deephaven/client/impl/HasTypedTicket.java
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright (c) 2016-2023 Deephaven Data Labs and Patent Pending
+ */
+package io.deephaven.client.impl;
+
+public interface HasTypedTicket extends HasTicketId {
```
niloc132, line 6: Given that this is explicitly for the low level "i already know how the underlying impl and wire format works" api, whats the purpose in interfaces like this, that only have a single implementation? I get it for the api's internals with multiple impls, or exposing types through the high level api with a low level impl (with hidden other details), but I don't see the benefit here.

java-client/session/src/main/java/io/deephaven/client/impl/ObjectService.java
```diff
@@ -6,11 +6,123 @@
...
+         */
+        MessageStream<DataAndTypedTickets> messageStream(MessageStream<DataAndExports> stream);
+    }
+
+    /**
+     * The sending and receiving interface for {@link #messageStream(HasTypedTicket, MessageStream)}.
+     *
+     * @param <Message> the message type
+     */
+    interface MessageStream<Message> {
```
niloc132, line 69: Name collision with server code (and both are on the same classpath if you are writing server code) - not necessarily actionable, but worth keeping an eye on.

Will come up again if this is the low level api and we end up with a high level client interface that likely has the same name...

java-client/session/src/main/java/io/deephaven/client/impl/Session.java
```diff
@@ -32,8 +32,8 @@ public interface Session
     // ----------------------------------------------------------
 
     /**
-     * Advanced usage, creates a new export ID for {@code this} session, but must be managed by the caller. Useful for
-     * more advanced integrations, particularly around doPut. Callers are responsible for {@link #release(ExportId)
+     * Advanced usage, creates a new table export ID for {@code this} session, but must be managed by the caller.Useful
```
niloc132, line 35: probably should put that removed space back - but why can't this be used for non-table exports any more?

java-client/session/src/main/java/io/deephaven/client/impl/Session.java
```diff
@@ -32,8 +32,8 @@ public interface Session
...
+     * for more advanced integrations, particularly around doPut. Callers are responsible for {@link #release(ExportId)
```
niloc132, line 36: I can't comment on release(), but it should also have its comment edited, esp the "**only**"

server/src/main/java/io/deephaven/server/session/SessionState.java
```diff
@@ -708,15 +708,14 @@ public T get() {
             if (session != null && session.isExpired()) {
                 throw Exceptions.statusRuntimeException(Code.UNAUTHENTICATED, "session has expired");
             }
-
+            final T localResult = result;
```
niloc132, line 711: tweaked for debugging, or does this serve a purpose?

devinrsmith: `result` is volatile, so we should only do the volatile read once in this method for correctness (also, performance, but I doubt that matters here).

java-client/session/src/main/java/io/deephaven/client/impl/Session.java
devinrsmith: The existing implementation is really catered around "Table"; mostly a lack of other use cases or naivety when these interfaces were created. To correct this in the future if necessary, we can add:

```java
ExportId newExportId(String type);
```

java-client/session/src/main/java/io/deephaven/client/impl/ObjectService.java
```diff
@@ -6,11 +6,123 @@
...
+         * Presently it is required that the server respond immediately, at least to acknowledge that the object was
+         * correctly contacted (as opposed to waiting for a pending ticket, or dealing with network lag, etc). This is a
+         * small (and possibly not required, but convenient) departure from a offering a gRPC stream (a server-streaming
+         * or bidi-streaming call need not send a message right away).
+         *
+         * @param stream the stream where the client will receive messages
+         * @return the stream where the client will send messages
+         * @see Session#messageStream(HasTypedTicket, MessageStream) for the lower-level interface
+         */
+        MessageStream<DataAndTypedTickets> messageStream(MessageStream<DataAndExports> stream);
```
niloc132, line 61: I missed your note about how this patch now includes a high level api, and if I'm reading the api properly, the `Bidirectional` type is returned by the high level api (so itself is part of the high level api). But it still returns a stream implementation that is dependent on high-level client code being able to transform tickets into objects (like tables, other plugins, etc)?

java-client/session/src/main/java/io/deephaven/client/impl/DataAndExports.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
+
+/**
+ * Data and exports are sent from the server to the client as part of a {@link Fetchable#fetch() fetch} or
+ * {@link Bidirectional#messageStream(MessageStream) bidirectional message stream}. The client is responsible for
+ * managing the {@link #exports()}.
+ */
+public final class DataAndExports implements Closeable {
```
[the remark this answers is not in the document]
devinrsmith: So, I see this class as being common for both the high- and low-level APIs. The high-level APIs are the entrypoints:

```java
CompletableFuture<? extends Fetchable> fetchable(HasTypedTicket typedTicket);
CompletableFuture<? extends Bidirectional> bidirectional(HasTypedTicket typedTicket);
```

while the low-level APIs are the entrypoints:

```java
CompletableFuture<DataAndExports> fetch(HasTypedTicket typedTicket);
MessageStream<DataAndTypedTickets> messageStream(HasTypedTicket typedTicket, MessageStream<DataAndExports> stream);
```

I'm not against choosing better names for it; but I _do_ find just `Data` a bit lacking. It's also in juxtaposition w/ `DataAndTypedTickets`.

java-client/session/src/main/java/io/deephaven/client/impl/HasTypedTicket.java
devinrsmith: I don't necessarily see `HasTypedTicket` as exclusively "low-level".

I think it's reasonable from a high-level api to do something like:

```java
objectService.fetchable(new ScopeId("my_custom_object").ticketId().toTypedTicket("my_object_type"))
```

(excuse the un-ergonimics of it; we _can_ make it better...)

In this case, the user knows they have a globally scoped variable named "my_custom_object" of custom type "my_object_type", and they didn't need to know the low-level wire details beyond that. 

Generally, I'm happy to make strides in the java client code to better hide abstractions the user's don't / shouldn't need to know about.

java-client/session/src/main/java/io/deephaven/client/impl/ObjectService.java
devinrsmith: `Fetchable` and `Bidirectional` are the high-level apis; the user's shouldn't need to know anything about transforming tickets. Both of these are client-managed exports, and both present `DataAndExports` to the users. The important part, also what I consider high-level, is `ServerObject` (`List<ServerObject> exports`); these are also client-managed exports, but provide users differentiation between: `TableObject`, `CustomObject extends Fetchable, Bidirectional`, and `UnknownObject`.

java-client/session/src/main/java/io/deephaven/client/impl/Session.java
devinrsmith: What's the suggestion for changing the release javadoc wrt "only"?

java-client/session/src/main/java/io/deephaven/client/impl/ObjectService.java
rcaudy, line 61: I think this should be called "connect". It maps to sending a ConnectRequest over the MessageStream RPC.
I think the argument should include something in the name to make it clear that this is the "receiveStream" or similar.

java-client/session/src/main/java/io/deephaven/client/impl/ObjectService.java
```diff
@@ -6,11 +6,123 @@
...
+        /**
+         * The type.
+         *
+         * @return the type
+         */
+        String type();
+
+        /**
+         * Opens a bidirectional message stream for {@code this}. The returned {@link DataAndExports} messages are
+         * managed separately than {@code this}.
```
rcaudy, line 43:
```suggestion
         * managed separately from {@code this}.
```

java-client/session/src/main/java/io/deephaven/client/impl/TableObject.java
```diff
@@ -0,0 +1,39 @@
...
+package io.deephaven.client.impl;
+
+import io.deephaven.client.impl.TableHandle.TableHandleException;
+import io.deephaven.qst.table.TableSpec;
+import io.deephaven.qst.table.TicketTable;
+
+/**
+ * A {@value TYPE} object.
+ */
+public final class TableObject extends ServerObjectBase implements ServerObject {
```
rcaudy, line 13:
```suggestion
public final class TableObject extends ServerObjectBase {
```

java-client/session/src/main/java/io/deephaven/client/impl/TableObject.java
```diff
@@ -0,0 +1,39 @@
...
+     * Creates a table handle.
+     *
+     * <p>
+     * Note: the table handle lifecycle is managed separately from {@code this}; the caller is still responsible for
+     * {@link #close() closing} {@code this}.
+     *
+     * @return the table handle.
+     * @see Session#execute(TableSpec)
+     */
+    public TableHandle executeTable() throws TableHandleException, InterruptedException {
```
rcaudy, line 32: Should this be `toTable()`, and give ownership to the `TableHandle`?

java-client/session/src/main/java/io/deephaven/client/impl/UnknownObject.java
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright (c) 2016-2023 Deephaven Data Labs and Patent Pending
+ */
+package io.deephaven.client.impl;
+
+import io.deephaven.client.impl.ObjectService.Bidirectional;
+import io.deephaven.client.impl.ObjectService.Fetchable;
+
+/**
+ * A server object that is <b>not</b> {@link Fetchable} nor {@link Bidirectional}; the server does not have a
```
rcaudy, line 10:
```suggestion
 * A server object that is <b>neither</b> {@link Fetchable} nor {@link Bidirectional}; the server does not have a
```

java-client/session/src/main/java/io/deephaven/client/impl/UnknownObject.java
```diff
@@ -0,0 +1,21 @@
...
+ * registration for the underlying object.
```
rcaudy, line 11: What does "the server does not have a registration for the underlying object" mean (to a user)?

java-client/session/src/main/java/io/deephaven/client/impl/UnknownObject.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+public final class UnknownObject extends ServerObjectBase implements ServerObject {
```
rcaudy, line 13:
```suggestion
public final class UnknownObject extends ServerObjectBase {
```

java-client/session/src/main/java/io/deephaven/client/impl/CustomObject.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * A custom server object that may be {@link Fetchable} or {@link Bidirectional}. The client is responsible for
+ * implementing the necessarily retrieval logic according to the specific {@link #type()}.
+ *
+ * @see <a href="https://github.com/deephaven/deephaven-core/issues/4486">deephaven-core#4486</a>
+ * @see <a href="https://github.com/deephaven/deephaven-core/issues/4487">deephaven-core#4487</a>
+ * @see <a href="https://github.com/deephaven/deephaven-core/issues/4488">deephaven-core#4488</a>
+ */
+public final class CustomObject extends ServerObjectBase
+        implements ServerObject, Fetchable, Bidirectional {
```
rcaudy, line 21:
```suggestion
        implements Fetchable, Bidirectional {
```

java-client/session/src/main/java/io/deephaven/client/impl/ObjectService.java
```diff
@@ -6,11 +6,123 @@
 import java.util.concurrent.CompletableFuture;
 
 public interface ObjectService {
+
+    /**
+     * A server object that supports fetching.
```
rcaudy, line 11: What's "fetching"? (A logical unary RPC)

java-client/session/src/main/java/io/deephaven/client/impl/ObjectService.java
```diff
@@ -6,11 +6,123 @@
...
+    interface Fetchable extends ServerObject {
+        /**
+         * The type.
+         *
+         * @return the type
+         */
+        String type();
+
+        /**
+         * Fetches {@code this}. The resulting object is managed separately than {@code this}.
```
rcaudy, line 22:
```suggestion
         * Fetches {@code this}. The resulting object is managed separately from {@code this}.
```

java-client/session/src/main/java/io/deephaven/client/impl/ObjectService.java
```diff
@@ -6,11 +6,123 @@
...
+     *
+     * @param <Message> the message type
+     */
+    interface MessageStream<Message> {
+        void onData(Message message);
+
+        void onClose();
+    }
+
     /**
```
rcaudy, line 75: I can't figure out when we would use these.

java-client/session/src/main/java/io/deephaven/client/impl/ObjectService.java
```diff
@@ -6,11 +6,123 @@
...
+     */
+    CompletableFuture<? extends Bidirectional> bidirectional(HasTypedTicket typedTicket);
+
+    /**
+     * The low-level interface for fetching data. See {@link #fetchable(HasTypedTicket)} for a higher-level interface.
+     *
+     * @param typedTicket the typed ticket
+     * @return the future
+     */
+    CompletableFuture<DataAndExports> fetch(HasTypedTicket typedTicket);
```
rcaudy, line 99: Maybe we could have overloads that just take typed tickets? Why do we need to "have" a typed ticket? I almost want something that is close to the gRPC.

java-client/session/src/main/java/io/deephaven/client/impl/Session.java
rcaudy: We should introduce the proposed `newExportId(String type)`, document that, and have this one delegate to it with documentation that it's defaulting type to "table".

java-client/session/src/main/java/io/deephaven/client/impl/Export.java
```diff
@@ -31,7 +31,12 @@ public final class Export implements AutoCloseable, HasExportId {
 
     @Override
     public ExportId exportId() {
-        return new ExportId("Table", state.exportId());
+        return new ExportId(TableObject.TYPE, state.exportId());
```
rcaudy, line 34: Maybe we should deprecate this class? It's name-squatting on something generic, but it's table-specific.

java-client/session/src/main/java/io/deephaven/client/impl/DataAndExports.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
+
+/**
+ * Data and exports are sent from the server to the client as part of a {@link Fetchable#fetch() fetch} or
+ * {@link Bidirectional#messageStream(MessageStream) bidirectional message stream}. The client is responsible for
+ * managing the {@link #exports()}.
```
rcaudy, line 24: What does "managing" mean? Seems like "keep track of and close when no longer needed".

java-client/session/src/main/java/io/deephaven/client/impl/DataAndExports.java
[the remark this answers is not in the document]
rcaudy: Couldn't we just call it "ServerMessage"?

java-client/session/src/main/java/io/deephaven/client/impl/DataAndExports.java
```diff
@@ -0,0 +1,98 @@
...
+    static DataAndExports of(Session session, FetchObjectResponse fetch) {
+        return new DataAndExports(fetch.getData(), toServerObjects(session, fetch.getTypedExportIdsList()));
+    }
+
+    private static List<ServerObject> toServerObjects(Session session,
+            List<io.deephaven.proto.backplane.grpc.TypedTicket> exportTickets) {
+        return exportTickets
+                .stream()
+                .map(TypedTicket::of)
+                .map(TypedTicket::toExportId)
```
rcaudy, line 41: This is enforcing a documented convention, but theoretically a TypedTicket could hold any ticket, right?

java-client/session/src/main/java/io/deephaven/client/impl/DataAndExports.java
```diff
@@ -0,0 +1,98 @@
...
+    private final ByteString data;
+    private final List<ServerObject> exports;
+
+    /**
+     * Constructs a new instance. Callers should not modify {@code exports} after construction.
+     *
+     * @param data the bytes
+     * @param exports the exports
+     */
+    DataAndExports(ByteString data, List<ServerObject> exports) {
```
rcaudy, line 55:
```suggestion
    private DataAndExports(ByteString data, List<ServerObject> exports) {
```

java-client/session/src/main/java/io/deephaven/client/impl/DataAndTypedTickets.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Data and typed tickets to be sent from the client to the server as part of a
+ * {@link Bidirectional#messageStream(MessageStream) bidirection message stream}.
+ */
+public final class DataAndTypedTickets {
```
rcaudy, line 19: Maybe we want to make the proto asymmetric, too?
Same kind of quibble regarding this class name. ClientMessage? MessageToServer?

devinrsmith: I agree - I think it makes sense to have two separate proto objects so they can be documented independently.

java-client/session/src/main/java/io/deephaven/client/impl/TableObject.java
devinrsmith: This was noted as an improvement / feature we could make in the future with some amount of non-trivial work.

java-client/session/src/main/java/io/deephaven/client/impl/ObjectService.java
devinrsmith: It's actually even less than an unary RPC - w/ an unary RPC you can at least have a request message; this is more like an unary RPC where you can't have any request message parameters.

java-client/session/src/main/java/io/deephaven/client/impl/ObjectService.java
devinrsmith: Anytime you might re-use the export it makes sense to export it first. A bit handwavy, but might theoretically want to keep your parent as an export b/c you need to reference it wrt the children:

```java
try (Fetchable myVar = fetchable("s/my_var")) {
  ServerObject child = myVar.fetch().exports().get(0);
  toChild = ((Bidirectional)child).connect(...);
  toChild.onData(new DataAndTypedTickets(..., List.of(myVar));
  ...
}
```

I can also theoretically come up w/ other patterns if necessary...

java-client/session/src/main/java/io/deephaven/client/impl/ObjectService.java
devinrsmith: I think we should leave that for larger java client abstraction redesigns if necessary.

java-client/session/src/main/java/io/deephaven/client/impl/DataAndTypedTickets.java
niloc132: If we're renaming the proto types, should be ClientData/ServerData, or we're renaming the message oneof (or we end up with foo.message.message, which we wanted to avoid).

java-client/session/src/main/java/io/deephaven/client/impl/DataAndExports.java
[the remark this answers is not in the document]
devinrsmith: Renamed to `ServerData` (and `ClientData`).

java-client/session/src/main/java/io/deephaven/client/impl/Session.java
devinrsmith: I don't think introducing that would be of any use right now b/c there is no low-level way to register that export id w/ a server instance (save from the Table case with a doPut).

java-client/session/src/main/java/io/deephaven/client/impl/Export.java
devinrsmith: https://github.com/deephaven/deephaven-core/issues/4534

java-client/session/src/main/java/io/deephaven/client/impl/TicketId.java
```diff
@@ -4,33 +4,35 @@
...
 
 import com.google.protobuf.ByteStringAccess;
-import io.deephaven.proto.backplane.grpc.Ticket;
 import io.deephaven.qst.table.TicketTable;
 
+import java.nio.charset.StandardCharsets;
 import java.util.Objects;
 
+;
+
```
rcaudy, line 13:
```suggestion
```

rcaudy (commented):
.

rcaudy (dismissed):

java-client/session/src/main/java/io/deephaven/client/impl/Session.java
niloc132: I'm saying that this method is now documented to be suitable only for tables, that's the substantive change here.

The javadoc for `release()` says that it should "**only** be called in combination with exports returned from newExportId" (not my emphasis) - and since newExportId now says that it is only for tables, release is also only for tables.

plugin/src/main/java/io/deephaven/plugin/type/EchoObjectType.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * An {@link ObjectType object type plugin} named {@value NAME} that echos back the received contents back to the
+ * client. When {@link #compatibleClientConnection(Object, MessageStream)} is called, an empty payload is sent to the
+ * client.
+ *
+ * <p>
+ * The only object that is compatible with this plugin is {@link #INSTANCE}.
+ */
+@AutoService(ObjectType.class)
+public final class EchoObjectType extends ObjectTypeBase {
```
niloc132, line 16: Do we intend this to be checked in? It probably should be in its own test jar, but not automatically part of every distribution?

java-client/session/src/main/java/io/deephaven/client/impl/Session.java
devinrsmith: I think the documentation is still valid as-is; currently, `release()` is only meant for table export ids, but in the future if we add a way to get exports in another way, ie `newExportId(String type)`, then the language at that time can be changed to say "only use it with `newExportId()` or `newExportId(String type)`".

plugin/src/main/java/io/deephaven/plugin/type/EchoObjectType.java
devinrsmith: I was thinking it could be generally useful for production deployments that wanted to test new plugin development; but if we don't want to promote it as such, happy to refactor.

java-client/session/src/main/java/io/deephaven/client/impl/Session.java
rcaudy: I'm not sure I follow, @niloc132 ; Devin updated the JavaDoc for `newExportId()` to reflect the current reality, but `ExportId` and `release()` are not inherently Table specific.

niloc132 (approved):

PR #4497, opened 2023-09-14
Enhance Python LivenessScope API

### LINKED ISSUE #4471: Liveness scopes are confusing and hard to use in python plugins
### Summary
Each gRPC call has its own liveness scope, but console commands specially use the ScriptSession singleton itself as a liveness scope, ostensibly to ensure that objects created in the script session remain live while still reachable. This means that when an ObjectType plugin is instantiated or receives a call from a client, any objects that it does not explicitly retain in another liveness scope will be released when the call finishes.

Add to this, the python deephaven.liveness_scope package only offers a way to open a liveness scope and close it right away, after some code synchronously ends. This makes it difficult/impossible to retain objects created for the lifetime of the bidi connection.

### Steps to reproduce:
* During creation of a bidi plugin or while reacting to a message, invoke `deephaven.table_listener.listen()` with a ticking table and some kind of listener function that will be obvious that it was called (for example calling `print()`), taking care to retain a reference to the new listener object.

Expected: The table listener should be called when the table changes.

Actual: The table listener is destroyed by the liveness scope immediately.

#### Leverage Python refcounting
The current gc implementation in cpython is to use refcounting to determine when an object can no longer be referenced and it should be cleaned up. Deephaven's own refcounting implementation came out of a need to stop objects from doing work immediately after they are no longer referenced rather than waiting for memory pressure to trigger GC. Python also has a mark-and-sweep gc that runs periodically to handle cyclical references.

The rough idea is to add a DH liveness node for each python interpreter (presently only one), and then change deephaven._wrapper.JObjectWrapper to add its wrapped object to that liveness node when it is created. Then, in the `__del__` method, remove the object from the interpreter liveness node. Quick experiments show that this lets DH promptly stop and release nodes when python is the only thing holding a reference to it, and would let us remove our "anything in python globals must be retained" code.

Unfortunately, our current weaknesses in promptly releasing PyObject references from Java make this untenable at this time - until Java GC runs, the references are often retained, and so Python can't know that it is safe to run `__del__`. Until we are better behaved with our python references, this option isn't tenable.

#### Better Python LivenessScope API
The existing deephaven.liveness_scope package only handles a few use cases, and isn't sufficient for this task.
 * Scope a set of synchronous actions in a single `with` block - when that block ends, the scope will close, and release their hold on any objects seen
 * Within such a scope, preserve an object by the next outermost scope, which is either the containing scope used when Java called into python, the global liveness scope, or another scope (as in the first point) which will also promptly close
 * Technically, it is also possible to create a liveness scope without using it in a `with`, and then invoke `close()`. Unfortunately, calling `liveness_scope()` will push the scope onto the scope stack, and calling `close()` will pop it. Since it is pushed, we can't return control back to Java (as Java will attempt to pop its own scope and fail), so we can't meaningfully change the lifetime of this scope. Confusingly, even if we could, calling `with` doesnt open the scope, but exiting does, so we can't use it in more than one `with` block, and the only way to manage an object is through `preserve`, which will peek at the previous item, so we would need _another_ scope to use the previous one.

We need a few more features, or at least changes to existing features. Proposal:
 * Creation of a liveness scope shouldn't push it, only entering a `with` block should do so
 * Exiting a `with` block should be capable of only popping it, but not necessarily closing it. The default should be to close, to keep existing code compatible.
 * Explicit calls to `close()` should close the scope, but not pop off the stack.
 * Consider an explicit `manage()` and `unmanage()` call that accepts JObjectWrappers or Java objects, rather than only using `with` to replace the current scope. For example this would allow transferring an object from one scope to another.

This should enable plugins to declare their own scopes and manage them across multiple calls.

niloc132:
The existing Python APIs for Liveness were difficult to use correctly in cases more complex than a single script - this brings several parts of the Java LivenessScope type to Python to enable more advanced use cases. Rather than introduce a LivenessScopeStack, this adds an open() method to LivenessScope itself, and through deprecation we're encouraging its use over the liveness_scope() factory method being directly passed to a with block. Likewise, close() is deprecated in favor of release(), which has different semantics, and matches the name used in Java.

Also makes TableListenerHandle a JObjectWrapper so it can be more easily unwrapped, can't accidentally be managed as a LivePyObjectWrapper in Java.

Fixes #4471


niloc132:
As part of this PR, I'll add a how-to to our docs, providing some better introduction to using liveness scopes.

py/server/deephaven/liveness_scope.py
```diff
@@ -27,44 +50,136 @@ def __init__(self, j_scope: jpy.JType):
...
+        True if the object is a LivenessReferent, False otherwise.
+    """
+    if isinstance(referent, jpy.JType) and _JLivenessReferent.jclass.isInstance(referent):
+        return True
+    if isinstance(referent, JObjectWrapper):
+        return is_liveness_referent(referent.j_object)
+    return False
 
 
 def liveness_scope() -> LivenessScope:
```
niloc132, line 185: @jmao-denver would it make sense to take this away (or deprecate it) and just encourage the LivenessScope constructor instead?

py/server/deephaven/liveness_scope.py
```diff
@@ -27,44 +50,136 @@ def __init__(self, j_scope: jpy.JType):
...
+        except Exception as e:
+            raise DHError(e, message="failed to manage object")
+
+    def unmanage(self, referent: Union[JObjectWrapper, jpy.JType]) -> None:
+        """
+        Explicitly unmanage the given java object from this scope. Must only be passed a Java LivenessReferent, or
+        a Python wrapper around a LivenessReferent
+
+        Args:
+            referent:
```
niloc132, line 154:
```suggestion
            referent: the object to unmanage from this scope
```

py/server/deephaven/liveness_scope.py
```diff
@@ -8,11 +8,34 @@
...
+        _JLivenessScopeStack.pop(scope)
+    except Exception as e:
+        raise DHError(e, message="failed to pop the LivenessScope from the stack.")
+
+
+def _unwrap_to_liveness_referent(referent: Union[JObjectWrapper, jpy.JType]) -> jpy.JType:
+    if isinstance(referent, jpy.JType) and _JLivenessReferent.jclass.isInstance(referent):
+        return referent
+    if isinstance(referent, JObjectWrapper):
+        _unwrap_to_liveness_referent(referent.j_object)
```
niloc132, line 37:
```suggestion
        return _unwrap_to_liveness_referent(referent.j_object)
```

Found by fixing tests - so far at least the other test failures have to do with the bad patterns that we forced on users (including our own tests).

py/server/deephaven/liveness_scope.py
```diff
@@ -27,44 +50,136 @@ def __init__(self, j_scope: jpy.JType):
...
+        referent = _unwrap_to_liveness_referent(referent)
         try:
             _JLivenessScopeStack.pop(self.j_scope)
-            _JLivenessScopeStack.peek().manage(wrapper.j_object)
-            _JLivenessScopeStack.push(self.j_scope)
+            _JLivenessScopeStack.peek().manage(_unwrap_to_liveness_referent(referent))
         except Exception as e:
             raise DHError(e, message="failed to preserve a wrapped object in this LivenessScope.")
+        finally:
+            _JLivenessScopeStack.push(self.j_scope)
```
niloc132, line 126: The original code was wrong, but this is differently wrong... working on a fix with more comments.

py/server/deephaven/liveness_scope.py
```diff
@@ -27,44 +50,136 @@ def __init__(self, j_scope: jpy.JType):
...
+            DHError if the referent isn't a LivenessReferent, or if it is no longer live
+
+        """
+        referent = _unwrap_to_liveness_referent(referent)
+        try:
+            self.j_scope.manage(referent)
+        except Exception as e:
+            raise DHError(e, message="failed to manage object")
+
+    def unmanage(self, referent: Union[JObjectWrapper, jpy.JType]) -> None:
```
rcaudy, line 148: Maybe we should be wrapping `tryManage` and `tryUnmanage`, too. Arguably, they are more useful inside of Java implementation classes.

py/server/deephaven/liveness_scope.py
```diff
@@ -74,5 +189,4 @@ def liveness_scope() -> LivenessScope:
         a LivenessScope
     """
     j_scope = _JLivenessScope()
-    _JLivenessScopeStack.push(j_scope)
```
rcaudy, line 77: If there's existing documentation, we should be sure to note this change.

py/server/deephaven/liveness_scope.py
niloc132: Yeah I wasn't sure if we should have these anyway - mostly meant as an easier (and doc'd) escape hatch for NextTimeThisHappens, rather than sending the user to figure out jpy fun As jpy goes, this isn't too bad at least.

py/server/deephaven/liveness_scope.py
niloc132: The only docs are in this this file, as part of this PR i'll make a real how-to doc.

py/server/tests/testbase.py
```diff
@@ -35,9 +35,11 @@ def tearDownClass(cls) -> None:
 
     def setUp(self) -> None:
         self._liveness_scope = liveness_scope()
+        self.opened_scope = liveness_scope().open()
```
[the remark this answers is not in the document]
rcaudy: What we had before was too simple, and prevented an important use case: Keeping a LivenessScope around in order to fully control the life cycle of the objects created under it, and/or re-using the same LivenessScope on multiple calls. That is, the old API was simply not fit for purpose.

py/server/deephaven/liveness_scope.py
rcaudy: I'm not sure what we're concluding. `tryManage` and `tryUnmanage` exist, but mostly for cases internal to the engine/operations where we have race conditions around whether a user may have caused an object's reference count to reach zero while we're concurrently doing things with the object. Meaning, I'd be OK:
1. leaving this as-is and not wrapping those methods at this time, or
2. wrapping them now to avoid future work.

rcaudy: We decided on #1.

rcaudy (commented):
.

py/server/deephaven/liveness_scope.py
```diff
@@ -27,44 +50,143 @@ def __init__(self, j_scope: jpy.JType):
         self.j_scope = j_scope
 
     def __enter__(self):
+        warn('Instead of passing liveness_scope() to with, call open() on it first',
+             DeprecationWarning, stacklevel=2)
```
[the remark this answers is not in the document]
niloc132: Welcome to python I guess? its already imported by default for some reason: https://docs.python.org/3/library/exceptions.html?highlight=deprecationwarning#DeprecationWarning
```
$ python
Python 3.10.11 (main, May 22 2023, 14:37:10) [GCC 13.1.1 20230429] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> print(DeprecationWarning)
<class 'DeprecationWarning'>
```

py/server/tests/testbase.py
```diff
@@ -35,9 +35,11 @@ def tearDownClass(cls) -> None:
 
     def setUp(self) -> None:
         self._liveness_scope = liveness_scope()
```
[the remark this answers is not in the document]
niloc132: Yes, I missed that.

py/server/deephaven/liveness_scope.py
```diff
@@ -27,44 +50,143 @@ def __init__(self, j_scope: jpy.JType):
...
         """
-        if self.j_scope:
-            try:
-                _JLivenessScopeStack.pop(self.j_scope)
-                self.j_scope.release()
-                self.j_scope = None
-            except Exception as e:
-                raise DHError(e, message="failed to close the LivenessScope.")
+        warn('This function is deprecated, prefer release(). Use cases that rely on this are likely to now fail.',
+             DeprecationWarning, stacklevel=2)
```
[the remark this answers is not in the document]
niloc132: The advantage (I leave it to you/Jianfeng if this is a good one or not) that the deprecation warning will tell you what to fix, but just removing the method will just break and let you sort it out yourself.

py/server/tests/testbase.py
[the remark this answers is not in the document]
niloc132: This particular case is bad, because apparently the unittest "fixtures" can't be wrapped in contextmanagers or a `with` expr, etc - you can only explicitly open something, then close it later. That isn't a pattern we want to encourage here, since you might screw up and not close it. Apparently pytest does this better, closer to how junit solves this.

The other changes made in this patch's tests are more representative of how we expect this to work.

py/server/deephaven/liveness_scope.py
```diff
@@ -27,44 +50,143 @@ def __init__(self, j_scope: jpy.JType):
...
+        _pop(self.j_scope)
+        self.release()
+
+    @contextlib.contextmanager
+    def open(self, release_after_block: bool = True) -> "LivenessScope":
```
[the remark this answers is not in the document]
niloc132: Both "work" (after `s/LivenessScope/livenessScope/`), in that the first one will warn "you're doing it wrong, this will not work correctly", thus "strongly encouraging" the user to call open(). We can outright remove this right away and do away with deprecation warnings, or wait a version or two for anyone who happens to be using this and needs to transition away.

The new version is strictly more powerful, as it doesn't conflate creating/releasing a scope with adding/removing it on the stack. We can't really have it both ways... The other option (discussed with Ryan) is to make open() an optional step so that you _can_ call it either way, but this makes the Java and Python usage different/confusing going back and forth.

One other option is that we deprecate liveness_scope() itself, and make the LivenessScope() constructor usable, so we can leave liveness_scope() intact.

py/server/deephaven/liveness_scope.py
```diff
@@ -27,44 +50,143 @@ def __init__(self, j_scope: jpy.JType):
...
+        Raises:
+            DHError
+        """
+        try:
+            self.j_scope.release()
+            self.j_scope = None
+        except Exception as e:
+            raise DHError(e, message="failed to close the LivenessScope.")
 
     def close(self):
```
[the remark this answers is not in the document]
niloc132: Deprecated, to be removed.

py/server/deephaven/liveness_scope.py
niloc132: Dropping this in favor of just a regular ctor call, updating class docs

niloc132 (commented):
Summary of call with @chipkent and @jmao-denver:

Two use cases to support, named "A2" and "B2" on the call:

Simplest example, where a scope is created+opened for a `with` block, and closed+released at the end:
```python
with liveness_scope(): # optional "as scope"
    # do things here that will be managed by the scope
    ...
```
This is the existing factory method, and the use case that it was intended for. The return value of that method is a LivenessScope, and can only be used to `manage(...)`/`unmanage(...)` or `preserve(...)` during that block.

Richer case, where the lifetime of the scope is longer than the time in which it was at the top of the liveness stack":
```python
scope = LivenessScope()
with scope.open():
    # do things here that will be managed by the scope
    ...

# later at some point release the scope (can be re-opened in the meantime)
scope.release()
```
Instead of the factory method, this uses a constructor and an `open()` instance method. Scopes can be opened more than once, and must then be released when the user is finished with them.

The `liveness_scope()` method will return a specific type that only supports manage/unmanage/preserve, whereas the `LivenessScope()`  constructor will also support the `open()` and `release()` methods.

jmao-denver (dismissed):

py/server/deephaven/liveness_scope.py
```diff
@@ -3,76 +3,249 @@
...
+            raise DHError(e, message="failed to unmanage object")
+
+    @property
+    def j_object(self) -> jpy.JType:
+        return self.j_scope
+
+
+class SimpleLivenessScope(_BaseLivenessScope):
+    """
+    A SimpleLivenessScope automatically managed reference counting of tables and other query resources that
```
chipkent, line 181:
```suggestion
    A SimpleLivenessScope automatically manages reference counting of tables and other query resources that
```

py/server/tests/test_liveness.py
```diff
@@ -80,6 +129,78 @@ def test_liveness_nested(self):
...
+
+        l_scope.release()
+
+        self.assertFalse(to_discard.j_table.tryRetainReference())
+        self.assertFalse(to_discard_2.j_table.tryRetainReference())
+
+        self.assertTrue(must_keep.j_table.tryRetainReference())
+        # drop extra reference
+        must_keep.j_table.dropReference()
+
```
chipkent, line 203: The `liveness_scope` pydoc indicates that `@liveness_scope` can be used to control scope, but I don't see a test for this case.

niloc132: added (and otherwise cleaned up tests a bit).

typehints were wrong for both contextmanager usages, i fixed them as well.

chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3277
Reference: https://github.com/deephaven/deephaven.io/issues/3276

PR #4500, opened 2023-09-15
Use new C++ base image that removes local from the directory layout.

jcferretti:
This simplifies our deployment layout, eliminated one unnecessary level of indirection (the intermediate `.../local/...` directory).
The C++ base image being adopted was already modified to do this, and also trim down the number of files
in the final deployment hierarchy; for builds that include debug information, we are still keeping the source files of 
all libraries, including dependent, but the object files and in general the intermediate results of compilation are not necessary after install is done, so this trims down our install footprint considerably.

### DIFF
--- a/cpp-client/README.md
+++ b/cpp-client/README.md
@@ -37,7 +37,7 @@ on them anymore so we do notguarantee they are current for those platforms.
    Get the `build-dependencies.sh` script from Deephaven's base images repository
    at the correct version.
    You can download it directly from the link
-   https://raw.githubusercontent.com/deephaven/deephaven-base-images/72427ce29901bf0419dd05db8e4abf31b57253d9/cpp-client/build-dependencies.sh
+   https://raw.githubusercontent.com/deephaven/deephaven-base-images/4eab90690888fbd5abfcbaf5dbacc750e90d9c2f/cpp-client/build-dependencies.sh
    (this script is also used from our automated tools, to generate a docker image to
    support tests runs; that's why it lives in a separate repo).
    The script downloads, builds and installs the dependent libraries
@@ -64,7 +64,7 @@ on them anymore so we do notguarantee they are current for those platforms.
    # If the directory already exists from a previous attempt, ensure is clean/empty
    mkdir -p $DHCPP
    cd $DHCPP
-   wget https://raw.githubusercontent.com/deephaven/deephaven-base-images/72427ce29901bf0419dd05db8e4abf31b57253d9/cpp-client/build-dependencies.sh
+   wget https://raw.githubusercontent.com/deephaven/deephaven-base-images/4eab90690888fbd5abfcbaf5dbacc750e90d9c2f/cpp-client/build-dependencies.sh
    chmod +x ./build-dependencies.sh
    # Maybe edit build-dependencies.sh to reflect choices of build tools and build target, if you
    # want anything different than defaults; defaults are tested to work,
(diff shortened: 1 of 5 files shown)


cpp-client/README.md
```diff
@@ -108,14 +108,27 @@ on them anymore so we do notguarantee they are current for those platforms.
...
+   (which may be helpful for debugging the paths being used etc),
+   you can set the environment variable `VERBOSE=1`.
+   This is true in general of cmake-generated Makefiles.
+   It is useful to keep the output of make for later reference
+   in case you need to be sure of the exact compiler flags
+   that were used to compile the library and its objects.
+   You can tweak the last command above as
+
+   ```
+   VERBOSE=1 make -j$NCPUS install 2>&1 | tee make-install.log
```
kosak, line 129: you can also say `make VERBOSE=1`...  maybe that's simpler or it doesn't matter.

proto/proto-backplane-grpc/src/main/proto/build-cpp-protos.sh
```diff
@@ -2,11 +2,11 @@
 
 set -euxo pipefail
 
-: ${DHCPP_LOCAL:=$HOME/dhcpp/local}
+: ${DHCPP_LOCAL:=$HOME/dhcpp}
 : ${CPP_PROTO_BUILD_DIR:=build}
 
 mkdir -p "${CPP_PROTO_BUILD_DIR}"
-$DHCPP_LOCAL/protobuf/bin/protoc \
+$DHCPP_LOCAL/bin/protoc \
```
kosak, line 9: I kind of wonder two things:
1. whether this should be DHCPP not DHCPP_LOCAL  (i.e. why "local")
2. whether this should not be set at all, and we should tell people to use the same $DHCPP variable they got when they ran the "env.sh" file we generated for them

alexpeters1208 (dismissed):

proto/proto-backplane-grpc/src/main/proto/build-cpp-protos.sh
jcferretti: I think you are right about 1, changed and pushed.
Ref 2, there is one level of indirection that is always necessary, something needs to point to where the env.sh file lives; so we need something to know where the $PREFIX deployment was made.  The way this code is written, it is essentially assuming a default (which I don't think is most of the time used... this script is meant for manual execution, and you/me should come here with the env.sh file sourced already).  The syntax
```
: ${POPO:=popo_default}
```
essentially tells the shell that if the `POPO` environment variable is already defined, use its value, otherwise if it is not defines, use `popo_default`.
So maybe I would question the default and we could change the logic to "check if the variable is not defined, and if it is not, bail with an error".
?

cpp-client/README.md
jcferretti: I feel a bit icky about `make VERBOSE=1 -j32` because of the order of definition versus flag, but it happens to work, at least in GNU make, and in practice all these linux distributiosn are GNU.
I believe I prefer the `environment variable for the process` regular shell syntax tho, because that's more general.
Do you care if we just leave it as is?

Update: Corey and myself discussed, we are keeping it as is for now.

proto/proto-backplane-grpc/src/main/proto/build-cpp-protos.sh
jcferretti: We agreed on a check of the var + fail instead of default.

alexpeters1208 (approved):

kosak (approved):

devinrsmith (approved):

rcaudy (approved):

PR #4502, opened 2023-09-16
auto convert DH nulls in params to py UDF

### LINKED ISSUE #4208: DH numeric nulls are not converted to  appropriate Python NaN or Pandas.NA when passing to Python UDFs in queries.
We also need to make sure return values of Python UDFs , if they Python float("nan"), Pandas.NaN/Pandas.NA, etc.,  are properly translated into DH null or Nan .

jmao-denver:
Fixes #4208


py/server/tests/test_pyfunc_param_null.py
```diff
@@ -0,0 +1,170 @@
...
+                    res = tbl.update("Col4 = custom_func4(Col2)")
+                    self.assertEqual(4, res.to_string().count("true"))
+
+                @dh_null_to_nan
+                @dh_vectorize
+                def custom_func5(col) -> bool:
+                    if np.isnan(col):
+                        return True
+                    else:
+                        return False
```
chipkent, line 97: It looks like scalars are being tested with (1) type hints and (2) with vectorize but not without type hints and without vectorize.  Seems like a missing test.

py/server/tests/test_pyfunc_param_null.py
```diff
@@ -0,0 +1,170 @@
...
+
+                with self.subTest(data_type):
+                    tbl = empty_table(100).update([col1_formula, col2_formula])
+                    res = tbl.update("Col5 = custom_func5(Col2)")
+                    self.assertEqual(4, res.to_string().count("true"))
+
+    def test_dh_null_to_na_array(self):
+        @dh_null_to_na
+        def custom_func(c0l1, col2):
```
chipkent, line 106:
```suggestion
        def custom_func(col1, col2):
```

py/server/tests/test_pyfunc_param_null.py
```diff
@@ -0,0 +1,170 @@
...
+                custom_func3 = eval(custom_func3_str, globals(), locals())
+
+                res = tbl.update("Col3 = custom_func(Col1, Col2)")
+                res2 = tbl.update("Col3 = custom_func2(Col1, Col2)")
+                res3 = tbl.update("Col3 = custom_func3(Col1, Col2)")
+                if data_type in ["double", "float"]:
+                    if data_type == "double":
+                        self.assert_table_equals(res2, res3)
+                else:
+                    self.assert_table_equals(res2, res3)
```
chipkent, line 130: Isn't this basically the same as:
```
if data_type != "float":
    do test...
else:
    don't test...
````
If so, I think this presentation is more clear.  Also why not testing float?

File: py/server/tests/test_pyfunc_param_null.py [hunk 4]
```diff
@@ -0,0 +1,170 @@
...
+                res = tbl.update("Col4 = custom_func4(Col2)")
+                self.assertEqual(4, res.to_string().count("true"))
+
+                @dh_null_to_na
+                @dh_vectorize
+                def custom_func5(col) -> bool:
+                    if col is pd.NA:
+                        return True
+                    else:
+                        return False
```
chipkent, line 161: same question here about not testing the non-hint non-vectorized case.

py/server/tests/test_vectorization.py
```diff
@@ -129,6 +130,7 @@ def py_const_str(s) -> str:
         self.assertEqual(deephaven.table._vectorized_count, expected_count)
 
     def test_multiple_formulas(self):
+        @dh_null_to_nan
```
chipkent, line 133: why is this added to this test?

py/server/tests/test_vectorization.py
```diff
@@ -143,9 +145,44 @@ def pyfunc(p1, p2, p3) -> int:
...
+        def pyfunc(p1, p2, p3) -> float:
+            return p1 + p2 + p3
+
+        t = empty_table(1).update("X = i").update(["Y = pyfunc(X, i, 33)", "Z = pyfunc(X, ii, 66)"])
+        self.assertEqual(deephaven.table._vectorized_count, 1)
+        self.assertIn("33", t.to_string(cols=["Y"]))
+        self.assertIn("66", t.to_string(cols=["Z"]))
+        self.assertEqual(t.columns[1].data_type, dtypes.double)
+
+    def test_null_conv(self):
```
chipkent, line 176: It seems like this test should go before the tests that have two annotations.

py/server/deephaven/pandas.py
chipkent, line 1: file not yet reviewed

chipkent (changes requested):

py/server/deephaven/pandas.py
```diff
@@ -279,3 +299,101 @@ def to_table(df: pd.DataFrame, cols: List[str] = None) -> Table:
...
+                        converted_args.append(arg)
+
+            return fn(*converted_args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+def dh_null_to_nan(fn):
```
chipkent, line 371: 1. Should this function be in `numpy.py`?
2. Should this handle strings and map null to `""`?

py/server/deephaven/pandas.py
```diff
@@ -279,3 +299,101 @@ def to_table(df: pd.DataFrame, cols: List[str] = None) -> Table:
...
+    function.
+
+    Scalar type parameters must be type-hinted with precise numpy dtypes in order to be converted correctly,
+    e.g. np.int32, np.float64, etc. If the type-hint is not precise, the conversion will fail. if no type-hint is
+    provided, the conversion will skip the parameter.
+
+    Array type parameters don't need to be type-hinted, their types can be automatically detected for the conversion
+    be done correctly. Note that the conversion automatically promotes arrays to pd.Series of corresponding nullable
+    types.
+    """
```
chipkent, line 398: same comments from above.

chipkent (changes requested):

py/server/deephaven/pandas.py
```diff
@@ -279,3 +299,101 @@ def to_table(df: pd.DataFrame, cols: List[str] = None) -> Table:
...
+        be done correctly.
+        Note that:
+            if the new_nv is np.nan, the conversion automatically promotes arrays of integer types to array of float64.
+            if the new_nv is pd.NA, the conversion automatically promotes arrays to pd.Series of corresponding nullable
+            types.
+        """
+        fn_sig = _encode_signature(fn)
+
+        @wraps(fn)
+        def wrapper(*args, **kwargs):
```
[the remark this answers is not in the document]
jmao-denver: The query engine does Vector -> Array conversion before passing a Vector to the UDF.

py/server/deephaven/pandas.py
```diff
@@ -279,3 +299,101 @@ def to_table(df: pd.DataFrame, cols: List[str] = None) -> Table:
...
+            for arg, np_dtype_char in zip(args, fn_sig):
+                if np_dtype := _J_ARRAY_NP_TYPE_MAP.get(type(arg)):
+                    np_array = np.frombuffer(arg, dtype=np_dtype)
+                    dtype = dtypes.from_np_dtype(np_dtype)
+                    if new_nv is np.nan:
+                        dh_null = dtypes.PRIMITIVE_DTYPE_NULL_MAP.get(dtype)
+                        if dtype in (dtypes.float32, dtypes.float64):
+                            np_array = np.copy(np_array)
+                        else:
+                            np_array = np_array.astype(np.float64)
```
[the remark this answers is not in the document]
jmao-denver: They will be skipped. The map tables should guarantee that.

py/server/deephaven/pandas.py
```diff
@@ -279,3 +299,101 @@ def to_table(df: pd.DataFrame, cols: List[str] = None) -> Table:
...
+    return decorator
+
+
+def dh_null_to_nan(fn):
+    """A decorator for user defined functions used in queries. It replaces Deephaven null values for primitive data
+    types (byte, short, char, int, long, float, double) in the arguments with numpy.nan before calling the decorated
+    function.
+
+    Scalar type parameters must be type-hinted with precise numpy dtypes in order to be converted correctly,
+    e.g. np.int32, np.float64, etc. If the type-hint is not precise, the conversion will fail. if no type-hint is
```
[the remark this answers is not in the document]
jmao-denver: Yes and no. For non-nullable arrays, Pandas pretty much uses Numpy's dtypes. But it also has its own nullable extension dtypes and supports pyarrow dtypes from 2.0 and above. So it is pretty confusing to use its predefined 'dtypes' to annotate a parameter. Numpy dtypes are much cleaner for our purpose.

py/server/deephaven/pandas.py
jmao-denver: 1. ideally yes, but that causes a cyclic import problem, we probably should just expose them in the top level so that we don't have this impression of misplacement
2. Nulls of Strings or non-primitive object types are returned as None via JPY.

py/server/deephaven/pandas.py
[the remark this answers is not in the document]
chipkent: I forgot about that.

py/server/deephaven/pandas.py
```diff
@@ -315,17 +315,17 @@ def to_table(df: pd.DataFrame, cols: List[str] = None) -> Table:
...
         """A decorator that replaces Deephaven null values in the parameters with the provided replacement value (
-        either numpy.nan or pandas.NA for now) before calling the decorated function.
+        either numpy.nan or pandas.NA) before calling the decorated function.
 
-        Scalar type parameters must be type-hinted with precise numpy dtypes in order to be converted correctly,
-        e.g. np.int32, np.float64, etc. If the type-hint is not precise, the conversion will fail. if no type-hint is
-        provided, the conversion will skip the parameter.
+        Scalar parameters must be type-hinted with precise numpy dtypes in order to be converted correctly,
+        e.g. np.int32, np.float64, etc. If the type-hint is not precise, the conversion will use the wrong Deephaven
+        null value and yield unintended result. if no type-hint is provided, the conversion will skip the parameter.
```
chipkent, line 322:
```suggestion
        null value and yield unintended result. If no type-hint is provided, the conversion will skip the parameter.
```

py/server/deephaven/pandas.py
```diff
@@ -374,11 +374,13 @@ def dh_null_to_nan(fn):
     function.
 
     Scalar type parameters must be type-hinted with precise numpy dtypes in order to be converted correctly,
-    e.g. np.int32, np.float64, etc. If the type-hint is not precise, the conversion will fail. if no type-hint is
-    provided, the conversion will skip the parameter.
+    e.g. np.int32, np.float64, etc. If the type-hint is not precise, the conversion will use the wrong Deephaven
+    null value and yield unintended result. if no type-hint is provided, the conversion will skip the parameter.
```
chipkent, line 378:
```suggestion
    null value and yield unintended result. If no type-hint is provided, the conversion will skip the parameter.
```

py/server/deephaven/pandas.py
```diff
@@ -389,11 +391,13 @@ def dh_null_to_na(fn):
...
```
chipkent, line 395:
```suggestion
    null value and yield unintended result. If no type-hint is provided, the conversion will skip the parameter.
```

py/server/tests/test_pyfunc_param_null.py
```diff
@@ -0,0 +1,222 @@
...
+    else:
+        return False
+"""
+                exec(func, globals())
+                with self.subTest(data_type):
+                    tbl = empty_table(100).update([col1_formula, col2_formula])
+                    res = tbl.update("Col4 = custom_func4(Col2)")
+                    self.assertEqual(4, res.to_string().count("true"))
+
+                # vectorization implicitly replaces scalar argument with Java arrays which dh_null_to_nam is able to
```
chipkent, line 91:
```suggestion
                # vectorization implicitly replaces scalar argument with Java arrays which dh_null_to_nan is able to
```

py/server/deephaven/pandas.py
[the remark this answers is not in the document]
jmao-denver: Me neither. how about just 'null_to_nan` and 'null_to_na`, or `np_nan` and `pd_na`?

py/server/deephaven/table.py
```diff
@@ -363,108 +368,306 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+                else:
+                    pa.encoded_types.add(p)
+                    if p in {"b", "h", "i", "l"}:
+                        pa.int_char = p
+                    if p in {"f", "d"}:
+                        pa.floating_char = p
+                p_annotations.append(pa)
+
+            if output_decl:
+                p_sig.ret_annotation.is_array = True
```
chipkent, line 527: I think this results from `( )` indicating a scalar output.  This is not obvious from the code and should have a comment.

py/server/deephaven/table.py
```diff
@@ -363,108 +368,306 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+            input_output_decl = fn.signature
+            input_decl, output_decl = input_output_decl.split("->")
+            input_decl = input_decl.strip("( )").split(",")
+            output_decl = output_decl.strip("( )")
+
+            for p, d in zip(params, input_decl):
+                pa = ParsedAnnotation()
+                if d:
+                    pa.encoded_types.add("[" + p)
+                    pa.is_array = True
```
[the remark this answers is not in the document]
chipkent: I think this results from ( ) indicating a scalar output. This is not obvious from the code and should have a comment.

py/server/deephaven/table.py
```diff
@@ -363,108 +368,306 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+            return _j_array_to_numpy_array(dtype, arg, conv_null=True, no_promotion=True)
+        # if the annotation is missing, or it is a generic object type, return the arg
+        elif "O" in param.encoded_types:
+            return arg
+        else:
+            raise TypeError(f"Argument {arg} is not compatible with annotation {param.encoded_types}")
+    else:  # if the arg is not Java array
+        # find the numpy dtype for the annotation
+        # if found, convert the arg to that type, take care of nulls (if null and optional, return None, else raise)
+        # if not found, if empyt annotation, return arg, else return
```
chipkent, line 594:
```suggestion
        # if not found, if empty annotation, return arg, else return
```

py/server/deephaven/table.py
```diff
@@ -363,108 +368,306 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+        # if dh_null := _PRIMITIVE_DTYPE_NULL_MAP.get(dtype):
```
chipkent, line 595: needed?

py/server/deephaven/table.py
```diff
@@ -363,108 +368,306 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+        specific_types = param.encoded_types - {"N", "O"}
```
chipkent, line 596: Not sure what the `N` type is.

py/server/deephaven/table.py
```diff
@@ -363,108 +368,306 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+            for t in specific_types:
+                if t.startswith("["):
+                    if isinstance(arg, np.ndarray):
+                        return arg
+                    continue
+
+                dtype = dtypes.from_np_dtype(np.dtype(t))
+                dh_null = _PRIMITIVE_DTYPE_NULL_MAP.get(dtype)
+
+                if param.int_char and isinstance(arg, int):
```
chipkent, line 607: How are `np.int32`, `np.int64`, etc. handled?

py/server/deephaven/table.py
```diff
@@ -363,108 +368,306 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+                    if arg == dh_null:
+                        if param.is_none_legal:
+                            return None
+                        else:
+                            raise DHError(f"Argument {arg} is not compatible with annotation {param.encoded_types}")
+                    else:
+                        return arg
+                elif param.floating_char and isinstance(arg, float):
```
chipkent, line 615: How are `np.float32` and `np.float64` handled?

py/server/deephaven/table.py
```diff
@@ -363,108 +368,306 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+            else:  # didn't return from inside the for loop
+                if "O" in param.encoded_types:
+                    return arg
+                else:
+                    raise TypeError(f"Argument {arg} is not compatible with annotation {param.orig_types}")
+        else:  # if no annotation or generic object, return arg
+            return arg
+
+
+def _convert_args(p_sig: ParsedSignature, args: Tuple[Any, ...]) -> List[Any, ...]:
```
chipkent, line 640: Does `List[Any,...]` need `...`?

py/server/deephaven/table.py
```diff
@@ -363,108 +368,306 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+    converted_args = []
+    for arg, param in zip(args, p_sig.params):
```
chipkent, line 642: does there need to be an assertion on zip arg lengths?

py/server/deephaven/table.py
```diff
@@ -363,108 +368,306 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+        converted_args.append(_convert_arg(param, arg))
```
chipkent, line 643: Could be simplified as a list comprehension.

py/server/deephaven/table.py
```diff
@@ -363,108 +368,306 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
-                    return_array = True
+    p_sig = _parse_signature(fn)
+    return_array = p_sig.ret_annotation.is_array
+    ret_dtype = dtypes.from_np_dtype(np.dtype(list(p_sig.ret_annotation.encoded_types)[0][-1]))
 
     @wraps(fn)
     def wrapper(*args, **kwargs):
-        ret = fn(*args, **kwargs)
+        converted_args = _convert_args(p_sig, args)
+        ret = fn(*converted_args, **kwargs)
```
chipkent, line 670: I would add a comment about why `kwargs` are not converted.  Potentially create a ticket to handle conversion after we support `kwargs` in the query lang.

py/server/deephaven/table.py
```diff
@@ -485,6 +688,13 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def _encode_signature(fn: Callable) -> str:
+    """Encode the signature of a function into a string for backwards compatibility with the vectorization implementation
+    in the server."""
+    p_sig = _parse_signature(fn)
+    return re.sub("[\[N,]", "", p_sig.encoded)
```
chipkent, line 695: Backwards compatibility seems weird for an internal feature.  Should things be done differently?

py/server/deephaven/table.py
```diff
@@ -500,14 +710,17 @@ def dh_vectorize(fn):
     The current vectorized function signature includes (1) the size of the input arrays, (2) the output array,
     and (3) the input arrays.
     """
-    signature = _encode_signature(fn)
-    ret_dtype = _udf_return_dtype(fn)
+    # sig_str = _encode_signature(fn)
+    # ret_dtype = _udf_return_dtype(fn, signature=fn_signature)
```
chipkent, line 714: remove?

py/server/deephaven/table.py
```diff
@@ -500,14 +710,17 @@ def dh_vectorize(fn):
...
+    p_sig = _parse_signature(fn)
+    sig_str = re.sub("[\[N,]", "", p_sig.encoded)
```
chipkent, line 716: DRY `_encode_signature`?

py/server/deephaven/dtypes.py
```diff
@@ -354,6 +377,8 @@ def _scalar(x: Any, dtype: DType) -> Any:
             elif x.dtype.char == 'M':
                 from deephaven.time import to_j_instant
                 return to_j_instant(x)
+            elif x.dtype.char == 'H':  # np.uint16
+                return jpy.get_type("java.lang.Character")(int(x))
```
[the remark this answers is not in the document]
jmao-denver: Yes, there is overhead with using boxed type but sadly we can't avoid it. Frankly this is scalar value and I value the ability to round-trip Java char type more here.

py/server/deephaven/pandas.py
```diff
@@ -51,29 +39,15 @@ def _column_to_series(table: Table, col_def: Column, conv_null: bool) -> pd.Seri
...
-        np_array = column_to_numpy_array(col_def, data_col.getDirect())
-        if conv_null and (null_pair := _DTYPE_NULL_MAPPING.get(col_def.data_type)) is not None:
-            nv = null_pair[0]
-            pd_ex_dtype = null_pair[1]
-            s = pd.Series(data=np_array, dtype=pd_ex_dtype(), copy=False)
-            s.mask(s == nv, inplace=True)
-        else:
-            s = pd.Series(data=np_array, copy=False)
-        return s
+        j_array = data_col.getDirect()
```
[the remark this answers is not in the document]
jmao-denver: `to_pandas` works only on a static table or the snapshot of a ticking table.

py/server/tests/test_udf_numpy_args.py
```diff
@@ -0,0 +1,184 @@
...
+}
+
+_J_TYPE_J_ARRAY_TYPE_MAP = {
+    "double": double_array,
+    "float": float32_array,
+    "int": int32_array,
+    "long": long_array,
+    "short": int16_array,
+    "byte": int8_array,
+}
```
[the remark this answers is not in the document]
jmao-denver: They are for dynamically generating Py UDFs with proper type hints for testing purpose. I don't see how they can be useful for the users.

py/server/tests/test_udf_numpy_args.py
```diff
@@ -0,0 +1,184 @@
...
+            with self.subTest(j_dtype):
+                tbl = empty_table(100).update([col1_formula, col2_formula]).group_by("Col1")
+
+                func_str = f"""
+def test_udf(col1, col2) -> bool:
+    j_array_type = _J_TYPE_J_ARRAY_TYPE_MAP[{j_dtype!r}].j_type
+    return ((isinstance(col1, int) or isinstance(col1, float)) and isinstance(col2, j_array_type)
+        and np.any(np.array(col2) == {null_name}))
+                        """
+                exec(f"from deephaven.constants import {null_name}", globals())
```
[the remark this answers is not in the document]
jmao-denver: yes, it is made standalone for symmetrical reason.

py/server/deephaven/jcompat.py
```diff
@@ -181,11 +198,119 @@ def to_sequence(v: Union[T, Sequence[T]] = None, wrapped: bool = False) -> Seque
...
+    """ Produces a numpy array from the DType and given Java array."""
+    if dtype.is_primitive:
+        np_array = np.frombuffer(j_array, dtype.np_type)
+    elif dtype == dtypes.Instant:
+        longs = _JPrimitiveArrayConversionUtility.translateArrayInstantToLong(j_array)
+        np_long_array = np.frombuffer(longs, np.int64)
+        np_array = np_long_array.view(dtype.np_type)
+    elif dtype == dtypes.bool_:
+        bytes_ = _JPrimitiveArrayConversionUtility.translateArrayBooleanToByte(j_array)
+        np_array = np.frombuffer(bytes_, dtype.np_type)
```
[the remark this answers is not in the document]
jmao-denver: It doesn't, the underlying byte array has NULL_BYTE

py/server/deephaven/jcompat.py
```diff
@@ -181,11 +198,119 @@ def to_sequence(v: Union[T, Sequence[T]] = None, wrapped: bool = False) -> Seque
...
+                if dtype is dtypes.bool_:  # promote boolean to float64
+                    np_array = np.frombuffer(np_array, np.byte)
+                if any(np_array[np_array == dh_null]):
+                    if no_promotion:
+                        raise DHError(f"Java array contains Deephaven nulls for dtype {dtype}")
+                    np_array = np_array.astype(np.float64)
+                    np_array[np_array == dh_null] = np.nan
+                else:
+                    if dtype is dtypes.bool_:  # promote boolean to float64
+                        np_array = np.frombuffer(np_array, np.bool_)
```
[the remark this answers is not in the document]
jmao-denver: Maybe, but the conv_null logic is split into two parts, which creates its own readability issue.

py/server/deephaven/table.py
```diff
@@ -363,108 +368,306 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+
+    tc = _np_dtype_char(t)
+    tc = tc if tc in _SUPPORTED_NP_TYPE_CODES else "O"
+
+    if component_type:
+        tc = "[" + tc
+    return tc
+
+
+def _parse_param_annotation(annotation: Any) -> ParsedAnnotation:
```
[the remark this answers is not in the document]
jmao-denver: No, but it is hard to pin down the exact types

py/server/deephaven/table.py
```diff
@@ -363,108 +368,306 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+        p_annotation.is_none_legal = True
+    if tc in {"b", "h", "i", "l"}:
+        if p_annotation.int_char and p_annotation.int_char != tc:
+            raise DHError(message=f"ambiguity detected: multiple integer types in annotation: {annotation}")
+        p_annotation.int_char = tc
+    if tc in {"f", "d"}:
+        if p_annotation.floating_char and p_annotation.floating_char != tc:
+            raise DHError(message=f"ambiguity detected: multiple floating types in annotation: {annotation}")
+        p_annotation.floating_char = tc
+    p_annotation.encoded_types.add(tc)
```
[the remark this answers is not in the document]
jmao-denver: is_array may be misnomer, changed to has_array

py/server/tests/test_udf_numpy_args.py
```diff
@@ -0,0 +1,184 @@
...
+                        tbl.update("Col3 = test_udf(Col1, Col2)")
+
+    def test_j_scalar_to_py_no_null(self):
+        col1_formula = "Col1 = i % 10"
+        for j_dtype, null_name in _J_TYPE_NULL_MAP.items():
+            col2_formula = f"Col2 = ({j_dtype})i"
+            with self.subTest(j_dtype):
+                np_type = _J_TYPE_NP_DTYPE_MAP[j_dtype]
+                func = f"""
+def test_udf(col: {np_type}) -> bool:
```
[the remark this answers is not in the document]
jmao-denver: Not sure what you mean. It is still a UDF unless you think `test_udf_scalar` is more readable?

py/server/deephaven/table.py
```diff
@@ -363,108 +368,306 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+    return p_annotation
+
+
+def _parse_type_no_nested(annotation, p_annotation, t):
+    """ Parse a type annotation in a function's signature, without handling nested types (Union) """
+    p_annotation.orig_types.add(t)
+    tc = _encode_param_type(t)
+    if "[" in tc:
+        p_annotation.is_array = True
+    if tc in {"N", "O", "?", "U", "M"}:
```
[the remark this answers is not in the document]
chipkent: Sounds like that needs some comments, since I did not pick up on that.

py/server/tests/test_udf_numpy_args.py
```diff
@@ -0,0 +1,384 @@
...
+
+        t = empty_table(10).update(["X1 = f3(i)"])
+        self.assertEqual(t.columns[0].data_type, dtypes.bool_)
+
+        def f4(p1: Union[np.int16, np.float64], p2=None) -> bool:
+            return bool(p1)
+
+        t = empty_table(10).update(["X1 = f4((double)i)"])
+        self.assertEqual(t.columns[0].data_type, dtypes.bool_)
+        with self.assertRaises(DHError) as cm:
```
chipkent, line 240: It would be good if the error assumptions checked the messages to make sure we are failing the way we expect.

py/server/tests/test_udf_numpy_args.py
```diff
@@ -0,0 +1,384 @@
...
+            t = t.update(["X1 = f5(X, null)"])
+
+        def f51(col1, col2: Optional[np.ndarray[np.int32]]) -> bool:
+            return np.nanmean(col2) == np.mean(col2)
+
+        t = empty_table(10).update(["X = i % 3", "Y = i"]).group_by("X")
+        t = t.update(["X1 = f51(X, Y)"])
+        with self.assertRaises(DHError) as cm:
+            t = t.update(["X1 = f51(X, null)"])
```
chipkent, line 270: This looks like it should work or the exception is in the mean calculation.  Not checking the error message hides the cause.

py/server/tests/test_udf_numpy_args.py
```diff
@@ -0,0 +1,384 @@
...
+            self.assertEqual(5, t2.to_string().count("false"))
+
+        with self.subTest("datetime"):
+            def f2(p1: np.datetime64, p2=None) -> bool:
+                return p1 is None
+
+            t = empty_table(10).update(["X = i % 3", "Y = i % 2 == 0? now() : null"])
+            t1 = t.update(["X1 = f2(Y)"])
+            self.assertEqual(t1.columns[2].data_type, dtypes.bool_)
+            self.assertEqual(5, t1.to_string().count("true"))
```
chipkent, line 356: This seems to violate the annotation. Are null inputs allowed for object types if they are not optional?

py/server/tests/test_udf_numpy_args.py
```diff
@@ -0,0 +1,384 @@
...
+            self.assertEqual(5, t2.to_string().count("false"))
+
+        with self.subTest("boolean"):
+            def f3(p1: np.bool_, p2=None) -> bool:
+                return p1 is None
+
+            t = empty_table(10).update(["X = i % 3", "Y = i % 2 == 0? true : null"])
+            t1 = t.update(["X1 = f3(Y)"])
+            self.assertEqual(t1.columns[2].data_type, dtypes.bool_)
+            self.assertEqual(5, t1.to_string("X1").count("false"))
```
chipkent, line 370: This seems to violate the annotation. Are null inputs allowed for object types if they are not optional?

py/server/deephaven/dtypes.py
```diff
@@ -382,20 +411,32 @@ def _component_np_dtype_char(t: type) -> Optional[str]:
...
     # Py3.9+, np.ndarray as a generic alias is only supported in Python 3.9+, also npt.NDArray is still available but a
     # specific alias (e.g. npt.NDArray[np.int64]) now is an instance of typing.GenericAlias.
     # when npt.NDArray is used, the 1st argument is typing.Any, the 2nd argument is another generic alias of which
     # the 1st argument is the component type
     # when np.ndarray is used, the 1st argument is the component type
-    if not component_type and sys.version_info.minor > 8:
+    if not component_type and sys.version_info.major == 3 and sys.version_info.minor > 8:
         import types
         if isinstance(t, types.GenericAlias) and (issubclass(t.__origin__, Sequence) or t.__origin__ == np.ndarray):
             nargs = len(t.__args__)
```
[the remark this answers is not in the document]
jmao-denver: For return type inference, if we can't deterministically arrive at one, we default to object, failing it is too breaking IMO.

See the docstring of 
```
def _parse_return_annotation(annotation: Any) -> _ParsedAnnotation:
    """ Parse an annotation in a function's signature

    The return annotation is treated differently from the parameter annotations. We don't apply the same check and are
    only interested in getting the array-like type right. Any nonsensical annotation will be treated as object type.
    This definitely can be improved in the future.
    """
```

py/server/deephaven/table.py
```diff
@@ -363,108 +368,319 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+        p_sig = _ParsedSignature(fn)
+
+        # for now, we only support one signature for a numba function because the query engine is not ready to handle
+        # multiple signatures for vectorization https://github.com/deephaven/deephaven-core/issues/4762
+        sig = sigs[0]
+        params, rt_char = sig.split("->")
+
+        p_sig.params = []
+        p_sig.ret_annotation = _ParsedAnnotation()
+        p_sig.ret_annotation.encoded_types.add(rt_char)
```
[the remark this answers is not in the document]
jmao-denver: Numba only allows one return type

py/server/deephaven/table.py
```diff
@@ -363,108 +368,319 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+                            return np.nan if "N" not in param.encoded_types else None
+                        else:
+                            return np.dtype(param.floating_char).type(arg)
+                elif t == "?" and isinstance(arg, bool):
+                    return arg
+                elif t == "M":
+                    try:
+                        return to_np_datetime64(arg)
+                    except Exception as e:
+                        if len(specific_types) == 1:
```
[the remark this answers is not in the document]
jmao-denver: The parameter type could be a Union.

py/server/tests/test_udf_numpy_args.py
jmao-denver: Yes, that's my understanding.

py/server/tests/test_udf_numpy_args.py
```diff
@@ -0,0 +1,384 @@
...
+
+    def test_str_bool_datetime_scalar(self):
+        with self.subTest("str"):
+            def f1(p1: str, p2=None) -> bool:
+                return p1 is None
+
+            t = empty_table(10).update(["X = i % 3", "Y = i % 2 == 0? `deephaven`: null"])
+            t1 = t.update(["X1 = f1(Y)"])
+            self.assertEqual(t1.columns[2].data_type, dtypes.bool_)
+            self.assertEqual(5, t1.to_string().count("true"))
```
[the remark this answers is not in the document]
jmao-denver: Yes, for scalars, that's my understanding.

py/server/tests/test_udf_numpy_args.py
```diff
@@ -0,0 +1,384 @@
...
+        with self.assertRaises(DHError) as cm:
+            t = empty_table(10).update(["X1 = f1(i)"])
+
+        def f11(p1: Union[float, np.float32]) -> bool:
+            return bool(p1)
+
+        with self.assertRaises(DHError) as cm:
+            t = empty_table(10).update(["X1 = f11(i)"])
+
+        def f2(p1: Union[np.int16, np.float64]) -> Union[Optional[bool]]:
```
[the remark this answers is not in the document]
jmao-denver: yes

py/server/tests/test_udf_numpy_args.py
```diff
@@ -0,0 +1,384 @@
...
+            return bool(p1)
+
+        t = empty_table(10).update(["X1 = f2(i)"])
+        self.assertEqual(t.columns[0].data_type, dtypes.bool_)
+        self.assertEqual(9, t.to_string().count("true"))
+
+        def f21(p1: Union[np.int16, np.float64]) -> Union[Optional[bool], int]:
```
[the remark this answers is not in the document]
jmao-denver: it is equivalent, but the inference comes to 'object' because it has two distinct types.

py/server/deephaven/table.py
```diff
@@ -363,108 +368,319 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
-            component_type = _component_np_dtype_char(return_annotation)
-            if component_type:
-                ret_dtype = dtypes.from_np_dtype(np.dtype(component_type))
-                if ret_dtype in _BUILDABLE_ARRAY_DTYPE_MAP:
-                    return_array = True
+    p_sig = _parse_signature(fn)
+    # build a signature string for vectorization by removing NoneType, array char '[', and comma from the encoded types
+    # since vectorization only supports UDFs with a single signature and enforces an exact match, any non-compliant
+    # signature (e.g. Union with more than 1 non-NoneTYpe) will be rejected by the vectorizer.
+    sig_str_vectorization = re.sub("[\[N,]", "", p_sig.encoded)
```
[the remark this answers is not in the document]
jmao-denver: This isn't any more brittle than what we had before the changes in this PR. The annotation parsing work in this PR actually lays the foundation for broadening the vectorization support in the future for what you mentioned b/c they capture the complete picture of a UDF's signature and can share that with the QLP once the improvement work starts.

py/server/deephaven/table.py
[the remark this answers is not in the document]
chipkent: I would add a comment, since it is not clear.

chipkent (commented):
I think the changes are good.  Just need to review open items to see if any require attention or if they are all done or WNI.

chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3514
Reference: https://github.com/deephaven/deephaven.io/issues/3515

PR #4507, opened 2023-09-18
DH to work with byte arrays

### LINKED ISSUE #4355: Support in Java client for publishing columns of type byte[]
As a user of the Java client, I want to be able to publish variable-length byte arrays into a Deephaven table column. Currently attempting to publish
```
ColumnHeader<byte[]> binary = ColumnHeader.of("B", byte[].class)
```
throws an exception from [here](https://github.com/deephaven/deephaven-core/blob/50bdba43c0290d75aad3dcfd70f4f7761ca8ba17/java-client/flight/src/main/java/io/deephaven/client/impl/FieldVectorAdapter.java#L159)

However, since Arrow supports `VarBinaryVector`, it might be possible to implement this in the same way that publishing a String-valued column is currently published.

ilyaberdnikov:
Fixes #4355.

This is currently a very rough hack, but it does work locally. Not sure what it breaks.

### DIFF
--- a/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/ChunkInputStreamGenerator.java
+++ b/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/ChunkInputStreamGenerator.java
@@ -30,6 +30,7 @@ import java.math.BigInteger;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+import java.util.Arrays;
 import java.util.Iterator;
 
 public interface ChunkInputStreamGenerator extends SafeCloseable {
@@ -204,8 +205,18 @@ public interface ChunkInputStreamGenerator extends SafeCloseable {
                         Double.BYTES, options,fieldNodeIter, bufferInfoIter, is, outChunk, outOffset, totalRows);
             case Object:
                 if (type.isArray()) {
-                   return VarListChunkInputStreamGenerator.extractChunkFromInputStream(
-                           options, type, fieldNodeIter, bufferInfoIter, is, outChunk, outOffset, totalRows);
+                    if(componentType == byte.class) {
+                        return VarBinaryChunkInputStreamGenerator.extractChunkFromInputStream(
+                              is,
+                              fieldNodeIter,
+                              bufferInfoIter,
+                              (buf, off, len) -> Arrays.copyOfRange(buf, off, off + len),
+                              outChunk, outOffset, totalRows
+                        );
+                    } else {
+                        return VarListChunkInputStreamGenerator.extractChunkFromInputStream(
+                              options, type, fieldNodeIter, bufferInfoIter, is, outChunk, outOffset, totalRows);
+                    }
                 }
                 if (Vector.class.isAssignableFrom(type)) {
                     //noinspection unchecked
--- a/java-client/flight-examples/src/main/java/io/deephaven/client/examples/AddToInputTable.java
+++ b/java-client/flight-examples/src/main/java/io/deephaven/client/examples/AddToInputTable.java
@@ -27,11 +27,21 @@ class AddToInputTable extends FlightExampleBase {
 
     @Override
     protected void execute(FlightSession flight) throws Exception {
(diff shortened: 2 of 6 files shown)


extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/ChunkInputStreamGenerator.java
```diff
@@ -204,8 +205,18 @@ static WritableChunk<Values> extractChunkFromInputStream(
                         Double.BYTES, options,fieldNodeIter, bufferInfoIter, is, outChunk, outOffset, totalRows);
             case Object:
                 if (type.isArray()) {
-                   return VarListChunkInputStreamGenerator.extractChunkFromInputStream(
-                           options, type, fieldNodeIter, bufferInfoIter, is, outChunk, outOffset, totalRows);
+                    if(componentType == byte.class) {
```
nbauernfeind, line 208: The style here is slightly off. If you run `./gradlew spotlessApply` at the top level, then everything non-comforming will be formatted automatically. (Otherwise, once CI runs, it will fail here and in a few other locations.)

java-client/flight-examples/src/main/java/io/deephaven/client/examples/AddToInputTable.java
```diff
@@ -44,9 +54,10 @@ protected void execute(FlightSession flight) throws Exception {
 
             while (true) {
                 // Add a new row, at least once every second
-                final NewTable newRow = header.row(Instant.now()).newTable();
+                final NewTable newRow = header.row(true, (byte) 42, 'a', (short) 32_000, 1234567, 1234567890123L, 3.14f, 3.14d, "Hello, World",
+                      Instant.now(), "abc".getBytes()).newTable();
                 flight.addToInputTable(timestampHandle, newRow, bufferAllocator).get(5, TimeUnit.SECONDS);
-                Thread.sleep(ThreadLocalRandom.current().nextLong(1000));
+                Thread.sleep(ThreadLocalRandom.current().nextLong(100000));
```
nbauernfeind, line 60: Let's set this value back to 1s. Alternatively, you could make this a flag.

Here is an example flag that is a similar type:
```java
    @Option(names = {"--input-table-size"}, description = "The input table size, default ${DEFAULT-VALUE}",
            defaultValue = "1000")
    int inputTableSize;
```
(Source for this flag is here: https://github.com/deephaven/deephaven-core/blob/main/java-client/flight-examples/src/main/java/io/deephaven/client/examples/AggInputBase.java#L30)

If you make as a flag; change the comment on line 56 to reference the flag rather than 'once every second'.

nbauernfeind (commented):
Please sign the CLA (see the automated message above). 

Once you do and address my comments, I'll kick of a "nightly" build which runs a suite of tests (many of which go through the ChunkInputStreamGenerator code path).

Thank you for the PR; once merged I'll probably tack on support for the remaining primitive arrays. (No reason the java client shouldn't be supporting these.)

ilyaberdnikov:

java-client/flight-examples/src/main/java/io/deephaven/client/examples/AddToInputTable.java
ilyaberdnikov: reverted to 1s

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/ChunkInputStreamGenerator.java
ilyaberdnikov: should now be spotless

ilyaberdnikov:
recheck

nbauernfeind:
Your PR looks good to me so far. I've just kicked off that suite of tests. When they pass (plus or minus a known random failure) then I'll get this merged. 

Thank you so much for your contribution!

nbauernfeind (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3202
Reference: https://github.com/deephaven/deephaven.io/issues/3201

PR #4511, opened 2023-09-18
Fix Python tests for FunctionGeneratedTable

rbasralian:


py/server/tests/test_function_generated_table.py
```diff
@@ -44,20 +44,18 @@ def test_generated_table_1trigger(self):
             "MyStr": dht.string
         }
         append_only_input_table = input_table(col_defs=col_defs)
+        input_table_lastby = append_only_input_table.last_by()
 
         def table_generator_function():
-            print("Running table_generator_function() at time: " + str(datetime.now()))
-            return append_only_input_table.last_by().update('ResultStr = MyStr')
+            with liveness_scope():
+                return input_table_lastby.update('ResultStr = MyStr')
```
rcaudy, line 51: What are you trying to accomplish here? That liveness scope is going to immediately kill your result. Maybe you want to return a `.snapshot()`?

rcaudy: You convinced me the existing pattern is better, since the downstream code will just do an effective snapshot.

rcaudy (approved):

jmao-denver (approved):
LGTM

PR #4513, opened 2023-09-18
Fix for failure in Head_By and AggAllBy after Format_Column

### LINKED ISSUE #4501: Calling Head_By After Format_Column Fails with Exception
Chaining operations after _format_columns_ works for some operations but not _head_by_.  We need to decide if we are going to support chaining on _format_columns_ or only allow it on a leaf table.  If not allowed, errors should be explicit.  If allowed, column formatting should be carried along through subsequent tables. (Easier said than done.)

Run the following code snippet (Tested in Deephaven 0.28.0+)
````
from deephaven.column import string_col, double_col

result = new_table([
    double_col("Doubles", [3.1, 5.45, -1.0]),
    string_col("Strings", ["Creating", "New", "Tables"])
])

result = result.format_columns(["Doubles=Decimal(`##0.00%`)"]).head_by(1)
````
You will get an exception like the following:
````
Value: table head_by operation failed. : RuntimeError: io.deephaven.engine.table.impl.NoSuchColumnException: Unknown column names [Doubles__TABLE_NUMBER_FORMAT], available column names are [Doubles,Strings]
  File "/opt/deephaven/venv/lib/python3.10/site-packages/deephaven/table.py", line 1487, in head_by
    return Table(j_table=self.j_table.headBy(num_rows, *by))
RuntimeError: io.deephaven.engine.table.impl.NoSuchColumnException: Unknown column names [Doubles__TABLE_NUMBER_FORMAT], available column names are [Doubles,Strings]
````
If you use head_by(1, ['Strings']) or head_by(1, ['Doubles','Strings']), you get essentially the same error

### PR DESCRIPTION
**Pull request by malhotrashivam**
Closes #4501

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
@@ -58,7 +58,6 @@ import io.deephaven.engine.table.iterators.*;
 import io.deephaven.engine.updategraph.DynamicNode;
 import io.deephaven.engine.util.*;
 import io.deephaven.engine.util.systemicmarking.SystemicObject;
-import io.deephaven.qst.table.AggregateAllTable;
 import io.deephaven.util.annotations.InternalUseOnly;
 import io.deephaven.util.annotations.ReferentialIntegrity;
 import io.deephaven.vector.Vector;
@@ -732,7 +731,7 @@ public class QueryTable extends BaseTable<QueryTable> {
             }
             final List<ColumnName> groupByList = Arrays.asList(groupByColumns);
             final List<ColumnName> tableColumns = definition.getTypedColumnNames();
-            final Optional<Aggregation> agg = AggregateAllTable.singleAggregation(spec, groupByList, tableColumns);
+            final Optional<Aggregation> agg = singleAggregation(spec, groupByList, tableColumns);
             if (agg.isEmpty()) {
                 throw new IllegalArgumentException(
                         "aggAllBy has no columns to aggregate: spec=" + spec + ", groupByColumns="
@@ -747,6 +746,29 @@ public class QueryTable extends BaseTable<QueryTable> {
                 return result;
             });
         }
+    }
+
+    /**
+     * Computes the single-aggregation from the agg-all implied by the {@code spec} and {@code groupByColumns} by
+     * removing the {@code groupByColumns} and any extra columns implied by the {@code spec}.
+     *
+     * @param spec the spec
+     * @param groupByColumns the group by columns
+     * @param tableColumns the table columns
+     * @return the aggregation, if non-empty
+     */
+    static Optional<Aggregation> singleAggregation(
+            AggSpec spec, Collection<? extends ColumnName> groupByColumns,
+            Collection<? extends ColumnName> tableColumns) {
+        Set<ColumnName> exclusions = AggregateAllExclusions.of(spec, groupByColumns, tableColumns);
+        List<ColumnName> columnsToAgg = new ArrayList<>(tableColumns.size());
+        for (ColumnName column : tableColumns) {
+            if (exclusions.contains(column)) {
+                continue;
+            }
+            columnsToAgg.add(column);
+        }
+        return columnsToAgg.isEmpty() ? Optional.empty() : Optional.of(spec.aggregation(columnsToAgg));
     }
 
     @Override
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

File: engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java [hunk 1]
```diff
@@ -835,31 +834,37 @@ private Table headOrTailBy(long nRows, boolean head, String... groupByColumns) {
             String colName = colNames.get(i);
             if (!groupByColsSet.contains(colName)) {
                 final Class<?> dataType = getDefinition().getColumn(colName).getDataType();
-                casting[j] = colName + " = " + getCastFormula(dataType) + colName;
+                casting[j] = String.format("%s = %s%s", colName, getCastFormula(dataType), colName);
```
rcaudy, line 837: Formats are nice, but they are definitely much more expensive than concatenation. This loop isn't so tight, though, so I suppose it's up to you if you want to keep this change or revert it.

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -835,31 +834,37 @@ private Table headOrTailBy(long nRows, boolean head, String... groupByColumns) {
...
                     updates[j++] =
                             // Get the first nRows rows:
                             // colName = isNull(colName) ? null
                             // : colName.size() > nRows ? colName.subVector(0, nRows)
                             // : colName
-                            colName + '=' + "isNull(" + colName + ") ? null" +
-                                    ':' + colName + ".size() > " + nRows + " ? " + colName + ".subVector(0, " + nRows
-                                    + ')' +
-                                    ':' + colName;
+                            String.format(
```
rcaudy, line 844: Wow, this is much more readable.

rcaudy (commented):
I think this is good, but I want @devinrsmith to give me an opinion on the two file moves to QST, and whether he has an alternate module he prefers.

qst/src/main/java/io/deephaven/qst/table/AggregateAllTable.java
```diff
@@ -35,7 +35,7 @@ public static Builder builder() {
     public static Optional<Aggregation> singleAggregation(
             AggSpec spec, Collection<? extends ColumnName> groupByColumns,
             Collection<? extends ColumnName> tableColumns) {
-        Set<ColumnName> exclusions = AggregateAllExclusions.of(spec, groupByColumns);
+        Set<ColumnName> exclusions = AggregateAllExclusions.of(spec, groupByColumns, tableColumns);
```
devinrsmith, line 38: I don't think we should move ColumnFormatting into qst; I'm thinking this should be the responsibility of the calling layer (engine). What about expanding this function to take in an additional parameter `Collection<? extends ColumnName> specialExclusions`?

devinrsmith (changes requested):

qst/src/main/java/io/deephaven/qst/table/AggregateAllTable.java
rcaudy: Then we won't get it right if we try to call `singleAggregation` from the QST layer.

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
malhotrashivam: How do we usually decide between the two: performance vs readability?
Is that why we always keep string.format in exceptions because performance is not a concern there?

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -732,24 +731,46 @@ public Table aggAllBy(AggSpec spec, ColumnName... groupByColumns) {
...
+    /**
+     * Computes the single-aggregation from the agg-all implied by the {@code spec} and {@code groupByColumns} by
+     * removing the {@code groupByColumns} and any extra columns implied by the {@code spec}.
+     *
+     * @param spec the spec
+     * @param groupByColumns the group by columns
+     * @param tableColumns the table columns
+     * @return the aggregation, if non-empty
+     */
+    static Optional<Aggregation> singleAggregation(
```
rcaudy, line 760:
```suggestion
    @VisibleForTesting
    static Optional<Aggregation> singleAggregation(
```

rcaudy (approved):

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
malhotrashivam: Will add this in my next PR to avoid another review cycle.

PR #4525, opened 2023-09-19
Modify layout of grouping files

### LINKED ISSUE #4283: Grouping files for parquet should not be written in the same directory as the main file
Currently, if we write a deephaven table with grouping columns to a parquet files, it creates separate grouping parquet files in the same directory as the main parquet file and the path to this grouping file is added as metadata of the main parquet file.

For example, if a table `A` has grouping column `g` and we write A to parquet file `some_path/tableA.parquet`, it will create additional parquet file at `some_path/tableA_g_grouping.parquet` containing the grouping information.

Therefore, if a user now reads all the parquet files for the directory `some_path`, they can mistakenly try to read the grouping file as regular data. We should fix this and come up with a new nested subdirectory structure for storing the grouping files.

malhotrashivam:
Closes #4283 

[Proposal document](https://docs.google.com/document/d/13a5kOVjUMXjUy8ThXLh9Cu9lnaxoAESCkINsCqkMXPI/edit#heading=h.5x0d5h95i329) for new layout.


extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -237,9 +237,18 @@ private static String minusParquetSuffix(@NotNull final String s) {
         return s;
     }
 
-    public static Function<String, String> defaultGroupingFileName(@NotNull final String path) {
-        final String prefix = minusParquetSuffix(path);
-        return columnName -> prefix + "_" + columnName + "_grouping.parquet";
+    /**
+     * Generates grouping file path relative to the table destination file path.
```
lbooker42, line 241:
```suggestion
     * Generates a grouping file path relative to the table destination file path.
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -237,9 +237,18 @@ private static String minusParquetSuffix(@NotNull final String s) {
...
+     *
+     * @param tableDest Destination path for the main table containing these grouping columns
+     * @param columnName Name of the grouping column
+     *
+     * @return The relative grouping file path. For example, for table {@code "A"} with destination
+     *         {@code "dir/A.parquet"} and grouping column {@code "g"}, the method will return
```
lbooker42, line 247: In the example `{@code "dir/A.parquet"}` , including `dir` is a bit confusing since it isn't actually used.  Or is it?  I think you can just leave that out.

```suggestion
     *         {@code "A.parquet"} and grouping column {@code "g"}, the method will return
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -346,7 +355,7 @@ private static File prepareDestinationFileLocation(@NotNull File destination) {
     }
 
     /**
-     * Helper function for building grouping column info for writing and deleting any backup grouping column files
+     * Helper function for building grouping column info for writing. Also, deletes any backup grouping column files.
```
lbooker42, line 358: Does this delete `any` backup files or just similarly named files?

```suggestion
     * Helper function for building grouping column info for writing. Also deletes any conflicting backup grouping column files.
```

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -665,14 +698,84 @@ public void groupingColumnsBasicWriteTestsImpl(TestParquetTableWriter writer) {
...
+        // Add another table in a hidden file in parent directory
+        final Table anotherTable = TableTools.emptyTable(5).update("A=(int)i");
+        final File pqDotFile = new File(parentDir, ".hiddenData.parquet");
+        ParquetTools.writeTable(anotherTable, pqDotFile);
+        fromDisk = ParquetTools.readTable(parentDir);
+        TstUtils.assertTableEquals(fromDisk, tableToSave);
+    }
+
+    @Test
+    public void partitionedParuetWithHiddenFilesTest() throws IOException {
```
lbooker42, line 747:
```suggestion
    public void partitionedParquetWithHiddenFilesTest() throws IOException {
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetFileHelper.java
```diff
@@ -1,12 +1,15 @@
...
 final class ParquetFileHelper {
-
-    public static boolean fileNameMatches(Path path) {
-        return path.getFileName().toString().endsWith(PARQUET_FILE_EXTENSION);
+    /**
+     * Used as a filter to select relevant parquet files while reading all files in a directory
+     */
+    static boolean parquetFileFilter(Path path) {
+        return path.getFileName().toString().endsWith(ParquetTableWriter.PARQUET_FILE_EXTENSION) &&
+                !path.toFile().isHidden();
```
lbooker42, line 13: I'm not sure this is correct on all OS (i.e. Windows is stupid).  We are treating the dot prefix as our marker for the folder names, should we just test for that instead of relying on OS-dependent file-hidden tests?

Or is this something totally different?  I may be missing the point.

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -665,14 +698,84 @@ public void groupingColumnsBasicWriteTestsImpl(TestParquetTableWriter writer) {
...
+
+        // Add an empty dot file in the parent directory
+        final File dotFile = new File(parentDir, ".hiddenData");
+        assertTrue(dotFile.createNewFile());
+        fromDisk = ParquetTools.readTable(parentDir);
+        TstUtils.assertTableEquals(fromDisk, tableToSave);
+
+        // Add another table in a hidden file in parent directory
+        final Table anotherTable = TableTools.emptyTable(5).update("A=(int)i");
+        final File pqDotFile = new File(parentDir, ".hiddenData.parquet");
```
lbooker42, line 740: This will fail on OS where `.` prefixes aren't honored as hidden (i.e. Windows is stupid).

lbooker42 (commented):
Only one actual comment, others are cosmetic.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
malhotrashivam: Just similarly named files.

lbooker42 (approved):
LGTM

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetFileHelper.java
```diff
@@ -1,12 +1,15 @@
...
+import java.nio.file.Path;
 
 final class ParquetFileHelper {
-
-    public static boolean fileNameMatches(Path path) {
-        return path.getFileName().toString().endsWith(PARQUET_FILE_EXTENSION);
+    /**
+     * Used as a filter to select relevant parquet files while reading all files in a directory.
+     */
+    static boolean parquetFileFilter(Path path) {
```
rcaudy, line 11: I'm not sure the method rename was necessary or makes it more clear.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -237,9 +237,18 @@ private static String minusParquetSuffix(@NotNull final String s) {
         return s;
     }
 
-    public static Function<String, String> defaultGroupingFileName(@NotNull final String path) {
-        final String prefix = minusParquetSuffix(path);
-        return columnName -> prefix + "_" + columnName + "_grouping.parquet";
+    /**
+     * Generates a grouping file path relative to the table destination file path.
```
rcaudy, line 241: I thought we were re-branding "grouping" to "indexing" in this PR for new/changed code, to avoid creating a subsequent task?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -285,7 +294,7 @@ private static void installShadowFile(@NotNull final File destFile, @NotNull fin
     }
 
     /**
-     * Roll back any changes made in the {@link #installShadowFile} in a best effort manner
+     * Roll back any changes made in the {@link #installShadowFile} in the best effort manner
```
rcaudy, line 297: Previous version was better. Maybe this is better than either?
```suggestion
     * Roll back any changes made in the {@link #installShadowFile} in best-effort manner
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -570,7 +580,12 @@ private static Table readTableInternal(
                 return readPartitionedTableWithMetadata(source, instructions);
             }
             final Path firstEntryPath;
-            try (final DirectoryStream<Path> sourceStream = Files.newDirectoryStream(sourcePath)) {
+            // Ignore dot files while looking for the first entry
+            try (final DirectoryStream<Path> sourceStream =
+                    Files.newDirectoryStream(sourcePath, (path) -> {
+                        final String filename = path.getFileName().toString();
+                        return !filename.isEmpty() && filename.charAt(0) != '.';
```
rcaudy, line 587: We could further improve this. The dot filter is good. We should also filter out regular files that don't end in `.parquet`.

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -607,6 +612,41 @@ private interface TestParquetTableWriter {
     TestParquetTableWriter multiWriter = (table, destFile) -> ParquetTools.writeTables(new Table[] {table},
             table.getDefinition(), new File[] {destFile});
 
+    /**
```
rcaudy, line 615: Have you tested backwards compatibility?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/location/ParquetColumnLocation.java
```diff
@@ -153,22 +154,19 @@ public <METADATA_TYPE> METADATA_TYPE getMetadata(
...
-        final Function<String, String> defaultGroupingFilenameByColumnName =
-                ParquetTools.defaultGroupingFileName(tl().getParquetFile().getAbsolutePath());
+        final File parquetFile = tl().getParquetFile();
         try {
-            final GroupingColumnInfo groupingColumnInfo =
-                    tl().getGroupingColumns().get(parquetColumnName);
+            final GroupingColumnInfo groupingColumnInfo = tl().getGroupingColumns().get(parquetColumnName);
+            final String groupingFileRelativePath = groupingColumnInfo == null
+                    ? ParquetTools.getRelativeGroupingFilePath(parquetFile, parquetColumnName)
+                    : groupingColumnInfo.groupingTablePath();
```
rcaudy, line 162: Seems to me like we should fall back to the old default path if we don't find explicit info or the new default path.

rcaudy (commented):
This looks pretty good to me. A few minor comments, and a suggestion for further fallback.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
malhotrashivam: I didn't add that because we don't want to filter directories not ending with `.parquet`

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
malhotrashivam: IIRC I did. Will add a unit test for that case.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
malhotrashivam: @lbooker42 mentioned that he will be doing renaming in his changes, so I didn't do it here.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
rcaudy: Exactly. I was saying to only filter out *regular files* that don't end in Parquet.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
rcaudy: My complaint was more that we're adding more work with some of the new code, and creating a "moving target" for him.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/location/ParquetColumnLocation.java
```diff
@@ -162,15 +162,24 @@ public <METADATA_TYPE> METADATA_TYPE getMetadata(
...
+            ParquetFileReader parquetFileReader;
             try {
                 parquetFileReader = new ParquetFileReader(groupingFilePath, tl().getChannelProvider());
-            } catch (IOException e) {
-                log.warn().append("Failed to read expected grouping file ").append(groupingFilePath)
-                        .append(" for table location ").append(tl()).append(", column ")
-                        .append(getName())
-                        .endl();
-                return null;
+            } catch (final UncheckedIOException e1) {
```
rcaudy, line 168: This isn't quite right. We would only want to fall back to legacy default if we had no explicit grouping column info, that is if `groupingColumnInfo == null`.
So the logic should be:
```
if groupingColumnInfo != null {
    try to load from there
    // no fallback
} else {
    try to load from new default path
    if that fails {
        try to load from legacy default path
    }
}
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
rcaudy: Now, that could be much more expensive, since we'd have to stat every file we look at, but since we're only looking for the first that matches our filter it may not be so bad.

malhotrashivam: Discussed offline, skipping this change.

lbooker42 (commented):
The renaming of group to index looks good.  Considering the code base still supports grouping (and not multi-column indexes) I think this is as far as we should go in this PR.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -238,7 +237,7 @@ private static String minusParquetSuffix(@NotNull final String s) {
     }
 
     /**
-     * Generates a grouping file path relative to the table destination file path.
+     * Generates the index file path relative to the table destination file path.
```
rcaudy, line 240: Incomplete. Still talks about grouping files, and uses "g" as the example column name.

rcaudy (approved):

PR #4528, opened 2023-09-20
Support ZonedDateTime columns in plots

### LINKED ISSUE #4466: Plotting doesn't respect ZonedDateTime as a numeric column
Steps to reproduce: 
1. Create a table with a ZonedDateTime column (instead of an Instant)
2. Attempt to create a plot using that column as if it were numeric/time

Expected: chart displays properly

Actual: error like `Attempting to create a dataseries with a non-numeric column: column=t`

Workaround: Replace the ZonedDateTime column with an Instant one, such as with a formula like `t=toInstant(t)`.

Example:
```python

t = empty_table(1).update_view(formulas=['t=(ZonedDateTime)null', 'y=(long)null'])
plot = Figure().plot_xy(t=t, x="t", y="y", series_name="sample").show()
```
This can be "fixed" by just changing the cast to Instant instead.

It looks like ColumnHandlerFactory should have specific support for ZonedDateTime.

chipkent:
Support ZonedDateTime columns in plots.

Resolves #4466

### DIFF
--- a/Plot/src/main/java/io/deephaven/plot/util/tables/ColumnHandlerFactory.java
+++ b/Plot/src/main/java/io/deephaven/plot/util/tables/ColumnHandlerFactory.java
@@ -410,7 +410,7 @@ public class ColumnHandlerFactory implements Serializable {
 
                 @Override
                 public double getDouble(int i) {
-                    final Instant value = DateTimeUtils.toInstant((ZonedDateTime) getDataColumn().get(i));
+                    final ZonedDateTime value = (ZonedDateTime) getDataColumn().get(i);
                     return value == null ? Double.NaN : DateTimeUtils.epochNanos(value);
                 }
 
@@ -628,6 +628,20 @@ public class ColumnHandlerFactory implements Serializable {
                 @Override
                 public double getDouble(int i) {
                     final Instant value = (Instant) getDataColumn().get(i);
+                    return value == null ? Double.NaN : DateTimeUtils.epochNanos(value);
+                }
+
+            };
+        } else if (type.equals(ZonedDateTime.class)) {
+            return new ColumnHandlerTable(table, columnName, type, plotInfo) {
+                @Override
+                public TypeClassification typeClassification() {
+                    return TypeClassification.TIME;
+                }
+
+                @Override
+                public double getDouble(int i) {
+                    final ZonedDateTime value = (ZonedDateTime) getDataColumn().get(i);
                     return value == null ? Double.NaN : DateTimeUtils.epochNanos(value);
                 }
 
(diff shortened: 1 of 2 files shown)


mofojed (approved):
Tested with the snippet in the ticket:
```

t = empty_table(1).update_view(formulas=['t=(ZonedDateTime)null', 'y=(long)null'])
plot = Figure().plot_xy(t=t, x="t", y="y", series_name="sample").show()
```

Plot/src/main/java/io/deephaven/plot/util/tables/ColumnHandlerFactory.java
```diff
@@ -399,6 +400,20 @@ public double getDouble(int i) {
...
+                @Override
+                public TypeClassification typeClassification() {
+                    return TypeClassification.TIME;
+                }
+
+                @Override
+                public double getDouble(int i) {
+                    final Instant value = DateTimeUtils.toInstant((ZonedDateTime) getDataColumn().get(i));
+                    return value == null ? Double.NaN : DateTimeUtils.epochNanos(value);
+                }
```
rcaudy, line 415: Does this even get used? I have the impression that web is snapshotting data directly rather than via any Plot code. Which is good, because this is a pretty sub-optimal way to fetch bulk data.

chipkent: Here is some test code I have been using.

```python
from deephaven.time import to_j_time_zone

tz = to_j_time_zone("ET")
t = time_table("PT00:00:01").update(["X=toZonedDateTime(Timestamp,tz)","Y=i"])
m = t.meta_table
p = Figure().plot_xy(series_name="Test", t=t, x="X", y="Y").show()
```

In the 0.28.1 demo, this fails with:

```
java.lang.UnsupportedOperationException: Unsupported numeric data type: columnName=X type=class java.time.ZonedDateTime
```

The equivalent Groovy is passing in my branch code.
```groovy

tz = timeZone("ET")
t = timeTable("PT00:00:01").update("X=toZonedDateTime(Timestamp,tz)","Y=i")
m = t.meta()
p = figure().plot_xy("Test", t, "X", "Y").show()
```

I will disable and test the code again.

chipkent: Confirmed that commenting out the new code results in an exception: ```java.lang.UnsupportedOperationException: Unsupported numeric data type: columnName=X type=class java.time.ZonedDateTime```

chipkent: A deep dive indicates that some, but not all of this code is called.  The methods to get type are called, but the `getDouble` methods are not called.  Removal of those functions should be part of a more comprehensive refactor and not this PR.

mofojed (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3229
Reference: https://github.com/deephaven/deephaven.io/issues/3230

PR #4529, opened 2023-09-20
Make em* pydocs consistent with one another, fix emstd* formulas

alexpeters1208:
I noticed some inconsistencies with variable stylings in the `em*` formulas, so I fixed them. I also noticed that the formulae documented for `emstd_tick` and `emstd_time` were not consistent with the Java operations and did not yield correct answers when I coded them up by hand, so I fixed those too.


lbooker42 (commented):
These changes look good to me.  Can you confirm that you corrected the indentation problems in the `Args:` for several operators?  I suspect you did but that white-space diffs are hiding the changes.

lbooker42 (approved):

jmao-denver (approved):
LGTM

PR #4530, opened 2023-09-20
Receive time for Kafka messages.

### LINKED ISSUE #4531: Add Kafka Received Time
As a data manager, I want to have my Kafka table include the time that we received the message so that I can better understand the performance of my data pipeline.

cpwright:
Fixes #4531.

### DIFF
--- a/extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
+++ b/extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
@@ -138,8 +138,8 @@ public class KafkaTools {
     public static final String OFFSET_COLUMN_NAME_DEFAULT = "KafkaOffset";
     public static final String TIMESTAMP_COLUMN_NAME_PROPERTY = "deephaven.timestamp.column.name";
     public static final String TIMESTAMP_COLUMN_NAME_DEFAULT = "KafkaTimestamp";
-    public static final String RECEIVETIME_COLUMN_NAME_PROPERTY = "deephaven.receivetime.column.name";
-    public static final String RECEIVETIME_COLUMN_NAME_DEFAULT = null;
+    public static final String RECEIVE_TIME_COLUMN_NAME_PROPERTY = "deephaven.receivetime.column.name";
+    public static final String RECEIVE_TIME_COLUMN_NAME_DEFAULT = null;
     public static final String KEY_BYTES_COLUMN_NAME_PROPERTY = "deephaven.keybytes.column.name";
     public static final String KEY_BYTES_COLUMN_NAME_DEFAULT = null;
     public static final String VALUE_BYTES_COLUMN_NAME_PROPERTY = "deephaven.valuebytes.column.name";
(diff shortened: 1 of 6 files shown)


extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -138,6 +138,12 @@ public class KafkaTools {
     public static final String OFFSET_COLUMN_NAME_DEFAULT = "KafkaOffset";
     public static final String TIMESTAMP_COLUMN_NAME_PROPERTY = "deephaven.timestamp.column.name";
     public static final String TIMESTAMP_COLUMN_NAME_DEFAULT = "KafkaTimestamp";
+    public static final String RECEIVETIME_COLUMN_NAME_PROPERTY = "deephaven.receivetime.column.name";
+    public static final String RECEIVETIME_COLUMN_NAME_DEFAULT = null;
```
rcaudy, line 142: Let's be consistent... if we're going to say "KEY_BYTES", we should say "RECEIVE_TIME".
```suggestion
    public static final String RECEIVE_TIME_COLUMN_NAME_PROPERTY = "deephaven.receivetime.column.name";
    public static final String RECEIVE_TIME_COLUMN_NAME_DEFAULT = null;
```

rcaudy (commented):
LGTM, barring minor quibbles. Highlights that this code sure does have a lot of touch points for each "common column".

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaStreamPublisher.java
```diff
@@ -127,7 +137,10 @@ public static ConsumerRecordToStreamPublisherAdapter make(
                 simpleKeyColumnIndex,
                 simpleValueColumnIndex,
                 parameters.getKeyToChunkObjectMapper(),
-                parameters.getValueToChunkObjectMapper());
+                parameters.getValueToChunkObjectMapper(),
+                parameters.getReceiveTimeColumnIndex(),
+                parameters.getKeyBytesColumnIndex(),
+                parameters.getValueBytesColumnIndex());
```
devinrsmith, line 143: More a note for Ryan and myself - wonder if we should just pass parameters into KafkaStreamPublisher instead of extracting all the values here.

devinrsmith (commented):
My one comment about using the system clock I'll leave up to the discretion of Charles; it's not a sticking point, although I may change it in the future if you don't.

Before I hit approve, I'm going to test out the code.

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaStreamPublisher.java
```diff
@@ -258,6 +299,15 @@ public synchronized long consumeRecords(@NotNull final List<? extends ConsumerRe
...
                 }
+                if (receiveTimeChunk != null) {
+                    receiveTimeChunk.add(receiveTime);
+                }
+                if (keyBytesChunk != null) {
+                    keyBytesChunk.add(record.serializedKeySize());
+                }
+                if (valueBytesChunk != null) {
+                    valueBytesChunk.add(record.serializedValueSize());
+                }
```
devinrsmith, line 310: Upon testing, we should treat -1 as NULL_INT.

devinrsmith (changes requested):

devinrsmith (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3217
Reference: https://github.com/deephaven/deephaven.io/issues/3218

PR #4538, opened 2023-09-22
Make Kafka KeyOrValueProducers public to enable external parsers.

cpwright:


extensions/kafka/src/main/java/io/deephaven/kafka/KafkaTools.java
```diff
@@ -313,19 +313,19 @@ public static class Consume {
...
-            abstract KeyOrValueIngestData getIngestData(
+            public abstract KeyOrValueIngestData getIngestData(
                     KeyOrValue keyOrValue,
                     SchemaRegistryClient schemaRegistryClient,
                     Map<String, ?> configs,
                     MutableInt nextColumnIndexMut,
                     List<ColumnDefinition<?>> columnDefinitionsOut);
 
-            abstract KeyOrValueProcessor getProcessor(
+            public abstract KeyOrValueProcessor getProcessor(
```
rcaudy, line 328: Protected, not public, I think.

rcaudy:
We'll merge this for expediency, but we're planning to change these interfaces in the October release; implementor beware.

rcaudy (approved):

Issue #4537, opened 2023-09-22
SourceTable removes itself from the UG instead of the updateSourceRegistrar


abaranec:
**Description**

When a SourceTable fails, it tries to remove itself from the UpdateGraph,  but it really should be removing itself from the updateSourceRegistrar it was created with.

**Steps to reproduce**

This is difficult to do without the ability to remove a location from a TableLocaftionProvider, but is like this:

1. Create a SourcePartitionedTable that initially has two locations
2. Force one of the to locations to fail (by pushing a negative size update or something)
3. Push a delete event through the TableLocationProvider

**Expected results**

The SourcePartitionedTable removes the location and life continues

**Actual results**

An assertion because the SourcePartitionedTable tries to refresh the table that has already failed

**Versions**
 - Deephaven: 0.28.0

PR #4539, opened 2023-09-22
SourceTable should remove itself from registrar, not UG

abaranec:
This fixes Issue #4537.

### DIFF
--- a/engine/table/src/test/java/io/deephaven/engine/table/impl/SourcePartitionedTableTest.java
+++ b/engine/table/src/test/java/io/deephaven/engine/table/impl/SourcePartitionedTableTest.java
@@ -213,6 +213,11 @@ public class SourcePartitionedTableTest extends RefreshingTableTestCase {
             updateGraph.refreshSources();
             registrar.run();
         });
+
+        assertEquals(1, partitionTable.size());
+        try (final CloseableIterator<Table> tableIt = partitionTable.columnIterator("LocationTable")) {
+            assertTableEquals(tableIt.next(), p2);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


engine/table/src/test/java/io/deephaven/engine/table/impl/SourcePartitionedTableTest.java
```diff
@@ -3,11 +3,16 @@
 import io.deephaven.configuration.Configuration;
 import io.deephaven.engine.context.ExecutionContext;
 import io.deephaven.engine.primitive.iterator.CloseableIterator;
+import io.deephaven.engine.rowset.RowSet;
+import io.deephaven.engine.rowset.RowSetFactory;
+import io.deephaven.engine.rowset.TrackingRowSet;
```
rcaudy, line 8: You added two unused imports
```suggestion
```

engine/table/src/test/java/io/deephaven/engine/table/impl/SourcePartitionedTableTest.java
```diff
@@ -166,10 +171,60 @@ public void testAddAndRemoveLocations() {
                         FindExceptionCause.isOrCausedBy(errors.get(0), IllegalStateException.class).isPresent());
     }
 
+    /**
+     * This is a test for PR 4537, where SourceTable removes itself from the wrong refresh provider
+     */
+    public void testRemoveAndFail() {
```
rcaudy, line 177: This is not annotated as a test...
```suggestion
    @Test
    public void testRemoveAndFail() {
```

rcaudy (approved):

PR #4541, opened 2023-09-22
Refactored parquet writing code

### LINKED ISSUE #4524: Parquet writing for arrays is slower than it should be, because the implementation relies on ungroup()
Right now, we use a `lengthsSource` and a `valuesSource` for vector-style writing:
```
                final Table lengthsTable = new QueryTable(tableRowSet, columns);
                final ColumnSource<?> lengthSource = lengthsTable
                        .view("len= ((Object)array) == null ? null : (int)array."
                                + (Vector.class.isAssignableFrom(valueSource.getType()) ? "size()" : "length"))
                        .getColumnSource("len");
                final Table ungroupedArrays = lengthsTable.ungroup("array");
                vectorHelper = new VectorColumnWriterHelper(lengthSource, ungroupedArrays.getRowSet());
                helper = vectorHelper;
                valueSource = ungroupedArrays.getColumnSource("array");
```

We should probably re-write this in terms of a `vectorSource` and consume one element at a time, rather than relying on the `ungroup()` operation.

malhotrashivam:
Closes #4524


extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/VariableWidthTransfer.java
```diff
@@ -0,0 +1,135 @@
...
+            this(data, numBytes);
+            this.numValues = numValues;
+        }
+    }
+
+    // TODO Add comments about what to be done before and after calling this method
+    final void transferOnePageToBufferHelper() {
+        if (!hasMoreDataToBuffer()) {
+            return;
+        }
```
rcaudy, line 78: If we're hitting this, maybe there's a bug in the calling code?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/VariableWidthTransfer.java
```diff
@@ -0,0 +1,135 @@
...
+                if (cachedValue == null) {
+                    nextEntry = encodeDataForBuffering(data);
+                } else {
+                    nextEntry = cachedValue;
+                    cachedValue = null;
+                }
+                int numBytesBuffered = getNumBytesBuffered();
+                // Always copy the first entry
+                if ((numBytesBuffered != 0 && numBytesBuffered + nextEntry.numBytes > targetPageSize) ||
+                        !addEncodedDataToBuffer(nextEntry)) {
```
rcaudy, line 109: I wonder if we should be passing a `force` boolean, instead of making the implementations responsible for knowing it's the first entry for a page and they *have* to buffer it.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/VariableWidthTransfer.java
```diff
@@ -0,0 +1,135 @@
...
+                    stop = true;
+                    cachedValue = nextEntry;
```
rcaudy, line 111: It's sort of clever that if this is `null`, it just falls through to the null entry case in the chunk the next time around. Maybe _too_ clever, since it took me a minute to catch on.

rcaudy:
Note that my comments so far are more "stuff I was thinking about while reading the code so far". I'll do a much more careful pass once this is more complete/polished. I like the direction a *lot*.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/VariableWidthTransfer.java
```diff
@@ -0,0 +1,135 @@
...
+                currentChunkIdx++;
+            }
+            if (currentChunkIdx == chunk.size()) {
+                chunk = null;
+            }
+        } while (!stop && tableRowSetIt.hasMore());
+    }
+
+    // TODO Add comments explaining what each does
+    abstract boolean addNullToBuffer();
```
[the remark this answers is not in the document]
malhotrashivam: False would mean that we could not add null to the buffer because its full) and true means we did.
We do have different handling logics for null for different types. For example, for dict encoded strings, we store `NULL_INT` in the IntBuffer, for plain encoded strings, we add `null` to the binary array, and for all vector and array types, we add NULL_INT in the repeatCounts buffer.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/VariableWidthTransfer.java
malhotrashivam: Yup, you are right, I can remove this check

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetInstructions.java
```diff
@@ -83,7 +83,7 @@ public static int getDefaltMaximumDictionarySize() {
         return defaultMaximumDictionarySize;
     }
 
-    private static final int MIN_TARGET_PAGE_SIZE =
+    public static final int MIN_TARGET_PAGE_SIZE =
```
malhotrashivam, line 86: Pending comments from https://github.com/deephaven/deephaven-core/pull/4344#discussion_r1326530910

rcaudy (commented):
This is a partial review we worked on on Friday and I forgot to submit.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/InstantArrayTransfer.java
```diff
@@ -0,0 +1,48 @@
...
+import io.deephaven.engine.rowset.RowSequence;
+import io.deephaven.engine.table.ColumnSource;
+import io.deephaven.time.DateTimeUtils;
+import org.jetbrains.annotations.NotNull;
+
+import java.nio.LongBuffer;
+import java.time.Instant;
+
+
+// TODO Ask Ryan if we need to do write something like ReinterpretUtils.instantToLongSource since that's what we do
```
malhotrashivam, line 15: Check with Ryan

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -416,53 +420,133 @@ public void testBigDecimalPrecisionScale() {
         }
     }
 
+    // TODO Delete this
```
malhotrashivam, line 423: Delete this when all types are added

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -270,6 +255,13 @@ private static Table pretransformTable(@NotNull final Table table, @NotNull fina
         return transformed;
     }
 
+    private static boolean isVectorDirectType(Class<?> type) {
+        // TODO Would need to add more types, check with Ryan if this seems okay
```
malhotrashivam, line 259: This is pending

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -341,458 +337,144 @@ private static ParquetFileWriter getParquetFileWriter(
...
-                        // Split a single page into multiple if we are not able to fit all the entries in one page
-                        do {
-                            int numValuesBuffered = transferObject.transferOnePageToBuffer();
-                            columnWriter.addPage(transferObject.getBuffer(), numValuesBuffered, statistics);
-                        } while (transferObject.hasMoreDataToBuffer());
-                    }
-                }
-            }
-        }
+    // TODO Is it okay to keep this method here? Its only called inside tryEncodeDictionary. I can also make it an
```
malhotrashivam, line 364: Check with Ryan if a better way to do this

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/metadata/ColumnTypeInfo.java
```diff
@@ -39,7 +39,7 @@ public abstract class ColumnTypeInfo {
      * Values for the {@link #specialType()} field.
      */
     public enum SpecialType {
-        StringSet, Vector
+        StringSet, Vector, VectorDirect
```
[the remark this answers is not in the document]
rcaudy: If we think we need this kind of functionality, let's record explicit types like we already do for codec'd fields. This is at best a single-purpose fix.

I would argue that it's a weird practice to end up with direct vectors as the explicit types in a column.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/ObjectTransfer.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    final int getNumBytesBuffered() {
+        return numBytesBuffered;
+    }
+
+    @Override
+    final boolean isBufferEmpty() {
+        return numBytesBuffered == 0;
```
rcaudy, line 50:
```suggestion
        return bufferedDataCount == 0;
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/ArrayAndVectorTransfer.java
```diff
@@ -0,0 +1,44 @@
...
+        if (!repeatCounts.hasRemaining()) {
+            return false;
+        }
+        repeatCounts.put(QueryConstants.NULL_INT);
+        return true;
+    }
+
+    @Override
+    final boolean isBufferEmpty() {
+        return getNumBytesBuffered() == 0 && repeatCounts.position() == 0;
```
rcaudy, line 41:
```suggestion
        return repeatCounts.position() == 0;
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/ObjectArrayAndVectorTransfer.java
```diff
@@ -0,0 +1,72 @@
...
+        // Clear any old buffered data
+        if (bufferedDataCount != 0) {
+            Arrays.fill(buffer, 0, bufferedDataCount, null);
+            bufferedDataCount = 0;
+            numBytesBuffered = 0;
+            repeatCounts.clear();
+        }
+        // Fill the buffer with data from the table
+        transferOnePageToBufferHelper();
+        repeatCounts.flip();
```
rcaudy, line 41: So, we flip repeats, but not nullOffsets?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/ObjectArrayAndVectorTransfer.java
```diff
@@ -0,0 +1,72 @@
...
+        if (!repeatCounts.hasRemaining()) {
+            Assert.eqFalse(force, "force");
+            return false;
+        }
+        final int numEncodedValues = data.numValues;
+        if (numEncodedValues > bufferSize - bufferedDataCount) {
+            if (force) {
+                // Resize the buffer
+                bufferSize = numEncodedValues + bufferedDataCount;
+                buffer = new Binary[bufferSize];
```
rcaudy, line 60: Since the layer above controls force, we should probably protectively copy the existing data. Or, we could assert that there is none.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/PrimitiveArrayAndVectorTransfer.java
```diff
@@ -0,0 +1,63 @@
...
+        repeatCounts.flip();
+        return buffer.limit();
+    }
+
+    final boolean addEncodedDataToBuffer(@NotNull final EncodedData<E> data, boolean force) {
+        if (!repeatCounts.hasRemaining()) {
+            Assert.eqFalse(force, "force");
+            return false;
+        }
+        if (data.numValues > buffer.remaining()) {
```
rcaudy, line 40: I don't think you should re-size the buffers down, but I do think you should refuse to write data that doesn't fit into maxValuesPerPage unless forced. Applies to all similar codecs.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/ByteArrayTransfer.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package io.deephaven.parquet.table.transfer;
+
+import io.deephaven.engine.rowset.RowSequence;
+import io.deephaven.engine.table.ColumnSource;
+import org.jetbrains.annotations.NotNull;
+
+import java.nio.IntBuffer;
+
+final class ByteArrayTransfer extends PrimitiveArrayAndVectorTransfer<byte[], byte[], IntBuffer> {
```
rcaudy, line 12: It looks to me like we should be adding these to `ReplicateParquetTransferObjects`.
You should replicate char to byte and short, and replicate int to long, float, and double.

Same logic applies for the vector codecs.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/CharArrayTransfer.java
```diff
@@ -0,0 +1,41 @@
...
+final class CharArrayTransfer extends PrimitiveArrayAndVectorTransfer<char[], char[], IntBuffer> {
+    CharArrayTransfer(@NotNull final ColumnSource<?> columnSource, @NotNull final RowSequence tableRowSet,
+                      final int targetPageSize) {
+        // We encode characters as integers
+        super(columnSource, tableRowSet, targetPageSize / Integer.BYTES, targetPageSize,
+                IntBuffer.allocate(targetPageSize / Integer.BYTES));
+    }
+
+    @Override
+    void encodeDataForBuffering(final char @NotNull [] data, @NotNull final EncodedData<char[]> encodedData) {
```
rcaudy, line 21:
```suggestion
    void encodeDataForBuffering(@NotNull final char [] data, @NotNull final EncodedData<char[]> encodedData) {
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/CharArrayTransfer.java
```diff
@@ -0,0 +1,41 @@
...
+import io.deephaven.engine.rowset.RowSequence;
+import io.deephaven.engine.table.ColumnSource;
+import org.jetbrains.annotations.NotNull;
+
+import java.nio.IntBuffer;
+
+final class CharArrayTransfer extends PrimitiveArrayAndVectorTransfer<char[], char[], IntBuffer> {
+    CharArrayTransfer(@NotNull final ColumnSource<?> columnSource, @NotNull final RowSequence tableRowSet,
+                      final int targetPageSize) {
+        // We encode characters as integers
```
rcaudy, line 15: I've noticed that this comment carried through (incorrectly) to `ByteArrayTransfer`. I think it might replicate cleanly this way, but you should double check.
```suggestion
        // We encode primitive chars as primitive ints
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/CharArrayTransfer.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    void resizeBuffer(final int length) {
+        buffer = IntBuffer.allocate(length);
+    }
+
+    @Override
+    void copyToBuffer(final char @NotNull [] data) {
+        for (char c : data) {
+            buffer.put((int) c);
```
rcaudy, line 38: Unnecessary int cast for `put`.
```suggestion
            buffer.put(c);
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/CharVectorTransfer.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    @Override
+    void resizeBuffer(final int length) {
+        buffer = IntBuffer.allocate(length);
+    }
+
+    @Override
+    void copyToBuffer(@NotNull final CharVector data) {
+        data.iterator().forEachRemaining((char value) -> buffer.put((int) value));
```
rcaudy, line 28: Vector iterators should be closed. Unnecessary int cast for `put`.
```suggestion
        try (final CloseablePrimitiveIteratorOfChar dataIterator = data.iterator()) {
            dataIterator.forEachRemaining((char value) -> buffer.put(value));
        }
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/IntVectorTransfer.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    @Override
+    void resizeBuffer(final int length) {
+        buffer = IntBuffer.allocate(length);
+    }
+
+    @Override
+    void copyToBuffer(@NotNull final IntVector data) {
+        data.iterator().forEachRemaining((int value) -> buffer.put(value));
```
rcaudy, line 27: Vector iterators must be closed.
```suggestion
        try (final CloseablePrimitiveIteratorOfInt dataIterator = data.iterator()) {
            dataIterator.forEachRemaining((int value) -> buffer.put(value));
        }
```

rcaudy (commented):
Comments on primitive array and vector transfers. Should replicate these.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/CharArrayTransfer.java
malhotrashivam: IntelliJ keeps giving me warnings that `Not annotated parameter overrides @NotNull parameter`, that's why I kept it like this.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/CharArrayTransfer.java
malhotrashivam: I wanted to make the conversion explicit, just for clarity. But I can change it back .

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/ObjectVectorTransfer.java
```diff
@@ -0,0 +1,52 @@
...
+                         final int targetPageSize) {
+        super(columnSource, tableRowSet, targetPageSize);
+        encodedDataBuf = null;
+    }
+
+    @Override
+    final void encodeDataForBuffering(final @NotNull ObjectVector<T> data, @NotNull final EncodedData<Binary[]> encodedData) {
+        int numObjects = data.intSize();
+        if (encodedDataBuf == null || numObjects > encodedDataBuf.length) {
+            encodedDataBuf = new Binary[numObjects];
```
rcaudy, line 34: Same issue as we've observed before, which is that we should be copying existing contents, or asserting that there are none.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/IntArrayTransfer.java
```diff
@@ -0,0 +1,39 @@
...
+        return buffer.position() * Integer.BYTES;
+    }
+
+    @Override
+    void resizeBuffer(final int length) {
+        buffer = IntBuffer.allocate(length);
+    }
+
+    @Override
+    void copyToBuffer(final int @NotNull [] data) {
```
rcaudy, line 36:
```suggestion
    void copyToBuffer(@NotNull final int [] data) {
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/IntArrayTransfer.java
```diff
@@ -0,0 +1,39 @@
...
+
+final class IntArrayTransfer extends PrimitiveArrayAndVectorTransfer<int[], int[], IntBuffer> {
+    IntArrayTransfer(@NotNull final ColumnSource<?> columnSource, @NotNull final RowSequence tableRowSet,
+                     final int targetPageSize) {
+        super(columnSource, tableRowSet, targetPageSize / Integer.BYTES, targetPageSize,
+                IntBuffer.allocate(targetPageSize / Integer.BYTES));
+    }
+
+    @Override
+    void encodeDataForBuffering(final int @NotNull [] data, @NotNull final EncodedData<int[]> encodedData) {
```
rcaudy, line 21:
```suggestion
    void encodeDataForBuffering(@NotNull final int [] data, @NotNull final EncodedData<int[]> encodedData) {
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/ObjectVectorTransfer.java
```diff
@@ -0,0 +1,52 @@
...
+ * encoding.
+ */
+abstract class ObjectVectorTransfer<T> extends ObjectArrayAndVectorTransfer<ObjectVector<T>> {
+    /**
+     * Used as a temporary buffer for storing encoded data for a single row before it is copied to the main buffer.
+     * Allocated lazily because of the high cost of construction of Binary objects.
+     */
+    private Binary[] encodedDataBuf;
+
+    ObjectVectorTransfer(final @NotNull ColumnSource<?> columnSource, final @NotNull RowSequence tableRowSet,
```
rcaudy, line 24: You might just want to do a search/replace for `final @NotNull` to `@NotNull final`.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/ObjectArrayTransfer.java
```diff
@@ -0,0 +1,51 @@
...
+                        final int targetPageSize) {
+        super(columnSource, tableRowSet, targetPageSize);
+        encodedDataBuf = null;
+    }
+
+    @Override
+    final void encodeDataForBuffering(final T @NotNull [] data, @NotNull final EncodedData<Binary[]> encodedData) {
+        int numObjects = data.length;
+        if (encodedDataBuf == null || numObjects > encodedDataBuf.length) {
+            encodedDataBuf = new Binary[numObjects];
```
rcaudy, line 32: Same resize concern.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/InstantArrayTransfer.java
rcaudy, line 15:
```suggestion
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/InstantVectorTransfer.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Override
+    void resizeBuffer(final int length) {
+        buffer = LongBuffer.allocate(length);
+    }
+
+    @Override
+    void copyToBuffer(@NotNull final ObjectVector<Instant> data) {
+        data.iterator().forEachRemaining((Instant value) -> buffer.put(DateTimeUtils.epochNanos(value)));
```
rcaudy, line 29:
```suggestion
        try (final CloseableIterator<Instant> dataIterator = data.iterator()) {
            dataIterator.forEachRemaining((Instant value) -> buffer.put(DateTimeUtils.epochNanos(value)));
        }
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/PrimitiveVectorTransfer.java
```diff
@@ -0,0 +1,36 @@
...
+import io.deephaven.engine.table.ColumnSource;
+import io.deephaven.vector.Vector;
+import org.jetbrains.annotations.NotNull;
+
+import java.nio.Buffer;
+
+/**
+ * Used as a base class of transfer objects for vectors of primitive types.
+ */
+abstract class PrimitiveVectorTransfer<T extends Vector<?>, B extends Buffer>
```
rcaudy, line 16: Why don't you have a similar `PrimitiveArrayTransfer` class? I would either remove this layer, or have the same layering in the array case, unless I'm missing some complexity there.

rcaudy (commented):
Everything but the dictionary stuff

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/DictEncodedStringTransferBase.java
```diff
@@ -0,0 +1,88 @@
...
+
+    DictEncodedStringTransferBase(@NotNull ColumnSource<?> columnSource, @NotNull RowSequence tableRowSet,
+            int targetPageSize, @NotNull StringDictionary dictionary, final int nullValue) {
+        super(columnSource, tableRowSet, targetPageSize / Integer.BYTES, targetPageSize,
+                IntBuffer.allocate(targetPageSize / Integer.BYTES));
+        this.pageHasNull = false;
+        this.dictionary = dictionary;
+        this.nullValue = nullValue;
+        this.dictEncodedValues = IntBuffer.allocate(targetPageSize);
+    }
```
rcaudy, line 34:
```suggestion

    private final StringDictionary dictionary;
    private final int nullValue; // The value to store in buffer for null strings

    private IntBuffer dictEncodedValues;
    private boolean pageHasNull;

    DictEncodedStringTransferBase(@NotNull ColumnSource<?> columnSource, @NotNull RowSequence tableRowSet,
            int targetPageSize, @NotNull StringDictionary dictionary, final int nullValue) {
        super(columnSource, tableRowSet, targetPageSize / Integer.BYTES, targetPageSize,
                IntBuffer.allocate(targetPageSize / Integer.BYTES));
        this.dictionary = dictionary;
        this.nullValue = nullValue;

        this.dictEncodedValues = IntBuffer.allocate(targetPageSize);
        this.pageHasNull = false;
    }
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/DictEncodedStringTransferBase.java
```diff
@@ -0,0 +1,88 @@
...
+        if (numStrings > dictEncodedValues.limit()) {
+            dictEncodedValues = IntBuffer.allocate(numStrings);
+        }
+        int numBytesEncoded = 0;
+        for (int i = 0; i < numStrings; i++) {
+            String value = strSupplier.get();
+            if (value == null) {
+                dictEncodedValues.put(nullValue);
+                pageHasNull = true;
+                numBytesEncoded += Integer.BYTES;
```
rcaudy, line 60: We're only doing this in the null case. That seems backwards from the other code.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/DictEncodedStringTransferBase.java
```diff
@@ -0,0 +1,88 @@
...
+        data.flip();
+        buffer.put(data);
+    }
+
+    @Override
+    final int getNumBytesBuffered() {
+        return buffer.position() * Integer.BYTES;
+    }
+
+    final public boolean pageHasNull() {
```
rcaudy, line 85:
```suggestion
    public final boolean pageHasNull() {
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/DictEncodedStringTransferBase.java
```diff
@@ -0,0 +1,88 @@
...
+ * with all the strings it encounters and generates an IntBuffer of dictionary position values. This class extends
+ * {@link PrimitiveArrayAndVectorTransfer} to manage the dictionary positions in an {@link IntBuffer} similar to an Int
+ * array/vector column.
+ */
+abstract public class DictEncodedStringTransferBase<T>
+        extends PrimitiveArrayAndVectorTransfer<T, IntBuffer, IntBuffer> {
+    private boolean pageHasNull;
+    private final StringDictionary dictionary;
+    private final int nullValue; // The value to store in buffer for null strings
+    private IntBuffer dictEncodedValues;
```
rcaudy, line 24: Seems to me like we might be better off with an array of ints, since we're bulk-putting this into another buffer.

rcaudy (commented):
First pass complete.

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/TestParquetTools.java
```diff
@@ -116,62 +116,63 @@ public static String toS(Object o) {
         }
     }
```
malhotrashivam, line 118: Check with Ryan if we need to support StringSet columns

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/ObjectArrayAndVectorTransfer.java
malhotrashivam: Some day we will flip nullOffsets too.

chipkent (approved):
I just reviewed the python file.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/PrimitiveArrayAndVectorTransfer.java
```diff
@@ -32,19 +36,52 @@ public int transferOnePageToBuffer() {
...
+        throw new UnsupportedOperationException("getSize() not implemented for " + getClass().getSimpleName());
+    }
+
+    @Override
+    final int getNumBytesBuffered() {
+        return buffer.position() * numBytesPerValue;
+    }
+
     final boolean addEncodedDataToBuffer(@NotNull final EncodedData<E> data, boolean force) {
+        if (force && (!repeatCounts.hasRemaining() || buffer.position() != 0)) {
```
rcaudy, line 65: Shouldn't `repeatCounts.position()` also be 0?

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -549,10 +510,255 @@ public void testArrayColumns() {
...
+        }
+        assertEquals(nullPos, dict.add(null));
+        try {
+            dict.add("Never before seen key which should take us over the allowed dictionary size");
+            TestCase.fail("Exception expected for exceeding dictionary size");
+        } catch (DictionarySizeExceededException expected) {
+        }
+    }
+
+    // TODO Remove all these
```
rcaudy, line 551: What's with all these? Don't forget to remove.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/ObjectArrayAndVectorTransfer.java
```diff
@@ -10,21 +10,38 @@
...
+     */
     private int bufferedDataCount;
+    /**
+     * Total number of bytes buffered
+     */
     private int numBytesBuffered;
 
+    /**
+     * Used as a temporary buffer for storing encoded data for a single row before it is copied to the main buffer.
+     * Allocated lazily because of the high cost of construction of Binary objects.
```
rcaudy, line 33: It's not a `Binary`. It's an array of references to `Binary`. Not sure why this one buffer needs to be allocated lazily.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/ObjectArrayAndVectorTransfer.java
```diff
@@ -47,17 +64,51 @@ final int getNumBytesBuffered() {
...
+        encodedData.fillRepeated(encodedDataBuf, numBytesEncoded, numObjects);
+    }
+
+    /**
+     * Encode a single value to binary
+     */
+    abstract Binary encodeToBinary(V value);
+
     final boolean addEncodedDataToBuffer(@NotNull final EncodedData<Binary[]> data, final boolean force) {
+        if (force && (!repeatCounts.hasRemaining() || bufferedDataCount != 0)) {
```
rcaudy, line 96: Shouldn't `repeatCounts.position()` also be 0?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/ObjectTransfer.java
```diff
@@ -47,26 +51,26 @@ final int getNumBytesBuffered() {
...
         }
         buffer[bufferedDataCount++] = null;
         return true;
     }
 
     final boolean addEncodedDataToBuffer(@NotNull final EncodedData<Binary> data, final boolean force) {
-        if (force && numBytesBuffered != 0) {
-            // This should never happen, because numBytesBuffered should be zero if bufferedDataCount is zero
+        if (force && bufferedDataCount != 0) {
+            // This should never happen, because "force" set by caller when adding the very first object
```
rcaudy, line 68:
```suggestion
            // This should never happen, because "force" is only set by the caller when adding the very first object
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/StringDictionary.java
```diff
@@ -44,13 +59,15 @@ public Binary[] getEncodedKeys() {
     }
 
     /**
-     * Add a string key to the dictionary and return its position in the dictionary. The function returns the value
-     * {@link #NULL_KEY_POS} if the key is null.
+     * Add a string key to the dictionary if it's not already present.
+     *
+     * @param key The key to add and/or find the position for
+     * @return {@code key}'s position in the dictionary, or special null key position (passed in constructor) is
```
rcaudy, line 65:
```suggestion
     * @return {@code key}'s position in the dictionary, or special null key position (passed in constructor) if
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/TransferObject.java
```diff
@@ -30,135 +29,164 @@
...
             @NotNull final RowSet tableRowSet,
-            @NotNull final ColumnSource<DATA_TYPE> columnSource,
-            @NotNull final ColumnDefinition<DATA_TYPE> columnDefinition,
-            @NotNull final ParquetInstructions instructions) {
+            @NotNull final ParquetInstructions instructions,
+            @NotNull final Map<String, Map<ParquetCacheTags, Object>> computedCache,
+            @NotNull final String columnName,
+            @NotNull final ColumnSource<DATA_TYPE> columnSource) {
         Class<DATA_TYPE> columnType = columnSource.getType();
         if (int.class.equals(columnType)) {
```
rcaudy, line 38: You don't need to use `.equals` when comparing class objects. Reference equality (`==`) is sufficient.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -404,14 +381,14 @@ private static <DATA_TYPE> boolean tryEncodeDictionary(
...
                     if (isArrayOrVector) {
-                        lengthsBuffers.add(copy(transferObject.getRepeatCount()));
+                        lengthsBuffers.add(makeCopy(transferObject.getRepeatCount()));
                     }
                 }
-                boolean pageHasNulls = transferObject.pageHasNull();
-                hasNulls = hasNulls || pageHasNulls;
-                pageBufferHasNull.set(curPage++, pageHasNulls);
+                if (transferObject.pageHasNull()) {
+                    pageBufferHasNull.set(curPage++);
```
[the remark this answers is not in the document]
malhotrashivam: My bad, will fix this.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/PrimitiveArrayAndVectorTransfer.java
```diff
@@ -14,9 +14,13 @@
  * Used as a base class of transfer objects for arrays/vectors of primitive types.
  */
 abstract class PrimitiveArrayAndVectorTransfer<T, E, B extends Buffer> extends ArrayAndVectorTransfer<T, E, B> {
```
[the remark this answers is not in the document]
malhotrashivam: The T=E condition doesn't hold true for the `DictEncodedStringArrayAndVectorTransfer` class because there we take a vector or array of strings and encode it as an array of integers.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/BooleanTransfer.java
```diff
@@ -15,9 +15,11 @@
 final class BooleanTransfer extends PrimitiveTransfer<WritableByteChunk<Values>, ByteBuffer> {
     // We encode booleans as bytes here and bit pack them with 8 booleans per byte at the time of writing.
     // Therefore, max values per page are (targetPageSize * 8).
-    static BooleanTransfer create(@NotNull final ColumnSource<?> columnSource, @NotNull final RowSet tableRowSet, int targetPageSize) {
+    static BooleanTransfer create(@NotNull final ColumnSource<?> columnSource, @NotNull final RowSet tableRowSet,
+                                  int targetPageSize) {
         final int NUM_BIT_PACKED_BOOLEANS_PER_BYTE = 8;
-        final int maxValuesPerPage = Math.toIntExact(Math.min(tableRowSet.size(), targetPageSize * NUM_BIT_PACKED_BOOLEANS_PER_BYTE));
+        final int maxValuesPerPage = Math.toIntExact(Math.min(tableRowSet.size(),
+                (long) targetPageSize * (long) NUM_BIT_PACKED_BOOLEANS_PER_BYTE));
```
rcaudy, line 22: No need to change this unless you need to make other changes, but you don't need to cast both operands. Java will perform numeric promotion, in this case a widening primitive conversion, on the operand you don't cast.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/IntTransfer.java
```diff
@@ -13,8 +13,9 @@
 import java.nio.IntBuffer;
 
 final class IntTransfer extends PrimitiveTransfer<WritableIntChunk<Values>, IntBuffer> {
-    static IntTransfer create(@NotNull final ColumnSource<?> columnSource, @NotNull final RowSet tableRowSet, final int targetPageSize) {
-        final int maxValuesPerPage = Math.toIntExact(Math.min(tableRowSet.size(), targetPageSize / Integer.BYTES));
+    static IntTransfer create(@NotNull final ColumnSource<?> columnSource, @NotNull final RowSet tableRowSet,
+                              final int targetPageSize) {
+        final int maxValuesPerPage = Math.toIntExact(Math.min(tableRowSet.size(), (long) (targetPageSize / Integer.BYTES)));
```
rcaudy, line 18: I'm not sure the long cast is needed for a division...

rcaudy (commented):
We need testing coverage for:
1. Transfer resize due to arrays and vectors that don't fit in the target page size
2. String arrays

rcaudy (approved):

PR #4542, opened 2023-09-22
Add kwargs to function generated tables

alexpeters1208:
In working on the function-generated tables blog, I wanted to be able to use arguments in the table-generating function. This PR adds that functionality.


py/server/deephaven/table_factory.py
```diff
@@ -396,7 +398,7 @@ def function_generated_table(table_generator: Callable[[], Table],
 
     def table_generator_function():
         with exec_ctx:
-            result = table_generator()
+            result = table_generator(**kwargs)
```
jmao-denver, line 401: The function call here could raise an exception of argument related error, such as unexpected argument etc. What should be our strategy to handle that? We probably should have some test cases to verify that the users won't be confused by the exceptions.

alexpeters1208: It seems that the precedent here is to assume that the table-generating function is going to be responsible and correct. We already don't try to intercept any errors from that function, don't do any checks for its correctness, etc. I think it's reasonable to let the function complain about wrong arguments if it's passed wrong arguments, and leave the  responsibility to the user. I agree that we need unit tests to make sure that the exceptions raised are the "wrong argument" exceptions that I expect.

py/server/tests/test_function_generated_table.py
```diff
@@ -111,6 +111,36 @@ def table_generator_function():
...
+        # we can be confident it's working.
+        self.assertGreater(later_time, initial_time)
+
+    def test_generated_table_wrong_args(self):
+        def table_generator_function(nrows, query_string):
+            return empty_table(nrows).update(query_string)
+
+        # nrow is not a valid argument, so we expect a type error
+        with update_graph.exclusive_lock(self.test_update_graph):
+            self.assertRaises(TypeError, table_generator_function, nrow=1, query_string="Timestamp = io.deephaven.base.clock.Clock.system().currentTimeMillis()")
```
jmao-denver, line 142:
```
       with self.assertRaises(TypeError) as cm:
            j_array = dtypes.array("java.lang.String", ["a", "b"])
        self.assertIn("expects a DType", str(cm.exception))
```
This is a pattern we use to additionally check if the error message is what we expect. In this case, we want the exception to clearly tell us the error is from calling table_generator_function not hidden inside the `function_generated_table` call.

alexpeters1208: Recent update has corrected this.

py/server/deephaven/table_factory.py
```diff
@@ -373,14 +373,14 @@ def function_generated_table(table_generator: Callable[[], Table],
         refresh_interval_ms (int): Interval (in milliseconds) at which the 'table_generator' function is rerun.
         exec_ctx (ExecutionContext) -> Table): A custom execution context. If 'None', the current
             execution context is used. If there is no current execution context, a ValueError is raised.
+        **kwargs (Dictionary): Optional keyword arguments to pass to table_generator.
```
jmao-denver, line 376: is `Dictionary` a valid type? Since `**kwargs` is so Python idiomatic, I don't think we need to provide type info for it.

alexpeters1208: I think this would be reasonable.

py/server/deephaven/table_factory.py
```diff
@@ -373,14 +373,14 @@ def function_generated_table(table_generator: Callable[[], Table],
         refresh_interval_ms (int): Interval (in milliseconds) at which the 'table_generator' function is rerun.
         exec_ctx (ExecutionContext) -> Table): A custom execution context. If 'None', the current
```
jmao-denver, line 374: Probably a mistake we missed when it was first added.

alexpeters1208: Should it be
```
def function_generated_table(table_generator: Callable,
           source_tables: Union[Table, List[Table]] = None,
           refresh_interval_ms: int = None,
           exec_ctx: ExecutionContext = None,
           **kwargs) -> Table:
```

jmao-denver (approved):
LGTM

py/server/deephaven/table_factory.py
```diff
@@ -342,12 +342,12 @@ def ring_table(parent: Table, capacity: int, initialize: bool = True) -> Table:
...
 
-def function_generated_table(table_generator: Callable[[], Table],
+def function_generated_table(table_generator: Callable[..., Table],
            source_tables: Union[Table, List[Table]] = None,
            refresh_interval_ms: int = None,
-           exec_ctx: ExecutionContext = None) -> Table:
-    """
-    Creates an abstract table that is generated by running the table_generator() function. The function will first be
+           exec_ctx: ExecutionContext = None,
+           **kwargs) -> Table:
```
chipkent, line 349: @jmao-denver should we support `*args` and `**kwargs` or is the latter the standard for something like this?
https://stackabuse.com/variable-length-arguments-in-python-with-args-and-kwargs/

py/server/tests/test_function_generated_table.py
```diff
@@ -111,6 +111,38 @@ def table_generator_function():
...
+    def test_generated_table_wrong_args(self):
+        def table_generator_function(nrows, query_string):
+            return empty_table(nrows).update(query_string)
+
+        # nrow is not a valid argument, so we expect a type error
+        with update_graph.exclusive_lock(self.test_update_graph):
+            with self.assertRaises(TypeError) as cm:
+                result_table = function_generated_table(table_generator_function, refresh_interval_ms=2000,
+                                                        nrow=1, query_string="Timestamp = io.deephaven.base.clock.Clock.system().currentTimeMillis()")
+            self.assertIn("table_generator_function() got an unexpected keyword argument 'nrow'", str(cm.exception))
```
chipkent, line 145: What about a test with too few args?

alexpeters1208: Good call

py/server/deephaven/table_factory.py
jmao-denver: I think requiring keyword arguments only isn’t too much of a burden for the users. Moving the parameters around in the factory function function_generated_table to support positional args for the table_generator function might hurt the usability of function_generated_table

py/server/deephaven/table_factory.py
```diff
@@ -358,29 +359,30 @@ def function_generated_table(table_generator: Callable[[], Table],
...
-        table_generator (Callable[[], Table]): The table generator function. This function must return a Table.
+        table_generator (Callable[..., Table]): The table generator function. This function must return a Table.
         source_tables (Union[Table, List[Table]]): Source tables used by the 'table_generator' function. The
             'table_generator' is rerun when any of these tables tick.
         refresh_interval_ms (int): Interval (in milliseconds) at which the 'table_generator' function is rerun.
-        exec_ctx (ExecutionContext) -> Table): A custom execution context. If 'None', the current
+        exec_ctx (ExecutionContext): A custom execution context. If 'None', the current
             execution context is used. If there is no current execution context, a ValueError is raised.
+        args (tuple): Optional tuple of positional arguments to pass to table_generator.
+        kwargs (dict): Optional dictionary of keyword arguments to pass to table_generator.
```
jmao-denver, line 378:
```suggestion
        args (Tuple): Optional tuple of positional arguments to pass to table_generator. Defaults to ()
        kwargs (Dict): Optional dictionary of keyword arguments to pass to table_generator. Defaults to {}
```

jmao-denver (changes requested):
Minor doc issues.

py/server/deephaven/table_factory.py
```diff
@@ -373,7 +374,8 @@ def function_generated_table(table_generator: Callable[..., Table],
         refresh_interval_ms (int): Interval (in milliseconds) at which the 'table_generator' function is rerun.
         exec_ctx (ExecutionContext): A custom execution context. If 'None', the current
             execution context is used. If there is no current execution context, a ValueError is raised.
-        **kwargs: Optional keyword arguments to pass to table_generator.
+        args (tuple): Optional tuple of positional arguments to pass to table_generator.
+        kwargs (dict): Optional dictionary of keyword arguments to pass to table_generator.
```
chipkent:
```suggestion
        args (Tuple): Optional tuple of positional arguments to pass to table_generator.
        kwargs (Dict): Optional dictionary of keyword arguments to pass to table_generator.
```

jmao-denver (approved):
LGTM

chipkent (approved):

PR #4547, opened 2023-09-25
Ensure that boolean partition columns are boxed

### LINKED ISSUE #4545: Creating parquet table partitioned on a boolean column throws error
**Description**

Creating a parquet table partitioned on a column of `boolean` type and then running `j_object.getColumnSources()` will cause an exception.

**Steps to reproduce**
1. Run the following code snippet:
```python

part = empty_table(4).update("II=ii")

from deephaven.parquet import write, read

write(part, "/tmp/bool-test/boolCol=true/part.parquet")
write(part, "/tmp/bool-test/boolCol=false/part.parquet")

bool_partition = read("/tmp/bool-test")
```
2. Run `bool_partition.j_object.getColumnSources()`

niloc132:
Fixes #4545

### DIFF
--- a/extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
+++ b/extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
@@ -685,8 +685,15 @@ public class ParquetTools {
                 throw new IllegalArgumentException("First location key " + firstKey
                         + " has null partition value at partition key " + partitionKey);
             }
-            allColumns.add(ColumnDefinition.fromGenericType(partitionKey,
-                    getUnboxedTypeIfBoxed(partitionValue.getClass()), null, ColumnDefinition.ColumnType.Partitioning));
+
+            // Primitives should be unboxed, except booleans
+            Class<?> dataType = partitionValue.getClass();
+            if (dataType != Boolean.class) {
+                dataType = getUnboxedTypeIfBoxed(partitionValue.getClass());
+            }
+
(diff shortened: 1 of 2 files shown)


georgecwan (approved):

rcaudy (dismissed):

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -685,8 +685,15 @@ public static Table readPartitionedTableInferSchema(
                 throw new IllegalArgumentException("First location key " + firstKey
                         + " has null partition value at partition key " + partitionKey);
             }
-            allColumns.add(ColumnDefinition.fromGenericType(partitionKey,
-                    getUnboxedTypeIfBoxed(partitionValue.getClass()), null, ColumnDefinition.ColumnType.Partitioning));
+
+            // Primitives should be unboxed, except booleans
```
mofojed, line 689: Why?

niloc132: I believe this is because other types have a null sentinel, but in order for booleans to have a possible null, we need to box them instead (it would be silly to have false mean "not set", etc). This is a convention followed in other parts of the platform, and the check was just missing here (but is present in another path to creating parquet tables, linked in the issue).

rcaudy (approved):

georgecwan (approved):

PR #4550, opened 2023-09-25
Fix missing AuthContext on New Partition

### LINKED ISSUE #4543: PerformanceEntry: Give ExecutionContext Error if AuthContext is null
DHE require that AuthContext will not be null if the PerformanceEntry's id is not NULL_INT. DHC uses a non-null AuthContext, too. We should be able to nip user errors during construction time by throwing an ExecutionContext style exception.

Otherwise, DHE gets an error later on when trying to process the performance entry, but it's harder to locate the context that the table was constructed (which causes the underlying issue and is the location of the user's fix).

nbauernfeind:
Fixes #4543 by improving the time of error when auth context is not provided and by installing an auth context when processing changes to a union source manager (which might create new listeners).

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/perf/PerformanceEntry.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/perf/PerformanceEntry.java
@@ -50,6 +50,9 @@ public class PerformanceEntry extends BasePerformanceEntry implements TableListe
         this.description = description;
         this.callerLine = callerLine;
         authContext = id == QueryConstants.NULL_INT ? null : ExecutionContext.getContext().getAuthContext();
+        if (id != QueryConstants.NULL_INT && authContext == null) {
+            throw new IllegalStateException("No Auth Context is Installed. Do you need to open an ExecutionContext?");
+        }
         this.updateGraphName = updateGraphName;
         startSample = new RuntimeMemory.Sample();
         endSample = new RuntimeMemory.Sample();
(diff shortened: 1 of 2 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/perf/PerformanceEntry.java
```diff
@@ -50,6 +50,9 @@ public class PerformanceEntry extends BasePerformanceEntry implements TableListe
         this.description = description;
         this.callerLine = callerLine;
         authContext = id == QueryConstants.NULL_INT ? null : ExecutionContext.getContext().getAuthContext();
+        if (id != QueryConstants.NULL_INT && authContext == null) {
+            throw new IllegalStateException("No Auth Context is Installed. Do you need to open an ExecutionContext?");
```
rcaudy, line 54: Do we think there is a path that gets us here besides engine bugs and thread mis-use? What AuthContext do clients get by default?
```suggestion
            throw new IllegalStateException("No AuthContext is installed. Do you need to open an ExecutionContext?");
```

engine/table/src/main/java/io/deephaven/engine/table/impl/sources/UnionSourceManager.java
```diff
@@ -61,10 +62,15 @@ public class UnionSourceManager {
...
     private final UpdateCommitter<UnionSourceManager> updateCommitter;
+    private final ExecutionContext executionContext;
 
     public UnionSourceManager(@NotNull final PartitionedTable partitionedTable) {
         constituentChangesPermitted = partitionedTable.constituentChangesPermitted();
         columnNames = partitionedTable.constituentDefinition().getColumnNamesArray();
+        executionContext = ExecutionContext.newBuilder()
+                .markSystemic()
+                .captureUpdateGraph()
+                .build();
```
rcaudy, line 73: You should do this after creating the `UpdateCommitter` in the `isRefreshing()` case, and assign it to `null` in the static case.

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/PerformanceEntry.java
nbauernfeind: A user could see this by constructing a listener on a new thread.

rcaudy (approved):
Let's merge this version, and keep the issue open while we think about more safety checks.

PR #4552, opened 2023-09-26
Enable figures to retain references to objects they rely on

### LINKED ISSUE #4540: Figures don't retain liveness to their tables or partitionedtables
Figures are also a great place if you want to delete some dead code...

niloc132:
Also turns down log levels when testing/generating figure code.

Fixes #4540


Plot/src/main/java/io/deephaven/plot/FigureWidget.java
```diff
@@ -19,10 +22,18 @@ public class FigureWidget extends FigureImpl implements LiveWidget, LiveWidgetVi
     private static final long serialVersionUID = 763409998768966385L;
     private String[] validGroups;
 
-    @SuppressWarnings("WeakerAccess") // this is used in the python integration
     public FigureWidget(final FigureImpl figure) {
         super(figure);
-        figure.getFigure().consolidatePartitionedTables();
+        getFigure().consolidatePartitionedTables();
```
rcaudy, line 27: We said the management should move up to `FigureImpl`.

Plot/src/test/java/io/deephaven/plot/TestLiveness.java
```diff
@@ -0,0 +1,44 @@
...
+
+import io.deephaven.engine.liveness.LivenessReferent;
+import io.deephaven.engine.liveness.LivenessScope;
+import io.deephaven.engine.liveness.LivenessScopeStack;
+import io.deephaven.engine.table.Table;
+import io.deephaven.engine.testutil.testcase.RefreshingTableTestCase;
+import io.deephaven.engine.util.TableTools;
+import io.deephaven.util.SafeCloseable;
+
+public class TestLiveness extends RefreshingTableTestCase {
```
rcaudy, line 11: You said you planned to delete this file.

Plot/src/main/java/io/deephaven/plot/FigureWidget.java
```diff
@@ -19,10 +22,18 @@ public class FigureWidget extends FigureImpl implements LiveWidget, LiveWidgetVi
...
+
+        getFigure().getTableHandles().stream()
+                .map(TableHandle::getTable)
+                .filter(DynamicNode::notDynamicOrIsRefreshing)
+                .forEach(this::manage);
+        getFigure().getPartitionedTableHandles().stream()
+                .map(PartitionedTableHandle::getPartitionedTable)
+                .filter(DynamicNode::notDynamicOrIsRefreshing)
```
rcaudy, line 35: This utility method doesn't work correctly as-is. PartitionedTable does not implement DynamicNode, because it doesn't make sense to `setRefreshing` or `addParentReference` on it. Maybe we need to factor out something like "MutableDynamicNode", and keep just `isRefreshing()` in `DynamicNode`.

rcaudy: We should ignore this issue and flag it for fixing later.

mofojed (dismissed):
Looks like this works with the following snippet, which was failing before:
```
import deephaven.ui as ui
from deephaven.ui import use_memo


@ui.component
def waves_with_plot():
    tt = time_table("PT1s").update("x=i")
    t = tt.update(["y_sin=Math.sin(x)"])
    p = Figure().plot_xy(series_name="Sine", t=t, x="x", y="y_sin").show()

    return ui.flex(p, flex_grow=1)


wp = waves_with_plot()
```

Plot/src/main/java/io/deephaven/plot/FigureWidget.java
niloc132: ..and have now changed our minds, going to leave it here, introduce a IsLivenessArtifact type, and have FigureWidget implement it.

engine/updategraph/src/main/java/io/deephaven/engine/liveness/IsLivenessNode.java
```diff
@@ -0,0 +1,40 @@
+package io.deephaven.engine.liveness;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.lang.ref.WeakReference;
+import java.util.stream.Stream;
+
+public interface IsLivenessNode extends LivenessNode {
+    LivenessNode asLivenessNode();
```
rcaudy, line 9: Should we have some JavaDoc for the class and the `asLivenessNode` method? Mainly, it should make it clear that these things have tightly coupled lifecycles, and that the `LivenessNode` returned must effectively be a proxy for `this`.

Generators/src/main/java/io/deephaven/plot/util/GenerateFigureImmutable.java
```diff
@@ -65,7 +67,7 @@ private GenerateFigureImmutable(final boolean isInterface, final String[] import
             log.info("Processing class: " + c);
 
             for (final Method m : c.getMethods()) {
-                log.info("Processing method (" + c + "): " + m);
+                log.fine("Processing method (" + c + "): " + m);
```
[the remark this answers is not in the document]
niloc132: Its like "debug" - there are "more spammy" log levels called "finer" and "finest" too, like "trace" or "spam" in other logging systems.

Basically "this is not useful most of the time". 

With that said, I tweaked the wrong thing that was too spammy - the INFO logging isnt even being written most of the time. I'll correct it.

engine/updategraph/src/main/java/io/deephaven/engine/liveness/IsLivenessNode.java
```diff
@@ -5,7 +5,17 @@
...
 
+/**
+ * Indicates that this class implements LivenessNode via a member rather than implementing it directly. The real
+ * LivenessNode is exposed via {@link #asLivenessNode()}, all other methods delegate to this instance.
+ */
 public interface IsLivenessNode extends LivenessNode {
+    /**
+     * Returns the "real" {@link LivenessNode} instance. When implementing this, care should be taken
+     * to match lifecycle of the {@code IsLivenessNode} instance with this instance, as the returned
+     * {@code LivenessNode} is behaves as a proxy for {@code this}.
```
rcaudy, line 16:
```suggestion
     * {@code LivenessNode} behaves as a proxy for {@code this}.
```

engine/updategraph/src/main/java/io/deephaven/engine/liveness/IsLivenessNode.java
```diff
@@ -5,7 +5,17 @@
 import java.lang.ref.WeakReference;
 import java.util.stream.Stream;
 
+/**
+ * Indicates that this class implements LivenessNode via a member rather than implementing it directly. The real
+ * LivenessNode is exposed via {@link #asLivenessNode()}, all other methods delegate to this instance.
+ */
 public interface IsLivenessNode extends LivenessNode {
```
rcaudy, line 12: Maybe rename `DelegatingLivenessNode` or something?

rcaudy (commented):
.

engine/updategraph/src/main/java/io/deephaven/engine/liveness/DelegatingLivenessNode.java
```diff
@@ -9,11 +9,12 @@
...
  * LivenessNode is exposed via {@link #asLivenessNode()}, all other methods delegate to this instance.
  */
-public interface IsLivenessNode extends LivenessNode {
+public interface DelegatingLivenessNode extends LivenessNode {
     /**
-     * Returns the "real" {@link LivenessNode} instance. When implementing this, care should be taken
-     * to match lifecycle of the {@code IsLivenessNode} instance with this instance, as the returned
-     * {@code LivenessNode} is behaves as a proxy for {@code this}.
+     * Returns the "real" {@link LivenessNode} instance. When implementing this, care should be taken to match lifecycle
+     * of the {@code IsLivenessNode} instance with this instance, as the returned {@code LivenessNode} behaves as a
```
rcaudy, line 15:
```suggestion
     * of the {@code DelegatingLivenessNode} instance with this instance, as the returned {@code LivenessNode} behaves as a
```

rcaudy (approved):

PR #4565, opened 2023-09-27
Better failure messages related to LFS

### LINKED ISSUE #4561: Improve error message for Git LFS related failures
As part of #4421, we moved all parquet files from the core repo to Git LFS. So now, on pulling the core repo, these files don't get fetched by default in case Git LFS is not installed on the system. This can lead to errors like the following while running tests locally:
```
    Caused by: io.deephaven.parquet.base.ParquetFileReaderException: extensions/parquet/table/build/resources/test/e1/uncompressed.parquet is not a Parquet file. expected magic number at tail [80, 65, 82, 49] but found [55, 49, 48, 10]
    at app//io.deephaven.parquet.base.ParquetFileReader.<init>(ParquetFileReader.java:62)
    at app//io.deephaven.parquet.table.ParquetTools.getParquetFileReaderChecked(ParquetTools.java:801)
    at app//io.deephaven.parquet.table.ParquetTools.getParquetFileReader(ParquetTools.java:784)
    ... 50 more
```
The fix is to run `git lfs install; git lfs pull` inside the repo.
We should make the error message more clear with the above instructions to fix it. Also, we should update the README to mention setup instructions for Git LFS.

malhotrashivam:
Closes #4561

### DIFF
--- a/extensions/parquet/base/src/main/java/io/deephaven/parquet/base/InvalidParquetFileException.java
+++ b/extensions/parquet/base/src/main/java/io/deephaven/parquet/base/InvalidParquetFileException.java
@@ -1,5 +1,8 @@
 package io.deephaven.parquet.base;
 
+/**
+ * Exception thrown when trying to read an invalid Parquet file.
+ */
 public class InvalidParquetFileException extends ParquetFileReaderException {
     InvalidParquetFileException(String message) {
         super(message);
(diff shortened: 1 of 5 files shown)


rcaudy (dismissed):

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/InvalidParquetFileException.java
```diff
@@ -0,0 +1,7 @@
+package io.deephaven.parquet.base;
+
+public class InvalidParquetFileException extends ParquetFileReaderException {
```
niloc132, line 3: Javadoc? alternatively, change the test to catch any ParquetFileReaderException and just check the `message` to see if it is this kind of failure?

malhotrashivam: I feel that a different exception is less error prone than relying on the message string. So added a Javadoc for now. Hope that's okay.

niloc132 (approved):

chipkent (approved):

PR #4579, opened 2023-10-01
auto convert Python return val to Java val

### LINKED ISSUE #4562: Support numba @guvectorize vector functions in the query language
Numba's `guvectorize` annotation is extremely powerful.  It provides a few key features not found in `vectorize`, which is currently supported by DH.
1. Support for vector inputs
2. Support for vector outputs
3. Support for numpy inputs and operations

The following code illustrates what is possible.  The one notable unsupported feature from the numba side is fixed size outputs (https://github.com/numba/numba/issues/1668).

On the DH side, the query language needs to support outputs from the guvectorize functions.  Currently, all guvectorize output columns are `java.lang.Object` , instead of  being the proper scalar or array type.

Making this fix will allow high performance creation of native java arrays that can be seamlessly be used in operations such as grouping.  Also, the array support makes it easier to directly use inputs as numpy arrays.

```python

import numpy as np
from numba import vectorize, jit, guvectorize, int64

a = np.arange(5, dtype=np.int64)

@jit(int64(int64[:]),nopython=True)
def f(x):
    r = 0
    for xi in x:
        r += xi
    return r


print(f(a))

#vector input to scalar output function (m)->()
@guvectorize([(int64[:],int64[:])],"(m)->()",nopython=True)
def g(x, res):
    res[0] = 0
    for xi in x:
        res[0] += xi

print(g(a))

t = empty_table(10).update(["X=i%3", "Y=i"]).group_by("X").update("Z=g(Y)")
m = t.meta_table

#vector and scalar input to vector ouput function
@guvectorize([(int64[:],int64,int64[:])],"(m),()->(m)",nopython=True)
def g2(x, y, res):
    for i in range(len(x)):
        res[i] = x[i] + y

print(g2(a,2))

t2 = empty_table(10).update(["X=i%3", "Y=i"]).group_by("X").update("Z=g2(Y,2)")
m2 = t2.meta_table

# NOTE: the following does not work according to this thread from 7 years ago:
# https://numba-users.continuum.narkive.com/7OAX8Suv/numba-guvectorize-with-fixed-size-output-array
# but the latest numpy Generalized Universal Function API does seem to support frozen dimensions
# https://numpy.org/doc/stable/reference/c-api/generalized-ufuncs.html#generalized-universal-function-api
# There is an old numba ticket about it
# https://github.com/numba/numba/issues/1668
# Possibly we could contribute a fix

# fails with: bad token in signature "2"

# #vector input to fixed-length vector ouput function
# @guvectorize([(int64[:],int64[:])],"(m)->(2)",nopython=True)
# def g3(x, res):
#     res[0] = min(x)
#     res[1] = max(x)

# print(g3(a))

# t3 = empty_table(10).update(["X=i%3", "Y=i"]).group_by("X").update("Z=g3(Y)")
# m3 = t3.meta_table

# ** Workaround **

dummy = np.array([0, 0], dtype=np.int64)

#vector input to fixed-length vector ouput function -- second arg is a dummy just to get a fixed size output
@guvectorize([(int64[:],int64[:],int64[:])],"(m),(n)->(n)",nopython=True)
def g4(x, dummy, res):
    res[0] = min(x)
    res[1] = max(x)

print(g4(a,dummy))

t4 = empty_table(10).update(["X=i%3", "Y=i"]).group_by("X").update("Z=g4(Y,dummy)")
m4 = t4.meta_table

# example using numpy

#vector input to fixed-length vector ouput function -- second arg is a dummy just to get a fixed size output
@guvectorize([(int64[:],int64[:],int64[:])],"(m),(n)->(n)",nopython=True)
def g5(x, dummy, res):
    res[0] = np.min(x)
    res[1] = np.max(x)

print(g5(a,dummy))

t5 = empty_table(10).update(["X=i%3", "Y=i"]).group_by("X").update("Z=g5(Y,dummy)")
m5 = t5.meta_table

# example using numpy

#vector input to fixed-length vector ouput function 
@guvectorize([(int64[:],int64[:])],"(m)->(m)",nopython=True)
def g6(x, res):
    res[:] = x+5

print(g6(a))

t6 = empty_table(10).update(["X=i%3", "Y=i"]).group_by("X").update("Z=g6(Y)")
m6 = t6.meta_table

```

### LINKED ISSUE #4569: Python type inference is not working
Type inference in python with type hints is not working:

```python

t = time_table("PT1s").update(["X=i%3","Y=i"]).group_by("X")

import numpy as np

def numpy_func(y) -> int:
    return int(np.sum(y))

t2 = t.update("Z=numpy_func(Y)")
```

This returns java.lang.Object for Z.

jmao-denver:
Only support 1d array and scalars for now.

Fixes #4562 
Fixes #4569

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
@@ -20,7 +20,6 @@ import com.github.javaparser.ast.body.VariableDeclarator;
 import com.github.javaparser.ast.comments.BlockComment;
 import com.github.javaparser.ast.comments.JavadocComment;
 import com.github.javaparser.ast.comments.LineComment;
-import com.github.javaparser.ast.comments.Comment;
 import com.github.javaparser.ast.expr.ArrayAccessExpr;
 import com.github.javaparser.ast.expr.ArrayCreationExpr;
 import com.github.javaparser.ast.expr.ArrayInitializerExpr;
@@ -1214,7 +1213,7 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
                     // as a single argument
                     if (ObjectVector.class.isAssignableFrom(expressionTypes[ei])) {
                         expressions[ei] = new CastExpr(
-                                new ClassOrInterfaceType("java.lang.Object"),
+                                StaticJavaParser.parseClassOrInterfaceType("java.lang.Object"),
                                 expressions[ei]);
                         expressionTypes[ei] = Object.class;
                     } else {
@@ -2337,6 +2336,33 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
                 callMethodCall.setData(QueryLanguageParserDataKeys.PY_CALLABLE_DETAILS,
                         new PyCallableDetails(null, methodName));
 
+                if (PyCallableWrapper.class.isAssignableFrom(method.getDeclaringClass())) {
+                    Optional<Class<?>> optionalRetType = pythonFuncReturnType(callMethodCall);
+                    if (optionalRetType.isPresent()) {
+                        Class<?> retType = optionalRetType.get();
+                        final Optional<CastExpr> optionalCastExpr = makeCastExpression(retType, callMethodCall);
+                        if (optionalCastExpr.isPresent()) {
+                            final CastExpr castExpr = optionalCastExpr.get();
+                            replaceChildExpression(
+                                    n.getParentNode().orElseThrow(),
+                                    n,
+                                    castExpr);
+
+                            callMethodCall.getData(QueryLanguageParserDataKeys.PY_CALLABLE_DETAILS).setCasted(true);
+                            try {
+                                return castExpr.accept(this, printer);
+                            } catch (Exception e) {
+                                replaceChildExpression(
+                                        castExpr.getParentNode().orElseThrow(),
+                                        castExpr,
+                                        callMethodCall);
+                                callMethodCall.getData(QueryLanguageParserDataKeys.PY_CALLABLE_DETAILS)
+                                        .setCasted(false);
+                                return callMethodCall.accept(this, printer);
+                            }
+                        }
+                    }
+                }
                 replaceChildExpression(
                         n.getParentNode().orElseThrow(),
                         n,
@@ -2428,6 +2454,52 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
 
         return calculateMethodReturnTypeUsingGenerics(scopeType, n.getScope().orElse(null), method, expressionTypes,
                 typeArguments);
+    }
+
+    private Optional<CastExpr> makeCastExpression(Class<?> retType, MethodCallExpr callMethodCall) {
+        if (retType.isPrimitive()) {
+            return Optional.of(new CastExpr(
+                    new PrimitiveType(PrimitiveType.Primitive
+                            .valueOf(retType.getSimpleName().toUpperCase())),
+                    callMethodCall));
+        } else if (retType.getComponentType() != null) {
+            Class<?> componentType = retType.getComponentType();
+            if (componentType.isPrimitive()) {
+                ArrayType arrayType;
+                if (componentType == boolean.class) {
+                    arrayType = new ArrayType(new ClassOrInterfaceType(null, "java.lang.Boolean"));
+                } else {
+                    arrayType = new ArrayType(new PrimitiveType(PrimitiveType.Primitive
+                            .valueOf(retType.getComponentType().getSimpleName().toUpperCase())));
+                }
+                return Optional.of(new CastExpr(arrayType, callMethodCall));
+            } else if (retType.getComponentType() == String.class || retType.getComponentType() == Boolean.class) {
+                ArrayType arrayType =
+                        new ArrayType(new ClassOrInterfaceType(null, retType.getComponentType().getSimpleName()));
+                return Optional.of(new CastExpr(arrayType, callMethodCall));
+            }
+        } else if (retType == Boolean.class) {
+            return Optional.of(new CastExpr(new ClassOrInterfaceType(null, "java.lang.Boolean"), callMethodCall));
+        } else if (retType == String.class) {
+            return Optional.of(new CastExpr(new ClassOrInterfaceType(null, "java.lang.String"), callMethodCall));
+        }
+        return Optional.empty();
+    }
+
+    private Optional<Class<?>> pythonFuncReturnType(@NotNull MethodCallExpr n) {
+        final PyCallableDetails pyCallableDetails = n.getData(QueryLanguageParserDataKeys.PY_CALLABLE_DETAILS);
+        final String pyMethodName = pyCallableDetails.pythonMethodName;
+        final QueryScope queryScope = ExecutionContext.getContext().getQueryScope();
+        final Object paramValueRaw = queryScope.readParamValue(pyMethodName, null);
+        if (paramValueRaw == null) {
+            return Optional.empty();
+        }
+        if (!(paramValueRaw instanceof PyCallableWrapper)) {
+            return Optional.empty();
+        }
+        final PyCallableWrapper pyCallableWrapper = (PyCallableWrapper) paramValueRaw;
+        pyCallableWrapper.parseSignature();
+        return Optional.ofNullable(pyCallableWrapper.getReturnType());
     }
 
     @NotNull
@@ -2563,13 +2635,27 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
         // expression evaluation code that expects singular values. This check makes sure that numba/dh vectorized
         // functions must be used alone as the entire expression after removing the enclosing parentheses.
         Node n1 = n;
+        boolean autoCastChecked = false;
         while (n1.hasParentNode()) {
             n1 = n1.getParentNode().orElseThrow();
             Class<?> cls = n1.getClass();
 
             if (cls == CastExpr.class) {
-                throw new PythonCallVectorizationFailure(
-                        "The return values of Python vectorized function can't be cast: " + n1);
+                if (!autoCastChecked && n.getData(QueryLanguageParserDataKeys.PY_CALLABLE_DETAILS).isCasted()) {
+                    autoCastChecked = true;
+                } else {
+                    throw new PythonCallVectorizationFailure(
+                            "The return values of Python vectorized function can't be cast: " + n1);
+                }
+            } else if (cls == MethodCallExpr.class) {
+                String methodName = ((MethodCallExpr) n1).getNameAsString();
+                if (!autoCastChecked && n.getData(QueryLanguageParserDataKeys.PY_CALLABLE_DETAILS).isCasted()
+                        && methodName.endsWith("Cast")) {
+                    autoCastChecked = true;
+                } else {
+                    throw new PythonCallVectorizationFailure(
+                            "Python vectorized function can't be used in another expression: " + n1);
+                }
             } else if (cls != EnclosedExpr.class && cls != WrapperNode.class) {
                 throw new PythonCallVectorizationFailure(
                         "Python vectorized function can't be used in another expression: " + n1);
@@ -3232,6 +3318,18 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
         private final String pythonScopeExpr;
         @NotNull
         private final String pythonMethodName;
+
+        @NotNull
+        private boolean isCasted = false;
+
+        public boolean isCasted() {
+            return isCasted;
+        }
+
+        public void setCasted(boolean casted) {
+            isCasted = casted;
+        }
+
 
         private PyCallableDetails(@Nullable String pythonScopeExpr, @NotNull String pythonMethodName) {
             this.pythonScopeExpr = pythonScopeExpr;
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
@@ -2337,10 +2337,11 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
                         new PyCallableDetails(null, methodName));
 
                 if (PyCallableWrapper.class.isAssignableFrom(method.getDeclaringClass())) {
-                    Optional<Class<?>> optionalRetType = pythonFuncReturnType(callMethodCall);
+                    Optional<Class<?>> optionalRetType = pyCallableReturnType(callMethodCall);
                     if (optionalRetType.isPresent()) {
                         Class<?> retType = optionalRetType.get();
-                        final Optional<CastExpr> optionalCastExpr = makeCastExpression(retType, callMethodCall);
+                        final Optional<CastExpr> optionalCastExpr =
+                                makeCastExpressionForPyCallable(retType, callMethodCall);
                         if (optionalCastExpr.isPresent()) {
                             final CastExpr castExpr = optionalCastExpr.get();
                             replaceChildExpression(
@@ -2457,7 +2458,7 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
                 typeArguments);
     }
 
-    private Optional<CastExpr> makeCastExpression(Class<?> retType, MethodCallExpr callMethodCall) {
+    private Optional<CastExpr> makeCastExpressionForPyCallable(Class<?> retType, MethodCallExpr callMethodCall) {
         if (retType.isPrimitive()) {
             return Optional.of(new CastExpr(
                     new PrimitiveType(PrimitiveType.Primitive
@@ -2487,7 +2488,7 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
         return Optional.empty();
     }
 
-    private Optional<Class<?>> pythonFuncReturnType(@NotNull MethodCallExpr n) {
+    private Optional<Class<?>> pyCallableReturnType(@NotNull MethodCallExpr n) {
         final PyCallableDetails pyCallableDetails = n.getData(QueryLanguageParserDataKeys.PY_CALLABLE_DETAILS);
         final String pyMethodName = pyCallableDetails.pythonMethodName;
         final QueryScope queryScope = ExecutionContext.getContext().getQueryScope();
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
@@ -2476,7 +2476,8 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
                             .valueOf(retType.getComponentType().getSimpleName().toUpperCase())));
                 }
                 return Optional.of(new CastExpr(arrayType, callMethodCall));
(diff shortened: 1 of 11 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -2563,13 +2635,27 @@ private void checkVectorizability(@NotNull final MethodCallExpr n,
...
             Class<?> cls = n1.getClass();
 
             if (cls == CastExpr.class) {
-                throw new PythonCallVectorizationFailure(
-                        "The return values of Python vectorized function can't be cast: " + n1);
+                if (!autoCastChecked && n.getData(QueryLanguageParserDataKeys.PY_CALLABLE_DETAILS).isCasted()) {
+                    autoCastChecked = true;
+                } else {
+                    throw new PythonCallVectorizationFailure(
+                            "The return values of Python vectorized function can't be cast: " + n1);
```
chipkent, line 2648:
```suggestion
                            "The return values of Python vectorized functions can't be cast: " + n1);
```

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -2563,13 +2635,27 @@ private void checkVectorizability(@NotNull final MethodCallExpr n,
...
+                }
+            } else if (cls == MethodCallExpr.class) {
+                String methodName = ((MethodCallExpr) n1).getNameAsString();
+                if (!autoCastChecked && n.getData(QueryLanguageParserDataKeys.PY_CALLABLE_DETAILS).isCasted()
+                        && methodName.endsWith("Cast")) {
```
chipkent, line 2653: Is there any chance that a python function with `Cast` at the end of the name will trigger weird behavior?  Does it need to be tested?

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -3233,6 +3319,18 @@ private static class PyCallableDetails {
...
+
+        public boolean isCasted() {
+            return isCasted;
+        }
+
+        public void setCasted(boolean casted) {
+            isCasted = casted;
+        }
+
+
```
chipkent, line 3333:
```suggestion

```
Looks like the standard spacing around this is one line.

engine/table/src/main/java/io/deephaven/engine/util/PyCallableWrapperJpyImpl.java
```diff
@@ -105,9 +118,25 @@ private void prepareSignature() {
...
+            }
+            // numba allows a vectorized function to have multiple signatures
+            if (params.size() > 1) {
+                throw new UnsupportedOperationException(
+                        pyCallable
+                                + " has multiple signatures; this is not currently supported for numba vectorized functions");
+            }
+            signature = params.get(0).getStringValue();
+            unwrapped = pyCallable;
+            numbaVectorized = false;
```
chipkent, line 138: I think possibly `numbaVectorized = true`.

See https://numba.pydata.org/numba-doc/dev/user/vectorize.html#the-guvectorize-decorator.
Look at the example that begins with `The nice thing is that NumPy will automatically dispatch over more complicated inputs, depending on their shapes:`.  In the example, the guvectorize function is `(n),()->(n)`, but it is applied to arrays of other shapes in a vectorized way.

py/server/deephaven/table.py
```diff
@@ -362,9 +366,24 @@ def _j_py_script_session() -> _JPythonScriptSession:
 _numpy_type_codes = ["i", "l", "h", "f", "d", "b", "?", "U", "O"]
 
 
+def _array_component_type(fn) -> str:
+    sig = inspect.signature(fn)
+    t = sig.return_annotation
+    if isinstance(t, _GenericAlias) and issubclass(t.__origin__, Sequence):
+        return np.dtype(t.__args__[0]).char
```
chipkent, line 373: What if the user does not provide an `__args__[0]` value in the type hint?  Does it default to giving `Any`, does it cause a syntax error, or do we get some un-understandable error here?

py/server/deephaven/table.py
```diff
@@ -362,9 +366,24 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+
+    if sys.version_info.minor > 8:
```
chipkent, line 375: what happens if the python version is `<=3.8`?  We just don't do numpy?  If so, add a comment.

py/server/tests/test_pyfunc_return_java_values.py
```diff
@@ -0,0 +1,80 @@
...
+_J_TYPE_NP_DTYPE_MAP = {
+    dtypes.double: "np.float64",
+    dtypes.float32: "np.float32",
+    dtypes.int32: "np.int32",
+    dtypes.long: "np.int64",
+    dtypes.short: "np.int16",
+    dtypes.byte: "np.int8",
+    dtypes.bool_: "np.bool_",
+    dtypes.string: "np.str_",
+    # dtypes.char: "np.uint16",
```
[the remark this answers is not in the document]
jmao-denver: I couldn't find a way to make a Java char/Character from a uint16. So I am not really sure what to do about it.

py/server/tests/test_pyfunc_return_java_values.py
```diff
@@ -0,0 +1,80 @@
...
+                func_str = f"""
+class Foo:
+    def fn(self, col) -> {np_dtype}:
+        return {np_dtype}(col)
+foo = Foo()
+"""
+                exec(func_str, globals())
+
+                t = empty_table(10).update("X = i").update(f"Y= foo.fn(X + 1)")
+                self.assertNotEqual(t.columns[1].data_type, dh_dtype)
```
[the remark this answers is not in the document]
jmao-denver: No. An exception like `java.lang.ClassCastException: class org.jpy.PyObject cannot be cast to class java.lang.Boolean ` will be thrown and the query fails.

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
jmao-denver: I don't think so. The `isCasted()` condition guarantees that it is the artificially inserted the `...Cast` method that we are checking.

engine/table/src/main/java/io/deephaven/engine/util/PyCallableWrapperJpyImpl.java
jmao-denver: It would be nice to support that. But the vectorization implementation currently doesn't support non scalar types.

py/server/deephaven/table.py
jmao-denver: if no `__args__` (meaning just `List`, not `List[float]`) is provided, then `isinstance(t, _GenericAlias)` would fail first.

py/server/deephaven/table.py
```diff
@@ -469,6 +489,93 @@ def _query_scope_ctx():
...
+
+    @wraps(fn)
+    def wrapper(*args, **kwargs):
+        ret = fn(*args, **kwargs)
+        if return_array:
+            return dtypes.array(dh_dtype, ret)
+        elif dh_dtype == dtypes.PyObject:
+            return ret
+        else:
+            return _from_numpy_scalar_to_py(ret)
```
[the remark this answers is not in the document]
jmao-denver: The existing vectorization unit tests verified  that it didn't break vectorization.

py/server/tests/test_pyfunc_return_java_values.py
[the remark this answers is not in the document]
chipkent: You couldn't find a way to test it or couldn't find a way to make it work?

engine/table/src/main/java/io/deephaven/engine/util/PyCallableWrapperJpyImpl.java
chipkent: Should this become a ticket and item on our project plan so that we don't try to tackle it in this PR?

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -2430,6 +2458,52 @@ public Class<?> visit(MethodCallExpr n, VisitArgs printer) {
                 typeArguments);
     }
 
+    private Optional<CastExpr> makeCastExpressionForPyCallable(Class<?> retType, MethodCallExpr callMethodCall) {
```
chipkent, line 2461: Looking through this list, I see that timestamps are missing. Are there any timestamp conversions that we should be transparently handling? Probably, but it may not be easy. I don't know.

py/server/tests/test_pyfunc_return_java_values.py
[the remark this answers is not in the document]
jmao-denver: I couldn't find a way to make it work nor does JPY, so the value comes to Java as a `org.jpy.PyObject` object.

py/server/deephaven/table.py
```diff
@@ -469,6 +495,93 @@ def _query_scope_ctx():
...
+        elif x.dtype.char == '?':
+            return bool(x)
+        elif x.dtype.char == 'U':
+            return str(x)
+        elif x.dtype.char == 'O':
+            return x
+        else:
+            raise TypeError(f"Unsupported dtype: {x.dtype}")
+    else:
+        return x
```
chipkent, line 518: Should this go in `dtypes`?

py/server/deephaven/table.py
```diff
@@ -469,6 +495,93 @@ def _query_scope_ctx():
...
+    """A decorator that acts as a transparent translator for Python UDFs used in Deephaven query formulas between
+    Python and Java. This decorator is intended for use by the Deephaven query engine and should not be used by
+    users.
+
+    For now, this decorator is only capable of converting Python function return values to Java values. It
+    does not yet convert Java values in arguments to usable Python object (e.g. numpy arrays) or properly translate
+    Deephaven primitive null values.
+
+    For properly annotated functions, including numba vectorized and guvectorized ones, this decorator inspects the
+    signature of the function and determines its return type including supported primitive types and arrays of
```
chipkent, line 531:
```suggestion
    signature of the function and determines its return type, including supported primitive types and arrays of
```

py/server/tests/test_pyfunc_return_java_values.py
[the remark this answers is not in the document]
chipkent: Not critical.

py/server/tests/test_pyfunc_return_java_values.py
```diff
@@ -0,0 +1,80 @@
...
+            "float": dtypes.double_array,
+            "np.int8": dtypes.byte_array,
+            "np.int16": dtypes.short_array,
+            "np.int32": dtypes.int32_array,
+            "np.int64": dtypes.long_array,
+            "np.float32": dtypes.float32_array,
+            "np.float64": dtypes.double_array,
+            "bool": dtypes.boolean_array,
+            "np.str_": dtypes.string_array,
+            # "np.uint16": dtypes.char_array,
```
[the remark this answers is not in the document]
chipkent: Not critical.

py/server/deephaven/dtypes.py
```diff
@@ -204,6 +207,33 @@ def null_remap(dtype: DType) -> Callable[[Any], Any]:
     return lambda v: null_value if v is None else v
 
 
+def _instant_array(data: Sequence) -> jpy.JType:
+    """Converts a sequence of either datetime64[ns], datetime.datetime, pandas.Timestamp, datetime strings,
+    integers in nanoseconds, to a Java array of Instant values. """
```
chipkent, line 212:
```suggestion
    or integers in nanoseconds, to a Java array of Instant values. """
```

py/server/deephaven/dtypes.py
```diff
@@ -286,3 +315,60 @@ def from_np_dtype(np_dtype: Union[np.dtype, pd.api.extensions.ExtensionDtype]) -
...
+        elif x.dtype.char in _numpy_floating_type_codes:
+            return float(x)
+        elif x.dtype.char == '?':
+            return bool(x)
+        elif x.dtype.char == 'U':
+            return str(x)
+        elif x.dtype.char == 'O':
+            return x
+        elif x.dtype.char == 'M':
+            from deephaven.time import to_j_instant
```
chipkent, line 338: Any reason this can't be at the top level?  If so, does it make things slow or is this optimized in python?

py/server/tests/test_pyfunc_return_java_values.py
```diff
@@ -75,6 +77,32 @@ def fn(self, col) -> {np_dtype}:
...
+        for np_dtype in dt_dtypes:
+            with self.subTest(np_dtype=np_dtype):
+                func_decl_str = f"""def fn(col) -> {np_dtype}:"""
+                if np_dtype == "np.dtype('datetime64[ns]')":
+                    func_body_str = f"""    return pd.Timestamp(col).to_numpy()"""
+                elif np_dtype == "datetime.datetime":
+                    func_body_str = f"""    return pd.Timestamp(col).to_pydatetime()"""
+                elif np_dtype == "pd.Timestamp":
+                    func_body_str = f"""    return pd.Timestamp(col)"""
+                func_body_str = f"""    return pd.Timestamp(col).to_numpy()"""
```
chipkent, line 97: test looks broken.  This looks like it overrides other cases..

py/server/deephaven/dtypes.py
jmao-denver: That would lead to circular imports.

py/server/tests/test_pyfunc_return_java_values.py
jmao-denver: Ah, I wasn't too careful with co-pilot.

rcaudy (commented):
In progress review from this morning.

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -2430,6 +2459,56 @@ public Class<?> visit(MethodCallExpr n, VisitArgs printer) {
...
+            return Optional.of(new CastExpr(
+                    new PrimitiveType(PrimitiveType.Primitive
+                            .valueOf(retType.getSimpleName().toUpperCase())),
+                    callMethodCall));
+        } else if (retType.getComponentType() != null) {
+            Class<?> componentType = retType.getComponentType();
+            if (componentType.isPrimitive()) {
+                ArrayType arrayType;
+                if (componentType == boolean.class) {
+                    arrayType = new ArrayType(new ClassOrInterfaceType(null, "java.lang.Boolean"));
```
rcaudy, line 2473: Should this use `StaticJavaParser.parseClassOrInterfaceType`? Same for other usages below.

engine/table/src/main/java/io/deephaven/engine/util/PyCallableWrapperJpyImpl.java
```diff
@@ -159,12 +183,17 @@ public void parseSignature() {
         }
 
         this.paramTypes = paramTypes;
+
+        returnType = (Class<?>) pyUdfDecoratedCallable.getAttribute("return_type", null);
+        if (returnType == null) {
```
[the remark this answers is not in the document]
jmao-denver: We always call `_py_udf` on a callable when this method is called, and  a return type is guaranteed to bet set.

engine/table/src/main/java/io/deephaven/engine/util/PyCallableWrapperJpyImpl.java
```diff
@@ -159,12 +183,17 @@ public void parseSignature() {
         }
 
         this.paramTypes = paramTypes;
+
+        returnType = (Class<?>) pyUdfDecoratedCallable.getAttribute("return_type", null);
```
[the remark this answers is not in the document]
jmao-denver: Because the `_py_udf` decorator returns the precise type in the case of return type is annotated as an array or other container types (list, tuple), not simply `PyObject`. As we are asked to support nD arrays (Avi), this is the direction we should go.  In the cases of return type being primitive, the two decorators get the same return type.

After this change, the `signature` field isn't relied upon for finding the return type but is still used for other purposes.

engine/table/src/main/java/io/deephaven/engine/util/PyCallableWrapperJpyImpl.java
```diff
@@ -105,9 +120,25 @@ private void prepareSignature() {
...
+            if (params.size() > 1) {
+                throw new UnsupportedOperationException(
+                        pyCallable
+                                + " has multiple signatures; this is not currently supported for numba vectorized functions");
+            }
+            signature = params.get(0).getStringValue();
+            unwrapped = pyCallable;
+            numbaVectorized = false;
+            vectorized = false;
         } else if (pyCallable.hasAttribute("dh_vectorized")) {
```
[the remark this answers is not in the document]
jmao-denver: The vectorized version will be used if the function call is vectorize-able.

py/server/tests/test_numba_guvectorize.py
```diff
@@ -0,0 +1,94 @@
...
+        self.assertEqual(t.columns[2].data_type, dtypes.long_array)
+
+    def test_fixed_length_vector_return(self):
+        # NOTE: the following does not work according to this thread from 7 years ago:
+        # https://numba-users.continuum.narkive.com/7OAX8Suv/numba-guvectorize-with-fixed-size-output-array
+        # but the latest numpy Generalized Universal Function API does seem to support frozen dimensions
+        # https://numpy.org/doc/stable/reference/c-api/generalized-ufuncs.html#generalized-universal-function-api
+        # There is an old numba ticket about it
+        # https://github.com/numba/numba/issues/1668
+        # Possibly we could contribute a fix
```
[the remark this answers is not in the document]
jmao-denver: #4605

py/server/tests/test_pyfunc_return_java_values.py
```diff
@@ -103,6 +105,62 @@ def test_datetime_scalar_return(self):
...
+            "np_array",
+        ]
+
+        for np_dtype in dt_dtypes:
+            for data in dt_data:
+                with self.subTest(np_dtype=np_dtype, data=data):
+                    func_decl_str = f"""def fn(col) -> {np_dtype}:"""
+                    # func_body_str_1 = f"""    nonlocal {data}"""
+                    func_body_str_1 = """"""
+                    func_body_str_2 = f"""    return {data}"""
```
chipkent, line 135: 1) Is the commented line needed?
2) This test is funky.  I'm not sure what it is trying to do.  For example, the value actually returned is either a list of mixed types or a numpy array of datetimes. ... but the type hint matches the return value in only 1 of 6 cases.  If this is intentional, there needs to be a better comment about the intention of the test and what is going on.

chipkent: Additionally, this weird test points out that we can potentially convert a union type like:
```def fn(col) -> List[Union[datetime.datetime,pd.Timestamp,np.datetime64]]`
Does this work?  Should we try to support it?  Should we test it?

I'm guessing that the guts work but there are a few places that would need a change.

rcaudy (commented):
.

py/server/tests/test_pyfunc_return_java_values.py
jmao-denver: I am not sure if we want to support the many combinations of type hints that certainly are valid but hard to cover completely. I don't think it is an unreasonable requirement for users to provide a more straightforward typehint for the return type, which is more Pythonic anyways. We currently parse it as just the generic object. I have added a test case .

chipkent (approved):

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3278
Reference: https://github.com/deephaven/deephaven.io/issues/3279

PR #4588, opened 2023-10-02
A few small cleanups to ring table and snapshot

devinrsmith:


engine/table/src/main/java/io/deephaven/engine/table/impl/remote/ConstructSnapshot.java
```diff
@@ -681,7 +678,7 @@ public static BarrageMessage constructBackplaneSnapshotInPositionSpace(
                 if (positionsToSnapshot == null && reversePositionsToSnapshot == null) {
                     keysToSnapshot = null;
                 } else {
-                    final RowSet rowSetToUse = usePrev ? table.getRowSet().copyPrev() : table.getRowSet();
+                    final RowSet rowSetToUse = usePrev ? table.getRowSet().prev() : table.getRowSet();
```
rcaudy, line 681: Subtle, but you've introduced a bug here. Look at the line below: if we're using prev, we'll close the row set, which we do not own since you haven't copied it. This can be remedied by removing the try-with-resources.

rcaudy (approved):

PR #4596, opened 2023-10-04
Enhance Groovy classloader experience

niloc132:
Introduces new features when using Groovy as a console language. This
should not remove or change existing features at all, but only serves
to backfill the removed source() and sourceOnce() functions with an
approach that encourages users to write real groovy scripts.

Makes it possible to load .groovy files from the classpath (or a custom
classloader) as classes. Standalone groovy scripts provide a static
main() method (and an instance run() method) to invoke one or more
times, and classes can be reimported to read fresh from disk. Includes
a feature where groovy files loaded from the classpath can use
Deephaven default imports, and can inherit imports from the console,
both disabled by default.

Replaces a list of import statements with the Groovy ImportCustomizer,
so that evaluated code doesn't have a variable number of prefix lines.
This has the advantage of also being applicable to scripts loaded from
the classpath.

Restored unit tests lost when migrating to the deephaven-core codebase,
which helped to validate some of these cases, and verify
console<->querylanguage classloader interactions.


engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -270,9 +366,9 @@ private Exception maybeRewriteStackTrace(String scriptName, String currentScript
         return e;
     }
 
-    private static Class<?> loadClass(String className) throws ClassNotFoundException {
+    private Class<?> loadClass(String className) throws ClassNotFoundException {
         try {
-            return Class.forName(className, false, GroovyDeephavenSession.class.getClassLoader());
+            return Class.forName(className, false, this.groovyShell.getClassLoader());
```
niloc132, line 371: As above, a custom classloader installed by the app before calling this wouldn't necessarily be called when a class needs to be loaded in this way. Between this and the above change, not only does groovy get a chance to compile if needed, but a custom classloader can also be used.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -288,7 +384,7 @@ private static Class<?> loadClass(String className) throws ClassNotFoundExceptio
         }
     }
 
-    private static boolean classExists(String className) {
+    private boolean classExists(String className) {
```
niloc132, line 387: Reviewer note: these could be made static again if the classloader was passed as an argument, it is needed in `loadClass`.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -345,6 +441,15 @@ public static String removeComments(String s) {
         return s.trim();
     }
 
+    /**
+     * Represents an import that can be added to an ImportCustomizer, as a valid return from
+     * {@link #isValidImportString(String)}.
+     */
+    @VisibleForTesting
+    public interface GroovyImport {
+        void append(ImportCustomizer imports);
```
devinrsmith, line 450: I was confused at first trying to figure out how `isValidImportString` works now. I think the method name `appendTo` makes more sense?

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -145,6 +192,55 @@ public GroovyDeephavenSession(
         }
     }
 
+    /**
+     * Adds the default imports that Groovy users assume to be present.
+     */
+    private void addDefaultImports(ImportCustomizer imports) {
+        // TODO (core#230): Remove large list of manual text-based consoleImports
+        // NOTE: Don't add to this list without a compelling reason!!! Use the user script import if possible.
+        imports.addImports(
```
niloc132, line 201: Reviewer note: I made this a list of class literals to assist in future refactoring, but can revert to just putting a list of strings for each of these.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -354,48 +459,74 @@ public static String removeComments(String s) {
      *         package.class.part.part[.*];"
      */
     @VisibleForTesting
-    public static String isValidImportString(Logger log, String importString) {
+    public Optional<GroovyImport> isValidImportString(String importString) {
```
niloc132, line 462: rename to validateImport or createImport?

Also, I left the structure of this method intact, but it could be made to have several returns instead of just one at the end. However, that would remove the `INFO` logging that happens whenever a new import is added... I'm not sure why that exists, but it must be important if we always log it?

engine/table/src/test/java/io/deephaven/engine/table/impl/select/TestConditionFilterGeneration.java
```diff
@@ -31,7 +32,11 @@ public void setUp() {
                 .captureQueryScope()
                 .captureUpdateGraph()
                 .build().open();
+    }
 
+    @After
+    public void tearDown() {
+        executionContext.close();
```
niloc132, line 39: I'm not sure how we got away with not closing the exec context before, are we possibly just opening other context on top of it and never checking that none were leaked?

engine/table/src/test/java/io/deephaven/engine/util/scripts/TestGroovyDeephavenSession.java
```diff
@@ -104,5 +124,486 @@ public void testUnloadedWildcardPackageImport() {
...
+        }
+    }
+
+
+    @Test
+    public void testMinInFormula() {
+        QueryScope.addParam("d", 5d);
+        session.evaluateScript("t = emptyTable(1).updateView(\"Y=1\", \"Z=min(Y,d)\")\n").throwIfError();
+        final Table t = (Table) session.getVariable("t");
+    }
```
devinrsmith, line 607: Do import aliases work now? Should we have a test for that?

engine/table/src/test/java/io/deephaven/engine/util/scripts/TestGroovyDeephavenSession.java
```diff
@@ -104,5 +124,486 @@ public void testUnloadedWildcardPackageImport() {
...
+                "static import io.deephaven.engine.util.TableTools", // static before import
+                "import // io.deephaven.engine.util.TableTools", // invalid after stripped comments
+                "import io.deephaven.engine.util.TableTools.", // trailing .
+                "import io.deephaven.engine.util.TableTools.**", // invalid wildcard
+                "import io.deephaven.engine.util.TableTools.*.*", // doubled wildcard
+                "import io.deephaven.engine. util.TableTools; // has a space", // internal space
+                "import io.deephaven.engine.xxx.util.TableTools", // nonexistent package
+                "import io.deephaven.engine.util.TablexxxTools", // nonexistant classname
+                "import static io.deephaven.engine.util.scripts.TestGroovyDeephavenSession.StaticClassstatic ", // includes
+                                                                                                                // "static
```
niloc132, line 250: Should update these comments (copied with the tests from dhe) so they appear on one line

engine/table/src/test/java/io/deephaven/engine/util/scripts/TestGroovyDeephavenSession.java
niloc132: Yes, they do, will add a test.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
niloc132: Agreed, just posted my own naming thoughts for the isValidImportString method itself, will change this as well.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -145,6 +192,55 @@ public GroovyDeephavenSession(
...
+                TableLoggers.class.getName(),
+                QueryConstants.class.getName(),
+                GroovyStaticImports.class.getName(),
+                DateTimeUtils.class.getName(),
+                QueryLanguageFunctionUtils.class.getName(),
+                Aggregation.class.getName(),
+                UpdateByOperation.class.getName(),
+                OperationControl.class.getName(),
+                DeltaControl.class.getName(),
+                BadDataBehavior.class.getName());
```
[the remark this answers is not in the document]
niloc132: They were simple imports before, same as now, the difference is that we don't rewrite your groovy string into a much longer string, but just tell groovy to behave as if these were already imported.

The only change (can revert and land separately) is that ExecutionContext is now in the list.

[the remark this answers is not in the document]
devinrsmith: Just to be a bit more explicit: I would have thought these would be under `addImports`, but now they are under `addStaticStars`. You are saying this _is_ correct?

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -145,6 +192,55 @@ public GroovyDeephavenSession(
...
     }
 
+    /**
+     * Adds the default imports that Groovy users assume to be present.
+     */
+    private void addDefaultImports(ImportCustomizer imports) {
+        // TODO (core#230): Remove large list of manual text-based consoleImports
+        // NOTE: Don't add to this list without a compelling reason!!! Use the user script import if possible.
+        imports.addImports(
+                DataColumn.class.getName(),
```
rcaudy, line 202: Remove `DataColumn`, please. Add `ColumnSource`, `RowSet`, and `TrackingRowSet`, maybe.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -145,6 +192,55 @@ public GroovyDeephavenSession(
...
+                Table.class.getName(),
+                TableFactory.class.getName(),
+                PartitionedTable.class.getName(),
+                PartitionedTableFactory.class.getName(),
+                Array.class.getName(),
+                TypeUtils.class.getName(),
+                ArrayTypeUtils.class.getName(),
+                DateTimeUtils.class.getName(),
+                CompressedString.class.getName(),
```
rcaudy, line 211: You might as well remove this. It's an old, failed experiment. We should probably let only enterprise have that code.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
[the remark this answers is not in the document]
niloc132: Thanks, fixed those three on the wrong list, and audited the rest (found one missing).

niloc132:
Here's an extremely hacky example classloader that produces groovy files on the fly, which could easily be a blocking call to read network/table data. Set this up before DHC has started, so that it will be used to load groovy files as requested by the DHE groovy IDE.

```java
        final Path dir = Files.createTempDirectory("dhc");

        Thread.currentThread().setContextClassLoader(new ClassLoader(Thread.currentThread().getContextClassLoader()) {
            @Nullable
            @Override
            public URL getResource(String name) {
                // if this is a groovy file, pretend to load contents remotely, and return a file: URL
                // to it, otherwise groovy won't parse it.
                if (name.startsWith("notebook/")) {// be sure not to take over deephaven/groovy/java files
                    Path path = Path.of(name);
                    String packageString = StreamSupport.stream(path.getParent().spliterator(), false)
                            .map(Path::toString).collect(Collectors.joining("."));
                    Path fileName = path.getParent().relativize(path);
                    if (fileName.endsWith("Script.groovy")) {// this check doesn't make sense for a real use case, but
                                                             // we're just going to make a fake file
                        Path result = dir.resolve(path);
                        try {
                            if (!Files.exists(result)) {
                                Files.createDirectories(result.getParent());
                                System.out.println(result);
                                Files.writeString(result, "package " + packageString + "\nprintln 'hello, world'\n");
                            }
                            return result.toFile().toURL();
                        } catch (IOException e) {
                            // instead of failing, possibly log, then pass the buck to the next link in the chain
                        }
                    }
                }

                return super.getResource(name);
            }
        });
```

engine/table/src/test/java/io/deephaven/engine/util/scripts/TestGroovyDeephavenSession.java
```diff
@@ -104,5 +136,497 @@ public void testUnloadedWildcardPackageImport() {
...
+            final double var2 = t.getColumnSource("Z").getDouble(0);
+            assertEquals(Numeric.max(Y, 5d), var2, 1e-10);
+        } catch (Exception e) {
+            e.printStackTrace();
+            fail("Fail for : \n" + c);
+        }
+
+        c = "t = emptyTable(1).updateView(\"Y=" + Y + "\", \"Z=max(Y, d)\")\n";
+        try {
+            QueryScope.addParam("z", z);
```
nbauernfeind, line 549: I don't understand why you're `addParam` on each test - in this particular case you're setting `z` but looks unused. Why not set all parameters once at the top of the method?

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -405,33 +547,38 @@ public static String isValidImportString(Logger log, String importString) {
...
                     } else {
                         log.warn().append("Package or class \"").append(body)
                                 .append("\" could not be verified.")
                                 .endl();
+                        return Optional.empty();
                     }
                 }
             } else {
-                okToImport = classExists(body);
+                if (classExists(body)) {
```
nbauernfeind, line 565: I'd collapse this with the line above to reduce a level of indentation:
```java
} else if (classExists(body)) {
// ...
} else {
// ...
}
```

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -354,48 +464,80 @@ public static String removeComments(String s) {
...
                     okToImport = classExists(body);
+                    typeName = body;
+                    memberName = null;
                 }
             }
+            if (okToImport) {
+                if (isWildcard) {
+                    result = Optional.of(imports -> imports.addStaticStars(body));
+                } else {
+                    if (alias != null) {
```
nbauernfeind, line 523: I would collapse this `if` with `else` above.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
nbauernfeind: You could introduce a helper method to invoke getName automatically, but this doesn't look like it will be gross to maintain as is =P.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -354,48 +464,80 @@ public static String removeComments(String s) {
      *         package.class.part.part[.*];"
      */
     @VisibleForTesting
-    public static String isValidImportString(Logger log, String importString) {
+    public Optional<GroovyImport> createImport(String importString) {
         // look for (ignoring whitespace): optional "import" optional "static" everything_else optional ".*" optional
+        // "as" optional ".*" optional
```
rbasralian, line 469: this comment looks like it might have gotten formatted strangely by spotless (maybe needs `@formatter:off`/`@formatter:on`?)

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -96,22 +131,15 @@ protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundE
 
     private final ScriptFinder scriptFinder;
 
-    private final ArrayList<String> scriptImports = new ArrayList<>();
+    private final ImportCustomizer consoleImports = new ImportCustomizer();
```
rbasralian, line 134: Just to be extra clear, could you add javadoc comments to these saying what they are? (`consoleImports` is imports to add to every command in the interactive console, and `loadedGroovyScriptImports` are imports to apply to any script being ~'source()d'~ loaded?)

engine/table/src/test/java/io/deephaven/engine/util/scripts/TestGroovyDeephavenSession.java
```diff
@@ -104,5 +136,497 @@ public void testUnloadedWildcardPackageImport() {
...
+                // missing spaces
+                "importstaticio.deephaven.engine.util.scripts.TestGroovyDeephavenSession;",
+                // missing spaces
+                "importstatic io.deephaven.engine.util.scripts.TestGroovyDeephavenSession;",
+                // missing spaces
+                "import staticio.deephaven.engine.util.scripts.TestGroovyDeephavenSession;",
+                // import is not optional, ; is optional;
+                "static    io.deephaven.engine.util.scripts.TestGroovyDeephavenSession.StaticClass",
+                // groovy is not cool with non-star package imports for packages that don't exist
+                "import com.illumon.foo.bar;",
```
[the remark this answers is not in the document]
niloc132: Sure, but the point is that the package _does not exist_...

engine/table/src/test/java/io/deephaven/engine/util/scripts/TestGroovyDeephavenSession.java
niloc132: Agreed, this seems a bit off, but is part of the original test from DHE. I tried to keep my changes here limited, ensuring that except where I explicitly changed something, the same things that pass for DHE passes in DHC.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
niloc132: Or drop the else entirely and just return once? no opinions offered either way, so I'm going to apply that change and your suggestions...

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -145,6 +196,56 @@ public GroovyDeephavenSession(
         }
     }
 
+    /**
+     * Adds the default imports that Groovy users assume to be present.
+     */
+    private void addDefaultImports(ImportCustomizer imports) {
+        // TODO (core#230): Remove large list of manual text-based consoleImports
```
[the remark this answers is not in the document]
niloc132: No, just moves it to here rather than a config entry.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -354,48 +464,80 @@ public static String removeComments(String s) {
...
+        @Nullable
+        final String alias = matcher.group("alias");
         if (body == null) {
-            return null;
+            return Optional.empty();
         }
+        final Optional<GroovyImport> result;
 
         boolean okToImport;
         if (isStatic) {
```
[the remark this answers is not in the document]
niloc132: Well it is slightly smaller than before, and from Nate's feedback I'll shrink it slightly more - take a look on the next rev?

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
niloc132: Aside from the line break, that's pretty much what it should say I think? The "as" syntax is new, was previously missing in DHE/DHC

engine/table/build.gradle
```diff
@@ -68,6 +68,7 @@ dependencies {
     testImplementation project(':extensions-parquet-table')
     testImplementation project(':extensions-source-support')
     testImplementation project(':Numerics')
+    testImplementation project(':extensions-suanshu')
```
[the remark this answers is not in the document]
niloc132: Added for test classpaths only, as an example used in DHE tests for groovy parameter disambiguation hit one of these methods.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
niloc132: Correct, added javadoc.

engine/table/src/test/java/io/deephaven/engine/util/scripts/TestGroovyDeephavenSession.java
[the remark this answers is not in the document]
nbauernfeind: Technically, `io.deephaven.foo.bar` does not exist. I'm just considering all of the effort we went through to remove references to illumon and db, for example.

devinrsmith (commented):
LGTM, leaving approval for others.

nbauernfeind (approved):

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -62,18 +90,25 @@ public class GroovyDeephavenSession extends AbstractScriptSession<GroovySnapshot
     private static final String DEFAULT_SCRIPT_PATH = Configuration.getInstance()
             .getStringWithDefault("GroovyDeephavenSession.defaultScriptPath", ".");
 
+    private static final boolean INCLUDE_DEFAULT_IMPORTS_IN_LOADED_GROOVY =
+            Configuration.getInstance()
+                    .getBooleanWithDefault("GroovyDeephavenSession.includeDefaultImportsInGroovyScripts", false);
+    private static final boolean INCLUDE_CONSOLE_IMPORTS_IN_LOADED_GROOVY =
+            INCLUDE_DEFAULT_IMPORTS_IN_LOADED_GROOVY && Configuration.getInstance()
```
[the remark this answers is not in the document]
niloc132: Perhaps - I wrote it this way under the assumption that a user who expects console imports to work in scripts would want that for the purpose of basically authoring notebooks (copied contents from the console), rather than "actually that script is missing a single import, so I'll run it manually."

That is, when would a user _not_ expect to need an import for `com.company.project.MyClass` if they had already used it once in the console, but _expect_ to write imports for TableTools etc (which are in the default imports list).

rbasralian (dismissed):

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
```diff
@@ -62,18 +90,25 @@ public class GroovyDeephavenSession extends AbstractScriptSession<GroovySnapshot
     private static final String DEFAULT_SCRIPT_PATH = Configuration.getInstance()
             .getStringWithDefault("GroovyDeephavenSession.defaultScriptPath", ".");
 
+    private static final boolean INCLUDE_DEFAULT_IMPORTS_IN_LOADED_GROOVY =
+            Configuration.getInstance()
+                    .getBooleanWithDefault("GroovyDeephavenSession.includeDefaultImportsInGroovyScripts", false);
+    private static final boolean INCLUDE_CONSOLE_IMPORTS_IN_LOADED_GROOVY =
```
arman-ddl, line 96: I view INCLUDE_CONSOLE_IMPORTS_IN_LOADED_GROOVY as having close to no (clear) value. To have effect, there would need to be scripts that rely on (or at least can be affected by) imports from the console, which is obviously not useful for prod, but also not useful for developing prod queries.

Maybe it can help some unlikely and/or fringe cases, but we might even purposefully not want to support those to advocate against such tight coupling in our system.

I'm requesting the removal of this prop on initial release. And if it does end up being wanted by a user later on, we can consider adding then.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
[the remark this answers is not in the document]
arman-ddl: Discussed with Colin off GH, which resulted in my request at https://github.com/deephaven/deephaven-core/pull/4596#discussion_r1361051744.

engine/table/src/main/java/io/deephaven/engine/util/GroovyDeephavenSession.java
niloc132: @rbasralian any objection? this will be easy to revert if we decide we need it. When gathering requirements from Arman to begin with, I had mistakenly added this as we reviewed what other features might matter, and I think I agree that the utility here is very limited.

rbasralian: I'm ok with that.

The benefit of this is that it lets you write scripts scripts (e.g. different notebook) without having to worry about copying the same imports to every single one, which is just a hassle to do, especially in the web UI where you can't open local scripts like you could in Swing (so you could just do all editing in IntelliJ and let it worry about the imports). But I agree that it's not the best practice, and now that we load Groovy scripts properly we should push users toward handling imports properly.

arman-ddl (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3304
Reference: https://github.com/deephaven/deephaven.io/issues/3303

PR #4598, opened 2023-10-04
Make newTable, emptyTable flat. Fix flatten.

devinrsmith:
Originally reported by @elijahpetty. Noticed that

```groovy
t1 = newTable(
    stringCol("Title", "content"),
    stringCol("ColumnName", "column_content")
)

println t1.isFlat()

t1.flatten()

println t1.isFlat()
```

printed out `false` then `true`. This change makes sure that `newTable` and `emptyTable` produces an `isFlat()` Table out of the box, as well as ensuring that `Table.flatten` doesn't modify the existing table.

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
@@ -1387,18 +1387,17 @@ public class QueryTable extends BaseTable<QueryTable> {
 
     @Override
     public Table flatten() {
+        if (!flat && !isRefreshing() && rowSet.isFlat()) {
+            // We're already flat, and we'll never update; so we can just return a flat copy
+            final QueryTable copyWithFlat = copy();
+            copyWithFlat.setFlat();
+            return copyWithFlat;
+        }
         final UpdateGraph updateGraph = getUpdateGraph();
         try (final SafeCloseable ignored = ExecutionContext.getContext().withUpdateGraph(updateGraph).open()) {
-            if (!isFlat() && !isRefreshing() && rowSet.size() - 1 == rowSet.lastRowKey()) {
-                // We're already flat, and we'll never update; so we can just return ourselves, after setting ourselves
-                // flat
-                setFlat();
-            }
-
             if (isFlat()) {
                 return prepareReturnThis();
             }
-
             return getResult(new FlattenOperation(this));
         }
     }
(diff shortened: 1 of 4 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -1387,18 +1387,17 @@ private Table whereInInternal(final Table rightTable, final boolean inclusion,
 
     @Override
     public Table flatten() {
+        if (!flat && !isRefreshing() && rowSet.isFlat()) {
+            // We're already flat, and we'll never update; so we can just return a flat copy
+            final QueryTable copyWithFlat = copy();
+            copyWithFlat.setFlat();
+            return copyWithFlat;
+        }
```
rcaudy, line 1395: Move this into the try-with-resources, so the copy picks up the correct update graph (even if it doesn't really matter for static tables).

engine/table/src/main/java/io/deephaven/engine/util/TableTools.java
```diff
@@ -665,8 +665,9 @@ public static ColumnHolder<Float> floatCol(String name, float... data) {
      * @return a Deephaven Table with no columns.
      */
     public static Table emptyTable(long size) {
-        return new QueryTable(RowSetFactory.flat(size).toTracking(),
-                Collections.emptyMap());
+        final QueryTable result = new QueryTable(RowSetFactory.flat(size).toTracking(), Collections.emptyMap());
+        result.setFlat();
+        return result;
```
rcaudy, line 670: This doesn't matter, but sometimes what we do to avoid the intermediate variables looks like:
```
    return new QueryTable(rowSet, visibleSources) {
            {
                setFlat();
            }
```

rcaudy (commented):
.

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
devinrsmith: `copy()` uses its own update graph / execution context setup; that should cover us here? I can still move it into the execution context if we feel like it more correct that way...

rcaudy: Yes, you're good.

rcaudy (approved):

PR #4610, opened 2023-10-09
Documentation for the R Client

alexpeters1208:
This client introduces documentation pages generated from existing docstrings that will be accessible from within RStudio. All documentation will have working examples, though these are not needed to assess language or formatting of the docs. RStudio is needed to get the full perspective on these docs.

Here are formatted doc previews, saved directly from R studio, as of 10/11:
Introduction to the package and important concepts: [rdeephaven.Rd](https://drive.google.com/file/d/1nsmnHgA6xuWQ5gSKU7ptBBHZp9PNJMlf/view?usp=share_link)
Reference doc on the Client class: [Client.Rd](https://drive.google.com/file/d/1v_LodO1HYV8Y8jwuvH25uzdYtE2eQx-T/view?usp=share_link)
Reference doc on the TableHandle class: [TableHandle.Rd](https://drive.google.com/file/d/1xHxobKmOMEw9yCSe7eYA2gsXTHymhkRo/view?usp=share_link)
Intro to AggBy: [AggBy.Rd](https://drive.google.com/file/d/1g1ybuMJ0M51169RXLKIPEMt4pJPhF2Jk/view?usp=share_link)
Intro to UpdateBy: [UpdateBy.Rd](https://drive.google.com/file/d/1fRifmCnLOq3KP0M0vgeJSJCG7cpF4RAA/view?usp=share_link)

For any who are using RStudio to view these files (Cristian and maybe others):
1. I'm interested to know whether `?thing` that you would expect to work actually work. The operator will work with the `thing` names I've selected, but I want to know if those names are actually what you expect them to be, so that you can run `?thing` by just _guessing_ what `thing` should be, and get the results you want.
2. `?rdeephaven` is my intended "portal" to the rest of the documentation. Start here.
3. Do the links work, do they take you to pages you expect?
4. Is it easy to find information you're looking for?


R/rdeephaven/R/exports.R
```diff
@@ -5,4 +5,64 @@
...
+#' `head()`, `tail()`, `update_by()`, and so on. So, you can typically use TableHandles _as if_ they are tables themselves,
+#' and all of the corresponding methods that you call on them will be executed on the server. TableHandles also support
+#' common functional methods for converting server-side Deephaven tables to R objects stored in local memory such as
+#' `as.data.frame()`, `as_tibble()`, and `as_arrow_table()`. For more information on these classes and all of their methods,
+#' see the reference documentation for [`Client`][Client] and [`TableHandle`][TableHandle] by clicking on their class names,
+#' or by running `?Client` or `?TableHandle`.
+#'
+#' @section
+#' Real-time data analysis:
+#' Since TableHandles are references to tables living on the Deephaven server, they may refer to streaming tables, or
```
jcferretti, line 40: "or or"

R/rdeephaven/R/exports.R
```diff
@@ -5,4 +5,64 @@
...
+#' or tables that are receiving new data every second. R objects like data frames or Dplyr tibbles do not have this property -
+#' they are always static objects stored in memory. However, a TableHandle referring to a streaming table may be converted
+#' to a data frame or tibble at any time, and the resulting object will be a snapshot of the table at the time of conversion.
+#' This means that you can use the Deephaven R Client to perform real-time data analysis on streaming data! Of course,
+#' there are performance and memory considerations when pulling data from the server, so it is best to use the provided
```
jcferretti, line 45: I think "of course" can be removed and it results in a better "tone" towards the user (slightly less hinting of "condescension").

R/rdeephaven/R/exports.R
```diff
@@ -5,4 +5,64 @@
...
+#' and [`agg_by()`][AggBy] methods. These table methods are important enough to warrant their own documentation pages, accessible
+#' by clicking on their names, or by running `?UpdateBy` or `?AggBy`. These methods come with their own suites of functions,
+#' prefixed with `agg_` and `uby_` respectively, that are discoverable from their documentation pages. Running `ls("package:rdeephaven")`
+#' will reveal that most of the functions included in this package are for these methods, so it is important to get acquainted
+#' with them.
+#'
+#' @section
+#' Getting help:
+#' While we've done our best to provide good documentation for this package, you may find you need more help than what
+#' this documentation has to offer. If this is the case, please visit the official Deephaven Community Core
```
jcferretti, line 61: "If this is the case" i think can be removed.

R/rdeephaven/R/operation_control.R
```diff
@@ -1,8 +1,7 @@
-#' @description
-#' An OperationControl defines the control parameters of some UpdateByOps used in an update_by table operation.
-#' The UpdateByOps can use OperationControl to handle erroneous data are `ema_tick`, `ema_time`, `emmax_tick`,
-#' `emmax_time`, `emmin_tick`, `emmin_time`, and `ems_time`.
-#' Note that OperationControl should not be instantiated directly by user code, but rather `by op_control()`.
+# An OperationControl defines the control parameters of some UpdateByOps used in an update_by table operation.
```
jcferretti, line 1: You have quoted with backticks in other parts of the documentation for language-level names (eg, parameter names to functions).  Perhaps you want to backtick-quote `OperationControl` ?

R/rdeephaven/R/operation_control.R
```diff
@@ -1,8 +1,7 @@
...
+# This is the return type of op_control(). It is a wrapper around an Rcpp_INTERNAL_OperationControl, which itself is a
+# wrapper around a C++ OperationControl. See rdeephaven/src/client.cpp for details.
```
jcferretti, line 3: May want to consider backtick-quoting the path `rdeephaven/...`

R/rdeephaven/R/operation_control.R
```diff
@@ -16,22 +15,56 @@ OperationControl <- R6Class("OperationControl",
...
 #' @description
-#' Creates an OperationControl that controls behavior on special values in `uby_em*` operations. The arguments `on_null`
-#' and `on_nan` can take the following values:
-#'   'poison': Allow bad data to poison the result. This is only valid for use with NaN.
-#'   'reset': Reset the state for the bucket to NULL when invalid data is encountered.
-#'   'skip': Skip and do not process the invalid data without changing state.
-#'   'throw': Throw an exception and abort processing when bad data is encountered.
-#' @param on_null Defines how an UpdateByOp handles null values it encounters. 'skip' is the default.
-#' @param on_nan Defines how an UpdateByOp handles NaN values it encounters. 'skip' is the default.
+#' Creates an OperationControl that controls behavior on special values in `uby_em*` operations.
```
jcferretti, line 25: Similar comment as above; if you decide in favor of backtick quoting `OperationControl` above, should be consistent here and below.

R/rdeephaven/R/operation_control.R
```diff
@@ -16,22 +15,56 @@ OperationControl <- R6Class("OperationControl",
...
+#' - `'throw'`: Throw an exception and abort processing when bad data is encountered.
+#'
+#' The argument `big_value_context` can take the following values:
+#'
+#' - `'decimal128'`: IEEE 754R Decimal128 format. 34 digits and rounding is half-even.
+#' - `'decimal32'`: IEEE 754R Decimal32 format. 7 digits and rounding is half-even.
+#' - `'decimal64'`: IEEE 754R Decimal64 format. 16 digits and rounding is half-even.
+#' - `'unlimited'`: Unlimited precision arithmetic. Rounding is half-up.
+#'
+#' This function is a generator function. That is, its output is another function that is intended to be used in a call
```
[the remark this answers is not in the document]
jcferretti: Humble suggestion, watch out for your use of connections with "that".   You can re-read your own sentence and make a decision on whether you find that the sentence 'works better' with or without it.   Goes along my previous comment about a balance between color and terseness.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -1,8 +1,51 @@
...
+#' TableHandle that gets created is _the only_ reference to the table that's been created. Importantly, the variable
+#' name given to the TableHandle is purely a _local_ variable, and has no relationship to that table's name on the server.
+#' For this reason, code like the following:
+#' ```r
+#' client <- Client$new(...)
+#' df1 <- data.frame(x = 1:10, y = 11:20)
+#' t1 <- client$import_table(df1)
+#' client$run_script("t2 = t1.update('z = x + y')")
+#' ```
+#' will not run, because the table referenced by the local variable `t1` is not actually named `t1` on the server. To
```
jcferretti, line 20: Saying "not actually named `t1` on the server" feels like an implied "but named something else".   I think the point is that is not named at all, so I suggest replacing:

"not actually named `t1` on the server" => "not actually named on the server"

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -1,8 +1,51 @@
...
+#' make the table referenced by `t1` accessible by name on the server (e.g., from within query strings), you must
+#' _bind it to a variable_ with the method `bind_to_variable()`. We also adopt the convention of calling
```
jcferretti, line 22: Consider removing "also".  "We also adopt the convention of..." => "We adopt the convention of..."

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -1,8 +1,51 @@
...
+#' df1 <- data.frame(x = 1:10, y = 11:20)
+#' th1 <- client$import_table(df1)
+#' th1$bind_to_variable("t1")
+#' client$run_script("t2 = t1.update('z = x + y')")
+#' ```
+#' You can then create a local TableHandle to reference `t2` as follows:
+#' ```r
+#' th2 <- client$open_table("t2")
+#' ```
+#' The above code is not best practice, and it would be preferred to call `update()` directly on `th1` rather than
```
jcferretti, line 36: Consider replacing 

"The above code is not best practice, and it would be preferred to call `update()` directly on `th1` rather..."

with

"The above code is not best practice; calling `update()` directly on `th1` would be preferred to running a script.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -74,7 +117,7 @@ TableHandle <- R6Class("TableHandle",
     },
 
     #' @description
-    #' Gets the dimensions of the table referenced by this TableHandle. Equivalent to c(nrow, ncol).
```
jcferretti, line 77: Consider backtick-quoting TableHandle.

jcferretti: Consider backtick-quoting `c(nrow, ncol)`

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -216,7 +259,7 @@ TableHandle <- R6Class("TableHandle",
     #' The aggregations are defined by the provided operations, which support incremental aggregations over the
     #' corresponding rows in the table. The aggregations will apply position or time-based windowing and compute the
     #' results over the entire table or each row group as identified by the provided key columns.
-    #' See `?UpdateBy` for more information.
+    #' See the documentation on UpdateBy operations [here][UpdateBy] or by running `?UpdateBy` for more information.
```
jcferretti, line 262: "or by running" => "or run"

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -230,7 +273,7 @@ TableHandle <- R6Class("TableHandle",
 
     #' @description
     #' Creates a new table containing grouping columns and grouped data. The resulting grouped data is defined by the
-    #' aggregation(s) specified. See `?Aggregations` for more information.
+    #' aggregation(s) specified. See the documentation on aggregations [here][AggBy] or by running `?AggBy` for more information.
```
jcferretti, line 276: "or by running" => "or run"

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -1,6 +1,87 @@
...
-#' Note that UpdateByOps should not be instantiated directly by user code, but rather by provided uby_* functions.
+#' Deephaven's `update_by()` table method and suite of UpdateBy functions enable cumulative and moving calculations
+#' on static _and_ streaming tables. Complex operations like cumulative minima and maxima, exponential moving averages,
+#' and rolling standard deviations are all possible and effortless to execute. As is always the case in Deephaven,
+#' the results of these calculations will continue to update as their parent tables are updated. Additionally, it's easy
+#' to group data by one or more columns, enabling complex group-wise calculations with a single line of code.
+#'
+#' @section
+#' Applying UpdateBy functions to a table:
+#' The table method `update_by()` is the entrypoint for UpdateBy operations. It takes two arguments: the first is an
```
jcferretti, line 19: entrypoint => entry point

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -14,76 +95,209 @@ UpdateByOp <- R6Class("UpdateByOp",
...
+#' function that is intended to be used in a call to `update_by()`. This detail is typically hidden from the user by
+#' `update_by()`, which calls the generated functions internally. However, it is important to understand this detail
+#' for debugging purposes, as the output of a `uby_*` function can otherwise seem unexpected.
+#'
 #' @param cols String or list of strings denoting the column(s) to operate on. Can be renaming expressions, i.e. “new_col = col”.
 #' Default is to compute the cumulative sum for all non-grouping columns.
 #' @return UpdateByOp to be used in `update_by()`.
+#'
+#' @examples
+#' print("hello!")
```
jcferretti, line 124: `print("hello!")` ?  Is that intentional?  :-)

jcferretti: There is some amount of repetition in the doc strings below for other functions like this one, so two comments:
One is to try to think about some way to try to reduce that repetition.  You could point to a common part of the documentation with a note, instead of copy&paste.  That is not necessarily "obviously better", but I think it is worth to think about doing it;  see if you feel one way or the other is better.  The issue with duplication is maintainability of us.  But is not only that, users coming to the documentation for a specific check/answer will see that comment multiple times; so it may also make sense to give them the option to not having to see the whole thing every time.

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -537,24 +1272,48 @@ uby_rolling_max_tick <- function(cols, rev_ticks, fwd_ticks = 0) {
...
+#' uby_rolling_max_time
+#' @title
+#' Rolling maximum with time as the windowing unit
+#' @md
+#'
 #' @description
 #' Creates a rolling maximum UpdateByOp for each column in `cols`, using time as the windowing unit.
+#'
+#' @details
 #' This uses ISO-8601 time strings as the reverse and forward window parameters.
```
jcferretti, line 1285: Consider adding a URL pointing to the wikipedia page for ISO-8601

R/rdeephaven/man/TableHandle.Rd
```diff
@@ -7,8 +7,54 @@
 \description{
 A TableHandle holds a reference to a Deephaven Table on the server, and provides methods for operating on that table.
 Note that TableHandles should not be instantiated directly by user code, but rather by server calls accessible from
-the `Client` class. See `?Client` for more information.
+the \code{\link{Client}} class. See \code{?Client} for more information.
 }
+\section{Naming tables on the server}{
+
+When a TableHandle is created, it is not automatically bound to a variable name on the server. This means that the
```
jcferretti, line 14: I have read this same text earlier; consider replacing the full explanation repeated here with a summary (short explanation)  and a pointer to the other part of the documentation with the full explanation.

jcferretti (commented):
I made a bunch of comments, most of them are suggestions and things to consider rather than hard black and white mandates, so ponder those and make your own calls.

I wanted to say, I think this is excellent and the level of detail and effort you put here is really above and beyond.  Thanks for caring this much!

chipkent (changes requested):

R/rdeephaven/man/AggBy.Rd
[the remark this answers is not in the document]
alexpeters1208: I'm strongly inclined to say that it's not a mistake. The `devtools::document()` function that generates the docs from Roxygen docstrings _does not_ get run automatically on package installation. As an R user, I never knew about this function before this project, and if I got a package that did not appear to have any documentation, I would not have run `devtools::document()` to see if that fixed things.
Also [arrow](https://github.com/apache/arrow/tree/main/r/man) does it, [dplyr](https://github.com/tidyverse/dplyr/tree/main/man) does it, [ggplot2](https://github.com/tidyverse/ggplot2/tree/main/man) does it, [caret](https://github.com/topepo/caret/tree/master/pkg/caret/man) does it, etc etc.

R/rdeephaven/R/agg_ops_wrapper.R
```diff
@@ -1,6 +1,73 @@
...
+#'
+#' @section
+#' `agg` functions:
+#' `agg` functions are used to perform aggregation calculations on grouped data by passing them to `agg_by()` or
+#' `agg_all_by()`. These functions are _generators_, meaning they return _functions_ that the Deephaven engine knows
+#' how to interpret. We call the functions that they return `AggOp`s. These `AggOp`s are not R-level functions, but
+#' wrappers around C++ functions that perform all of the intensive calculations. Here is a list of all `agg` functions
+#' available in Deephaven:
+#'
+#' - [`agg_first()`][agg_first]
```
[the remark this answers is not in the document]
alexpeters1208: Yes.

R/rdeephaven/R/agg_ops_wrapper.R
```diff
@@ -1,6 +1,73 @@
...
+#'
+#' For more details on each aggregation function, click on one of the methods above or see the reference documentation
+#' by running `?agg_first`, `?agg_last`, etc.
+#'
+#' @section
+#' Examples:
+#'
+#' ```
+#' NULL
+#' ```
```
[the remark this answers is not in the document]
alexpeters1208: Yes, that is on today's menu.

R/rdeephaven/R/agg_ops_wrapper.R
```diff
@@ -1,6 +1,73 @@
...
+#'
+NULL
+
+# An AggOp represents an aggregation operator that can be passed to agg_by() or agg_all_by(). This is the return type
+# of all of the `agg` functions. It is a wrapper around an Rcpp_INTERNAL_AggOp, which itself is a wrapper around a
+# C++ AggregateWrapper, which is finally a wrapper around a C++ Aggregate. See rdeephaven/src/client.cpp for details.
+# Note that AggOps should not be instantiated directly by user code, but rather by provided `agg` functions.
```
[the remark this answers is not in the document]
alexpeters1208: These are regular comments and not Roxygen comments. They don't get turned into documentation. This is here for the next dev reading the source code, because it's kind of convoluted if you haven't seen anything like it before.

R/rdeephaven/man/AggBy.Rd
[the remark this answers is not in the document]
alexpeters1208: There may be a subtlety in your statement that I'm missing. Are you suggesting it would be better to have CI generate and check in these files? That might be a good solution, since there's no worry about the issues mentioned in the latter half of your comment.

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -427,24 +1123,48 @@ uby_rolling_avg_tick <- function(cols, rev_ticks, fwd_ticks = 0) {
...
+#'
+#' This function, like other Deephaven `uby` functions, is a generator function. That is, its output is another
+#' function called an [`UpdateByOp`][UpdateByOp] intended to be used in a call to `update_by()`. This detail is typically
+#' hidden from the user. However, it is important to understand this detail for debugging purposes, as the output of
+#' a `uby` function can otherwise seem unexpected.
+#'
 #' @param ts_col String denoting the column to use as the timestamp.
 #' @param cols String or list of strings denoting the column(s) to operate on. Can be renaming expressions, i.e. “new_col = col”.
-#' Default is to compute the rolling average for all non-grouping columns.
+#' Default is to compute the rolling sum for all non-grouping columns.
```
margaretkennedy, line 1160: you've changed group to sum, and average to sum here - just checking that's deliberate

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -454,26 +1174,50 @@ uby_rolling_avg_time <- function(ts_col, cols, rev_time, fwd_time = "PT0s") {
...
+#' `A` and `B` columns.
+#'
+#' This function, like other Deephaven `uby` functions, is a generator function. That is, its output is another
+#' function called an [`UpdateByOp`][UpdateByOp] intended to be used in a call to `update_by()`. This detail is typically
+#' hidden from the user. However, it is important to understand this detail for debugging purposes, as the output of
+#' a `uby` function can otherwise seem unexpected.
+#'
 #' @param cols String or list of strings denoting the column(s) to operate on. Can be renaming expressions, i.e. “new_col = col”.
-#' Default is to compute the rolling minimum for all non-grouping columns.
+#' Default is to compute the rolling sum for all non-grouping columns.
```
margaretkennedy, line 1213: ?

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -564,26 +1380,50 @@ uby_rolling_max_time <- function(ts_col, cols, rev_time, fwd_time = "PT0s") {
   return(UpdateByOp$new(INTERNAL_rolling_max_time(ts_col, cols, rev_time, fwd_time)))
 }
 
+#' @name
+#' uby_rolling_prod_tick
+#' @title
+#' Rolling prod with ticks as the windowing unit
```
margaretkennedy, line 1386:
```suggestion
#' Rolling product with ticks as the windowing unit
```

R/rdeephaven/man/AggBy.Rd
```diff
@@ -0,0 +1,101 @@
...
+Here are some details on each method:
+\itemize{
+\item \code{TableHandle$agg_by(aggs, by)}: Creates a new table containing grouping columns and grouped data.
+The resulting grouped data is defined by the aggregation(s) specified.
+\item \code{TableHandle$agg_all_by(agg, by)}: Creates a new table containing grouping columns and grouped data.
+The resulting grouped data is defined by the aggregation specified. This method applies the aggregation to all
+non-grouping columns of the table, so it can only accept one aggregation at a time.
+}
+
+The \code{agg_by()} and \code{agg_all_by()} methods themselves do not know anything about the columns you want to perform
```
margaretkennedy, line 27:
```suggestion
The \code{agg_by()} and \code{agg_all_by()} methods themselves do not know anything about the columns on which you want to perform
```

R/rdeephaven/man/AggBy.Rd
```diff
@@ -0,0 +1,101 @@
...
+aggregations on. Rather, the desired columns are passed to individual \code{agg} functions, enabling you to apply
```
margaretkennedy, line 28:
```suggestion
aggregations. Rather, the desired columns are passed to individual \code{agg} functions, enabling you to apply
```

R/rdeephaven/man/AggBy.Rd
margaretkennedy: I think my suggested change in the previous file will actually change it here, too?

R/rdeephaven/man/AggBy.Rd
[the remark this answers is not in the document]
jcferretti: Is a complicated topic and there are schools of thought about this in the "abstract".  I was "raised" in the "never check in generated files" school.  However, I believe we in DH are strongly of the "check in all generated files and when there are new versions have a check that fails and forces the PR to re-generate and add to git the generated changes".

So a general discussion here is possible, but if I understand correctly how we treat generated files (eg, protobuf compiler output, java expression parser in the engine, chunked types classes, and a lot of others), we consistently check in / add to git everything necessary for a build that we ourselves make.  Meaning all generated files.

R/rdeephaven/R/exports.R
```diff
@@ -7,4 +7,103 @@
...
+#' The Deephaven Community R Client
+#' @md
+#' @usage NULL
+#' @format NULL
+#'
+#' @description
+#' The Deephaven Community R Client provides an R interface to Deephaven's powerful real-time data engine, [_Deephaven Core_](https://deephaven.io/community/).
+#' To use this package, you must have a Deephaven server running and be able to connect to it. For more information on
+#' how to set up a Deephaven server, see the documentation [here](https://deephaven.io/core/docs/tutorials/quickstart/).
+#' For a brief overview of the Deephaven R Client, see the [blog post](https://deephaven.io/blog/2023/10/04/r-client-updates/)
```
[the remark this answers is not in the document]
chipkent: I would lean away from linking to blog posts in ref docs.

R/rdeephaven/man/AggBy.Rd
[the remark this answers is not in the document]
niloc132: Summary:
 * In Java, we try not to check in generated sources, since usually we can get Gradle to do it for us. There are exceptions, where we either don't have the workflow finished, or the code that is generated needs some tweaking, or where we just can't be bothered for legacy reasons.
 * In Go, in order for the ecosystem's packaging tools to work, all source code must be committed to a git repo, so generated sources are checked in. No option here if we want to be part of the ecosystem.
 * Our python client code uses checked in sources because it was decided that we wouldn't follow gRPC/protobuf conventions in naming, and while we have a process that lets a user invoke Gradle to generate the appropriate python output and then mangle names with the corresponding sed-in-groovy magic, the thinking was that a Python developer shouldn't need to use Gradle to make small changes to our code. 
 * Our c++ client code uses checked in sources for reasons I don't recall, but at least we just run protoc without extra magic. Likely the thinking was the same - while a c++ developer would need to follow another 18 steps to get a working build, that 19th of using Gradle was a doozy. 

As such I don't strongly object, but I would love to revisit some of these past decisions. If we're going to use "check in the sources, and validate that they are correct by running Gradle", it seems like we could abbreviate this to "run Gradle to have the right sources", but perhaps verifying correctness only in CI is okay.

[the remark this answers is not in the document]
chipkent: The Go statement isn't true... but that isn't relevant here.

R/rdeephaven/R/agg_ops_wrapper.R
```diff
@@ -1,6 +1,134 @@
...
+#'
+#' - `TableHandle$agg_by(aggs, by)`: Creates a new table containing grouping columns and grouped data.
+#'   The resulting grouped data is defined by the aggregation(s) specified.
+#' - `TableHandle$agg_all_by(agg, by)`: Creates a new table containing grouping columns and grouped data.
+#'   The resulting grouped data is defined by the aggregation specified. This method applies the aggregation to all
+#'   non-grouping columns of the table, so it can only accept one aggregation at a time.
+#'
+#' The `agg_by()` and `agg_all_by()` methods themselves do not know anything about the columns on which you want to
+#' perform aggregations. Rather, the desired columns are passed to individual `agg` functions, enabling you to apply
+#' various kinds of aggregations to different columns or groups of columns as needed.
```
chipkent, line 37: I'm not sure if `agg_all_by` operators accept column args.  I think they don't, but this wording sounds like they do.

R/rdeephaven/R/agg_ops_wrapper.R
```diff
@@ -1,6 +1,134 @@
...
+NULL
+
+
+#' Name AggOp
+#' @title Deephaven AggOps
+#' @md
 #' @description
-#' An AggOp represents an aggregation operation that can be passed to `agg_by()` or `agg_all_by()`.
-#' Note that AggOps should not be instantiated directly by user code, but rather by provided agg_* functions.
+#' An `AggOp` is the return type of one of Deephaven's [`agg`][AggBy] functions. It is a function that performs the
```
chipkent, line 113: Should indicate that these are "deephaven" functions and not general R functions.

R/rdeephaven/R/exports.R
```diff
@@ -7,4 +7,101 @@
...
+#' ```
+#' There are performance and memory considerations when pulling data from the server, so it is best to use the provided
+#' TableHandle methods to perform as much of your analysis as possible on the server, and to only pull the data when
+#' something _must_ be done in R, like plotting or writing to a local file.
+#'
+#' @section
+#' Powerful table operations:
+#' Much of the power of Deephaven's suite of table operations is achieved through the use of the [`update_by()`][UpdateBy]
+#' and [`agg_by()`][AggBy] methods. These table methods are important enough to warrant their own documentation pages, accessible
+#' by clicking on their names, or by running `?UpdateBy` or `?AggBy`. These methods come with their own suites of functions,
```
chipkent, line 92: It seems weird that `UpdateBy` and `AggBy` are links here as well as `?` operator targets.

If this is as problem, I would search the rest of the docs for similar problems.

R/rdeephaven/R/exports.R
```diff
@@ -7,4 +7,101 @@
...
+#' prefixed with `agg_` and `uby_` respectively, that are discoverable from their documentation pages. Running `ls("package:rdeephaven")`
+#' will reveal that most of the functions included in this package are for these methods, so it is important to get acquainted
+#' with them.
+#'
+#' @section
+#' Getting help:
+#' While we've done our best to provide good documentation for this package, you may find you need more help than what
+#' this documentation has to offer. Please visit the official Deephaven Community Core [documentation](https://deephaven.io/core/docs/tutorials/quickstart/)
```
chipkent, line 100:
```suggestion
#' this documentation has to offer. Please visit the official Deephaven Community Core [documentation](https://deephaven.io/core/docs/)
```

R/rdeephaven/R/operation_control.R
```diff
@@ -1,8 +1,32 @@
...
+#' @name OperationControl
+#' @title Deephaven OperationControl
+#' @md
 #' @description
-#' An OperationControl defines the control parameters of some UpdateByOps used in an update_by table operation.
-#' The UpdateByOps can use OperationControl to handle erroneous data are `ema_tick`, `ema_time`, `emmax_tick`,
-#' `emmax_time`, `emmin_tick`, `emmin_time`, and `ems_time`.
-#' Note that OperationControl should not be instantiated directly by user code, but rather `by op_control()`.
+#' An `OperationControl` is the return type of Deephaven's [`op_control`][op_control] function. It is a function that
+#' determines how special values will be handled in the context of an [`update_by`][UpdateBy] operation.
```
chipkent, line 12: Here is another `UpdateBy`.

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -216,8 +305,8 @@ TableHandle <- R6Class("TableHandle",
     #' The aggregations are defined by the provided operations, which support incremental aggregations over the
     #' corresponding rows in the table. The aggregations will apply position or time-based windowing and compute the
     #' results over the entire table or each row group as identified by the provided key columns.
-    #' See `?UpdateBy` for more information.
-    #' @param ops UpdateByOp or list of UpdateByOps to perform on non-grouping columns.
+    #' See more detailed documentation [here][UpdateBy] or run `?UpdateBy`.
```
chipkent, line 308: Here again

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -230,8 +319,8 @@ TableHandle <- R6Class("TableHandle",
 
     #' @description
     #' Creates a new table containing grouping columns and grouped data. The resulting grouped data is defined by the
-    #' aggregation(s) specified. See `?Aggregations` for more information.
-    #' @param aggs AggOp or list of AggOps to perform on non-grouping columns.
+    #' aggregation(s) specified. See more detailed documentation [here][AggBy] or run `?AggBy`.
```
chipkent, line 322: here too

R/rdeephaven/R/table_handle_wrapper.R
```diff
@@ -249,10 +338,10 @@ TableHandle <- R6Class("TableHandle",
 
     #' @description
     #' Creates a new table containing grouping columns and grouped data. The resulting grouped data is defined by the
-    #' aggregation(s) specified. See `?Aggregations` for more information.
-    #' This method applies the aggregation to all columns of the table, so it can only
+    #' aggregation(s) specified. See more detailed documentation [here][AggBy] or run `?AggBy`.
```
chipkent, line 341: and here

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -1,6 +1,144 @@
+# An UpdateByOp represents a window-based operator that can be passed to update_by(). This is the return type of
+# all of the uby functions. It is a wrapper around an Rcpp_INTERNAL_UpdateByOp, which itself is a wrapper around a
+# C++ UpdateByOpWrapper, which is finally a wrapper around a C++ UpdateByOperation. See rdeephaven/src/client.cpp for details.
+# Note that UpdateByOps should not be instantiated directly by user code, but rather by provided uby functions.
+
+
+#' @name
+#' UpdateBy
```
chipkent, line 8: Maybe this is the source of my concerns.  Should it be `update_by`?

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -1,6 +1,144 @@
...
+
+#' @name UpdateByOp
+#' @title Deephaven UpdateByOps
+#' @md
 #' @description
-#' An UpdateByOp represents a window-based operator that can be passed to `update_by()`.
-#' Note that UpdateByOps should not be instantiated directly by user code, but rather by provided uby_* functions.
+#' An `UpdateByOp` is the return type of one of Deephaven's [`uby`][UpdateBy] functions. It is a function that performs
+#' the computation specified by the `uby` function. These are intended to be passed directly to `update_by()`,
+#' and should never be instantiated directly be user code.
```
chipkent, line 125: Should this make clear tht it is not a general purpose R function?

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -14,76 +152,390 @@ UpdateByOp <- R6Class("UpdateByOp",
...
+#'
+#' This function, like other Deephaven `uby` functions, is a generator function. That is, its output is another
+#' function called an [`UpdateByOp`][UpdateByOp] intended to be used in a call to `update_by()`. This detail is typically
+#' hidden from the user. However, it is important to understand this detail for debugging purposes, as the output of
+#' a `uby` function can otherwise seem unexpected.
+#'
 #' @param cols String or list of strings denoting the column(s) to operate on. Can be renaming expressions, i.e. “new_col = col”.
 #' Default is to compute the cumulative product for all non-grouping columns.
-#' @return UpdateByOp to be used in `update_by()`.
+#' @return `UpdateByOp` to be used in a call to `update_by()`.
```
chipkent, line 241: `update_by()` is listed multiple times in here without a link.  Seems like it should have a link.  Change probably needed in other places as well.

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -111,18 +623,78 @@ uby_ema_tick <- function(decay_ticks, cols = character(), operation_control = op
...
-#'   `a_current = e^(-dt_current / decay_time)`
-#'   `ema_first = first_value`
-#'   `ema_current = a_current * ema_prev + (1 - a_current) * current_value`
+#' \deqn{a_i = e^{\frac{-dt_i}{\tau}}}
+#' \deqn{\bar{x}_0 = x_0}
+#' \deqn{\bar{x}_i = a_i*\bar{x}_{i-1} + (1-a_i)*x_i}
+#'
+#' Where:
+#' - \eqn{dt_i} is the difference between time \eqn{t_i} and \eqn{t_{i-1}} in nanoseconds.
+#' - \eqn{\tau} is `decay_time` in nanoseconds, an input parameter to the method.
```
chipkent, line 643: I find "in nanoseconds" to be confusing here since users are specifying the duration string, which could be specified in hours or maybe days.

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -150,18 +782,77 @@ uby_ems_tick <- function(decay_ticks, cols = character(), operation_control = op
...
-#'   `ems_current = a_current * ems_prev + current_value`
-#' @param ts_col String denoting the column to use as the timestamp.
-#' @param decay_time ISO-8601-formatted string specifying the decay rate.
+#' \deqn{a_i = e^{\frac{-dt_i}{\tau}}}
+#' \deqn{\mathcal{S}_0 = x_0}
+#' \deqn{\mathcal{S}_i = a_i*\mathcal{S}_{i-1} + x_i}
+#'
+#' Where:
+#' - \eqn{dt_i} is the difference between time \eqn{t_i} and \eqn{t_{i-1}} in nanoseconds.
+#' - \eqn{\tau} is `decay_time` in nanoseconds, an input parameter to the method.
```
chipkent, line 802: same here.  I'm guessing that there are other cases as well.  Could probably also remove "in nanoseconds" from the prior line.

R/rdeephaven/R/update_by_ops_wrapper.R
```diff
@@ -427,24 +1746,78 @@ uby_rolling_avg_tick <- function(cols, rev_ticks, fwd_ticks = 0) {
...
+#' @title
+#' Rolling average with time as the windowing unit
+#' @md
+#'
 #' @description
 #' Creates a simple moving average (SMA) UpdateByOp for each column in `cols`, using time as the windowing unit.
-#' This uses ISO-8601 time strings as the reverse and forward window parameters.
+#'
+#' @details
+#' This uses [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) time strings as the reverse and forward window parameters.
```
chipkent, line 1759: It supports these and other formats.  I think the examples below fall into the other formats.

chipkent (approved):
I am approving under the assumption that the open comments will be addressed in a follow-on PR.

PR #4611, opened 2023-10-09
Make extensions-kafka source-compatible with confluent 5.5.x

devinrsmith:
This is a simple change that removes our usage of a few "newer" confluent APIs. This in effect makes our code source-compitable with confluent platform 5.5.x. There has been minimal testing to ensure this code compiles with 5.5.x dependencies.


rcaudy (commented):
So the upshot is:
1. We'll now log the configuration
2. We lose some mocking functionality for schema registry client

Right?

devinrsmith:
Yes; I didn't notice it at the time, but w/ the protobuf PR we lost logging out the props passed to the schema registry. We do lose some mocking hooks for the schema registry client, but it's not something we were using anyways.

rcaudy (approved):

PR #4625, opened 2023-10-11
R dockerized test run gradle target and CI

### LINKED ISSUE #4091: Continuous Integration for the R Client
When R is ready for CI setup, we need to figure out how that gets done w.r.t CRAN repository. @niloc132 has discussed doing unit testing inside of Docker containers so that everything stays isolated, which probably involves some Gradle magic. This will become a top priority when the R API is more complete than it is today.

We will also have issues with the new PSK default, as the current R unit tests assume anon default. That will be my responsibility to overhaul.

jcferretti:
Fixes #4091


jcferretti:
Not ready for review yet, thus "draft".
At this point is mostly working:
* cpp-client build.gradle refactoring is working; separate targets for image building (that can be used from the R build) and test run work.
* Need to add xml2 package to r-base-image for R junit test reporter to work https://github.com/deephaven/deephaven-base-images/pull/94/files
* Need to fix dependency at the beginning of `cpp-client/build.gradle` that says `evaluationDependsOn Docker.registryProject('cpp-client-base')` to say `evaluationDependsOn Docker.registryProject('r-client-base')`.  The later is failing and I don't understand why.

R/build.gradle
```diff
@@ -0,0 +1,96 @@
...
+        copyFile('rdeephaven/etc/', "${prefix}/src/rdeephaven/etc/")
+        copyFile('rdeephaven/R/', "${prefix}/src/rdeephaven/R/")
+        copyFile('rdeephaven/src/*.cpp', "${prefix}/src/rdeephaven/src/")
+        copyFile('rdeephaven/src/Makevars', "${prefix}/src/rdeephaven/src/")
+        copyFile('r-tests.sh', "${prefix}/bin/rdeephaven")
+        runCommand("PREFIX=\"" + prefix + "\"; \\\n" +
+                  '''set -eux ; \
+                     cd "${PREFIX}/src/rdeephaven"; \
+                     . "${PREFIX}/env.sh"; \
+                     echo 'install.packages(".", repos=NULL, type="source")' | \
```
alexpeters1208, line 77: There might be reason to want to do this via
```
R CMD build rdeephaven
R CMD check rdeephaven*.tar.gz
R CMD INSTALL rdeephaven*.tar.gz
```
from terminal in the `${PREFIX}/src` directory, rather than getting into an R console and running `install.packages` directly.
`R CMD check` runs a laundry list of [checks](https://kbroman.org/pkg_primer/pages/check.html) that, if we ever make it on CRAN, CRAN will run as well. This gives us an extremely verbose and detailed look at what are package is doing, if all the documentation is as it should be, if the package structure is as it should be, etc etc.

I think it would not be the case that we want to call CI "failed" if `R CMD check` complains about something but all of the unit tests pass. Maybe we would, or maybe there's an intermediate stage between passing and failing CI that says "tests don't fail, but you should look at the output from R CMD check and see what's up".

jcferretti: We discussed on slack and agreed that for automated CI checks, we want to be doing the same thing that we are asking the customer to do in the README regarding the manual steps they would run (so that we continue to check that what we are asking users to do to build works).

mofojed (changes requested):
I'm missing some context. What ticket is this fixing? It looks like this is just running tests, is there a reason it needs to make this release?

alexpeters1208 (approved):

cpp-client/build.gradle
```diff
@@ -108,19 +110,36 @@ def testCppClient = Docker.registerDockerTask(project, 'testCppClient') {
...
-        environmentVariable 'LD_LIBRARY_PATH', "${prefix}/lib"
     }
     containerDependencies.dependsOn = [deephavenDocker.healthyTask]
     containerDependencies.finalizedBy = deephavenDocker.endTask
     network = deephavenDocker.networkName.get()
-    parentContainers = [ Docker.registryTask(project, 'cpp-client-base') ]
+    parentContainers = [ project.tasks.getByName('cppClient') ]
     entrypoint = ["${prefix}/bin/dhcpp/cpp-tests-to-junit.sh", '/out/cpp-test.xml', '/out/cpp-test.log']
 }
+tasks.getByName('testCppClient').dependsOn project.tasks.getByName('cppClient')
```
JamesXNelson, line 142: when you do this, you forcibly initialize both of these tasks regardless of whether either of them will actually be running.

Instead consider updating the closure sent to `Docker.registerDockerTask(project, 'testCppClient') {}` (circa line 123)  to additionally do:
```
dependsOn "$project.path:cppClient"
```
or
```
dependsOn buildCppClientImage
```
(the latter references the TaskProvider return value of `Docker.registerDockerTask`

cpp-client/cpp-tests-to-junit.sh
```diff
@@ -11,7 +11,7 @@ if [ "$#" -ne 2 ]; then
 fi
 
 if [ -z "${DH_PREFIX}" ]; then
-    echo "$0: Environment variable DHCPP_PREFIX is not set, aborting." 1>&2
+    echo "$0: Environment variable DH_PREFIX is not set, aborting." 1>&2
```
JamesXNelson, line 14: Would `DH_CPP_PREFIX` not be more appropriate? `DH_PREFIX` is pretty generic, but since I don't really understand the full scope of this variable, feel free to say "no, I like what I did" / just resolve my comment.

.github/workflows/tag-base-images.yml
```diff
@@ -60,6 +60,7 @@ jobs:
         if: ${{ startsWith(github.ref, 'refs/heads/release/v') }}
         run: |
           ./docker/registry/cpp-client-base/build/crane/retag.sh
```
devinrsmith, line 62: Should we remove cpp-client-base? Here, but also delete the registry directory and make sure nothing references it?

devinrsmith (commented):
I made a bad change to R/rdeephaven/src/client.cpp to see what would happen, and gradle still reports "success".

Internally, I see:

> ERROR: compilation failed for package 'rdeephaven'

I'm assuming we need something that will complete the loop when something internally goes wrong.

cpp-client/cpp-tests-to-junit.sh
jcferretti: I changed it because we are installing there now more than just C++, eg, also R and potentially python/cython.

cpp-client/build.gradle
jcferretti:
```
cfs@caicai 23:28:59 ~/dh/oss1/deephaven-core
$ git diff
diff --git a/cpp-client/build.gradle b/cpp-client/build.gradle
index 358337b1f8..d0944760a5 100644
--- a/cpp-client/build.gradle
+++ b/cpp-client/build.gradle
@@ -119,6 +119,7 @@ def buildCppClientImage = Docker.registerDockerTask(project, 'cppClient') {
 }
 
 def testCppClient = Docker.registerDockerTask(project, 'testCppClient') {
+    dependsOn "$project.path:cppClient"
     // Only tested on x86-64, and we only build dependencies for x86-64
     platform = 'linux/amd64'
     copyIn {}
@@ -139,7 +140,6 @@ def testCppClient = Docker.registerDockerTask(project, 'testCppClient') {
     parentContainers = [ project.tasks.getByName('cppClient') ]
     entrypoint = ["${prefix}/bin/dhcpp/cpp-tests-to-junit.sh", '/out/cpp-test.xml', '/out/cpp-test.log']
 }
-tasks.getByName('testCppClient').dependsOn project.tasks.getByName('cppClient')
 
 deephavenDocker.shouldLogIfTaskFails testCppClient
 tasks.check.dependsOn(testCppClient)

cfs@caicai 23:29:02 ~/dh/oss1/deephaven-core
$ ./gradlew :cpp-client:testCppClient

* Where:
Build file '/home/cfs/dh/oss1/deephaven-core/cpp-client/build.gradle' line: 122

* What went wrong:
A problem occurred evaluating project ':cpp-client'.
> Could not find method dependsOn() for arguments [:cpp-client:cppClient] on object of type Docker$DockerTaskConfig.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.

* Get more help at https://help.gradle.org

Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

See https://docs.gradle.org/7.6.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD FAILED in 312ms
10 actionable tasks: 10 up-to-date
```

And

```
cfs@caicai 23:29:55 ~/dh/oss1/deephaven-core
$ git diff
diff --git a/cpp-client/build.gradle b/cpp-client/build.gradle
index 358337b1f8..e621e47126 100644
--- a/cpp-client/build.gradle
+++ b/cpp-client/build.gradle
@@ -119,6 +119,7 @@ def buildCppClientImage = Docker.registerDockerTask(project, 'cppClient') {
 }
 
 def testCppClient = Docker.registerDockerTask(project, 'testCppClient') {
+    dependsOn buildCppClientImage
     // Only tested on x86-64, and we only build dependencies for x86-64
     platform = 'linux/amd64'
     copyIn {}
@@ -139,7 +140,6 @@ def testCppClient = Docker.registerDockerTask(project, 'testCppClient') {
     parentContainers = [ project.tasks.getByName('cppClient') ]
     entrypoint = ["${prefix}/bin/dhcpp/cpp-tests-to-junit.sh", '/out/cpp-test.xml', '/out/cpp-test.log']
 }
-tasks.getByName('testCppClient').dependsOn project.tasks.getByName('cppClient')
 
 deephavenDocker.shouldLogIfTaskFails testCppClient
 tasks.check.dependsOn(testCppClient)

cfs@caicai 23:29:57 ~/dh/oss1/deephaven-core
$ ./gradlew :cpp-client:testCppClient

* Where:
Build file '/home/cfs/dh/oss1/deephaven-core/cpp-client/build.gradle' line: 122

* What went wrong:
A problem occurred evaluating project ':cpp-client'.
> Could not find method dependsOn() for arguments [provider(task 'cppClient', class org.gradle.api.DefaultTask)] on object of type Docker$DockerTaskConfig.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.

* Get more help at https://help.gradle.org

Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

See https://docs.gradle.org/7.6.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD FAILED in 301ms
10 actionable tasks: 10 up-to-date
```

Which essentially are the same error.

.github/workflows/tag-base-images.yml
jcferretti: r-base is based on cpp-client-base, so I don't think we can remove it from the base-images repo.   The tagging here I am not sure what it does, maybe we should have a short chat on slack or VC.

jcferretti:
I fixed the problem Devin's noted (thanks!) about a failure of compilation of the C++ code inside R's install.packages not triggering a gradle task failure.   Interestingly enough, if some source fails compilation (which means the whole package won't build), install.packages only gives a warning... what?  Sigh.

cpp-client/build.gradle
niloc132: I think this is what he's suggesting:
```patch

diff --git a/cpp-client/build.gradle b/cpp-client/build.gradle
index 358337b1f8..332726996c 100644
--- a/cpp-client/build.gradle
+++ b/cpp-client/build.gradle
@@ -139,7 +139,9 @@ def testCppClient = Docker.registerDockerTask(project, 'testCppClient') {
     parentContainers = [ project.tasks.getByName('cppClient') ]
     entrypoint = ["${prefix}/bin/dhcpp/cpp-tests-to-junit.sh", '/out/cpp-test.xml', '/out/cpp-test.log']
 }
-tasks.getByName('testCppClient').dependsOn project.tasks.getByName('cppClient')
+testCppClient.configure {
+    dependsOn buildCppClientImage
+}
 
 deephavenDocker.shouldLogIfTaskFails testCppClient
 tasks.check.dependsOn(testCppClient)
```

JamesXNelson: colin's comment is the shortest form to make it do what I was suggesting.  I didn't understand the purpose of the closure passed to Docker.registerDockerTask properly, I thought it was applied to a task and executed, which would provide `dependsOn`, but colin's answer sidesteps that and gets to the meat of the issue w/ the least complexity / effort.

R/build.gradle
```diff
@@ -0,0 +1,107 @@
...
+            include 'rdeephaven/LICENSE'
+            include 'rdeephaven/NAMESPACE'
+            include 'rdeephaven/README.md'
+            include 'rdeephaven/inst/**'
+            include 'rdeephaven/man/**'
+            include 'rdeephaven/etc/**'
+            include 'rdeephaven/R/**'
+            include 'rdeephaven/src/*.cpp'
+            include 'rdeephaven/src/Makevars'
+        }
```
niloc132, line 47: if i'm reading this right, we're excluding the following:
 * build/
 * build.gradle
 * build.properties
 * rdeephaven/.gitignore
 * src/CMakeLists.txt
 * src/.gitignore

Any thought as to instead just copying everything (minus `build/`) to keep this a bit more brief, less confusing if a new file needs to be added later?

Possibly like this:
```patch
diff --git a/R/build.gradle b/R/build.gradle
index ee6b2c1b82..a66752b6a0 100644
--- a/R/build.gradle
+++ b/R/build.gradle
@@ -32,18 +32,7 @@ def buildRClient = Docker.registerDockerTask(project, 'rClient') {
 
     copyIn {
         from(layout.projectDirectory) {
-            include 'r-build.sh'
-            include 'r-tests.sh'
-            include 'rdeephaven/DESCRIPTION'
-            include 'rdeephaven/LICENSE'
-            include 'rdeephaven/NAMESPACE'
-            include 'rdeephaven/README.md'
-            include 'rdeephaven/inst/**'
-            include 'rdeephaven/man/**'
-            include 'rdeephaven/etc/**'
-            include 'rdeephaven/R/**'
-            include 'rdeephaven/src/*.cpp'
-            include 'rdeephaven/src/Makevars'
+            exclude 'build'
         }
     }
     dockerfile {
@@ -57,24 +46,14 @@ def buildRClient = Docker.registerDockerTask(project, 'rClient') {
                         ${prefix}/bin/rdeephaven \\
                         ${prefix}/src/rdeephaven/{inst,man,etc,src,R,bin}
                    """)
-        copyFile('rdeephaven/DESCRIPTION',  "${prefix}/src/rdeephaven/")
-        copyFile('rdeephaven/LICENSE',  "${prefix}/src/rdeephaven/")
-        copyFile('rdeephaven/NAMESPACE',  "${prefix}/src/rdeephaven/")
-        copyFile('rdeephaven/README.md',  "${prefix}/src/rdeephaven/")
-        copyFile('rdeephaven/inst/',  "${prefix}/src/rdeephaven/inst/")
-        copyFile('rdeephaven/man/', "${prefix}/src/rdeephaven/man/")
-        copyFile('rdeephaven/etc/', "${prefix}/src/rdeephaven/etc/")
-        copyFile('rdeephaven/R/', "${prefix}/src/rdeephaven/R/")
-        copyFile('rdeephaven/src/*.cpp', "${prefix}/src/rdeephaven/src/")
-        copyFile('rdeephaven/src/Makevars', "${prefix}/src/rdeephaven/src/")
+        copyFile('rdeephaven',  "${prefix}/src/rdeephaven")
         copyFile('r-tests.sh', "${prefix}/bin/rdeephaven")
         copyFile('r-build.sh', "${prefix}/bin/rdeephaven")
-        runCommand("PREFIX=\"" + prefix + "\"; \\\n" +
-                  '''set -eux ; \
-                     cd "${PREFIX}/src/rdeephaven"; \
-                     . "${PREFIX}/env.sh"; \
-                     ${PREFIX}/bin/rdeephaven/r-build.sh
-                   ''')
+        runCommand("""set -eux ; \
+                     cd "${prefix}/src/rdeephaven"; \
+                     . "${prefix}/env.sh"; \
+                     ${prefix}/bin/rdeephaven/r-build.sh
+                   """)
     }
     parentContainers = [ project.tasks.getByPath(':cpp-client:cppClient') ]
 }
```

cpp-client/build.gradle
jcferretti: Done, it worked, thank you guys.

niloc132: (We would love to have it be a proper task configuring closure, but as far as I can tell it is impossible to define a task as "run these tasks", there's no such thing as task composition to produce a task. As we need to provide different configuration details to different constituent tasks, another option could be to define an extension... but as in projects like this one, we need to define at least two instances of "do stuff with docker, using inputs and outputs, etc.." as an extension... and i'm not aware of a nice way to apply the same extension more than once. So, we return the "most-downstream" task from the function, and as needed have other tasks depend on it.)

Yep  I'm sorry, what I gave you is equivalent to what you had before, but also wrong - the testCppClient task is the very last step in "run the test in docker and then finally copy the results out to gradle", which means that this dependency will not make sense, Gradle would be within its rights to not build the cpp client image until just before R test output is attempted to be copied from docker.

However, if I am reading this all right, it is also unnecessary - this line already covers the requirement you're after:
```groovy
    parentContainers = [ project.tasks.getByName('cppClient') ]
```
This ensures that before `testCppClient` tries to _start_ to build its own image, cppClient must be complete.

niloc132: To restate, because parentContainers= already has a reference to :cppClient, this line which I "corrected" is entirely unnecessary.

R/build.gradle
```diff
@@ -0,0 +1,109 @@
...
+        environmentVariable 'DH_HOST', deephavenDocker.containerName.get()
+        environmentVariable 'DH_PORT', '10000'
+    }
+    containerDependencies.dependsOn = [deephavenDocker.healthyTask]
+    containerDependencies.finalizedBy = deephavenDocker.endTask
+    network = deephavenDocker.networkName.get()
+    parentContainers = [ project.tasks.getByName('rClient') ]
+    entrypoint = ["${prefix}/bin/rdeephaven/r-tests.sh", '/out/r-test.xml', '/out/r-test.log']
+}
+testRClient.configure {
```
niloc132, line 104: as in cpp-client/build.gradle, this is also unnecessary, the parentContainers= above handles it already

R/build.gradle
```diff
@@ -0,0 +1,106 @@
...
+
+    copyIn {
+        from(layout.projectDirectory) {
+            include 'r-build.sh'
+            include 'r-tests.sh'
+            include 'rdeephaven/DESCRIPTION'
+            include 'rdeephaven/LICENSE'
+            include 'rdeephaven/NAMESPACE'
+            include 'rdeephaven/README.md'
+            include 'rdeephaven/inst/**'
```
niloc132, line 41: if this is only tests, it should probably be removed here (and in the copyFile command), and instead added to the copyIn of testRClient, so that changing a test doesnt require rebuilding the rclient itself, just re-running the test

R/build.gradle
```diff
@@ -0,0 +1,107 @@
...
+        copyFile('rdeephaven/README.md',  "${prefix}/src/rdeephaven/")
+        copyFile('rdeephaven/inst/',  "${prefix}/src/rdeephaven/inst/")
+        copyFile('rdeephaven/man/', "${prefix}/src/rdeephaven/man/")
+        copyFile('rdeephaven/etc/', "${prefix}/src/rdeephaven/etc/")
+        copyFile('rdeephaven/R/', "${prefix}/src/rdeephaven/R/")
+        copyFile('rdeephaven/src/*.cpp', "${prefix}/src/rdeephaven/src/")
+        copyFile('rdeephaven/src/Makevars', "${prefix}/src/rdeephaven/src/")
+        copyFile('r-tests.sh', "${prefix}/bin/rdeephaven")
+        copyFile('r-build.sh', "${prefix}/bin/rdeephaven")
+        runCommand("PREFIX=\"" + prefix + "\"; \\\n" +
```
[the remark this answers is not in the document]
jcferretti: I simplified the syntax of that first line a bit trying to capture the spirit of your comment.

jcferretti:
> I'm missing some context. What ticket is this fixing? It looks like this is just running tests, is there a reason it needs to make this release?

Discussed with Mike over slack.   He said he is ok deferring to Devin and/or Ryan on the .github/workflows part of the change.

mofojed (commented):
I defer to Devin/Ryan for changes to the deephaven-core workflows.

alexpeters1208 (approved):
I only looked at the R files. LGTM.

devinrsmith (approved):
Verified the points I care about - gradle now properly fails when error in R :)

R/build.gradle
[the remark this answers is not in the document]
niloc132: The double quotes were a deliberate change on my part, to ensure that no shell variables existed at all in this code, but that the shell was finished while still in groovy.

That said, this is just intended as an observation/idea/suggestion and not a requirement for merging.

niloc132 (approved):

.github/workflows/tag-base-images.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       - name: Tag upstream images
         if: ${{ startsWith(github.ref, 'refs/heads/release/v') }}
         run: |
-          ./docker/registry/cpp-client-base/build/crane/retag.sh
```
chipkent, line 62: Did you mean to remove the cpp?

chipkent (approved):

.github/workflows/tag-base-images.yml
jcferretti: Yes, I Devin request I remove earlier in this same review.  The C++ build is now using the r-client-base image (which we plan to rename `cpp-clients-fat-base` on the py ticking coming PR).

Issue #4627, opened 2023-10-11
io.grpc.StatusRuntimeException: UNAVAILABLE: RST_STREAM closed stream. HTTP/2 error code: NO_ERROR


niloc132:
Using envoy as a proxy seems to prevent some streams from completing correctly - instead of receiving an `onCompleted` call, we get an error message and stack trace that looks like this:
```
io.grpc.StatusRuntimeException: UNAVAILABLE: RST_STREAM closed stream. HTTP/2 error code: NO_ERROR
	at io.grpc.Status.asRuntimeException(Status.java:539)
	at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:487)
	at io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
	at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
	at io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
```

This happens with either netty or jetty as our http server, but doesn't happen in grpc-web or our websocket transport. Watching envoy logs, we can confirm that envoy sees the `grpc-status: 0` trailer message, and seems to claim that it sends the headers to the client, but the client seems to only receive the RST_STREAM. Watching wireshark traffic, we can see that there are many DATA messages left unforwarded to the client, and that the HEADERS(end_stream=true) is also left behind, with only a RST_STREAM sent to the client and the upstream server.

The okhttp grpc client also gets an error here:
```
io.grpc.StatusRuntimeException: INTERNAL: No error: A GRPC status of OK should have been sent
Rst Stream
	at io.grpc.Status.asRuntimeException(Status.java:537)
	at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:481)
	at io.grpc.internal.DelayedClientCall$DelayedListener$3.run(DelayedClientCall.java:489)
	at io.grpc.internal.DelayedClientCall$DelayedListener.delayOrExecute(DelayedClientCall.java:453)
	at io.grpc.internal.DelayedClientCall$DelayedListener.onClose(DelayedClientCall.java:486)
	at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:567)
```

[...]

PR #4632, opened 2023-10-12
Always half-close DoExchange snapshot requests

niloc132:
Partial mitigation for #4627


rcaudy (approved):

nbauernfeind (approved):

niloc132:
Something is wrong with this, do not merge.

niloc132:
Okay, I misremembered how we implemented hierarchical table data - this actually is a subscription and not snapshot based, so it was incorrect to half close right away. As far as I can tell, the java client is the only official client that used the snapshot api in this way - JS api is the only other consumer, and it already half-closed right away (...and doesn't use h2 anyway for DoExchange).

rcaudy (approved):

Issue #4628, opened 2023-10-11
Parquet reading code crashes while reading arrays for this scenario


malhotrashivam:
Consider the following python scrip which creates a parquet file using pyarrow and then reads it in deephaven:
```
import pandas
import pyarrow.parquet
from deephaven.parquet import write, read

ARRAY_SIZE = 3
NUM_ROWS = 100000

numbers = list(range(ARRAY_SIZE))
numbersCol = []
for i in range(NUM_ROWS):
    numbersCol.append(numbers)

pa_table = pyarrow.table({'numbersCol': numbersCol})
pyarrow.parquet.write_table(pa_table, 'data_from_pa.parquet', compression="NONE", use_dictionary=False)

dh_table = read('data_from_pa.parquet').select()
```

It fails with the exception at 

> RuntimeError: java.lang.IndexOutOfBoundsException
> 	at java.base/java.nio.Buffer.checkIndex(Buffer.java:687)
> 	at java.base/java.nio.HeapIntBuffer.put(HeapIntBuffer.java:209)
> 	at io.deephaven.parquet.base.LevelsController$RepeatingLevel.addValues(LevelsController.java:165)
> 	at io.deephaven.parquet.base.LevelsController.addElements(LevelsController.java:56)
> 	at io.deephaven.parquet.base.ColumnPageReaderImpl.getOffsetsAndNulls(ColumnPageReaderImpl.java:645)
> 	at io.deephaven.parquet.base.ColumnPageReaderImpl.materializeWithNulls(ColumnPageReaderImpl.java:503)
> 	at io.deephaven.parquet.base.ColumnPageReaderImpl.materialize(ColumnPageReaderImpl.java:386)

Note that the exception goes away if I change NUM_ROWS from 100_000 to 10_000, or if I change array size from 3 to 2. So seems like something related to larger sizes.

(Issue found during testing for #4541 )

PR #4640, opened 2023-10-13
Added exception for an unsupported case while reading parquet file

malhotrashivam:
Related to #4628


engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -757,6 +757,7 @@ public Table aggAllBy(AggSpec spec, ColumnName... groupByColumns) {
      * @param tableColumns the table columns
      * @return the aggregation, if non-empty
      */
+    @VisibleForTesting
```
malhotrashivam, line 760: Pending change from https://github.com/deephaven/deephaven-core/pull/4513#discussion_r1333624721

rcaudy (approved):

PR #4648, opened 2023-10-16
Add KafkaTools.Consume#objectProcessorSpec for custom KeyOrValueSpecs

### LINKED ISSUE #4346: Extensible Kafka KeyOrValueSpecs
As a data administrator/developer, I want to be able to substitute my own key or value specifications instead of ones that Deephaven has built so that I can customize my ingestion for new formats or as a developer experiment with alternative processors  for existing formats.

The current architecture has switch statements that are not conducive to extending the ingester framework at runtime.

devinrsmith:
This introduces a new public interface, ObjectProcessor, that provides a means for end-users to create custom KeyOrValueSpec for use with Kafka ingestion. The interface isn't specific to kafka though, and should be extensible to other domains in the future.

The existing KeyOrValue specs are left unchanged; but arguably, there may be value in porting our existing implementations over to it in the future.

Fixes #4346


extensions/kafka/src/main/java/io/deephaven/kafka/KeyOrValueSpecObjectProcessorImpl.java
```diff
@@ -0,0 +1,129 @@
...
+ * take advantage of these better interfaces.
+ */
+class KeyOrValueSpecObjectProcessorImpl<T> extends KeyOrValueSpec {
+    private final Deserializer<? extends T> deserializer;
+    private final ObjectProcessor<? super T> processor;
+    private final List<String> columnNames;
+
+    KeyOrValueSpecObjectProcessorImpl(
+            Deserializer<? extends T> deserializer, ObjectProcessor<? super T> processor, List<String> columnNames) {
+        if (columnNames.size() != processor.outputTypes().size()) {
```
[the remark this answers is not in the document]
devinrsmith: So; your own processor implementation _can_ know about the column names, but I don't know if it makes sense to elevate that to the interface level. While it would simplify some calls sites (like this constructor), I think there are legitimate use cases where it won't make sense and would limit compose-ability.

As such, I _think_ the interface should be based around the minimal information needed for `void processAll(ObjectChunk<? extends T, ?> in, List<WritableChunk<?>> out)` to work.

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+     * Processes {@code in} into {@code out} by appending {@code in.size()} values to each chunk. The size of each
+     * {@code out} chunk will be incremented by {@code in.size()}. Implementations are free to process the data in a
+     * row-oriented, column-oriented, or mix-oriented fashion. Implementations must not keep any references to the
+     * passed-in chunks.
+     *
+     * <p>
+     * If an exception thrown the output chunks will be in an unspecified state.
+     *
+     * @param in the input objects
+     * @param out the output chunks as specified by {@link #outputTypes()}; each chunk must have remaining capacity of
```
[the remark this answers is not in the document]
devinrsmith: Not an unreasonable thought; that said, you may be keeping a large byte array in memory if you only have one small string output that references it. Either way, I should be more explicit about the *data* inside the chunks vs the chunks themselves. Let me think about this some more; the statement was really originally around not wanting the implementation to keep around a chunk that was later going to be released to a pool.

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessorRowLimitedImpl.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    // Ideally, these would be built into Chunk impls
+
+    private static <T, ATTR extends Any> Iterable<ObjectChunk<T, ATTR>> iterable(
+            ObjectChunk<T, ATTR> source, int sliceSize) {
+        if (source.size() <= sliceSize) {
+            return List.of(source);
+        }
+        // Note: we need to create the "for pool" version to guarantee we are getting a plain version; we know we don't
```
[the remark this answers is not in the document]
devinrsmith: So, `io.deephaven.chunk.ResettableObjectChunk#makeResettableChunk` doesn't actually use a pool by default; our property for that is false.

That said, I _could_ implemented this w/ a thread local pattern and/or with processAll handling resetFromTypedChunk to do without the Iterable allocation...

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+     * <td>All other {@link GenericType}</td>
+     * <td>{@link ChunkType#Object}</td>
+     * </tr>
+     * </table>
+     */
+    static ChunkType chunkType(Type<?> type) {
+        return ObjectProcessorTypes.of(type);
+    }
+
+    /**
```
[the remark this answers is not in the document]
devinrsmith: The types in TableDefinition are

```
  Class<TYPE> dataType,
  Class<?> componentType
```

- `Type` is our new canonical class for representing that information. I would very much like to change TableDefinition impl to use `Type<TYPE>`. Given reasons already mentioned, I don't think it's appropriate to have column name associations in the ObjectProcessor interface, so that's the reason for not using `TableDefinition`.

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessor.java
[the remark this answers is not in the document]
devinrsmith: This may also be a particularly relevant concern for Capn Proto format.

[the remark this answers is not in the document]
devinrsmith: That said, it's beginning to feel more like formula views than copying data into output chunks...

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessorRowBase.java
```diff
@@ -0,0 +1,49 @@
...
+ * A specialization that {@link #processAll(ObjectChunk, List) processes all} one row at a time.
+ *
+ * <p>
+ * In particular, this is a useful construct when {@code T} is a {@link java.nio.ByteBuffer} or {@code byte[]} type and
+ * it makes sense to parse straight from bytes into output chunks (as opposed to parsing into a more intermediate state
+ * and then processing into chunks in a column-oriented fashion).
+ *
+ * @param <T> the object type
+ */
+public abstract class ObjectProcessorRowBase<T> implements ObjectProcessorRowLimited<T> {
```
devinrsmith, line 21: I'm not sure this abstraction will actually be used; thinking more about how the raw avro handler was written, it first casts all of the writeable chunks and then proceeds row-by-row. As written, the implementations here will need to cast once per row / cell, and that's probably not desirable.

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessorStrict.java
```diff
@@ -0,0 +1,72 @@
...
+            throw new IllegalArgumentException(String.format(
+                    "Expected delegate.outputTypes().size() == out.size(). delegate.outputTypes().size()=%d, out.size()=%d",
+                    numColumns, out.size()));
+        }
+        final int[] originalSizes = new int[numColumns];
+        for (int i = 0; i < numColumns; ++i) {
+            final WritableChunk<?> chunk = out.get(i);
+            if (chunk.capacity() - chunk.size() < in.size()) {
+                throw new IllegalArgumentException(String.format(
+                        "out chunk does not have enough remaining capacity. i=%d, in.size()=%d, chunk.size()=%d, chunk.capacity()=%d",
```
rcaudy, line 48: What's `i`? A chunk index in `in`?

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessorStrict.java
```diff
@@ -0,0 +1,72 @@
...
+                        i, in.size(), chunk.size(), chunk.capacity()));
+            }
+            final Type<?> type = delegate.outputTypes().get(i);
+            final ChunkType expectedChunkType = ObjectProcessor.chunkType(type);
+            final ChunkType actualChunkType = chunk.getChunkType();
+            if (!expectedChunkType.equals(actualChunkType)) {
```
rcaudy, line 54:
```suggestion
            if (expectedChunkType != actualChunkType) {
```

extensions/kafka/src/main/java/io/deephaven/processor/functions/ObjectProcessorFunctionsImpl.java
```diff
@@ -0,0 +1,459 @@
...
+import java.util.Objects;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+final class ObjectProcessorFunctionsImpl<T> implements ObjectProcessor<T> {
+
+    interface Appender<T> {
+
+        void append(WritableChunk<?> dest, ObjectChunk<? extends T, ?> src);
```
rcaudy, line 59: Maybe reverse the order here, to match processAll?

extensions/kafka/src/main/java/io/deephaven/kafka/KeyOrValueSpecObjectProcessorImpl.java
```diff
@@ -0,0 +1,129 @@
...
+            // we except isInOrder to be true, so apply should be an O(1) op no matter how many columns there are.
+            processor.processAll(in, offsetsAdapter.apply(publisherChunks));
+        }
+    }
+
+    private static <T> Function<T[], List<T>> offsetsFunction(int[] offsets) {
+        return offsets.length == 0
+                ? array -> Collections.emptyList()
+                : isInOrder(offsets)
+                        ? array -> Arrays.asList(array).subList(offsets[0], offsets[0] + offsets.length)
```
rcaudy, line 109: I dislike this allocation. We could just advise of which `out` chunk index to start from.

devinrsmith: I'm very much against this idea. My hope is that `ObjectProcessor#processAll` is very straightforward. Adding indexes as part of that feels like a cop-out based on our current implementation. I very much expect that a re-write of some internal implementation details (KafkaStreamPublisher) and continuing along w/ our path of making KafkaTools better will make this whole adapter layer unnecessary in the future.

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessorRowBase.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Override
+    public final int rowLimit() {
+        return 1;
+    }
+
+    /**
+     * Implementations are responsible for parsing {@code in} into {@code out} <b>without</b> increasing the output
+     * chunk size; as such implementations are responsible for setting the cell at position {@code chunk.size() + i} for
```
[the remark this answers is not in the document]
devinrsmith: So, I think based on our current implementations, this "seems reasonable". Looking forward a little bit though, I've got ideas for allowing publishing on a 1-to-N basis. These generalized 1-to-N interfaces we'll still want to have an easy adapter on top of the 1-to-1 interface (ObjectProcessor); but I can _theoretically_ see the 1-to-N interfaces as preferring to create different sized chunks for different column types - ie, maybe we want 4096 sized chunks for int chunks, but 2048 sized chunks for long chunks.

I think if we really want to make the guarantee that all out chunks have the same initialSize, we _could_ add that guarantee in the future, as we are primarily the controllers of the publishing side. If we _do_ make that guarantee, I agree that ObjectProcessorRowBase could have a better interface. 

Regardless, I think that ObjectProcessorRowBase could almost always be better implemented by the end-user w/ a single cast. I'm going to remove it for now.

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessorRowLimited.java
```diff
@@ -12,9 +12,8 @@
  * {@link #processAll(ObjectChunk, List)}.
  *
  * <p>
- * For use-cases that are naturally performed one row at a time, callers are encouraged to extend
- * {@link ObjectProcessorRowBase}. For use-cases where the implementation can be columnar but callers would like to add
- * a row limit, callers are encouraged to create a columnar implementation against {@link ObjectProcessor} and use
+ * For use-cases where the implementation can be columnar but callers would like to add a row limit, callers are
```
rcaudy, line 15: This comment looks orphaned without the previous sentence. What about row-oriented use cases? Maybe delete the whole thing?

devinrsmith: I think it's still relevant. Essentially, it provides guidance around "should my implementation extend ObjectProcessorRowLimited?". If an implementation _can_ be column oriented, but you want to add a row limit for reasons outlined, you'll be encouraged to still create the column oriented version, but call into `ObjectProcessorRowLimited#of` instead of extending `ObjectProcessorRowLimited`..

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessorRowBase.java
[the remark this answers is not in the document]
cpwright: We actually should not want to require .add on chunks.  At least in profiles we've seen that it is worse to add to a chunk than to set its size once, and then fill it in.  If you were to have something that is "simple" like Captain proto where we are reading from one block per row and doing scattered writes to N others, then you are instead doing the read from one block, a scaterred read for the original size, a 2*N scattered writes.

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessorRowLimited.java
rcaudy: We discussed getting rid of the interface and squashing it into the impl. Should clarify that this is not a meaningful thing to do for row-oriented impls.

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3314
Reference: https://github.com/deephaven/deephaven.io/issues/3315

PR #4651, opened 2023-10-16
KafkaStreamPublisher null out input chunks after flushing

devinrsmith:


extensions/kafka/src/main/java/io/deephaven/kafka/ingest/KafkaStreamPublisher.java
```diff
@@ -189,6 +189,7 @@ public synchronized long consumeRecords(long receiveTime,
                         : null) {
             WritableObjectChunk<Object, Values> keyChunk;
             if (keyChunkCloseable != null) {
+                keyChunkCloseable.fillWithValue(0, keyChunkCloseable.size(), null);
```
rcaudy, line 192: At a minimum, the first two fills are redundant, I think.
The second two are also mostly-obviated by the behavior of the pool on close, but you could argue they do make objects unreachable a little bit earlier.

devinrsmith:
This was pointed out as "mostly unnecessary" do to the closing of the chunks.

PR #4661, opened 2023-10-17
Write parquet file headers to a bufferedOutputStream

malhotrashivam:
In the original code, we used a regular output stream for writing the parquet file header, which led to multiple `write` system calls for writing each field of the header. After this change, we use a `BufferedOutputStream` to buffer all the header data and flush it together.
Found during experiments for #946

Benchmark results:
Screenshot 2023-10-17 at 6 05 49 PM

### DIFF
--- a/extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnWriterImpl.java
+++ b/extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnWriterImpl.java
@@ -43,9 +43,14 @@ public class ColumnWriterImpl implements ColumnWriter {
     private static final int MIN_SLAB_SIZE = 64;
     private final SeekableByteChannel writeChannel;
     /**
-     * The buffered output stream {@link #bout} will stream data to the {@link #writeChannel}
+     * Buffered stream that writes to {@link #writeChannel}. Used for buffering small writes, particularly page headers.
      */
-    private final BufferedOutputStream bout;
+    private final BufferedOutputStream bufferedOutput;
+    /**
+     * Following is used to set the size of buffer for {@link #bufferedOutput}. In our testing, we found the page
+     * headers to be much smaller than 16 KB, so using that as the default size.
+     */
+    private static final int EXPECTED_PAGE_HEADER_SIZE = 16 << 10;
     private final ColumnDescriptor column;
     private final RowGroupWriterImpl owner;
     private final CompressorAdapter compressorAdapter;
@@ -79,12 +84,7 @@ public class ColumnWriterImpl implements ColumnWriter {
             final int targetPageSize,
             final ByteBufferAllocator allocator) {
         this.writeChannel = writeChannel;
-        // The buffered output stream is used to buffer the page header. Therefore, we set the size of the buffer as
-        // 16 KB because that is the expected page header size (ref
-        // https://github.com/apache/arrow/blob/ac581fd2a87b35c872cf334bb147851fe1287714/cpp/src/parquet/column_reader.h#L57)
-        // Also, in our testing, we found that the headers we write are much smaller than 16 KB.
-        final int EXPECTED_PAGE_HEADER_SIZE = 16 << 10;
-        bout = new BufferedOutputStream(Channels.newOutputStream(writeChannel), EXPECTED_PAGE_HEADER_SIZE);
+        bufferedOutput = new BufferedOutputStream(Channels.newOutputStream(writeChannel), EXPECTED_PAGE_HEADER_SIZE);
         this.column = column;
         this.compressorAdapter = compressorAdapter;
         this.targetPageSize = targetPageSize;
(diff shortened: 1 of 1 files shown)


extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnWriterImpl.java
```diff
@@ -41,6 +42,10 @@ public class ColumnWriterImpl implements ColumnWriter {
 
     private static final int MIN_SLAB_SIZE = 64;
     private final SeekableByteChannel writeChannel;
+    /**
+     * The buffered output stream {@link #bout} will stream data to the {@link #writeChannel}
+     */
+    private final BufferedOutputStream bout;
```
rcaudy, line 48:
```suggestion
    private final BufferedOutputStream bufferedOutput;
```

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnWriterImpl.java
```diff
@@ -41,6 +42,10 @@ public class ColumnWriterImpl implements ColumnWriter {
 
     private static final int MIN_SLAB_SIZE = 64;
     private final SeekableByteChannel writeChannel;
+    /**
+     * The buffered output stream {@link #bout} will stream data to the {@link #writeChannel}
```
rcaudy, line 46:
```suggestion
     * Buffered stream that writes to {@link #writeChannel}. Used for buffering small writes.
```

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnWriterImpl.java
```diff
@@ -74,6 +79,12 @@ public class ColumnWriterImpl implements ColumnWriter {
             final int targetPageSize,
             final ByteBufferAllocator allocator) {
         this.writeChannel = writeChannel;
+        // The buffered output stream is used to buffer the page header. Therefore, we set the size of the buffer as
+        // 16 KB because that is the expected page header size (ref
+        // https://github.com/apache/arrow/blob/ac581fd2a87b35c872cf334bb147851fe1287714/cpp/src/parquet/column_reader.h#L57)
```
rcaudy, line 84: It seems like this is the default unit of growth for page size buffers in the arrow reading code. I would rather not reference another project from our code, unless it's Parquet or Java, etc. If you think our usual size is much smaller, we might want to default to something smaller, but I'm not uncomfortable with 16KB.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnWriterImpl.java
```diff
@@ -74,6 +79,12 @@ public class ColumnWriterImpl implements ColumnWriter {
...
+        // Also, in our testing, we found that the headers we write are much smaller than 16 KB.
+        final int EXPECTED_PAGE_HEADER_SIZE = 16 << 10;
```
rcaudy, line 86: I would make this a static constant, rather than in-line in this method.

rcaudy (commented):
Minor comments only. Looks pretty good.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnWriterImpl.java
malhotrashivam: Leaving the buffer size as 16 KB just to be safe side.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnWriterImpl.java
```diff
@@ -43,9 +43,14 @@ public class ColumnWriterImpl implements ColumnWriter {
...
-     * The buffered output stream {@link #bout} will stream data to the {@link #writeChannel}
+     * Buffered stream that writes to {@link #writeChannel}. Used for buffering small writes, particularly page headers.
      */
-    private final BufferedOutputStream bout;
+    private final BufferedOutputStream bufferedOutput;
+    /**
+     * Following is used to set the size of buffer for {@link #bufferedOutput}. In our testing, we found the page
+     * headers to be much smaller than 16 KB, so using that as the default size.
+     */
+    private static final int EXPECTED_PAGE_HEADER_SIZE = 16 << 10;
```
rcaudy, line 53: I would group this with the other static constants, but not worth pushing another commit and restarting CI.

rcaudy (approved):

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnWriterImpl.java
malhotrashivam: Will add this change in my next PR

PR #4669, opened 2023-10-19
Added buffered stream for all parquet writes

malhotrashivam:
Related to #946


extensions/parquet/base/src/main/java/io/deephaven/parquet/base/BufferedStreamOverWriteChannel.java
```diff
@@ -0,0 +1,24 @@
...
+import java.io.BufferedOutputStream;
+import java.io.IOException;
+import java.nio.channels.Channels;
+import java.nio.channels.SeekableByteChannel;
+
+final class BufferedStreamOverWriteChannel extends BufferedOutputStream {
+
+    private final SeekableByteChannel writeChannel;
+
+    BufferedStreamOverWriteChannel(final SeekableByteChannel writeChannel, final int size) {
```
malhotrashivam, line 12: Please suggest a better name

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileWriter.java
```diff
@@ -18,22 +18,19 @@
...
 import java.util.List;
 import java.util.Map;
 
 import static org.apache.parquet.format.Util.writeFileMetaData;
 
-public class ParquetFileWriter {
+public final class ParquetFileWriter {
     private static final ParquetMetadataConverter metadataConverter = new ParquetMetadataConverter();
     private static final int VERSION = 1;
+    private static final int OUTPUT_BUFFER_SIZE = 1 << 18; // TODO Benchmark this
```
malhotrashivam, line 31: Benchmark this

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileWriter.java
```diff
@@ -54,67 +51,70 @@ public ParquetFileWriter(
         this.targetPageSize = targetPageSize;
         this.allocator = allocator;
         this.extraMetaData = new HashMap<>(extraMetaData);
-        writeChannel = channelsProvider.getWriteChannel(filePath, false); // TODO add support for appending
-        writeChannel.write(ByteBuffer.wrap(ParquetFileReader.MAGIC));
+        // TODO add support for appending
```
rcaudy, line 54: Remove the TODO, or provide a link to a GitHub issue for it.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileWriter.java
```diff
@@ -54,67 +51,70 @@ public ParquetFileWriter(
...
+                OUTPUT_BUFFER_SIZE);
+        bufferedOutput.write(ParquetFileReader.MAGIC);
         this.type = type;
         this.channelsProvider = channelsProvider;
         this.compressorAdapter = DeephavenCompressorAdapterFactory.getInstance().getByName(codecName);
     }
 
+    /**
+     * Following method is unused and untested.
+     */
```
rcaudy, line 65: Maybe don't JavaDoc that. ;-)

I am 98% sure that this method is for adding a row group to a different file that will be referenced from the footer of the "main" file. Parquet allows this, we technically support reading it, but I've never seen it in the wild. It might be useful for parallel writing. I am suspicious that nothing is recording the offset indices.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileWriter.java
```diff
@@ -54,67 +51,70 @@ public ParquetFileWriter(
...
     @SuppressWarnings("unused")
     RowGroupWriter addRowGroup(final String path, final boolean append) throws IOException {
-        RowGroupWriterImpl rowGroupWriter =
-                new RowGroupWriterImpl(path, append, channelsProvider, type, targetPageSize, allocator,
-                        compressorAdapter);
+        bufferedOutput.close();
+        bufferedOutput =
+                new BufferedStreamOverWriteChannel(channelsProvider.getWriteChannel(path, append), OUTPUT_BUFFER_SIZE);
```
rcaudy, line 70: This isn't the appropriate thing to do. You should pass a new channel to the new writer.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileWriter.java
```diff
@@ -18,22 +18,19 @@
...
 import static org.apache.parquet.format.Util.writeFileMetaData;
 
-public class ParquetFileWriter {
+public final class ParquetFileWriter {
     private static final ParquetMetadataConverter metadataConverter = new ParquetMetadataConverter();
     private static final int VERSION = 1;
+    private static final int OUTPUT_BUFFER_SIZE = 1 << 18; // TODO Benchmark this
 
-    private final SeekableByteChannel writeChannel;
+    private BufferedStreamOverWriteChannel bufferedOutput;
```
rcaudy, line 33: `final` (if you are convinced by my other comments)

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileWriter.java
malhotrashivam: This is what was happening earlier as well in the file RowGroupWriterImpl.java [here]( https://github.com/deephaven/deephaven-core/pull/4669/files/50dfd20d8595efaf3783583690aaa2a28a4f5a7c#diff-cbfe79bd13d25860ebfbea418157f838ea092c5562ff57fa82f7ad74df5296b0L40:~:text=this(channelsProvider.getWriteChannel(path%2C%20append)%2C%20type%2C%20targetPageSize%2C%20allocator%2C%20blockWithPath(path)%2C ). I just moved it up one layer.
Do you want me to change it?

I added a close in line 68 so that we close and flush the old channel and create a new one.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/BufferedStreamOverWriteChannel.java
malhotrashivam: Interestingly there already exist a class `PositionOutputStream` that does exactly this. But the challenging part is that all of these are abstract classes and not interface (even `OutputStream` is abstract). And I need to extend from `BufferedOutputStream` since I need to access the `count` field. And I couldn't extend more than one abstract class.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileWriter.java
```diff
@@ -54,67 +51,66 @@ public ParquetFileWriter(
...
         this.channelsProvider = channelsProvider;
         this.compressorAdapter = DeephavenCompressorAdapterFactory.getInstance().getByName(codecName);
     }
 
     @SuppressWarnings("unused")
     RowGroupWriter addRowGroup(final String path, final boolean append) throws IOException {
-        RowGroupWriterImpl rowGroupWriter =
-                new RowGroupWriterImpl(path, append, channelsProvider, type, targetPageSize, allocator,
-                        compressorAdapter);
+        bufferedOutput.close();
```
rcaudy, line 64: This should also validate that `path` is not equal to `filePath`.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileWriter.java
```diff
@@ -54,67 +51,66 @@ public ParquetFileWriter(
...
+        bufferedOutput =
+                new PositionedBufferedOutputStream(channelsProvider.getWriteChannel(path, append), OUTPUT_BUFFER_SIZE);
+        final RowGroupWriter rowGroupWriter =
+                new RowGroupWriterImpl(bufferedOutput, type, targetPageSize, allocator, compressorAdapter, path);
```
rcaudy, line 68:
```suggestion
        if (Objects.equals(filePath, path)) {
            throw new IllegalArgumentException(String.format(
                    "Cannot add a new row group with the same path %s as the primary file", path));
        }
        final RowGroupWriter rowGroupWriter = new RowGroupWriterImpl(
                new PositionedBufferedOutputStream(channelsProvider.getWriteChannel(path, append), OUTPUT_BUFFER_SIZE),
                type, targetPageSize, allocator, compressorAdapter, path);
```

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileWriter.java
rcaudy, line 64: Maybe we should delete this method, since it's unused and we both have correctness questions about it?

rcaudy (approved):

PR #4673, opened 2023-10-19
Add Groovy container w/ examples and Redpanda

jjbrosnan:
We're adding Kafka Groovy docs, so having this to link to will be helpful, even if it's basically identical to the Python one.


containers/groovy-examples-redpanda/README.md
```diff
@@ -0,0 +1,24 @@
...
+For instructions on launching and using Kafka, see [Kafka in Deephaven: an introduction](https://deephaven.io/core/docs/conceptual/kafka-in-deephaven/) and [Simple Kafka import](https://deephaven.io/core/docs/how-to-guides/kafka-simple/).  For full instructions to work with Deephaven, see the [Quick start](https://deephaven.io/core/docs/tutorials/quickstart).
+
+To launch Deephaven, execute the following in your deployment directory:
+
+```bash
+compose_file=https://raw.githubusercontent.com/deephaven/deephaven-core/main/containers/groovy-examples-redpanda/docker-compose.yml
+curl  -O "${compose_file}"
+
+docker-compose pull
+docker-compose up -d
```
devinrsmith, line 23: We should not produce new documentation w/ `docker-compose`, `docker compose` is the correct way.

containers/groovy-examples-redpanda/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+    - PLAINTEXT://0.0.0.0:29092,OUTSIDE://0.0.0.0:9092
+    - --advertise-kafka-addr
+    - PLAINTEXT://redpanda:29092,OUTSIDE://localhost:9092
+    image: docker.redpanda.com/vectorized/redpanda:v22.2.2
+    ports:
+    - 9092:9092
+    - 29092:29092
+
+  registry:
+    image: apicurio/apicurio-registry-mem:1.2.2.Final
```
devinrsmith, line 40: We should update our code - we shouldn't need apicurio registry; it's built into redpanda.

jjbrosnan:
This PR now not only adds a Groovy-examples-redpanda docker compose deployment, it also adds:

- A Python + Redpanda deployment without examples
- A Groovy + Redpanda deployment without examples

It also removes any `docker-compose` shell commands in favor of docker compose v2 syntax, `docker compose`.

jjbrosnan:
I should also mention this removes our All-AI deployments, as we decided to no longer publish those images.

containers/python-redpanda/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    - '1'
+    - --reserve-memory
+    - 0M
+    - --overprovisioned
+    - --node-id
+    - '0'
+    - --kafka-addr
+    - PLAINTEXT://0.0.0.0:29092,OUTSIDE://0.0.0.0:9092
+    - --advertise-kafka-addr
+    - PLAINTEXT://redpanda:29092,OUTSIDE://localhost:9092
```
devinrsmith, line 27: I think we should adopt the formatting style redpanda uses for this:

```yaml
    command:
      - redpanda
      - start
      - --kafka-addr internal://0.0.0.0:9092,external://0.0.0.0:19092
      - --advertise-kafka-addr internal://redpanda:9092,external://localhost:19092
      - --schema-registry-addr internal://0.0.0.0:8081,external://0.0.0.0:18081
      - --smp 1
      - --memory 1G
      - --mode dev-container
```

I've modified this from https://docs.redpanda.com/current/get-started/quick-start/; we should test it in all of the different compose files to make sure it works. Do we use the schema registry for any of our examples?

devinrsmith (changes requested):
Couple of comments that I think apply to a lot of files - didn't repeat the advice for every single file though...

containers/python-redpanda/docker-compose.yml
jjbrosnan: I don't think we do. I've made updates that make our syntax consistent with Redpanda and tested it to make sure it works with our Kafka documentation.

devinrsmith: I don't think we're consistent w/ the redpanda quickstart - they use `internal` and `external`, and use `dev-container`.

devinrsmith (commented):
See item I re-opened.

devinrsmith (approved):
Assuming you've run these examples successfully, LGTM.

mattrunyon:
Run the action to run all examples against the edge container in CI to verify the nightly task isn't going to fail.

PR #4674, opened 2023-10-19
Clean up some legacy code and improve build

### LINKED ISSUE #3962: Eliminate `:Net` from Build
There are only a few classes in `:Net`, but they are barely hanging on. This is code that we intended to delete (that Ryan and I both thought was already deleted). These files are `CommBase`, `FastNIODriver`, and `NIODriver`.

`NIODriver and `FastNIODriver` are only used by `CommBase`. `CommBase` is used by:
- `PeriodicUpdateGraph` - this is being used as a ScheduledExecutorService to facilitate watch dog alerting; we could potentially add one to the builder and pass along via Dagger.
- `StatsDriver` - this is being used as a ScheduledExecutorService and we initialize this during `EngineMetrics#maybeStartStatsCollection`, so we could pass one along.
- `TrackedFileHandleFactory` is also using it as a ScheduledExecutorService, but it is being statically initialized. It might be easiest to create a new single threaded executor and not bother with the static initialization.

We should consider dropping the `YASchedulerImpl` and `Scheduler` classes from `:IO` as well.

### LINKED ISSUE #4460: Clean up FishUtil
The top-level project FishUtil has a lot of code that appears to be dead:
 * `cpp/` appears to be entirely unused - the MicroTimer java code no longer exists, and the SignalUtils java classes are unused
 * `src/main/java/io/deephaven/util/`
     * `formatters/ISO8601.java` is only used by StatsDriver, and then only to access `ISO8601.serverTimeZone()`, so need not exist any more, could be moved and reduced or just inlined.
     * `signals/` appears entirely unused
     * `threads/ThreadDump.java` is used in `process.ShutdownManagerImpl`, but only the `ThreadDump.threadDump() method ` is used, could be moved/inlined.
     * `OSUtil` is used only by StatsCPUCollector, and then only to see if mac or windows is being used.
     * DateUtil, ExceptionUtil, Mailer, ThreadSafeDateFormat, Validate classes are not used.
     * The `process` package is still used, perhaps should live somewhere else.

rcaudy:
* Addresses a build issue on ARM macs by specifying the platform for the proto-backplane-grpc image in proto/raw-js-openapi/Dockerfile
* Closes #4460 by deleting, inlining, or moving FishUtil code and deleting the FishUtil module
* Closes #3962 by deleting usages of :Net code


proto/raw-js-openapi/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM deephaven/proto-backplane-grpc:local-build AS proto-backplane-grpc
+FROM --platform=linux/amd64 deephaven/proto-backplane-grpc:local-build AS proto-backplane-grpc
```
devinrsmith, line 1: cc @niloc132 . I debugged w/ Ryan a while ago, and we needed to make this change. I'm guessing some change in default Docker behavior.

Stats/src/main/java/io/deephaven/stats/StatsDriver.java
```diff
@@ -150,23 +159,30 @@ public StatsDriver(Clock clock, StatsIntradayLogger intraday, boolean getFdStats
             clockValue = null;
         }
 
-        long now = System.currentTimeMillis();
-        long delay = STEP - (now % STEP);
-        nextInvocation = now + delay;
+        final long now = System.currentTimeMillis();
+        final long delay = STEP - (now % STEP);
+        nextCpuUpdate = now + delay + CPU_INTERVAL;
+        nextMemUpdate = now + delay + MEM_INTERVAL;
```
devinrsmith, line 165: This appears to be a slight change in logic from before - approximately, the previous logic was `now + <X>_INTERVAL`. Assuming this was intentional?

rcaudy: It was intentional, yes. Basically, it didn't make sense (to me) that when we next call `update` (fka `timedOut`) and evaluate these, they'll be `delay` behind, and continue to be `delay` behind for the life of the driver.

The whole thing is a little on the silly side, since the intervals match the step. Arguably, I should just remove them and make updatings CPU and memory stats unconditional.

proto/raw-js-openapi/Dockerfile
rcaudy: After a discussion with @niloc132 , I have reverted this change and cherry-picked a better change from his fork.

devinrsmith (approved):

niloc132 (approved):
Approved for changes to proto/raw-js-openapi/Dockerfile

PR #4676, opened 2023-10-19
`BarrageSession` Subscription/Snapshot Methods now Return `Future<Table>`

### LINKED ISSUE #4675: Barrage snapshot can return 0 rows to the client
**Description**
When trying to complete a Barrage snapshot,  you can get a table with 0 rows in it instead of the actual data.

**Steps to reproduce**
In one Community session create a table
```groovy
import static io.deephaven.engine.util.TableTools.*;
t = newTable(intCol("Sentinel", 10, 20, 30), stringCol("S2", "40", "50", "or more"))
```

In another worker try to fetch a snapshot
```groovy
import io.grpc.netty.NettyChannelBuilder
import io.deephaven.proto.DeephavenChannelImpl
import io.deephaven.client.impl.BarrageSession
import io.deephaven.client.impl.SessionImpl
import io.deephaven.client.impl.SessionImplConfig
import java.util.concurrent.Executors
import org.apache.arrow.memory.RootAllocator
import io.deephaven.qst.table.TicketTable
import io.deephaven.server.uri.BarrageTableResolver

managedChannel = NettyChannelBuilder.forTarget("dh://REMOTE_HOST").build()
dhChannel = new DeephavenChannelImpl(managedChannel)
session = SessionImpl.create(SessionImplConfig.builder()
    .executor(Executors.newScheduledThreadPool(4))
    .channel(dhChannel)
    .build())
barrage = BarrageSession.of(session, new RootAllocator(), managedChannel)

sub = barrage.snapshot(TicketTable.of("s/t"), BarrageTableResolver.SNAP_OPTIONS)
table = sub.entireTable()
```

**Expected results**

You get an identical table to t

**Actual results**

You get a blank table.

**Versions**
 - Deephaven: 0.29.0

nbauernfeind:
Fix #4675 

This includes breaking changes to the barrage java-client API. The snapshot/subscription methods now return `Future<Table>` instead of `BarrageTable` and users are no longer required to `blockUntilCompletion` as this is handled by `Future#get`.

Subscribed tables can now be canceled only through `Liveness` destruction such as using an explicit `LivenessScope` or invoking `ReferenceCounted#forceReferenceCountToZero()`.

### DIFF
--- a/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphAwareCompletableFuture.java
+++ b/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphAwareCompletableFuture.java
@@ -1,0 +1,164 @@
+package io.deephaven.engine.updategraph;
+
+import io.deephaven.util.SafeCloseable;
+import io.deephaven.util.function.ThrowingSupplier;
+import io.deephaven.util.locks.FunctionalLock;
+import io.deephaven.util.locks.FunctionalReentrantLock;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.concurrent.locks.Condition;
+
+public class UpdateGraphAwareCompletableFuture<T> implements Future<T> {
+
+    private final UpdateGraph updateGraph;
+
+    /** This condition is used to signal any threads waiting on the UpdateGraph exclusive lock. */
+    private volatile Condition updateGraphCondition;
+
+    private final FunctionalLock lock = new FunctionalReentrantLock();
+    private volatile Condition lockCondition;
+
+    private volatile ThrowingSupplier<T, ExecutionException> resultSupplier;
+    @SuppressWarnings("rawtypes")
+    private static final AtomicReferenceFieldUpdater<UpdateGraphAwareCompletableFuture, ThrowingSupplier> RESULT_UPDATER =
+            AtomicReferenceFieldUpdater.newUpdater(
+                    UpdateGraphAwareCompletableFuture.class, ThrowingSupplier.class, "resultSupplier");
+
+    /** The encoding of the cancelled supplier. */
+    private static final ThrowingSupplier<?, ExecutionException> CANCELLATION_SUPPLIER = () -> {
+        throw new CancellationException();
+    };
+
+    public UpdateGraphAwareCompletableFuture(@NotNull final UpdateGraph updateGraph) {
+        this.updateGraph = updateGraph;
+    }
+
+    ////////////////
+    // Future API //
+    ////////////////
+    @Override
+    public boolean cancel(boolean mayInterruptIfRunning) {
+        // noinspection unchecked
+        return trySignalCompletion((ThrowingSupplier<T, ExecutionException>) CANCELLATION_SUPPLIER);
+    }
+
+    @Override
+    public boolean isCancelled() {
+        return resultSupplier == CANCELLATION_SUPPLIER;
+    }
+
+    @Override
+    public boolean isDone() {
+        return resultSupplier != null;
+    }
+
+    @Override
+    public T get() throws InterruptedException, ExecutionException {
+        if (resultSupplier != null) {
+            return resultSupplier.get();
+        }
+        try {
+            return getInternal(0, null);
+        } catch (TimeoutException toe) {
+            throw new IllegalStateException("Unexpected TimeoutException", toe);
+        }
+    }
+
+    @Override
+    public T get(final long timeout, @NotNull final TimeUnit unit)
+            throws InterruptedException, ExecutionException, TimeoutException {
+        if (resultSupplier != null) {
+            return resultSupplier.get();
+        }
+        if (timeout <= 0) {
+            throw new TimeoutException();
+        }
+        return getInternal(timeout, unit);
+    }
+
+    private T getInternal(long timeout, @Nullable TimeUnit unit)
+            throws InterruptedException, ExecutionException, TimeoutException {
+        // test lock conditions
+        if (updateGraph.sharedLock().isHeldByCurrentThread()) {
+            throw new UnsupportedOperationException(
+                    "Cannot Future#get while holding the " + updateGraph + " shared lock");
+        }
+
+        final boolean holdingUpdateGraphLock = updateGraph.exclusiveLock().isHeldByCurrentThread();
+        if (updateGraphCondition == null && holdingUpdateGraphLock) {
+            try (final SafeCloseable ignored = lock.lockCloseable()) {
+                if (updateGraphCondition == null) {
+                    updateGraphCondition = updateGraph.exclusiveLock().newCondition();
+                }
+            }
+        } else if (lockCondition == null) {
+            try (final SafeCloseable ignored = lock.lockCloseable()) {
+                if (lockCondition == null) {
+                    lockCondition = lock.newCondition();
+                }
+            }
+        }
+
+        if (holdingUpdateGraphLock) {
+            waitForResult(updateGraphCondition, timeout, unit);
+        } else {
+            try (final SafeCloseable ignored = lock.lockCloseable()) {
+                waitForResult(lockCondition, timeout, unit);
+            }
+        }
+
+        return resultSupplier.get();
+    }
+
+    private void waitForResult(final Condition condition, final long timeout, @Nullable final TimeUnit unit)
+            throws InterruptedException, TimeoutException {
+        if (unit == null) {
+            while (resultSupplier == null) {
+                condition.await();
+            }
+            return;
+        }
+
+        long nanosLeft = unit.toNanos(timeout);
+        while (resultSupplier == null) {
+            nanosLeft = condition.awaitNanos(nanosLeft);
+            if (nanosLeft <= 0) {
+                throw new TimeoutException();
+            }
+        }
+    }
+
+    ////////////////////////////////////////////////////
+    // Completion API modeled after CompletableFuture //
+    ////////////////////////////////////////////////////
+
+    public boolean complete(T value) {
+        return trySignalCompletion(() -> value);
+    }
+
+    public boolean completeExceptionally(Throwable ex) {
+        if (ex == null)
+            throw new NullPointerException();
+        return trySignalCompletion(() -> {
+            throw new ExecutionException(ex);
+        });
+    }
+
+    private boolean trySignalCompletion(@NotNull final ThrowingSupplier<T, ExecutionException> result) {
+        if (!RESULT_UPDATER.compareAndSet(UpdateGraphAwareCompletableFuture.this, null, result)) {
+            return false;
+        }
+
+        final Condition localCondition = updateGraphCondition;
+        try (final SafeCloseable ignored = lock.lockCloseable()) {
+            if (localCondition != null) {
+                updateGraph.requestSignal(localCondition);
+            }
+        }
+
+        return true;
+    }
+}
--- a/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphAwareCompletableFuture.java
+++ b/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphAwareCompletableFuture.java
@@ -138,8 +138,9 @@ public class UpdateGraphAwareCompletableFuture<T> implements Future<T> {
     }
 
     public boolean completeExceptionally(Throwable ex) {
-        if (ex == null)
+        if (ex == null) {
             throw new NullPointerException();
+        }
         return trySignalCompletion(() -> {
             throw new ExecutionException(ex);
         });
--- a/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphAwareCompletableFuture.java
+++ b/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphAwareCompletableFuture.java
@@ -7,6 +7,7 @@ import io.deephaven.util.locks.FunctionalReentrantLock;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.Objects;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.concurrent.locks.Condition;
@@ -57,6 +58,8 @@ public class UpdateGraphAwareCompletableFuture<T> implements Future<T> {
 
     @Override
     public T get() throws InterruptedException, ExecutionException {
+        checkSharedLockState();
+
         if (resultSupplier != null) {
             return resultSupplier.get();
         }
@@ -70,6 +73,8 @@ public class UpdateGraphAwareCompletableFuture<T> implements Future<T> {
     @Override
     public T get(final long timeout, @NotNull final TimeUnit unit)
             throws InterruptedException, ExecutionException, TimeoutException {
+        checkSharedLockState();
+
         if (resultSupplier != null) {
             return resultSupplier.get();
         }
@@ -81,12 +86,6 @@ public class UpdateGraphAwareCompletableFuture<T> implements Future<T> {
 
     private T getInternal(final long timeout, @Nullable final TimeUnit unit)
             throws InterruptedException, ExecutionException, TimeoutException {
-        // test lock conditions
-        if (updateGraph.sharedLock().isHeldByCurrentThread()) {
-            throw new UnsupportedOperationException(
-                    "Cannot Future.get(...) while holding the " + updateGraph + " shared lock");
-        }
-
         final boolean holdingUpdateGraphLock = updateGraph.exclusiveLock().isHeldByCurrentThread();
         if (holdingUpdateGraphLock) {
             if (updateGraphCondition == null) {
@@ -109,6 +108,13 @@ public class UpdateGraphAwareCompletableFuture<T> implements Future<T> {
         }
 
         return resultSupplier.get();
+    }
+
+    private void checkSharedLockState() {
+        if (updateGraph.sharedLock().isHeldByCurrentThread()) {
+            throw new UnsupportedOperationException(
+                    "Cannot Future.get(...) while holding the " + updateGraph + " shared lock");
+        }
     }
 
     private void waitForResult(final Condition condition, final long timeout, @Nullable final TimeUnit unit)
@@ -138,9 +144,7 @@ public class UpdateGraphAwareCompletableFuture<T> implements Future<T> {
     }
 
     public boolean completeExceptionally(Throwable ex) {
-        if (ex == null) {
-            throw new NullPointerException();
-        }
+        Objects.requireNonNull(ex);
         return trySignalCompletion(() -> {
             throw new ExecutionException(ex);
         });
--- a/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphAwareCompletableFuture.java
+++ b/engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphAwareCompletableFuture.java
@@ -79,20 +79,18 @@ public class UpdateGraphAwareCompletableFuture<T> implements Future<T> {
         return getInternal(timeout, unit);
     }
 
-    private T getInternal(long timeout, @Nullable TimeUnit unit)
+    private T getInternal(final long timeout, @Nullable final TimeUnit unit)
             throws InterruptedException, ExecutionException, TimeoutException {
         // test lock conditions
         if (updateGraph.sharedLock().isHeldByCurrentThread()) {
             throw new UnsupportedOperationException(
-                    "Cannot Future#get while holding the " + updateGraph + " shared lock");
+                    "Cannot Future.get(...) while holding the " + updateGraph + " shared lock");
         }
 
         final boolean holdingUpdateGraphLock = updateGraph.exclusiveLock().isHeldByCurrentThread();
-        if (updateGraphCondition == null && holdingUpdateGraphLock) {
-            try (final SafeCloseable ignored = lock.lockCloseable()) {
-                if (updateGraphCondition == null) {
-                    updateGraphCondition = updateGraph.exclusiveLock().newCondition();
-                }
+        if (holdingUpdateGraphLock) {
+            if (updateGraphCondition == null) {
+                updateGraphCondition = updateGraph.exclusiveLock().newCondition();
             }
         } else if (lockCondition == null) {
             try (final SafeCloseable ignored = lock.lockCloseable()) {
@@ -148,15 +146,17 @@ public class UpdateGraphAwareCompletableFuture<T> implements Future<T> {
     }
 
     private boolean trySignalCompletion(@NotNull final ThrowingSupplier<T, ExecutionException> result) {
-        if (!RESULT_UPDATER.compareAndSet(UpdateGraphAwareCompletableFuture.this, null, result)) {
+        if (!RESULT_UPDATER.compareAndSet(this, null, result)) {
             return false;
         }
 
-        final Condition localCondition = updateGraphCondition;
-        try (final SafeCloseable ignored = lock.lockCloseable()) {
-            if (localCondition != null) {
-                updateGraph.requestSignal(localCondition);
-            }
+        final Condition localUpdateGraphCondition = updateGraphCondition;
+        if (localUpdateGraphCondition != null) {
+            updateGraph.requestSignal(localUpdateGraphCondition);
+        }
+        final Condition localLockCondition = lockCondition;
+        if (localLockCondition != null) {
+            lock.doLocked(localLockCondition::signalAll);
         }
 
         return true;
--- a/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/table/BarrageTable.java
+++ b/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/table/BarrageTable.java
@@ -396,6 +396,24 @@ public abstract class BarrageTable extends QueryTable implements BarrageMessage.
         }
     }
 
+    public void unsubscribe() {
+        unsubscribed = true;
+
+        if (!isRefreshing()) {
+            try {
+                realRefresh();
+            } catch (Throwable err) {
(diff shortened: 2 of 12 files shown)


java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
```diff
@@ -41,6 +42,30 @@
...
+ * It is not an error to create derived tables from the {@code BarrageTable} prior to the subscription being complete,
+ * as all changes are propagated to downstream tables.
+ * <p>
+ * For Snapshots (static tables):
+ * <p>
+ * Users may call {@code snapshotEntireTable}, or {@code snapshotPartialTable}, to initiate the gRPC call to the server.
+ * These methods return the eventually populated {@code BarrageTable} to the user. The user must either set
+ * {@code blockUntilComplete} to {@code true} during initiation or call {@code blockUntilComplete} to ensure that the
+ * {@code BarrageTable} is fully populated before using it. Noting that snapshots are static tables that do not
+ * propagate changes on any update graph.
```
rcaudy, line 67:
```suggestion
 * This class is an intermediary helper class that uses a {@code DoExchange} to populate a {@link BarrageTable} using
 * subscription data from a remote server, propagating updates if the request is a subscription.
 * <p>
 * For Subscriptions (refreshing result tables):
 * <p>
 * Users may call {@link #entireTable} or {@link #partialTable} to initiate the gRPC call to the server. These methods
 * return the eventually populated {@code BarrageTable} to the user.
 * <p>
 * If the user wants to ensure that the table is completely populated with an initial state of the remote table prior to
 * using the result they must either set {@code blockUntilComplete} to {@code true} or call
 * {@link #blockUntilComplete()} to ensure that the {@code BarrageTable} is respecting the requested subscription.
 * <p>
 * It is not an error to create derived tables from the {@code BarrageTable} prior to the subscription being complete,
 * as all changes are propagated to downstream tables.
 * <p>
 * For Snapshots (static result tables):
 * <p>
 * Users may call {@link #snapshotEntireTable} or {@link #snapshotPartialTable} to initiate the gRPC call to the server.
 * These methods return the eventually populated {@code BarrageTable} to the user. The user must either set
 * {@code blockUntilComplete} to {@code true} during initiation or call {@link #blockUntilComplete()} to ensure that the
 * {@code BarrageTable} is fully populated before using it. Noting that snapshots are static tables that do not
 * propagate changes on any update graph.
```

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
```diff
@@ -37,6 +37,7 @@
 import java.nio.ByteBuffer;
 import java.util.BitSet;
 import java.util.concurrent.ScheduledExecutorService;
```
rcaudy, line 39: We have unused imports.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
```diff
@@ -154,17 +156,18 @@ public void onError(final Throwable t) {
...
-            handleDisconnect();
+
+            // this error will always be propagated to our CheckForCompletion#onError callback
+            resultTable.handleBarrageError(t);
         }
 
         @Override
         public void onCompleted() {
-            handleDisconnect();
+            // this will always be propagated to our CheckForCompletion#onClose callback
```
rcaudy, line 169: We don't need to record that we're not connected here?

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
```diff
@@ -274,41 +277,37 @@ protected void destroy() {
...
+     * This method will only be invoked once using the CAS of {@code connected} from 1 to 0 as a guard.
+     */
+    private synchronized void signalCompletion(@Nullable final Throwable t) {
         completed = true;
-        signalCompletion();
-        cleanup();
-    }
 
-    private synchronized void signalCompletion() {
+        if (t != null) {
```
rcaudy, line 290: Same kind of race condition as in subscribe.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/table/BarrageTable.java
```diff
@@ -242,8 +233,8 @@ public BitSet getServerColumns() {
 
     @Override
     public void handleBarrageMessage(final BarrageMessage update) {
-        if (unsubscribed) {
-            beginLog(LogLevel.INFO).append(": Discarding update for unsubscribed table!").endl();
+        if (pendingError != null) {
```
rcaudy, line 236:
```suggestion
        if (pendingError != null || isFailed()) {
```

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/table/BarrageTable.java
```diff
@@ -337,34 +329,19 @@ protected boolean isSubscribedColumn(int i) {
     }
 
     private synchronized void realRefresh() {
+        if (isFailed()) {
+            return;
```
rcaudy, line 333: We may want to discard pending updates here.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/table/BarrageTable.java
```diff
@@ -271,6 +259,13 @@ public void handleBarrageError(Throwable t) {
         enqueueError(t);
     }
 
+    private synchronized void tryToDeliverErrorToCallback(final Throwable err) {
```
rcaudy, line 262: I think we want to synchronize `updateServerViewport`, too. Or at least the part that deals with the callback.

java-client/barrage-examples/src/main/java/io/deephaven/client/examples/SnapshotExampleBase.java
rcaudy, line 1: Multiple failures due to the AutoCloseable change.
Maybe we want to bring back `close()` as `cancel()`?
Or maybe we want to change usages to use a `LivenessScope`, which seems better and more idomatic?

rcaudy (dismissed):

rcaudy:
Approved, but needs thorough testing before we merge.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshot.java
```diff
@@ -38,6 +38,13 @@ BarrageSnapshot snapshot(TableSpec tableSpec, BarrageSnapshotOptions options)
         BarrageSnapshot snapshot(TableHandle tableHandle, BarrageSnapshotOptions options);
     }
 
+    /**
+     * This call will return false until all rows for the result table are available or the snapshot failed.
+     *
+     * @return true when all rows for the result table are available or the snapshot failed, false otherwise
+     */
+    boolean isCompleted();
```
rcaudy, line 46: Maybe we should return an enum here? Success or failure? Or maybe we should just be returning a `Future` from `partialTable` and `entireTable` anyway. That would avoid letting the user get a reference before it's safe to do so.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
```diff
@@ -68,7 +68,9 @@ public class BarrageSnapshotImpl extends ReferenceCountedLivenessNode implements
     private volatile int connected = 1;
     private static final AtomicIntegerFieldUpdater<BarrageSnapshotImpl> CONNECTED_UPDATER =
             AtomicIntegerFieldUpdater.newUpdater(BarrageSnapshotImpl.class, "connected");
-
+    private volatile int completed = 0;
```
rcaudy, line 71: No need to bother changing this, since you're planning to converge these impls in the next issue, but it would be better if the parallel variables between snapshot and sub were declared in the same order, etc.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
```diff
@@ -256,77 +320,63 @@ public BarrageTable blockUntilComplete() throws InterruptedException {
...
 
-    @Override
-    public BarrageTable snapshotPartialTable(RowSet viewport, BitSet columns, boolean reverseViewport,
-            boolean blockUntilComplete) throws InterruptedException {
-        isSnapshot = true;
-        return partialTable(viewport, columns, reverseViewport, blockUntilComplete);
+        // if we are building a snapshot via a growing viewport subscription, then cancel our subscription
+        if (isSnapshot) {
+            tryRecordDisconnect();
+            GrpcUtil.safelyCancel(observer, "Barrage snapshot is complete", null);
```
rcaudy, line 349:
```suggestion
            if (tryRecordDisconnect()) {
                GrpcUtil.safelyCancel(observer, "Barrage snapshot is complete", null);
            }
```

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshot.java
```diff
@@ -38,28 +38,12 @@ BarrageSnapshot snapshot(TableSpec tableSpec, BarrageSnapshotOptions options)
...
-
-    /**
-     * Request a full snapshot of the data and populate a {@link Table} with the data that is received.
-     *
-     * @param blockUntilComplete Whether to block execution until all rows for the subscribed table are available
-     *
-     * @return the {@code Table}
-     */
-    Table entireTable(boolean blockUntilComplete) throws InterruptedException;
+    Future<Table> entireTable() throws InterruptedException;
```
rcaudy, line 46: We don't throw an `InterruptedException` anymore. We have a `Future` for that. Applies to all these interfaces.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshot.java
```diff
@@ -38,28 +38,12 @@ BarrageSnapshot snapshot(TableSpec tableSpec, BarrageSnapshotOptions options)
...
-     * This call will return false until all rows for the result table are available or the snapshot failed.
-     *
-     * @return true when all rows for the result table are available or the snapshot failed, false otherwise
-     */
-    boolean isCompleted();
-
     /**
      * Request a full snapshot of the data and populate a {@link Table} with the data that is received.
      *
      * @return the {@code Table}
```
rcaudy, line 44: This is not the right return documentation. Applies to all, again/

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
```diff
@@ -328,34 +236,19 @@ private boolean tryRecordDisconnect() {
...
     }
 
-    private void trySignalCompletion() {
-        trySignalCompletion(null);
-    }
-
-    private void trySignalCompletion(@Nullable final Throwable t) {
-        if (!COMPLETED_UPDATER.compareAndSet(BarrageSubscriptionImpl.this, 0, 1)) {
-            return;
+    private boolean onFutureCancel(boolean cancelResult) {
```
rcaudy, line 239: It seems to me like it would be better to get rid of the argument and only call this if the argument would have been `true`. That way it matches `onFutureComplete()`, and we only have one pattern.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
```diff
@@ -127,66 +141,74 @@ public void onNext(final BarrageMessage barrageMessage) {
...
+                throw new UncheckedDeephavenException("Barrage subscription objects cannot be reused");
             }
             subscribed = true;
         }
 
+        if (isSnapshot) {
+            future = new CompletableFutureAdapter();
+        } else {
+            future = new UpdateGraphAwareFutureAdapter(resultTable.getUpdateGraph());
+        }
```
rcaudy, line 210: I'm concerned we have lurking NPEs since we delay this initialization. :-(

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
```diff
@@ -318,14 +232,15 @@ protected void destroy() {
...
+    private void cancel() {
         if (!tryRecordDisconnect()) {
             return;
         }
 
-        resultTable.handleBarrageError(new RequestCancelledException("Barrage subscription cancelled by client"));
-        GrpcUtil.safelyCancel(observer, "Barrage snapshot is cancelled", exceptionWhileCompleting);
+        final RequestCancelledException requestCancelled =
+                new RequestCancelledException("Barrage subscription cancelled by client");
+        resultTable.handleBarrageError(requestCancelled);
```
rcaudy, line 242: We don't need this handleBarrageError on a future cancel, right? Nothing can get the Table in that case. And we don't really need it on a destroy, because the table had better also be dead at that point; we just need to complete the future exceptionally, if it's not already complete.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
```diff
@@ -78,7 +87,7 @@ public BarrageSnapshotImpl(
         final BarrageUtil.ConvertedArrowSchema schema = BarrageUtil.convertArrowSchema(tableHandle.response());
         final TableDefinition tableDefinition = schema.tableDef;
         resultTable = BarrageTable.make(executorService, tableDefinition, schema.attributes, new CheckForCompletion());
-        resultTable.addParentReference(this);
+        future = new CompletableFuture<>();
```
rcaudy, line 90: Don't you need the `onFutureCancel` callback?

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
```diff
@@ -203,130 +225,51 @@ public BarrageTable partialTable(RowSet viewport, BitSet columns, boolean revers
...
-        cleanup();
-    }
 
-    @Override
-    public synchronized void close() {
-        if (!connected) {
-            return;
-        }
-        GrpcUtil.safelyComplete(observer);
+        resultTable.handleBarrageError(new RequestCancelledException("Barrage subscription cancelled by client"));
```
rcaudy, line 266: We don't need this handleBarrageError on a future cancel, right? Nothing can get the Table in that case. And we don't really need it on a destroy, because the table had better also be dead at that point; we just need to complete the future exceptionally, if it's not already complete.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
```diff
@@ -289,26 +218,11 @@ private boolean tryRecordDisconnect() {
...
-        trySignalCompletion(null);
-    }
-
-    private void trySignalCompletion(@Nullable final Throwable t) {
-        if (!COMPLETED_UPDATER.compareAndSet(BarrageSnapshotImpl.this, 0, 1)) {
-            return;
-        }
-        if (t != null) {
-            exceptionWhileCompleting = t;
+    private void tryCompleteFuture(@Nullable final Throwable t) {
```
rcaudy, line 221: Unused. Delete me.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
nbauernfeind: The future is read in the vp-change callback which cannot be invoked until after the `observer.onNext` (which happens below). This looks safe to me. Note that the future is volatile due to the var's need to cross thread boundaries.

We cannot create this much earlier as we do not know whether the user wants a snapshot or a subscription until right here (after checking that the object is not-yet-used and still connected).

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
nbauernfeind: In the snapshot case, you are right this table is not published anywhere and should be inaccessible. We do still want to do this in the subscription impl though, right? If the rpc has queued up some messages but the UG has not run the source yet, then we are able to drop those messages on the floor and do not need to rely on GC.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
nbauernfeind: I think this one we want if `!isSnapshot`. Maybe we want to `resultTable.forceReferenceCountToZero()` instead?

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshot.java
```diff
@@ -41,9 +41,11 @@ BarrageSnapshot snapshot(TableSpec tableSpec, BarrageSnapshotOptions options)
     /**
      * Request a full snapshot of the data and populate a {@link Table} with the data that is received.
      *
-     * @return the {@code Table}
+     * @return a {@link Future
+     *         <Table>
+     *         } that will be populated with the snapshot
```
rcaudy, line 46: * @return a {@link Future} that will be populated with the result {@link Table}

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscription.java
```diff
@@ -40,68 +40,78 @@ BarrageSubscription subscribe(TableSpec tableSpec, BarrageSubscriptionOptions op
 
     /**
      * Request a full subscription of the data and populate a {@link Table} with the incrementally updating data that is
-     * received. This call will block until all rows for the subscribed table are available.
```
rcaudy: You could add that the *future* will block...

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
```diff
@@ -263,8 +254,13 @@ private void cancel() {
             return;
         }
 
-        resultTable.handleBarrageError(new RequestCancelledException("Barrage subscription cancelled by client"));
-        GrpcUtil.safelyCancel(observer, "Barrage subscription is cancelled", null);
+        final RequestCancelledException cancelledException =
+                new RequestCancelledException("Barrage subscription cancelled by client");
+        if (!isSnapshot) {
+            // Stop our result table from processing any more data.
+            resultTable.forceReferenceCountToZero();
```
rcaudy, line 261: We said that `BarrageTable.destroy` should call `cleanup()`, and that the assertion from cleanup should move to `realRefresh`.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
```diff
@@ -263,8 +254,13 @@ private void cancel() {
             return;
         }
 
-        resultTable.handleBarrageError(new RequestCancelledException("Barrage subscription cancelled by client"));
-        GrpcUtil.safelyCancel(observer, "Barrage subscription is cancelled", null);
+        final RequestCancelledException cancelledException =
```
rcaudy, line 257: We're missing a future.completeExceptionally in the destroy case. But the future might be null. How dystopian.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
```diff
@@ -263,8 +254,13 @@ private void cancel() {
...
+                new RequestCancelledException("Barrage subscription cancelled by client");
```
rcaudy, line 258: Maybe we want to have a string argument to distinguish between "no longer live" and "cancelled by client" ?

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSnapshotImpl.java
```diff
@@ -237,10 +229,8 @@ private void cancel() {
             return;
         }
 
-        final RequestCancelledException requestCancelled =
-                new RequestCancelledException("Barrage subscription cancelled by client");
-        resultTable.handleBarrageError(requestCancelled);
-        GrpcUtil.safelyCancel(observer, "Barrage snapshot is cancelled", requestCancelled);
+        GrpcUtil.safelyCancel(observer, "Barrage snapshot is cancelled",
```
rcaudy, line 232: Same thing about a string arg.

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphAwareCompletableFuture.java
```diff
@@ -0,0 +1,164 @@
...
+        return getInternal(timeout, unit);
+    }
+
+    private T getInternal(long timeout, @Nullable TimeUnit unit)
+            throws InterruptedException, ExecutionException, TimeoutException {
+        // test lock conditions
+        if (updateGraph.sharedLock().isHeldByCurrentThread()) {
+            throw new UnsupportedOperationException(
+                    "Cannot Future#get while holding the " + updateGraph + " shared lock");
+        }
```
rcaudy, line 88: Put this in a method and call it first thing in both `get` overloads. We want to be consistent in our error delivery.
```suggestion
        if (updateGraph.sharedLock().isHeldByCurrentThread()) {
            throw new UnsupportedOperationException(
                    "Cannot Future.get(...) while holding the " + updateGraph + " shared lock");
        }
```

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphAwareCompletableFuture.java
```diff
@@ -0,0 +1,164 @@
...
+        if (resultSupplier != null) {
+            return resultSupplier.get();
+        }
+        if (timeout <= 0) {
+            throw new TimeoutException();
+        }
+        return getInternal(timeout, unit);
+    }
+
+    private T getInternal(long timeout, @Nullable TimeUnit unit)
```
rcaudy, line 82:
```suggestion
    private T getInternal(final long timeout, @Nullable final TimeUnit unit)
```

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphAwareCompletableFuture.java
```diff
@@ -0,0 +1,164 @@
...
+    public boolean completeExceptionally(Throwable ex) {
+        if (ex == null)
+            throw new NullPointerException();
+        return trySignalCompletion(() -> {
+            throw new ExecutionException(ex);
+        });
+    }
+
+    private boolean trySignalCompletion(@NotNull final ThrowingSupplier<T, ExecutionException> result) {
+        if (!RESULT_UPDATER.compareAndSet(UpdateGraphAwareCompletableFuture.this, null, result)) {
```
rcaudy, line 151:
```suggestion
        if (!RESULT_UPDATER.compareAndSet(this, null, result)) {
```

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphAwareCompletableFuture.java
```diff
@@ -0,0 +1,164 @@
...
+                    "Cannot Future#get while holding the " + updateGraph + " shared lock");
+        }
+
+        final boolean holdingUpdateGraphLock = updateGraph.exclusiveLock().isHeldByCurrentThread();
+        if (updateGraphCondition == null && holdingUpdateGraphLock) {
+            try (final SafeCloseable ignored = lock.lockCloseable()) {
+                if (updateGraphCondition == null) {
+                    updateGraphCondition = updateGraph.exclusiveLock().newCondition();
+                }
+            }
```
rcaudy, line 96:
```suggestion
        if (holdingUpdateGraphLock) {
            if (updateGraphCondition == null) {
                updateGraphCondition = updateGraph.exclusiveLock().newCondition();
            }
```

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphAwareCompletableFuture.java
```diff
@@ -0,0 +1,164 @@
...
+        if (!RESULT_UPDATER.compareAndSet(UpdateGraphAwareCompletableFuture.this, null, result)) {
+            return false;
+        }
+
+        final Condition localCondition = updateGraphCondition;
+        try (final SafeCloseable ignored = lock.lockCloseable()) {
+            if (localCondition != null) {
+                updateGraph.requestSignal(localCondition);
+            }
+        }
```
rcaudy, line 160:
```suggestion
        final Condition localUpdateGraphCondition = updateGraphCondition;
        if (localUpdateGraphCondition != null) {
            updateGraph.requestSignal(localUpdateGraphCondition);
        }
        final Condition localLockCondition = lockCondition;
        if (localLockCondition != null) {
            lock.doLocked(localLockCondition::signalAll);
        }
```

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/UpdateGraphAwareCompletableFuture.java
```diff
@@ -140,23 +138,26 @@ public boolean complete(T value) {
     }
 
     public boolean completeExceptionally(Throwable ex) {
-        if (ex == null)
+        if (ex == null) {
```
rcaudy, line 141: You can use `Objects.requireNonNull` for this. No need to make the change if there's nothing else to change.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
```diff
@@ -246,21 +249,23 @@ private void onFutureComplete() {
     @Override
     protected void destroy() {
         super.destroy();
-        cancel();
+        cancel("no longer live");
+        if (future != null) {
```
rcaudy, line 253: I'd use a local here, but it's OK; we never re-assign `future`.

rcaudy (approved):
Two cosmetic complaints. No need to make further changes if tests pass.

rcaudy (approved):

PR #4680, opened 2023-10-20
Server-generated column statistics

niloc132:
Reintroduces the column statistics feature from DHE. This code is copied, with a few changes to how it behaves and how it functions:
 * Results are sent to the client in a static table rather than a simple object payload.
 * Client can control the max number of unique values to display, up to a point (previously the max was technically 19, new default is 20).
 * Non-Comparable objects now have their "popularity" counts as well as Comparable, until a max limit is reached. Only the most common 20 entries will be sent to the client.
 * Deephaven "null" values are used to represent that a standard deviation or average cannot be computed.

Note that until #188 is solved, the JS API and web UI will not display the unique value list.


replication/static/src/main/java/io/deephaven/replicators/ReplicateColumnStats.java
```diff
@@ -0,0 +1,46 @@
...
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.util.List;
+
+import static io.deephaven.replication.ReplicatePrimitiveCode.shortToAllIntegralTypes;
+import static io.deephaven.replication.ReplicationUtils.globalReplacements;
+
+public class ReplicateColumnStats {
```
devinrsmith, line 15: Is this hooked up to gradle replication tasks?

devinrsmith (commented):
Proto changes look structurally sound; but I have a larger question. Are we fundamentally performing computations here that _can't_ be performed with our existing APIs / table operations? If the answer is "yes", would it be better to try and improve our fundamental operations to handle this use case? ColumnStatisticsRequest looks like a heavy hammer, and I'm sad to see the bespoke logic in `server/src/main/java/io/deephaven/server/table/stats/`. (I understand this is mostly a copy of DHE code, but worth raising Qs IMO.)

py/client/pydeephaven/proto/table_pb2_grpc.py
```diff
@@ -602,6 +607,14 @@ def MetaTable(self, request, context):
         context.set_details('Method not implemented!')
         raise NotImplementedError('Method not implemented!')
 
+    def ComputeColumnStatistics(self, request, context):
+        """*
+        Returns a new table representing statistics about a single column of the provided table.
+        """
+        context.set_code(grpc.StatusCode.UNIMPLEMENTED)
+        context.set_details('Method not implemented!')
+        raise NotImplementedError('Method not implemented!')
```
chipkent, line 616: Does this need implementation?

niloc132: Yes - every generated implementation looks like this. If you were to implement a python server, you would override this method and offer a real implementation.

replication/static/src/main/java/io/deephaven/replicators/ReplicateColumnStats.java
niloc132: No, thanks for the reminder, I'll update gradle.

niloc132:
@devinrsmith 

> Are we fundamentally performing computations here that can't be performed with our existing APIs / table operations?

Yes, and deliberately so - we do not have a "Count" operation that only counts non-nulls, and we do not have a "unique" operation that only selects the N most unique. Neither is impossible to add, but @rcaudy specifically prefers that for now we just stick to a specialized implementation like DHE has.

As one alternative, we could implement the two above aggregations, which would give us these same operations, except now ticking - however, the only consumer of this API is the web UI's column stats popup, which doesn't expect a Table instance, and so doesn't support ticking data. So the only win here would be that we would implement two new aggregation types, and then not need the rest of this new code.

server/src/main/java/io/deephaven/server/table/stats/BigDecimalChunkedNumericalStats.java
```diff
@@ -0,0 +1,106 @@
...
+                TableTools.longCol("SIZE", rowSet.size()),
+                new ColumnHolder<>("SUM", BigDecimal.class, null, false, sum),
+                new ColumnHolder<>("SUM_ABS", BigDecimal.class, null, false, absSum),
+                new ColumnHolder<>("SUM_SQRD", BigDecimal.class, null, false, sqrdSum),
+                new ColumnHolder<>("MIN", BigDecimal.class, null, false, min),
+                new ColumnHolder<>("MAX", BigDecimal.class, null, false, max),
+                new ColumnHolder<>("MIN_ABS", BigDecimal.class, null, false, absMin),
+                new ColumnHolder<>("MAX_ABS", BigDecimal.class, null, false, absMax),
+                new ColumnHolder<>("AVG", BigDecimal.class, null, false, avg),
+                new ColumnHolder<>("AVG_SUM", BigDecimal.class, null, false, absAvg),
```
rcaudy, line 103:
```suggestion
                new ColumnHolder<>("AVG_ABS", BigDecimal.class, null, false, absAvg),
```

server/src/main/java/io/deephaven/server/table/stats/BigDecimalChunkedNumericalStats.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private BigDecimal min = null;
+    private BigDecimal max = null;
+    private BigDecimal absMin = null;
+    private BigDecimal absMax = null;
+
+    @Override
+    public Table processChunks(final RowSet rowSet, final ColumnSource<?> columnSource, boolean usePrev) {
+
+        try (final ChunkSource.GetContext getContext = columnSource.makeGetContext(CHUNK_SIZE)) {
```
rcaudy, line 36: All of these could be written with less boilerplate using column iterators.

server/src/main/java/io/deephaven/server/table/stats/ObjectChunkedStats.java
```diff
@@ -0,0 +1,117 @@
...
+                        uniqueValues.add(val);
+                    } else if (uniqueCount > maxUniqueToCollect) {
+                        // we no longer need to track counts for these items; fall back to a Set to get at least a count
+                        uniqueValues.addAll(countValues.keySet());
+                        countValues.clear();
+                        uniqueValues.add(val);
+                        useSet = true;
+                    } else if (countValues.adjustOrPutValue(val, 1, 1) == 1) {
+                        uniqueCount++;
+                    }
```
rcaudy, line 77: Seems to me like we do this pessimistically, before knowing if the new value increases our unique count. We can probably simplify:
```suggestion
                    } else if (countValues.adjustOrPutValue(val, 1, 1) == 1 && ++uniqueCount > maxUniqueToCollect) {
                        // we no longer need to track counts for these items; fall back to a Set to just get a count
                        uniqueValues.addAll(countValues.keySet());
                        countValues.clear();
                        useSet = true;
                    }
```

server/src/main/java/io/deephaven/server/table/stats/ObjectChunkedStats.java
```diff
@@ -0,0 +1,117 @@
...
+            final int numUnique;
+            String[] uniqueKeys;
+            long[] uniqueCounts;
+            if (useSet) {
+                numUnique = uniqueValues.size();
+                uniqueKeys = CollectionUtil.ZERO_LENGTH_STRING_ARRAY;
+                uniqueCounts = CollectionUtil.ZERO_LENGTH_LONG_ARRAY;
+            } else {
+                numUnique = countValues.size();
+                List<Map.Entry<String, Long>> sorted = new ArrayList<>();
```
rcaudy, line 90:
```suggestion
                List<Map.Entry<String, Long>> sorted = new ArrayList<>(numUnique);
```

server/src/main/java/io/deephaven/server/table/stats/ObjectChunkedStats.java
```diff
@@ -0,0 +1,117 @@
...
+                    sorted.add(Map.entry(Objects.toString(o), c));
+                    return true;
+                });
+                sorted.sort(Map.Entry.comparingByValue());
+
+                int resultCount = Math.min(maxUniqueToDisplay, sorted.size());
+                uniqueKeys = new String[resultCount];
+                uniqueCounts = new long[resultCount];
+                Iterator<Map.Entry<String, Long>> iter = sorted.iterator();
+                for (int i = 0; i < resultCount && iter.hasNext(); i++) {
```
rcaudy, line 102: If you're short-circuiting on `!iter.hasNext()`, something is wrong...

server/src/main/java/io/deephaven/server/table/stats/ObjectChunkedStats.java
```diff
@@ -0,0 +1,117 @@
...
+                    uniqueCounts[i] = entry.getValue();
+                }
+            }
+
+            return TableTools.newTable(
+                    TableTools.longCol("COUNT", count),
+                    TableTools.longCol("SIZE", rowSet.size()),
+                    TableTools.intCol("UNIQUE_VALUES", numUnique),
+                    new ColumnHolder<>("UNIQUE_KEYS", String[].class, String.class, false, uniqueKeys),
+                    new ColumnHolder<>("UNIQUE_COUNTS", long[].class, long.class, false, uniqueCounts));
```
rcaudy, line 114: Do we really need to send these when we overflowed?

server/src/main/java/io/deephaven/server/table/stats/CharacterChunkedStats.java
```diff
@@ -0,0 +1,112 @@
...
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+public class CharacterChunkedStats implements ChunkedStatsKernel {
```
rcaudy, line 24: Be sure to port changes from `ObjectChunkedStats`.

server/src/main/java/io/deephaven/server/table/stats/CharacterChunkedStats.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @Override
+    public Table processChunks(final RowSet rowSet, final ColumnSource<?> columnSource, boolean usePrev) {
+        long count = 0;
+        int uniqueCount = 0;
+
+        final TObjectLongHashMap<Object> countValues = new TObjectLongHashMap<>();
+        boolean useSet = false;
+        final Set<Object> uniqueValues = new HashSet<>();
```
rcaudy, line 40:
```suggestion
        final TCharLongHashMap countValues = new TCharLongHashMap();
        boolean useSet = false;
        final TCharSet uniqueValues = new TCharHashSet();
```

server/src/main/java/io/deephaven/server/table/stats/CharacterChunkedStats.java
```diff
@@ -0,0 +1,112 @@
...
+                        uniqueValues.add(val);
+                    } else if (uniqueCount > maxUniqueToCollect) {
+                        // we no longer need to track counts for these items; fall back to a Set to get at least a count
+                        uniqueValues.addAll(countValues.keySet());
+                        countValues.clear();
+                        uniqueValues.add(val);
+                        useSet = true;
+                    } else if (countValues.adjustOrPutValue(val, 1, 1) == 1) {
+                        uniqueCount++;
+                    }
```
rcaudy, line 72: Explanation on object version.
```suggestion
                    } else if (countValues.adjustOrPutValue(val, 1, 1) == 1 && ++uniqueCount > maxUniqueToCollect) {
                        // we no longer need to track counts for these items; fall back to a Set to just get a count
                        uniqueValues.addAll(countValues.keySet());
                        countValues.clear();
                        useSet = true;
                    }
```

server/src/main/java/io/deephaven/server/table/stats/CharacterChunkedStats.java
```diff
@@ -0,0 +1,112 @@
...
+            final int numUnique;
+            String[] uniqueKeys;
+            long[] uniqueCounts;
+            if (useSet) {
+                numUnique = uniqueValues.size();
+                uniqueKeys = CollectionUtil.ZERO_LENGTH_STRING_ARRAY;
+                uniqueCounts = CollectionUtil.ZERO_LENGTH_LONG_ARRAY;
+            } else {
+                numUnique = countValues.size();
+                List<Map.Entry<String, Long>> sorted = new ArrayList<>();
```
rcaudy, line 85:
```suggestion
                List<Map.Entry<String, Long>> sorted = new ArrayList<>(numUnique);
```

server/src/main/java/io/deephaven/server/table/stats/CharacterChunkedStats.java
```diff
@@ -0,0 +1,112 @@
...
+                    uniqueKeys[i] = entry.getKey();
+                    uniqueCounts[i] = entry.getValue();
+                }
+            }
+
+            return TableTools.newTable(
+                    TableTools.longCol("COUNT", count),
+                    TableTools.longCol("SIZE", rowSet.size()),
+                    TableTools.intCol("UNIQUE_VALUES", numUnique),
+                    new ColumnHolder<>("UNIQUE_KEYS", String[].class, String.class, false, uniqueKeys),
```
rcaudy, line 108: Same question... can we skip these when empty? What even *happens* if we create these mismatched column sizes?

server/src/main/java/io/deephaven/server/table/stats/DateTimeChunkedStats.java
```diff
@@ -0,0 +1,61 @@
...
+import io.deephaven.engine.rowset.RowSequence;
+import io.deephaven.engine.rowset.RowSet;
+import io.deephaven.engine.table.ChunkSource;
+import io.deephaven.engine.table.ColumnSource;
+import io.deephaven.engine.table.Table;
+import io.deephaven.engine.util.TableTools;
+import io.deephaven.time.DateTimeUtils;
+import io.deephaven.util.QueryConstants;
+
+public class DateTimeChunkedStats implements ChunkedStatsKernel {
```
rcaudy, line 14: Seems to me like we could have just done these as objects.

rcaudy (commented):
This is a review of everything but the JS* classes.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -254,6 +254,10 @@ service TableService {
    */
   rpc MetaTable(MetaTableRequest) returns (ExportedTableCreationResponse) {}
 
+  /**
+   * Returns a new table representing statistics about a single column of the provided table.
+   */
+  rpc ComputeColumnStatistics(ColumnStatisticsRequest) returns (ExportedTableCreationResponse) {}
```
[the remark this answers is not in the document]
niloc132: It takes a table as input and returns a table, so it seemed easy to put here, but I'm happy to make its own service too, if you'd prefer that.

server/src/main/java/io/deephaven/server/table/ops/ColumnStatisticsGrpcImpl.java
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.commons.lang3.mutable.MutableObject;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import java.time.Instant;
+import java.time.ZonedDateTime;
+import java.util.List;
+
+@Singleton
+public class ColumnStatisticsGrpcImpl extends GrpcTableOperation<ColumnStatisticsRequest> {
```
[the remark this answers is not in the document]
niloc132: I think we could only validate that the column name is actually a valid name, and that the unique count isn't negative? I had thought this wasn't worth it, as those are plainly wrong (and will just fail slightly slower).

server/src/main/java/io/deephaven/server/table/ops/ColumnStatisticsGrpcImpl.java
```diff
@@ -0,0 +1,105 @@
...
+    public ColumnStatisticsGrpcImpl(final TableServiceContextualAuthWiring authWiring) {
+        super(authWiring::checkPermissionComputeColumnStatistics, BatchTableRequest.Operation::getColumnStatistics,
+                ColumnStatisticsRequest::getResultId, ColumnStatisticsRequest::getSourceId);
+    }
+
+    @Override
+    public Table create(ColumnStatisticsRequest request, List<SessionState.ExportObject<Table>> sourceTables) {
+        Table table = sourceTables.get(0).get();
+        String columnName = request.getColumnName();
+        final Class<?> type = table.getDefinition().getColumn(columnName).getDataType();
```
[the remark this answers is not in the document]
niloc132: Can do that - though cannot do it in `validateRequest`, only in `create` (since we need the `Table` to see if the column exists).

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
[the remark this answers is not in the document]
devinrsmith: I prefer it here (as-is).

server/src/main/java/io/deephaven/server/table/stats/ChunkedNumericalStatsKernel.java
```diff
@@ -0,0 +1,51 @@
...
+import io.deephaven.engine.table.Table;
+import io.deephaven.util.QueryConstants;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+
+public interface ChunkedNumericalStatsKernel extends ChunkedStatsKernel {
+
+    static ChunkedNumericalStatsKernel makeChunkedNumericalStatsFactory(final Class<?> type) {
+        if (type == Long.class || type == long.class) {
```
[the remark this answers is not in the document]
niloc132: This just happens to be the order that DHE had.

server/src/main/java/io/deephaven/server/table/stats/ObjectChunkedStats.java
niloc132: That's true - this was a late addition and I thought I was adding extra safety, but you're right that it doesn't make sense to exist.

server/src/main/java/io/deephaven/server/table/stats/DateTimeChunkedStats.java
niloc132: That could make sense, a few notes:
 * Collecting unique datetimes is _probably_ not constructive for many use cases
 * We would either lose min/max for datetime columns, or we could add min/max to character/object? It might be weird to have a min/max character, but maybe we don't care
 * We would need to stop reinterpreting to long (which I think means we would recreate each object on the way in), which I think is more expensive to compute?

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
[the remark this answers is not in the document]
rcaudy: My only complaint is that it's very specific. The schemas are variable and undocumented, and nobody but our web UI knows what to expect.

server/src/main/java/io/deephaven/server/table/ops/ColumnStatisticsGrpcImpl.java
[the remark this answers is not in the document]
rcaudy: True enough.

server/src/main/java/io/deephaven/server/table/ops/ColumnStatisticsGrpcImpl.java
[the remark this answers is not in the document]
rcaudy: I do think it's worth validating in create. Better to return a proper error message.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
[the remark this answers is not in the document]
devinrsmith: I would prefer we solve that w/ more documentation at this site (mentioning only really meant for use by web UI).

server/src/main/java/io/deephaven/server/table/stats/DateTimeChunkedStats.java
rcaudy: Yeah, I suppose you're right. Maybe we need comparables, and then we can have min-max *and* a sampling/count.

rcaudy:
> @devinrsmith
> 
> > Are we fundamentally performing computations here that can't be performed with our existing APIs / table operations?
> 
> Yes, and deliberately so - we do not have a "Count" operation that only counts non-nulls, and we do not have a "unique" operation that only selects the N most unique. Neither is impossible to add, but @rcaudy specifically prefers that for now we just stick to a specialized implementation like DHE has.
> 
> As one alternative, we could implement the two above aggregations, which would give us these same operations, except now ticking - however, the only consumer of this API is the web UI's column stats popup, which doesn't expect a Table instance, and so doesn't support ticking data. So the only win here would be that we would implement two new aggregation types, and then not need the rest of this new code.

We discussed this at some length before doing the porting. I would prefer to use our aggs, and thus standardize on one set of code for this. What convinced me that we should just port is:
1. The client cannot consume ticking stats (if it could, I would insist we use aggs)
2. This code can be "tighter", only scanning each chunk a single time
3. We can ship faster this way

server/src/main/java/io/deephaven/server/table/stats/CharacterChunkedStats.java
niloc132: Skipping when we don't populate makes more sense, the client can handle that case.

What happens if we create this mismatched intentionally? The client will have an error when it tries to use them... but at least with the current code, how is that possible?

Is this like asking what would happen if you grouped two columns on a table and they ended up with different sizes...?

rcaudy: Scratch the size-mismatch question, that was reader-error.

rcaudy (commented):
.

server/src/main/java/io/deephaven/server/table/stats/ObjectChunkedStats.java
niloc132: what do you think about this instead, to avoid adding to one collection, copying to the other, clearing the first:

```
            while (iterator.hasNext()) {
                char val = iterator.next();
                if (val == QueryConstants.NULL_CHAR) {
                    continue;
                }
                count++;

                if (countValues.adjustOrPutValue(val, 1, 1) == 1 && ++uniqueCount > maxUniqueToCollect) {
                    // we no longer want to track counts for these items; fall back to a Set to get at least a count
                    uniqueValues.addAll(countValues.keySet());
                    countValues.clear();
                    useSet = true;
                    break;
                }
            }
            while (iterator.hasNext()) {
                char val = iterator.next();
                if (val == QueryConstants.NULL_CHAR) {
                    continue;
                }
                count++;

                uniqueValues.add(val);
            }
```

server/src/main/java/io/deephaven/server/table/stats/BigDecimalChunkedNumericalStats.java
niloc132: I didn't change the short/int/long or float/double, since they collect stats per-chunk to avoid greater loss of precision.

server/src/main/java/io/deephaven/server/table/stats/ObjectChunkedStats.java
rcaudy: This seems perfectly nice, but I don't understand what we're avoiding. You *seem* to have the same logic as my suggestion, just splitting into two iterations to avoid the `if` inside the iteration.

server/src/main/java/io/deephaven/server/table/stats/BigDecimalChunkedNumericalStats.java
rcaudy: Seems like a fair point.

server/src/main/java/io/deephaven/server/table/stats/CharacterChunkedStats.java
```diff
@@ -35,76 +32,67 @@ public Table processChunks(final RowSet rowSet, final ColumnSource<?> columnSour
...
-                        useSet = true;
-                    } else if (countValues.adjustOrPutValue(val, 1, 1) == 1) {
-                        uniqueCount++;
-                    }
+        final TCharSet uniqueValues = new TCharHashSet();
+
+        try (CharacterColumnIterator iterator =
+                new ChunkedCharacterColumnIterator(usePrev ? columnSource.getPrevSource() : columnSource, rowSet)) {
+            while (iterator.hasNext()) {
+                char val = iterator.next();
```
rcaudy, line 42: This is important to change, or you will NPE when `NULL_CHAR` is converted to `null` and then unboxed. I wonder if this means the null case has not been tested?
```suggestion
                char val = iterator.nextChar();
```

server/src/main/java/io/deephaven/server/table/stats/CharacterChunkedStats.java
```diff
@@ -35,76 +32,67 @@ public Table processChunks(final RowSet rowSet, final ColumnSource<?> columnSour
...
+                    // we no longer want to track counts for these items; fall back to a Set to get at least a count
+                    uniqueValues.addAll(countValues.keySet());
+                    countValues.clear();
+                    useSet = true;
+                    break;
                 }
             }
+            while (iterator.hasNext()) {
+                // items still remain, count non-nulls and uniques
+                char val = iterator.next();
```
rcaudy, line 58:
```suggestion
                char val = iterator.nextChar();
```

server/src/main/java/io/deephaven/server/table/stats/CharacterChunkedStats.java
```diff
@@ -35,76 +32,67 @@ public Table processChunks(final RowSet rowSet, final ColumnSource<?> columnSour
...
-                    Map.Entry<String, Long> entry = iter.next();
-                    uniqueKeys[i] = entry.getKey();
-                    uniqueCounts[i] = entry.getValue();
-                }
+        if (useSet) {
+            return TableTools.newTable(
+                    TableTools.longCol("COUNT", count),
+                    TableTools.longCol("SIZE", rowSet.size()),
+                    TableTools.intCol("UNIQUE_VALUES", uniqueValues.size()));
+        } else {
```
rcaudy, line 74: Could unindent this block, since you returned in the if.

replication/static/src/main/java/io/deephaven/replicators/ReplicateColumnStats.java
```diff
@@ -37,10 +40,22 @@ private static void fixupIntegerChunkName(final String intPath) throws IOExcepti
...
+        lines = globalReplacements(lines,
                 "QueryConstants.NULL_OBJECT", "null",
                 "\\? extends Attributes.Values", "?, ? extends Attributes.Values",
-                "ObjectChunk<[?] ", "ObjectChunk<?, ? "));
+                "ObjectChunk<[?] ", "ObjectChunk<?, ? ",
+                " TObjectLongHashMap", " TObjectLongHashMap<Object>",
+                " TObjectSet", " Set<Object>",
+                " TObjectHashSet", " HashSet<>",
+                "new ChunkedObjectColumnIterator", "new ChunkedObjectColumnIterator<>",
+                "\\(ObjectColumnIterator", "(ObjectColumnIterator<Object>");
```
rcaudy, line 58: You'll need `nextObject()` -> `next()`.

server/src/main/java/io/deephaven/server/table/stats/ObjectChunkedStats.java
```diff
@@ -40,76 +38,67 @@ public Table processChunks(final RowSet rowSet, final ColumnSource<?> columnSour
         long count = 0;
         int uniqueCount = 0;
 
-        final TObjectLongHashMap<Object> countValues = new TObjectLongHashMap<>();
+        final TObjectLongHashMap<Object> countValues = new TObjectLongHashMap<Object>();
```
rcaudy, line 41: You could make this cleaner without unduly stressing the replication if you used the interface type on the LHS.
```suggestion
        final TObjectLongMap<Object> countValues = new TObjectLongHashMap<>();
```

server/src/main/java/io/deephaven/server/table/stats/ObjectChunkedStats.java
```diff
@@ -5,26 +5,24 @@
...
-import io.deephaven.engine.rowset.RowSequence;
 import io.deephaven.engine.rowset.RowSet;
-import io.deephaven.engine.table.ChunkSource;
 import io.deephaven.engine.table.ColumnSource;
 import io.deephaven.engine.table.Table;
 import io.deephaven.engine.table.impl.util.ColumnHolder;
+import io.deephaven.engine.table.iterators.ObjectColumnIterator;
+import io.deephaven.engine.table.iterators.ChunkedObjectColumnIterator;
 import io.deephaven.engine.util.TableTools;
 import io.deephaven.util.QueryConstants;
```
rcaudy, line 19: Can drop this import, if you want.

rcaudy (commented):
One minor bug, and some code cleanup suggestions.

server/src/main/java/io/deephaven/server/table/stats/CharacterChunkedStats.java
niloc132: Good catch from the col iterator change. I've added a simple test for the remaining types to check these.

rcaudy (approved):

PR #4684, opened 2023-10-20
Fix a bug in the use of liveness_scope in tests

### LINKED ISSUE #4681: More frequent nightly checks failure with OOM in Python Kafka tests
The tests right now relies on Java GC to clean up the resources used for consuming/publishing to Kafka streams. 
Things we can try:
1. setting the target table to None and calling garbage_collect
2. another level of liveliness scoping the target table
3. create a safe way to stop consuming from/publishing to Kafka

jmao-denver:
Fixes #4681

### DIFF
--- a/py/server/tests/testbase.py
+++ b/py/server/tests/testbase.py
@@ -37,10 +37,10 @@ class BaseTestCase(unittest.TestCase):
         # Note that this is technically not a supported way to use liveness_scope, but we are deliberately leaving
         # the scope open across separate method calls, which we would normally consider unsafe.
         self.opened_scope = liveness_scope()
+        self.opened_scope.__enter__()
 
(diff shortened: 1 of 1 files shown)


niloc132 (approved):

py/server/tests/testbase.py
```diff
@@ -37,10 +37,10 @@ def setUp(self) -> None:
         # Note that this is technically not a supported way to use liveness_scope, but we are deliberately leaving
         # the scope open across separate method calls, which we would normally consider unsafe.
         self.opened_scope = liveness_scope()
```
rcaudy, line 39: Shouldn't this be LivenessScope()?

rcaudy (approved):

PR #4685, opened 2023-10-20
Add ObjectProcessor noop impl and ObjectProcessor tests

devinrsmith:
Unit test follow-up to #4648

### DIFF
--- a/extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessor.java
+++ b/extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessor.java
@@ -74,7 +74,7 @@ public interface ObjectProcessor<T> {
      * output chunks sizes.
      *
      * @param outputTypes the output types
-     * @param fillWithNullValue if the output chunks should be filled with the appropriate null value;
+     * @param fillWithNullValue if the output chunks should be filled with the appropriate null value
      * @return the no-op object processor
      * @param <T> the object type
      */
--- a/extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessor.java
+++ b/extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessor.java
@@ -67,6 +67,22 @@ public interface ObjectProcessor<T> {
     }
 
     /**
+     * Creates a "no-operation" object processor that ignores the input object chunk. If {@code fillWithNullValue} is
+     * {@code true}, during {@link ObjectProcessor#processAll(ObjectChunk, List) processAll}
+     * {@link WritableChunk#fillWithNullValue(int, int) fillWithNullValue} will be invoked on each output chunk;
+     * otherwise, the output chunk contents will not be modified. In either case, the processing will increment the
+     * output chunks sizes.
+     *
+     * @param outputTypes the output types
+     * @param fillWithNullValue if the output chunks should be filled with the appropriate null value;
+     * @return the no-op object processor
+     * @param <T> the object type
+     */
+    static <T> ObjectProcessor<T> noop(List<Type<?>> outputTypes, boolean fillWithNullValue) {
+        return new ObjectProcessorNoop<>(outputTypes, fillWithNullValue);
+    }
+
+    /**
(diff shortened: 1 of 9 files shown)


extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessor.java
```diff
@@ -66,6 +66,22 @@ static <T> ObjectProcessor<T> rowLimited(ObjectProcessor<T> delegate, int rowLim
...
 
+    /**
+     * Creates a "no-operation" object processor that ignores the input object chunk. If {@code fillWithNullValue} is
+     * {@code true}, during {@link ObjectProcessor#processAll(ObjectChunk, List) processAll}
+     * {@link WritableChunk#fillWithNullValue(int, int) fillWithNullValue} will be invoked on each output chunk;
+     * otherwise, the output chunk contents will not be modified. In either case, the processing will increment the
+     * output chunks sizes.
+     *
+     * @param outputTypes the output types
+     * @param fillWithNullValue if the output chunks should be filled with the appropriate null value;
```
rcaudy, line 77:
```suggestion
     * @param fillWithNullValue if the output chunks should be filled with the appropriate null value
```

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessor.java
```diff
@@ -66,6 +66,22 @@ static <T> ObjectProcessor<T> rowLimited(ObjectProcessor<T> delegate, int rowLim
...
+     * @return the no-op object processor
+     * @param <T> the object type
+     */
+    static <T> ObjectProcessor<T> noop(List<Type<?>> outputTypes, boolean fillWithNullValue) {
```
rcaudy, line 81: Does this belong in production sources? Same question for `ObjectProcessorNoop`, of course.

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessorStrict.java
```diff
@@ -22,22 +22,28 @@ static <T> ObjectProcessor<T> create(ObjectProcessor<T> delegate) {
...
+        this.outputTypes = List.copyOf(delegate.outputTypes());
     }
 
     @Override
     public List<Type<?>> outputTypes() {
-        return delegate.outputTypes();
+        final List<Type<?>> outputTypes = delegate.outputTypes();
+        if (!this.outputTypes.equals(outputTypes)) {
+            throw new UncheckedDeephavenException("Implementation is returning a different list of outputTypes");
+        }
```
rcaudy, line 37: Do you think this is actually a real worry?

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessor.java
devinrsmith: I think yes - it's potentially useful scaffolding to test out a given publisher before you have your own ObjectProcessor impl.

For example, let's say I want to develop my own Capn Proto ObjectProcessor, but in parallel I still want to allow other developers to work on the downstream queries while the impl is still in progress - they can become "unblocked" by faking it w/ noop:

```groovy
my_table = consumeToTable(..., objectProcessorSpec(noop([Type.stringType(), Type.intType()], true), ["Name", "Age"]))

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessorStrict.java
devinrsmith: No, but the strict impl _is_ being a bit pedantic on purpose.

rcaudy (commented):
Placeholder

extensions/kafka/src/main/java/io/deephaven/processor/ObjectProcessor.java
rcaudy: That seems pretty tenuous.

rcaudy (approved):

PR #4688, opened 2023-10-22
Gradle build target and CI checks for python ticking client

### LINKED ISSUE #4642: Add a gradle build and test targets for python ticking client; include in CI checks.
We have a dockerized build and test run that is included in CI checks for the C++ client thanks to work done by Colin and Devin.   The one for the R client is in the making.   The cython based python ticking client (similar to the R client in that it uses wrapping of the C++ client) needs the same.

jcferretti:
Fixes #4642


py/client-ticking/tests/test_ticking_basic.py
```diff
@@ -0,0 +1,28 @@
...
+        half_second_in_nanos = 500 * 1000 * 1000
+        table = session.time_table(period=half_second_in_nanos).update(formulas=["Col1 = i"])
+        table_added_count = 0
+        def update_table_added(added):
+            nonlocal table_added_count
+            table_added_count += len(added)
+        listener_handle = dhtl.listen(table, lambda update : update_table_added(update.added()))
+        listener_handle.start()
+        time.sleep(3)
+        self.assertTrue(4 <= table_added_count and table_added_count <= 10)
```
jmao-denver, line 23: prone to failing, no? 
```suggestion
        time.sleep(3)
        self.assertTrue(4 <= table_added_count  <= 10)
```

jcferretti: I improved the test to make its asserts have more precise checks on what we expect to see, PTAL.

py/client-ticking/demos/demo_simple.py
```diff
@@ -7,7 +7,7 @@
 import pyarrow as pa
 import time
 from typing import Dict, List, Tuple
-from pydeephaven import TableListener, TableUpdate, listen
+from pydeephaven_ticking import TableListener, TableUpdate, listen
```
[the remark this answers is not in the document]
jcferretti: We discussed over slack; it doesn't work for me (the test fails) without that change; maybe I have something wrong in the build script?

py/client-ticking/build.gradle
```diff
@@ -0,0 +1,108 @@
...
+}
+
+evaluationDependsOn(':cpp-client')
+
+configurations {
+    python {}
+}
+
+dependencies {
+    python project(':cpp-client')
```
niloc132, line 14: If this does something, we shouldn't need the `evaluationDependsOn` above. With that said, I don't see that cpp-client actually declares any artifacts, and this project doesn't use `configurations.python` for anything to put those artifacts to use, so I think you can safely delete the `configurations` and `dependencies` blocks.

Using artifacts is "better" in that it means gradle can actually see the files moving back and forth, figure out when changes don't have to happen. Other places where we have a `python` configuration are usually when we have a `.whl` file copied out of the container and ready for another task (typically another docker container) to copy in. I think we could use that approach here if the cpp-client:cppClient tasks set up some `copyOut` of compiled cpp output files... but if we're just depending on the docker image produced in cpp-client, `evaluationDependsOn` and `parentContainers` is probably sufficient.

---


Okay, that being said, the `parentContainers` reference below points at py-client, not cpp-client, and py-client points at cpp-client. This means that technically gradle is allow to build this project after cpp-client but before py-client, which will result in an NPE when that parentContainers is evaluated. To fix that, change the `evaluationDependsOn` to point at py-client instead of cpp-client.

py/client-ticking/demos/demo_simple.py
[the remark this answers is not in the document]
niloc132: yeah the py/client-ticking/src/ directory seems to explicitly have a `pydeephaven_ticking/__init__.py`, which tells me as a naive reader that the appropriate thing to do is to import things from `pydeephaven_ticking`...

py/client-ticking/build.gradle
jcferretti: Fixed, thanks.

py/client-ticking/build.gradle
```diff
@@ -0,0 +1,100 @@
...
+        //
+        // Setup for test run.
+        //
+        environmentVariable 'DH_HOST', deephavenDocker.containerName.get()
+        environmentVariable 'DH_PORT', '10000'
+    }
+    containerDependencies.dependsOn = [deephavenDocker.healthyTask]
+    containerDependencies.finalizedBy = deephavenDocker.endTask
+    network = deephavenDocker.networkName.get()
+    parentContainers = [ project.tasks.getByName('pyClientTicking') ]
```
niloc132, line 92:
```suggestion
    parentContainers = [ buildPyClientTicking ]
```

py/client/build.gradle
```diff
@@ -99,24 +96,39 @@ def testPyClient = Docker.registerDockerTask(project, 'testPyClient') {
...
+        workingDir('/project')
+        runCommand '''set -eux; \
+                      pip3 install --upgrade pip; \
+                      pip3 install -r requirements-dev.txt; \
+                      pip3 install .; \
+                      cd /; \
+                      rm -fr /project
+                   '''
+    }
+    parentContainers = [ project.tasks.getByPath(':cpp-client:cppClient') ]
```
niloc132, line 111: this file now also needs an evaluationDependsOn :cpp-client

On the existing evaluationDependsOn line: I don't think we need to depend on the python project any more.

py/client/build.gradle
```diff
@@ -99,24 +96,39 @@ def testPyClient = Docker.registerDockerTask(project, 'testPyClient') {
...
-                      pip3 install --upgrade pip; \\
-                      pip3 install -r requirements-dev.txt'''
-
-        runCommand 'pip install .'
         environmentVariable 'DH_HOST', deephavenDocker.containerName.get()
         environmentVariable 'DH_PORT', '10000'
     }
-    parentContainers = [ Docker.registryTask(project, 'python') ]
-    entrypoint = ['python', '-m', 'xmlrunner', 'discover', 'tests', '-v', '-o', '/out/report']
+    parentContainers = [ project.tasks.getByName('pyClient') ]
```
niloc132, line 130:
```suggestion
    parentContainers = [ buildPyClientImage ]
```

kosak (dismissed):

py/client-ticking/README.md
```diff
@@ -80,10 +80,11 @@ Run python from the venv while in this directory, and try this sample Python pro
 
 ```
 import pydeephaven as dh
+import pydeephaven_ticking.table_listener as dhtl
```
jmao-denver, line 83: remove?

py/client-ticking/README.md
```diff
@@ -80,10 +80,11 @@ Run python from the venv while in this directory, and try this sample Python pro
...
 import time
 session = dh.Session() # assuming Deephaven Community Edition is running locally with the default configuration
 table = session.time_table(period=1000000000).update(formulas=["Col1 = i"])
-listener_handle = dh.listen(table, lambda update : print(update.added()))
+listener_handle = dhtl.listen(table, lambda update : print(update.added()))
```
jmao-denver, line 87: revert?

py/client/build.gradle
```diff
@@ -81,14 +81,11 @@ deephavenDocker {
     networkName.set "pydeephaven-network-${randomSuffix}"
 }
 
-def testPyClient = Docker.registerDockerTask(project, 'testPyClient') {
+def buildPyClientImage = Docker.registerDockerTask(project, 'pyClient') {
```
niloc132, line 84: The "main/test" split that you're doing shouldn't be necessary in this project - the py-client project already uses our home-grown "python-wheel" plugin, which will make a wheel file on disk ready to be used by the test. 

If you want to run the tests against the wheel, I'd suggest dropping this task here, and just copy the wheel into the test docker image. Otherwise it is appropriate to do the test against the sources instead of the already-installed wheel (but you might have reasons to not do that?).

Very roughly, entirely remove this task, and copy the already-built wheel and requirements-dev.txt, and pip install both. Here's an example of what testPyClient could look like (it runs and seems to pass locally) - I copied the dev dependencies first under the assumption that they change less frequently.

```
def testPyClient = Docker.registerDockerTask(project, 'testPyClient') {
    copyIn {
        from(tasks.named('buildWheel').get().outputs.files.singleFile)
        from ('requirements-dev.txt') {
            into 'project/'
        }
        from('tests') {
            into 'project/tests'
        }
    }
    containerDependencies.dependsOn = [deephavenDocker.healthyTask]
    dockerfile {
        from('deephaven/cpp-client:local-build')
        copyFile('project/requirements-dev.txt', '/project/requirements-dev.txt')
        runCommand('pip3 install -r /project/requirements-dev.txt')
        copyFile('*.whl', '/')
        runCommand('pip3 install /pydeephaven*.whl')
        workingDir('/project')
        environmentVariable 'DH_HOST', deephavenDocker.containerName.get()
        environmentVariable 'DH_PORT', '10000'
    }
    parentContainers = [ project.tasks.getByPath(':cpp-client:cppClient') ]
    entrypoint = ['python3', '-m', 'xmlrunner', 'discover', 'tests', '-v', '-o', '/out/report']
    copyOut {
        into layout.buildDirectory.dir('test-results')
    }
}
```

py/client-ticking/build.gradle
jcferretti: We discussed over slack, doing that results in gradle failure:
```
cfs@caicai 16:33:34 ~/dh/oss1/deephaven-core
$ git diff
diff --git a/py/client-ticking/build.gradle b/py/client-ticking/build.gradle
index 650973a94c..95584ddbe6 100644
--- a/py/client-ticking/build.gradle
+++ b/py/client-ticking/build.gradle
@@ -89,7 +89,7 @@ def testPyClientTicking = Docker.registerDockerTask(project, 'testPyClientTickin
     containerDependencies.dependsOn = [deephavenDocker.healthyTask]
-    parentContainers = [ project.tasks.getByName('pyClientTicking') ]
+    parentContainers = [ buildPyClientTicking ]
     entrypoint = ['python3', '-m', 'xmlrunner', 'discover', 'tests', '-v', '-o', '/out/report']
     copyOut {
         into layout.buildDirectory.dir('test-results')


```

```
cfs@caicai 16:33:36 ~/dh/oss1/deephaven-core
$ ./gradlew :py-client-ticking:testPyClientTicking

* What went wrong:
A problem occurred configuring project ':py-client-ticking'.
> Could not create task ':py-client-ticking:testPyClientTicking'.
   > Could not create task ':py-client-ticking:testPyClientTickingMakeImage'.
      > class org.gradle.api.internal.tasks.DefaultTaskContainer$TaskCreatingProvider_Decorated cannot be cast to class org.gradle.api.Task (org.gradle.api.internal.tasks.DefaultTaskContainer$TaskCreatingProvider_Decorated and org.gradle.api.Task are in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader @66cd51c3)
[...]
```

py/client/build.gradle
jcferretti: See previous answer.

py/client/build.gradle
jcferretti: Done.

niloc132 (approved):

kosak (approved):

jmao-denver (approved):

devinrsmith (approved):

py/client-ticking/build.gradle
```diff
@@ -0,0 +1,114 @@
...
+                        /out \\
+                        ${prefix}/log \\
+                        ${prefix}/src/py-client-ticking/src \\
+                        ${prefix}/src/py-client-ticking/in-wheels
+
+                   """)
+        copyFile('setup.py', "${prefix}/src/py-client-ticking")
+        copyFile('README.md', "${prefix}/src/py-client-ticking")
+        copyFile('src/', "${prefix}/src/py-client-ticking/src/")
+        copyFile('wheels/', "${prefix}/src/py-client-ticking/in-wheels")
```
mofojed, line 58: Any reason we change the dest dir to `in-wheels` instead of just keeping the same name `wheels`?

At least we should have the trailing slash in the destination for consistency.

jcferretti: The `in-wheels` was because in that image creation script, the wheel is not one that is being created in the image and for output, but one that is coming from a different gradle target as input (to be installed in the image, as opposed to be created by it)


It is a temporary directory in the image that gets removed before the image is published; just a staging area for the wheel coming form py/client.

PR #4699, opened 2023-10-24
Missing TableUpdateValidator unit test.

cpwright:


engine/table/src/test/java/io/deephaven/engine/table/impl/TestTableUpdateValidator.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.Random;
+
+import static io.deephaven.engine.testutil.TstUtils.*;
+import static io.deephaven.engine.util.TableTools.intCol;
+
+/**
+ * Test that the TableUpdateValidator can be put in the middle of an operation.
+ */
+public class TestTableUpdateValidator extends QueryTableTestBase {
+    @Test
```
rcaudy, line 27: If you feel like it, you could delete these annotations, since IntelliJ doesn't like them on a JUnit3 style test.

rcaudy (approved):

PR #4707, opened 2023-10-24
Support Optional annotation

### LINKED ISSUE #3622: Python methods annotated with `Optional` return typing
I'd expect that explicitly annotating a method with an `Optional` return type should be supported, and correctly deduce the return type (given that Deephaven columns support null values).

```python
from typing import Optional
from deephaven.column import bool_col

def my_bool_func(foo, bar) -> Optional[bool]:
    if foo:
        return True
    if bar:
        return False
    return None

def my_bool_func_bad_hint(foo, bar) -> bool:
    return my_bool_func(foo, bar)

bool_results = new_table(
    [
        bool_col("X", [True, True, False, False]),
        bool_col("Y", [True, False, True, False]),
    ]
).update_view(
    [
        "MyBoolFunc=my_bool_func(X, Y)",
        "MyBoolFuncBadHint=my_bool_func_bad_hint(X, Y)",
        "MyBoolFuncExplicit=(Boolean)my_bool_func(X, Y)",
    ]
)
```

In this case, `MyBoolFunc` is a `PyObject`, while `MyBoolFuncBadHint` / `MyBoolFuncExplicit` are more correctly `Boolean`. I think all `MyBoolFunc` should ideally operate the same way `MyBoolFuncExplicit` does.

image

```python
from typing import Optional
from deephaven.column import long_col

def my_long_func(foo, bar) -> Optional[int]:
    if foo >= 100:
        return foo
    if bar >= 100:
        return bar
    return None

def my_long_func_bad_hint(foo, bar) -> int:
    return my_long_func(foo, bar)

long_results = new_table(
    [
        long_col("X", [101, 102, 99, 98]),
        long_col("Y", [103, 99, 104, 97]),
    ]
).update_view(
    [
        "MyLongFunc=my_long_func(X, Y)",
        "MyLongFuncBadHint=my_long_func_bad_hint(X, Y)",
        "MyLongFuncExplicit=(long)my_long_func(X, Y)",
    ]
)
```

In this case, `MyLongFunc` is a `PyObject`, ` MyLongFuncBadHint` is a `long` column but has mapped `None` to 0, and `MyLongFuncExplicit` is a `long` column and has mapped `None` to `null`. I think `MyLongFunc` and `MyLongFuncBadHint` should ideally operate the same way that `MyLongFuncExplicit` does.

image

jmao-denver:
Fixes #3622

### DIFF
--- a/py/server/deephaven/dtypes.py
+++ b/py/server/deephaven/dtypes.py
@@ -333,6 +333,7 @@ def _scalar(x: Any, dtype: DType) -> Any:
     """Converts a Python value to a Java scalar value. It converts the numpy primitive types, string to
     their Python equivalents so that JPY can handle them. For datetime values, it converts them to Java Instant.
     Otherwise, it returns the value as is."""
+
     # NULL_BOOL will appear in Java as a byte value which causes a cast error. We just let JPY converts it to Java null
     # and the engine has casting logic to handle it.
     if x is None and dtype != bool_ and _PRIMITIVE_DTYPE_NULL_MAP.get(dtype):
@@ -353,8 +354,7 @@ def _scalar(x: Any, dtype: DType) -> Any:
             elif x.dtype.char == 'M':
                 from deephaven.time import to_j_instant
                 return to_j_instant(x)
-        else:
-            if isinstance(x, (datetime.datetime, pd.Timestamp)):
+        elif isinstance(x, (datetime.datetime, pd.Timestamp)):
                 from deephaven.time import to_j_instant
                 return to_j_instant(x)
         return x
--- a/py/server/deephaven/dtypes.py
+++ b/py/server/deephaven/dtypes.py
@@ -235,7 +235,7 @@ def _instant_array(data: Sequence) -> jpy.JType:
     return jpy.array(Instant.j_type, data)
 
 
-def array(dtype: DType, seq: Sequence, remap: Callable[[Any], Any] = None) -> jpy.JType:
+def array(dtype: DType, seq: Optional[Sequence], remap: Callable[[Any], Any] = None) -> Optional[jpy.JType]:
     """ Creates a Java array of the specified data type populated with values from a sequence.
 
     Note:
@@ -253,6 +253,9 @@ def array(dtype: DType, seq: Sequence, remap: Callable[[Any], Any] = None) -> jp
     Raises:
         DHError
     """
+    if seq is None:
+        return None
+
     if isinstance(seq, np.ndarray) and seq.ndim > 1:
         raise ValueError("array() does not support multi-dimensional arrays")
 
--- a/py/server/deephaven/dtypes.py
+++ b/py/server/deephaven/dtypes.py
@@ -329,30 +329,36 @@ _NUMPY_INT_TYPE_CODES = ["i", "l", "h", "b"]
 _NUMPY_FLOATING_TYPE_CODES = ["f", "d"]
 
 
-def _scalar(x):
+def _scalar(x: Any, dtype: DType) -> Any:
     """Converts a Python value to a Java scalar value. It converts the numpy primitive types, string to
     their Python equivalents so that JPY can handle them. For datetime values, it converts them to Java Instant.
     Otherwise, it returns the value as is."""
-    if hasattr(x, "dtype"):
-        if x.dtype.char in _NUMPY_INT_TYPE_CODES:
-            return int(x)
-        elif x.dtype.char in _NUMPY_FLOATING_TYPE_CODES:
-            return float(x)
-        elif x.dtype.char == '?':
-            return bool(x)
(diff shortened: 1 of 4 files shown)


py/server/deephaven/table.py
```diff
@@ -366,6 +366,20 @@ def _j_py_script_session() -> _JPythonScriptSession:
...
+def _parse_annotation(annotation: Any) -> Any:
+    """Parse a Python annotation, for now mostly to extract the non-None type from an Optional(Union) annotation,
+    otherwise return the original annotation. """
+    if isinstance(annotation, _GenericAlias) and annotation.__origin__ == Union and len(annotation.__args__) == 2:
+        if annotation.__args__[1] == type(None):  # noqa: E721
+            return annotation.__args__[0]
+        elif annotation.__args__[0] == type(None):  # noqa: E721
+            return annotation.__args__[1]
+        else:
+            return annotation
```
chipkent, line 378: Is it possible to get a pathological `Union[None,None]` type, and do we handle it well?

py/server/tests/test_vectorization.py
```diff
@@ -265,6 +266,15 @@ def sinc2(x):
         t = empty_table(100).update(["X = 0.1 * i", "SincXS=((sinc2(X)))"])
         self.assertEqual(t.columns[1].data_type, dtypes.PyObject)
 
+    def test_optional_annotations(self):
```
chipkent, line 269: Should we also be testing `None` return values getting converted to DH Nulls?

devinrsmith (dismissed):
Tested against my original issue, and it works as expected. :+1:

py/server/deephaven/table.py
jmao-denver: yes, in that case, the NoneType will be returned and eventually falls back to being 'object'.

py/server/tests/test_vectorization.py
```diff
@@ -265,6 +266,16 @@ def sinc2(x):
...
+        def pyfunc(p1: np.int32, p2: np.int32, p3: Optional[np.int32]) -> Optional[int]:
+            total = p1 + p2 + p3
+            return None if total % 3 == 0 else total
+
+        t = empty_table(10).update("X = i").update(["Y = pyfunc(X, i, 13)", "Z = pyfunc(X, ii, 66)"])
+        self.assertEqual(deephaven.table._vectorized_count, 2)
+        self.assertIn("13", t.to_string(cols=["Y"]))
+        self.assertIn("null", t.to_string())
```
chipkent, line 277: Should assert the output column types.

py/server/deephaven/dtypes.py
```diff
@@ -326,30 +329,36 @@ def from_np_dtype(np_dtype: Union[np.dtype, pd.api.extensions.ExtensionDtype]) -
...
+                return bool(x)
+            elif x.dtype.char == 'U':
+                return str(x)
+            elif x.dtype.char == 'O':
+                return x
+            elif x.dtype.char == 'M':
+                from deephaven.time import to_j_instant
+                return to_j_instant(x)
         else:
-            raise TypeError(f"Unsupported dtype: {x.dtype}")
```
chipkent, line 348: I pulled up the rendered code, and I think it would be more clear as:
```python
if hasattr(x, "dtype"):
  ...
elif isinstance(x, (datetime.datetime, pd.Timestamp)):
  ...
else:
  return x

File: py/server/deephaven/dtypes.py [hunk 5]
```diff
@@ -326,30 +329,36 @@ def from_np_dtype(np_dtype: Union[np.dtype, pd.api.extensions.ExtensionDtype]) -
...
-        elif x.dtype.char == '?':
-            return bool(x)
-        elif x.dtype.char == 'U':
-            return str(x)
-        elif x.dtype.char == 'O':
-            return x
-        elif x.dtype.char == 'M':
-            from deephaven.time import to_j_instant
-            return to_j_instant(x)
+    # NULL_BOOL will appear in Java as a byte value which causes a cast error. We just let JPY converts it to Java null
```
chipkent, line 336:
```suggestion

    # NULL_BOOL will appear in Java as a byte value which causes a cast error. We just let JPY converts it to Java null
```
Makes the rendered code more clear.

py/server/deephaven/dtypes.py
chipkent, line 348: Are there any bad scenarios that resut from returning unsupported values instead of erroring?

jmao-denver: Actually we should not fail a query just because we can't convert a value to a Java object of a specific type other than the catch-all PyObject.

chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3350
Reference: https://github.com/deephaven/deephaven.io/issues/3349

Issue #4689, opened 2023-10-23
R Client Doc Generation


alexpeters1208:
There was discussion in the R Client slack channel on how the generated `.Rd` files that live in `deephaven-core/R/rdeephaven/man` should be checked in. There are three options:
1. Generate files locally and check them into GitHub explicitly.
2. Set up a CI task to generate the files from source, and have CI check them in.
3. Don't check anything in at all, and leave this file generation to the user.

The first option presents the possibility of a dev forgetting to regenerate the files, forgetting to check them in, etc. This could lead to an inconsistency in source code vs generated documentation, which is not ideal.
The third option requires users to have the `devtools` tool chain to generate the docs, as well as the knowledge that the docs need to be generated. No existing popular R package does this - they all have their generated files sitting in GitHub. This is the least viable option.

That leaves option two. This is probably the best option, and ensures that the generated documentation will always be up-to-date with the source code. Documentation in this case should still be generated locally to verify formatting and correctness, but should not be checked into github directly.

The documentation gets generated by running `devtools::document()` from the root dictory of the R package, and it should be relatively easy to get Docker to do this for us. Running this command successfully requires an already-built package (why??), so it cannot be easily separated from the CI for the R client itself.

alexpeters1208:
This part of the process is complete, and what is left to be done is described [here](https://github.com/deephaven/deephaven-core/issues/4778).

PR #4713, opened 2023-10-24
Create a dockerized gradle task to update the generated R client documentation.

jcferretti:
Towards #4689

This PR adds two new gradle tasks to `R/build.gradle`:
* `rClientDoc` creates a docker image with R doc support and runs it in a container generating a tar bundle in `build/man/man.tar.gz` with the updated documentation.
* `updateRClientDoc` overwrites the contents of `R/rdeephaven/man` with the newly generated doc by the task above.

You can run either (they have proper dependencies) like so, from the root of a `deephave-core` clone:
`./gradlew :R:rClientDoc`
`./gradlew :R:updateRClientDoc`

Notes

* Something that is less than ideal in terms of dependencies is that R/roxygen2 (the package/tool that generates doc) requires the package that we intend to document (eg, rdeephaven) to be already installed.  This implies we are forced to have the image that supports generating documentation depend on the image that builds the rdeephaven package; this is unfortunate because the image that supports generating documentation has to install a lot of base OS packages and then build the roxygen2 R package; all this does not depend on any way on our own rdeephaven sources, however, since the image depends on a base image that builds from those sources, any changes in rdeephaven sources will trigger dependencies in a way that will regenerate the documentation base image which will require reinstalling the OS base packages and roxygen2 in R, etc.  We could make this less expensive by making all these additional OS packages and R roxygen package installation be part already of the `cpp-clients-multi-base` image from the `deephaven-base-images` repository... but then we will be taxing every deephaven developer who does not need to work on R related changes with a constant tax.  So I think the tradeoff right now is to make this slightly less convenient to R developers only on the specific instance when doc needs to be regenerated in favor of avoiding that tax for everybody else on any and every change.

### DIFF
--- a/R/build.gradle
+++ b/R/build.gradle
@@ -33,7 +33,6 @@ def buildRClient = Docker.registerDockerTask(project, 'rClient') {
     copyIn {
         from(layout.projectDirectory) {
             include 'r-build.sh'
-            include 'r-tests.sh'
             include 'rdeephaven/DESCRIPTION'
             include 'rdeephaven/LICENSE'
             include 'rdeephaven/NAMESPACE'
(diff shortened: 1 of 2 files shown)


alexpeters1208 (approved):
LGTM

PR #4715, opened 2023-10-25
Barrage Protocol: Fix Writer of `byte[]` to Use Var Binary Encoding

nbauernfeind:
Thanks to @abaranec who noticed the lack of symmetry between writer and reader.

Indeed the transport VarBinary encoding is not equal to VarList<Byte> encoding as verified by the test I added that broke.

I also verified that the UI continues to properly handle byte[] (it uses a stringified preview column).


lbooker42 (approved):
LGTM

PR #4722, opened 2023-10-25
Vector Expansion Kernel O(n^2) -> O(n)

nbauernfeind:
This is, unfortunately, an identical issue to #2458 but for the `Vector` kernel instead of the `Array` kernel.

This is what caused @lbooker42's performance issues found while testing grouping changes.


extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/vector/ByteVectorExpansionKernel.java
```diff
@@ -37,24 +39,31 @@ public <A extends Any> WritableChunk<A> expand(
...
+
         perElementLengthDest.setSize(source.size() + 1);
         for (int i = 0; i < typedSource.size(); ++i) {
             final ByteVector row = typedSource.get(i);
-            final int len = row == null ? 0 : row.intSize("ByteVectorExpansionKernel");
-            perElementLengthDest.set(i, lenWritten);
-            final WritableByteChunk<A> result = resultWrapper.ensureCapacityPreserve(lenWritten + len);
-            for (int j = 0; j < len; ++j) {
-                result.set(lenWritten + j, row.get(j));
+            perElementLengthDest.set(i, result.size());
```
rcaudy, line 55: To confirm, these are cumulative lengths?

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/vector/ByteVectorExpansionKernel.java
```diff
@@ -37,24 +39,31 @@ public <A extends Any> WritableChunk<A> expand(
...
+            if (row == null) {
+                continue;
+            }
+            final ByteConsumer consumer = result::add;
```
rcaudy, line 59: Charles made a claim recently that using set and a single setSize was more performant. I don't fully believe it, especially in a case like this one where the counter has to live somewhere.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/vector/ObjectVectorExpansionKernel.java
```diff
@@ -4,15 +4,17 @@
...
-import io.deephaven.chunk.ChunkType;
 import io.deephaven.chunk.IntChunk;
 import io.deephaven.chunk.ObjectChunk;
 import io.deephaven.chunk.WritableChunk;
 import io.deephaven.chunk.WritableIntChunk;
 import io.deephaven.chunk.WritableObjectChunk;
 import io.deephaven.chunk.attributes.Any;
 import io.deephaven.chunk.attributes.ChunkPositions;
-import io.deephaven.chunk.sized.SizedChunk;
+import io.deephaven.engine.primitive.function.CharConsumer;
```
rcaudy, line 14: Unused imports here. Not replicated.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/vector/ByteVectorExpansionKernel.java
nbauernfeind: Yes, these are offsets into the inner buffer.

rcaudy (approved):

PR #4735, opened 2023-10-27
Ensure defaults for float/double generation result in random values

niloc132:
Also ensures generating BigIntegers with a default range can contain some nulls.

Nightlies are currently running at https://github.com/niloc132/deephaven-core/actions/runs/6671871431

### DIFF
--- a/engine/test-utils/src/main/java/io/deephaven/engine/testutil/generator/DoubleGenerator.java
+++ b/engine/test-utils/src/main/java/io/deephaven/engine/testutil/generator/DoubleGenerator.java
@@ -74,7 +74,7 @@ public class DoubleGenerator extends AbstractGenerator<Double> {
                 return Double.POSITIVE_INFINITY;
             }
         }
-        return from + (random.nextDouble() * to - from);
+        return (from / 2 + (random.nextDouble() * (to / 2 - from / 2))) * 2;
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


rcaudy (approved):
I like it enough, if Cristian does.

jcferretti (approved):
Maybe also remove that "1" in the constructor we discussed about, has no value/place in there.

niloc132:
One more set of nightly runs with the better default lower bound https://github.com/niloc132/deephaven-core/actions/runs/6710144326

rcaudy (approved):

engine/test-utils/src/main/java/io/deephaven/engine/testutil/generator/DoubleGenerator.java
```diff
@@ -22,7 +22,7 @@ public class DoubleGenerator extends AbstractGenerator<Double> {
     private final double posInfFraction;
 
     public DoubleGenerator() {
-        this(QueryConstants.NULL_DOUBLE + 1, Double.MAX_VALUE);
+        this(Math.nextUp(-Double.MAX_VALUE), Double.MAX_VALUE);
```
rcaudy, line 25: By the way, we have a constant for this:
`QueryConstants.MIN_FINITE_DOUBLE`

rcaudy (approved):

PR #4748, opened 2023-10-30
Add FilterIn support via java-client

devinrsmith:


java-client/session/src/main/java/io/deephaven/client/impl/BatchTableRequestBuilder.java
```diff
@@ -807,9 +808,15 @@ public Condition visit(FilterIsNull isNull) {
         @Override
         public Condition visit(FilterComparison comparison) {
             FilterComparison preferred = comparison.maybeTranspose();
+            Operator operator = preferred.operator();
+            if (operator == Operator.EQUALS) {
```
rcaudy, line 812: What about not equals?

java-client/session/src/main/java/io/deephaven/client/impl/BatchTableRequestBuilder.java
```diff
@@ -818,8 +825,18 @@ public Condition visit(FilterComparison comparison) {
 
         @Override
         public Condition visit(FilterIn in) {
-            // TODO(deephaven-core#3609): Update gRPC expression / filter / literal structures
```
rcaudy, line 821: Lots more cases still UOE. Do we expect users to hit them soon?

java-client/session/src/main/java/io/deephaven/client/impl/BatchTableRequestBuilder.java
```diff
@@ -818,8 +825,18 @@ public Condition visit(FilterComparison comparison) {
...
-            throw new UnsupportedOperationException("Can't build Condition with FilterIn");
+            // Note: we might think that simplifying into an equals comparison would be simpler than a FilterIn, but the
+            // current implementation does not optimize an EQUALS FilterComparison, but _does_ optimize InCondition.
+            // See io.deephaven.server.table.ops.filter.FilterFactory
+            // if (in.values().size() == 1) {
+            // return visit(in.asEquals());
+            // }
+            final InCondition.Builder builder = InCondition.newBuilder()
```
rcaudy, line 834: What happens if LHS isn't a column?

devinrsmith: The server side handles this:

```
        // if a "in" expression has a non-reference on the left or reference on the right, flip it, and split
        // up values so these can be left as INs or remade into EQs, and join them together with OR/ANDs.
        filter = FlipNonReferenceMatchExpression.exec(filter);

...
        // for any "in" expression (at this point, all have a reference on the left), if they have a reference
        // value on the left it must be split into its own "equals" instead.
        filter = ConvertInvalidInExpressions.exec(filter);
```

java-client/session/src/main/java/io/deephaven/client/impl/BatchTableRequestBuilder.java
devinrsmith: Probably. Would like to consider https://github.com/deephaven/deephaven-core/issues/3609 sooner rather than later.

rcaudy (commented):
Caught up

java-client/session/src/main/java/io/deephaven/client/impl/BatchTableRequestBuilder.java
devinrsmith: I'm not sure what you mean. A FilterIn that represents `1 in Foo, Bar, Baz` *is* valid.

devinrsmith: Ideally, this layer would not need to do _any_ "fixing up", and it would simple pass the representation over the wire and back into the server exactly the same as it received.

The only reason I'm "fixing" the equals case to be an filter-in at _this_ level (and not the server) is b/c the current server implementation layer is outdated IMO and it's much easier to work w/ the `io.deephaven.api.filter.Filter` objects. If we'd prefer, I can remove my optimization case and fix the server side impl instead:

```java
        // replace any EQ-type expression with its corresponding IN-type expression. this preserves the changes
        // made above, could be moved earlier in this list, but must come before "in"/"not in"s are merged
        // TODO (deephaven-core#733)
```

Although, I'm hoping we can deprecate this layer and implement https://github.com/deephaven/deephaven-core/issues/3609 sooner rather than later.

rcaudy (approved):

PR #4752, opened 2023-10-31
Select: Reuse Results When Possible

### LINKED ISSUE #4751: SELECT_STATIC May Flatten a Source Column More Than Once
Under a debugger `TestSelectPreserveGrouping#testPreserveDeferredGrouping` demonstrates that a select on a static table re-flattens a source column more than once.

Here is a groovy example:
```groovy
source = emptyTable(10).updateView("I = ii").where("I % 2 == 0")
result = source.select("Foo = I", "Bar = Foo", "Baz = I")
println(result.getRowSet()) // ensure flat
println("Foo == Bar ?? " + result.getColumnSource("Foo") == result.getColumnSource("Bar"))
println("Foo == Baz ?? " + result.getColumnSource("Foo") == result.getColumnSource("Baz"))
```

Related (kind of) to #3438 which would prevent this kind of optimization as we split formula-array access into three separate operations without being smart about where exactly we compute the select columns.

nbauernfeind:
Fixes #4751; enabling re-use of an already flattened parent column.

This also fixes:
- an infinite recursion when flattening is reverted (it was recursively calling the wrong version of the method)
- the loss of internally flattening if a SourceColumn refers to a new, already flattened, result column (cause this is safe)

Nightlies: https://github.com/nbauernfeind/deephaven-core/actions/runs/6746760106/

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/PreserveColumnLayer.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/PreserveColumnLayer.java
@@ -77,10 +77,10 @@ final public class PreserveColumnLayer extends DependencyLayerBase {
 
     @Override
     public boolean flattenedResult() {
-        // a preserve layer is only flattened if the inner is flattened
-        // the "flattenedResult" means that we are flattening the table as part of select.  For a pre-existing column, we
-        // could not preserve a layer while flattening, but if we are preserving a newly generated column; it is valid for
-        // the result to have been flattened as part of select.
+        // preserve layer is only flattened if the inner is flattened
+        // the "flattenedResult" means that we are flattening the table as part of select. For a pre-existing column, we
+        // could not preserve a layer while flattening, but if we are preserving a newly generated column; it is valid
+        // for the result to have been flattened as part of select.
         return inner.flattenedResult();
     }
 
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/PreserveColumnLayer.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/PreserveColumnLayer.java
@@ -24,13 +24,11 @@ import java.util.Map;
  */
 final public class PreserveColumnLayer extends DependencyLayerBase {
     private final BitSet dependencyBitSet;
-    private final boolean flattenedResult;
 
     PreserveColumnLayer(SelectAndViewAnalyzer inner, String name, SelectColumn sc, ColumnSource<?> cs, String[] deps,
-            ModifiedColumnSet mcsBuilder, boolean flattenedResult) {
+            ModifiedColumnSet mcsBuilder) {
         super(inner, name, sc, cs, deps, mcsBuilder);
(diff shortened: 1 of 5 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/SelectAndViewAnalyzer.java
```diff
@@ -189,6 +222,12 @@ public static SelectAndViewAnalyzerWrapper create(
                     final WritableColumnSource<?> scs =
                             flatResult || flattenedResult ? sc.newFlatDestInstance(targetDestinationCapacity)
                                     : sc.newDestInstance(targetDestinationCapacity);
+
+                    if (flattenedResult && realColumn != null && !resultColumns.contains(realColumn.getSourceName())) {
+                        // this source column a candidate for preservation if referenced again
+                        resultAlias.put(realColumn.getSourceName(), scs);
```
cpwright, line 228: Why is the source column a candidate for preservation and not the destination name?

cpwright: I think the optimization holds true for any preserved column, no matter the mode.  It is a bit odd to make copies of a column, as you are not adding any meaningful data.  If you were trying to capture some kind of stateful function it would change the answer; but I think if you do that maybe you get what you deserve.

cpwright (dismissed):

cpwright (approved):

engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/PreserveColumnLayer.java
```diff
@@ -75,6 +75,17 @@ public LogOutput append(LogOutput logOutput) {
                 .append("}");
     }
 
+    @Override
+    public boolean flattenedResult() {
+        // a preserve layer is only flattened if the inner is flattened
```
cpwright, line 80:
```suggestion
        // a preserve layer is only flattened if the inner is flattened
        // the "flattenedResult" means that we are flattening the table as part of select.  For a pre-existing column, we
        // could not preserve a layer while flattening, but if we are preserving a newly generated column; it is valid for
        // the result to have been flattened as part of select.
```

engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/SelectAndViewAnalyzer.java
```diff
@@ -156,30 +163,59 @@ public static SelectAndViewAnalyzerWrapper create(
...
-                            false, selectColumns);
+                            useShiftedColumns, false, selectColumns);
+                }
+
+                analyzer = analyzer.createLayerForPreserve(
+                        sc.getName(), sc, sc.getDataView(), distinctDeps, mcsBuilder);
+
+                if (!sourceIsNew) {
+                    // we can not flatten future columns because we are preserving a column that may not be flat
+                    flattenedResult = false;
```
cpwright, line 189: Move into if after line 180.

cpwright (approved):

PR #4755, opened 2023-11-01
Added support for Parquet DATE

### LINKED ISSUE #4753: Support for Parquet DATE and TIME types
A customer needs us to support the Parquet [`DATE`](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#date) type. (A Java `java.time.LocalDate` column should be written as a Parquet `DATE`, and a Parquet `DATE` column should be read back into Java `LocalDate` columns.)

We might as well handle `java.time.LocalTime`/Parquet [`TIME`](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#time) while we're at it.

Relates to #3367.

malhotrashivam:
Related to #4753

### DIFF
--- a/engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
+++ b/engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
@@ -1037,6 +1037,19 @@ public class DateTimeUtils {
     }
 
     /**
+     * Returns number of days from the Epoch for a {@link LocalDate} value as an int.
+     *
+     * @param date date to compute the Epoch offset for
+     * @return days since Epoch, or a {@link QueryConstants#NULL_INT NULL_INT} value if the instant is {@code null}
+     */
+    public static int epochDays(@Nullable final LocalDate date) {
+        if (date == null) {
+            return io.deephaven.util.QueryConstants.NULL_INT;
+        }
+        return Math.toIntExact(date.toEpochDay());
+    }
+
+    /**
      * Converts nanoseconds from the Epoch to an {@link Instant}.
      *
      * @param nanos nanoseconds since Epoch
--- a/engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
+++ b/engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
@@ -24,6 +24,7 @@ import java.util.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import static io.deephaven.util.QueryConstants.NULL_INT;
 import static io.deephaven.util.QueryConstants.NULL_LONG;
 import static java.time.format.DateTimeFormatter.*;
 
@@ -1037,14 +1038,27 @@ public class DateTimeUtils {
     }
 
     /**
-     * Returns number of days from the Epoch for a {@link LocalDate} value as an int.
+     * Returns number of days from the Epoch for a {@link LocalDate} value.
+     *
+     * @param date date to compute the Epoch offset for
+     * @return days since Epoch, or a {@link QueryConstants#NULL_LONG NULL_LONG} value if the instant is {@code null}
+     */
+    public static long epochDays(@Nullable final LocalDate date) {
+        if (date == null) {
+            return NULL_LONG;
+        }
+        return date.toEpochDay();
+    }
+
+    /**
+     * Returns number of days (as an {@code int}) from the Epoch for a {@link LocalDate} value.
      *
      * @param date date to compute the Epoch offset for
      * @return days since Epoch, or a {@link QueryConstants#NULL_INT NULL_INT} value if the instant is {@code null}
      */
-    public static int epochDays(@Nullable final LocalDate date) {
+    public static int epochDaysAsInt(@Nullable final LocalDate date) {
         if (date == null) {
-            return io.deephaven.util.QueryConstants.NULL_INT;
+            return NULL_INT;
         }
         return Math.toIntExact(date.toEpochDay());
     }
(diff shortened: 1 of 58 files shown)


extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetSchemaReader.java
```diff
@@ -337,7 +337,7 @@ public Optional<Class<?>> visit(
 
             @Override
             public Optional<Class<?>> visit(final LogicalTypeAnnotation.DateLogicalTypeAnnotation dateLogicalType) {
-                return Optional.of(int.class);
+                return Optional.of(java.time.LocalDate.class);
```
rcaudy, line 340: Might as well import the class, rather than having a FQN here.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/topage/ToDatePageFromInt.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    @NotNull
+    public final LocalDate[] convertResult(final Object result) {
+        final int[] from = (int[]) result;
+        final LocalDate[] to = new LocalDate[from.length];
+
+        for (int i = 0; i < from.length; ++i) {
+            to[i] = LocalDate.ofEpochDay(from[i]);
```
rcaudy, line 54: I think this requires `null` handling, unless I'm mistaken. You'll need to map `QueryConstants.NULL_INT` to `null`. I think you should be adding a method in `DateTimeUtils` that you can call for this.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/topage/ToDatePageFromInt.java
```diff
@@ -0,0 +1,59 @@
...
+        throw new IllegalArgumentException("The native type for a Date column is " + nativeType.getCanonicalName());
+    }
+
+    private ToDatePageFromInt() {}
+
+    @Override
+    @NotNull
+    public final Class<LocalDate> getNativeType() {
+        return LocalDate.class;
+    }
```
rcaudy, line 33: This is correct as-is. At some point we might want to consider whether to keep things as `int` longer, but that's a bigger change.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetSchemaReader.java
rcaudy, line 340: This represents a behavior change. We need to be sure it's noted in release notes as a potentially breaking change for users, and that any associated documentation is updated.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/DateArrayTransfer.java
```diff
@@ -0,0 +1,37 @@
...
+        return data.length;
+    }
+
+    @Override
+    void resizeBuffer(final int length) {
+        buffer = IntBuffer.allocate(length);
+    }
+
+    @Override
+    void copyToBuffer(final @NotNull EncodedData<LocalDate[]> data) {
```
rcaudy, line 32:
```suggestion
    void copyToBuffer(@NotNull final EncodedData<LocalDate[]> data) {
```

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -2650,4 +2652,36 @@ private void assertBigIntegerColumnStatistics(SerialObjectColumnIterator<BigInte
...
+            dateArr[ii] = LocalDate.now();
+            // timeArr[ii] = LocalTime.now();
+        }
+        QueryScope.addParam("dateArr", dateArr);
+        // QueryScope.addParam("timeArr", timeArr);
+        final Table table = TableTools.emptyTable(NUM_ROWS).view(
+                "Date = java.time.LocalDate.now()",
+                // "nullDate = (java.time.LocalDate)null"
+                "DateArray = dateArr");
+        final File dest = new File("/Users/shivammalhotra/Documents/dates.parquet"); // TOD change it back to local dir
```
rcaudy, line 2673: This is clearly not ready to merge as committed. Please be sure to fix both test file paths, and test coverage for the new/changed code.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -1036,6 +1036,19 @@ public static long epochSeconds(@Nullable final ZonedDateTime dateTime) {
         return dateTime.toEpochSecond();
     }
 
+    /**
+     * Returns number of days from the Epoch for a {@link LocalDate} value as an int.
+     *
+     * @param date date to compute the Epoch offset for
+     * @return days since Epoch, or a {@link QueryConstants#NULL_INT NULL_INT} value if the instant is {@code null}
+     */
+    public static int epochDays(@Nullable final LocalDate date) {
```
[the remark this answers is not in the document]
rcaudy: My preference is to do 2, rather than 1.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/DateArrayTransfer.java
```diff
@@ -12,10 +12,14 @@
 import java.time.LocalDate;
 
 final class DateArrayTransfer extends PrimitiveArrayAndVectorTransfer<LocalDate[], LocalDate[], IntBuffer> {
+    private static int getMaxValuesPerPage(@NotNull final RowSequence tableRowSet, final int targetPageSize) {
```
rcaudy, line 15: Why do we do this for date array/vector, and not other array/vector transfers?

rcaudy (commented):
.

py/server/tests/test_parquet.py
```diff
@@ -354,6 +354,38 @@ def test_dictionary_encoding(self):
...
+        df_from_pandas_as_str = df_from_pandas.astype(str)
+        self.assertTrue((df_from_disk_as_str == df_from_pandas_as_str).all().values.all())
+
+        # Rewrite the dataframe back to parquet using pyarrow and read it back using deephaven.parquet to compare
+        df_from_pandas.to_parquet('data_from_pandas.parquet', compression='SNAPPY')
+        from_disk_pandas = read('data_from_pandas.parquet')
+
+        # Compare only the date column since the null column is written as different logical types by pandas and
+        # deephaven
+        self.assertTrue(dh_table.select("someDateColumn"), from_disk_pandas.select("someDateColumn"))
```
jmao-denver, line 387:
```suggestion
        self.assertEqual(dh_table.select("someDateColumn"), from_disk_pandas.select("someDateColumn"))
```

rcaudy (dismissed):

rcaudy (approved):

PR #4756, opened 2023-11-01
Add java client TableServiceAsync

devinrsmith:


java-client/session/src/main/java/io/deephaven/client/impl/ExportStates.java
```diff
@@ -133,19 +135,39 @@ synchronized List<Export> export(ExportsRequest requests) {
...
-            ++batchCount;
-        }
+            @Override
+            public void send() {
+                send.run();
+            }
 
-        return results;
+            @Override
+            public void close() {
```
rcaudy, line 157: We have the potential for state leak if the user calls close without calling send. Do we need to clean those up? What about on errors?

java-client/session/src/main/java/io/deephaven/client/impl/SessionImpl.java
```diff
@@ -282,26 +278,10 @@ public CompletableFuture<Void> closeFuture() {
...
-        if (this.config.mixinStacktrace() == mixinStacktrace) {
-            return batchManager;
-        }
-        return TableHandleManagerBatch.of(this, mixinStacktrace);
-    }
-
-    @Override
-    public TableHandleManager serial() {
-        return serialManager;
+    protected TableServices delegate() {
```
rcaudy, line 281: So the pattern is we construct a whole bunch of state, and then throw it away...

devinrsmith: Yes. The internal state _is_ needed even for a single query - there may be shared sub-DAGs that we don't want to export multiple times in a single Batch request. One might argue "we should re-implement all of this logic, only keeping the state/Map alive for the duration of the execute method", but re-working  as such would add additional maintenance burdens for the sake of saving the construction of a extra object, which is not even close to worth it IMO.

java-client/session/src/main/java/io/deephaven/client/impl/TableServiceAsync.java
```diff
@@ -0,0 +1,137 @@
...
+         * @return the table handle
+         * @throws InterruptedException if the current thread was interrupted while waiting
+         * @throws ExecutionException if the computation threw an exception
+         * @throws CancellationException if the computation was cancelled
+         */
+        static TableHandle get(TableHandleFuture future) throws InterruptedException, ExecutionException {
+            try {
+                return future.get();
+            } catch (InterruptedException e) {
+                cancelOrClose(future, true);
```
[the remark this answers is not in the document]
devinrsmith: I've built-in as `getOrCancel`

java-client/session/src/main/java/io/deephaven/client/impl/TableServiceAsync.java
```diff
@@ -0,0 +1,137 @@
...
+         * @param futures the futures
+         * @param mayInterruptIfRunning {@code true} if the thread executing the task should be interrupted; otherwise,
+         *        in-progress tasks are allowed to complete
+         */
+        static void cancelOrClose(Iterable<? extends TableHandleFuture> futures, boolean mayInterruptIfRunning) {
+            for (TableHandleFuture future : futures) {
+                future.cancel(mayInterruptIfRunning);
+            }
+            for (TableHandleFuture future : futures) {
+                cancelOrClose(future, mayInterruptIfRunning);
```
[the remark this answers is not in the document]
devinrsmith: I've changed the semantics and impl to make it a bit cleaner.

java-client/session/src/main/java/io/deephaven/client/impl/Session.java
```diff
@@ -73,6 +73,78 @@ public interface Session
...
+     * TableServices ts = session.tableServices();
+     * TableHandle h1 = ts.execute(TableSpec.emptyTable(42).view("I=ii").where("I % 2 == 0"));
+     * // This execution will throw an "unreferenceable table" exception.
+     * TableHandle h2 = ts.execute(TableSpec.emptyTable(42).view("I=ii"));
+     * </pre>
+     * 
+     * @return a new stateful table services
+     * @see <a href="https://github.com/deephaven/deephaven-core/issues/4733">deephaven-core#4733</a>
+     */
+    TableServices tableServices();
```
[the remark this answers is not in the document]
devinrsmith: I've updated it as such; the `Session#batch` and `Session#serial` are _not_ stateful now. To get them as stateful, you can do `Session.newStatefulTableServices().batch()` etc.

java-client/session/src/main/java/io/deephaven/client/impl/ExportStates.java
devinrsmith: Good catch. I've added some new logic to handle the case where the user does not call send. After sent, any errors are the handled by respective `io.deephaven.client.impl.ExportRequest.listener`.

java-client/session/src/main/java/io/deephaven/client/impl/ExportServiceRequest.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+package io.deephaven.client.impl;
+
+import java.io.Closeable;
+import java.util.List;
+
+interface ExportServiceRequest extends Closeable {
+    List<Export> exports();
+
+    // Note: not providing cancel() or Runnable as return here; we don't really want to cancel a Batch request, we'll
```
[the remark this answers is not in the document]
devinrsmith: I think we can debate this point once we hook up the on cancel handlers for the server, after the semantics of said cancel is clearly document.

java-client/session/src/main/java/io/deephaven/client/impl/ExportStates.java
```diff
@@ -224,7 +261,7 @@ class State {
         State(TableSpec table, int exportId) {
             this.table = Objects.requireNonNull(table);
             this.exportId = exportId;
-            this.children = new LinkedHashSet<>();
+            this.children = new CopyOnWriteArraySet<>();
```
rcaudy, line 264: I understand we want this to avoid CME's on remove during an iteration. Can we build a list to remove and do a removeAll instead? I'm worried about adding COW data structures with user-facing adds.

server/src/main/java/io/deephaven/server/util/Scheduler.java
```diff
@@ -68,12 +68,16 @@ public DelegatingImpl(ExecutorService serialExecutor, ScheduledExecutorService c
...
         }
 
-        ExecutorService serialDelegate() {
-            return serialDelegate;
-        }
-
-        ScheduledExecutorService concurrentDelegate() {
-            return concurrentDelegate;
+        @VisibleForTesting
+        public void shutdown() throws InterruptedException {
```
rcaudy, line 72: This is leaking in from another PR.

java-client/session/src/main/java/io/deephaven/client/impl/ExportStates.java
devinrsmith: I more thoroughly traced exactly the part that was causing concurrent mod, and it was actually the mitigation; now that the Batch / ETCR race has been fixed, I can get rid of the mitigation and revert this change.

java-client/session/src/main/java/io/deephaven/client/impl/FutureHelper.java
```diff
@@ -0,0 +1,79 @@
...
+            return getCompleted(future, true);
+        } catch (TimeoutException e) {
+            if (future.cancel(true)) {
+                throw e;
+            }
+            return getCompleted(future, false);
+        }
+    }
+
+    static <T> void cancelOrConsume(Iterable<? extends Future<T>> futures, BiConsumer<T, ExecutionException> consumer,
```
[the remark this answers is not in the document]
devinrsmith: FutureConsumer works nicely.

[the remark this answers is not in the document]
devinrsmith: I added some javadoc to cancelOrConsume b/c it's easy, but I don't know how to easily document getOrCancel without getting into the weeds of the exact implementation.

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3399
Reference: https://github.com/deephaven/deephaven.io/issues/3400

PR #4757, opened 2023-11-01
Invoke typedoc to produce browsable JS API documentation

niloc132:
Replaces the annotation processor previously used with a doclet implementation. This still emits a `.d.ts` file using Java mirror/element types for accurate type information, but now is able to traverse Javadoc trees as well, and produces corresponding typedoc tags or markdown. 

Additionally, this branch produces a tarball suitable for deployment as an npm module, containing only those generated types. However, this output is still untested.

CI is updated to deploy the results of this to https://deephaven.io/core/client-api/javascript/.

This branch is a draft, but the results of this can currently be seen at https://colinalworth.com/documentation/.


niloc132:
There are also a few javadoc changes to both test out features we expect to work now in typedoc, and also to work around deliberate shortcomings in typedoc (unclosed tags in a member's documentation can interrupt navigational aspects of the page). I'll cherry-pick and land those as their own branches, but for the time being they are included here to ensure that the rest of the branch is working correctly.

.github/workflows/docs-ci.yml
```diff
@@ -52,6 +52,12 @@ jobs:
         with:
           name: javadocs
           path: 'combined-javadoc/build/docs/javadoc/'
+      - name: Upload JavaScript/TypeScript docs
+        if: ${{ github.ref == 'refs/heads/main' }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: javadocs
```
devinrsmith, line 59: re-using existing name.

.github/workflows/docs-ci.yml
```diff
@@ -39,11 +39,11 @@ jobs:
           .github/scripts/gradle-properties.sh >> gradle.properties
           cat gradle.properties
 
-      - name: All Javadoc
+      - name: All Javadoc, ts/js docs
         uses: burrunan/gradle-cache-action@v1
         with:
           job-id: allJavadoc
-          arguments: --scan combined-javadoc:allJavadoc
+          arguments: --scan combined-javadoc:allJavadoc :web-client-api:types:typedoc
```
devinrsmith, line 46: So far, we've separated the different documentation parts into their own jobs. Is there a reason we don't want to do that w/ ts/js?

.github/workflows/docs-ci.yml
```diff
@@ -64,6 +70,17 @@ jobs:
           remote_port: ${{ secrets.DOCS_PORT }}
           remote_user: ${{ secrets.DOCS_USER }}
           remote_key: ${{ secrets.DEEPHAVEN_CORE_SSH_KEY }}
+      - name: Deploy JavaScript/TypeScript docs
+        if: ${{ github.ref == 'refs/heads/main' }}
+        uses: burnett01/rsync-deployments@5.2
+        with:
+          switches: -avzr --delete
+          path: web/client-api/docs/build/documentation/
+          remote_path: deephaven-core/client-api/javascript/
```
devinrsmith, line 79: Not sure if remote path needs to exist before this job is run, just FYI. Obv sync expectations w/ deephaven.io.

niloc132: I'm not following - how could this be omitted, where would the CI job rsync the data to?

.github/workflows/docs-ci.yml
devinrsmith: Historically, it was just a decision I made to parallelize the docs building process for what I felt where distinct builds. I think there is some value in seeing them fail independently of one another. I think it makes sense to continue the distinct jobs pattern _or_ combine them all into a single job.

.github/workflows/docs-ci.yml
devinrsmith: Sorry - didn't mean to imply you don't need to set `remote_path`. I'm not sure if the _physical_ "remote path" needs to exist or not before this job will succeed. Just commenting on the fact that there is some coordination w/ deephaven.io that will be necessary.

niloc132: Ok, will verify with docs team.

web/client-api/types/src/main/docker/theme.css
```diff
@@ -0,0 +1,18 @@
+:root {
+    --dark-color-background: #040427;
+    --dark-color-background-secondary: #0d1b37;
+    --dark-color-background-warning: #fffbef;
```
mofojed, line 4: Where are these colours coming from? This one in particular I can't see referenced anywhere else in the code base.
Pretty sure we should match the colours from other docs (defined in `root.scss`). @dsmmcken ?

dsmmcken: deephaven.io colors.

.github/workflows/docs-ci.yml
niloc132: Ok, its a separate job now.

mofojed (approved):

.github/workflows/docs-ci.yml
dsmmcken: Running it creates the folder, nothing needs to be done. You should create a ticket to add a link in the sidebar back to this though once it merges though

niloc132: I will do that in a few minutes after seeing if I broke main...

PR #4759, opened 2023-11-02
Support calling numpy ufuncs assigned to top-level names in formulas

### LINKED ISSUE #4737: Calling imported signature-less numpy ufuncs would raise a formula compilation exception
```
import numpy as np
from numba import vectorize, float64
from numpy import sin as npsin
from time import time

@vectorize([float64(float64)])
def npsin2(x):
    return np.sin(x)

n_tries = 1
source = empty_table(100_000).update("X = 0.1 * i")

start = time()
for idx in range(n_tries):
    result_builtin = source.update("Y = sin(X)")
end = time()
elapsed = (end - start) / n_tries
print(elapsed)

start = time()
for idx in range(n_tries):
    result_builtin = source.update("Y = npsin(X)")
end = time()
elapsed = (end - start) / n_tries
print(elapsed)
```
This script fails with the following exception.  Function `npsin2` is the workaround.
```
Full expression           : npsin(X)
Expression having trouble : 
Exception type            : java.lang.RuntimeException
Exception message         : Error in Python interpreter:
Type: <class 'ValueError'>
Value: callable <ufunc 'sin'> is not supported by signature
Line: 2558
Namespace: _signature_from_callable
File: /usr/lib/python3.10/inspect.py
  File "/opt/deephaven/venv/lib/python3.10/site-packages/deephaven/table.py", line 390, in _encode_signature
  File "/usr/lib/python3.10/inspect.py", line 3247, in signature
  File "/usr/lib/python3.10/inspect.py", line 2995, in from_callable
  File "/usr/lib/python3.10/inspect.py", line 2558, in _signature_from_callable
```

This looks to be a regression introduced in the Py UDF return-type inference change.

jmao-denver:
Fixes #4737

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -390,6 +390,9 @@ def _encode_signature(fn: Callable) -> str:
     try:
         sig = inspect.signature(fn)
     except:
+        # in case inspect.signature() fails, we'll just use the default 'O' - object type.
+        # numpy ufuncs actually have signature encoded in their 'types' attribute, we want to better support
+        # them in the future (https://github.com/deephaven/deephaven-core/issues/4762)
         if type(fn) == np.ufunc:
             return "O"*fn.nin + "->" + "O"
         return "->O"
@@ -447,7 +450,11 @@ def _py_udf(fn: Callable):
         try:
             return_annotation = _parse_annotation(inspect.signature(fn).return_annotation)
         except ValueError:
-            ...
+            # the function has no return annotation, and since we can't know what the exact type is, the return type
+            # defaults to the generic object type therefore it is not an array of a specific type,
+            # but see (https://github.com/deephaven/deephaven-core/issues/4762) for future imporvement to better support
+            # numpy ufuncs.
+            pass
         else:
             component_type = _component_np_dtype_char(return_annotation)
             if component_type:
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -387,7 +387,12 @@ def _encode_signature(fn: Callable) -> str:
 
     If a parameter or the return of the function is not annotated, the default 'O' - object type, will be used.
     """
-    sig = inspect.signature(fn)
+    try:
+        sig = inspect.signature(fn)
+    except:
+        if type(fn) == np.ufunc:
+            return "O"*fn.nin + "->" + "O"
+        return "->O"
 
     np_type_codes = []
     for n, p in sig.parameters.items():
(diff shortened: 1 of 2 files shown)


py/server/deephaven/table.py
```diff
@@ -387,7 +387,12 @@ def _encode_signature(fn: Callable) -> str:
 
     If a parameter or the return of the function is not annotated, the default 'O' - object type, will be used.
     """
-    sig = inspect.signature(fn)
+    try:
+        sig = inspect.signature(fn)
+    except:
+        if type(fn) == np.ufunc:
```
chipkent, line 393: I would comment this block

py/server/deephaven/table.py
```diff
@@ -439,12 +444,16 @@ def _py_udf(fn: Callable):
...
-        return_annotation = _parse_annotation(inspect.signature(fn).return_annotation)
-        component_type = _component_np_dtype_char(return_annotation)
-        if component_type:
-            ret_dtype = dtypes.from_np_dtype(np.dtype(component_type))
-            if ret_dtype in _BUILDABLE_ARRAY_DTYPE_MAP:
-                return_array = True
+        try:
+            return_annotation = _parse_annotation(inspect.signature(fn).return_annotation)
+        except ValueError:
+            ...
```
chipkent, line 450: What happens in this case?  I would comment this block.

py/server/tests/test_pyfunc_return_java_values.py
```diff
@@ -271,6 +271,27 @@ def f1(col) -> Optional[List[int]]:
         self.assertEqual(t.columns[0].data_type, dtypes.long_array)
         self.assertEqual(t.to_string().count("null"), 5)
 
+    def test_np_ufunc(self):
+        # no vectorization and no type inference
+        npsin = np.sin
+        t = empty_table(10).update(["X1 = npsin(i)"])
+        self.assertEqual(t.columns[0].data_type, dtypes.PyObject)
+        t2 = t.update("X2 = X1.getDoubleValue()")
```
chipkent, line 279: Does a double cast work?

py/server/deephaven/table.py
```diff
@@ -387,7 +387,12 @@ def _encode_signature(fn: Callable) -> str:
 
     If a parameter or the return of the function is not annotated, the default 'O' - object type, will be used.
     """
-    sig = inspect.signature(fn)
+    try:
+        sig = inspect.signature(fn)
+    except:
+        if type(fn) == np.ufunc:
+            return "O"*fn.nin + "->" + "O"
+        return "->O"
```
chipkent, line 395: Are the output types just PyObject?  If so, that's not ideal.  Can we do better?

py/server/tests/test_pyfunc_return_java_values.py
jmao-denver: No, the column is really of PyObject type.

py/server/deephaven/table.py
jmao-denver: Yes, but that's left for #4762.

chipkent (approved):

PR #4760, opened 2023-11-02
QueryPerformanceRecorder: Group Batched Operations as a Single Query

### LINKED ISSUE #4696: Performance logging from Java client splits up queries by operation
If you run a query from a client, such as the following, the individual table operations get broken up into separate "queries" in the performance logging (with distinct EvaluationNumbers and separate lines in the query performance log). This makes it impossible to measure the overall performance of a query.

Everything submitted as part of the same query should have the same EvaluationNumber (and just different OperatinNumbers).

```
query = session.scopeTable("LatestSnapshot").where("Key != null").view("ASIACalcSnapTime").maxBy()

result = session.snapshotOf(query)

session.closeTable(result)
```

image

The lines with EvaluationNumbers 256-258 should really all be one evaluation. (Possibly the flatten() too, if it's related to the `snapshotOf()`.)

### LINKED ISSUE #4765: Performance Instrumentation Should Include Ticket Resolution
Currently, both one-shot-operations and batch-operations do not performance instrument ticket resolution. This is important for DHE, and possibly for extreme-super-users with custom resolvers, that perform non-trivial work during certain ticket resolution.

nbauernfeind:
Fixes #4696.
Fixes #4765.

Change Highlights:
- Introduce a parent-child relationship for `EvaluationNumber` on the QueryPerformanceLog enabling a batch query to kick off multiple sub queries.
- Introduce a parent-child relationship for `OperationNumber` on the QueryOperationPerfomanceLog enabling easier interaction for users.
- Convert all `EvaluationNumber` usages (and blink-table columns) to `Long`.
- Convert `PerformanceEntry#id` to a `long` (and propagates to blink-table columns).
- I've removed the evaluation number / operation number from the Performance Log Logger as these are accessible from the `QueryPerformanceNugget` directly.

Nightlies: https://github.com/nbauernfeind/deephaven-core/actions/runs/6900780591/

Here is a script that can be used to create the hierarchical table view -- perhaps it should be available to all via an `EngineMetrics` accessor method.

Groovy:
```groovy
import io.deephaven.engine.table.impl.util.PerformanceQueries
qptt = PerformanceQueries.queryPerformanceAsTreeTable()
qoptt = PerformanceQueries.queryOperationPerformanceAsTreeTable()
```

Python:
```python
from deephaven import perfmon
qptt = perfmon.query_performance_tree_table()
qoptt = perfmon.query_operation_performance_tree_table()
```

And for fun here's a lovely screen shot using the java-client TableManagerExample which sends several batches:
Screenshot 2023-11-09 at 1 04 34 PM


engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceRecorder.java
```diff
@@ -714,4 +429,63 @@ public static boolean setCallsite() {
...
+
+        if (needClear) {
+            clearCallsite();
+        }
+    }
+
+    /**
+     * Dummy recorder for use when no recorder is installed.
+     */
+    private static class DummyQueryPerformanceRecorder extends QueryPerformanceRecorder {
```
rcaudy, line 452: We might be able to streamline this into an impl that always throws UOE for most ops.

rcaudy (commented):
We talked about:
1. Separating new from start, for close support, to stop exposing reset
2. “ending” outer nugget when creating inner, accumulating and restarting when ending inner
3. Maybe combining end and suspend; and resume and start? Separating is nice for state management, but otherwise unnecessary; you can distinguish start from resume by value of start time.
4. Probably move helper/statics out of QueryPerformanceRecorder to a package-private helper, to hide them from “users”
5. That lets QPR be an interface
6. Maybe make end/suspend do the reset internally?

rcaudy (commented):
.

py/server/deephaven/perfmon.py
chipkent, line 1: Ideally the new functionality should have a unit test to make sure it actually functions.

py/server/deephaven/perfmon.py
```diff
@@ -95,6 +95,38 @@ def query_performance_log() -> Table:
     except Exception as e:
         raise DHError(e, "failed to obtain the query performance log table.") from e
 
+def query_operation_performance_log_as_tree_table() -> TreeTable:
```
chipkent, line 98:
```suggestion
def query_operation_performance_tree_table() -> TreeTable:
```
Name is already way way long.

py/server/deephaven/perfmon.py
```diff
@@ -95,6 +95,38 @@ def query_performance_log() -> Table:
...
+        DHError
+    """
+    try:
+        return TreeTable(j_tree_table=_JPerformanceQueries.queryOperationPerformanceAsTreeTable(),
+                         id_col = "EvalKey", parent_col = "ParentEvalKey")
+    except Exception as e:
+        raise DHError(e, "failed to obtain the query operation performance log as tree table.") from e
+
+
+def query_performance_log_as_tree_table() -> TreeTable:
```
chipkent, line 114:
```suggestion
def query_performance_tree_table() -> TreeTable:
```

py/server/deephaven/perfmon.py
```diff
@@ -95,6 +95,38 @@ def query_performance_log() -> Table:
...
+    """ Returns a tree table with Deephaven query performance data. Performance data for individual sub-operations as
+    a tree table is available from calling `query_operation_performance_log_as_tree_table`.
```
chipkent, line 116:
```suggestion
    a tree table is available from calling `query_operation_performance_tree_table`.
```
Should the other method include a reference to this method as well?

nbauernfeind: Technically the other tree includes the QPL level entries. They are parents of the operations; so I didn't think it was necessary/useful.

chipkent (approved):
Python LGTM.  Did not review the rest.  Let me know if I need to.

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceNugget.java
```diff
@@ -221,43 +243,44 @@ private QueryPerformanceNugget() {
...
 
-        startClockEpochNanos = NULL_LONG;
-
         state = null; // This turns close into a no-op.
         shouldLogThisAndStackParents = false;
     }
 
-    public void done() {
-        done(QueryPerformanceRecorder.getInstance());
+    public void markStartTime() {
```
rcaudy, line 261: `start()`? Collapse with base entry start? Document as doing a base entry start? Should probably be synchronized.

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceNugget.java
```diff
@@ -279,14 +301,18 @@ private boolean close(final QueryState closingState, final QueryPerformanceRecor
...
+
             onBaseEntryEnd();
             endClockEpochNanos = DateTimeUtils.millisToNanos(System.currentTimeMillis());
 
             final RuntimeMemory runtimeMemory = RuntimeMemory.getInstance();
             runtimeMemory.read(endMemorySample);
 
             state = closingState;
-            return recorderToNotify.releaseNugget(this);
+            return onCloseCallback.test(this);
```
rcaudy, line 339: This is weird. Couldn't we just have the recorder check *after*? Or inside the callback? That way done and abort stop returning a boolean, too.

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceRecorder.java
```diff
@@ -100,128 +51,184 @@ public static void resetInstance() {
...
      *
      * @return Either a "catch-all" nugget, or the top of the user nugget stack.
      */
-    public abstract QueryPerformanceNugget getEnclosingNugget();
+    QueryPerformanceNugget getEnclosingNugget();
+
+
+    interface EntrySetter {
+        void set(long evaluationNumber, int operationNumber, boolean uninstrumented);
+    }
```
rcaudy: QueryDataConsumer?
set -> accept?
setQueryData -> supplyQueryData?

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceRecorder.java
```diff
@@ -100,128 +51,184 @@ public static void resetInstance() {
...
+            @NotNull final QueryPerformanceNugget.Factory nuggetFactory) {
+        return new QueryPerformanceRecorderImpl(description, parent, nuggetFactory);
     }
 
     /**
-     * Get the total bytes of pool-allocated memory attributed to this thread via
-     * {@link #recordPoolAllocation(Supplier)}.
+     * Starts a query.
+     * <p>
+     * It is an error to start a query more than once or while another query is running on this thread.
```
rcaudy, line 171: A query is running if it has been {@link #startQuery() started} or {@link #resumed() resumed} without a subsequent {@link #endQuery() end}, {@link #suspendQuery() suspend}, or {@link #abortQuery() abort}).

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceRecorder.java
```diff
@@ -100,128 +51,184 @@ public static void resetInstance() {
...
+
+    /**
+     * This getter should be called by exclusive owners of the recorder, and never concurrently with mutators.
+     *
+     * @return A list of loggable operation performance data.
+     */
+    List<QueryPerformanceNugget> getOperationLevelPerformanceData();
+
+    /**
+     * Accumulate the values from another recorder into this one. The provided recorder will not be mutated.
```
rcaudy, line 217: Be explicit about what's being accumulated. Aggregate performance information.

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceRecorderImpl.java
```diff
@@ -234,17 +226,20 @@ synchronized boolean releaseNugget(@NotNull final QueryPerformanceNugget nugget)
...
-            if (!userNuggetStack.isEmpty()) {
-                userNuggetStack.getLast().setShouldLogThisAndStackParents();
+            if (removed.shouldLogThisAndStackParents()) {
+                parent.setShouldLogThisAndStackParents();
             }
-        } else {
+
+            // resume the parent
+            parent.onBaseEntryStart();
+        }
```
rcaudy, line 254: You were thinking about accumulating depth==0 nuggets into the enclosing query.

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceRecorderImpl.java
```diff
@@ -204,22 +189,29 @@ public synchronized QueryPerformanceNugget getNugget(@NotNull final String name,
...
-     * <b>Note:</b> Do not call this directly - it's for nugget use only. Call {@link QueryPerformanceNugget#done()} or
-     * {@link QueryPerformanceNugget#close()} instead.
+     * This is our onCloseCallback from the nugget.
      *
      * @param nugget the nugget to be released
      * @return If the nugget passes criteria for logging.
      */
-    synchronized boolean releaseNugget(@NotNull final QueryPerformanceNugget nugget) {
+    private synchronized boolean releaseNugget(@NotNull final QueryPerformanceNugget nugget) {
         boolean shouldLog = nugget.shouldLogNugget(nugget == catchAllNugget);
```
rcaudy, line 229: I think this should internally update `shouldLogThisAndStackParents`.

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceRecorderImpl.java
```diff
@@ -234,17 +226,20 @@ synchronized boolean releaseNugget(@NotNull final QueryPerformanceNugget nugget)
...
 
-        if (shouldLog) {
-            // It is entirely possible, with parallelization, that this nugget should be logged while the outer nugget
-            // has a wall clock time less than the threshold for logging. If we ever want to log this nugget, we must
-            // log
-            // all of its parents as well regardless of the shouldLogNugget call result.
-            if (!userNuggetStack.isEmpty()) {
-                userNuggetStack.getLast().setShouldLogThisAndStackParents();
+            if (removed.shouldLogThisAndStackParents()) {
+                parent.setShouldLogThisAndStackParents();
```
rcaudy, line 249: Rename to should log?

rcaudy (commented):
Comments on two commits

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceNugget.java
```diff
@@ -4,121 +4,234 @@
...
-
-        startClockTime = System.currentTimeMillis();
-        startTimeNanos = System.nanoTime();
-
-        startCpuNanos = ThreadProfiler.DEFAULT.getCurrentThreadCpuTime();
-        startUserCpuNanos = ThreadProfiler.DEFAULT.getCurrentThreadUserTime();
-
-        state = QueryState.RUNNING;
-        shouldLogMeAndStackParents = false;
+        state = QueryState.NOT_STARTED;
```
rcaudy, line 233: I think we should do a quick audit of instance members and constructors.
1. I don't think we should mix inline initialization with constructor initialization.
2. We should make sure the order of members matches the order of initializations in the constructors, so it's easy to verify that we didn't make any mistakes.
3. I prefer not to explicitly initialize to default values, maybe. Or, do so consistently.

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceNugget.java
```diff
@@ -127,57 +240,71 @@ public class QueryPerformanceNugget implements Serializable, AutoCloseable {
...
 
         authContext = null;
         callerLine = null;
 
-        startAllocatedBytes = NULL_LONG;
-        startPoolAllocatedBytes = NULL_LONG;
-
-        startClockTime = NULL_LONG;
-        startTimeNanos = NULL_LONG;
+        state = null; // This turns close into a no-op.
```
rcaudy, line 257: Comment belongs on a JavaDoc on state.

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceNugget.java
```diff
@@ -4,121 +4,234 @@
...
+            final int operationNumber,
+            final int parentOperationNumber,
+            final int depth,
+            @NotNull final String description,
+            @Nullable final String sessionId,
+            final boolean isUser,
+            final long inputSize,
+            @NotNull final Predicate<QueryPerformanceNugget> onCloseCallback) {
         startMemorySample = new RuntimeMemory.Sample();
         endMemorySample = new RuntimeMemory.Sample();
```
rcaudy, line 210: Why would we want to do the samples here, rather than in start?

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceNugget.java
```diff
@@ -127,57 +240,71 @@ public class QueryPerformanceNugget implements Serializable, AutoCloseable {
...
 
     /**
      * Mark this nugget {@link QueryState#FINISHED} and notify the recorder.
      *
-     * @param recorder The recorder to notify
      * @return if the nugget passes logging thresholds.
      */
-    public boolean done(final QueryPerformanceRecorder recorder) {
-        return close(QueryState.FINISHED, recorder);
+    public boolean done() {
```
rcaudy, line 292: Can we safely remove this? Just document on close()?

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceNugget.java
```diff
@@ -201,42 +325,51 @@ private boolean close(final QueryState closingState, final QueryPerformanceRecor
                 return false;
             }
 
-            diffUserCpuNanos = minus(currentThreadUserTime, startUserCpuNanos);
-            diffCpuNanos = minus(currentThreadCpuTime, startCpuNanos);
+            if (startClockEpochNanos == NULL_LONG) {
```
rcaudy, line 328: We have state for this...

engine/table/src/main/java/io/deephaven/engine/table/impl/perf/QueryPerformanceNugget.java
```diff
@@ -201,42 +325,51 @@ private boolean close(final QueryState closingState, final QueryPerformanceRecor
...
+                + ":" + (isQueryLevel() ? "query_level" : operationNumber)
                 + ":" + description
                 + ":" + callerLine;
     }
 
-    public int getEvaluationNumber() {
+    @Override
+    public LogOutput append(@NotNull final LogOutput logOutput) {
+        // override BasePerformanceEntry's impl and match toString()
+        return logOutput.append(toString());
```
rcaudy, line 354: Backwards. Implement toString using append.

engine/table/src/main/java/io/deephaven/engine/table/impl/util/QueryPerformanceStreamPublisher.java
```diff
@@ -64,69 +63,68 @@ public void register(@NotNull StreamConsumer consumer) {
...
 
         // ColumnDefinition.ofLong("EvaluationNumber")
-        chunks[1].asWritableLongChunk().add(evaluationNumber);
+        chunks[0].asWritableLongChunk().add(nugget.getEvaluationNumber());
+
+        // ColumnDefinition.ofLong("ParentEvaluationNumber")
+        chunks[1].asWritableLongChunk().add(nugget.getParentEvaluationNumber());
+
+        // ColumnDefinition.ofString("Description")
+        chunks[2].<String>asWritableObjectChunk().add(nugget.getName());
```
rcaudy, line 76: name -> description on nugget?

extensions/performance/src/main/java/io/deephaven/engine/table/impl/util/PerformanceQueriesGeneral.java
```diff
@@ -35,20 +37,21 @@ public static Table queryPerformance(Table queryPerformanceLog, final long evalu
...
-                        "QueryMemUsed", "QueryMemFree", "QueryMemUsedPct",
-                        "EndTime", "TimeSecs", "NetMemoryChange");
+                        // Memory in use by the query. (Only includes active heap memory.)
+                        "QueryMemUsed = TotalMemory - FreeMemory",
+                        // Memory usage as a percenage of max heap size (-Xmx)
+                        "QueryMemUsedPct = QueryMemUsed / WorkerHeapSize",
+                        // Remaining memory until the query runs into the max heap size
+                        "QueryMemFree = WorkerHeapSize - QueryMemUsed");
+
+        queryPerformanceLog = maybeMoveColumnsUp(queryPerformanceLog,
```
rcaudy, line 50: Remember `maybeMoveColumnsUp` should have a list of allowed missing.

extensions/performance/src/main/java/io/deephaven/engine/table/impl/util/PerformanceQueriesGeneral.java
```diff
@@ -310,6 +313,33 @@ public static Map<String, Object> serverStateWithPlots(final Table pml) {
...
 
+    public static TreeTable queryPerformanceAsTreeTable(@NotNull final Table qpl) {
+        return qpl.tree("EvaluationNumber", "ParentEvaluationNumber");
+    }
+
+    public static TreeTable queryOperationPerformanceAsTreeTable(
+            @NotNull final Table qpl, @NotNull final Table qopl) {
+        Table mergeWithAggKeys = TableTools.merge(
+                qpl.updateView(
+                        "EvalKey = `` + EvaluationNumber",
```
rcaudy, line 324: Use `Long.toString()` or `Integer.toString()` for these cases.

extensions/performance/src/main/java/io/deephaven/engine/table/impl/util/PerformanceQueriesGeneral.java
```diff
@@ -310,6 +313,33 @@ public static Map<String, Object> serverStateWithPlots(final Table pml) {
...
+                        "ParentEvalKey = ParentEvaluationNumber == null ? null : (`` + ParentEvaluationNumber)",
+                        "OperationNumber = NULL_INT",
+                        "ParentOperationNumber = NULL_INT",
+                        "Depth = NULL_INT",
```
rcaudy, line 328: If you want to assign these to OperationNumber, fine, but do comment it

server/src/main/java/io/deephaven/server/arrow/ArrowFlightUtil.java
```diff
@@ -71,37 +73,45 @@ public static void DoGetCustom(
...
-                                    table.getDefinition(), table.getAttributes())));
-
-                    // shared code between `DoGet` and `BarrageSnapshotRequest`
-                    BarrageUtil.createAndSendSnapshot(streamGeneratorFactory, table, null, null, false,
-                            DEFAULT_SNAPSHOT_DESER_OPTIONS, listener, metrics);
-
-                    listener.onCompleted();
-                });
+        final String description =
+                "FlightService#DoGet(request=" + ticketRouter.getLogNameFor(request, "request") + ")";
```
rcaudy, line 77: Do we want to standardize the log names?

server/src/main/java/io/deephaven/server/partitionedtable/PartitionedTableServiceGrpcImpl.java
```diff
@@ -55,20 +58,28 @@ public void partitionBy(
...
+        final String description = "PartitionedTableServiceGrpcImpl#partitionBy(source="
+                + ticketRouter.getLogNameFor(request.getTableId(), "source") + ")";
+        final QueryPerformanceRecorder queryPerformanceRecorder = QueryPerformanceRecorder.newQuery(
+                description, session.getSessionId(), QueryPerformanceNugget.DEFAULT_FACTORY);
+
+        try (final SafeCloseable ignored = queryPerformanceRecorder.startQuery()) {
+            final SessionState.ExportObject<Table> targetTable =
+                    ticketRouter.resolve(session, request.getTableId(), "partition.tableId");
+
+            session.newExport(request.getResultId(), "partition.resultId")
```
rcaudy, line 70: You added a prefix to these strings. It's cool, but inconsistent. Maybe remove?

server/src/main/java/io/deephaven/server/table/ops/TableServiceGrpcImpl.java
```diff
@@ -491,65 +522,87 @@ public void batch(
...
+                    final QueryProcessingResults results = new QueryProcessingResults(queryPerformanceRecorder);
+                    final StatusRuntimeException failure = firstFailure.get();
+                    if (failure != null) {
+                        results.setException(failure.getMessage());
+                        safelyError(responseObserver, failure);
+                    } else {
+                        safelyComplete(responseObserver);
+                    }
+                    if (queryPerformanceRecorder.endQuery()) {
+                        EngineMetrics.getInstance().logQueryProcessingResults(results);
```
rcaudy, line 563: Maybe we should delete results and pass two things?

rcaudy (commented):
.

rcaudy (approved):

PR #4764, opened 2023-11-02
Unit-test thread cleanup

devinrsmith:
During testing of #4756, I discovered that `io.deephaven.server.util.Scheduler` and `io.deephaven.util.reference.CleanupReferenceProcessor` threads were leaking. CI testing was throwing "can't create more threads" errors due to the addition of a large number of new java client async tests which pushed the leaked thread counts over some JVM limit.

As concrete evidence of the leaks, I ran

```
./gradlew java-client-session-dagger:test -PforceTest=true --max-workers 1
```

```
jps
jstack <pid-for-GradleWorkerMain> | grep os_prio | awk -F " " '{print $1}' | sort | uniq -c | sort -nr
```

and got

```
    432 "DeephavenApiServer-Scheduler-Concurrent-4"
    432 "DeephavenApiServer-Scheduler-Concurrent-3"
    432 "DeephavenApiServer-Scheduler-Concurrent-2"
    432 "DeephavenApiServer-Scheduler-Concurrent-1"
     43 "CleanupReferenceProcessor-liveness-drainingThread"
     18 "GC
     10 "G1
      4 "C1
      3 "C2
      3 "/127.0.0.1:42968
      2 "VM
      1 "Test
      1 "Sweeper
      1 "StrDedup"
      1 "Signal
      1 "Service
      1 "Reference
      1 "process
      1 "PeriodicUpdateGraph-updateExecutor-1"
      1 "grpc-timer-0"
      1 "grpc-shared-destroyer-0"
      1 "grpc-default-executor-6"
      1 "grpc-default-executor-5"
      1 "grpc-default-executor-4"
      1 "grpc-default-executor-3"
      1 "grpc-default-executor-2"
      1 "grpc-default-executor-1"
      1 "grpc-default-executor-0"
      1 "Finalizer"
      1 "Common-Cleaner"
      1 "Attach
```

This PR adds an unit-test-only shutdown procedure for the Scheduler; if we wanted to actually change the Scheduler interface to support this, it would require more work and testing. This also ensures that CleanupReferenceProcessor are promptly cleaned up (instead of eventually).


Util/src/main/java/io/deephaven/util/reference/CleanupReferenceProcessor.java
```diff
@@ -37,6 +37,7 @@ public interface ExceptionHandler {
      * The reference queue from the most recent initialization.
      */
     private volatile ReferenceQueue<?> referenceQueue;
+    private Thread cleanerThread;
```
rcaudy, line 40:
```suggestion

    /**
     * The cleaner thread from the most initialization, guarded by the lock on {@code this}.
     */
    private Thread cleanerThread;
```

server/src/test/java/io/deephaven/server/util/SchedulerShutdown.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public void run() throws InterruptedException {
+        scheduler.concurrentDelegate().shutdownNow();
+        scheduler.serialDelegate().shutdownNow();
+        if (!scheduler.concurrentDelegate().awaitTermination(5, TimeUnit.SECONDS)) {
+            throw new RuntimeException("concurrentDelegate not shutdown within 5 seconds");
+        }
+        if (!scheduler.serialDelegate().awaitTermination(5, TimeUnit.SECONDS)) {
+            throw new RuntimeException("serialDelegate not shutdown within 5 seconds");
+        }
```
rcaudy, line 27: We should move this logic into `DelegatingImpl`. I understand why you don't want to do expose `shutdown()` as a public method on `Scheduler`.

nbauernfeind (approved):

rcaudy (approved):

PR #4771, opened 2023-11-03
Add topic, partition, and timestamp column kafka publishing support

### LINKED ISSUE #4767: Support kafka publishing into specific partition on row-by-row basis
There is not currently an easy way for users to publish a Deephaven table row into a specific Kafka topic partition. Right now, users can set a kafka-specific strategy to choose the correct partition for a record [partitioner.class](https://docs.confluent.io/platform/current/installation/configuration/producer-configs.html#partitioner-class), but that's a layer of indirection the user shouldn't typically need to interact with. Alternatively, the user can create N kafka publishers (still setting `partition.class`) and partition their tables accordingly, but that is also less-than-ideal.

It should be relatively easy to add a feature where the user specifies a source column as a "partition" column, and for the record to be published with that partition as appropriate.

We could also consider adding support for a row-by-row topic instead of requiring a fixed topic up front as this could be similarly useful depending on kafka architectures. (Trying to create a different _payload_ depending on topic would be out-of-scope for this issue.)

devinrsmith:
This adds optional configuration for publishing a kafka record with the topic, partition, and/or timestamp as specified by their respective columns.

Additionally, this also adds the ability to specify a default partition.

Fixes #4767


py/server/deephaven/stream/kafka/producer.py
```diff
@@ -61,6 +66,13 @@ def produce(
             aggregation on table grouped by the input columns of key_spec and publish to Kafka from the result.
         publish_initial (bool): whether the initial data in table should be published. When False, table.is_refreshing
             must be True. By default, is True.
+        partition (Optional[int]): the default partition, None by default.
+        topic_column (Optional[str]): the topic column, None by default. When set, uses the the given string column from
+            table as the first source for setting the kafka record topic.
+        partition_column (Optional[str]): the partition column, None by default. When set, uses the the given int column
+            from table as the first source for setting the kafka record partition.
```
rcaudy, line 73:
```suggestion
            from table as the first source for setting the kafka record partition, falling back to partition on null, and to Kafka default logic if partition is null.
```

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -61,6 +66,13 @@ def produce(
...
+        timestamp_column (Optional[str]): the timestamp column, None by default. When set, uses the the given timestamp
+            column from table as the first source for setting the kafka record timestamp.
```
rcaudy, line 75: And what's the second source?

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -61,6 +66,13 @@ def produce(
             aggregation on table grouped by the input columns of key_spec and publish to Kafka from the result.
         publish_initial (bool): whether the initial data in table should be published. When False, table.is_refreshing
             must be True. By default, is True.
+        partition (Optional[int]): the default partition, None by default.
+        topic_column (Optional[str]): the topic column, None by default. When set, uses the the given string column from
+            table as the first source for setting the kafka record topic.
```
rcaudy, line 71: Fallback?

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -77,7 +89,7 @@ def produce(
             )
         if not publish_initial and not table.is_refreshing:
             raise ValueError("publish_initial == False and table.is_refreshing == False")
-        options = (
+        options_builder = (
             _JKafkaPublishOptions.builder()
             .table(table.j_table)
             .topic(topic)
```
rcaudy, line 95: Shouldn't this be in an `if`, since it's optional now?

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaPublishOptions.java
```diff
@@ -34,12 +39,20 @@ public static Builder builder() {
     public abstract Table table();
 
     /**
-     * The kafka topic to publish to.
+     * The default kafka topic to publish to.
      *
-     * @return the kafka topic
+     * @return the default kafka topic
      */
+    @Nullable
```
rcaudy, line 46: Maybe you want to add a method that converts this with an `Optional.ofNullable()`?

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaPublishOptions.java
```diff
@@ -93,6 +106,30 @@ public boolean publishInitial() {
         return true;
     }
 
+    /**
+     * The topic column. When set, uses the the given {@link CharSequence}-compatible column from {@link #table()} as
```
rcaudy, line 110: Need to propagate documentation review points from Python to this file.

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaPublishOptions.java
```diff
@@ -114,12 +151,42 @@ final void checkLastBy() {
...
+    final void checkTopic() {
+        if (topic() == null && topicColumn().isEmpty()) {
+            throw new IllegalArgumentException("Must set topic or topicColumn (or both)");
+        }
+    }
+
+    @Check
+    final void checkTopicColumn() {
+        if (topicColumn().isPresent()) {
+            table().getColumnSource(topicColumn().get().name(), CharSequence.class);
```
rcaudy, line 164: Better to validate based on the definition.

extensions/kafka/src/main/java/io/deephaven/kafka/publish/PublishToKafka.java
```diff
@@ -89,26 +124,46 @@ public class PublishToKafka<K, V> extends LivenessArtifact {
...
             final Serializer<K> kafkaKeySerializer,
             final KeyOrValueSerializer<K> keyChunkSerializer,
             final String[] valueColumns,
             final Serializer<V> kafkaValueSerializer,
             final KeyOrValueSerializer<V> valueChunkSerializer,
+            final ColumnName topicColumn,
+            final ColumnName partitionColumn,
+            final ColumnName timestampColumn,
             final boolean publishInitial) {
         this.table = table;
```
rcaudy, line 149: Maybe make the RHS of the assignment `.coalesce()`, and then use `this.table` throughout.

extensions/kafka/src/main/java/io/deephaven/kafka/publish/PublishToKafka.java
```diff
@@ -133,6 +188,38 @@ private static ModifiedColumnSet getModifiedColumnSet(@NotNull final Table table
...
+        final CharSequence charSequence = topicChunk.get(index);
+        return charSequence == null ? defaultTopic : charSequence.toString();
+    }
+
+    private Integer partition(IntChunk<?> partitionChunk, int index) {
+        if (partitionChunk == null) {
+            return defaultPartition;
+        }
+        final int partition = partitionChunk.get(index);
+        return partition == QueryConstants.NULL_INT ? defaultPartition : (Integer) partition;
```
rcaudy, line 204:
```suggestion
        return partition == QueryConstants.NULL_INT ? defaultPartition : Integer.valueOf(partition);
```

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -36,11 +37,15 @@ def j_object(self) -> jpy.JType:
...
     key_spec: KeyValueSpec,
     value_spec: KeyValueSpec,
     last_by_key_columns: bool = False,
     publish_initial: bool = True,
+    partition: Optional[int] = None,
+    topic_column: Optional[str] = None,
+    partition_column: Optional[str] = None,
+    timestamp_column: Optional[str] = None,
 ) -> Callable[[], None]:
     """Produce to Kafka from a Deephaven table.
```
chipkent, line 50: I would go over all of the new pydocs to make sure they answer all questions a user may have.  They seem light on details for a number of cases, and I may not know enough cases to fully say where they are not comprehensive enough.

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -66,13 +66,22 @@ def produce(
...
-        partition (Optional[int]): the default partition, None by default.
+        partition (Optional[int]): the default partition, None by default. See partition_column for partition behavior.
         topic_column (Optional[str]): the topic column, None by default. When set, uses the the given string column from
-            table as the first source for setting the kafka record topic.
+            table as the first source for setting the kafka record topic. When None, or the column value is null, topic
+            will be used.
         partition_column (Optional[str]): the partition column, None by default. When set, uses the the given int column
-            from table as the first source for setting the kafka record partition.
+            from table as the first source for setting the kafka record partition. When None, or the column value is null,
+            partition will be used if present. If a valid partition number is specified that partition will be used when
```
chipkent, line 75:
```suggestion
            partition will be used if present. If a valid partition number is specified, that partition will be used when
```

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -66,13 +66,22 @@ def produce(
...
+            sending the record. If no partition is specified but a key is present a partition will be chosen using a hash
```
chipkent, line 76:
```suggestion
            sending the record. If no partition is specified but a key is present, a partition will be chosen using a hash
```

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -66,13 +66,22 @@ def produce(
...
+            of the key. If neither key nor partition is present a partition will be assigned in a round-robin fashion.
```
chipkent, line 77:
```suggestion
            of the key. If neither key nor partition is present, a partition will be assigned in a round-robin fashion.
```

chipkent (dismissed):

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -61,6 +66,22 @@ def produce(
             aggregation on table grouped by the input columns of key_spec and publish to Kafka from the result.
         publish_initial (bool): whether the initial data in table should be published. When False, table.is_refreshing
             must be True. By default, is True.
+        partition (Optional[int]): the default partition, None by default. See partition_column for partition behavior.
+        topic_column (Optional[str]): the topic column, None by default. When set, uses the the given string column from
```
jmao-denver, line 70: I think `topic_col` is more consistent with our naming convention.

engine/api/src/main/java/io/deephaven/engine/table/ColumnDefinition.java
```diff
@@ -504,6 +504,28 @@ public void describeDifferences(@NotNull List<String> differences, @NotNull fina
         }
     }
 
+    /**
+     * Checks if {@link #getDataType() dataType} can be cast to {@code destDataType}. If not, this throws a
+     * {@link ClassCastException}.
+     *
+     * @param destDataType the destination data type
+     */
+    public final void checkCastTo(Class<?> destDataType) {
```
rcaudy, line 513: Usually we just have a `cast` method that returns the type we like and (maybe) validates internally.

engine/api/src/main/java/io/deephaven/engine/table/ColumnDefinition.java
```diff
@@ -504,6 +504,28 @@ public void describeDifferences(@NotNull List<String> differences, @NotNull fina
...
+        TypeHelper.checkCastTo("[" + name + "]", dataType, destDataType);
```
rcaudy, line 514: This context being passed does not mesh well with the exception message as currently composed.

rcaudy, line 514: This context being passed does not mesh well with the exception message as currently composed.

engine/api/src/main/java/io/deephaven/engine/table/TableDefinition.java
```diff
@@ -313,6 +313,40 @@ public final void checkHasColumn(@NotNull String columnName) {
         NoSuchColumnException.throwIf(getColumnNameSet(), columnName);
     }
 
+    /**
+     * Checks if {@code columnName} exists and if supports {@link ColumnDefinition#checkCastTo(Class)} with
```
rcaudy, line 317:
```suggestion
     * Checks if {@code columnName} exists and supports {@link ColumnDefinition#checkCastTo(Class)} with
```

engine/api/src/main/java/io/deephaven/engine/table/TableDefinition.java
```diff
@@ -313,6 +313,40 @@ public final void checkHasColumn(@NotNull String columnName) {
...
+    public final void checkHasColumn(@NotNull String columnName, @NotNull Class<?> clazz) {
+        final ColumnDefinition<?> cd = getColumn(columnName);
+        if (cd == null) {
+            throw new NoSuchColumnException(getColumnNameSet(), columnName);
+        }
+        cd.checkCastTo(clazz);
+    }
+
+    /**
+     * Checks if {@code columnName} exists and if supports {@link ColumnDefinition#checkCastTo(Class, Class)} with
```
rcaudy, line 333:
```suggestion
     * Checks if {@code columnName} exists and supports {@link ColumnDefinition#checkCastTo(Class, Class)} with
```

engine/api/src/main/java/io/deephaven/engine/table/TableDefinition.java
```diff
@@ -323,6 +357,7 @@ public final void checkHasColumns(@NotNull Collection<String> columns) {
         NoSuchColumnException.throwIf(getColumnNameSet(), columns);
     }
 
+
```
rcaudy, line 360:
```suggestion
```

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaPublishOptions.java
```diff
@@ -108,25 +110,33 @@ public boolean publishInitial() {
 
     /**
      * The topic column. When set, uses the the given {@link CharSequence}-compatible column from {@link #table()} as
-     * the first source for setting the kafka record topic.
+     * the first source for setting the kafka record topic. When not present, or the column value is null,
```
rcaudy, line 113:
```suggestion
     * the first source for setting the kafka record topic. When not present, or if the column value is null,
```

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaPublishOptions.java
```diff
@@ -108,25 +110,33 @@ public boolean publishInitial() {
...
+     * {@link #topic()} will be used.
      *
      * @return the topic column name
      */
     public abstract Optional<ColumnName> topicColumn();
 
     /**
      * The partition column. When set, uses the the given {@code int} column from {@link #table()} as the first source
-     * for setting the kafka record partition.
+     * for setting the kafka record partition. When not present, or the column value is null, {@link #partition()} will
```
rcaudy, line 122:
```suggestion
     * for setting the kafka record partition. When not present, or if the column value is null, {@link #partition()} will
```

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaPublishOptions.java
```diff
@@ -108,25 +110,33 @@ public boolean publishInitial() {
...
+     * be used if present. If a valid partition number is specified, that partition will be used when sending the
+     * record. If no partition is specified but a key is present, a partition will be chosen using a hash of the key. If
```
rcaudy, line 124: OK. Maybe we should clarify that the last two bits are "Kafka's default behavior" that we don't control?

extensions/kafka/src/main/java/io/deephaven/kafka/KafkaPublishOptions.java
```diff
@@ -108,25 +110,33 @@ public boolean publishInitial() {
...
+     * neither key nor partition is present, a partition will be assigned in a round-robin fashion.
      *
      * @return the partition column name
      */
     public abstract Optional<ColumnName> partitionColumn();
 
     /**
      * The timestamp column. When set, uses the the given {@link Instant} column from {@link #table()} as the first
-     * source for setting the kafka record timestamp.
+     * source for setting the kafka record timestamp. When not present, or the column value is null, the producer will
```
rcaudy, line 133:
```suggestion
     * source for setting the kafka record timestamp. When not present, or if the column value is null, the producer will
```

py/server/deephaven/stream/kafka/producer.py
```diff
@@ -66,13 +66,22 @@ def produce(
...
             must be True. By default, is True.
-        partition (Optional[int]): the default partition, None by default.
+        partition (Optional[int]): the default partition, None by default. See partition_column for partition behavior.
         topic_column (Optional[str]): the topic column, None by default. When set, uses the the given string column from
-            table as the first source for setting the kafka record topic.
+            table as the first source for setting the kafka record topic. When None, or the column value is null, topic
+            will be used.
         partition_column (Optional[str]): the partition column, None by default. When set, uses the the given int column
-            from table as the first source for setting the kafka record partition.
+            from table as the first source for setting the kafka record partition. When None, or the column value is null,
```
rcaudy, line 74: Same comments as in Java.

engine/api/src/main/java/io/deephaven/engine/table/ColumnDefinition.java
devinrsmith: I think you took back this point.

engine/api/src/main/java/io/deephaven/engine/table/ColumnDefinition.java
devinrsmith: I think it's a little bit different w/ `ColumnDefinition`. In other cases cases where `cast` is used, it is typically done to match types b/c the consumer wants _to get data out_. In this case, `ColumnDefinition` itself doesn't hold any data, it is just parameterized to support explicit typing around `Class <TYPE> getDataType()`.

I don't see a practical use for wanting to actually cast a definition. Furthermore, it breaks the contract:

```java
ColumnDefinition<Integer> intColDef = ...;
ColumnDefinition<Number> numColDef = intColDef.cast(Number.class);

Class<Number> numberClass = numColDef.getDataType();
// This will fail
assertEquals(Number.class, numberClass);
```

We do play loose and fast with this in other places; `ColumnSource` suffers from the same `getType` problem - although again, `ColumnSource#cast` is usually used to get _data_ out (not b/c somebody wants `getType`).

In some more recent code, I've tried to "do the right thing"; `io.deephaven.functions.ToObjectFunction#cast`

rcaudy (approved):

jmao-denver (approved):
The Python changes LGTM

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3450
Reference: https://github.com/deephaven/deephaven.io/issues/3451

PR #4772, opened 2023-11-04
ExportObject: Add onSuccess Callback for use in TableService#batch

### LINKED ISSUE #4754: Batch ETCR and Release race
It's currently possible for the server to send two ETCR responses for the same export to a client; the first one claiming the ETCR is successful, and the second one claiming the ETCR was not successful. This was seen during the development of an asynchronous interface for the java client, with logs that looked like

```
[grpc-default-executor-0] INFO io.deephaven.client.impl.ExportRequest$LoggingListener - ExportedTableCreationResponse for 'export/36' was successful
[grpc-default-executor-2] INFO io.deephaven.client.impl.ExportRequest$LoggingListener - ExportedTableCreationResponse for 'export/36' was not successful: 682bd4a4-7c9f-4d7e-a6c5-eac2cb2edc57
```

Note: this is not the typical server UUID obfuscated error - it's specific to ExportObject, and doesn't have any other corresponding server logs (which may be another issue we want to look into).

The test in question creates a chain of exports executed via a Batch and then Releases each export as soon as the successful ETCR comes in.

The server side in question looks like:

```java
exportBuilder.exportBuilder.onError((result, errorContext, cause, dependentId) -> {
    ...
    final ExportedTableCreationResponse response = ExportedTableCreationResponse.newBuilder()
            .setResultId(resultId)
            .setSuccess(false)
            .setErrorInfo(errorInfo)
            .build();
    safelyOnNext(responseObserver, response);
    onOneResolved.run();
}).submit(() -> {
    final Table table = exportBuilder.doExport();
    final ExportedTableCreationResponse response =
            ExportUtil.buildTableCreationResponse(resultId, table);
    safelyOnNext(responseObserver, response);
    onOneResolved.run();
    return table;
});
```

Above there are two paths to call onNext for a given export which matches up with our bug report; looking down into the implementation, we see that submit callable `capturedExport` is executed in 

```java
private void doExport() {
    ...
        setResult(capturedExport.call());
    ...
}

private void setResult(final T result) {
    ...
    synchronized (this) {
        // client may race a cancel with setResult
        if (!isExportStateTerminal(state)) {
            this.result = result;
            if (result instanceof LivenessReferent && DynamicNode.notDynamicOrIsRefreshing(result)) {
                manage((LivenessReferent) result);
            }
            setState(ExportNotification.State.EXPORTED);
        }
    }
    ...
}
```

Note the comment "client may race a cancel with setResult" aims to solve the issue we are seeing; but there is no synchronization at this state around the callable (where the onNext is actually sent).

In the meantime a Release RPC comes in before the doWork is able to advance to the synchronized block in setResult; and the release gets into

```java
public synchronized void cancel() {
    ...
    } else if (!isExportStateTerminal(state)) {
        setState(ExportNotification.State.CANCELLED);
    }
}
```

where the export state gets changed to CANCELLED and the onError gets called and the extra onNext gets sent.

A quick change of 

```java
synchronized(this) {
  setResult(capturedExport.call());
}
```

_does_ appear to fix the issue, but this may be a heavy hammer, and we probably don't want to synchronize around a potentially long call to `capturedExport.call()`.

nbauernfeind:
Fixes #4754.

### DIFF
--- a/server/src/main/java/io/deephaven/server/session/SessionState.java
+++ b/server/src/main/java/io/deephaven/server/session/SessionState.java
@@ -822,7 +822,7 @@ public class SessionState {
                     errorHandler.onError(state, errorId, toReport, failedDependencyLogIdentity);
                 } catch (final Throwable err) {
                     // this is a serious error; crash the jvm to ensure that we don't miss it
-                    log.error().append("Unexpected error while reporting failure: ").append(err).endl();
+                    log.error().append("Unexpected error while reporting ExportObject failure: ").append(err).endl();
                     ProcessEnvironment.getGlobalFatalErrorReporter().reportAsync(
                             "Unexpected error while reporting ExportObject failure", err);
                 }
@@ -834,7 +834,7 @@ public class SessionState {
                     successHandler.accept(result);
                 } catch (final Throwable err) {
                     // this is a serious error; crash the jvm to ensure that we don't miss it
-                    log.error().append("Unexpected error while reporting success: ").append(err).endl();
+                    log.error().append("Unexpected error while reporting ExportObject success: ").append(err).endl();
                     ProcessEnvironment.getGlobalFatalErrorReporter().reportAsync(
                             "Unexpected error while reporting ExportObject success", err);
                 }
--- a/server/src/main/java/io/deephaven/server/session/SessionState.java
+++ b/server/src/main/java/io/deephaven/server/session/SessionState.java
@@ -825,6 +825,7 @@ public class SessionState {
                 } catch (final Exception err) {
                     log.error().append("Unexpected error while reporting success: ").append(err).endl();
                 }
+                successHandler = null;
             }
 
             if (isNowExported || isExportStateTerminal(state)) {
@@ -834,6 +835,7 @@ public class SessionState {
                 parents = Collections.emptyList();
                 exportMain = null;
                 errorHandler = null;
+                successHandler = null;
             }
 
             if ((isNowExported && isNonExport()) || isExportStateTerminal(state)) {
--- a/server/src/main/java/io/deephaven/server/session/SessionState.java
+++ b/server/src/main/java/io/deephaven/server/session/SessionState.java
@@ -56,6 +56,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.function.Consumer;
 
 import static io.deephaven.base.log.LogOutput.MILLIS_FROM_EPOCH_FORMATTER;
 import static io.deephaven.extensions.barrage.util.GrpcUtil.safelyComplete;
@@ -543,7 +544,11 @@ public class SessionState {
         /** This is a reference of the work to-be-done. It is non-null only during the PENDING state. */
         private Callable<T> exportMain;
         /** This is a reference to the error handler to call if this item enters one of the failure states. */
+        @Nullable
         private ExportErrorHandler errorHandler;
+        /** This is a reference to the success handler to call if this item successfully exports. */
+        @Nullable
+        private Consumer<? super T> successHandler;
 
         /** used to keep track of which children need notification on export completion */
         private List<ExportObject<?>> children = Collections.emptyList();
@@ -647,7 +652,10 @@ public class SessionState {
          * @param exportMain the exportMain callable to invoke when dependencies are satisfied
          * @param errorHandler the errorHandler to notify so that it may propagate errors to the requesting client
          */
-        private synchronized void setWork(final Callable<T> exportMain, final ExportErrorHandler errorHandler,
+        private synchronized void setWork(
+                @NotNull final Callable<T> exportMain,
+                @Nullable final ExportErrorHandler errorHandler,
+                @Nullable final Consumer<? super T> successHandler,
                 final boolean requiresSerialQueue) {
             if (hasHadWorkSet) {
                 throw new IllegalStateException("export object can only be defined once");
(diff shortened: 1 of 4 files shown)


devinrsmith (commented):
I've confirmed that this PR appears to fix the race that I've been able to reliably reproduce otherwise.

server/src/main/java/io/deephaven/server/table/ops/TableServiceGrpcImpl.java
```diff
@@ -544,14 +544,12 @@ public void batch(
                         .build();
                 safelyOnNext(responseObserver, response);
                 onOneResolved.run();
-            }).submit(() -> {
-                final Table table = exportBuilder.doExport();
+            }).onSuccess(table -> {
```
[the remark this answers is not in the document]
rcaudy: I think we should strongly consider the following as a matter of policy:
1. Require that all exports and non-exports have an explicit, non-null, `onError` and `onSuccess` set before `submit`.
2. Move our success responses to `onSuccess` and out of the work set in `submit`.

I think this establishes a pattern of safer code, and eliminates some racy responses. Maybe. OK to defer this to a separate PR.

server/src/main/java/io/deephaven/server/session/SessionState.java
```diff
@@ -801,20 +814,31 @@ private synchronized void setState(final ExportNotification.State state) {
 
                     errorHandler.onError(state, errorId, toReport, dependentHandle);
                 } catch (final Exception err) {
-                    log.error().append("Unexpected error while reporting state failure: ").append(err).endl();
+                    log.error().append("Unexpected error while reporting failure: ").append(err).endl();
                 }
             }
 
-            if (state == ExportNotification.State.EXPORTED || isExportStateTerminal(state)) {
+            final boolean isNowExported = state == ExportNotification.State.EXPORTED;
```
rcaudy, line 821: 1. I wonder if we should have more validation of state transitions, e.g. can't go from exported to exported.
2. Is `RELEASED` terminal? Not in the `isExportStateTerminal` impl.

server/src/main/java/io/deephaven/server/session/SessionState.java
```diff
@@ -801,20 +814,31 @@ private synchronized void setState(final ExportNotification.State state) {
...
+            if (isNowExported && successHandler != null) {
+                try {
+                    successHandler.accept(result);
+                } catch (final Exception err) {
+                    log.error().append("Unexpected error while reporting success: ").append(err).endl();
```
rcaudy, line 826: Should we take some action? Release the export? FATAL?

server/src/main/java/io/deephaven/server/session/SessionState.java
```diff
@@ -801,20 +814,31 @@ private synchronized void setState(final ExportNotification.State state) {
...
+                }
+                successHandler = null;
```
rcaudy, line 828: Redundant
```suggestion
```

server/src/main/java/io/deephaven/server/session/SessionState.java
```diff
@@ -904,11 +928,22 @@ private void doExport() {
...
-                    return; // had a cancel race with client
+                // check for some sort of cancel race with client
+                if (state != ExportNotification.State.QUEUED
+                        || session.isExpired()
+                        || capturedExport == null
+                        || !tryRetainReference()) {
+                    if (errorHandler != null) {
+                        // fulfill promise to client that we will notify them whether we succeed or fail
+                        assignErrorId();
+                        errorHandler.onError(ExportNotification.State.CANCELLED, errorId, null, null);
```
rcaudy, line 939: 1. Should `assignErrorId()` be idempotent? If we've already assigned an ID, should we still be calling `onError`?
2. `null, null`?

rcaudy: Should `dependentHandle` be called something more like `failedDependencyHandle`?

server/src/main/java/io/deephaven/server/table/ops/TableServiceGrpcImpl.java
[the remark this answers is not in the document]
nbauernfeind: https://github.com/deephaven/deephaven-core/issues/4779

server/src/main/java/io/deephaven/server/session/SessionState.java
```diff
@@ -812,20 +819,25 @@ private synchronized void setState(final ExportNotification.State state) {
...
                     }
 
-                    errorHandler.onError(state, errorId, toReport, dependentHandle);
-                } catch (final Exception err) {
+                    errorHandler.onError(state, errorId, toReport, failedDependencyLogIdentity);
+                } catch (final Throwable err) {
+                    // this is a serious error; crash the jvm to ensure that we don't miss it
                     log.error().append("Unexpected error while reporting failure: ").append(err).endl();
+                    ProcessEnvironment.getGlobalFatalErrorReporter().reportAsync(
+                            "Unexpected error while reporting ExportObject failure", err);
```
rcaudy, line 827: Should first match the second?

server/src/main/java/io/deephaven/server/session/SessionState.java
```diff
@@ -812,20 +819,25 @@ private synchronized void setState(final ExportNotification.State state) {
...
             final boolean isNowExported = state == ExportNotification.State.EXPORTED;
             if (isNowExported && successHandler != null) {
                 try {
                     successHandler.accept(result);
-                } catch (final Exception err) {
+                } catch (final Throwable err) {
+                    // this is a serious error; crash the jvm to ensure that we don't miss it
                     log.error().append("Unexpected error while reporting success: ").append(err).endl();
+                    ProcessEnvironment.getGlobalFatalErrorReporter().reportAsync(
+                            "Unexpected error while reporting ExportObject success", err);
```
rcaudy, line 839: Propagate

rcaudy (approved):

PR #4780, opened 2023-11-06
JS API must support SKIP as an aggregation type

### LINKED ISSUE #4182: No way for UI to specify skipping a column in aggregations
Related to Web UI issue: https://github.com/deephaven/web-client-ui/issues/1355

UI is unable to specify columns to ignore in aggregations, as the default operation is `Sum`, and `Skip` was removed.

**Steps to Reproduce:**
1. Open up http://localhost:10000/jsapi/rollup.html, and set a breakpoint in the `displayTableDetails` function at the top using the developer tools.
2. Refresh the page. When the breakpoint is hit, highlight the `table` variable and assign it to a global variable, then resume operation
3. In the developer console, execute the following snippet:
```
// `temp1` should be the name of your global variable the table is assigned to
var totalsTable = await temp1.getTotalsTable({ operationMap: { I: ['Sum', 'Count'] } })
console.log('totals columns: ', totalsTable.columns.map(c => c.name))
```

**Expected Results:**
3. Only two columns `I__Sum` and `I__Count` should exist in the output, since aggregations were only specified for `I`

**Actual Results:**
3. There are three columns, `['I__Sum', 'K', 'I__Count']`. `K` should not be in the totals output as the column was not specified

There is no `Skip` operation that can be used either, so there does not appear to be a way to _not_ aggregate a specific column.

niloc132:
Fixes #4182

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/JsTotalsTableConfig.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/JsTotalsTableConfig.java
@@ -83,7 +83,8 @@ public class JsTotalsTableConfig {
             JsAggregationOperation.LAST,
             JsAggregationOperation.COUNT_DISTINCT,
             JsAggregationOperation.DISTINCT,
-            JsAggregationOperation.UNIQUE);
+            JsAggregationOperation.UNIQUE,
+            JsAggregationOperation.SKIP);
 
     /**
      * Specifies if a Totals Table should be expanded by default in the UI. Defaults to false.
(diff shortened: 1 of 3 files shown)


mofojed (approved):
Looks good, tested with the steps in the ticket #4182 but with this snippet specifying the `defaultOperation`:
```
// `temp1` should be the name of your global variable the table is assigned to
var totalsTable = await temp1.getTotalsTable({ defaultOperation: 'Skip', operationMap: { I: ['Sum', 'Count'] } })
console.log('totals columns: ', totalsTable.columns.map(c => c.name))
```
I do find it slightly odd that our default operation is `Sum` instead of `Skip`, but that seems to be consistent with Enterprise.

Issue #4778, opened 2023-11-06
R Client HTML Docs


alexpeters1208:
Remaining work for the R Client lies in getting the source documentation into an HTML website, discoverable from deephaven.io. This will be broken into the following tasks:
1. Get HTML doc source reviewed and merged (DONE in [4782](https://github.com/deephaven/deephaven-core/pull/4782))
2. Get gradle to generate the non-check-ables from the HTML doc source, to generate the .html files (DONE in [4782](https://github.com/deephaven/deephaven-core/pull/4782))
3. Get the .html files that are produced hooked into [deephaven.io](http://deephaven.io/) like the other client docs ([4851](https://github.com/deephaven/deephaven-core/pull/4851))

The first item falls on me (and Chip for review), the second may be a me-Cristian-Colin problem, and the third sounds like a me-Don-Devin problem.

Additionally, the new docsite is made much more navigable when the docs previously called `rdeephaven.Rd`, `AggBy.Rd`, and `UpdateBy.Rd` are moved to vignettes. These are longer-form documentation and belong in a separate tab on the website. Making this change introduces the problem that `install.packages(...build_vignettes = TRUE)` does not build and install the vignettes when the package is built from source. `R CMD build` and `R CMD INSTALL` do take care of the vignette build and install, so we need to move the gradle builds and README to use these commands. This will happen in two steps.
1. Verify that these commands can be made to return a non-zero exit status, so that gradle picks up on failure. (DONE in [4782](https://github.com/deephaven/deephaven-core/pull/4782))
2. Change the build scripts and README to make this the "proper" method of installation. (DONE in [4782](https://github.com/deephaven/deephaven-core/pull/4782))

I will link PRs here as they are opened.

PR #4782, opened 2023-11-07
R HTML docs source code

alexpeters1208:
This adds the necessary source code infrastructure for generating an HTML docsite from R documentation. This _does not_ actually generate the .html files. That will require another PR with CI integration. This PR resolved the first step in #4778.

Most of the changes here _are not_ content changes, but formatting changes and moving things around to get the docsite to work and be easy to use. The best way to review this PR would be to check the docsite. I've generated a preview of this docsite [here](https://alexpeters1208.github.io/rdeephaven/index.html) to make reviewing easier, and so that reviewers may see the final product, and not _only_ the large diff it requires.


R/r-build.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+echo "HELLO THERE"
```
chipkent, line 3: Remove or change the message.

chipkent (approved):

jcferretti (approved):

Issue #4738, opened 2023-10-29
Non-deterministic partition parquet reading


devinrsmith:
I ran into a case where my partitioned parquet reading code would sometimes execute successfully, and sometimes error out.

The non-determinism comes in based on our logic which places special emphasis on the first path returned from `Files.newDirectoryStream`, which has javadoc:

> The elements returned by the iterator are in no specific order

At a minimum, we should be sorting the paths in some consistent fashion to ensure deterministic execution.

In my specific case, some of the parquet files were incorrectly written down w/ the partitioned columns, and when one of those files was the first, I would get the error:

```
RuntimeError: java.lang.IllegalArgumentException: Supplied ColumnDefinitions include duplicate names [KafkaPartition, Date]
	at io.deephaven.engine.table.TableDefinition.checkForNullOrDuplicates(TableDefinition.java:131)
	at io.deephaven.engine.table.TableDefinition.<init>(TableDefinition.java:118)
	at io.deephaven.engine.table.TableDefinition.of(TableDefinition.java:36)
	at io.deephaven.parquet.table.ParquetTools.readPartitionedTableInferSchema(ParquetTools.java:693)
	at io.deephaven.parquet.table.ParquetTools.readTableInternal(ParquetTools.java:585)
	at io.deephaven.parquet.table.ParquetTools.readTable(ParquetTools.java:97)
```

We could argue that this is "user error", but I suggest that we still want to be consistent and potentially improve how we handle sourcing the initial definition for partitioned parquet (maybe the user should be able to explicitly supply definition?).

PR #4783, opened 2023-11-07
Use last lexigraphical key for parquet schema inference

devinrsmith:
In support of https://github.com/deephaven/deephaven-core/issues/4738

### DIFF
--- a/extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
+++ b/extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
@@ -548,8 +548,8 @@ public class ParquetTools {
     /**
      * This method attempts to "do the right thing." It examines the source to determine if it's a single parquet file,
      * a metadata file, or a directory. If it's a directory, it additionally tries to guess the layout to use. Unless a
-     * metadata file is supplied or discovered in the directory, the first found parquet file will be used to infer
-     * schema.
+     * metadata file is supplied or discovered in the directory, the highest (by {@link ParquetTableLocationKey location
+     * key} order) location found will be used to infer schema.
      *
      * @param source The source file or directory
      * @param instructions Instructions for reading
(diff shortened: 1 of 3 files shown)


extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -699,7 +698,7 @@ public static Table readPartitionedTableInferSchema(
         }
         // TODO (https://github.com/deephaven/deephaven-core/issues/877): Support schema merge when discovering multiple
         // parquet files
-        final ParquetTableLocationKey firstKey = foundKeys.get(0);
+        final ParquetTableLocationKey firstKey = sortedKeys.getKnownKeys().get(0);
```
rcaudy, line 701: It would be best to use the *last* key, not the first. We should also update the JavaDoc:
"using a definition built from the highest (by {@link ParquetTableLocationKey location key} order) location found"

In general, locations are sorted such that the most recent comes last; using the last one to define the definition for the data set makes sense. The previous approach allowed the key finder to determine which one was most important, but this approach isn't unreasonable.

devinrsmith: I thought about that, but shouldn't we consider that a breaking change? Happy to make it, just wanting to make sure I'm thinking about it correctly.

Do any of the existing key finders apply any specific orderings to the keys?

rcaudy: 1. I don't think we should consider it a breaking change; I bet 99.9% of our users have never seen that JavaDoc, and are using us on datasets where all files have the same schema. Anyone who is using us "in production" should probably be specifying a definition or using metadata files.
2. In all cases but one, they supply locations in directory traversal order. The exception is the metadata file-based key finder, which will use the order found in the `_metadata` file.
3. Sorting on the TLKs is applied *later* when accepting non-empty locations to include in the table row set, and that order is strictly defined.

rcaudy (approved):

PR #4784, opened 2023-11-07
Added in an experimental Windowed table.

supernomad:
This implements a windowing partitioned table, that can take in timestamped data and window across any period of time for any amount of backlook desired.

Generally the implementation:
- Leverages an input Blink table that has a timestamp column.
- Creates a new `__WINDOWED__` column that it partitions the input by using a call to `partitionedAggBy`. 
- Attaches a custom `InstrumentedTableUpdateListenerAdapter` implementation to copy tables from the resulting `PartitionedTable` to a managed version.
- Leveraging a `PriorityQueue` to track and expire old windows.

The build is just tangentially connected up here, and I will be honest I am not a Java developer so not 100% versed on how to get this all wired in properly. It does however build and passes the `spotless` check.

### DIFF
--- a/extensions/window/build.gradle
+++ b/extensions/window/build.gradle
@@ -1,0 +1,21 @@
+plugins {
+    id 'java-library'
+    id 'io.deephaven.project.register'
+}
+
+description 'Window: Integrating a time bucketing partitioned table'
+
+dependencies {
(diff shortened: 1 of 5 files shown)


supernomad:

PR #4789, opened 2023-11-07
Add key column and column properties to JsPartitionedTable

### LINKED ISSUE #4773: JsPartitionedTable key columns not accessible
I am trying to create a UI for PartitionedTable in web client, but I don't have a way of accessible the key columns. I would like to be able to call a function similar to `PartitionedTable.key_columns` and get information (i.e. names) about the columns. Currently, I am only able to access the key values with `getKeys` and column types with `keyColumnTypes`.

georgecwan:
Added `getKeyColumns`, `getColumns`, and `getKeyTable` to `JsPartitionedTable`.

Resolves #4773.

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
@@ -47,10 +47,10 @@ public class JsPartitionedTable extends HasLifecycle implements ServerObject {
     public static final String EVENT_KEYADDED = "keyadded",
             EVENT_DISCONNECT = JsTable.EVENT_DISCONNECT,
             EVENT_RECONNECT = JsTable.EVENT_RECONNECT,
-    /**
-     * Indicates that an error has occurred while communicating with the server.
-     */
-    EVENT_RECONNECTFAILED = JsTable.EVENT_RECONNECTFAILED;
+            /**
+             * Indicates that an error has occurred while communicating with the server.
+             */
+            EVENT_RECONNECTFAILED = JsTable.EVENT_RECONNECTFAILED;
 
     private final WorkerConnection connection;
     private final JsWidget widget;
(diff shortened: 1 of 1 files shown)


web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
```diff
@@ -83,12 +88,18 @@ public Promise<JsPartitionedTable> refetch() {
             descriptor = PartitionedTableDescriptor.deserializeBinary(w.getDataAsU8());
 
             keyColumnTypes = new ArrayList<>();
+            InitialTableDefinition tableDefinition = WebBarrageUtils.readTableDefinition(
+                    WebBarrageUtils.readSchemaMessage(descriptor.getConstituentDefinitionSchema_asU8()));
             ColumnDefinition[] columnDefinitions = WebBarrageUtils.readColumnDefinitions(
```
niloc132, line 93: this line can be removed, since column definitions are part of the table definitions

web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
```diff
@@ -83,12 +88,18 @@ public Promise<JsPartitionedTable> refetch() {
...
                     WebBarrageUtils.readSchemaMessage(descriptor.getConstituentDefinitionSchema_asU8()));
-            for (int i = 0; i < columnDefinitions.length; i++) {
-                ColumnDefinition columnDefinition = columnDefinitions[i];
+            columns = new Column[0];
+            keyColumns = new Column[0];
+            for (ColumnDefinition columnDefinition : columnDefinitions) {
```
niloc132, line 97: java iterators don't translate as cleanly to patterns that make sense in JS, this is better left as a `for(int i =...` loop

web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
```diff
@@ -83,12 +88,18 @@ public Promise<JsPartitionedTable> refetch() {
...
+                Column column = columnDefinition.makeJsColumn(columns.length, tableDefinition.getColumnsByName());
+                columns[columns.length] = column;
                 if (descriptor.getKeyColumnNamesList().indexOf(columnDefinition.getName()) != -1) {
                     keyColumnTypes.add(columnDefinition.getType());
+                    keyColumns[keyColumns.length] = column;
                 }
             }
```
niloc132, line 104: After all of the items are added to the new arrays, freeze them with `JsObject.freeze(...)`, so they are safe to return to the user.

web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
```diff
@@ -248,6 +259,26 @@ public int size() {
         return tables.size();
     }
 
+    /**
+     * An array of all the key columns used by the table
```
niloc132, line 263: Javadoc should be sentences, end in a `.` - I think it might be more accurate to say that these are the key columns that the tables are partitioned by, or something to that effect?

web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
```diff
@@ -248,6 +259,26 @@ public int size() {
...
+     *
+     * @return Array of Column
+     */
+    @JsProperty
+    public Column[] getKeyColumns() {
+        return keyColumns;
+    }
+
+    /**
+     * An array of all the columns used by the table
```
niloc132, line 273: Might be clearer as "An array of the columns in the tables that can be retrieved from this partitioned table" or something to that effect?

web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
```diff
@@ -248,6 +262,37 @@ public int size() {
...
+        return columns;
+    }
+
+    /**
+     * A Table object containing all currently known keys used for this partitioned table.
+     *
+     * @return Table
+     */
+    @JsProperty
+    public JsTable getKeyTable() {
```
niloc132, line 292: This is probably not safe to return - what are you trying to do with this as an api consumer?

georgecwan: The new UI being implemented for Partitioned Tables ([spec](https://github.com/deephaven/web-client-ui/issues/1143)) need to be able to display the table containing all the possible keys.

niloc132 (approved):

PR #4775, opened 2023-11-06
Added support for Parquet TIME

malhotrashivam:
Closes #4753 

Interesting reference: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#temporal-types

PR #4801, opened 2023-11-09
Bug fixes and improved support for Parquet TIMESTAMP

malhotrashivam:
As part of #4421, we started throwing an exception on reading Parquet TIMESTAMP fields with isAdjustedToUTC set as false. After this change:
- Such fields will be read as `java.time.LocalDateTime`
- `java.time.LocalDateTime` columns will be written as Parquet TIMESTAMP fields with `isAdjustedToUTC=false`. Earlier they were written as binary data with a codec.

Also, this PR fixes the bugs introduced in #4775 and #4755 that can lead to a `ClassCastException` in some cases on reading Parquet DATE and TIME columns.

Related to #976


jmao-denver (approved):
The Python changes LGTM.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileReader.java
```diff
@@ -368,23 +358,23 @@ private static LogicalTypeAnnotation getLogicalTypeAnnotation(ConvertedType type
...
-                // TODO(deephaven-core#976) Assuming that time is adjusted to UTC
+                // isAdjustedToUTC parameter is ignored while reading Parquet TIME type, so assume it whatever
                 return LogicalTypeAnnotation.timeType(true, LogicalTypeAnnotation.TimeUnit.MILLIS);
             case TIME_MICROS:
-                // TODO(deephaven-core#976) Assuming that time is adjusted to UTC
                 return LogicalTypeAnnotation.timeType(true, LogicalTypeAnnotation.TimeUnit.MICROS);
             case TIMESTAMP_MILLIS:
-                // TODO(deephaven-core#976) Assuming that time is adjusted to UTC
-                return LogicalTypeAnnotation.timestampType(true, LogicalTypeAnnotation.TimeUnit.MILLIS);
+                // Converted type doesn't have isAdjustedToUTC parameter, so use the information from logical type
```
malhotrashivam, line 372: This is an ugly hack that I had to do because of the code inside `ParquetFileReader::buildChildren` which looks like this:

```
if (schemaElement.isSetConverted_type()) {
    LogicalTypeAnnotation originalType =
            getLogicalTypeAnnotation(schemaElement.converted_type, schemaElement);
    LogicalTypeAnnotation newOriginalType = schemaElement.isSetLogicalType()
            && getLogicalTypeAnnotation(schemaElement.logicalType) != null
                    ? getLogicalTypeAnnotation(schemaElement.logicalType)
                    : null;
    if (!originalType.equals(newOriginalType)) {
        ((Types.Builder) childBuilder).as(originalType);
    }
}
```

Mainly we are trying to deduce the type of the column from `schemaElement.converted_type` and `schemaElement.logicalType` and in case of mismatch, we go with type deduced from `schemaElement.converted_type`.

The problem is that `schemaElement.converted_type` doesn't have any information about adjustments to UTC. So to deduce it correctly, I had to read the `isAdjustedToUTC` from the logical type.

Ideally, I would have liked to change the above code to prefer the type deduced from `schemaElement.logicalType` over `schemaElement.converted_type` because `converted_type` is deprecated and superseded by `logicalType` (based on the Javadoc for converted_type).
But this would be a much bigger change and impact all parquet types. So I added this hack which just impacts the TIMESTAMP type.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/DateTransfer.java
```diff
@@ -12,7 +12,7 @@
 
 import java.time.LocalDate;
 
-final class DateTransfer extends IntCastablePrimitiveTransfer<WritableObjectChunk<LocalDate, Values>> {
+final class DateTransfer extends IntCastablePrimitiveTransfer<ObjectChunk<LocalDate, Values>> {
```
malhotrashivam, line 15: Some of my tests were failing with `WritableObjectChunk` for `LocalDateTimeTransfer`. So I replaced it with `ObjectChunk` and also changed the same for `DateTransfer`.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -981,6 +981,20 @@ public static long epochNanos(@Nullable final ZonedDateTime dateTime) {
...
+     *
+     * @param localDateTime the local date time to compute the Epoch offset for
+     * @return nanoseconds since Epoch, or a NULL_LONG value if the local date time is null
+     */
+    @ScriptApi
+    public static long epochNanosUTC(@Nullable final LocalDateTime localDateTime) {
+        if (localDateTime == null) {
+            return NULL_LONG;
+        }
+        return epochNanos(localDateTime.toInstant(ZoneOffset.UTC));
```
rcaudy, line 995: We should probably just compute nanos as:
```
        TimeUnit.SECOND.toNanos(localDateTime.toEpochSecond(offset)) +  localDateTime.toLocalTime().getNano();
```

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -1399,6 +1413,52 @@ public static ZonedDateTime excelToZonedDateTime(final double excel, @Nullable f
...
+     *         Epoch converted to a {@link LocalDateTime} in UTC timezone
+     */
+    public static @Nullable LocalDateTime epochMillisToLocalDateTimeUTC(final long millis) {
+        if (millis == QueryConstants.NULL_LONG) {
+            return null;
+        }
+        final long millisPerSecond = 1_000;
+        return LocalDateTime.ofEpochSecond(millis / millisPerSecond, (int) ((millis % millisPerSecond) * MILLI),
+                ZoneOffset.UTC);
+    }
```
rcaudy, line 1460: I would convert mills or micros to nanos and just delegate.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileReader.java
```diff
@@ -368,23 +358,23 @@ private static LogicalTypeAnnotation getLogicalTypeAnnotation(ConvertedType type
...
-                int scale = schemaElement == null ? 0 : schemaElement.scale;
-                int precision = schemaElement == null ? 0 : schemaElement.precision;
+                final int scale = schemaElement == null ? 0 : schemaElement.scale;
+                final int precision = schemaElement == null ? 0 : schemaElement.precision;
                 return LogicalTypeAnnotation.decimalType(scale, precision);
             case DATE:
                 return LogicalTypeAnnotation.dateType();
             case TIME_MILLIS:
-                // TODO(deephaven-core#976) Assuming that time is adjusted to UTC
+                // isAdjustedToUTC parameter is ignored while reading Parquet TIME type, so assume it whatever
```
rcaudy, line 367:
```suggestion
                // isAdjustedToUTC parameter is ignored while reading Parquet TIME type, so disregard it here
```

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileReader.java
rcaudy: What would be the side effect of switching to use the logical type?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/TypeInfos.java
```diff
@@ -376,11 +378,33 @@ public PrimitiveBuilder<PrimitiveType> getBuilder(boolean required, boolean repe
...
+        @Override
+        public Set<Class<?>> getTypes() {
+            return clazzes;
+        }
+
+        @Override
+        public PrimitiveBuilder<PrimitiveType> getBuilder(boolean required, boolean repeating, Class<?> dataType) {
+            if (!isValidFor(dataType)) {
+                throw new IllegalArgumentException("Invalid data type " + dataType);
+            }
```
rcaudy, line 401: We copy this everywhere. Is there a way to move it up to an interface (defaulted), or at least cut it down to a one-liner?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/transfer/DateTransfer.java
rcaudy: `ObjectChunk` is correct. Please check all of the `Getting` transfers and make sure they don't use the writable interface. We're going to want to point release for that.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileReader.java
malhotrashivam: I have added a test commit [here](https://github.com/deephaven/deephaven-core/compare/main...malhotrashivam:deephaven-core:nightly/sm-test-branch) and started the [nightly jobs](https://github.com/malhotrashivam/deephaven-core/actions/runs/6827594756). Let me see how this work.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -1438,9 +1440,7 @@ public static ZonedDateTime excelToZonedDateTime(final double excel, @Nullable f
         if (micros == QueryConstants.NULL_LONG) {
             return null;
         }
-        final long microsPerSecond = 1_000_000;
-        return LocalDateTime.ofEpochSecond(micros / microsPerSecond, (int) ((micros % microsPerSecond) * MICRO),
-                ZoneOffset.UTC);
+        return epochNanosToLocalDateTimeUTC(micros * MICRO);
```
rcaudy, line 1443: I was wrong; the previous code better matches our existing implementations. Below matches "epochMicrosToInstant`:
```
        return micros == NULL_LONG ? null : LocalDateTime.ofEpochSecond(micros / 1_000_000L, (int) ((micros % 1_000_000L) * MICRO), ZoneOffset.UTC);
```

And the following for millis:
```
return millis == NULL_LONG ? null : LocalDateTime.ofEpochSecond(millis / 1_000L, (int) ((millis % 1_000L) * MILLI), ZoneOffset.UTC);
```

rcaudy: This would argue for the nanos one to be:
```
        return nanos == NULL_LONG ? null : LocalDateTime.ofEpochSecond(nanos / 1_000_000_000L, (int) (nanos % 1_000_000_000L), ZoneOffset.UTC);
```

rcaudy (approved):

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetFileReader.java
malhotrashivam: The nightly jobs run fine.

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -981,6 +982,21 @@ public static long epochNanos(@Nullable final ZonedDateTime dateTime) {
         return safeComputeNanos(dateTime.toEpochSecond(), dateTime.getNano());
     }
 
+    /**
```
chipkent, line 985: In case my comments seem strong, this library is very exposed to users, so it needs to be ultra currated.  Functions should only get added when there is a compelling reason.

1. I am not a fan of having hard-coded methods for any specific timezone.  
2. The methods should accept a time zone as an input.
3. If methods are added for `LocalDateTime`, `LocalDateTime` signatures should be added to all relevant methods.

Issue #4798, opened 2023-11-08
TableServiceAsyncTest timeout unit test failure


devinrsmith:
It appears that the following test is sometimes failing:

```
2023-11-08T18:52:13.9139040Z io.deephaven.client.impl.TableServiceAsyncTest > longChainAsyncExportOnlyLast FAILED
2023-11-08T18:52:13.9142554Z     java.util.concurrent.TimeoutException at TableServiceAsyncTest.java:79
```

It suggests that there is an underlying race (potentially Batch / ETCR or related), or client implementation bug; but it's possible the timeout is too short given CI resources, TBD.

devinrsmith:
As seen in https://github.com/deephaven/deephaven-core/actions/runs/6802272060/job/18495015941

devinrsmith:
I've done 8 CI runs and so far have been unable to reproduce this issue. After #4802 merges, if this issues triggers in CI again we should have much more details to go off of.

devinrsmith:
I _think_ #4802 fixes this - leaving open for a week to see if we here anymore cases.

devinrsmith:
Updated comment to specifically make sure this is about the _timeout_ nature of this issue.

PR #4802, opened 2023-11-09
Properly pass along exceptional state in BatchHandler

devinrsmith:
This is in support of #4798. While this _is_ a fix to BatchHandler logic, it isn't the underlying cause of 4798. Next time 4798 is triggered in CI though, we should have more details to go off of.

### DIFF
--- a/java-client/session/src/main/java/io/deephaven/client/impl/ExportStates.java
+++ b/java-client/session/src/main/java/io/deephaven/client/impl/ExportStates.java
@@ -381,9 +381,11 @@ final class ExportStates implements ExportService {
         private static final Logger log = LoggerFactory.getLogger(BatchHandler.class);
 
         private final Map<Integer, State> newStates;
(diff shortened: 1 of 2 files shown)


java-client/session/src/main/java/io/deephaven/client/impl/ExportStates.java
```diff
@@ -398,11 +400,16 @@ public void onNext(ExportedTableCreationResponse value) {
                         "Not expecting export creation responses for empty tickets");
             }
             final int exportId = ExportTicketHelper.ticketToExportId(value.getResultId().getTicket(), "export");
-            final State state = newStates.remove(exportId);
+            final State state = newStates.get(exportId);
```
rcaudy, line 403: Should we keep some state to catch dup ETCRs from the server?

rcaudy (approved):

PR #4804, opened 2023-11-09
Server logs should be written when integration tests fail

### LINKED ISSUE #4766: Server logs not written out after go:testGoClientRun failure
Noticed as part of failed check run https://github.com/deephaven/deephaven-core/actions/runs/6735593616/attempts/1?pr=4764,

```
2023-11-02T16:52:45.8435194Z > Task :go:testGoClientRun FAILED
2023-11-02T16:52:45.9433547Z > Task :go:testGoClientSyncAfterFail
2023-11-02T16:52:45.9435887Z > Task :go:writeServerLogs SKIPPED
```

that `go:writeServerLogs` was _not_ executed.

The build logic is setup such that:

```groovy
deephavenDocker.shouldLogIfTaskFails testGoClient
```

It's possible that this is a regression due to #4692. I haven't checked if other docker testing setups exhibit the same issue.

niloc132:
Fixes #4766

### DIFF
--- a/buildSrc/src/main/groovy/Docker.groovy
+++ b/buildSrc/src/main/groovy/Docker.groovy
@@ -300,7 +300,34 @@ class Docker {
 
         if (cfg.copyOut && !cfg.showLogsOnSuccess) {
             // Single task with explicit inputs and outputs, to let gradle detect if it is up to date, and let docker
-            // cache what it can.
+            // cache what it can. While far more efficient for gradle to run (or rather, know when it does not need to
+            // run), it is also more bug-prone while we try to get various competing features working.
+            //
+            // To handle these use cases, we're not using dependsOn as we typically would do, but instead using
+            // finalizedBy and onlyIf. Here's a mermaid diagram:
+            //
+            // graph LR;
+            //     MakeImage -. finalizedBy .-> Run
+            //     Sync -- dependsOn --> MakeImage
+            //     Run -. finalizedBy .-> Sync
+            //
+            //
+            // Unlike "A dependsOn B", "B finalized A" will let B run if A failed, and will not run A if B must run.
+            // Combining the chain of finalizedBys between MakeImage <- Run <- Sync with the dependsOn from
+            // Sync -> MakeImage lets us handle the following cases:
+            // * Successful run, output is sync'd afterwards, final task succeeds
+            // * Failed run, output is sync'd afterwards, final task fails
+            // * Failed image creation, no run, no sync, no final task
+            // * Previously successful run with no source changes, no tasks run (all "UP-TO-DATE")
+            //
+            // Tests to run to confirm functionality:
+            // * After changes, confirm that :web:assemble runs (isn't all "UP-TO-DATE")
+            // * Then run again with no changes, confirm all are UP-TO-DATE, roughly 2s build time
+            // * Edit a test that uses deephavenDocker to fail, confirm that the test fails, that the test-reports
+            //   are copied out, and that server logs are written to console
+            // * Ensure that if the test is set to pass that the test-reports are copied out, and server logs are
+            //   not written.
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/groovy/Docker.groovy
```diff
@@ -337,35 +364,34 @@ class Docker {
                 }
             }
 
-            // Handle copying failure. This is now distinct from the "actual" Sync task that depends directly
-            // on the CombinedDockerRunTask.
-            def syncAfterFail = project.tasks.register("${taskName}SyncAfterFail", Sync) { sync ->
+            // Specify that makeImage is finalized by buildAndRun - that is, in this configuration buildAndRun
+            // must run after makeImage finishes
```
devinrsmith, line 368: Your comment is a bit at odds with the typical gradle terminology. I'm digging in now; I don't _know_ if finalizedBy is the technically correct solution here (it may very well be...).

https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:finalizer_tasks
> Finalizer tasks are useful in situations where the build creates a resource that has to be cleaned up regardless of the build failing or succeeding. 

https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:ordering_tasks
> When you use the “must run after” ordering rule you specify that taskB must always run after taskA, whenever both taskA and taskB will be run. This is expressed as taskB.mustRunAfter(taskA).

buildSrc/src/main/groovy/Docker.groovy
```diff
@@ -337,35 +364,34 @@ class Docker {
...
                     // run the provided closure first
                     cfg.copyOut.execute(sync)
 
                     // then set the from location
                     from dockerCopyLocation
 
-                    onlyIf { buildAndRun.get().state.failure != null }
+                    doLast {
+                        if (buildAndRun.get().state.failure != null) {
+                            throw new GradleException('Docker task failed, see earlier task failures for details')
```
devinrsmith, line 385: I might add a comment this isn't an illegal state, we just want to preserve / propagate the failure, correct? I wonder if we should have a separate task that propagates this failure, instead of attaching it to the Sync task? (which make have specific semantics attached to sync-related failures.)

devinrsmith (commented):
I'm going to dig in a bit more - posted initial thoughts.

buildSrc/src/main/groovy/Docker.groovy
niloc132: `mustRunAfter`/`shouldRunAfter` only sets an ordering requirement, but does not mandate that the earlier or later task run at all, they can run or not run independently.

On the other hand, `b.dependsOn(a)` means that if `b` is requested to run, `a` must run first and must be successful, whereas `a.finalizedBy(b)` means that if `a` will run, `b` must run _afterwards_ (regardless of whether or not `a` is successful).

buildSrc/src/main/groovy/Docker.groovy
niloc132: Correct - not an illegal state (and the choice of GradleException is I thought intended to convey that), but just a failure. If this task did not throw an exception (or otherwise fail), it wouldn't be clear to other tasks that they should/shouldn't continue (e.g. "if it failed, dont continue to build downstream artifacts" vs "if it failed, make sure we dump the server's logs").

jcferretti (approved):

devinrsmith (dismissed):

buildSrc/src/main/groovy/Docker.groovy
niloc132: Note also that a new task with actions (i.e. "fail if that other thing failed") but no inputs or outputs can never be up to date. I'm not aware of Sync specific errors, but can add a comment here to call out what is going on, besides the wall of text written at the top of this section.

devinrsmith (approved):

PR #4805, opened 2023-11-09
Update protoc and cpp client base images.

jcferretti:
This should fix the issue with the protoc image generation that was introduced in the R related refactor for deephaven-base-images.

We are also updating the cpp image to include the change in build-dependencies.sh (and link to it from the cpp-client/README) that allows for customizing the github URL (per Raffi's request).

### DIFF
--- a/cpp-client/README.md
+++ b/cpp-client/README.md
@@ -37,7 +37,7 @@ on them anymore so we do notguarantee they are current for those platforms.
    Get the `build-dependencies.sh` script from Deephaven's base images repository
    at the correct version.
    You can download it directly from the link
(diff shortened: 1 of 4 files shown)


devinrsmith:
FYI @niloc132; https://github.com/deephaven/deephaven-base-images/pull/102 look OK to you?

niloc132 (approved):
I looked at this and the linked PR in base-images - I don't think I have anything constructive to add or to call out. The changes might not appear ideal, but they come from a place of "this is the clearest/quickest way to keep our build wiring consistent", and its hard to argue with that.

devinrsmith (approved):

PR #4808, opened 2023-11-10
`TableServiceAsyncTest`: Eliminate Compiler Overhead Variance; Close `QueryCompiler`'s `JavaFileManager`

### LINKED ISSUE #4798: TableServiceAsyncTest timeout unit test failure

nbauernfeind:
Fixes #4798.
Follow up work #4814.

Whenever I could catch a timeout in the debugger the server was not blocked, always making progress, but spending time in compilation. I've eliminated the need to recompile formulas for each operation in the chain and added some validation that the data in the result tables are as expected.

While testing ways to reproduce and/or fix the original issue, I discovered that the `JavaFileManager` that we're creating on every compilation request opened file handles for many of the jars on my classpath. After around 50'ish queries the `QueryCompiler` broke complaining of `Too many open files`. Running a quick `lsof | cut | sort | uniq -c` showed 40+ open file handles each on many of the jars on my classpath. I've done the bare minimum in this PR, will write a follow up issue, and discuss via team-slack.

Multiplexed this PR's Check CI here: https://github.com/nbauernfeind/deephaven-core/actions/runs/6830355652
Multiplexed current main's Check CI here: https://github.com/nbauernfeind/deephaven-core/actions/runs/6830357421/
(I expect to see a failure or three in the current main and none for this PR as it's hard to tell if the issue has been fixed.)

### DIFF
--- a/engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
+++ b/engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
@@ -697,7 +697,7 @@ public class QueryCompiler {
         final String classPathAsString = getClassPath() + File.pathSeparator + getJavaClassPath();
         final List<String> compilerOptions = Arrays.asList("-d", tempDirAsString, "-cp", classPathAsString);
 
-        final StandardJavaFileManager fileManager = compiler.getStandardFileManager(null, null, null);
+        final JavaFileManager fileManager = compiler.getStandardFileManager(null, null, null);
 
         boolean result = false;
         boolean exceptionThrown = false;
--- a/engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
+++ b/engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
@@ -683,19 +683,60 @@ public class QueryCompiler {
         }
     }
 
+    private static JavaCompiler JAVA_COMPILER = null;
+
+    private static JavaCompiler getJavaCompiler() {
+        JavaCompiler localCompiler = JAVA_COMPILER;
+        if (localCompiler != null) {
+            return localCompiler;
+        }
+        synchronized (QueryCompiler.class) {
+            localCompiler = JAVA_COMPILER;
+            if (localCompiler != null) {
+                return localCompiler;
+            }
+            localCompiler = JAVA_COMPILER = ToolProvider.getSystemJavaCompiler();
+            if (localCompiler == null) {
+                throw new RuntimeException("No Java compiler provided - are you using a JRE instead of a JDK?");
+            }
+        }
+        return localCompiler;
+    }
+
+    /**
+     * While the JavaFileManager should be closed to clean up resources, using a singleton avoids repeated processing of
+     * the classpath, which is <b>very</b> expensive.
+     */
+    private static JavaFileManager JAVA_FILE_MANAGER = null;
+
+    private static JavaFileManager getJavaFileManager() {
+        JavaFileManager localManager = JAVA_FILE_MANAGER;
+        if (localManager != null) {
+            return localManager;
+        }
+        synchronized (QueryCompiler.class) {
+            localManager = JAVA_FILE_MANAGER;
+            if (localManager != null) {
+                return localManager;
+            }
+            localManager = JAVA_FILE_MANAGER = getJavaCompiler().getStandardFileManager(null, null, null);
+            if (localManager == null) {
+                throw new RuntimeException("No Java compiler provided - are you using a JRE instead of a JDK?");
+            }
+        }
+        return localManager;
+    }
+
     private void maybeCreateClassHelper(String fqClassName, String finalCode, String[] splitPackageName,
             String rootPathAsString, String tempDirAsString) {
         final StringWriter compilerOutput = new StringWriter();
 
-        final JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
-        if (compiler == null) {
-            throw new RuntimeException("No Java compiler provided - are you using a JRE instead of a JDK?");
-        }
+        final JavaCompiler compiler = getJavaCompiler();
 
         final String classPathAsString = getClassPath() + File.pathSeparator + getJavaClassPath();
         final List<String> compilerOptions = Arrays.asList("-d", tempDirAsString, "-cp", classPathAsString);
 
-        final StandardJavaFileManager fileManager = compiler.getStandardFileManager(null, null, null);
+        final JavaFileManager fileManager = getJavaFileManager();
 
         final boolean result = compiler.getTask(compilerOutput,
                 fileManager,
@@ -735,7 +776,7 @@ public class QueryCompiler {
      * @return a Pair of success, and the compiler output
      */
     private Pair<Boolean, String> tryCompile(File basePath, Collection<File> javaFiles) throws IOException {
-        final JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+        final JavaCompiler compiler = getJavaCompiler();
         if (compiler == null) {
             throw new RuntimeException("No Java compiler provided - are you using a JRE instead of a JDK?");
         }
--- a/engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
+++ b/engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
@@ -683,22 +683,19 @@ public class QueryCompiler {
         }
     }
 
-    private static JavaCompiler JAVA_COMPILER = null;
+    private static volatile JavaCompiler JAVA_COMPILER = null;
 
     private static JavaCompiler getJavaCompiler() {
-        JavaCompiler localCompiler = JAVA_COMPILER;
-        if (localCompiler != null) {
-            return localCompiler;
+        JavaCompiler localCompiler;
+        if ((localCompiler = JAVA_COMPILER) == null) {
+            synchronized (QueryCompiler.class) {
+                if ((localCompiler = JAVA_COMPILER) == null) {
+                    localCompiler = JAVA_COMPILER = ToolProvider.getSystemJavaCompiler();
+                }
+            }
         }
-        synchronized (QueryCompiler.class) {
-            localCompiler = JAVA_COMPILER;
-            if (localCompiler != null) {
-                return localCompiler;
-            }
-            localCompiler = JAVA_COMPILER = ToolProvider.getSystemJavaCompiler();
-            if (localCompiler == null) {
-                throw new RuntimeException("No Java compiler provided - are you using a JRE instead of a JDK?");
-            }
+        if (localCompiler == null) {
+            throw new RuntimeException("No Java compiler provided - are you using a JRE instead of a JDK?");
         }
         return localCompiler;
     }
@@ -707,21 +704,15 @@ public class QueryCompiler {
      * While the JavaFileManager should be closed to clean up resources, using a singleton avoids repeated processing of
      * the classpath, which is <b>very</b> expensive.
      */
-    private static JavaFileManager JAVA_FILE_MANAGER = null;
+    private static volatile JavaFileManager JAVA_FILE_MANAGER = null;
 
     private static JavaFileManager getJavaFileManager() {
-        JavaFileManager localManager = JAVA_FILE_MANAGER;
-        if (localManager != null) {
-            return localManager;
-        }
-        synchronized (QueryCompiler.class) {
-            localManager = JAVA_FILE_MANAGER;
-            if (localManager != null) {
-                return localManager;
-            }
-            localManager = JAVA_FILE_MANAGER = getJavaCompiler().getStandardFileManager(null, null, null);
-            if (localManager == null) {
-                throw new RuntimeException("No Java compiler provided - are you using a JRE instead of a JDK?");
+        JavaFileManager localManager;
+        if ((localManager = JAVA_FILE_MANAGER) == null) {
+            synchronized (QueryCompiler.class) {
+                if ((localManager = JAVA_FILE_MANAGER) == null) {
+                    localManager = JAVA_FILE_MANAGER = getJavaCompiler().getStandardFileManager(null, null, null);
+                }
             }
         }
         return localManager;
--- a/engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
+++ b/engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
@@ -683,55 +683,23 @@ public class QueryCompiler {
         }
     }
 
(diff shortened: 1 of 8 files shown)


java-client/session-dagger/src/test/java/io/deephaven/client/impl/TableServiceAsyncTest.java
```diff
@@ -93,8 +93,11 @@ private static List<TableSpec> createLongChain(int numColumns, int numRows) {
             if (i == 0) {
                 longChain.add(TableSpec.empty(numRows).view("I_0=ii"));
             } else {
+                // create a chain using the previous table even though we don't need the data
                 final TableSpec prev = longChain.get(i - 1);
-                longChain.add(prev.updateView("I_" + i + " = 1 + I_" + (i - 1)));
+                // to avoid timeouts due to compilation variance reuse the same expression, including the destination
+                // column which shows up in the generated formula's FormulaEvaluationException
+                longChain.add(prev.updateView("I_1 = 1 + I_0"));
```
devinrsmith, line 100: I'm not sure how this is directly related to #4798; that is _specifically_ around a timeout error. I'd assume in the case of too many open files in the compilation path there will be some _other_ error that causes TableServiceAsyncTest to fail, correct?

devinrsmith, line 100: I think reworking as `prev.updateView("Next = Current + 1", "Current = Next")` better matches the original intentions of the query.

engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -683,19 +683,60 @@ private void maybeCreateClass(String className, String code, String packageName,
         }
     }
 
+    private static JavaCompiler JAVA_COMPILER = null;
```
devinrsmith, line 686: Needs to be volatile for this pattern.

engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -683,19 +683,60 @@ private void maybeCreateClass(String className, String code, String packageName,
...
+    private static JavaCompiler getJavaCompiler() {
+        JavaCompiler localCompiler = JAVA_COMPILER;
+        if (localCompiler != null) {
+            return localCompiler;
+        }
+        synchronized (QueryCompiler.class) {
+            localCompiler = JAVA_COMPILER;
+            if (localCompiler != null) {
+                return localCompiler;
+            }
```
devinrsmith, line 697: Bit more compact to do:

```java
JavaCompiler localCompiler;
if ((localCompiler = JAVA_COMPILER) == null) {
   sync (...) {
     if ((localCompiler = JAVA_COMPILER) == null) {
     ...
     }
}
return localCompiler;
}
```

java-client/session-dagger/src/test/java/io/deephaven/client/impl/TableServiceAsyncTest.java
nbauernfeind: Everytime it was timed out the server was in the middle of compiling. I did a bunch of instrumentation - there is no real issue with the test, just that the variance of compilation causes it to time out. Repeated local runs are actually harder to time out because the classes have been precompiled.

The change that you are commenting on here, is the actual fix -- run the long chain without repeatedly invoking the compiler. If you prefer, we can split this out from the singleton creation.

nbauernfeind: The test currently does not validate the data. As far as I can tell, the purpose is to ensure that some table returns at the end of the chain.

I like the proposed change as it eliminates the need to recompile the formulas repeatedly while preserving the original implementation. Though, perhaps we should be adding an assert table equals to validate the expected data is returned.

devinrsmith: Ah, makes sense; I never saw that high of variation likely due to the points you bring up.

devinrsmith: Shouldn't my proposal not need repeated compilation? The strings are fixed.

nbauernfeind: > I like the proposed change [...]

I meant to reference your suggestion which is a nice improvement.

engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -683,19 +683,51 @@ private void maybeCreateClass(String className, String code, String packageName,
...
     }
 
+    private static volatile JavaCompiler JAVA_COMPILER = null;
+
+    private static JavaCompiler getJavaCompiler() {
+        JavaCompiler localCompiler;
+        if ((localCompiler = JAVA_COMPILER) == null) {
+            synchronized (QueryCompiler.class) {
+                if ((localCompiler = JAVA_COMPILER) == null) {
+                    localCompiler = JAVA_COMPILER = ToolProvider.getSystemJavaCompiler();
```
devinrsmith, line 693: One small improvement:

```java
JAVA_COMPILER = localCompiler = ...
```

does a volatile write, whereas the existing one does volatile write + read.

engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -683,19 +683,51 @@ private void maybeCreateClass(String className, String code, String packageName,
...
+     * the classpath, which is <b>very</b> expensive.
+     */
+    private static volatile JavaFileManager JAVA_FILE_MANAGER = null;
+
+    private static JavaFileManager getJavaFileManager() {
+        JavaFileManager localManager;
+        if ((localManager = JAVA_FILE_MANAGER) == null) {
+            synchronized (QueryCompiler.class) {
+                if ((localManager = JAVA_FILE_MANAGER) == null) {
+                    localManager = JAVA_FILE_MANAGER = getJavaCompiler().getStandardFileManager(null, null, null);
```
devinrsmith, line 714: Ditto.

java-client/session/src/main/java/io/deephaven/client/impl/Session.java
```diff
@@ -22,6 +24,12 @@ public interface Session
     @Override
     void close();
 
+    /**
+     * Returns the current auth token.
+     */
+    @VisibleForTesting
+    UUID getCurrentToken();
```
devinrsmith, line 31: I don't think we need this on the interface; SessionImpl is OK.

java-client/session-dagger/src/test/java/io/deephaven/client/impl/TableServiceAsyncTest.java
```diff
@@ -79,22 +85,30 @@ private static TableHandle get(TableHandleFuture future)
         return future.getOrCancel(GETTIME);
     }
 
-    private static void checkSucceeded(TableHandle x) {
+    private void checkSucceeded(TableHandle x, int chainLength) {
         assertThat(x.isSuccessful()).isTrue();
+        try (final SafeCloseable ignored = getExecutionContext().open()) {
+            final Table result = getSession(session.getCurrentToken()).<Table>getExport(x.exportId().id()).get();
```
devinrsmith, line 91: I think we should capture the server-side session in `DeephavenSessionTestBase` field; you should have access to `SessionImpl` w/ return call of `io.deephaven.client.SessionSubcomponent#newSession`

engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -683,27 +683,63 @@ private void maybeCreateClass(String className, String code, String packageName,
...
+        JavaCompiler localCompiler;
+        if ((localCompiler = JAVA_COMPILER) == null) {
+            synchronized (QueryCompiler.class) {
+                if ((localCompiler = JAVA_COMPILER) == null) {
+                    JAVA_COMPILER = localCompiler = ToolProvider.getSystemJavaCompiler();
+                }
+            }
+        }
+        if (localCompiler == null) {
+            throw new RuntimeException("No Java compiler provided - are you using a JRE instead of a JDK?");
```
rcaudy, line 698:
```suggestion
            throw new UncheckedDeephavenException("No Java compiler provided - are you using a JRE instead of a JDK?");
```

engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -735,7 +771,7 @@ private void maybeCreateClassHelper(String fqClassName, String finalCode, String
      * @return a Pair of success, and the compiler output
      */
     private Pair<Boolean, String> tryCompile(File basePath, Collection<File> javaFiles) throws IOException {
-        final JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+        final JavaCompiler compiler = getJavaCompiler();
         if (compiler == null) {
             throw new RuntimeException("No Java compiler provided - are you using a JRE instead of a JDK?");
         }
```
rcaudy, line 777: This is no longer necessary.
```suggestion
```

engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -683,27 +683,63 @@ private void maybeCreateClass(String className, String code, String packageName,
...
+        // the java file manager is not thread safe
+        synchronized (QueryCompiler.class) {
+            result = compiler.getTask(compilerOutput,
+                    fileManager,
+                    null,
+                    compilerOptions,
+                    null,
+                    Collections.singletonList(new JavaSourceFromString(fqClassName, finalCode)))
+                    .call();
+        }
```
rcaudy, line 742: No way can we merge this unless you can convince me that compilation was already single-threaded somewhere under the hood (which I don't believe to be the case).

java-client/session/src/main/java/io/deephaven/client/impl/Session.java
rcaudy, line 31: Do we want to make this less obscure?

java-client/session-dagger/src/test/java/io/deephaven/client/impl/TableServiceAsyncTest.java
```diff
@@ -79,22 +85,30 @@ private static TableHandle get(TableHandleFuture future)
...
+            ExecutionContext.getContext().getQueryScope().putParam("ChainLength", chainLength);
+            final Table expected = TableTools.emptyTable(CHAIN_ROWS).update("Current = ii - 1 + ChainLength");
```
rcaudy, line 93: I don't understand this test.

rcaudy (changes requested):

engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
devinrsmith: I was going to make a similar comment.

nbauernfeind: It's definitely not synchronized under the hood - go-client tests caught this via some concurrent modification exceptions.

engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -697,13 +697,29 @@ private void maybeCreateClassHelper(String fqClassName, String finalCode, String
 
         final StandardJavaFileManager fileManager = compiler.getStandardFileManager(null, null, null);
```
rcaudy, line 698: Should be `JavaFileManager`. No need to change in this PR.

engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -697,13 +697,29 @@ private void maybeCreateClassHelper(String fqClassName, String finalCode, String
...
+        boolean exceptionThrown = false;
+        try {
+            result = compiler.getTask(compilerOutput,
+                    fileManager,
+                    null,
+                    compilerOptions,
+                    null,
+                    Collections.singletonList(new JavaSourceFromString(fqClassName, finalCode)))
+                    .call();
+        } catch (final Exception err) {
```
rcaudy, line 710: Would you really want to suppress an `Error`? Probably better to catch `Throwable`, given what's happening here.

engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -683,62 +683,42 @@ private void maybeCreateClass(String className, String code, String packageName,
...
-    }
-
     private void maybeCreateClassHelper(String fqClassName, String finalCode, String[] splitPackageName,
             String rootPathAsString, String tempDirAsString) {
         final StringWriter compilerOutput = new StringWriter();
 
-        final JavaCompiler compiler = getJavaCompiler();
+        final JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+        if (compiler == null) {
+            throw new RuntimeException("No Java compiler provided - are you using a JRE instead of a JDK?");
```
rcaudy, line 692: Still better as `UncheckedDeephavenException`.

engine/context/src/main/java/io/deephaven/engine/context/QueryCompiler.java
```diff
@@ -771,7 +751,7 @@ private void maybeCreateClassHelper(String fqClassName, String finalCode, String
      * @return a Pair of success, and the compiler output
      */
     private Pair<Boolean, String> tryCompile(File basePath, Collection<File> javaFiles) throws IOException {
-        final JavaCompiler compiler = getJavaCompiler();
+        final JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
         if (compiler == null) {
             throw new RuntimeException("No Java compiler provided - are you using a JRE instead of a JDK?");
```
rcaudy, line 756: Here, too... UDE.

rcaudy (commented):
Only the comment regarding suppressing Errors is really important/actionable.

rcaudy (approved):

rcaudy (approved):

devinrsmith (approved):

PR #4813, opened 2023-11-10
Add TableDefinition column name helpers

devinrsmith:
Adds some helper methods for on `TableDefinition#checkHasColumn`, `TableDefinition#checkHasColumns`, and `TableDefinition#getColumnNameSet`.

Additionally, fixes up call sites that were (ab)using `Table#getColumnSourceMap` to simply get the keySet. This invokes a potentially extraneous `Table#coalesce` which can be avoided in these cases.

In support of common scaffolding so https://github.com/deephaven/deephaven-core/pull/4771 won't need to call `Table#getColumnSource` for validation purposes.

### DIFF
--- a/engine/api/src/main/java/io/deephaven/engine/table/TableDefinition.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/TableDefinition.java
@@ -305,12 +305,25 @@ public class TableDefinition implements LogOutputAppendable {
     }
 
     /**
+     * Check this definition to ensure that {@code columnName} is present.
+     *
+     * @param columnName The column name to check
+     * @throws NoSuchColumnException If {@code columnName} is missing
+     */
+    public final void checkColumn(@NotNull String columnName) {
+        final Set<String> columnNames = getColumnNameSet();
+        if (!columnNames.contains(columnName)) {
+            throw new NoSuchColumnException(columnNames, columnName);
+        }
+    }
+
+    /**
      * Check this definition to ensure that all {@code columns} are present.
      *
      * @param columns The column names to check
-     * @throws NoSuchColumnException If any columns were missing
+     * @throws NoSuchColumnException If any {@code columns} were missing
      */
-    public final void checkAvailableColumns(@NotNull Collection<String> columns) {
+    public final void checkColumns(@NotNull Collection<String> columns) {
         final Set<String> columnNames = getColumnNameSet();
         final List<String> missingColumns = columns
                 .stream()
--- a/engine/api/src/main/java/io/deephaven/engine/table/TableDefinition.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/TableDefinition.java
@@ -310,7 +310,7 @@ public class TableDefinition implements LogOutputAppendable {
      * @param columnName The column name to check
      * @throws NoSuchColumnException If {@code columnName} is missing
      */
-    public final void checkColumn(@NotNull String columnName) {
+    public final void checkHasColumn(@NotNull String columnName) {
         final Set<String> columnNames = getColumnNameSet();
         if (!columnNames.contains(columnName)) {
             throw new NoSuchColumnException(columnNames, columnName);
@@ -323,7 +323,7 @@ public class TableDefinition implements LogOutputAppendable {
      * @param columns The column names to check
      * @throws NoSuchColumnException If any {@code columns} were missing
      */
-    public final void checkColumns(@NotNull Collection<String> columns) {
+    public final void checkHasColumns(@NotNull Collection<String> columns) {
         final Set<String> columnNames = getColumnNameSet();
         final List<String> missingColumns = columns
                 .stream()
--- a/engine/api/src/main/java/io/deephaven/engine/table/TableDefinition.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/TableDefinition.java
@@ -17,7 +17,6 @@ import org.jetbrains.annotations.NotNull;
 import java.util.Map.Entry;
 import java.util.*;
 import java.util.function.Function;
-import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
@@ -311,10 +310,7 @@ public class TableDefinition implements LogOutputAppendable {
      * @throws NoSuchColumnException If {@code columnName} is missing
      */
     public final void checkHasColumn(@NotNull String columnName) {
-        final Set<String> columnNames = getColumnNameSet();
-        if (!columnNames.contains(columnName)) {
-            throw new NoSuchColumnException(columnNames, columnName);
-        }
+        NoSuchColumnException.throwIf(getColumnNameSet(), columnName);
     }
 
     /**
@@ -324,14 +320,7 @@ public class TableDefinition implements LogOutputAppendable {
      * @throws NoSuchColumnException If any {@code columns} were missing
      */
     public final void checkHasColumns(@NotNull Collection<String> columns) {
-        final Set<String> columnNames = getColumnNameSet();
-        final List<String> missingColumns = columns
-                .stream()
-                .filter(Predicate.not(columnNames::contains))
-                .collect(Collectors.toList());
-        if (!missingColumns.isEmpty()) {
-            throw new NoSuchColumnException(columnNames, missingColumns);
-        }
+        NoSuchColumnException.throwIf(getColumnNameSet(), columns);
     }
 
     /**
--- a/engine/api/src/main/java/io/deephaven/engine/table/TableDefinition.java
+++ b/engine/api/src/main/java/io/deephaven/engine/table/TableDefinition.java
@@ -9,6 +9,7 @@ import io.deephaven.base.cache.OpenAddressedCanonicalizationCache;
 import io.deephaven.base.log.LogOutput;
 import io.deephaven.base.log.LogOutputAppendable;
 import io.deephaven.base.verify.Assert;
+import io.deephaven.engine.table.impl.NoSuchColumnException;
 import io.deephaven.io.log.impl.LogOutputStringImpl;
 import io.deephaven.qst.column.header.ColumnHeader;
 import org.jetbrains.annotations.NotNull;
@@ -16,6 +17,7 @@ import org.jetbrains.annotations.NotNull;
 import java.util.Map.Entry;
 import java.util.*;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
@@ -205,6 +207,13 @@ public class TableDefinition implements LogOutputAppendable {
     }
 
     /**
+     * @return An unmodifiable set of column names
+     */
+    public Set<String> getColumnNameSet() {
+        return getColumnNameMap().keySet();
+    }
+
+    /**
      * @return A list of {@link ColumnDefinition column definitions} for all
      *         {@link ColumnDefinition.ColumnType#Partitioning partitioning} columns in the same relative order as the
      *         column definitions list
@@ -293,6 +302,23 @@ public class TableDefinition implements LogOutputAppendable {
     @SuppressWarnings("unused")
     public String getColumnNamesAsString() {
         return getColumnStream().map(ColumnDefinition::getName).collect(Collectors.joining(","));
+    }
+
+    /**
+     * Check this definition to ensure that all {@code columns} are present.
+     *
+     * @param columns The column names to check
+     * @throws NoSuchColumnException If any columns were missing
+     */
+    public final void checkAvailableColumns(@NotNull Collection<String> columns) {
+        final Set<String> columnNames = getColumnNameSet();
+        final List<String> missingColumns = columns
+                .stream()
+                .filter(Predicate.not(columnNames::contains))
+                .collect(Collectors.toList());
+        if (!missingColumns.isEmpty()) {
+            throw new NoSuchColumnException(columnNames, missingColumns);
+        }
     }
 
     /**
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/CrossJoinHelper.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/CrossJoinHelper.java
@@ -195,7 +195,7 @@ public class CrossJoinHelper {
                 jsm.startTrackingPrevValues();
                 final ModifiedColumnSet.Transformer leftTransformer = leftTable.newModifiedColumnSetTransformer(
                         resultTable,
-                        leftTable.getDefinition().getColumnNameSet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY));
+                        leftTable.getDefinition().getColumnNamesArray());
 
                 leftTable.addUpdateListener(new BaseTable.ListenerImpl(bucketingContext.listenerDescription,
                         leftTable, resultTable) {
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/CrossJoinHelper.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/CrossJoinHelper.java
@@ -195,7 +195,7 @@ public class CrossJoinHelper {
                 jsm.startTrackingPrevValues();
                 final ModifiedColumnSet.Transformer leftTransformer = leftTable.newModifiedColumnSetTransformer(
                         resultTable,
-                        leftTable.getColumnSourceMap().keySet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY));
+                        leftTable.getDefinition().getColumnNameSet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY));
 
                 leftTable.addUpdateListener(new BaseTable.ListenerImpl(bucketingContext.listenerDescription,
                         leftTable, resultTable) {
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
@@ -1787,7 +1787,7 @@ public class QueryTable extends BaseTable<QueryTable> {
 
                             copyAttributes(resultTable, CopyAttributeOperation.DropColumns);
                             copySortableColumns(resultTable,
-                                    resultTable.getDefinition().getColumnNameMap()::containsKey);
+                                    resultTable.getDefinition().getColumnNameSet()::contains);
                             maybeCopyColumnDescriptions(resultTable);
 
                             if (snapshotControl != null) {
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
@@ -1772,7 +1772,7 @@ public class QueryTable extends BaseTable<QueryTable> {
             return memoizeResult(MemoizedOperationKey.dropColumns(columnNames), () -> QueryPerformanceRecorder
                     .withNugget("dropColumns(" + Arrays.toString(columnNames) + ")", sizeForInstrumentation(), () -> {
                         final Mutable<Table> result = new MutableObject<>();
-                        definition.checkColumns(Arrays.asList(columnNames));
+                        definition.checkHasColumns(Arrays.asList(columnNames));
                         final Map<String, ColumnSource<?>> newColumns = new LinkedHashMap<>(columns);
                         for (String columnName : columnNames) {
                             newColumns.remove(columnName);
@@ -1793,8 +1793,7 @@ public class QueryTable extends BaseTable<QueryTable> {
                             if (snapshotControl != null) {
                                 final ModifiedColumnSet.Transformer mcsTransformer =
                                         newModifiedColumnSetTransformer(resultTable,
-                                                resultTable.getDefinition().getColumnNameSet()
-                                                        .toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY));
+                                                resultTable.getDefinition().getColumnNamesArray());
                                 final ListenerImpl listener = new ListenerImpl(
                                         "dropColumns(" + Arrays.deepToString(columnNames) + ')', this, resultTable) {
                                     @Override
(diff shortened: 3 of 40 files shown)


engine/api/src/main/java/io/deephaven/engine/table/TableDefinition.java
```diff
@@ -295,6 +304,36 @@ public String getColumnNamesAsString() {
         return getColumnStream().map(ColumnDefinition::getName).collect(Collectors.joining(","));
     }
 
+    /**
+     * Check this definition to ensure that {@code columnName} is present.
+     *
+     * @param columnName The column name to check
+     * @throws NoSuchColumnException If {@code columnName} is missing
+     */
+    public final void checkColumn(@NotNull String columnName) {
```
rcaudy, line 313:
```suggestion
    public final void checkHasColumn(@NotNull String columnName) {
```

engine/api/src/main/java/io/deephaven/engine/table/TableDefinition.java
```diff
@@ -295,6 +304,36 @@ public String getColumnNamesAsString() {
...
+        }
+    }
+
+    /**
+     * Check this definition to ensure that all {@code columns} are present.
+     *
+     * @param columns The column names to check
+     * @throws NoSuchColumnException If any {@code columns} were missing
+     */
+    public final void checkColumns(@NotNull Collection<String> columns) {
```
rcaudy, line 326:
```suggestion
    public final void checkHasColumns(@NotNull Collection<String> columns) {
```

engine/table/src/main/java/io/deephaven/engine/table/impl/CrossJoinHelper.java
```diff
@@ -195,7 +195,7 @@ private static QueryTable internalJoin(
                 jsm.startTrackingPrevValues();
                 final ModifiedColumnSet.Transformer leftTransformer = leftTable.newModifiedColumnSetTransformer(
                         resultTable,
-                        leftTable.getColumnSourceMap().keySet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY));
+                        leftTable.getDefinition().getColumnNameSet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY));
```
rcaudy, line 198:
```suggestion
                        leftTable.getDefinition().getColumnNamesArray());
```

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -1800,7 +1793,7 @@ public Table dropColumns(String... columnNames) {
                             if (snapshotControl != null) {
                                 final ModifiedColumnSet.Transformer mcsTransformer =
                                         newModifiedColumnSetTransformer(resultTable,
-                                                resultTable.getColumnSourceMap().keySet()
+                                                resultTable.getDefinition().getColumnNameSet()
```
rcaudy, line 1796: Use `io.deephaven.engine.table.TableDefinition#getColumnNamesArray`

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -2400,7 +2393,7 @@ private Table snapshotIncrementalInternal(final Table base, final boolean doInit
 
         // Use the given columns (if specified); otherwise an empty array means all of my columns
         final String[] useStampColumns = stampColumns.length == 0
-                ? getColumnSourceMap().keySet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY)
+                ? definition.getColumnNameSet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY)
```
rcaudy, line 2396:
```suggestion
                ? definition. getColumnNamesArray()
```

engine/table/src/main/java/io/deephaven/engine/table/impl/TableUpdateValidator.java
```diff
@@ -61,11 +61,14 @@ private TableUpdateValidator(final String description, final QueryTable tableToV
         this.description = description == null ? tableToValidate.getDescription() : description;
         this.tableToValidate = tableToValidate;
         this.validationMCS = tableToValidate.newModifiedColumnSet(
-                tableToValidate.getColumnSourceMap().keySet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY));
+                tableToValidate.getDefinition().getColumnStream().map(ColumnDefinition::getName)
+                        .toArray(String[]::new));
```
rcaudy, line 65:
```suggestion
                tableToValidate.getDefinition(). getColumnNamesArray());
```

engine/table/src/main/java/io/deephaven/engine/table/impl/by/ChunkedOperatorAggregationHelper.java
```diff
@@ -86,34 +86,14 @@ public static QueryTable aggregation(
...
-                    .filter(Predicate.not(colNames::contains))
-                    .toArray(String[]::new);;
-
-            throw new IllegalArgumentException("aggregation: not all group-by columns " + Arrays.toString(keyNames)
-                    + " are present in input table with columns "
-                    + Arrays.toString(input.getDefinition().getColumnNamesArray()) + ". Missing columns: "
-                    + Arrays.toString(missingColumns));
-        }
+        final List<String> keyNames = groupByColumns.stream().map(ColumnName::name).collect(Collectors.toList());
+        input.checkAvailableColumns(keyNames);
```
rcaudy, line 90: Maybe we can create a constructor for `NoSuchColumnException` that lets us specify part of the message? We've lost some valuable content with this change.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/ChunkedOperatorAggregationHelper.java
```diff
@@ -86,34 +86,14 @@ public static QueryTable aggregation(
...
-            final String[] missingColumns = Arrays.stream(keyNames)
-                    .filter(Predicate.not(colNames::contains))
-                    .toArray(String[]::new);;
-
-            throw new IllegalArgumentException("aggregation: not all group-by columns " + Arrays.toString(keyNames)
-                    + " are present in input table with columns "
-                    + Arrays.toString(input.getDefinition().getColumnNamesArray()) + ". Missing columns: "
-                    + Arrays.toString(missingColumns));
-        }
+        final List<String> keyNames = groupByColumns.stream().map(ColumnName::name).collect(Collectors.toList());
```
rcaudy, line 89: We were probably better off with an array that can be wrapped as a list in O(1) storage, rather than a list that can only be converted to a list in O(n) storage... not a big deal, but not an improvement.

engine/table/src/main/java/io/deephaven/engine/table/impl/by/ChunkedOperatorAggregationHelper.java
```diff
@@ -86,34 +86,14 @@ public static QueryTable aggregation(
...
-                final String[] missingColumns = Arrays.stream(keyNames)
-                        .filter(Predicate.not(colNames::contains))
-                        .toArray(String[]::new);;
-
-                throw new IllegalArgumentException("aggregation: not all group-by columns " + Arrays.toString(keyNames)
-                        + " are present in initial groups table with columns "
-                        + Arrays.toString(initialKeys.getDefinition().getColumnNamesArray()) + ". Missing columns: "
-                        + Arrays.toString(missingColumns));
-            }
+            initialKeys.getDefinition().checkColumns(keyNames);
```
rcaudy, line 96: Same comment regarding specialization of the error message. Using `checkAvailableColumns` is less user-friendly, especially when there are two inputs to be concerned with.

engine/table/src/main/java/io/deephaven/engine/table/impl/remote/ConstructSnapshot.java
```diff
@@ -1480,8 +1479,8 @@ private static boolean serializeAllTable(
             snapshot.rowsIncluded = snapshot.rowsAdded.copy();
         }
 
-        final Map<String, ? extends ColumnSource<?>> sourceMap = table.getColumnSourceMap();
-        final String[] columnSources = sourceMap.keySet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY);
+        final String[] columnSources =
+                table.getDefinition().getColumnNameSet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY);
```
rcaudy, line 1483:
```suggestion
        final String[] columnSources = table.getDefinition().getColumnNamesArray();
```

engine/table/src/main/java/io/deephaven/engine/table/impl/remote/ConstructSnapshot.java
```diff
@@ -1403,9 +1403,8 @@ private static boolean serializeAllTable(
         }
 
         LongSizedDataStructure.intSize("construct snapshot", snapshot.rowsIncluded.size());
-
-        final Map<String, ? extends ColumnSource<?>> sourceMap = table.getColumnSourceMap();
-        final String[] columnSources = sourceMap.keySet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY);
+        final String[] columnSources =
+                table.getDefinition().getColumnNameSet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY);
```
rcaudy, line 1407:
```suggestion
        final String[] columnSources = table.getDefinition(). getColumnNamesArray();
```

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateBy.java
```diff
@@ -1198,7 +1199,10 @@ public static Table updateBy(@NotNull final QueryTable source,
 
         final MutableObject<String> timestampColumnName = new MutableObject<>(null);
         // create an initial set of all source columns
-        final Set<String> preservedColumnSet = new LinkedHashSet<>(source.getColumnSourceMap().keySet());
+        final LinkedHashSet<String> preservedColumnSet = source.getDefinition()
```
rcaudy, line 1202: Why not just use the map constructor that takes a collection, with `getColumnNameSet()`? I guess we don't have to be populating the name map...

engine/table/src/main/java/io/deephaven/engine/util/TotalsTableBuilder.java
```diff
@@ -546,12 +546,7 @@ public static Table makeTotalsTable(Table source, TotalsTableBuilder builder, St
     }
 
     private static void ensureColumnsExist(Table source, Set<String> columns) {
-        if (!source.getColumnSourceMap().keySet().containsAll(columns)) {
-            final Set<String> missing = new LinkedHashSet<>(columns);
-            missing.removeAll(source.getColumnSourceMap().keySet());
-            throw new IllegalArgumentException("Missing columns for totals table " + missing + ", available columns "
-                    + source.getColumnSourceMap().keySet());
-        }
+        source.getDefinition().checkColumns(columns);
```
rcaudy, line 549: Another case for specialization?

engine/table/src/main/java/io/deephaven/engine/util/WindowCheck.java
```diff
@@ -195,7 +195,7 @@ public int getPos(Entry el) {
             this.rowKeyToEntry = new TLongObjectHashMap<>();
 
             this.mcsTransformer = source.newModifiedColumnSetTransformer(result,
-                    source.getColumnSourceMap().keySet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY));
+                    source.getDefinition().getColumnNameSet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY));
```
rcaudy, line 198:
```suggestion
                    source.getDefinition().getColumnNamesArray());
```

engine/table/src/test/java/io/deephaven/engine/table/impl/QueryTableAggregationTest.java
```diff
@@ -887,7 +887,7 @@ public void testKeyColumnTypes() {
                         new BigDecimalGenerator(),
                         new IntGenerator()));
 
-        final Set<String> keyColumnSet = new LinkedHashSet<>(table.getColumnSourceMap().keySet());
+        final Set<String> keyColumnSet = new LinkedHashSet<>(table.getDefinition().getColumnNames());
```
rcaudy, line 890: I raised a discussion around a similar bit of code earlier. In that case, you chose to stream into a map collector. Here, you're using a newly-created list. Maybe both are better like this one, or using a `getColumnNameSet()` as the input.

engine/table/src/test/java/io/deephaven/engine/table/impl/SelectOverheadLimiter.java
```diff
@@ -131,7 +131,7 @@ public static Table clampSelectOverhead(Table input, double permittedOverhead) {
             {
                 inputRecorder.getValue().setMergedListener(this);
                 inputTransformer = ((QueryTable) input).newModifiedColumnSetTransformer(result,
-                        result.getColumnSourceMap().keySet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY));
+                        result.getDefinition().getColumnNameSet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY));
```
rcaudy, line 134:
```suggestion
                        result.getDefinition().getColumnNamesArray());
```

engine/table/src/test/java/io/deephaven/engine/table/impl/SelectOverheadLimiter.java
```diff
@@ -171,7 +171,7 @@ protected void process() {
                         new ListenerRecorder("clampSelectOverhead.flatResult()", flatResult, result);
                 flatRecorder.setMergedListener(this);
                 flatTransformer = ((QueryTable) flatResult).newModifiedColumnSetTransformer(result,
-                        result.getColumnSourceMap().keySet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY));
+                        result.getDefinition().getColumnNameSet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY));
```
rcaudy, line 174:
```suggestion
                        result.getDefinition().getColumnNamesArray());
```

engine/table/src/test/java/io/deephaven/engine/table/impl/TestAggBy.java
```diff
@@ -818,7 +818,7 @@ public void testAggAllByWithFormatColumn() {
         assertEquals(2.0, cs.get(0));
 
         result = dataTable.formatColumns("Doubles=Decimal(`##0.00%`)").headBy(1);
-        Set<String> columnNames = result.getColumnSourceMap().keySet();
+        List<String> columnNames = result.getDefinition().getColumnNames();
```
rcaudy, line 821: Why not use your new set method?

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -1378,7 +1378,7 @@ private void assertTableStatistics(Table inputTable, File dest) {
         final ParquetMetadata metadata = new ParquetTableLocationKey(dest, 0, null).getMetadata();
 
         final String[] colNames =
-                inputTable.getColumnSourceMap().keySet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY);
+                inputTable.getDefinition().getColumnNameSet().toArray(CollectionUtil.ZERO_LENGTH_STRING_ARRAY);
```
rcaudy, line 1381:
```suggestion
                inputTable.getDefinition().getColumnNamesArray();
```

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -460,7 +460,7 @@ public void testVectorColumns() {
         writeReadTableTest(vectorTable, dest);
 
         // Convert the table from vector to array column
-        final Table arrayTable = vectorTable.updateView(vectorTable.getColumnSourceMap().keySet().stream()
+        final Table arrayTable = vectorTable.updateView(vectorTable.getDefinition().getColumnNameSet().stream()
```
rcaudy, line 463: Better with column stream?

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateBy.java
devinrsmith: Yeah, this saves an little bit of extra allocation. Happy to take the less-efficient but prettier approach if you prefer. I don't think we need modifiable results often enough to bake it into TableDefinition, but I could be swayed that way too if we like...

devinrsmith: I'll take the prettier approach for now; I'm usually the one to argue for maintainability over perceived efficiencies.

engine/table/src/test/java/io/deephaven/engine/table/impl/TestAggBy.java
devinrsmith: test impl only checks size and iterates through it - no need for Set afaict.

rcaudy: Sure. Just a question of an on-demand allocation vs. a cached one.

devinrsmith: Ah, I see the "concern"; this brings up the larger question of do we want / need to document the APIs in terms of whether they are cached or not / implementation details? I don't necessarily want to box us into defining `TableDefinition#getColumnNames()` as on-demand, `TableDefinition#getColumnNamesSet` as cached.

devinrsmith: It's not unreasonable IMO that we could make the case that TableDefinition impl should go from

```java
    private final List<ColumnDefinition<?>> columns;
    private Map<String, ColumnDefinition<?>> columnNameMap;
```

to

```java
    private final Map<String, ColumnDefinition<?>> columnNameMap;
```

. Majority of work is probably auditing usage of `io.deephaven.engine.table.TableDefinition#getColumns()` to see who actually _needs_ a List vs a Collection (which could be achieved efficiently by `columnNameMap.values()`).

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateBy.java
```diff
@@ -1198,7 +1199,7 @@ public static Table updateBy(@NotNull final QueryTable source,
 
         final MutableObject<String> timestampColumnName = new MutableObject<>(null);
         // create an initial set of all source columns
-        final Set<String> preservedColumnSet = new LinkedHashSet<>(source.getColumnSourceMap().keySet());
+        final LinkedHashSet<String> preservedColumnSet = new LinkedHashSet<>(source.getDefinition().getColumnNames());
```
rcaudy, line 1202: Which is the right standard to follow here? We have at least one instance of "the old way", and a mix of list vs. set.
[](url)
Screenshot 2023-11-13 at 7 35 58 PM

devinrsmith: The FormulaColumnPython example isn't the same interface. I'm happy to standardize on `getColumnNameSet` in these cases.

I did find another class of maps that could be optimized to sets as well.

engine/test-utils/src/main/java/io/deephaven/engine/testutil/TstUtils.java
```diff
@@ -173,11 +175,12 @@ public static void addToTable(final Table table, final RowSet rowSet, final Colu
             }
         }
 
-        if (!usedNames.containsAll(table.getDefinition().getColumnNameSet())) {
-            final Set<String> expected = new LinkedHashSet<>(table.getDefinition().getColumnNameSet());
-            expected.removeAll(usedNames);
-            throw new IllegalStateException("Not all columns were populated, missing " + expected);
-        }
+        NoSuchColumnException.throwIf(
+                table.getDefinition().getColumnNameSet(),
```
rcaudy, line 179: I think this is backwards, and I think the description requires revision. This complains if we didn't use all of the table columns, so you want available to be `usedColumns` and requested to be the definition's columns.

engine/table/src/test/java/io/deephaven/engine/table/impl/QueryTableAggregationTest.java
```diff
@@ -116,12 +116,12 @@ private static Table individualStaticByTest(@NotNull final Table input,
...
             if (keyColumns.length == 0) {
-                expectedKeys = TableTools.emptyTable(adjustedInput.size() > 0 ? 1 : 0);
+                expectedKeys = TableTools.emptyTable(!adjustedInput.isEmpty() ? 1 : 0);
                 expected = adjustedInput;
             } else {
                 final Set<String> retainedColumns =
-                        new LinkedHashSet<>(adjustedInput.getDefinition().getColumnNameMap().keySet());
-                retainedColumns.removeAll(Arrays.stream(keyNames).collect(Collectors.toSet()));
+                        new LinkedHashSet<>(adjustedInput.getDefinition().getColumnNameSet());
+                Arrays.asList(keyNames).forEach(retainedColumns::remove);
```
rcaudy, line 124: Isn't this better?
```suggestion
                retainedColumns.removeAll(Arrays.asList(keyNames));
```

devinrsmith: So, IJ warned me and apparently the semantics (as explicitly called out in) `java.util.AbstractSet#removeAll` is *BAD* because it may end up doing contains call against the RHS. It was a big oversight in java development, and they've been unable to fix this.

If RHS is a set `AbstractSet.removeAll` is _ok_.

engine/test-utils/src/main/java/io/deephaven/engine/testutil/TstUtils.java
devinrsmith: I think you are correct; I had this backwards (tests still pass b/c they actually need to be equal). That said, I think the error message is still correct.

engine/table/src/test/java/io/deephaven/engine/table/impl/TestAggBy.java
rcaudy: I suspect you're right that the vast majority of the time an ordered collection is sufficient. No need to scope creep this PR for that.

rcaudy (approved):

PR #4815, opened 2023-11-12
change default dtype_backend for to_pandas

### LINKED ISSUE #3149: to_pandas function transforms null/None values in a String column into "None" string
**Description**

Instead of keeping values as null/None, the to_pandas function creates a "None" string.

**Steps to reproduce**
Run this script
```
from deephaven.pandas import to_pandas

dh_table = empty_table(10).update("I=(String)null")
dataframe = to_pandas(dh_table)
```

**Expected results**

The visible dataframe should show `null` in italics to represent the absence of a value instead of the text None. 

**Actual results**

The visible dataframe shows None.

**Versions**
 - Deephaven: 0.18+
 - OS: Linux
 - Browser: N/A
 - Docker: N/A

### LINKED ISSUE #4810: deephaven.pandas.to_pandas does not properly convert string columns
A user reported problems with doing round trips between DH and pandas when column types are strings.  I have confirmed that DH is not properly setting string column types in pandas tables.  See below:

```python

a=empty_table(5).update_view('A=`value`')

meta1 = a.meta_table # java.lang.String

 
df = pd.DataFrame({'a':['a', 'b', 'c']})

print(df.dtypes)

 

from deephaven.pandas import to_table

t = to_table(df)

meta2 = t.meta_table # PyObject


# See https://pandas.pydata.org/docs/user_guide/text.html

import pandas as pd
s_a = pd.Series(["a", "b", "c"])
s_b = pd.Series(["a", "b", "c"], dtype="string")
print(s_a)
print(s_b)
df2 = pd.DataFrame({'a':s_a, 'b':s_b})
print(df2.dtypes)
t2 = to_table(df2)
meta3 = t2.meta_table

# Now look at the original case with a twist
df['b'] = df['a'].astype("string")

print(df.dtypes)
t3 = to_table(df)
meta4 = t3.meta_table

# Test DH to_pandas conversion
from deephaven.pandas import to_pandas
df_a = to_pandas(a)
print(df_a.dtypes)
```

jmao-denver:
Fixes #4810
Fixes #3149 

This is a breaking change:
1. the new default value for `dtype_backend` is `numpy_nullable`, which directs the conversion to automatically use Pandas Nullable Extension Types.
2. when `dtype_backend` is `numpy_nullable`, the conversion exports the table data into a pyarrow table as an intermediate step and then converts the pyarrow table to a Pandas Dataframe. This is a more efficient and deterministic way to convert DH data and DH nulls, however DH server still has some gaps in mapping between DH data types and Arrow types (e.g. a column of PyObject will be converted to strings, java.time.LocalDate/LocalTime to bytes). This isn't yet a problem because the unsupported data types aren't what users normally would use and there is no complaints from them so far.

### DIFF
--- a/py/server/deephaven/pandas.py
+++ b/py/server/deephaven/pandas.py
@@ -112,7 +112,8 @@ _PYARROW_TO_PANDAS_TYPE_MAPPERS = {
 }
 
 
-def to_pandas(table: Table, cols: List[str] = None, dtype_backend: Literal[None, "pyarrow", "numpy_nullable"] = None,
+def to_pandas(table: Table, cols: List[str] = None,
+              dtype_backend: Literal[None, "pyarrow", "numpy_nullable"] = "numpy_nullable",
               conv_null: bool = True) -> pd.DataFrame:
     """Produces a pandas DataFrame from a table.
 
@@ -125,8 +126,8 @@ def to_pandas(table: Table, cols: List[str] = None, dtype_backend: Literal[None,
         cols (List[str]): the source column names, default is None which means include all columns
         dtype_backend (str): Which dtype_backend to use, e.g. whether a DataFrame should have NumPy arrays,
             nullable dtypes are used for all dtypes that have a nullable implementation when “numpy_nullable” is set,
-            pyarrow is used for all dtypes if “pyarrow” is set. default is None, meaning Numpy backed DataFrames with
-            no nullable dtypes.
+            pyarrow is used for all dtypes if “pyarrow” is set. None means Numpy backed DataFrames with no nullable
+            dtypes. Default is "numpy_nullable".
         conv_null (bool): When dtype_backend is not set, whether to check for Deephaven nulls in the data and
             automatically replace them with pd.NA. default is True.
 
--- a/py/server/deephaven/pandas.py
+++ b/py/server/deephaven/pandas.py
@@ -127,8 +127,9 @@ def to_pandas(table: Table, cols: List[str] = None,
         dtype_backend (str): which dtype_backend to use, e.g. whether a DataFrame should have NumPy arrays,
             nullable dtypes are used for all dtypes that have a nullable implementation when “numpy_nullable” is set,
             pyarrow is used for all dtypes if “pyarrow” is set. None means Numpy backed DataFrames with no nullable
-            dtypes. Both "numpy_nullable" and "pyarrow" automatically converts Deephaven nulls and enable the correct
-            mapping of Java String. default is "numpy_nullable".
+            dtypes. Both "numpy_nullable" and "pyarrow" automatically convert Deephaven nulls to Pandas NA and enable
+            Pandas extension types.  Extension types are needed to support types beyond NumPy's type system.  Extension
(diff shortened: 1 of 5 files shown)


py/server/deephaven/pandas.py
```diff
@@ -125,8 +126,8 @@ def to_pandas(table: Table, cols: List[str] = None, dtype_backend: Literal[None,
         cols (List[str]): the source column names, default is None which means include all columns
         dtype_backend (str): Which dtype_backend to use, e.g. whether a DataFrame should have NumPy arrays,
             nullable dtypes are used for all dtypes that have a nullable implementation when “numpy_nullable” is set,
-            pyarrow is used for all dtypes if “pyarrow” is set. default is None, meaning Numpy backed DataFrames with
-            no nullable dtypes.
+            pyarrow is used for all dtypes if “pyarrow” is set. None means Numpy backed DataFrames with no nullable
+            dtypes. Default is "numpy_nullable".
```
chipkent, line 130: Docstring does not explain that "numpy_nullable" also allows support for string types, and whatever other voodoo can happen.

py/server/tests/test_parquet.py
```diff
@@ -347,7 +347,7 @@ def test_dates_and_time(self):
         from_disk = read('data_from_dh.parquet')
         self.assert_table_equals(dh_table, from_disk)
 
-        df_from_disk = to_pandas(from_disk)
+        df_from_disk = to_pandas(from_disk, dtype_backend=None)
         if pandas.__version__.split('.')[0] == "1":
             df_from_pandas = pandas.read_parquet("data_from_dh.parquet", use_nullable_dtypes=True)
         else:
```
chipkent, line 353: This looks suspect.  We are doing our on data treatment with `dtype_backend=None`, and then we read with `use_nullable_dtypes=True`.

py/server/tests/test_parquet.py
```diff
@@ -384,7 +384,7 @@ def time_test_helper(pa_table, new_schema, dest):
             # Write the provided pyarrow table type-casted to the new schema
             pyarrow.parquet.write_table(pa_table.cast(new_schema), dest)
             from_disk = read(dest)
-            df_from_disk = to_pandas(from_disk)
+            df_from_disk = to_pandas(from_disk, dtype_backend=None)
             original_df = pa_table.to_pandas()
             # Compare the dataframes as strings
             self.assertTrue((df_from_disk.astype(str) == original_df.astype(str)).all().values.all())
```
chipkent, line 390: Isn't this `astype(str)` stuff indicating a problem in the test or library?  If the new default backend is used, shouldn't this be unneeded?

jmao-denver: See the response above.

py/server/tests/test_parquet.py
jmao-denver: Debugged a bit with @malhotrashivam and we found an issue with `to_arrow`. https://github.com/deephaven/deephaven-core/issues/4823

chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3429
Reference: https://github.com/deephaven/deephaven.io/issues/3428

PR #4816, opened 2023-11-13
Always export scope objects before performing other interactions

### LINKED ISSUE #4604: Hierarchical tables not creating an export in JS API
**Description**

According to @niloc132 the JS API is not creating an export ticket for hierarchical tables. This can cause an unhandled promise rejection `Could not resolve 'view.target': no variable exists with name 'result'` if rows are expanded or there is a pending expansion request and the table has been deleted/closed.

Issue popped up in the docs snapshot tool and is likely an issue of a missing return statement there, but the error could be better. There should be an error about trying to expand rows on a closed table.

**Steps to reproduce**

This is what the docs snapshot tool does. It creates a race condition between expanding the table and deleting it in Python/Groovy.

1. Create a table
2. Run `del t` in Python
3. Expand a row on `t` in the JS API

**Expected results**

4. No error or error about trying to expand a row of a closed table (if `table.close()` called in JS API).

**Actual results**

4. Unhandled promise rejection `Could not resolve 'view.target': no variable exists with name 'result'` from the server.

**Versions**
 - Deephaven: 0.28.0

niloc132:
Also ensures that widget type strings are checked without regard to case, as the server will also do.

Fixes #4604

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
@@ -1420,8 +1420,7 @@ public class JsTable extends HasLifecycle implements HasTableBinding, JoinableTa
         typedTicket.setType(JsVariableType.PARTITIONEDTABLE);
         typedTicket.setTicket(partitionedTableTicket);
         Promise<JsPartitionedTable> fetchPromise = new JsWidget(workerConnection, typedTicket).refetch().then(
-                widget -> Promise.resolve(new JsPartitionedTable(workerConnection, widget))
-        );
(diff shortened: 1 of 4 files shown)


mofojed (approved):
Looks good to me, tested with my deephaven.ui branch: https://github.com/deephaven/deephaven-plugins/pull/88
Just a basic panel:
```
from deephaven import ui

@ui.component
def foo():
    return ui.flex([])

f = foo()
```
Opened and closed it, widget closed on panel close, then was able to re-open it without errors.

mattrunyon:
Also confirmed this removes the error when closing the widget for the plotly-express plugin

PR #4819, opened 2023-11-13
Moved Parquet specific methods out of DateTimeUtils

malhotrashivam:


extensions/parquet/table/src/main/java/io/deephaven/parquet/table/util/TransferUtils.java
```diff
@@ -0,0 +1,66 @@
...
+     * Returns nanoseconds from the Epoch for a {@link LocalDateTime} value in UTC timezone.
+     *
+     * @param localDateTime the local date time to compute the Epoch offset for
+     * @return nanoseconds since Epoch, or a NULL_LONG value if the local date time is null
+     */
+    public static long epochNanosUTC(@Nullable final LocalDateTime localDateTime) {
+        if (localDateTime == null) {
+            return QueryConstants.NULL_LONG;
+        }
+        return DateTimeUtils.epochNanos(localDateTime.atZone(ZoneId.of("UTC")));
```
malhotrashivam, line 25: Earlier this was done as
```        
return TimeUnit.SECONDS.toNanos(localDateTime.toEpochSecond(ZoneOffset.UTC))
                + localDateTime.toLocalTime().getNano();
```
Chip suggested the new approach.

rcaudy: No, this new approach is worse; it causes us to lookup the time zone (worse), allocate a ZonedDateTime (much worse), repeat the null check (slightly worse), etc.

The only argument in favor of using the new method is some overflow testing it does. I'm willing for us to do that overflow testing if we can make the methods or constants public. We could also just use `Math.addExact`.

Actually, we could use `DateTimeUtils.secondsToNanos()` to help; if we do that, we're only duplicating the null check. Making a suggestion accordingly.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/util/TransferUtils.java
```diff
@@ -0,0 +1,66 @@
...
```
rcaudy, line 25:
```suggestion
        return DateTimeUtils.secondsToNanos(localDateTime.toEpochSecond(ZoneOffset.UTC)) + localDateTime.toLocalTime().getNano();
```

rcaudy (approved):

PR #4826, opened 2023-11-14
Use chunks in DynamicWhereFilter, fix a chunk leak, and cleanup some code generally

rcaudy:


engine/table/src/main/java/io/deephaven/engine/table/impl/select/DynamicWhereFilter.java
```diff
@@ -70,46 +71,77 @@ public DynamicWhereFilter(final QueryTable setTable, final boolean inclusion, fi
         if (setRefreshing) {
             this.setTable = setTable;
             setTupleSource = TupleSourceFactory.makeTupleSource(setColumns);
-            setTable.getRowSet().forAllRowKeys((final long v) -> addKey(makeKey(v)));
+            try (final CloseableIterator<?> initialKeysIterator = ChunkedColumnIterator.make(
+                    setTupleSource, setTable.getRowSet())) {
```
devinrsmith, line 75: Should we be passing our own `CHUNK_SIZE`? (also, other call sites)

engine/table/src/main/java/io/deephaven/engine/table/impl/select/DynamicWhereFilter.java
```diff
@@ -70,46 +71,77 @@ public DynamicWhereFilter(final QueryTable setTable, final boolean inclusion, fi
...
+                    // Remove removed keys
+                    if (hasRemoves) {
+                        try (final CloseableIterator<?> removedKeysIterator = ChunkedColumnIterator.make(
+                                setTupleSource.getPrevSource(), upstream.removed())) {
+                            removedKeysIterator.forEachRemaining(DynamicWhereFilter.this::removeKey);
+                        }
+                    }
 
-                    upstream.added().forAllRowKeys((final long v) -> addKey(makeKey(v)));
-                    upstream.removed().forAllRowKeys((final long v) -> removeKey(makePrevKey(v)));
```
devinrsmith, line 90: I see the order changed from added, removed, modified to added, modified, removed. Just double checking, that's intentional (looks OK to do based on code).

engine/table/src/main/java/io/deephaven/engine/table/impl/select/DynamicWhereFilter.java
```diff
@@ -241,79 +264,56 @@ public WritableRowSet filter(
...
-            while (rsIt.hasMore()) {
-                final RowSequence chunkOk = rsIt.getNextRowSequenceWithLength(CHUNK_SIZE);
+            while (selectionIterator.hasMore()) {
+                final RowSequence selectionChunk = selectionIterator.getNextRowSequenceWithLength(maxChunkSize);
 
-                final Chunk<Values> chunk = Chunk.downcast(keyColumn.getChunk(getContext, chunkOk));
-                setInclusionKernel.matchValues(chunk, matches);
+                final Chunk<Values> keyChunk = Chunk.downcast(tupleSource.getChunk(keyGetContext, selectionChunk));
+                final int thisChunkSize = keyChunk.size();
+                setInclusionKernel.matchValues(keyChunk, matches);
```
devinrsmith, line 304: There isn't anything "special" about `SetInclusionKernel`; do we always know it will be an object chunk now?

There doesn't seem to be any value in copying into matches, to then just use it later as filter. I think we can simply check against `liveValues` as we iterate and build index.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/DynamicWhereFilter.java
```diff
@@ -241,79 +264,56 @@ public WritableRowSet filter(
...
-                keyIndices.setSize(chunk.size());
-                chunkOk.fillRowKeyChunk(keyIndices);
+                selectionRowKeyChunk.setSize(thisChunkSize);
+                selectionChunk.fillRowKeyChunk(selectionRowKeyChunk);
 
-                for (int ii = 0; ii < chunk.size(); ++ii) {
+                for (int ii = 0; ii < thisChunkSize; ++ii) {
                     if (matches.get(ii)) {
-                        indexBuilder.appendKey(keyIndices.get(ii));
+                        indexBuilder.appendKey(selectionRowKeyChunk.get(ii));
```
devinrsmith, line 311: Is there value in trying to use `appendRange` instead? I know we've got arbitrary filtering here, but it's possible there is underlying structure and so we have large contiguous inclusions?

More generally, it seems like a utility wrapper around RowSetBuilderSequential could be generally useful in this regards?

engine/table/src/main/java/io/deephaven/engine/table/impl/select/DynamicWhereFilter.java
rcaudy: It's actually removes, then mods, then adds. These orders are equally correct. Note that setTable always has distinct keys.

I changed it for two reasons:
1. This is a more canonical order.
2. This order keeps the hash table at its minimum possible size, which may avoid unnecessary expansions/rehashes/collisions that will be mooted by subsequent operations.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/DynamicWhereFilter.java
rcaudy: Because I combined the single column case and the tuple case, we do *not* know it will always be an object chunk. Set inclusion let's us do non-boxing inclusion tests. That is, it *is* special.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/DynamicWhereFilter.java
rcaudy: No. RowSetBuilderSequential should be optimizing range coalescing internally.

devinrsmith (approved):

Issue #871, opened 2021-07-20
Support repetition levels > 1 in Parquet files


rcaudy:
Currently, we regard nested repetition as uncommon and hard to map into a columnar data table like Deephaven's.
This feature request is intended to capture views to the contrary.

Linked to #294 , although intended for a later effort.

rcaudy:
Likely the first step is some kind of "flattening", but this is contrary to the intent of the Dremel design, so maybe we can think of a better solution.

rcaudy:
I'll be improving our error messages with a PR shortly. New messages:
For:
```
t = io.deephaven.db.tables.utils.ParquetTools.readTable("/data/parquetFiles/nonnullable_nested_v1_IMPALA_NULLS_NONE.parquet")
```
We'll see:
```
java.lang.UnsupportedOperationException: Unsupported maximum repetition level 2 in column int_array_array/list/element/list/element
```
For:
```
t = io.deephaven.db.tables.utils.ParquetTools.readTable("/data/parquetFiles/repeated_nested_RUST_NONE.parquet")
```
We'll see:
```
java.lang.UnsupportedOperationException: Encountered unsupported multi-column field phoneNumbers: found columns phoneNumbers/phone/number and phoneNumbers/phone/kind
```

devinrsmith:
It might be nice to be able to specify which columns you care about for your Table - in which case, the user can choose to not include the nested columns.

There's a mechanism right now to provide column instructions:

```python
from deephaven.parquet import read, ColumnInstruction

t = read(
    path="/snappy.parquet",
    col_instructions=[
        ColumnInstruction(column_name="date", parquet_column_name="date")
    ],
)
```

but this currently throws the error:

```
java.lang.UnsupportedOperationException: Encountered unsupported multi-column field outputs: found columns outputs/list/element/address and outputs/list/element/index
	at io.deephaven.parquet.table.ParquetSchemaReader.lambda$readParquetSchema$1(ParquetSchemaReader.java:174)
	at java.base/java.util.HashMap.compute(HashMap.java:1316)
	at io.deephaven.parquet.table.ParquetSchemaReader.readParquetSchema(ParquetSchemaReader.java:169)
	at io.deephaven.parquet.table.ParquetTools.convertSchema(ParquetTools.java:647)
	at io.deephaven.parquet.table.ParquetTools.readTableInternal(ParquetTools.java:384)
[...]

PR #4831, opened 2023-11-15
Add parquet read TableDefinition support

### LINKED ISSUE #4746: Initially empty hive partitioned parquet support
I'd like an easy way to create a refreshing, hive partitioned parquet table without the backing directory structures needing to exist - ie, the ability to create an initially empty hive parquet partitioned table. This is useful when bootstrapping a query starting from nothing. There might be low-level support for this via `io.deephaven.parquet.table.ParquetTools#readPartitionedTable` or creating `PartitionAwareSourceTable` directly, but it's not easy from high-level user tooling.

Essentially, the user would have to provide the root directory and the table definition with partition columns up front.

Something like this might also solve #4738.

devinrsmith:
Additionally, adds explicit entry points for single, flat-partitioned, and kv-partitioned reads.

Fixes #4746
Partial workaround for #871


extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -741,6 +838,127 @@ public static Table readPartitionedTableWithMetadata(
...
+     * Callers wishing to be more explicit (for example, to skip some columns) may prefer to call
+     * {@link #readSingleTable(File, ParquetInstructions, TableDefinition)}.
+     *
+     * @param file the parquet file
+     * @param readInstructions the instructions for customizations while reading
+     * @return the table
+     * @see ParquetTableLocationKey#ParquetTableLocationKey(File, int, Map)
+     * @see #readSingleFileTable(ParquetTableLocationKey, ParquetInstructions, TableDefinition)
+     */
+    public static Table readSingleTable(
```
rcaudy, line 935: `readSingleFileTable`?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -684,22 +774,25 @@ public static Table readPartitionedTable(
...
-            if (readInstructions.isRefreshing()) {
-                throw new IllegalArgumentException(
-                        "Unable to infer schema for a refreshing partitioned parquet table when there are no initial parquet files");
-            }
-            return TableTools.emptyTable(0);
+        return readPartitionedTable(locationKeyFinder, readInstructions);
+    }
+
+    /**
+     * TODO(deephaven-core#877): Support schema merge when discovering multiple parquet files
```
rcaudy, line 787: Maybe don't put TODOs in JavaDoc.

py/server/deephaven/parquet.py
```diff
@@ -76,30 +79,55 @@ def _build_parquet_instructions(
...
-    builder.setIsRefreshing(is_refreshing)
+    if is_refreshing:
+        builder.setIsRefreshing(is_refreshing)
 
     return builder.build()
 
+class ParquetType(Enum):
+    """ The parquet file layout type. """
+
+    SINGLE = 1
```
rcaudy, line 96: SINGLE_FILE

rcaudy (commented):
Minimal comments from me. I think this is a nice change. Deferring approval for Python reviewers.

py/server/deephaven/parquet.py
devinrsmith: So, I do think it's a _bit_ redundant to explicitly call out "FILE"; the equivalent for the others would be `FLAT_PARTITIONED_DIRECTORY`, etc. The docs do mention "file" or "directory" as appropriate.

I do think these types also cleanly map into what we might expect w/ s3. So, `s3://my-bucket/foo.parquet` should work with `SINGLE`. And I do think the existing documentation probably still works in the s3 case - the big different from the parquet side is that `path (str): the file or directory to examine` needs to be expanded to include other URIs.

I'm not wholly against changing, just not sure it's helpful.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
devinrsmith: I do think "File" is a bit redundant here; but I could also argue that same point for "Table". I was following the existing nomenclature wrt `readTable`; but it might make sense to go to:

`readSingle`, `readFlatPartitioned`, `readKeyValuePartitioned`.

py/server/deephaven/parquet.py
```diff
@@ -76,30 +80,72 @@ def _build_parquet_instructions(
...
+                Column(name=name, data_type=dtype).j_column_definition
+                for name, dtype in table_definition.items()
+            ]
+        )
+    elif isinstance(table_definition, List):
+        return _JTableDefinition.of(
+            [col.j_column_definition for col in table_definition]
+        )
+    else:
+        raise DHError("not type")
```
chipkent, line 109: Unhelpful error message.

py/server/deephaven/parquet.py
```diff
@@ -76,30 +80,72 @@ def _build_parquet_instructions(
...
     if target_page_size is not None:
         builder.setTargetPageSize(target_page_size)
 
-    builder.setIsRefreshing(is_refreshing)
+    if is_refreshing:
+        builder.setIsRefreshing(is_refreshing)
 
     return builder.build()
 
+def _j_table_definition(table_definition: Union[Dict[str, DType], List[Column], None]):
```
chipkent, line 94: @jmao-denver Should this go in `table.py` or some other location where it could be reused?

chipkent, line 94: Should have a return type hint.  I think it is `jpy.JType` for java types.

py/server/deephaven/parquet.py
chipkent: I think SINGLE_FILE is more clear.

py/server/deephaven/parquet.py
```diff
@@ -76,30 +80,72 @@ def _build_parquet_instructions(
...
+            ]
+        )
+    elif isinstance(table_definition, List):
+        return _JTableDefinition.of(
+            [col.j_column_definition for col in table_definition]
+        )
+    else:
+        raise DHError("not type")
+
+class ParquetType(Enum):
```
chipkent, line 111: Class name is not obvious for the user.  e.g. `ParqueFileLayout` would be more clear.

py/server/deephaven/parquet.py
```diff
@@ -76,30 +80,72 @@ def _build_parquet_instructions(
...
+    """ A directory containing a _metadata parquet file and an optional _common_metadata parquet file. """
+
 
 def read(
     path: str,
-    col_instructions: List[ColumnInstruction] = None,
+    col_instructions: Optional[List[ColumnInstruction]] = None,
     is_legacy_parquet: bool = False,
     is_refreshing: bool = False,
+    type: Optional[ParquetType] = None,
```
chipkent, line 132: `type`->`file_layout`?

py/server/deephaven/parquet.py
```diff
@@ -76,30 +80,72 @@ def _build_parquet_instructions(
...
 
     Args:
         path (str): the file or directory to examine
-        col_instructions (List[ColumnInstruction]): instructions for customizations while reading
+        col_instructions (Optional[List[ColumnInstruction]]): instructions for customizations while reading, None by
+            default.
         is_legacy_parquet (bool): if the parquet data is legacy
         is_refreshing (bool): if the parquet data represents a refreshing source
-
+        type (Optional[ParquetType]): the parquet type, by default None. When None, the type is inferred.
```
chipkent, line 143: "type" is not obvious.  The variable and docs will not be understandable by users.

py/server/deephaven/parquet.py
```diff
@@ -76,30 +80,72 @@ def _build_parquet_instructions(
...
+        table_definition (Union[Dict[str, DType], List[Column], None]): the table definition, by default None. When None,
+            the definition is inferred from the parquet file(s). Setting a definition guarantees the returned table will
+            have that definition. This is useful for bootstrapping purposes when the initial partitioned directory is
+            empty and is_refreshing=True. It is also useful for specifying a subset of the parquet definition. When set,
+            type must also be set.
```
chipkent, line 148: "type"

py/server/deephaven/parquet.py
```diff
@@ -113,12 +159,36 @@ def read(
...
+            if not type:
+                raise DHError("Must provide type when table_definition is set")
+            if type == ParquetType.SINGLE:
+                j_table = _JParquetTools.readSingleTable(_JFile(path), read_instructions, j_table_definition)
+            elif type == ParquetType.FLAT_PARTITIONED:
+                j_table = _JParquetTools.readFlatPartitionedTable(_JFile(path), read_instructions, j_table_definition)
+            elif type == ParquetType.KV_PARTITIONED:
+                j_table = _JParquetTools.readKeyValuePartitionedTable(_JFile(path), read_instructions, j_table_definition)
+            elif type == ParquetType.METADATA_PARTITIONED:
+                raise DHError(f"{ParquetType.METADATA_PARTITIONED} with table_definition not currently supported")
```
chipkent, line 175: Do you want the f-string here?

py/server/tests/test_parquet.py
```diff
@@ -51,15 +49,15 @@ def test_crd(self):
         with self.subTest(msg="write_table(Table, str)"):
             write(table, file_location)
             self.assertTrue(os.path.exists(file_location))
-            table2 = read(file_location)
+            table2 = read(file_location, type=ParquetType.SINGLE)
```
chipkent, line 52: These type specs are not needed, correct?  If they are needed, I'm sure user code will break.

malhotrashivam (dismissed):

py/server/tests/test_parquet.py
devinrsmith: This is mostly about having the test code be more targeted - no reason to test general parquet layout inference for every single test IMO. There are tests specifically designed to test layout inference.

chipkent: I'm good with that.  Just confirming that we weren't breaking users.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
rcaudy: SingleTable seems weird. All of these return a "single table". 
"Single location table" might be more accurate, if you don't want to say "file".

py/server/deephaven/parquet.py
rcaudy: That s3 object is a parquet file in every meaningful away.

rcaudy: Look at language Amazon uses:
https://docs.aws.amazon.com/glue/latest/dg/aws-glue-programming-etl-format-parquet-home.html

py/server/deephaven/parquet.py
devinrsmith: Just cleanly support refactoring if `METADATA_PARTITIONED` was renamed.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
devinrsmith: I've acquiesced (after looking up how to spell...).

py/server/deephaven/parquet.py
devinrsmith: You've convinced me.

py/server/deephaven/parquet.py
chipkent: Sounds good.

chipkent (approved):

jmao-denver (approved):
The Python changes LGTM

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3446
Reference: https://github.com/deephaven/deephaven.io/issues/3447

PR #4839, opened 2023-11-15
Apply query scope ctx on formula aggregation

### LINKED ISSUE #4834: Python scope does not work  properly in agg formulas
my_file1.py:
```python
def my_fn(vals)
    import deephaven.dtypes as dht
    return dht.array(dht.double, [i+2 for i in vals])
```

my_file2.py
```python
from my_file1 import my_fn

def run_func_works():
    t = empty_table(1000).update_view(["A=i%2", "B=A+3"])
    t = t.group_by('A').update('C=(double[])my_fn(B)').ungroup()
    return t

def run_func_not_work():
    t = empty_table(1000).update_view(["A=i%2", "B=A+3"])
    t = t.agg_by([agg.formula("(double[])my_fn(each)", formula_param='each', cols=['C=B'])], by='A') # This fails, my_fn cannot be found
    return t
```
 
From Console:
```python
from my_file2 import run_func_works, run_func_not_work

run_func_works()
run_func_not_work()
```
 
The second one gives ‘Cannot find method my_fn’. This one only works if the run_func_not_work is defined in global scope itself (i.e., not imported).

jmao-denver:
Fixes #4834

### DIFF
--- a/py/server/tests/test_table.py
+++ b/py/server/tests/test_table.py
@@ -943,7 +943,7 @@ class TableTestCase(BaseTestCase):
 
     def test_callable_attrs_in_query(self):
         input_cols = [
-            datetime_col(name="DTCol", data=[1,10000000]),
+            datetime_col(name="DTCol", data=[1, 10000000]),
         ]
         test_table = new_table(cols=input_cols)
         rt = test_table.update("Year = (int)year(DTCol, timeZone(`ET`))")
@@ -1025,7 +1025,7 @@ class TableTestCase(BaseTestCase):
             unique(cols=["ua = a", "ub = b"], include_nulls=True, non_unique_sentinel=-1),
             count_distinct(cols=["csa = a", "csb = b"], count_nulls=True),
             distinct(cols=["da = a", "db = b"], include_nulls=True),
-            ]
+        ]
         rt = test_table.agg_by(aggs=aggs, by=["c"])
         self.assertEqual(rt.size, test_table.select_distinct(["c"]).size)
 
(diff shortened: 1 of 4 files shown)


py/server/tests/test_table.py
```diff
@@ -1063,6 +1063,19 @@ def test_agg_count_and_partition_error(self):
...
         self.assertIn("string value", str(cm.exception))
 
+    def test_agg_formula_scope(self):
+        def run_func_work():
+            def my_fn(vals):
+                import deephaven.dtypes as dht
+                return dht.array(dht.double, [i + 2 for i in vals])
+
+            t = empty_table(1000).update_view(["A=i%2", "B=A+3"])
+            t = t.agg_by([formula("(double[])my_fn(each)", formula_param='each', cols=['C=B'])], by='A') # This
```
chipkent, line 1073: "# This"?

py/server/tests/test_table.py
```diff
@@ -1063,6 +1063,19 @@ def test_agg_count_and_partition_error(self):
...
+            return t
+
+        t = run_func_work()
+        self.assertIsNotNone(t)
+
```
chipkent, line 1078: The way this feature is implemented, I think there is a case we fail in, but maybe we don't care.

```python

def make_agg():
    def my_fn(x):
        return x

    return formula("(double[])my_fn(each)", formula_param='each', cols=['C=B'])

t = t.agg_by([make_agg()], by='A')
```
Here there is a disconnect in the `agg_by` and the `formula` scopes, which leads to problems.

jmao-denver: Right, only the scope in which `Table.agg_by` is called is visible to the formulas. I don't think we need to care about this case, especially when it can be easily worked around.

chipkent (approved):

py/server/tests/test_table.py
rcaudy: This is the same as any and all formula evaluations in the engine. We substitute for scope variables at operation initialization time, not when the expression is composed.

Issue #4718, opened 2023-10-25
Improve parallelization of reads through VariablePageSizeColumnChunkPageStore


malhotrashivam:
It looks like while reading a parquet file using`VariablePageSizeColumnChunkPageStore`, to reach a particular page, we’re materializing every prior page sequentially. This makes the reading process extremely serial since pages cannot be read independently in parallel.

We should check if we can avoid the materialization for pages before the target page by simply reading headers to determine the page size boundaries instead of materializing whole pages prematurely.

We can also looking at widening the lock so we have less switching, too… right now it’s in extendOnePage.

malhotrashivam:
On further inspection, its clear that we don't materialize whole pages to reach any particular page but just iterate over all the headers to calculate the number of rows. We are fixing this issue by introducing a new offset index based page store for faster page access as part of #4844

malhotrashivam:
As noted earlier, `VariablePageSizeColumnChunkPageStore` does not materialize full pages but just the page headers.
Also, going forward, we are using `OffsetIndexBasedColumnChunkPageStore` whenever offset index is present in the file, since it provides faster access for pages. 
So we don't have any immediate need or clear path to optimize `VariablePageSizeColumnChunkPageStore` any further.

PR #4844, opened 2023-11-15
Added offset index based parquet reading support

### LINKED ISSUE #4717: Properly use fixed page size reader for parquet
For reading the pages of a parquet file, we have two approaches:
- `FixedPageSizeColumnChunkPageStore` : If the page sizes are fixed
- `VariablePageSizeColumnChunkPageStore`: if the page sizes are variable. This logic is more complex and time consuming since we need to determine page size boundaries properly for each page.

Currently, we always use `VariablePageSizeColumnChunkPageStore`. 

As part of this issue, we should properly use and test `FixedPageSizeColumnChunkPageStore`.

malhotrashivam:
Currently, while reading a particular row from a parquet file,
- We first check if the row is in any of the pages read so far.
- If not, we keep on extending one page till we reach the required page containing the row. Extending the page means computing the number of rows and setting up some internal data structures for each page.
All of this happens inside `VariablePageSizeColumnChunkPageStore`.

Some of the above operations can be optimized using `OffsetIndex` data structure.
[OffsetIndex](https://github.com/apache/parquet-format/blob/master/PageIndex.md#:~:text=a%20scan%20predicate-,OffsetIndex,-%3A%20this%20allows%20navigation) is an optional field in Parquet which gives us three things for each page in a parquet file:
1. Byte offset
2. Compressed page size
3. Index of first row _(<- Most useful for this PR)_

So for locating a page containing a row, we can directly binary search inside the offset index and don't need to extend pages up to that page. Doing this helps speed up sparse reads and improves the parallelization.

This PR makes the change and also fixes a bug in writing offset index for array and vector columns. Because of this bug fix, we can not use the old code to read the parquet files generated with the new code. So the PR is backwards compatible but not forwards compatible.

Also, as part of this PR, we are deleting the `FixedPageSizeColumnChunkPageStore` because to detect if page sizes are fixed, we need offset index. So fixed page size support is added in the new `OffsetIndexBasedColumnChunkPageStore`.

Benchmarking results: https://docs.google.com/document/d/19AIp6UpHTWpbmxo0ldkBYyuRKOIhlGLP2NHU26jOeOM/edit?usp=sharing

Closes #4717, #4718 
Related to #4879


extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReader.java
```diff
@@ -32,9 +34,24 @@ public interface ColumnChunkReader {
      */
     int getMaxRl();
 
+    /**
+     * @return The offset index for this column chunk, or null if it not found in the metadata.
+     */
+    @Nullable
+    OffsetIndex getOffsetIndex();
+
     interface ColumnPageReaderIterator extends Iterator<ColumnPageReader>, AutoCloseable {
```
malhotrashivam, line 43: Should I rename this class to `ColumnPageReaderAccessor` because now there are two ways to create a ColumnPageReader for a particular page, we can either iterate and get it from `next()` or we can directly create one using the page number?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/FixedPageSizeColumnChunkPageStore.java
```diff
@@ -17,6 +17,10 @@
 import java.lang.ref.WeakReference;
 import java.util.Arrays;
 
+/**
+ * Unused class, use {@link OffsetIndexBasedColumnChunkPageStore} instead because to check if the page sizes are fixed
+ * without actually reading the page headers, we need offset index.
+ */
 class FixedPageSizeColumnChunkPageStore<ATTR extends Any> extends ColumnChunkPageStore<ATTR> {
```
malhotrashivam, line 24: Should I delete this class? I also hit exceptions while using it but didn't dig too deep. So it has some unidentified bugs too.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    private ChunkPage<ATTR> getPage(final int pageNum) {
+        if (pageNum < 0 || pageNum >= numPages) {
+            throw new IllegalArgumentException("pageNum " + pageNum + " is out of range [0, " + numPages + ")");
+        }
+        final PageCache.IntrusivePage<ATTR> page;
+        WeakReference<PageCache.IntrusivePage<ATTR>> pageRef = pages[pageNum];
+        if (pageRef == null || pageRef.get() == null) {
+            synchronized (objectsForSynchronizingPageAccess[pageNum]) {
```
malhotrashivam, line 101: Is there a better way to synchronize access here?
Basically, I want to prevent two threads from extending the same page.

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -614,6 +614,94 @@ private Table maybeFixBigDecimal(Table toFix) {
...
+        try {
+            ParquetTools.readTable(new File(path)).select();
+        } catch (RuntimeException e) {
+            if (e.getCause() instanceof InvalidParquetFileException) {
+                final String InvalidParquetFileErrorMsgString = "Invalid parquet file detected, please ensure the " +
+                        "file is fetched properly from Git LFS. Run commands 'git lfs install; git lfs pull' inside " +
+                        "the repo to pull the files from LFS. Check cause of exception for more details.";
+                throw new UncheckedDeephavenException(InvalidParquetFileErrorMsgString, e.getCause());
+            }
+            throw e;
```
devinrsmith, line 653: Looks like this block is repeated a couple of times, would be nice to make method for it.

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -614,6 +614,94 @@ private Table maybeFixBigDecimal(Table toFix) {
...
+        }
+        final ParquetMetadata metadata = new ParquetTableLocationKey(new File(path), 0, null).getMetadata();
+        assertTrue(metadata.getFileMetaData().getKeyValueMetaData().get("deephaven").contains("\"version\":\"0.4.0\""));
+
+        path = ParquetTableReadWriteTest.class.getResource("/ReferenceParquetVectorData.parquet").getFile();
+        ParquetTools.readTable(new File(path)).select();
```
devinrsmith, line 659: Given it's easy to generate the source data, I think we should check that the table contents are equal - this will also save you from needing to materialize via select.

Going to put a plug in for `readSingleFileTable` once #4831 merges

Ditto for other call site like this.

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -614,6 +614,94 @@ private Table maybeFixBigDecimal(Table toFix) {
...
+        assertTrue(ColumnChunkPageStore.satisfiesMinimumVersionRequirements("0.32.0"));
+        assertTrue(ColumnChunkPageStore.satisfiesMinimumVersionRequirements("1.3.0"));
+    }
+
+    @Test
+    public void testWritingDifferentPageSizes() {
+        // Make a table with arrays of decreasing sizes such that different pages will have different number of rows
+        Table arrayTable = TableTools.emptyTable(100).update(
+                "intArrays = java.util.stream.IntStream.range(0, i).toArray()").reverse();
+        final File dest = new File(rootFile + File.separator + "testWritingDifferentPageSizes.parquet");
```
devinrsmith, line 685: Typical idiom is to do `new File(rootFile, name)`

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -614,6 +614,94 @@ private Table maybeFixBigDecimal(Table toFix) {
...
+        final ParquetInstructions writeInstructions = new ParquetInstructions.Builder()
+                .setTargetPageSize(ParquetInstructions.MIN_TARGET_PAGE_SIZE)
+                .build();
+        ParquetTools.writeTable(arrayTable, dest, writeInstructions);
+        Table fromDisk = ParquetTools.readTable(dest).select();
+        TstUtils.assertTableEquals(arrayTable, fromDisk);
```
devinrsmith, line 691: This pattern should also be easier after PR lands.

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -614,6 +614,94 @@ private Table maybeFixBigDecimal(Table toFix) {
...
+        // Make a table such that only the last page has different number of rows, all else have equal number
+        final long NUM_ROWS = 1000;
+        arrayTable = TableTools.emptyTable(NUM_ROWS).update(
+                "intArrays = (i <= 900) ? java.util.stream.IntStream.range(i, i+50).toArray() : " +
+                        "java.util.stream.IntStream.range(i, i+2).toArray()");
+        ParquetTools.writeTable(arrayTable, dest, writeInstructions);
+        fromDisk = ParquetTools.readTable(dest);
+        // Access something on the last page to make sure we can read it
+        final int[] data = (int[]) fromDisk.getColumnSource("intArrays").get(998);
+        assertTrue(data.length == 2 && data[0] == 998 && data[1] == 999);
```
devinrsmith, line 702: I'm not a fan of chaining multiple asserts together w/ an `&&` like this; each should be tested w/ more specific method:

```java
assertEquals(2, data.length);
assertEquals(998, data[0]);
...
```

devinrsmith (commented):
This is not my full review. Just took a quick look at the tests for now. I'll wait until after Ryan's review to dig down deeper.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReaderImpl.java
```diff
@@ -300,6 +319,17 @@ nullMaterializerFactory, path, getFilePath(), fieldTypes, offsetIndex.getOffset(
             return columnPageReader;
         }
 
+        @Override
+        public ColumnPageReader getPageReader(final int pageNum) {
+            if (pageNum > offsetIndex.getPageCount()) {
+                throw new RuntimeException(
```
rcaudy, line 325: Prefer `IllegalArgumentException` or some flavor or `UncheckedDeephavenException`.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
devinrsmith: Good question. How many pages are there typically? It is unfortunate we need to pre-create this state per-page. See other comment I made, may be better way to structure this.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -0,0 +1,154 @@
...
+     */
+    private boolean isPageSizeFixed;
+    /**
+     * Fixed number of rows per page, only valid if {@link #isPageSizeFixed} is true. Used to map from row index to page
+     * number.
+     */
+    private final long fixedPageSize;
+    private final Object[] objectsForSynchronizingPageAccess;
+    private final ColumnPageReader[] columnPageReaders;
+    private final WeakReference<PageCache.IntrusivePage<ATTR>>[] pages;
```
devinrsmith, line 35: I wonder if it's better to organize this w/ a single array, potentially do away w/ the Object.

```java
class PageState {
  ColumnPageReader reader;
  WeakReference<PageCache.IntrusivePage<ATTR>> page;
}
private final PageState[] states;
```

I think there's a two-stage sync model where you wouldn't even have to allocate the states up front:

```java
PageState get(int i) {
  if (states[i] == null) {
    // global sync if not allocated
    synchronize(this) {
      if (states[i] == null) {
        states[i] = new PageState();
      }
    }
  }
}

class PageState {
  ...
  PageCache.IntrusivePage<ATTR> get() {
    PageCache.IntrusivePage<ATTR> localPage;
    if ((localPage = page) == null)) {
      synchronize(this) {
          if ((localPage = page) == null)) {
           ...
         }
      }
      return localPage;
    }
```

I think this is better than maintaining Object sidecar model.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java
```diff
@@ -577,6 +577,15 @@ private void ensureNumValues() throws IOException {
         if (numValues >= 0) {
             return;
         }
+        if (pageHeader == null) {
+            try (final SeekableByteChannel readChannel = channelsProvider.getReadChannel(filePath)) {
+                readChannel.position(offset);
+                ensurePageHeader(readChannel);
```
rcaudy, line 583: Something is wrong with the flow in this code. All of the places that `ensurePageHeader(SeekableByteChannel)` are positioning the channel to `offset`. But `offset` is non-final, and set in both the constructor and in `ensurePageHeader`. I'm hoping that overwriting it in `ensurePageHeader` is wrong and can be removed....

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/FixedPageSizeColumnChunkPageStore.java
rcaudy: Yes, delete it. I understand that that's a feature in the offset-indexed version.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
devinrsmith: Does `ColumnPageReader` need to be a hard ref? In this model I posted, can we have `WeakRef<PageState>` instead?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -0,0 +1,154 @@
...
+        isPageSizeFixed = true;
+        final long firstPageSize;
+        if (numPages > 1) {
+            firstPageSize = offsetIndex.getFirstRowIndex(1) - offsetIndex.getFirstRowIndex(0);
+        } else {
+            firstPageSize = numRows();
+        }
+        objectsForSynchronizingPageAccess = new Object[numPages];
+        for (int i = 0; i < numPages; ++i) {
+            objectsForSynchronizingPageAccess[i] = new Object();
```
rcaudy, line 59: Instead of using an array of locks (which is what this is), we could probably race to construct readers and CAS them into place in an `AtomicReferenceArray`, then synchronize on the readers as the previous implementations did.

To clarify, the patterns you might consider look like:
```
localReader = array.get(index);
if (localReader == null && !array.compareAndSet(index, null, localReader = constructReader(index)) {
    localReader = array.get(index);
} 
```
You might be able to simplify:
```
localReader = array.get(index);
if (localReader == null) {
    localReader = array.compareAndGet(index, e -> e == null ? constructReader(index) : e);
} 
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/FixedPageSizeColumnChunkPageStore.java
devinrsmith: Yes, I would delete if unused.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReader.java
```diff
@@ -32,9 +34,24 @@ public interface ColumnChunkReader {
...
         void close() throws IOException;
+
+        /**
+         * Directly access a page reader for a given page number. This is an optional method that may not be
+         * implemented. Note that the user should either use {@link Iterator} methods or this method, but not both.
+         */
+        @Nullable
+        default ColumnPageReader getPageReader(final int pageNum) {
+            return null;
+        };
```
devinrsmith, line 54: extraneous `;`

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -0,0 +1,154 @@
...
+            if (isPageSizeFixed && i > 0
```
rcaudy, line 60: The logic as you have it is checking the first page unnecessarily while processing the second page.

You should comment that you don't care about the size of the last page, because we can always infer that a row that would come from past the end of the last page (were it the fixed size) must be read from the last page.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReader.java
```diff
@@ -32,9 +34,24 @@ public interface ColumnChunkReader {
...
     interface ColumnPageReaderIterator extends Iterator<ColumnPageReader>, AutoCloseable {
         @Override
         void close() throws IOException;
+
+        /**
+         * Directly access a page reader for a given page number. This is an optional method that may not be
+         * implemented. Note that the user should either use {@link Iterator} methods or this method, but not both.
+         */
+        @Nullable
+        default ColumnPageReader getPageReader(final int pageNum) {
```
devinrsmith, line 52: IMO, we should add a method `boolean supportsDirectAccess()` and make sure our implementations have proper assertions (for example, in `OffsetIndexBasedColumnChunkPageStore`)

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
rcaudy: See answer regarding atomic CAS for reader objects.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReader.java
devinrsmith: The other half of me thinks this should be a _separate_ interface, not merged w/ `ColumnPageReaderIterator`. Are there any cases where a consumer doesn't know their exact access patterns a priori?

```java
...
    interface ColumnPageDirectorAccessor extends AutoCloseable {
        ColumnPageReader getPageReader(final int pageNum);
        
        @Override
        void close() throws IOException;
    }
    
    ColumnPageReaderIterator getPageIterator() throws IOException;
    
    ColumnPageDirectorAccessor getPageAccessor() throws IOException;
    ...
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -0,0 +1,154 @@
...
+        if (pageRef == null || pageRef.get() == null) {
+            synchronized (objectsForSynchronizingPageAccess[pageNum]) {
+                // Make sure no one materialized this page as we waited for the lock
+                pageRef = pages[pageNum];
+                if (pageRef == null || pageRef.get() == null) {
+                    if (columnPageReaders[pageNum] == null) {
+                        columnPageReaders[pageNum] = columnPageReaderIterator.getPageReader(pageNum);
+                    }
+                    try {
+                        page = new PageCache.IntrusivePage<>(toPage(offsetIndex.getFirstRowIndex(pageNum),
```
rcaudy, line 109: I do kind of wonder if we shouldn't just allocate readers on demand when we are forced to read pages, or just get rid of readers altogether.

You sort of "need" to keep one in the legacy case with no offset indexes, or at least remember the offset index information you constructed during iteration for future use.

In this case, you get nothing but a lock and some garbage from having a reader object.

Maybe the answer is to change the legacy code to just populate offset index data on demand and use that to do your reads.

Note that reader objects currently hold:
1. 7 references that are shared by all readers for a particular column chunk (wasteful)
2. 4 fields specific to the page

I think we could refactor the "read one page" code to live inside a column chunk reader, passing the 4 "per page fields" down as needed. Or, we could have short-lived parameter objects to hold those fields. We ought not to be caching page headers indefinitely, etc.

Then you could do the `updateAndGet` pattern around the actual reading, maybe.

I'm OK with divorcing the "use offset indexes" PR from the "get rid of long-lived reader objects" PR, but I do think we should be making this change; these kinds of overheads add up rapidly at the scale we see in enterprise use-cases.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/ColumnChunkPageStore.java
```diff
@@ -48,25 +53,60 @@ private CreatorResult(@NotNull final ColumnChunkPageStore<ATTR> pageStore,
...
     public static <ATTR extends Any> CreatorResult<ATTR> create(
             @NotNull final PageCache<ATTR> pageCache,
             @NotNull final ColumnChunkReader columnChunkReader,
             final long mask,
-            @NotNull final ToPage<ATTR, ?> toPage) throws IOException {
-        final boolean fixedSizePages = columnChunkReader.getPageFixedSize() >= 1;
-        final ColumnChunkPageStore<ATTR> columnChunkPageStore = fixedSizePages
-                ? new FixedPageSizeColumnChunkPageStore<>(pageCache, columnChunkReader, mask, toPage)
+            @NotNull final ToPage<ATTR, ?> toPage,
+            @NotNull final ColumnDefinition<?> columnDefinition) throws IOException {
```
devinrsmith, line 96: This is _technically_ a "breaking" change; another reason I don't like how our implementation is structured right now. Not your issue, but just noting. https://github.com/deephaven/deephaven-core/issues/4850

rcaudy (commented):
Mostly high-level review pass. Be sure that we keep enterprise in the loop regarding forwards compatibility issue, and look for more external validation of our new approach to writing and interpreting offset indexes.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/ColumnChunkPageStore.java
```diff
@@ -48,25 +53,60 @@ private CreatorResult(@NotNull final ColumnChunkPageStore<ATTR> pageStore,
...
+        final boolean canUseOffsetIndex = canUseOffsetIndexBasedPageStore(columnChunkReader, columnDefinition);
```
devinrsmith, line 97: It's unfortunate we have to weave the logic like this - the pervasive passing of `version` now is less than ideal IMO. Maybe we want to have the implementations have access to the full `TableInfo` instead?

devinrsmith (commented):
Will take another pass soon; need to take a break.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/metadata/TableInfo.java
```diff
@@ -61,11 +61,11 @@ public final Map<String, ColumnTypeInfo> columnTypeMap() {
...
 
     /**
-     * @return The Deephaven release version when this metadata format was defined
+     * @return The Deephaven release version when this metadata format was updated
      */
     @Value.Default
     public String version() {
-        return "0.4.0";
+        return "0.31.0";
     }
```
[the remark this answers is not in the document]
malhotrashivam: I think we made the change assuming it would be Larry's change that will change the metadata.
But you are right Devin, its not obvious that this needs to be changed every time we change the metadata.

@rcaudy  can you please look at this comment once if its okay to change this field's semantics to writeVersion?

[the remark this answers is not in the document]
rcaudy: I think we can safely re-word the semantics of the version field we have, and not introduce a new one.
It might be useful to always reflect the version we wrote with, but I think that requires some automation we don't currently have.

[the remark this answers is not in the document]
devinrsmith: @rcaudy : `TableInfo.class.getPackage().getImplementationVersion()`

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/ColumnChunkPageStore.java
malhotrashivam: I tried passing `TableInfo` instead of `version` to `ParquetFileReader` class but that would require adding a dependency of `Deephaven_Community_Core.extensions-parquet-base.main` on `Deephaven_Community_Core.extensions-parquet-table.main` leading to circular dependency.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java
malhotrashivam: The offset field is not final because it can store the offset of page header or the offset of data (which comes after the page header). So once we read the page header, we increment the offset so that following reads directly access the data. 
I have added some comments to explain in the code. Let me know if anything is not clear.

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
malhotrashivam: I didn't want to generate the data because if we change the method `getTableFlat` in future (which we do change very frequently as we add new types), these tests will start to fail. 

Also, currently, `getTablFlat` returns a column `"someTime = DateTimeUtils.now() + i"`
So we cannot regenerate that data because of the `now`.   This is still fixable since I can change the column. But in general didn't like the overall thing.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
malhotrashivam: No, ColumnPageReader doesn't need to be a hard ref. In fact, we are thinking of completely removing the stored reference to the reader object. This comment has the details: https://github.com/deephaven/deephaven-core/pull/4844#discussion_r1396299390

So in the latest patch, I have removed storing reference to the reader.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -0,0 +1,147 @@
...
+        if (numPages == 1) {
+            fixedPageSize = numRows();
+            return;
+        }
+        boolean isPageSizeFixed = true;
+        final long firstPageSize = offsetIndex.getFirstRowIndex(1) - offsetIndex.getFirstRowIndex(0);
+        for (int i = 2; i < numPages && isPageSizeFixed; ++i) {
+            if (offsetIndex.getFirstRowIndex(i) - offsetIndex.getFirstRowIndex(i - 1) != firstPageSize) {
+                isPageSizeFixed = false;
+            }
```
devinrsmith, line 69: A more typical pattern IMO is to not have the boolean checked in the for loop, but just use a `break` here after setting `isPageSizeFixed = false`.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -0,0 +1,147 @@
...
+        Assert.assertion(numPages > 0, "numPages > 0");
+        pageStates = new AtomicReferenceArray<>(numPages);
+        columnPageDirectAccessor = columnChunkReader.getPageAccessor();
+
+        if (numPages == 1) {
+            fixedPageSize = numRows();
+            return;
+        }
+        boolean isPageSizeFixed = true;
+        final long firstPageSize = offsetIndex.getFirstRowIndex(1) - offsetIndex.getFirstRowIndex(0);
```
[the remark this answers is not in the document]
devinrsmith: Ah, I see the assertion earlier.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -0,0 +1,147 @@
...
+
+    private ChunkPage<ATTR> getPage(final int pageNum) {
+        if (pageNum < 0 || pageNum >= numPages) {
+            throw new IllegalArgumentException("pageNum " + pageNum + " is out of range [0, " + numPages + ")");
+        }
+        PageCache.IntrusivePage<ATTR> page;
+        PageState pageState = pageStates.get(pageNum);
+        if (pageState == null) {
+            pageState = pageStates.updateAndGet(pageNum, p -> p == null ? new PageState() : p);
+        }
```
devinrsmith, line 104: While this pattern does work, it's a bit sloppy IMO; we should be able to implement the same sort of loop-CASing here a bit more efficiently.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -0,0 +1,147 @@
...
+        if (fixedPageSize == PAGE_SIZE_NOT_FIXED) {
+            pageNum = findPageNumUsingOffsetIndex(offsetIndex, row);
+            if (pageNum < 0) {
+                pageNum = -2 - pageNum;
+            }
+        } else {
+            pageNum = (int) (row / fixedPageSize);
+            if (pageNum >= numPages) {
+                // This can happen if the last page is of different size from rest of the pages.
+                // We have already checked that row is less than numRows.
```
devinrsmith, line 139: I'm trying to think if this is actually true and could happen.

Let's say we have `fixedPageSize = 100` and `numPages = 3`, and the last page only has 1 row; so, `numRows = 100 * 2 + 1 = 201`.

```
(int)(x / 100)
0, x=[0, 100)
1, x=[100, 200)
2, x=[200, 201)
```

At no point did we ever get back `>=3`.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
malhotrashivam: My latest patch makes this change where we created readers only at the time of reading the page and forget them after that. So every time we read the page, we will create a new reader, which is a very small operation.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java
```diff
@@ -108,9 +116,17 @@ public IntBuffer readKeyValues(IntBuffer keyDest, int nullPlaceholder) throws IO
         }
     }
 
+    /**
+     * If {@link #pageHeader} is {@code null}, read it from the file and increment the {@link #offset} and file position
+     * by the length of page header. This method assumes that file position is set to {@link #offset} before calling.
+     * This method also read the number of values in the page from the header.
+     */
```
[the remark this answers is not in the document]
malhotrashivam: Added this check

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java
```diff
@@ -577,6 +593,15 @@ private void ensureNumValues() throws IOException {
         if (numValues >= 0) {
             return;
         }
+        if (pageHeader == null) {
```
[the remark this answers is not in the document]
malhotrashivam: Good catch. I made the structure of numValues() similar to other methods in the class to avoid the race condition. Let me know if it makes sense now.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
malhotrashivam: Resolving this conversation since we have the discussion on other threads.
https://github.com/deephaven/deephaven-core/pull/4844#discussion_r1396284477
https://github.com/deephaven/deephaven-core/pull/4844#discussion_r1396273371

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java
```diff
@@ -585,34 +592,34 @@ private ValuesReader getDataReader(Encoding dataEncoding, ByteBuffer in, int val
 
     @Override
     public int numValues() throws IOException {
-        ensureNumValues();
-        return numValues;
+        if (numValues > 0) {
```
devinrsmith, line 595: `numValues >= 0`?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -129,13 +130,10 @@ public ChunkPage<ATTR> getPageContaining(@NotNull final FillContext fillContext,
...
         if (fixedPageSize == PAGE_SIZE_NOT_FIXED) {
             pageNum = findPageNumUsingOffsetIndex(offsetIndex, row);
-            if (pageNum < 0) {
-                pageNum = -2 - pageNum;
-            }
         } else {
             pageNum = (int) (row / fixedPageSize);
             if (pageNum >= numPages) {
-                // This can happen if the last page is of different size from rest of the pages.
+                // This can happen if the last page is larger than rest of the pages, which are all the same size.
```
devinrsmith, line 136: This is a weird condition to account for; do we regularly expect that all of the pages will be the same size, _except_ the last page is _larger_?

IMO, if all the page sizes are the same, except the last is larger, we should just use `PAGE_SIZE_NOT_FIXED`.

We could equivalently come up with a formulation to account for the equally strange condition "All of the pages are the same size, except the first which is (larger|smaller)." and make efficient math for that, but it's weird.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
devinrsmith: I'm not worried about the atomic array; I think that `get` + `updateAndGet` is a bit sloppy. We can write tighter code that accomplishes more specifically what we want. Look at the implementation for `updateAndGet` - happy to walk through this with you if you want.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
malhotrashivam: I don't think its a common case.
The challenging part is that we cannot find the size of last page using offset index. Offset index only has the index of first row on the page. So we don't know how many rows there are on the last page without actually reading the page.  So keeping it like this doesn't need the number of rows in the last page.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+        PageCache.IntrusivePage<ATTR> page;
+        PageState pageState = pageStates.get(pageNum);
+        if (pageState == null) {
+            pageState = pageStates.updateAndGet(pageNum, p -> p == null ? new PageState() : p);
+        }
+        if (pageState.pageRef == null || (page = pageState.pageRef.get()) == null) {
+            synchronized (pageState) {
+                // Make sure no one materialized this page as we waited for the lock
+                if (pageState.pageRef == null || (page = pageState.pageRef.get()) == null) {
```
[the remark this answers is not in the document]
devinrsmith: If that's the case, then `PageState` can be static. That said, I'll run something by you regarding this soon. I _think_ there are double checked locking issues I'm looking into. `pageRef` should be volatile, but I'm not sure that's actually enough to satisfy what we are trying to do here...

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
malhotrashivam: The other approach to do this would look like:
```
PageState pageState = pageStates.get(pageNum);
if (pageState == null && !pageStates.compareAndSet(pageNum, null, pageState = new PageState()) {
    pageState = pageStates.get(pageNum);
} 
```
Do you think this would be better than `updateAndGet`?

devinrsmith: Well, there will need to be a loop, this is not correct w/ CAS semantics. The issue is that as written, in the case where our inner function `p != null`, we're doing a spurious volatile write.

Something like:

```java
PageState pageState;
while ((pageState = pageStates.get(pageNum)) == null) {
   pageState = new PageState();
   if (pageStates.weakCompareAndSetVolatile(pageNum, null, pageState)) {
     break;
   }
}

File: extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java [hunk 29]
```diff
@@ -569,25 +592,34 @@ private ValuesReader getDataReader(Encoding dataEncoding, ByteBuffer in, int val
...
+        try (final SeekableByteChannel readChannel = channelsProvider.getReadChannel(filePath)) {
+            readChannel.position(offset);
+            ensurePageHeader(readChannel);
+            // Above will automatically populate numValues
+            Assert.geq(numValues, "numValues", 0);
+            return numValues;
+        }
     }
 
     private void ensureNumValues() throws IOException {
```
malhotrashivam, line 607: Since this method is now only called from inside ensurePageHeader, I want to move it closer to that method in the file. Will make this change as a single commit once rest of the review is done.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
malhotrashivam: Added a comment explaining this

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
malhotrashivam: Made this change.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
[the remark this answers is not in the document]
malhotrashivam: Made the class static.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
malhotrashivam: The pattern we have kept in the final patch is similar to how `updateAndGet` is internally implemented using `weakCompareAndSetVolatile` inside a loop. 
Thanks to Devin for suggesting that approach.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/metadata/TableInfo.java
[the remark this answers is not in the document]
rcaudy: That works? Nice!

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReader.java
malhotrashivam: Done

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/ColumnChunkPageStore.java
```diff
@@ -48,25 +51,60 @@ private CreatorResult(@NotNull final ColumnChunkPageStore<ATTR> pageStore,
...
+        if (columnChunkReader.getOffsetIndex() == null) {
+            return false;
+        }
+        final String version = columnChunkReader.getVersion();
+        if (version == null) {
+            // Parquet file not written by deephaven, can use offset index
+            return true;
+        }
+        // For vector and array column types, versions before 0.31.0 had a bug in offset index calculation, fixed as
+        // part of deephaven-core#4844
```
devinrsmith, line 65: I think we should eventually plan to delete this logic `canUseOffsetIndexBasedPageStore` to workaround bad DHC written parquet. Instead, we should eventually have a more appropriate error message ("please use this tool <link to DHC issue> to re-write the parquet file"). Probably worth filing a new issue to eventually remove this workaround.

malhotrashivam: Created this issue: https://github.com/deephaven/deephaven-core/issues/4879
Will also add a comment in the code.

devinrsmith (approved):

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java
```diff
@@ -569,25 +592,34 @@ private ValuesReader getDataReader(Encoding dataEncoding, ByteBuffer in, int val
...
-        ensureNumValues();
-        return numValues;
+        if (numValues >= 0) {
+            return numValues;
+        }
+        try (final SeekableByteChannel readChannel = channelsProvider.getReadChannel(filePath)) {
+            readChannel.position(offset);
+            ensurePageHeader(readChannel);
+            // Above will automatically populate numValues
+            Assert.geq(numValues, "numValues", 0);
```
rcaudy, line 602:
```suggestion
            Assert.geqZero(numValues, "numValues");
```

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java
rcaudy, line 607: You said you want to move this, but you should just inline it. The logic in `ensurePageHeader` will make more sense if you do that and then cleanup.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java
```diff
@@ -108,9 +116,17 @@ public IntBuffer readKeyValues(IntBuffer keyDest, int nullPlaceholder) throws IO
...
+     * by the length of page header. This method assumes that file position is set to {@link #offset} before calling.
+     * This method also read the number of values in the page from the header.
+     */
     private synchronized void ensurePageHeader(SeekableByteChannel file) throws IOException {
         if (pageHeader == null) {
-            offset = file.position();
+            if (file.position() != offset) {
+                throw new IllegalStateException("File position = " + file.position() + " not equal to expected offset ="
+                        + offset);
+            }
```
rcaudy, line 129: This is silly. All of the callers invoke `readChannel.position(offset)` immediately before calling this method. Move that logic in here, and delete the check.
If you want to get fancier, you could read the current file position before hand and then restore it afterwards (right now you unconditionally restore it to `offset`).

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/metadata/TableInfo.java
```diff
@@ -61,11 +61,16 @@ public final Map<String, ColumnTypeInfo> columnTypeMap() {
...
-     * @return The Deephaven release version when this metadata format was defined
+     * @return The Deephaven release version used to write the parquet file
      */
     @Value.Default
     public String version() {
-        return "0.4.0";
+        final String version = TableInfo.class.getPackage().getImplementationVersion();
+        if (version == null) {
+            // For unit tests
+            return "0.dev.0";
```
rcaudy, line 71: When will this happen? What's special about the unit test environment? Seems like this would be bad in a lot of cases.
Would "unknown" satisfy the test requirement?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/ColumnChunkPageStore.java
```diff
@@ -48,25 +51,63 @@ private CreatorResult(@NotNull final ColumnChunkPageStore<ATTR> pageStore,
...
+        return true;
+    }
+
+    private static final Pattern VERSION_PATTERN = Pattern.compile("(\\d+)\\.(\\d+)\\.(\\d+)");
+
+    /**
+     * Check if the version is greater than or equal to 0.31.0, or it doesn't follow the versioning schema X.Y.Z
+     */
+    @VisibleForTesting
+    public static boolean satisfiesMinimumVersionRequirements(@NotNull final String version) {
```
rcaudy, line 79: Name seems a little off.
`hasCorrectVectorOffsetIndexes`?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/ColumnChunkPageStore.java
rcaudy: We don't expect anyone to use our parquet libraries at this level but us. Java does not make it easy to specify that.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -0,0 +1,151 @@
...
+    private final AtomicReferenceArray<PageState<ATTR>> pageStates;
+    private final ColumnChunkReader.ColumnPageDirectAccessor columnPageDirectAccessor;
+
+    OffsetIndexBasedColumnChunkPageStore(@NotNull final PageCache<ATTR> pageCache,
+            @NotNull final ColumnChunkReader columnChunkReader,
+            final long mask,
+            @NotNull final ToPage<ATTR, ?> toPage) throws IOException {
+        super(pageCache, columnChunkReader, mask, toPage);
+        offsetIndex = columnChunkReader.getOffsetIndex();
+        Assert.assertion(offsetIndex != null, "offsetIndex != null");
```
rcaudy, line 56:
```suggestion
        Assert.neqNull(offsetIndex, "offsetIndex");
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -0,0 +1,151 @@
...
+        numPages = offsetIndex.getPageCount();
+        Assert.assertion(numPages > 0, "numPages > 0");
```
rcaudy, line 58:
```suggestion
        Assert.gtZero(numPages, "numPages");
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -0,0 +1,151 @@
...
+        while (low <= high) {
+            final int mid = (low + high) >>> 1;
+            final long midVal = offsetIndex.getFirstRowIndex(mid);
+
+            if (midVal < row)
+                low = mid + 1;
+            else if (midVal > row)
+                high = mid - 1;
+            else
+                return mid; // 'row' is the first row of page
```
rcaudy, line 94:
```suggestion
            if (midVal < row) {
                low = mid + 1;
            } else if (midVal > row) {
                high = mid - 1;
            } else {
                return mid; // 'row' is the first row of page
            }
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -0,0 +1,151 @@
...
+        int pageNum;
+        if (fixedPageSize == PAGE_SIZE_NOT_FIXED) {
+            pageNum = findPageNumUsingOffsetIndex(offsetIndex, row);
+        } else {
+            pageNum = (int) (row / fixedPageSize);
+            if (pageNum >= numPages) {
+                // This can happen if the last page is larger than rest of the pages, which are all the same size.
+                // We have already checked that row is less than numRows.
+                Assert.assertion(row >= offsetIndex.getFirstRowIndex(numPages - 1),
+                        "row >= offsetIndex.getFirstRowIndex(numPages - 1)");
```
rcaudy, line 145:
```suggestion
                Assert.geq(row, "row",
                        offsetIndex.getFirstRowIndex(numPages - 1), "offsetIndex.getFirstRowIndex(numPages - 1)");
```

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -627,6 +628,121 @@ private Table maybeFixBigDecimal(Table toFix) {
...
+
+        path = ParquetTableReadWriteTest.class.getResource("/ReferenceParquetVectorData.parquet").getFile();
+        readParquetFileFromGitLFS(new File(path)).select();
+
+        path = ParquetTableReadWriteTest.class.getResource("/ReferenceParquetArrayData.parquet").getFile();
+        readParquetFileFromGitLFS(new File(path)).select();
+    }
+
+    @Test
+    public void testVersionChecks() {
```
rcaudy, line 684: Fails:
```
junit.framework.AssertionFailedError: Exception expected for invalid version string
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.TestCase.fail(TestCase.java:223)
	at io.deephaven.parquet.table.ParquetTableReadWriteTest.testVersionChecks(ParquetTableReadWriteTest.java:689)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at io.deephaven.engine.testutil.junit4.EngineCleanup$1.evaluate(EngineCleanup.java:34)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.runTestClass(JUnitTestClassExecutor.java:108)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:57)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:39)
	at org.gradle.api.internal.tasks.testing.junit.AbstractJUnitTestClassProcessor.processTestClass(AbstractJUnitTestClassProcessor.java:62)
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.processTestClass(SuiteTestClassProcessor.java:52)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:33)
	at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:94)
	at com.sun.proxy.$Proxy2.processTestClass(Unknown Source)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker$2.run(TestWorker.java:176)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.executeAndMaintainThreadName(TestWorker.java:129)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:100)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:60)
	at org.gradle.process.internal.worker.child.ActionExecutionWorker.execute(ActionExecutionWorker.java:56)
	at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:113)
	at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:65)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.run(GradleWorkerMain.java:69)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.main(GradleWorkerMain.java:74)
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/metadata/TableInfo.java
devinrsmith: This is for anytime the code is run from class files as opposed to jars. I think "unknown" is better in that case as well too.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReader.java
```diff
@@ -69,4 +84,10 @@ public int getMaxId() {
     }
 
     PrimitiveType getType();
+
+    /**
+     * @return The "version" string from deephaven specific parquet metadata, or null if it's not present.
+     */
+    @Nullable
+    String getVersion();
```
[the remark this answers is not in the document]
malhotrashivam: I tried passing TableInfo instead of version to ParquetFileReader class but that would require adding a dependency of Deephaven_Community_Core.extensions-parquet-base.main on Deephaven_Community_Core.extensions-parquet-table.main leading to circular dependency.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java
malhotrashivam: Its important to set the file position to offset because if we have already read the header, offset will store the offset for the data in the page. So we want to set file position as the location of data.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReader.java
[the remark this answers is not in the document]
rcaudy: OK

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReaderImpl.java
```diff
@@ -264,24 +266,21 @@ public ColumnPageReader next() {
...
                                 "Unknown parquet data page header type " + pageHeader.type);
                 }
-                final Supplier<Dictionary> pageDictionarySupplier =
-                        (encoding == PLAIN_DICTIONARY || encoding == RLE_DICTIONARY)
-                                ? dictionarySupplier
-                                : () -> NULL_DICTIONARY;
-                return new ColumnPageReaderImpl(
-                        channelsProvider, decompressor, pageDictionarySupplier,
+                if ((encoding == PLAIN_DICTIONARY || encoding == RLE_DICTIONARY)
+                        && dictionarySupplier.get() == NULL_DICTIONARY) {
```
rcaudy, line 270: This is probably a bad idea. I'd need to look at the dictionary suppliers, but they do a lot of work, and I don't recall if they cache internally.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java
```diff
@@ -110,53 +122,66 @@ public int readRowCount() throws IOException {
...
-                    offset += bufferedIS.position();
-                    success = true;
-                } catch (IOException e) {
-                    success = false;
-                    if (maxHeader > MAX_HEADER) {
-                        throw e;
+    private void ensurePageHeader(final SeekableByteChannel file) throws IOException {
+        // Set this channel's position to appropriate offset for reading. If pageHeader is null, this offset would be
+        // the offset of page header, else it would be the offset of data.
+        file.position(offset);
```
rcaudy, line 137: Should this be inside the synchronized block? I guess the file is never shared, and consequently need not have access synchronized? So I think the answer is "no"?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/metadata/TableInfo.java
```diff
@@ -67,8 +67,8 @@ public final Map<String, ColumnTypeInfo> columnTypeMap() {
     public String version() {
         final String version = TableInfo.class.getPackage().getImplementationVersion();
         if (version == null) {
-            // For unit tests
-            return "0.dev.0";
+            // When the code is run from class files as opposed to jars, like in unit tests
```
rcaudy, line 70: Better. But does this create issues if someone re-uses our code?

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java
malhotrashivam: Yes, that's right. And it wasn't inside the synchronized block earlier, so keeping it consistent.

rcaudy (approved):

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReaderImpl.java
```diff
@@ -287,20 +296,39 @@ public boolean hasNext() {
...
+        ColumnPageDirectAccessorImpl() {}
+
+        @Override
+        public ColumnPageReader getPageReader(final int pageNum) {
+            if (pageNum < 0 || pageNum >= offsetIndex.getPageCount()) {
+                throw new IndexOutOfBoundsException(
+                        "pageNum=" + pageNum + ", offsetIndex.getPageCount()=" + offsetIndex.getPageCount());
+            }
+            final int numValues = -1; // Will be populated properly when we read the page header
+            return new ColumnPageReaderImpl(channelsProvider, decompressor, dictionarySupplier, nullMaterializerFactory,
```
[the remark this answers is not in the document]
malhotrashivam: Reverted back my change

devinrsmith (approved):

PR #4853, opened 2023-11-16
Docs CI Upload Versioned Artifacts

stanbrub:
Updated Docs CI workflow to produce the following artifacts:
- cppclient-docs-0.31.0
- cppclient-examples-0.31.0
- javadocs-0.31.0
- pyclient-docs-0.31.0	
- pyserver-docs-0.31.0	
- rdoc-0.31.0	
- typedoc-0.31.0

Also, added appropriate verbiage to the RELEASE.MD.

These end up in the Summary for the Workflow and can be download/uploaded to Github release assets similar to the ones produced by the Publish CI.


.github/workflows/docs-ci.yml
```diff
@@ -38,6 +38,13 @@ jobs:
         run: |
           .github/scripts/gradle-properties.sh >> gradle.properties
           cat gradle.properties
+          
+      - name: Set DHC Version
+        id: dhc-version
+        run: |
+          ./gradlew printVersion -q 
+          ./gradlew printVersion -q > dhc_version.txt
```
devinrsmith, line 46: We've got an `outputVersion` task we can use instead; it can be invoked during the respective `burrunan/gradle-cache-action@v1` step. ie:

```
arguments: --scan combined-javadoc:allJavadoc outputVersion
```

There's an example of this in build-ci.yml. (Ditto for the other places you are using printVersion.)

.github/workflows/docs-ci.yml
```diff
@@ -50,7 +57,7 @@ jobs:
         if: ${{ github.ref == 'refs/heads/main' }}
         uses: actions/upload-artifact@v3
         with:
-          name: javadocs
+          name: javadocs-${{ steps.dhc-version.outputs.version }}
```
devinrsmith, line 60: Separately... we should probably use this technique for our release artifacts too - much nicer w/ them having version attached...

RELEASE.md
```diff
@@ -150,7 +150,8 @@ Once the workflow job is done, ensure all publication sources have the new artif
 ### 5. Download artifacts
 
 Once the full publish-ci.yml worflow is done, the release artifacts can be downloaded from the GitHub Artifacts (located in the "Summary" tab of the action).
-These are currently manual steps taken from the browser.
+Similarly, release artifacts can be downloaded from the docs-ci.yml workflow.
```
devinrsmith, line 153: I would add a new last section under `## Artifacts` that describes the documentation artifacts.

I do wonder about the javadoc ones though, b/c they are very big...

devinrsmith (changes requested):

.github/workflows/docs-ci.yml
stanbrub: Do you mean like _artifacts-0.30.1_?  Or are you suggesting producing separate artifacts for what that contains?

RELEASE.md
devinrsmith: Can you comment on the size of the other documentation artifacts? I could be in favor of only adding javadoc workflow artifact on release - would want to specifically note it in the yml file.

devinrsmith (commented):
Looks like you've got some conflicts w/ main...

Happy to approve this soon. The next step after this PR would be to actually publish based on release branches to named subdirectories of deephaven.io.

RELEASE.md
stanbrub: Below are compressed file sizes:

Docs Artifacts
- 2.4MB  pyserver-docs-0.31.0.zip
- 748kB  pyclient-docs-0.31.0.zip
- 200kB  cppclient-examples-0.31.0.zip
- 39MB  javadocs-0.31.0.zip
- 694kB  cppclient-docs-0.31.0.zip

Other Assets
- 182MB  server-jetty-0.30.1.zip
- 194MB  server-jetty-0.30.1.tar
- 98kB  pydeephaven-0.30.1-py3-none-any.whl
- 149kB  deephaven_core-0.30.1-py3-none-any.whl
- 19MB   deephaven-core-0.30.1.zip (source)
- 9.2MB  deephaven-core-0.30.1.tar.gz (source)

.github/workflows/docs-ci.yml
devinrsmith: Yeah, just making it so when the server does the publish workflow, the artifacts that it produces has unique names

RELEASE.md
devinrsmith: Okay; 39MB is big but not unreasonable. I don't think the workflow artifacts stick around for an extended period of time (I forget exactly how often we've set them to expire), so I wouldn't be too worried about that right now.

RELEASE.md
```diff
@@ -208,7 +219,8 @@ Create a new [GitHub release](https://github.com/deephaven/deephaven-core/releas
 
 The convention is to have the Release title of the form `vX.Y.Z` and to autogenerate the release notes in comparison to the previous release tag. Question: should we always generate release notes based off of the previous minor release, instead of patch? Our git release workflow suggests we may want to do it always minor to minor.
 
-Upload the Deephaven server application, deephaven-core wheel, pydeephaven wheel, and SBOM artifacts.
+Upload the Deephaven server application, deephaven-core wheel, pydeephaven wheel, and SBOM artifacts. Also, upload the C++, Java, and Python docs artifacts.
```
devinrsmith, line 222: Update this w/ the typescript + R.

RELEASE.md
```diff
@@ -150,7 +160,8 @@ Once the workflow job is done, ensure all publication sources have the new artif
 ### 5. Download artifacts
 
 Once the full publish-ci.yml worflow is done, the release artifacts can be downloaded from the GitHub Artifacts (located in the "Summary" tab of the action).
-These are currently manual steps taken from the browser.
+Similarly, release artifacts can be downloaded from the docs-ci.yml workflow.
```
devinrsmith, line 163: We'll need to update the docs-ci.yml workflow to run on push to `release/v*` branches.

devinrsmith (changes requested):

RELEASE.md
stanbrub: We can add "release/v*" to the push trigger, but what about Upload Artifact and Deploy?  

Should it be:
- Upload Artifact on main and release
- Deploy only on release

.github/workflows/docs-ci.yml
stanbrub: Added a ticket for this: https://github.com/deephaven/deephaven-core/issues/4869.  Feel free to change if it's not what you are thinking.

.github/workflows/docs-ci.yml
```diff
@@ -51,14 +51,14 @@ jobs:
...
-        if: ${{ github.ref == 'refs/heads/main' }}
+        if: ${{ (github.ref == 'refs/heads/main') || startsWith(github.ref, 'refs/heads/release/v') }}
         uses: actions/upload-artifact@v3
         with:
           name: javadocs-${{ steps.dhc-version.outputs.version }}
           path: 'combined-javadoc/build/docs/javadoc/'
 
       - name: Deploy Javadoc
-        if: ${{ github.ref == 'refs/heads/main' }}
+        if: ${{ (github.ref == 'refs/heads/main') || startsWith(github.ref, 'refs/heads/release/v') }}
```
devinrsmith, line 61: We should not be _deploying_ on release in this PR; just uploading.

.github/workflows/docs-ci.yml
```diff
@@ -2,7 +2,7 @@ name: Docs CI
 
 on:
   pull_request:
-    branches: [ main ]
```
devinrsmith, line 5: We don't typically do PRs into release/v*; this should be on the `push` side.

RELEASE.md
devinrsmith: It should be "Upload on push to main or release/v*; deploy on push main" until we have a proper directory structure in place (future PR).

devinrsmith (changes requested):

devinrsmith (approved):

PR #4862, opened 2023-11-20
QueryPerformanceRecorder: Changes for CorePlus Integration

nbauernfeind:


rcaudy (approved):

rcaudy (approved):

cpwright:
The performance queries should reintroduce Durations and an IntervalDuration as derived columns.

cpwright (approved):

rcaudy (approved):
This is good. I do want to raise that we're now calculating UPL interval duration by delta of  wall-clock time, rather than delta of "monotonic" time. Are we sure we like this change?

nbauernfeind:
Just to close the loop on the discussion, in case we ever look at this PR again:

We discussed offline that we're happy with duration being computed using wall-clock time given that usage nanos is the sum of [parallelized] System.nanoTime.

Issue #3466, opened 2023-02-28
Simplify version bump


devinrsmith:
Right now, there are a variety of files where the deephaven-core version are specified. After the release process, it's a manual find and replace to update the relevant versions. It would be much improved if we could specify the version once and either have automation update other version numbers as appropriate, or have that single version as the source-of-truth.

mofojed:
One tool that might be handy is [Cocogitto](https://docs.cocogitto.io/). Using that tool in combination with conventional commits (as enforced with #5528) you could automatically bump the version/build release notes.

Also, I would consider doing a branch named `release/v0.33` instead of having branches for each patch release as we do now; as I have a change that I'd like to submit on the v0.33 in case we have any future patches, but I have no branch to submit it on right now. It doesn't make sense (to me) to have a branch for each patch; branch should be for the minor version, then each patch release is tagged at point of release.

Issue #2233, opened 2022-04-08
upgrade setup.py to use setup.cfg, pyproject.toml


jmao-denver:

devinrsmith:
A specific point from a review; this upgrade would allow us to potentially reduce some duplication we currently have among our setup.py.

PR #4865, opened 2023-11-20
Add in-development version support

### LINKED ISSUE #4654: Add in-development friendly versioning scheme
There's potential confusing when our main branch refers to the version of our _next_ release, as opposed to some scheme that better denotes its in-development status.

In practice, Java patterns typically label in-development versions with the `-SNAPSHOT` suffix; for example `0.30.0-SNAPSHOT`. Historically, there were reasons why this didn't work well with our python tooling, but I'm not sure those concerns are applicable anymore.

Python versioning allows for the `.devX` suffix (https://peps.python.org/pep-0440/#public-version-identifiers).

We may want to explore adopting these schemes or something similar to better represent in-development versions.

Related to https://github.com/deephaven/deephaven-core/issues/3704

devinrsmith:
This adds support for in-development versioning of jars and wheels. Specifically, this PR translates the java version qualifier (anything after "-") to a python "local version label" (`<public version identifier>[+<local version label>]`). For example, the java version "1.2.3-SNAPSHOT" becomes python version "1.2.3+snapshot"; the java version "4.5.6-my-custom-build123" becomes python version "4.5.6+my.custom.build123". By using a "local version identifier", these wheels become non-public artifacts - they can't be publicly published. While we could contemplate using "development release segments" (ie, python versios that look like "X.Y.Z.devN"), we wouldn't be able to generally translate java versions to python versions without enforcing more structure onto our java versions.

As an added benefit, this PR also reduces some of the duplication of version numbers by sourcing the version number as appropriate during the wheel building process; and then also in the runtime python code, sourcing the version information via the appropriate importlib metadata APIs.

This also improves our versioning support (now supported via property changes instead of hard-coded buildSrc changes), and gives re-builders options when defining their own version.

See https://packaging.python.org/en/latest/specifications/version-specifiers/#
See https://peps.python.org/pep-0440/

Partial #3466
Fixes #4654

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -247,18 +247,9 @@ mention the version explicitly. These files are listed below:
 
 ```
 #
-# Edit files for version change, updating from 0.31.0 to 0.32.0
+# Edit files for version change, updating from 0.31.0-SNAPSHOT to 0.32.0-SNAPSHOT
 #
-authorization-codegen/protoc-gen-contextual-auth-wiring
-authorization-codegen/protoc-gen-service-auth-wiring
 buildSrc/src/main/groovy/io.deephaven.common-conventions.gradle
-py/client-ticking/README.md
-py/client-ticking/setup.py
-py/client/README.md
-py/client/pydeephaven/__init__.py
-py/client/setup.py
-py/embedded-server/deephaven_server/__init__.py
-py/server/deephaven/__init__.py
 R/rdeephaven/DESCRIPTION
 ```
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -247,9 +247,9 @@ mention the version explicitly. These files are listed below:
 
 ```
 #
-# Edit files for version change, updating from 0.31.0-SNAPSHOT to 0.32.0-SNAPSHOT
+# Edit files for version change
 #
-buildSrc/src/main/groovy/io.deephaven.common-conventions.gradle
+gradle.properties
 R/rdeephaven/DESCRIPTION
 ```
 
--- a/authorization/README.md
+++ b/authorization/README.md
@@ -27,11 +27,12 @@ Here is a sample bash script to generate the provided authorizing wiring if you 
 ```bash
 ./gradlew :authorization-codegen:shadowJar
 
+DEEPHAVEN_VERSION="$(./gradlew printVersion -q)"
 OUT_DIR=authorization/src/main/java/
 PROTO_DIR=proto/proto-backplane-grpc/src/main/proto/
 ROOT_DIR=$PROTO_DIR/deephaven/proto
 
-PATH=authorization-codegen:$PATH protoc --service-auth-wiring_out=$OUT_DIR -I $PROTO_DIR    \
+DEEPHAVEN_VERSION=${DEEPHAVEN_VERSION} PATH=authorization-codegen:$PATH protoc --service-auth-wiring_out=$OUT_DIR -I $PROTO_DIR    \
      $ROOT_DIR/application.proto                                        \
      $ROOT_DIR/console.proto                                            \
      $ROOT_DIR/config.proto                                             \
@@ -41,7 +42,7 @@ PATH=authorization-codegen:$PATH protoc --service-auth-wiring_out=$OUT_DIR -I $P
      $ROOT_DIR/storage.proto                                            \
      $ROOT_DIR/ticket.proto
 
-PATH=authorization-codegen:$PATH protoc --contextual-auth-wiring_out=$OUT_DIR -I $PROTO_DIR \
+DEEPHAVEN_VERSION=${DEEPHAVEN_VERSION} PATH=authorization-codegen:$PATH protoc --contextual-auth-wiring_out=$OUT_DIR -I $PROTO_DIR \
      $ROOT_DIR/table.proto                                              \
      $ROOT_DIR/inputtable.proto                                         \
      $ROOT_DIR/partitionedtable.proto                                   \
@@ -51,6 +52,8 @@ OUT_DIR=authorization/src/main/java/
 PROTO_DIR=../grpc/src/proto/grpc/health/v1/
 ROOT_DIR=$PROTO_DIR
 
-PATH=authorization-codegen:$PATH protoc --service-auth-wiring_out=$OUT_DIR -I $PROTO_DIR    \
+DEEPHAVEN_VERSION=${DEEPHAVEN_VERSION}  PATH=authorization-codegen:$PATH protoc --service-auth-wiring_out=$OUT_DIR -I $PROTO_DIR    \
      $ROOT_DIR/health.proto
+
+./gradlew :authorization:spotlessApply
 ```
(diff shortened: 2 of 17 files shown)


authorization/README.md
```diff
@@ -27,11 +27,12 @@ Here is a sample bash script to generate the provided authorizing wiring if you
 ```bash
 ./gradlew :authorization-codegen:shadowJar
 
+DEEPHAVEN_VERSION="$(./gradlew printVersion -q)"
```
stanbrub, line 30: If a branch has just been checked out, the stdout of the  grade script is not just the version, it's also some other output.

devinrsmith: Interesting; I would have thought that goes to stderr, but I can see this as a potential issue. There is _probably_ some way to disable that if it's true. That said, the previous line `./gradlew :authorization-codegen:shadowJar` _might_ take care of any first-time startup stdout logging?

cpwright:
The properties, extra suffix, and default of using SNAPSHOT should be suitable for Enterprise and eliminate the confusion we have now about whether "0.31.0" is really the released version.

authorization/README.md
stanbrub: Ah, good point.

stanbrub (dismissed):

jmao-denver (approved):
Py changes LGTM

jcferretti (changes requested):
It is not clear to me from the description or the code what are the changes to the release process in terms of tagging before/after release; as of before this PR, the procedure is to edit sources mentioned versions and put the version that will be released next.  This means that when the release comes the version to be released is already in those files.
Since there are no changes in `RELEASE.md` mentioning any of that, it would seem that nothing has changed about that.  But I don't understand how can that be the case given the description.

devinrsmith:
> It is not clear to me from the description or the code what are the changes to the release process in terms of tagging before/after release; as of before this PR, the procedure is to edit sources that mention versions (the list of files to edit is given in `RELEASE.md`) and put in them the version number that will be released "next" (eg, if we just released 0.29.0, the files are edited right after release to point to 0.30.0). This means that when the release comes the version to be released is already in those files, so before a release there is no need to change any versions in any files.
> 
> Since there are no changes in `RELEASE.md` mentioning any of that in this PR, it would seem that nothing has changed about that on this PR. But I don't understand how can that be the case given the description.

There are no changes to the release part of the process - the publish workflow removes the "-SNAPSHOT" suffix automatically. After release, the manual editing of version numbers has been greatly simplified (from ~10 files -> 2 files).

jcferretti (dismissed):

RELEASE.md
```diff
@@ -247,18 +247,9 @@ mention the version explicitly. These files are listed below:
...
-authorization-codegen/protoc-gen-service-auth-wiring
-buildSrc/src/main/groovy/io.deephaven.common-conventions.gradle
-py/client-ticking/README.md
-py/client-ticking/setup.py
-py/client/README.md
-py/client/pydeephaven/__init__.py
-py/client/setup.py
-py/embedded-server/deephaven_server/__init__.py
-py/server/deephaven/__init__.py
+gradle.properties
```
mofojed, line 252: Glorious. Also kind of neat how the example version bump above is 0.31.0 to 0.32.0 and was written a few months ago, which is assumedly going to actually be our next bump after this merges.

py/client-ticking/setup.py
```diff
@@ -1,21 +1,39 @@
...
 from setuptools import find_packages, setup, Extension
 from Cython.Build import cythonize
 
-# The directory containing this file
-HERE = pathlib.Path(__file__).parent
+def _get_readme() -> str:
+    # The directory containing this file
+    HERE = pathlib.Path(__file__).parent
+    # The text of the README file
+    return (HERE / "README.md").read_text()
```
mofojed, line 18: If I'm not mistaken, you should specify utf8 encoding here: https://github.com/pypa/sampleproject/commit/aeeb50a948addcb712ad4261df472263514991e1#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L15

```suggestion
    return (HERE / "README.md").read_text(encoding="utf-8")
```
It probably defaults to utf-8 on most platforms, but I believe it should be explicit to be safe in case the platform locale is different.

py/client/setup.py
```diff
@@ -1,20 +1,36 @@
...
+from pkg_resources import parse_version
 from setuptools import find_packages, setup
 
-# The directory containing this file
-HERE = pathlib.Path(__file__).parent
+def _get_readme() -> str:
+    # The directory containing this file
+    HERE = pathlib.Path(__file__).parent
+    # The text of the README file
+    return (HERE / "README.md").read_text()
```
mofojed, line 17:
```suggestion
    return (HERE / "README.md").read_text(encoding="utf-8")
```

py/embedded-server/setup.py
```diff
@@ -3,33 +3,36 @@
...
+# to use an external library `packaging`. From the context of building a wheel though, we'd prefer to not have to
+# install extra dependencies, at least until we can more properly manage the build environment (pyproject.toml)
+from pkg_resources import parse_version
+from setuptools import find_namespace_packages, setup
 
+def _get_readme() -> str:
+    # The directory containing this file
+    HERE = pathlib.Path(__file__).parent
+    # The text of the README file
+    return (HERE / "README_PyPi.md").read_text()
```
mofojed, line 17:
```suggestion
    return (HERE / "README_PyPi.md").read_text(encoding="utf-8")
```

py/server/setup.py
```diff
@@ -3,33 +3,34 @@
...
-# https://packaging.python.org/en/latest/single_source_version.html
-# todo: does DH versions align w/ PEP440?
-# see https://github.com/pypa/setuptools/blob/v40.8.0/setuptools/dist.py#L470
-def normalize_version(version):
-    return str(packaging.version.Version(version))
+def _get_readme() -> str:
+    # The directory containing this file
+    HERE = pathlib.Path(__file__).parent
+    # The text of the README file
+    return (HERE / "README.md").read_text()
```
mofojed, line 17:
```suggestion
    return (HERE / "README.md").read_text(encoding="utf-8")
```

mofojed (dismissed):
Would be nice if `_get_readme` and `_normalize_version` were somehow shared between projects, but not a big deal.
Just a minor suggestion about being explicit about encoding.

RELEASE.md
devinrsmith: Agreed. I think https://github.com/deephaven/deephaven-core/issues/2233 would allow this to be more easily solved - much easier to specify build related dependencies then.

devinrsmith:
> Would be nice if `_get_readme` and `_normalize_version` were somehow shared between projects, but not a big deal. Just a minor suggestion about being explicit about encoding.

Agreed. I think https://github.com/deephaven/deephaven-core/issues/2233 would allow this to be more easily solved - much easier to specify build related dependencies then.

jcferretti (approved):

jmao-denver (approved):

stanbrub (approved):

mofojed (approved):

PR #4876, opened 2023-11-21
ConstructSnapshot + BarrageMessageProducer: Use Static ConstructSnapshot Fast Path 

nbauernfeind:
This fixes the following query:
```groovy
def static_right = emptyTable(1).update("Test = `test`")
Thread.start {
    result = emptyTable(1000).join(static_right)
}.join()
```

The result table is a static table built without an `ExecutionContext`, as should be allowed. 

However, the captured update graph is the `PoisonedUpdateGraph` which throws in `ConstructSnapshot`. While `ConstructSnapshot` allows for a `null` `UpdateGraph`, the `SnapshotControl` does not "go through the motions" of an otherwise typical snapshot.

Conveniently, this fix will snapshot static tables (even those that capture a valid UpdateGraph) to work their way through `ConstructSnapshot` with less hassle, no locks, and no unnecessary retries.

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/remote/ConstructSnapshot.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/remote/ConstructSnapshot.java
@@ -1173,6 +1173,17 @@ public class ConstructSnapshot {
      * Invokes the snapshot function in a loop until it succeeds with provably consistent results, or until
      * {@code MAX_CONCURRENT_ATTEMPTS} or {@code MAX_CONCURRENT_ATTEMPT_DURATION_MILLIS} are exceeded. Falls back to
      * acquiring a shared update graph lock for a final attempt.
+     * <p>
+     * The supplied {@link SnapshotControl}'s {@link SnapshotControl#usePreviousValues usePreviousValues} will be
+     * invoked at the start of any snapshot attempt, and its {@link SnapshotControl#snapshotCompletedConsistently
+     * snapshotCompletedConsistently} will be invoked at the end of any snapshot attempt that is not provably
+     * inconsistent.
+     * <p>
+     * If the supplied {@link SnapshotControl} provides a null {@link SnapshotControl#getUpdateGraph UpdateGraph}, then
+     * this method will perform a static snapshot without locks or retrying. In this case, the {@link SnapshotControl}'s
+     * {@link SnapshotControl#usePreviousValues usePreviousValues} must return {@code false}, and
+     * {@link SnapshotControl#snapshotCompletedConsistently snapshotCompletedConsistently} must return {@code true}.
+     *
      *
      * @param logPrefix A prefix for our log messages
      * @param control A {@link SnapshotControl} to define the parameters and consistency for this snapshot
@@ -1190,6 +1201,9 @@ public class ConstructSnapshot {
 
         if (updateGraph == null) {
             // This is a snapshot of static data. Just call the function with no frippery.
+            final boolean controlUsePrev = control.usePreviousValues(LogicalClock.NULL_CLOCK_VALUE);
+            Assert.eqFalse(controlUsePrev, "controlUsePrev");
+
             final boolean functionSuccessful = function.call(false, LogicalClock.NULL_CLOCK_VALUE);
             Assert.assertion(functionSuccessful, "functionSuccessful");
             if (log.isDebugEnabled()) {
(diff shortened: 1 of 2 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/remote/ConstructSnapshot.java
```diff
@@ -1173,6 +1173,17 @@ public static long callDataSnapshotFunction(
...
+     * <p>
+     * The supplied {@link SnapshotControl}'s {@link SnapshotControl#usePreviousValues usePreviousValues} will be
+     * invoked at the start of any snapshot attempt, and its {@link SnapshotControl#snapshotCompletedConsistently
+     * snapshotCompletedConsistently} will be invoked at the end of any snapshot attempt that is not provably
+     * inconsistent.
+     * <p>
+     * If the supplied {@link SnapshotControl} provides a null {@link SnapshotControl#getUpdateGraph UpdateGraph}, then
+     * this method will perform a static snapshot without locks or retrying. In this case, the {@link SnapshotControl}'s
+     * {@link SnapshotControl#usePreviousValues usePreviousValues} must return {@code false}, and
+     * {@link SnapshotControl#snapshotCompletedConsistently snapshotCompletedConsistently} must return {@code true}.
```
rcaudy, line 1185: Should we also clarify that the clock value will be NULL?

engine/table/src/main/java/io/deephaven/engine/table/impl/remote/ConstructSnapshot.java
```diff
@@ -1190,13 +1201,21 @@ public static long callDataSnapshotFunction(
 
         if (updateGraph == null) {
             // This is a snapshot of static data. Just call the function with no frippery.
+            final boolean controlUsePrev = control.usePreviousValues(LogicalClock.NULL_CLOCK_VALUE);
+            Assert.eqFalse(controlUsePrev, "controlUsePrev");
```
rcaudy, line 1205: Feels like this should be an ISE with a user-facing error message.

engine/table/src/main/java/io/deephaven/engine/table/impl/remote/ConstructSnapshot.java
```diff
@@ -1190,13 +1201,21 @@ public static long callDataSnapshotFunction(
...
             if (log.isDebugEnabled()) {
                 final long duration = System.currentTimeMillis() - overallStart;
                 log.debug().append(logPrefix)
                         .append(" Static snapshot function elapsed time ").append(duration).append(" ms").endl();
             }
+
+            // notify control of successful snapshot
+            final boolean controlSuccessful =
+                    control.snapshotCompletedConsistently(LogicalClock.NULL_CLOCK_VALUE, false);
+            Assert.assertion(controlSuccessful, "controlSuccessful");
```
rcaudy, line 1218: Same as above.

server/src/main/java/io/deephaven/server/barrage/BarrageMessageProducer.java
```diff
@@ -1488,6 +1501,11 @@ private void updateSubscriptionsSnapshotAndPropagate() {
         }
 
         if (numGrowingSubscriptions > 0) {
+            if (log.isDebugEnabled()) {
+                log.info().append(logPrefix).append("Have ").append(numGrowingSubscriptions)
+                        .append(" growing subscriptions; scheduling next snapshot immediately.").endl();
+            }
+            updatePropagationJob.scheduleImmediately();
```
rcaudy, line 1508: Why have you doubled this up?

server/src/main/java/io/deephaven/server/barrage/BarrageMessageProducer.java
```diff
@@ -2024,6 +2042,11 @@ private void finalizeSnapshotForSubscriptions(final List<Subscription> subscript
                     || subscription.growingRemainingViewport.firstRowKey() >= parentTableSize
                     || isBlinkTable;
 
+            if (log.isDebugEnabled()) {
+                log.debug().append(logPrefix).append("finalizing snapshot for subscription: ")
+                        .append(subscription.toString()).append(", isComplete=").append(isComplete).endl();
```
rcaudy, line 2047: Don't subscriptions have a log prefix?

rcaudy (approved):

PR #4880, opened 2023-11-22
Introduce Blink-To-Append-Only Memoization and Apply to TableLoggers

nbauernfeind:
The primary purpose is to make `PerformanceQueries` work as intended (best-effort w.r.t. memory usage when not in use) according to existing documentation.

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
@@ -15,7 +15,6 @@ import io.deephaven.util.SafeCloseable;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
@@ -148,8 +147,8 @@ public class BlinkTableTools {
 
         @Override
         public Result<QueryTable> initialize(boolean usePrev, long beforeClock) {
-            final Map<String, WritableColumnSource<?>> resultColumns = new LinkedHashMap<>();
-            final Map<String, ? extends ColumnSource<?>> columnSourceMap = parent.getColumnSourceMap();
+            final Map<String, ? extends ColumnSource<?>> sourceColumns = parent.getColumnSourceMap();
+            final Map<String, WritableColumnSource<?>> resultColumns = new LinkedHashMap<>(sourceColumns.size());
 
             // note that we do not need to enable prev tracking for an add-only table
             int colIdx = 0;
@@ -171,7 +170,6 @@ public class BlinkTableTools {
             resultTable.setAttribute(Table.APPEND_ONLY_TABLE_ATTRIBUTE, true);
             resultTable.setFlat();
 
-            // statistically a blink table is likely to be empty; but let's copy whatever we can
             appendRows(usePrev ? parent.getRowSet().prev() : parent.getRowSet(), usePrev).close();
             rowSet.initializePreviousValue();
             Assert.leq(rowSet.size(), "rowSet.size()", sizeLimit, "sizeLimit");
@@ -194,8 +192,7 @@ public class BlinkTableTools {
             if (upstream.modified().isNonempty() || upstream.shifted().nonempty()) {
                 throw new IllegalStateException("Blink tables should not modify or shift!");
             }
-            long newRowsSize = upstream.added().size();
-            if (newRowsSize == 0) {
+            if (upstream.added().isEmpty()) {
                 return;
             }
 
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
@@ -24,18 +24,25 @@ import java.util.Map;
  * @see Table#BLINK_TABLE_ATTRIBUTE
  */
 public class BlinkTableTools {
+    public static final Object DEFAULT_MEMO_KEY = new Object() {
+        @Override
+        public String toString() {
+            return "DEFAULT_MEMOIZATION_KEY";
+        }
+    };
 
     /**
      * Convert a Blink Table to an in-memory append only table.
      * <p>
-     * Note, this table will grow without bound as new blink table rows are encountered.
+     * Note, this table will grow without bound as new blink table rows are encountered. The result is memoized under
+     * {@link #DEFAULT_MEMO_KEY}.
      *
      * @param blinkTable The input blink table
      * @return An append-only in-memory table representing all data encountered in the blink table across all cycles
      */
     public static Table blinkToAppendOnly(final Table blinkTable) {
         // Setting the size limit as maximum allowed value
-        return blinkToAppendOnly(blinkTable, Long.MAX_VALUE);
+        return blinkToAppendOnly(blinkTable, Long.MAX_VALUE, DEFAULT_MEMO_KEY);
     }
 
     /**
@@ -44,7 +51,8 @@ public class BlinkTableTools {
      * Note, this table will grow without bound as new blink table rows are encountered.
      *
      * @param blinkTable The input blink table
-     * @param memoKey A memo key to use to enable memoization (or {@code null} to disable memoization)
+     * @param memoKey saves a weak reference to the result of the given operation under the given memoization key (null
+     *        to disable memoization)
      * @return An append-only in-memory table representing all data encountered in the blink table across all cycles
      */
     public static Table blinkToAppendOnly(@NotNull final Table blinkTable, @Nullable final Object memoKey) {
@@ -55,6 +63,9 @@ public class BlinkTableTools {
     /**
      * Convert a Blink Table to an in-memory append only table with a limit on maximum size. Any updates beyond that
      * limit won't be appended to the table.
+     * <p>
+     * Note, this table will grow without bound as new blink table rows are encountered. The result is memoized under
+     * {@link #DEFAULT_MEMO_KEY}.
      *
      * @param blinkTable The input blink table
      * @param sizeLimit The maximum number of rows in the append-only table
@@ -62,7 +73,7 @@ public class BlinkTableTools {
      *         till maximum row count
      */
     public static Table blinkToAppendOnly(@NotNull final Table blinkTable, long sizeLimit) {
-        return blinkToAppendOnly(blinkTable, sizeLimit, null);
+        return blinkToAppendOnly(blinkTable, sizeLimit, DEFAULT_MEMO_KEY);
     }
 
     /**
@@ -71,7 +82,8 @@ public class BlinkTableTools {
      *
      * @param blinkTable The input blink table
      * @param sizeLimit The maximum number of rows in the append-only table
-     * @param memoKey A memo key to use to enable memoization (or {@code null} to disable memoization)
+     * @param memoKey saves a weak reference to the result of the given operation under the given size limit and
+     *        memoization key (null to disable memoization)
      * @return An append-only in-memory table representing all data encountered in the blink table across all cycles
      *         till maximum row count
      */
@@ -82,14 +94,12 @@ public class BlinkTableTools {
         if (sizeLimit < 0) {
             throw new IllegalArgumentException("Size limit cannot be negative, limit=" + sizeLimit);
         }
+        if (!isBlink(blinkTable)) {
+            throw new IllegalArgumentException("Input is not a blink table!");
+        }
         final UpdateGraph updateGraph = blinkTable.getUpdateGraph();
         try (final SafeCloseable ignored = ExecutionContext.getContext().withUpdateGraph(updateGraph).open()) {
-            final QueryTable coalesced = (QueryTable) blinkTable.coalesce();
-
-            if (!isBlink(coalesced)) {
-                throw new IllegalArgumentException("Input is not a blink table!");
-            }
-
+            final QueryTable coalesced = (QueryTable) (blinkTable.coalesce());
             return coalesced.getResult(new BlinkToAppendOnlyOperation(coalesced, sizeLimit, memoKey));
         }
     }
@@ -132,7 +142,9 @@ public class BlinkTableTools {
 
         @Override
         public String getDescription() {
-            return "BlinkTableTools.blinkToAppendOnly(" + (sizeLimit == Long.MAX_VALUE ? "" : sizeLimit) + ")";
+            final String sizeLimitStr = sizeLimit == Long.MAX_VALUE ? "Long.MAX_VALUE" : Long.toString(sizeLimit);
+            final String memoKeyStr = memoKey == null ? "NO_MEMOIZATION_KEY" : memoKey.toString();
+            return String.format("BlinkTableTools.blinkToAppendOnly(%s, %s)", sizeLimitStr, memoKeyStr);
         }
 
         @Override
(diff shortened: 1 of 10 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -52,122 +62,37 @@ public static Table blinkToAppendOnly(final Table blinkTable) {
...
         try (final SafeCloseable ignored = ExecutionContext.getContext().withUpdateGraph(updateGraph).open()) {
-            return internalBlinkToAppendOnly(blinkTable, sizeLimit);
-        }
-    }
+            final QueryTable coalesced = (QueryTable) blinkTable.coalesce();
 
-    private static Table internalBlinkToAppendOnly(final Table blinkTable, long sizeLimit) {
-        return QueryPerformanceRecorder.withNugget("blinkToAppendOnly", () -> {
-            if (!isBlink(blinkTable)) {
+            if (!isBlink(coalesced)) {
```
devinrsmith, line 90: I see the order has changed here - coalescing _first_ then checking blink. Does it make more sense to keep the check first?

More generally, are we really using coalesce() b/c we need to make sure manage livenessscope stack stuff? If so, are we maybe abusing coalesce in this context b/c we already know a blinktable is coalesced? (And maybe, we should have a more appropriate method on Table for this situation?)

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -183,4 +108,133 @@ public static boolean isBlink(Table table) {
...
+            this.sizeLimit = sizeLimit;
+            this.memoKey = memoKey;
+
+            this.sourceColumns = new ColumnSource<?>[parent.numColumns()];
+            this.destColumns = new WritableColumnSource<?>[parent.numColumns()];
+        }
+
+        @Override
+        public String getDescription() {
+            return "BlinkTableTools.blinkToAppendOnly(" + (sizeLimit == Long.MAX_VALUE ? "" : sizeLimit) + ")";
```
devinrsmith, line 136: Should we include memoKey? String.format may make it easier to read.

engine/table/src/main/java/io/deephaven/engine/table/impl/MemoizedOperationKey.java
```diff
@@ -672,4 +676,38 @@ protected static boolean equalWeakRefsByReferentIdentity(final WeakReference<?>
...
         return t1 == t2;
     }
+
+    private static class BlinkToAppendOnly extends AttributeAgnosticMemoizedOperationKey {
+        private final long sizeLimit;
+        private final Object key;
+
+        private BlinkToAppendOnly(final long sizeLimit, @NotNull final Object key) {
+            this.sizeLimit = sizeLimit;
+            this.key = key;
```
devinrsmith, line 686: Defensive `Objects.requireNonNull` is appropriate IMO.

engine/table/src/main/java/io/deephaven/engine/table/impl/MemoizedOperationKey.java
```diff
@@ -672,4 +676,38 @@ protected static boolean equalWeakRefsByReferentIdentity(final WeakReference<?>
...
+            if (this == other) {
+                return true;
+            }
+            if (other == null || getClass() != other.getClass()) {
+                return false;
+            }
+
+            final BlinkToAppendOnly blinkToAppendOnly = (BlinkToAppendOnly) other;
+
+            return key.equals(blinkToAppendOnly.key);
```
devinrsmith, line 700: Don't we need to check sizeLimit too?

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -183,4 +108,133 @@ public static boolean isBlink(Table table) {
...
+        }
+
+        private void onUpdate(final TableUpdate upstream) {
+            if (upstream.modified().isNonempty() || upstream.shifted().nonempty()) {
+                throw new IllegalStateException("Blink tables should not modify or shift!");
+            }
+            long newRowsSize = upstream.added().size();
+            if (newRowsSize == 0) {
+                return;
+            }
```
rcaudy, line 200:
```suggestion
            if (upstream.added().isEmpty()) {
                return;
            }
```

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -183,4 +108,133 @@ public static boolean isBlink(Table table) {
...
+
+            final TableUpdateImpl downstream = new TableUpdateImpl();
+            downstream.added = appendRows(upstream.added(), false);
```
rcaudy, line 203: You should assert that this is non-empty.

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -183,4 +108,133 @@ public static boolean isBlink(Table table) {
...
+            resultTable.notifyListeners(downstream);
+
+            if (resultTable.size() == sizeLimit) {
+                // No more rows can be appended, so remove the listener and remove all references
+                resultListener.forceReferenceCountToZero();
+                resultListener = null;
+            }
+        }
+
+        private RowSet appendRows(final RowSet newRows, final boolean usePrev) {
```
rcaudy, line 217: * This could easily accept the `currentSize` as an argument.
* That would let you defer creating the result table until after the `copyData`.
* That, in turn, would let you avoid making the result row set tracking prematurely, avoid the need to use `initializePreviousValue` (which is almost always a code smell), and let you always use the returned `newRange`, because you can just convert it to tracking to be the initial row set, and insert it *outside* of this method in the updating case.
* (Note that you can do the `insertRange` using `newRange.firstRowKey()` and `newRange.lastRowKey()`.)

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
nbauernfeind: Good catch. That is my mistake. I needed the QueryTable to invoke `getResult` and I got the order wrong.

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -183,4 +108,133 @@ public static boolean isBlink(Table table) {
...
+            resultTable.setAttribute(Table.APPEND_ONLY_TABLE_ATTRIBUTE, true);
+            resultTable.setFlat();
+
+            // statistically a blink table is likely to be empty; but let's copy whatever we can
+            appendRows(usePrev ? parent.getRowSet().prev() : parent.getRowSet(), usePrev).close();
+            rowSet.initializePreviousValue();
+            Assert.leq(rowSet.size(), "rowSet.size()", sizeLimit, "sizeLimit");
+
+            if (resultTable.size() < sizeLimit) {
+                resultTable.setRefreshing(true);
```
rcaudy, line 180: It should never be ambiguous whether we will get a refreshing result. This should not be conditional.

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -183,4 +108,133 @@ public static boolean isBlink(Table table) {
...
+
+                resultListener = new BaseTable.ListenerImpl(getDescription(), parent, resultTable) {
```
rcaudy, line 182: I understand why you might choose to make this conditional. Seems OK.

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -183,4 +108,133 @@ public static boolean isBlink(Table table) {
...
+                destColumns[colIdx++] = ReinterpretUtils.maybeConvertToWritablePrimitive(newColumn);
+            }
+
+            final TrackingWritableRowSet rowSet = RowSetFactory.empty().toTracking();
+            resultTable = new QueryTable(rowSet, resultColumns);
+            resultTable.setAttribute(Table.ADD_ONLY_TABLE_ATTRIBUTE, true);
+            resultTable.setAttribute(Table.APPEND_ONLY_TABLE_ATTRIBUTE, true);
+            resultTable.setFlat();
+
+            // statistically a blink table is likely to be empty; but let's copy whatever we can
```
rcaudy, line 174: "statistically"? This is kind of a silly comment.
```suggestion
```

rcaudy (commented):
Mostly looking good. I wish you had separated out some of the refactoring to its own commit.

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -116,7 +116,7 @@ default boolean snapshotNeeded() {
          */
         class Result<T extends DynamicNode & NotificationStepReceiver> {
             public final T resultNode;
-            public final TableUpdateListener resultListener; // may be null if parent is non-ticking
+            public final TableUpdateListener resultListener; // may be null if parent or result are non-refreshing
```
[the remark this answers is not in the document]
rcaudy: I see what you had in mind. I think this comment should be revised, but I can accept that there's a case for no listener.

engine/table/src/main/java/io/deephaven/engine/table/impl/OperationSnapshotControl.java
```diff
@@ -132,7 +129,7 @@ public synchronized boolean snapshotCompletedConsistently(
 
         // Be sure to record initial last notification step before subscribing
         eventualResult.setLastNotificationStep(lastNotificationStep);
-        return subscribeForUpdates(eventualListener);
+        return eventualListener == null || subscribeForUpdates(eventualListener);
```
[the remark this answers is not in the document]
rcaudy: I understand now.

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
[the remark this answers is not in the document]
nbauernfeind: In this case .. the append-only table could hit its size limit during initialization. Maybe this is an over optimization, but it does allow the result to be static. I can't think of anything other than a snapshot.

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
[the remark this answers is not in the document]
nbauernfeind: Ack.

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -183,4 +117,134 @@ public static boolean isBlink(Table table) {
...
+            this.sizeLimit = sizeLimit;
+            this.memoKey = memoKey;
+
+            this.sourceColumns = new ColumnSource<?>[parent.numColumns()];
+            this.destColumns = new WritableColumnSource<?>[parent.numColumns()];
+        }
+
+        @Override
+        public String getDescription() {
+            final String sizeLimitStr = sizeLimit == Long.MAX_VALUE ? "Long.MAX_VALUE" : Long.toString(sizeLimit);
```
rcaudy, line 145: Why not say "unbounded" or "unlimited" for max?

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -183,4 +117,134 @@ public static boolean isBlink(Table table) {
...
+            final String memoKeyStr = memoKey == null ? "NO_MEMOIZATION_KEY" : memoKey.toString();
```
rcaudy, line 146: Similarly "none"?

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -183,4 +117,134 @@ public static boolean isBlink(Table table) {
...
+            int colIdx = 0;
+            for (Map.Entry<String, ? extends ColumnSource<?>> sourceEntry : sourceColumns
+                    .entrySet()) {
+                final ColumnSource<?> sourceColumn = sourceEntry.getValue();
+                final WritableColumnSource<?> newColumn = ArrayBackedColumnSource.getMemoryColumnSource(
+                        0, sourceColumn.getType(), sourceColumn.getComponentType());
+                resultColumns.put(sourceEntry.getKey(), newColumn);
+
+                // read and write primitives whenever possible
+                this.sourceColumns[colIdx] = ReinterpretUtils.maybeConvertToPrimitive(sourceColumn);
```
rcaudy, line 175: I'm sorry I suggested the previous rename, now that I realize this consequence. Maybe I should have suggested parentColumns?

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -183,4 +117,134 @@ public static boolean isBlink(Table table) {
...
+            final RowSet newRange = RowSetFactory.fromRange(currentSize, totalSize - 1);
+
+            try (final SafeCloseable ignored = rowsToAdd) {
+                if (rowsToAdd == null) {
+                    rowsToAdd = newRows;
+                }
+                ChunkUtils.copyData(sourceColumns, rowsToAdd, destColumns, newRange, usePrev);
+            }
+            if (resultTable != null) {
+                resultTable.getRowSet().writableCast().insertRange(newRange.firstRowKey(), newRange.lastRowKey());
```
rcaudy, line 243: I would just do this unconditionally in the listener, instead of in this method. Cleaner, and separates concerns.

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -26,150 +24,86 @@
  * @see Table#BLINK_TABLE_ATTRIBUTE
  */
 public class BlinkTableTools {
+    public static final Object DEFAULT_MEMO_KEY = new Object() {
+        @Override
+        public String toString() {
+            return "DEFAULT_MEMOIZATION_KEY";
+        }
+    };
```
rcaudy, line 32: Could just be
```
new String("DEFAULT_MEMOIZATION_KEY");
```

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -26,150 +24,86 @@
...
+    public static Table blinkToAppendOnly(@NotNull final Table blinkTable, @Nullable final Object memoKey) {
+        // Setting the size limit as maximum allowed value
+        return blinkToAppendOnly(blinkTable, Long.MAX_VALUE, memoKey);
     }
 
     /**
      * Convert a Blink Table to an in-memory append only table with a limit on maximum size. Any updates beyond that
      * limit won't be appended to the table.
+     * <p>
+     * Note, this table will grow without bound as new blink table rows are encountered. The result is memoized under
```
rcaudy, line 67: This is incorrect; this is size limited.

rcaudy (commented):
.

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
devinrsmith: I'm a fan of how it's setup here w/ Object; even though unlikely, a string-based memo-key is "public" and the user can just use the same string.

rcaudy: I think that's an argument for using object identity when comparing memo keys. Also, why is it bad if the user really wants to explicitly use our default memo key? It's public here, for some reason, although I'd think it could be private.

devinrsmith: Oh, I see it's explicitly public. 

I don't think we'd want to use object identity for all memo, otherwise it wouldn't be obvious how one achieves memo over API today.

rcaudy: OK. I'm convinced enough. We stick to Object equality, keep this default as an Object so it will rely on reference equality, and maybe make it private (but I'm ambivalent about that).

rcaudy (approved):

engine/table/src/main/java/io/deephaven/engine/table/impl/BlinkTableTools.java
```diff
@@ -26,21 +24,55 @@
...
+            return "DEFAULT_MEMOIZATION_KEY";
+        }
+    };
 
     /**
      * Convert a Blink Table to an in-memory append only table.
      * <p>
-     * Note, this table will grow without bound as new blink table rows are encountered.
+     * Note, this table will grow without bound as new blink table rows are encountered. The result is memoized under
+     * {@link #DEFAULT_MEMO_KEY}.
```
devinrsmith, line 38: If we're going to link to it, it _should_ be public IMO. I think it's _ok_ to have it public.

devinrsmith (changes requested):

devinrsmith (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3488
Reference: https://github.com/deephaven/deephaven.io/issues/3487

PR #4885, opened 2023-11-27
Collect JsPlugin instances before starting Jetty

niloc132:
This reverses a DI dependency, preventing requiring that the server start up before plugins can be collected.

Partial #4040

### DIFF
--- a/server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
+++ b/server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
@@ -3,7 +3,6 @@
  */
 package io.deephaven.server.jetty;
 
-import io.deephaven.plugin.js.JsPlugin;
 import io.deephaven.server.browserstreaming.BrowserStreamInterceptor;
 import io.deephaven.server.runner.GrpcServer;
 import io.deephaven.ssl.config.CiphersIntermediate;
@@ -73,7 +72,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Consumer;
 import java.util.function.Supplier;
 
 import static io.grpc.servlet.web.websocket.MultiplexedWebSocketServerStream.GRPC_WEBSOCKETS_MULTIPLEX_PROTOCOL;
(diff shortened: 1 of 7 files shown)


server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
```diff
@@ -85,20 +83,15 @@ public class JettyBackedGrpcServer implements GrpcServer {
...
     private final Server jetty;
-    private final JsPlugins jsPlugins;
     private final boolean websocketsEnabled;
 
     @Inject
     public JettyBackedGrpcServer(
             final JettyConfig config,
-            final GrpcFilter filter) {
+            final GrpcFilter filter,
+            final JsPlugins jsPlugins) {
```
devinrsmith, line 92: This adds a package private class to a public constructor, essentially rendering it non-public to anybody but us. Given the way we construct JettyBackedGrpcServer, it's probably reasonable to make this constructor package private. The other option is to make `JsPlugins` public.

I wish there were a way to force java compilation to fail in these cases?

devinrsmith: image

server/jetty/src/main/java/io/deephaven/server/jetty/JettyServerModule.java
```diff
@@ -67,9 +67,15 @@ static ServletAdapter provideGrpcServletAdapter(
...
     }
 
+    @Binds
+    JsPluginRegistration bindJsPlugins(JsPlugins plugins);
+
     @Provides
-    @Named(PluginsModule.JS_PLUGIN_CONSUMER_NAME)
-    static Consumer<JsPlugin> providesJsPluginConsumer(JettyBackedGrpcServer server) {
-        return server.jsPluginConsumer();
+    static JsPlugins providesJsPluginRegistration() {
```
devinrsmith, line 74: These also add `JsPlugins` to the public API.

server/jetty/src/main/java/io/deephaven/server/jetty/JettyBackedGrpcServer.java
niloc132: I bet https://errorprone.info/ can make this fail (or at least has an option for it).

Agreed, I can change this (as discussed briefly in slack) in either direction. I'd lean towards both being public, do you have a preferece?

server/jetty/src/main/java/io/deephaven/server/jetty/JettyServerModule.java
niloc132: Ok, seems like making it public is the "better" option?

devinrsmith (approved):

PR #4887, opened 2023-11-28
Add notes on releasing jsapi-types

niloc132:
Note that this isn't yet part of CI, either for nightly builds or for tagged releases.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -64,7 +64,11 @@ and are published as the following:
 * [C++ Client API](https://deephaven.io/core/client-api/cpp/)
 * [C++ Examples](https://deephaven.io/core/client-api/cpp-examples/)
 * [R Client API](https://deephaven.io/core/client-api/r/)
-* [TypeScript Client API](https://deephaven.io/core/client-api/javascript/)
+* [JavaScript/TypeScript Client API](https://deephaven.io/core/client-api/javascript/)
+
+### Deephaven JS API types
+The npm package `@deephaven/jsapi-types` allow TypeScript clients to use their IDE's autocomplete and compiler's typechecks.
+It is released to [npm](https://www.npmjs.com/package/@deephaven/jsapi-types).
 
 ## Release process
 
@@ -265,6 +269,7 @@ mention the version explicitly. These files are listed below:
 #
 gradle.properties
 R/rdeephaven/DESCRIPTION
+web/client-api/types/package.json
 ```
 
 This leaves the files "ready" for the next regular release, and also ensures any build done from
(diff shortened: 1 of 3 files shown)


RELEASE.md
```diff
@@ -265,6 +269,7 @@ mention the version explicitly. These files are listed below:
 #
 gradle.properties
 R/rdeephaven/DESCRIPTION
+web/client-api/types/package.json
```
devinrsmith, line 272: Does package-lock.json need to be updated too?

niloc132: I don't _think_ so, since I never had it in there previously, but I used the `npm version` command to perform this particular update and it emitted that. @mofojed can you weigh in?

mofojed: Yes, otherwise the next `npm install` will cause a `package-lock.json` diff (when it adds the version info to it). `package-lock.json` shouldn't typically be modified by hand though...
Side note - what version of `node`/`npm` are you using Colin? Doesn't seem to match the engines specified in the `package.json` (`"node": ">=16"`). We should update that version in https://github.com/deephaven/deephaven-core/blob/e09bcfdaca4264ecd06a73b0e4acab5f850e88ff/docker/registry/node/gradle.properties#L2 (probably use `node:18` now instead of 14).
We should have a [.nvmrc](https://github.com/deephaven/web-client-ui/blob/main/.nvmrc) file defining the node version of an [.npmrc](https://github.com/deephaven/web-client-ui/blob/main/.npmrc) specifying the engine version should be strict and enable provenance as well, so we can sign the builds using GitHub Actions: https://github.blog/2023-04-19-introducing-npm-package-provenance/
I guess then we should just have a gradle action that runs the `npm version`?

niloc132: I used npm 10.2.3 locally to run the `npm version` command that wrote that update to the package-lock.json, and must have created the file using the older node/npm from the container, since it produces a different package-lock.json output if i run it again outside that image.

If you're suggesting running npm version from CI, wouldn't that imply either creating a commit, or not updating the version itself?

Also, does provenance buy us anything here, given that it is just the .d.ts file?

niloc132: Bumping the node image breaks the raw-js-openapi build, i'll file an issue and get that worked out separately. I'll back out the `version` change to package-lock.json for the moment, and work out the node bump and other CI actions later?

niloc132: https://github.com/deephaven/deephaven-core/issues/4888 to follow up on the node image and updating these js projects.

devinrsmith (dismissed):

RELEASE.md
mofojed: Looks like this fixes #4888 as well now? Looks good to me, but the instructions should be to run the a gradle bump action instead of manually editing this package.json file

mofojed (commented):
Changes look good,

niloc132:
Removed last commit, will make a separate PR for that after the release.

devinrsmith (approved):

PR #4889, opened 2023-11-28
Address PartitionedTable transform and partitionedTransform dependency issues

### LINKED ISSUE #4881: PartitionedTable transforms need a way to depend on additional inputs for upstream satisfaction
If I want to use an additional `Table` or similar as input to a `PartitionedTable.transform` or `PartitionedTable.partitionedTransform`, there's currently no good way to make sure that new constituents are safely instantiated.

Expert users can work around by:
1. `copy()`ing the `PartitionedTable.table()`
2. adding a parent reference to the desired dependency
3. re-wrapping the result as a `PartitionedTable`
4. applying the desired transformation

rcaudy:
Fixes #4881

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/partitioned/PartitionedTableProxyImpl.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/partitioned/PartitionedTableProxyImpl.java
@@ -13,20 +13,18 @@ import io.deephaven.api.updateby.UpdateByOperation;
 import io.deephaven.api.updateby.UpdateByControl;
 import io.deephaven.engine.context.ExecutionContext;
 import io.deephaven.engine.liveness.LivenessArtifact;
+import io.deephaven.engine.liveness.LivenessScopeStack;
+import io.deephaven.engine.table.*;
 import io.deephaven.engine.table.impl.MatchPair;
-import io.deephaven.engine.table.PartitionedTable;
-import io.deephaven.engine.table.Table;
-import io.deephaven.engine.table.TableDefinition;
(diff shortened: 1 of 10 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/partitioned/PartitionedTableProxyImpl.java
```diff
@@ -553,11 +573,28 @@ public PartitionedTable.Proxy aggAllBy(AggSpec spec, ColumnName... groupByColumn
...
+            return basicTransform(
+                    true,
+                    ct -> ct.aggBy(aggregations, preserveEmpty, null, groupByColumns));
+        }
+        if (initialGroups instanceof Table) {
+            // Force a consistent view of initial groups table to be used for all current and future constituents
+            final Table initialGroupsTable = LivenessScopeStack.computeEnclosed(
+                    () -> ((Table) initialGroups).selectDistinct(groupByColumns).snapshot(),
+                    () -> ((Table) initialGroups).isRefreshing(),
+                    Table::isRefreshing);
```
cpwright, line 585: So how could this actually do anything other than return false (definitionally basd on the snapshot()?

cpwright (approved):

engine/table/src/main/java/io/deephaven/engine/table/impl/partitioned/PartitionedTableProxyImpl.java
rcaudy: It could't. I just felt like it was better to go through the motions, especially in case the code is ever copy-pasted.

PR #4891, opened 2023-11-29
Add the optional dependencies parameters in the transform/partitioned_transform methods

### LINKED ISSUE #4890: Add the new 'dependencies' parameter to the Python transform/partitioned_transform wrapper methods

jmao-denver:
Fixes #4890

### DIFF
--- a/py/server/deephaven/table.py
+++ b/py/server/deephaven/table.py
@@ -2673,8 +2673,8 @@ class PartitionedTable(JObjectWrapper):
             func (Callable[[Table], Table]): a function which takes a Table as input and returns a new Table
             dependencies (Optional[Sequence[Union[Table, PartitionedTable]]]): additional dependencies that must be
                 satisfied before applying the provided transform function to added or modified constituents during
-                update processing, use this when the transform function uses additional Table or Partitioned Table
-                inputs besides the constituents of this PartitionedTable. Defaults to None.
+                update processing. If the transform function uses any other refreshing Table or refreshing Partitioned
+                Table, they must be included in this argument. Defaults to None.
 
         Returns:
             a PartitionedTable
@@ -2685,8 +2685,8 @@ class PartitionedTable(JObjectWrapper):
         try:
             j_operator = j_unary_operator(func, dtypes.from_jtype(Table.j_object_type.jclass))
             dependencies = to_sequence(dependencies, wrapped=True)
-            j_dependencies = [d.j_table for d in dependencies if isinstance(d, Table)]
-            j_dependencies.extend([d.table.j_table for d in dependencies if isinstance(d, PartitionedTable)])
+            j_dependencies = [d.j_table for d in dependencies if isinstance(d, Table) and d.is_refreshing]
+            j_dependencies.extend([d.table.j_table for d in dependencies if isinstance(d, PartitionedTable) and d.is_refreshing])
             with auto_locking_ctx(self, *dependencies):
                 j_pt = self.j_partitioned_table.transform(j_operator, j_dependencies)
                 return PartitionedTable(j_partitioned_table=j_pt)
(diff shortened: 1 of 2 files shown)


py/server/deephaven/table.py
```diff
@@ -2679,13 +2684,18 @@ def transform(self, func: Callable[[Table], Table]) -> PartitionedTable:
         """
         try:
             j_operator = j_unary_operator(func, dtypes.from_jtype(Table.j_object_type.jclass))
-            with auto_locking_ctx(self):
-                j_pt = self.j_partitioned_table.transform(j_operator)
+            dependencies = to_sequence(dependencies, wrapped=True)
+            j_dependencies = [d.j_table for d in dependencies if isinstance(d, Table)]
+            j_dependencies.extend([d.table for d in dependencies if isinstance(d, PartitionedTable)])
```
rcaudy, line 2689:
```suggestion
            j_dependencies.extend([d.table.j_table for d in dependencies if isinstance(d, PartitionedTable)])
```

py/server/tests/test_partitioned_table.py
```diff
@@ -128,6 +129,10 @@ def test_transform(self):
             pt = self.partitioned_table.transform(Transformer)
             self.assertIn("f", [col.name for col in pt.constituent_table_columns])
 
+            with shared_lock(self.test_table):
```
rcaudy, line 132: Why wasn't the auto-locking context good enough for this?

jmao-denver: Because the test table is not ticking I think.

rcaudy: OK, I understand the issue. My code expects you to only add extra dependencies if they are refreshing.

rcaudy (approved):

py/server/deephaven/table.py
```diff
@@ -2663,13 +2663,18 @@ def constituent_tables(self) -> List[Table]:
...
         """Apply the provided function to all constituent Tables and produce a new PartitionedTable with the results
         as its constituents, with the same data for all other columns in the underlying partitioned Table. Note that
         if the Table underlying this PartitionedTable changes, a corresponding change will propagate to the result.
 
         Args:
             func (Callable[[Table], Table]): a function which takes a Table as input and returns a new Table
+            dependencies (Optional[Sequence[Union[Table, PartitionedTable]]]): additional dependencies that must be
+                satisfied before applying the provided transform function to added or modified constituents during
+                update processing, use this when the transform function uses additional Table or Partitioned Table
+                inputs besides the constituents of this PartitionedTable. Defaults to None.
```
chipkent, line 2677: This will not be understandable to a user:
```
additional dependencies that must be
                satisfied before applying the provided transform function to added or modified constituents during
                update processing
```

py/server/deephaven/table.py
```diff
@@ -2679,13 +2684,18 @@ def transform(self, func: Callable[[Table], Table]) -> PartitionedTable:
         """
         try:
             j_operator = j_unary_operator(func, dtypes.from_jtype(Table.j_object_type.jclass))
-            with auto_locking_ctx(self):
-                j_pt = self.j_partitioned_table.transform(j_operator)
+            dependencies = to_sequence(dependencies, wrapped=True)
+            j_dependencies = [d.j_table for d in dependencies if isinstance(d, Table)]
+            j_dependencies.extend([d.table.j_table for d in dependencies if isinstance(d, PartitionedTable)])
```
chipkent, line 2689: There should be a size assert because this will happily run if non Table / PartitionedTable garbage is included.

py/server/deephaven/table.py
```diff
@@ -2707,8 +2721,12 @@ def partitioned_transform(self, other: PartitionedTable, func: Callable[[Table,
         """
         try:
             j_operator = j_binary_operator(func, dtypes.from_jtype(Table.j_object_type.jclass))
-            with auto_locking_ctx(self, other):
-                j_pt = self.j_partitioned_table.partitionedTransform(other.j_partitioned_table, j_operator)
+            dependencies = to_sequence(dependencies, wrapped=True)
+            j_dependencies = [d.j_table for d in dependencies if isinstance(d, Table)]
+            j_dependencies.extend([d.table.j_table for d in dependencies if isinstance(d, PartitionedTable)])
```
chipkent, line 2726: See other comment on assertion.

py/server/tests/test_partitioned_table.py
chipkent: Seems like we want a test with a live dependency to make sure that the auto-locking works as expected.

py/server/deephaven/table.py
jmao-denver: It might be an overkill, I think it is safe to assume that users who need to use this parameter knows what they are doing.

rcaudy: Will Python's type system (hah) allow that?

rcaudy (approved):

chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3496
Reference: https://github.com/deephaven/deephaven.io/issues/3495

PR #4907, opened 2023-12-01
Deploy `@deephaven/jsapi-types` during release, provide tools for nightly builds

### LINKED ISSUE #4888: Update node image to 20
We're using 14 now, which is EOL. 16 and 18 are "maintenance", and 20 just entered LTS, so we should probably move to 20.

This will require updating our proto:raw-js-openapi build, and the web:client-api:types as well. Both should have .npmrc and .nvmrc files added to specify expected node/npm versions.

### LINKED ISSUE #5081: Autogenerated JS API type necessary fixes
Changes needed to work with my branch (https://github.com/mattrunyon/web-client-ui/tree/jsapi-types) updating web-client-ui to use the generated JS API types

- [x] Add "types": "dist/index.d.ts" to package.json
- [x] Remove (or change to something like >=16) Node engines from package.json
- [x] `LayoutHints.frontColumns` should be nullable (back/frozen/hidden are nullable)
- [x] `IdeConnection` is missing the `HasEventHandling` methods in the generated types
- [x] `TableData` and `SubscriptionData` should return `Array<Row>` (I know there's some Java generics issue, but even a non-generic version works for TS sake)

niloc132:
Like other packages, the npmjs package for jsapi types will now get its version assigned as it is built from gradle. Also reads properties from gradle to specify `--preid` to npm (e.g. `-PnpmPreid=nightly`), to be given a date stamp suffix.

Public job for CI will now deploy tagged versions to npmjs.

Fixes #4888
Fixes #5081


niloc132:
Marked as draft until 0.31 goes out.

Two attached example builds for manual review:

"beta" for a nightly builds:
```
./gradlew :web-client-api:types:typedoc -PnpmPreid=beta
```
[deephaven-jsapi-types-1.0.0-dev0.31.0beta20231201.tgz](https://github.com/deephaven/deephaven-core/files/13531702/deephaven-jsapi-types-1.0.0-dev0.31.0beta20231201.tgz)

no "preid" for a release:
```
./gradlew :web-client-api:types:typedoc
```
[deephaven-jsapi-types-1.0.0-dev0.31.0.tgz](https://github.com/deephaven/deephaven-core/files/13531710/deephaven-jsapi-types-1.0.0-dev0.31.0.tgz)

.github/workflows/publish-ci.yml
```diff
@@ -13,6 +13,7 @@ concurrency:
 jobs:
   publish:
     runs-on: ubuntu-22.04
+    if: ${{ github.repository_owner == 'deephaven' }}
```
niloc132, line 16: Added because I dont think we want to attempt any deploys for tags that exist on forks

.github/workflows/publish-ci.yml
```diff
@@ -98,3 +106,9 @@ jobs:
           user: __token__
           password: ${{ secrets.PYDEEPHAVEN_PYPI_TOKEN }}
           packages_dir: py/client/build/wheel/
+
+      - name: Publish @deephaven/jsapi-types to npmjs
```
niloc132, line 110: Note that I haven't tested this, all lifted from web-client-ui

web/client-api/types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@deephaven/jsapi-types",
-  "version": "1.0.0-dev0.31.0",
+  "version": "0.0.0-dummy",
```
niloc132, line 3: explicit dummy version to be clear that developers need not update this, it will automatically be updated in the gradle+docker wiring

.github/workflows/publish-ci.yml
```diff
@@ -39,23 +40,29 @@ jobs:
           .github/scripts/gradle-properties.sh >> gradle.properties
           cat gradle.properties
 
+      - name: Setup Node for deployment to npmjs
+        uses: actions/setup-node@v3
```
devinrsmith, line 44: I don't have insights into specifics, but v4 is the latest.

.github/workflows/publish-ci.yml
```diff
@@ -98,3 +106,9 @@ jobs:
           user: __token__
           password: ${{ secrets.PYDEEPHAVEN_PYPI_TOKEN }}
           packages_dir: py/client/build/wheel/
+
+      - name: Publish @deephaven/jsapi-types to npmjs
+        if: ${{ startsWith(github.ref, 'refs/heads/release/v') }}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.DEEPHAVENBOT_NPM_TOKEN }}
+        run: npm publish --tag latest web/client-api/types/build/deephaven-jsapi-types-*.tgz
```
devinrsmith, line 114: Does it make sense to add this to our GH release asset? If so, can you please update RELEASE.md (I think there will be a couple places) and make sure to note the updated procedure to `#dhc-release` channel?

.github/workflows/publish-ci.yml
```diff
@@ -39,23 +40,29 @@ jobs:
...
 
+      - name: Setup Node for deployment to npmjs
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: 'web/client-api/types/.nvmrc'
+          registry-url: 'https://registry.npmjs.org'
+
       # TODO(deephaven-core#2614): Improve gradle/CI assemble and publishing of artifacts
 
       - name: Publish to Maven Local
```
devinrsmith, line 51: I'd go ahead and update this name in the same way you've updated the latter.

.github/workflows/publish-ci.yml
stanbrub: I'm not opposed, but it would have to be commented out to test _Publish to Maven Local_ and _Upload Artifacts_ steps.  Maybe it would be clearer with a global variable like _allowPublish_ applied to just the steps that need the protection?

proto/raw-js-openapi/package.json
```diff
@@ -9,10 +9,11 @@
   "devDependencies": {
     "ts-loader": "^6.0.1",
     "typescript": "3.9.7",
-    "webpack": "^5.38.1",
+    "webpack": "^5.76.0",
```
stanbrub, line 12: Forgive my dumb questions here, but why 5.76.0 and not one of the patches?

web/client-api/types/package.json
```diff
@@ -10,9 +10,6 @@
     "url": "https://github.com/deephaven/deephaven-core.git",
     "directory": "web/client-api/types"
   },
-  "engines": {
-    "node": ">=16"
-  },
   "devDependencies": {
     "typedoc": "^0.24.8",
     "typescript": "^5.1.6"
```
stanbrub, line 15: Why is this version different than the one in _proto/raw-js-openapi/package.json_?

proto/raw-js-openapi/package.json
niloc132: Only because my experience with these tools is limited, and the output does what we need with this version.

web/client-api/types/package.json
niloc132: I'm not going to go so far as to say that I fear change, but I haven't kept up with the changes, and the "inputs" into this typescript process are generated by a tool that are no longer getting updates.

.github/workflows/publish-ci.yml
niloc132: I think I've updated those sections, see if I missed anything?

.github/workflows/publish-ci.yml
```diff
@@ -98,3 +105,9 @@ jobs:
           user: __token__
           password: ${{ secrets.PYDEEPHAVEN_PYPI_TOKEN }}
           packages_dir: py/client/build/wheel/
+
+      - name: Publish @deephaven/jsapi-types to npmjs
+        if: ${{ startsWith(github.ref, 'refs/heads/release/v') }}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.DEEPHAVENBOT_NPM_TOKEN }}
```
devinrsmith, line 112: We'll (I can do this) need to provide deephaven-core repo access to this secret before this action will succeed; I don't know how the npmjs secret management works, but I'm assuming this is a "heavy hammer", but we are okay w/ it @mofojed? The alternative would be a secret that is scoped explicitly for `@deephaven/jsapi-types`.

.github/workflows/publish-ci.yml
```diff
@@ -39,23 +39,29 @@ jobs:
...
       # TODO(deephaven-core#2614): Improve gradle/CI assemble and publishing of artifacts
 
-      - name: Publish to Maven Local
+      - name: Build all artifacts, publish to Maven Local
         if: ${{ !startsWith(github.ref, 'refs/heads/release/v') }}
         uses: burrunan/gradle-cache-action@v1
         with:
           job-id: publish-local
-          arguments: server-netty-app:build server-jetty-app:build py-server:build py-embedded-server:build py-client:build publishToMavenLocal
+          arguments: server-netty-app:build server-jetty-app:build py-server:build py-embedded-server:build py-client:build web-client-api:types:typedoc publishToMavenLocal
```
devinrsmith, line 55: I think `assemble.dependsOn(typedoc)` should be added to the web-client-api:types build logic; in which case, this could be updated to `web-client-api:types:build`.

.github/workflows/publish-ci.yml
```diff
@@ -39,23 +39,29 @@ jobs:
...
 
-      - name: Publish
+      - name: Build all artifacts, publish to Sonatype for staging to Maven Central
         if: ${{ startsWith(github.ref, 'refs/heads/release/v') }}
         uses: burrunan/gradle-cache-action@v1
         with:
           job-id: publish
           # We need to be explicit here about no parallelism to ensure we don't create disjointed staging repositories.
-          arguments: --no-parallel server-netty-app:build server-jetty-app:build py-server:build py-client:build py-embedded-server:build publish
+          arguments: --no-parallel server-netty-app:build server-jetty-app:build py-server:build py-client:build py-embedded-server:build web-client-api:types:typedoc publish
```
devinrsmith, line 64: ditto

mattrunyon:
Builds and passes tests in my web-client-ui branch that consumes these types. Output looks good from the web side

.github/workflows/publish-ci.yml
mofojed: Man I wish npmjs had [trusted publishers](https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/) like PyPi does. Way better workflow.

Yes, the `DEEPHAVENBOT_NPM_TOKEN` is not scoped to a particular package, it is a "heavy hammer". You can create granular tokens that are scoped to specific packages, but I think this token scoped to our organization should be fine (organization wide secret) as it would be a pain to have to create new tokens for each package. Or we can have different tokens for the web-client-ui repo, deephaven-plugins repo, and whatever is published by deephaven-core, but then we need to update those tokens anytime we want to publish a new package in that repo (which happens from time to time in both web-client-ui and deephaven-plugins).
https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry

devinrsmith: Done - secret now accessible from deephaven-core.

.github/workflows/publish-ci.yml
```diff
@@ -98,3 +105,9 @@ jobs:
           user: __token__
           password: ${{ secrets.PYDEEPHAVEN_PYPI_TOKEN }}
           packages_dir: py/client/build/wheel/
+
+      - name: Publish @deephaven/jsapi-types to npmjs
+        if: ${{ startsWith(github.ref, 'refs/heads/release/v') }}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.DEEPHAVENBOT_NPM_TOKEN }}
+        run: npm publish --tag latest web/client-api/types/build/deephaven-jsapi-types-*.tgz
```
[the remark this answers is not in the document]
niloc132: Yes - this is the way to say "here's a finished tarball, go ahead and deploy it". The work of making the tarball is done inside of docker (so that we guarantee it builds the way we expect, that it will do on our own machines), so this npm invocation just has access to the finished tarball and the secret.

web/client-api/types/build.gradle
```diff
@@ -28,10 +36,12 @@ Docker.registerDockerTask(project, 'typedoc') {
         from 'deephaven/node:local-build'
 
         copyFile('.', '/project')
+        environmentVariable('VERSION', npmVersion)
 
         runCommand('''set -eux; \\
                       cd /project/; \\
                       mv dist/types.d.ts dist/index.d.ts; \\
+                      npm version $VERSION; \\
```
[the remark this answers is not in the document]
niloc132: No - there is no change to the source files, this happens inside docker (and doesn't use volumes/bind mounts).

proto/raw-js-openapi/package.json
niloc132: Ok - should I update package.json then, or is it sufficient to say "the combination of the committed package.json and package-lock.json result in a deterministic build"?

web/client-api/types/build.gradle
```diff
@@ -10,6 +10,13 @@ dependencies {
     dts project(path: ':web-client-api', configuration: 'dts')
 }
 
+// Apply our temporary convention until dhc hits 1.0.0
+def npmVersion = '1.0.0-dev' + project.property('deephavenBaseVersion').toString().trim()
+def preId = project.findProperty('npmPreid')// i.e. 'nightly'
+if (preId) {
+    npmVersion += preId + new Date().format("yyyyMMdd")
```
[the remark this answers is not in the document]
mofojed: Yea probably want to include hhmmss as well, just to avoid any collisions

proto/raw-js-openapi/package.json
mofojed: No need to update, the combination of package.json/package-lock.json results in a deterministic build.

.github/workflows/publish-ci.yml
niloc132: Okay, changed - I went a bit further and declared both the npm module tarball and the doc directory as artifacts, which automatically hooks them up as dependent on assemble, just in case we use either in the future in that way.

mattrunyon (approved):
Approved for consuming these types. The types built from this branch work w/ my web-client-ui branch https://github.com/mattrunyon/web-client-ui/tree/jsapi-types

mofojed (approved):

devinrsmith (approved):

niloc132:
Pushing a new merge commit to get CI to re-run, since clicking re-run wasn't doing anything...

PR #4909, opened 2023-12-02
fix: Clean up how objects are managed when they are fetched

### LINKED ISSUE #4903: Widgets cause leaks, do not release all resources
**Description**

Some of the objects we use (JsTreeTable, Pandas tables, JsPartitionedTable) do not manage their tickets correctly, and leak memory.

**Steps to reproduce**
1. Restart your server from a clean slate with no applications so that export tickets can be counted a little bit easier.
2. Go to a JSAPI page (http://localhost:10000/jsapi/), and run the following code in the browser console to create a connection and then print out all the export messages (replace the token `iris` with your token):
```
// REPLACE THE TOKEN WITH YOUR PSK
var token = 'iris';
var dh = (await import('./dh-core.js')).default

var c = new dh.CoreClient(window.location.protocol + "//" + window.location.host)
await c.login({type: 'io.deephaven.authentication.psk.PskAuthenticationHandler', token})

console.log('Logged in...')

var connection = await c.getAsIdeConnection()
var ide = await connection.startSession('python')

console.log('Session started...')

// Keep track of the states for all tickets. Map from ID to state.
var allStates = new Map();

// Just track the current states for the current test run.
var currentStates = new Map();

var STATE_COLORS = {
    'EXPORTED': 'red',
    'RELEASED': 'green',
};

// Listen for SessionState messages
ide.onLogMessage(({ message }) => {
    // We only want to see messages that contain export messages
    const matches = message.match(/export \'(.+)\' is ExportState\.(.+)/);
    if (matches == null) {
        return;
    }

    const [ , exportId, state ] = matches;
    if (exportId.length > 4) {
        // Non-exports are given hex strings for their exportId: https://github.com/deephaven/deephaven-core/blob/6d1d1c96c79d7121763365ceff59f1004b3539fd/server/src/main/java/io/deephaven/server/session/SessionState.java#L590
        // We don't care about these exports
        return;
    }
    allStates.set(exportId, state);
    currentStates.set(exportId, state);
    
    const stateColor = STATE_COLORS[state] ?? 'yellow';
    console.log('[State Update] %s => %c%s', exportId, `background: ${stateColor}`, state);
});

// Add debug logging for SessionState
await ide.runCode(`
import jpy
logger = jpy.get_type('org.slf4j.LoggerFactory').getLogger('io.deephaven.server.session.SessionState')
jpy.cast(logger, jpy.get_type('ch.qos.logback.classic.Logger')).setLevel(jpy.get_type('ch.qos.logback.classic.Level').DEBUG)
`)

async function fetchWidgetExportedObjects(widget) {
    const fetchedObjects = await Promise.all(
        widget.exportedObjects.map(eo => {
            console.log('Fetching exported object', eo);
            return eo.fetch();
        })
    );

    await Promise.all(
        fetchedObjects.filter(o => o.exportedObjects != null).map(o => fetchWidgetExportedObjects(o))        
    );

    fetchedObjects.forEach(fo => { 
        console.log('Closing exported object', fo);
        fo.close() 
    });
}

// Add some helper functions that will be used below
async function fetchAndCloseVariable(varDef, fetchExportedObjects) {
    console.log('Fetching', varDef.title)
    
    const o = await connection.getObject(varDef)

    console.log('Obtained', varDef.title, o, 'closing...')

    if (fetchExportedObjects) {
        await fetchWidgetExportedObjects(o);
    }

    o.close()
    
    console.log('Closed', varDef.title)

    // We need to wait for the close to actually finish. 1s should be more than enough for these purposes, since there's no promise we can actually await on.
    await new Promise(r => setTimeout(r, 1000));
}

async function testCode(code, fetchExportedObjects = false) {
    currentStates.clear()
    var result = await ide.runCode(code);
    var updates = [...result.changes.created, ...result.changes.updated]
    try {
        await Promise.all(updates.map(update => fetchAndCloseVariable(update, fetchExportedObjects)));
    } catch (e) {
        console.error('Error while trying to run tests', e);
    }
    
    let releasedCount = 0;
    const testStates = new Map(currentStates);
    testStates.forEach((state, exportId) => {
        if (state === 'RELEASED') {
            releasedCount++;
        }
    });

    const isSuccess = releasedCount === testStates.size;
    
    console.groupCollapsed('[TEST RESULTS]: %c %s: %s/%s objects released', `background: ${isSuccess ? 'green' : 'red'}`, isSuccess ? 'PASS': 'FAIL', releasedCount, testStates.size);
    console.dir(testStates);
    console.groupEnd();
}
```
3. Run the following snippet to create a regular table, fetch it, and then close it:
```
await testCode(`
t = empty_table(1)
`)
```
4. Run the following snippet to create a Pandas table, fetch it, and then close it:
```
await testCode(`
import pandas as pd
df = pd.DataFrame()
`)
```
5. Run the following snippet to create a HierarchicalTable, fetch it, then close it:
```
await testCode(`
r = empty_table(1).rollup(aggs=None, by=[])
`)
```

6. Run the following snippet to create a Plotly Express widget, fetch it, and then close it:
```
await testCode(`
from deephaven.plot import express as dx

w = dx.line(empty_table(100).update(["x=i", "y=i*(-i)"]), x="x", y="y")
`, true)
```
7. Run the following to create a PartitionedTable, fetch it, and then close it:
```
await testCode(`
pt = empty_table(10000).update(["x=(int)Math.floor(i/100)", "y=(int)Math.floor(i/10)", "z=i"]).partition_by(["x", "y"])
`)
```
8. Figure:
```
await testCode(`

p = Figure().plot_xy(series_name="Test", t=empty_table(100).update(["x=i", "y=i*(-i)"]), x="x", y="y").show()
`)
```
9. deephaven.ui widget exporting another widget:
```
await testCode(`
from deephaven import empty_table, ui
from deephaven.plot import express as dx

@ui.component
def foo():
    return dx.line(empty_table(100).update(["x=i", "y=i*(-i)"]), x="x", y="y")

f = foo()
`, true)
```

**Expected results**
3,4,5,6: All tickets allocated should be released after a close

**Actual results**
3. All tickets are released (2)
4. Export ticket 3 is never released, and there is a ticket with ID -1 exported? (though -1 is released)
5. Export ticket 4 is never released
6. Export ticket -2 is never released
7. Export tickets 6 and -3 are never released

A clear and concise description of what actually happened.

**Versions**
 - Deephaven: 7e596a603b9e685c09b948bf63d7b9a6a7642854

mofojed:
- Clean up memory leak in JsPartitionedTable, JsFigure, PandasTable with widgets not being closed after being used
- Add a `takeTicket` method to JsWidgetExportedObject to make it clear when ownership of the ticket has been passed on
- Close exported objects that are still owned when closing a widget
- Tested using the steps and scripts provided in the ticket, as well as testing with creating some tables and opening/closing them in the UI
- Fixes #4903

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
@@ -760,7 +760,7 @@ public class WorkerConnection {
     }
 
     /**
-     * Export a table from the provided ticket
+     * Export a table from the provided ticket.
      * 
      * @param ticket Ticket of the table to export
      * @param fetchSummary Description for the fetch. Used for logging and errors.
@@ -809,7 +809,7 @@ public class WorkerConnection {
     }
 
     /**
-     * Retrieve an object from it's definition
+     * Retrieve an object from its definition.
      * 
      * @param definition Variable definition to get
      * @return The object requested
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
@@ -767,8 +767,6 @@ public class WorkerConnection {
      * @return Promise for the table
      */
     public Promise<JsTable> getTable(TypedTicket ticket, String fetchSummary) {
-        // Note that creating a CTS like this means we can't actually refetch it, but that's okay, we can't
-        // reconnect in this way without refetching the entire figure anyway.
         return Callbacks.<ExportedTableCreationResponse, Object>grpcUnaryPromise(c -> {
             tableServiceClient().getExportedTableCreationResponse(ticket.getTicket(),
                     metadata(),
@@ -984,42 +982,34 @@ public class WorkerConnection {
                 .then(widget -> Promise.resolve(new JsTreeTable(this, widget)));
     }
 
+    private JsFigure.FigureFetch getWidgetFigureFetch(Promise<JsWidget> widgetPromise) {
+        return c -> widgetPromise.then(widget -> {
+            FetchObjectResponse legacyResponse = new FetchObjectResponse();
+            legacyResponse.setData(widget.getDataAsU8());
+            legacyResponse.setType(widget.getType());
+            legacyResponse.setTypedExportIdsList(
+                    Arrays.stream(widget.getExportedObjects()).map(JsWidgetExportedObject::takeTicket)
+                            .toArray(TypedTicket[]::new));
+            c.apply(null, legacyResponse);
+            widget.close();
+            return null;
+        }, error -> {
+            c.apply(error, null);
+            return null;
+        });
+    }
+
+    private Promise<JsFigure> getFigure(Promise<JsWidget> widgetPromise) {
+        return whenServerReady("get a figure").then(server -> new JsFigure(this, getWidgetFigureFetch(widgetPromise)).refetch());
+
+    }
+
     public Promise<JsFigure> getFigure(JsVariableDefinition varDef) {
-        return whenServerReady("get a figure")
-                .then(server -> new JsFigure(this,
-                        c -> getWidget(varDef).then(widget -> {
-                            FetchObjectResponse legacyResponse = new FetchObjectResponse();
-                            legacyResponse.setData(widget.getDataAsU8());
-                            legacyResponse.setType(widget.getType());
-                            legacyResponse.setTypedExportIdsList(
-                                    Arrays.stream(widget.getExportedObjects()).map(JsWidgetExportedObject::takeTicket)
-                                            .toArray(TypedTicket[]::new));
-                            c.apply(null, legacyResponse);
-                            widget.close();
-                            return null;
-                        }, error -> {
-                            c.apply(error, null);
-                            return null;
-                        })).refetch());
+        return getFigure(getWidget(varDef));
     }
 
     public Promise<JsFigure> getFigure(TypedTicket ticket) {
-        return whenServerReady("get a figure")
-                .then(ignore -> new JsFigure(this,
-                        c -> getWidget(ticket, true).then(widget -> {
-                            FetchObjectResponse legacyResponse = new FetchObjectResponse();
-                            legacyResponse.setData(widget.getDataAsU8());
-                            legacyResponse.setType(widget.getType());
-                            legacyResponse.setTypedExportIdsList(
-                                    Arrays.stream(widget.getExportedObjects()).map(JsWidgetExportedObject::takeTicket)
-                                            .toArray(TypedTicket[]::new));
-                            c.apply(null, legacyResponse);
-                            widget.close();
-                            return null;
-                        }, error -> {
-                            c.apply(error, null);
-                            return null;
-                        })).refetch());
+        return getFigure(getWidget(ticket, true));
     }
 
     /**
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
@@ -946,6 +946,7 @@ public class WorkerConnection {
                                 legacyResponse.setTypedExportIdsList(Arrays.stream(widget.getExportedObjects())
                                         .map(JsWidgetExportedObject::typedTicket).toArray(TypedTicket[]::new));
                                 c.apply(null, legacyResponse);
+                                widget.close();
                                 return null;
                             }, error -> {
                                 c.apply(error, null);
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
@@ -1088,8 +1088,8 @@ public class WorkerConnection {
      */
     public Promise<JsWidget> getWidget(TypedTicket typedTicket, boolean refetch) {
         return whenServerReady("get a widget")
-                .then(response -> Promise.resolve(new JsWidget(this, typedTicket)))
-                .then(widget -> {
+                .then(response -> {
+                    JsWidget widget = new JsWidget(this, typedTicket);
                     if (refetch) {
                         return widget.refetch();
                     }
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
@@ -991,7 +991,6 @@ public class WorkerConnection {
                     Arrays.stream(widget.getExportedObjects()).map(JsWidgetExportedObject::takeTicket)
                             .toArray(TypedTicket[]::new));
             c.apply(null, legacyResponse);
-            widget.close();
             return null;
         }, error -> {
             c.apply(error, null);
@@ -1021,10 +1020,6 @@ public class WorkerConnection {
      */
     private Promise<JsTable> getPandasTable(JsWidget widget) {
         TypedTicket typedTicket = widget.getExportedObjects()[0].takeTicket();
-
-        // We don't need to keep the original widget open, we can just close it right away
-        widget.close();
-
         return getTable(typedTicket, "table for pandas").then(table -> {
             // TODO(deephaven-core#3604) if using a new session don't attempt a reconnect
             // never attempt a reconnect, since we might have a different widget schema entirely
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
@@ -1077,7 +1077,7 @@ public class WorkerConnection {
      * Retrieve a widget using a ticket. Does not create a new ticket.
      * 
      * @param typedTicket Ticket to fetch
-     * @param refetch Whether to refetch the widget. If false, you must ensure you call {@JsWidget#refetch()} before
+     * @param refetch Whether to refetch the widget. If false, you must ensure you call {@link JsWidget#refetch()} before
      *        using the widget.
      * @return Promise of the widget
      */
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
@@ -1077,8 +1077,8 @@ public class WorkerConnection {
      * Retrieve a widget using a ticket. Does not create a new ticket.
      * 
      * @param typedTicket Ticket to fetch
-     * @param reexport Whether to re-export the widget. If false, you must ensure you call {@link JsWidget#refetch()} before
-     *        using the widget.
+     * @param reexport Whether to re-export the widget. If false, you must ensure you call {@link JsWidget#refetch()}
+     *        before using the widget.
      * @return Promise of the widget
      */
     private Promise<JsWidget> getExportedWidget(TypedTicket typedTicket, boolean reexport) {
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
@@ -759,7 +759,7 @@ public class WorkerConnection {
         });
     }
 
-    public Promise<?> getObject(TypedTicket typedTicket) {
+    public Promise<?> getObject(TypedTicket typedTicket, boolean exportNew) {
         return getObject(
                 new JsVariableDefinition(typedTicket.getType(), null, typedTicket.getTicket().getTicket_asB64(), null));
     }
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
@@ -759,6 +759,11 @@ public class WorkerConnection {
         });
     }
 
+    public Promise<?> getObject(TypedTicket typedTicket) {
+        return getObject(
+                new JsVariableDefinition(typedTicket.getType(), null, typedTicket.getTicket().getTicket_asB64(), null));
+    }
+
     public Promise<?> getObject(JsVariableDefinition definition) {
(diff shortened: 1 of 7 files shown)


web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -759,14 +759,68 @@ public Promise<JsTable> getTable(JsVariableDefinition varDef, @Nullable Boolean
...
+                    c::apply);
+        }).then(etcr -> {
+            ClientTableState cts = newStateFromUnsolicitedTable(etcr, fetchSummary);
+            JsTable table = new JsTable(this, cts);
+            return Promise.resolve(table);
+        });
+    }
+
+    /**
+     * Get an object from it's ticket
```
niloc132, line 782:
```suggestion
     * Get an object from its ticket.
```

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -759,14 +759,68 @@ public Promise<JsTable> getTable(JsVariableDefinition varDef, @Nullable Boolean
         });
     }
 
+    /**
+     * Export a table from the provided ticket
```
niloc132, line 763:
```suggestion
     * Export a table from the provided ticket.
```

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -759,14 +759,68 @@ public Promise<JsTable> getTable(JsVariableDefinition varDef, @Nullable Boolean
...
+     * 
+     * @param ticket Ticket of the table to export
+     * @param fetchSummary Description for the fetch. Used for logging and errors.
+     * @return Promise for the table
+     */
+    public Promise<JsTable> getTable(TypedTicket ticket, String fetchSummary) {
```
niloc132, line 769: maybe a name that is clear about how this is intended for server-side exports only (i.e. note most table tickets)? the api would totally let you pass in a plain typedticket and summary, but would fail for client tickets/etc.

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -759,14 +759,68 @@ public Promise<JsTable> getTable(JsVariableDefinition varDef, @Nullable Boolean
...
+            if (JsVariableType.TREETABLE.equalsIgnoreCase(typedTicket.getType())) {
+                JsLog.warn(
+                        "TreeTable is now HierarchicalTable, fetching as a plain widget. To fetch as a HierarchicalTable use that as this type.");
+            }
+            return getWidget(typedTicket, true);
+        }
+    }
+
+    /**
+     * Retrieve an object from it's definition
```
niloc132, line 812:
```suggestion
     * Retrieve an object from its definition.
```

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
```diff
@@ -979,8 +979,7 @@ public void close() {
 
         connection.unregisterSimpleReconnectable(this);
 
-        // Presently it is never necessary to release widget tickets, since they can't be export tickets.
-        // connection.releaseTicket(widget.getTicket());
+        connection.releaseTicket(widget.getTicket());
```
niloc132, line 982: alternatively, right?
```suggestion
        widget.close();
```

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
```diff
@@ -47,24 +46,46 @@ public TypedTicket typedTicket() {
...
+        verifyIsOwner();
+        isOwner = false;
+        return typedTicket();
+    }
+
+    /**
+     * Fetches the object from the server.
+     * 
+     * @param takeOwnership Whether to take ownership of the object. Defaults to false.
+     *
```
niloc132, line 76:
```suggestion
```

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
```diff
@@ -47,24 +46,46 @@ public TypedTicket typedTicket() {
...
+     * @return a promise that resolves to the object
+     */
+    @JsMethod
+    public Promise<?> fetch(@JsOptional Boolean takeOwnership) {
+        if (takeOwnership == null) {
+            takeOwnership = false;
+        }
+        if (takeOwnership) {
+            return this.connection.getObject(takeTicket());
```
niloc132, line 85: this means that raced calls to `fetch(true)` and `fetch(false)` are going to intermittently fail - probably need some clearer docs on this and other general usage of ownership.

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
```diff
@@ -47,24 +46,46 @@ public TypedTicket typedTicket() {
...
         }
+        return this.connection.getObject(
+                new JsVariableDefinition(ticket.getType(), null, ticket.getTicket().getTicket_asB64(), null));
```
niloc132, line 88: I'm surprised this works - VariableDefinition is meant for scope tickets...

Will think on it more, but I think I'd prefer we decompose this (or provide an interface and second impl)  to let this be reexported-and-fetched.

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -948,19 +1048,53 @@ private TypedTicket createTypedTicket(JsVariableDefinition varDef) {
...
+    public Promise<JsWidget> getWidget(TypedTicket typedTicket, boolean refetch) {
         return whenServerReady("get a widget")
-                .then(response -> new JsWidget(this, typedTicket).refetch());
+                .then(response -> Promise.resolve(new JsWidget(this, typedTicket)))
+                .then(widget -> {
+                    if (refetch) {
+                        return widget.refetch();
+                    }
+                    return Promise.resolve(widget);
+                });
```
niloc132, line 1097:
```suggestion
                .then(response -> {
                    JsWidget = w = new JsWidget(this, typedTicket);
                    if (refetch) {
                        return widget.refetch();
                    }
                    return Promise.resolve(widget);
                });
```

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -907,38 +961,84 @@ private TicketAndPromise<?> exportScopeTicket(JsVariableDefinition varDef) {
...
+
+        // We don't need to keep the original widget open, we can just close it right away
+        widget.close();
+
+        return getTable(typedTicket, "table for pandas").then(table -> {
+            // TODO(deephaven-core#3604) if using a new session don't attempt a reconnect
+            // never attempt a reconnect, since we might have a different widget schema entirely
+            table.addEventListener(JsTable.EVENT_DISCONNECT, ignore -> table.close());
+            return Promise.resolve(table);
+        });
```
niloc132, line 1033: Is this necessary? Can't we just do something sorta like

```suggestion
        return widget.getExportedObjects()[0].fetch(true).then(table -> {
            // TODO(deephaven-core#3604) if using a new session don't attempt a reconnect
            // never attempt a reconnect, since we might have a different widget schema entirely
            table.addEventListener(JsTable.EVENT_DISCONNECT, ignore -> table.close());
            return Promise.resolve(table);
        });
```

web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java
mofojed: Yes much better

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
mofojed: We still need to close the `JsWidget`. Maybe could have a method on JsWidget like `Promise<?> takeObjects()` which returns a `Promise.all` of all the exported objects fetch objects and closes the parent widget as well?

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
mofojed: Was debating just having a separate method `take` instead of making it a boolean.
Really you should only be calling one or the other (`true` or `false`) depending on how you're managing the object, I wouldn't expect you'd want to call both.
In any case, I can add a `verifyIsOwner()` here in the `false` case, and throw if it's been called after a `fetch(false)` so it reduces a race that way, but you could still race `fetch(false)` then `fetch(true)` I suppose.

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
mofojed: Yea, that was already there. Looks like I added this back when implementing JsWidget: https://github.com/deephaven/deephaven-core/pull/1888#discussion_r793091444
I may have misinterpreted that comment then.
Should be able to refactor the `WorkerConnection#getObject` methods; the only part of the `JsVariableDefinition` that is actually used is the `type` and the `id` used in `exportScopeTicket`

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -948,19 +1043,53 @@ private TypedTicket createTypedTicket(JsVariableDefinition varDef) {
...
-    public Promise<JsWidget> getWidget(TypedTicket typedTicket) {
+    /**
+     * Retrieve a widget using a ticket. Does not create a new ticket.
+     * 
+     * @param typedTicket Ticket to fetch
+     * @param refetch Whether to refetch the widget. If false, you must ensure you call {@JsWidget#refetch()} before
+     *        using the widget.
+     * @return Promise of the widget
+     */
+    public Promise<JsWidget> getWidget(TypedTicket typedTicket, boolean refetch) {
```
niloc132, line 1084: Many of these could be private, they are not called outside this class

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -948,19 +1043,53 @@ private TypedTicket createTypedTicket(JsVariableDefinition varDef) {
...
+                    return getWidget(typedTicket, refetch);
                 }).promise();
     }
 
-    public Promise<JsWidget> getWidget(TypedTicket typedTicket) {
+    /**
+     * Retrieve a widget using a ticket. Does not create a new ticket.
+     * 
+     * @param typedTicket Ticket to fetch
+     * @param refetch Whether to refetch the widget. If false, you must ensure you call {@JsWidget#refetch()} before
```
niloc132, line 1080:
```suggestion
     * @param refetch Whether to refetch the widget. If false, you must ensure you call {@link JsWidget#refetch()} before
```

(plus spotless, which will probably push a word to the next line)

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidget.java
```diff
@@ -85,13 +85,26 @@ private void closeStream() {
...
+    }
+
     /**
      * Ends the client connection to the server.
      */
     @JsMethod
     public void close() {
         suppressEvents();
         closeStream();
+        closeExportedObjects();
```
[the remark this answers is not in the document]
mofojed: I'll change the default

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
mofojed: Updated names

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
mofojed: Now we are releasing the ticket when the widget gets the close event from the server, so this is no longer necessary.

mofojed:
Rollups are screwed up, need to fix...

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -948,19 +1043,53 @@ private TypedTicket createTypedTicket(JsVariableDefinition varDef) {
...
+
+    /**
+     * Retrieve a widget using a variable definition. Creates a new export ticket to reference the variable.
+     * Intentionally races the ticket export with the variable export to parallelize the requests.
+     * 
+     * @param varDef Variable definition to get the widget for
+     * @param refetch Whether to refetch the widget before resolving or not
+     * @return Promise of the widget
+     */
+    public Promise<JsWidget> getWidget(JsVariableDefinition varDef, boolean refetch) {
```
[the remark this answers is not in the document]
mofojed: The `refetch` in this case is triggering `widget.refetch()`, which was already named... but essentially is restarting the stream and fetching the data. I think the `reexport` name probably doesn't fit here either, and `refetch` makes more sense (or `JsWidget.refetch` should be renamed?). Thoughts?

mofojed:
@niloc132 So the issue now with releasing the widget ticket when the stream ends (as suggested in our previous discussion for fetch-only widgets) actually breaks JsTreeTable implementation, as it's using the ticket right from the widget for the hierarchical table?? https://github.com/deephaven/deephaven-core/blob/160760690998bb67ebe71e5977c12223a5071d58/web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java#L512

So I changed HierarchicalTableTypePlugin to export a reference to the HierarchicalTable as an exported object... but I think JsPartitionedTable still has a problem... they're not expecting the widget ticket to be released after the widget has been fetched, which is at odds with what we're doing with Pandas (I'm actually not sure where the Pandas plugin is/how that works?).

I favour what I had previously, where the consumer needs to explicitly close the widget when they're done with it to release the ticket, rather than implicitly releasing the ticket when the stream ends... thoughts?

mofojed:
Closed in favour of #4939

PR #4919, opened 2023-12-06
Make OperationInitializer non-static

niloc132:
This patch makes the OperationInitializer part of the ExecutionContext, allowing consumers to specify a preferred instance. ThreadInitializationFactory is also changed to be built in Dagger instead of only by configuration. Combined, these changes make it easier to reason about startup, requiring that dependencies are declared explicitly rather than implicitly referenced and used.

It is likely that some of the work done here should be refactored further into some JobScheduler factory or assisted injection, but OperationInitializer is still referenced directly by InitialFilterExecution, so this would require more refactoring.

Ideally, OperationInitializer.NON_PARALLELIZABLE would be used instead of the OperationInitializerThreadPool's threadlocal, but some services restore the user's own exec context - this is a shortcoming of the patch, and possibly should be resolved before merging.

Downstream users will need to consider if they want to capture the OperationInitializer when they create new exec contexts - generally this will be desired.

Partial #4040


engine/table/src/main/java/io/deephaven/engine/updategraph/impl/PeriodicUpdateGraph.java
```diff
@@ -2084,6 +2091,16 @@ public Builder numUpdateThreads(int numUpdateThreads) {
             return this;
         }
 
+        /**
+         *
```
niloc132, line 2095: finish javadoc

engine/test-utils/src/main/java/io/deephaven/engine/testutil/ControlledUpdateGraph.java
```diff
@@ -1,13 +1,14 @@
...
+import io.deephaven.util.thread.ThreadInitializationFactory;
 
 // TODO (deephaven-core#3886): Extract test functionality from PeriodicUpdateGraph
 public class ControlledUpdateGraph extends PeriodicUpdateGraph {
-
-    public static final ControlledUpdateGraph INSTANCE = new ControlledUpdateGraph();
-
-    private ControlledUpdateGraph() {
-        super("TEST", true, 1000, 25, -1);
+    public ControlledUpdateGraph(String name, boolean allowUnitTestMode, long targetCycleDurationMillis,
```
niloc132, line 8: Probably should restore the default ctor, but I think it makes sense to move the singleton instance?

engine/test-utils/src/main/java/io/deephaven/engine/testutil/QueryTableTestBase.java
```diff
@@ -118,7 +119,7 @@ public String toString() {
         public void step(int leftSize, int rightSize, QueryTable leftTable, QueryTable rightTable,
                 ColumnInfo<?, ?>[] leftColumnInfo, ColumnInfo<?, ?>[] rightColumnInfo, EvalNuggetInterface[] en,
                 Random random) {
-            final ControlledUpdateGraph updateGraph = ExecutionContext.getContext().getUpdateGraph().cast();
+            final PeriodicUpdateGraph updateGraph = ExecutionContext.getContext().getUpdateGraph().cast();
```
niloc132, line 122: Should restore these

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/OperationInitializer.java
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     *
+     * @return
+     */
+    int parallelismFactor();
+
+    /**
+     *
+     */
+    void start();
```
niloc132, line 55: needs javadoc, should rename to prestart? optionally add a shutdown method?

props/configs/src/main/resources/dh-defaults.prop
```diff
@@ -63,4 +63,4 @@ client.version.list=deephaven=io.deephaven.engine.table.Table,barrage=io.deephav
 
 
 # Specifies additional setup to run on threads that can perform table operations with user code. Comma-separated list, instances must be of type io.deephaven.util.thread.ThreadInitializationFactory
-thread.initialization=io.deephaven.server.console.python.DebuggingInitializer
+thread.initialization=
```
niloc132, line 66: should add a note that this is no longer used

engine/table/src/main/java/io/deephaven/engine/table/impl/OperationInitializationThreadPool.java
```diff
@@ -31,56 +36,47 @@ public class OperationInitializationThreadPool {
...
+        return NUM_THREADS;
     }
 
     /**
-     * Start the OperationInitializationThreadPool. In practice, this just pre-starts all threads in the
-     * {@link #executorService()}.
+     * Start the OperationInitializationThreadPool. In practice, this just pre-starts all threads.
      */
-    public static void start() {
+    public void start() {
```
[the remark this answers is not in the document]
niloc132: Also, should be called at all...

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/OperationInitializer.java
```diff
@@ -0,0 +1,56 @@
+package io.deephaven.engine.updategraph;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
+
+/**
+ * alt naming: OperationParallelismControl?
```
niloc132, line 7: Bikeshed time?

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/OperationInitializer.java
niloc132: resolved - method is removed, OITP starts itself

engine/context/src/main/java/io/deephaven/engine/context/PoisonedOperationInitializer.java
```diff
@@ -0,0 +1,35 @@
+package io.deephaven.engine.context;
+
+import io.deephaven.engine.updategraph.OperationInitializer;
+import io.deephaven.util.ExecutionContextRegistrationException;
+
+import java.util.concurrent.Future;
+
+public class PoisonedOperationInitializer implements OperationInitializer {
```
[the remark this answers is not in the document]
niloc132: Looks like we don't javadoc sibling classes, so resolving this.

engine/context/src/main/java/io/deephaven/engine/context/ExecutionContext.java
```diff
@@ -208,8 +227,10 @@ public static class Builder {
         private QueryScope queryScope = PoisonedQueryScope.INSTANCE;
         private QueryCompiler queryCompiler = PoisonedQueryCompiler.INSTANCE;
         private UpdateGraph updateGraph = PoisonedUpdateGraph.INSTANCE;
+        private OperationInitializer operationInitializer = PoisonedOperationInitializer.INSTANCE;
 
         private Builder() {
+            // why automatically propagate this, but not other things?
```
niloc132, line 233: remove

rcaudy:
`PartitionedTableImpl` L299 and L356:
We must not use the same operation initializer as the currently-installed one, unless it's the non-parallelizable one.
Only really necessary if we get rid of the `isInitializationThread` `ThreadLocal` to dictate `canParallelize()`... might be a nice improvement....

chipkent (approved):
Python LGTM

engine/table/src/main/java/io/deephaven/engine/table/impl/OperationInitializationThreadPool.java
```diff
@@ -31,56 +36,47 @@ public class OperationInitializationThreadPool {
...
-    public static ExecutorService executorService() {
-        return executorService;
+    @Override
+    public boolean canParallelize() {
+        return NUM_THREADS > 1 && !isInitializationThread.get();
+    }
+
+    @Override
+    public Future<?> submit(Runnable runnable) {
+        return executorService.submit(runnable);
```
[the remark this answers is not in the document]
niloc132: Assertion doesn't work it turns out - SelectColumnLayer appears to deliberately submit more work from on the OITP threads to that thread pool. Crucially it doesn't block on that work being returned, so this should be safe.

I'm going to defer other refactoring here to minimize risk of introducing bugs that I can't guard against like this.

engine/table/src/main/java/io/deephaven/engine/table/impl/partitioned/PartitionedTableImpl.java
```diff
@@ -318,6 +319,26 @@ public PartitionedTable transform(
...
+            return null;
+        }
+        ExecutionContext current = ExecutionContext.getContext();
+        if (!provided.getInitializer().canParallelize()) {
+            return provided;
+        }
+        if (current.getInitializer() != provided.getInitializer()) {
+            return provided;
+        }
+        return provided.withOperationInitializer(OperationInitializer.NON_PARALLELIZABLE);
```
nbauernfeind, line 339: I am finding the name and javadoc a little confusing.

I'd like to suggest renaming to `disableRecursiveParallelOperationInitialization()` with the javadoc indicating that it returns the provided execution context after possibly disabling parallelism if further parallelism could result in deadlock.

Then probably finish off the method with the comment making it clear this is a disable:
```java
// disable recursive parallelism
return provided.withOperationInitializer(...);
```

engine/updategraph/src/main/java/io/deephaven/engine/updategraph/OperationInitializer.java
```diff
@@ -0,0 +1,56 @@
...
+        @Override
+        public void start() {
+            // no-op
+        }
+    };
+
+    /**
+     * @return Whether the current thread can parallelize operations using this OperationInitialization.
+     */
+    boolean canParallelize();
```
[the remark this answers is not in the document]
niloc132: As above with the SelectColumnLayer, this doesnt look feasible at this time.

nbauernfeind (approved):

rcaudy (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3545
Reference: https://github.com/deephaven/deephaven.io/issues/3544

PR #4922, opened 2023-12-07
UPDATE: fixed formatting of hyperlink in pydoc

elijahpetty:
Having issues with my previews, but this should be correct syntax for a hyperlink using Sphinx.


jmao-denver (dismissed):
LGTM

chipkent (approved):

PR #4931, opened 2023-12-11
fix: JsPartitionedTable keyColumn and keyColumnTypes ordering

georgecwan:
Fixes issue where JsPartitionedTable returns the keyColumns and keyColumn types in the constituent table's column order which doesn't always match the partitioned table's column order.


web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
```diff
@@ -95,19 +92,25 @@ public Promise<JsPartitionedTable> refetch() {
...
+            JsArray<String> keyColumnNames = descriptor.getKeyColumnNamesList();
+            for (int i = 0; i < keyColumnNames.length; i++) {
+                String name = keyColumnNames.getAt(i);
+                ColumnDefinition columnDefinition = tableDefinition.getColumnsByName().get(false).get(name);
+                int index = 0;
+                while (!columns[index].getName().equals(name)) {
+                    index++;
                 }
+                keyColumnTypes.add(columnDefinition.getType());
+                keyColumns[keyColumns.length] = columns[index];
```
niloc132, line 110: No need to loop twice, i think this would do the job too?
```suggestion
                ColumnDefinition columnDefinition = tableDefinition.getColumnsByName().get(false).get(name);
                keyColumnTypes.add(columnDefinition.getType());
                keyColumns[keyColumns.length] = columns[columnDefinition.getIndex()];
```

niloc132 (approved):

PR #4936, opened 2023-12-11
Share PyArrow table data with the server through ByteBuffer

### LINKED ISSUE #3709: Add ByteBuffer support in JPY to give DH  direct access to pyarrow data
1. cut the number of data copies by 1 or achieve zero-copy when ingesting arrow data from Python
2. possibly allow arrow data created by in Python to be arrow-backed ColumnSource

### LINKED ISSUE #4935: Improve arrow.to_table() by utilizing the new ByteBuffer support in JPY
https://github.com/jpy-consortium/jpy/pull/112

It will reduce the number of copies in memory by 1 and be a bit faster.

jmao-denver:
Fixes #4935
Fixes #3709

depends on https://github.com/jpy-consortium/jpy/pull/112
depends on #5022

### DIFF
--- a/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/ArrowToTableConverter.java
+++ b/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/ArrowToTableConverter.java
@@ -71,10 +71,8 @@ public class ArrowToTableConverter {
     @ScriptApi
     public synchronized void setSchema(final ByteBuffer ipcMessage) {
         // The input ByteBuffer instance (especially originated from Python) can't be assumed to be valid after the
-        // return
-        // of this method. Until https://github.com/jpy-consortium/jpy/issues/126 is resolved, we need to copy the data
-        // out of
-        // the input ByteBuffer to use after the return of this method.
+        // return of this method. Until https://github.com/jpy-consortium/jpy/issues/126 is resolved, we need to copy
+        // the data out of the input ByteBuffer to use after the return of this method.
         if (completed) {
             throw new IllegalStateException("Conversion is complete; cannot process additional messages");
         }
@@ -88,10 +86,8 @@ public class ArrowToTableConverter {
     @ScriptApi
     public synchronized void addRecordBatches(final ByteBuffer... ipcMessages) {
         // The input ByteBuffer instance (especially originated from Python) can't be assumed to be valid after the
-        // return
-        // of this method. Until https://github.com/jpy-consortium/jpy/issues/126 is resolved, we need to copy the data
-        // out of
-        // the input ByteBuffer to use after the return of this method.
+        // return of this method. Until https://github.com/jpy-consortium/jpy/issues/126 is resolved, we need to copy
+        // the data out of the input ByteBuffer to use after the return of this method.
         for (final ByteBuffer ipcMessage : ipcMessages) {
             addRecordBatch(ipcMessage);
         }
@@ -100,10 +96,8 @@ public class ArrowToTableConverter {
     @ScriptApi
     public synchronized void addRecordBatch(final ByteBuffer ipcMessage) {
         // The input ByteBuffer instance (especially originated from Python) can't be assumed to be valid after the
-        // return
-        // of this method. Until https://github.com/jpy-consortium/jpy/issues/126 is resolved, we need to copy the data
-        // out of
-        // the input ByteBuffer to use after the return of this method.
+        // return of this method. Until https://github.com/jpy-consortium/jpy/issues/126 is resolved, we need to copy
+        // the data out of the input ByteBuffer to use after the return of this method.
         if (completed) {
             throw new IllegalStateException("Conversion is complete; cannot process additional messages");
         }
--- a/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/ArrowToTableConverter.java
+++ b/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/ArrowToTableConverter.java
@@ -48,7 +48,7 @@ public class ArrowToTableConverter {
 
     private volatile boolean completed = false;
 
-    private BarrageProtoUtil.MessageInfo parseArrowIpcMessage(final ByteBuffer bb) throws IOException {
+    private static BarrageProtoUtil.MessageInfo parseArrowIpcMessage(final ByteBuffer bb) throws IOException {
         final BarrageProtoUtil.MessageInfo mi = new BarrageProtoUtil.MessageInfo();
 
         bb.order(ByteOrder.LITTLE_ENDIAN);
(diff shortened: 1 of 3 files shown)


chipkent (approved):

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/ArrowToTableConverter.java
```diff
@@ -48,10 +48,9 @@ public class ArrowToTableConverter {
 
     private volatile boolean completed = false;
 
-    private static BarrageProtoUtil.MessageInfo parseArrowIpcMessage(final byte[] ipcMessage) throws IOException {
```
devinrsmith, line 51: Why remove static?

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/ArrowToTableConverter.java
```diff
@@ -70,7 +69,7 @@ private static BarrageProtoUtil.MessageInfo parseArrowIpcMessage(final byte[] ip
     }
 
     @ScriptApi
-    public synchronized void setSchema(final byte[] ipcMessage) {
+    public synchronized void setSchema(final ByteBuffer ipcMessage) {
```
devinrsmith, line 72: So, I think we need to add a big warning here and link to https://github.com/jpy-consortium/jpy/issues/126; no strong references to this buffer can exist by the end of this method. I think we need to extend this warning into the implementation of parseSchema as well - it *can't* hold any references to `Schema header` (since that has references to the byte buffer).

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/ArrowToTableConverter.java
devinrsmith: So, this is what scares me. Flatbuffer almost certainly retains references to ByteBuffer. We need to be very careful that nothing retains this flatbuffer object.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/ArrowToTableConverter.java
```diff
@@ -192,7 +198,7 @@ protected BarrageMessage createBarrageMessage(BarrageProtoUtil.MessageInfo mi, i
         return msg;
     }
 
-    private BarrageProtoUtil.MessageInfo getMessageInfo(byte[] ipcMessage) {
+    private BarrageProtoUtil.MessageInfo getMessageInfo(ByteBuffer ipcMessage) {
```
devinrsmith, line 201: Flatbuffer almost certainly retains references to ByteBuffer. We need to be very careful that nothing retains this flatbuffer object. I think we need to add a big warning in `io.deephaven.extensions.barrage.util.ArrowToTableConverter#createBarrageMessage` that nothing can retain reference to `BarrageProtoUtil.MessageInfo mi`

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/ArrowToTableConverter.java
```diff
@@ -82,7 +81,14 @@ public synchronized void setSchema(final byte[] ipcMessage) {
...
     @ScriptApi
-    public synchronized void addRecordBatch(final byte[] ipcMessage) {
+    public synchronized void addRecordBatches(final ByteBuffer... ipcMessages) {
+        for (final ByteBuffer ipcMessage : ipcMessages) {
+            addRecordBatch(ipcMessage);
+        }
+    }
+
+    @ScriptApi
+    public synchronized void addRecordBatch(final ByteBuffer ipcMessage) {
```
devinrsmith, line 91: Need to add same warnings here.

devinrsmith (changes requested):

devinrsmith:
I think we need to carefully audit all of the parsing logic here to make sure no ByteBuffer / Flatbuffer object is retained. It's not obvious to me without digging in further to the implementations, which is why trying to use `jpy.byte_buffer` as it exists today is so dangerous. I think @nbauernfeind to review with this context in mind and go down through the implementations to double check.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/ArrowToTableConverter.java
```diff
@@ -69,8 +68,13 @@ private static BarrageProtoUtil.MessageInfo parseArrowIpcMessage(final byte[] ip
         return mi;
     }
 
+    /**
+     * The input ByteBuffer instance (especially originated from Python) can't be assumed to be valid after the return
+     * of this method. Until jpy-consortium/jpy#126 is resolved, we need to copy the input ByteBuffer to a new
```
chipkent, line 73: I suggest the full URL.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/ArrowToTableConverter.java
```diff
@@ -81,8 +85,25 @@ public synchronized void setSchema(final byte[] ipcMessage) {
...
-    public synchronized void addRecordBatch(final byte[] ipcMessage) {
+    public synchronized void addRecordBatches(final ByteBuffer... ipcMessages) {
+        for (final ByteBuffer ipcMessage : ipcMessages) {
+            addRecordBatch(ipcMessage);
+        }
+    }
+
+    /**
+     * The input ByteBuffer instance (especially originated from Python) can't be assumed to be valid after the return
+     * of this method. Until jpy-consortium/jpy#126 is resolved, we need to copy the input ByteBuffer to a new
```
chipkent, line 102: I suggest the full URL.

chipkent (approved):
OK with me if all of @devinrsmith 's suggestions have been addressed.  My comments are optional.

nbauernfeind (approved):
I dug in a bit. All of the flatbuffer usages that could reference the byte buffer (such as for strings) perform copies. (In particular the internal flatbuffer `__string` method copies during decoding). The `extractChunkFromInputStream` implementations similarly will copy via `new String(buf, offset, length)` invocations. Similarly any arrays extract a copy from the contiguous body payload rather than an effective slice.

AFAICT this change looks safe.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/ArrowToTableConverter.java
```diff
@@ -81,8 +85,25 @@ public synchronized void setSchema(final byte[] ipcMessage) {
         parseSchema((Schema) mi.header.header(new Schema()));
     }
 
+    /**
```
devinrsmith, line 88: I don't think it's appropriate as javadoc; this is more of an implementation note. I would move these inside the method.

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/util/ArrowToTableConverter.java
```diff
@@ -81,8 +85,25 @@ public synchronized void setSchema(final byte[] ipcMessage) {
...
+     * The input ByteBuffer instance (especially originated from Python) can't be assumed to be valid after the return
+     * of this method. Until jpy-consortium/jpy#126 is resolved, we need to copy the input ByteBuffer to a new
```
devinrsmith, line 90: We don't need to copy the ByteBuffer to a *new* ByteBuffer, we just need to copy the data out in _some_ format. (In this case, likely primitives derived from flatbuffers).

devinrsmith (changes requested):
Please add warning comment inside of parseSchema and createBarrageMessage as well.

devinrsmith (dismissed):

chipkent (approved):

PR #4939, opened 2023-12-12
Fix JS client leaking exported references, clarify dh.Widget API

### LINKED ISSUE #4903: Widgets cause leaks, do not release all resources

niloc132:
This PR is an update to https://github.com/deephaven/deephaven-core/pull/4909.

JsPartitionedTable, JsFigure, PandasTable were previously being leaked even with correct usage, ensure that those are released when client code calls close(). Additionally, partitioned tables would be re-exported unnecessarily, but only the re-export was released, leaking the original reference.

Widget and WidgetExportedObject had documentation indicating how to consume exported objects, but these rules were not enforced. The rules have changed to be more consistent and are enforced, which should avoid later confusion, and new documentation added to propose potential patterns to follow when authoring plugins.

Fixes #4903

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidget.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidget.java
@@ -4,7 +4,6 @@
 package io.deephaven.web.client.api.widget;
 
 import com.vertispan.tsdefs.annotations.TsName;
-import com.vertispan.tsdefs.annotations.TsTypeRef;
 import com.vertispan.tsdefs.annotations.TsUnion;
 import com.vertispan.tsdefs.annotations.TsUnionMember;
 import elemental2.core.ArrayBuffer;
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidget.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidget.java
@@ -218,6 +218,10 @@ public class JsWidget extends HasEventHandling implements ServerObject, WidgetMe
         return new String(Js.uncheckedCast(response.getData().getPayload_asU8()), StandardCharsets.UTF_8);
     }
 
+    /**
+     * @return the exported objects sent in the initial message from the server. The client is responsible for closing
+     *         them when finished using them.
+     */
     @Override
     @JsProperty
     public JsWidgetExportedObject[] getExportedObjects() {
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidget.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidget.java
@@ -34,15 +34,57 @@ import java.util.function.Supplier;
 /**
  * A Widget represents a server side object that sends one or more responses to the client. The client can then
  * interpret these responses to see what to render, or how to respond.
- *
+ * <p>
  * Most custom object types result in a single response being sent to the client, often with other exported objects, but
  * some will have streamed responses, and allow the client to send follow-up requests of its own. This class's API is
  * backwards compatible, but as such does not offer a way to tell the difference between a streaming or non-streaming
  * object type, the client code that handles the payloads is expected to know what to expect. See
- * dh.WidgetMessageDetails for more information.
- *
+ * {@link WidgetMessageDetails} for more information.
+ * <p>
  * When the promise that returns this object resolves, it will have the first response assigned to its fields. Later
- * responses from the server will be emitted as "message" events. When the connection with the server ends
+ * responses from the server will be emitted as "message" events. When the connection with the server ends, the "close"
+ * event will be emitted. In this way, the connection will behave roughly in the same way as a WebSocket - either side
+ * can close, and after close no more messages will be processed. There can be some latency in closing locally while
+ * remote messages are still pending - it is up to implementations of plugins to handle this case.
+ * <p>
+ * Also like WebSockets, the plugin API doesn't define how to serialize messages, and just handles any binary payloads.
+ * What it does handle however, is allowing those messages to include references to server-side objects with those
+ * payloads. Those server side objects might be tables or other built-in types in the Deephaven JS API, or could be
+ * objects usable through their own plugins. They also might have no plugin at all, allowing the client to hold a
+ * reference to them and pass them back to the server, either to the current plugin instance, or through another API.
+ * The {@code Widget} type does not specify how those objects should be used or their lifecycle, but leaves that
+ * entirely to the plugin. Messages will arrive in the order they were sent.
+ * <p>
+ * This can suggest several patterns for how plugins operate:
+ * <ul>
+ * <li>The plugin merely exists to transport some other object to the client. This can be useful for objects which can
+ * easily be translated to some other type (like a Table) when the user clicks on it. An example of this is
+ * {@code pandas.DataFrame} will result in a widget that only contains a static
+ * {@link io.deephaven.web.client.api.JsTable}. Presently, the widget is immediately closed, and only the Table is
+ * provided to the JS API consumer.</li>
+ * <li>The plugin provides references to Tables and other objects, and those objects can live longer than the object
+ * which provided them. One concrete example of this could have been
+ * {@link io.deephaven.web.client.api.JsPartitionedTable} when fetching constituent tables, but it was implemented
+ * before bidirectional plugins were implemented. Another example of this is plugins that serve as a "factory", giving
+ * the user access to table manipulation/creation methods not supported by gRPC or the JS API.</li>
+ * <li>The plugin provides reference to Tables and other objects that only make sense within the context of the widget
+ * instance, so when the widget goes away, those objects should be released as well. This is also an example of
+ * {@link io.deephaven.web.client.api.JsPartitionedTable}, as the partitioned table tracks creation of new keys through
+ * an internal table instance.</li>
+ * </ul>
+ *
+ * Handling server objects in messages also has more than one potential pattern that can be used:
+ * <ul>
+ * <li>One object per message - the message clearly is about that message, no other details required.</li>
+ * <li>Objects indexed within their message - as each message comes with a list of objects, those objects can be
+ * referenced within the payload by index. This is roughly how {@link io.deephaven.web.client.api.widget.plot.JsFigure}
+ * behaves, where the figure descriptor schema includes an index for each created series, describing which table should
+ * be used, which columns should be mapped to each axis.</li>
+ * <li>Objects indexed since widget creation - each message would append its objects to a list created when the widget
+ * was first made, and any new exports that arrive in a new message would be appended to that list. Then, subsequent
+ * messages can reference objects already sent. This imposes a limitation where the client cannot release any exports
+ * without the server somehow signaling that it will never reference that export again.</li>
+ * </ul>
  */
 // TODO consider reconnect support? This is somewhat tricky without understanding the semantics of the widget
 @TsName(namespace = "dh", name = "Widget")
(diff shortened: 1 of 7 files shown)


web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidget.java
```diff
@@ -34,15 +33,57 @@
...
+ * the user access to table manipulation/creation methods not supported by gRPC or the JS API.</li>
+ * <li>The plugin provides reference to Tables and other objects that only make sense within the context of the widget
+ * instance, so when the widget goes away, those objects should be released as well. This is also an example of
+ * {@link io.deephaven.web.client.api.JsPartitionedTable}, as the partitioned table tracks creation of new keys through
+ * an internal table instance.</li>
+ * </ul>
+ *
+ * Handling server objects in messages also has more than one potential pattern that can be used:
+ * <ul>
+ * <li>One object per message - the message clearly is about that message, no other details required.</li>
```
mofojed, line 77:
```suggestion
 * <li>One object per message - the message clearly is about that object, no other details required.</li>
```

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
```diff
@@ -29,13 +34,46 @@ public class JsWidgetExportedObject implements ServerObject {
...
     private final TypedTicket ticket;
 
+    private final JsLazy<Promise<?>> fetched;
+
     public JsWidgetExportedObject(WorkerConnection connection, TypedTicket ticket) {
         this.connection = connection;
         this.ticket = ticket;
+        this.fetched = JsLazy.of(() -> {
+            if (getType() == null) {
+                return Promise.reject("Exported object has no type, can't be fetched");
```
mofojed, line 44: Is this a case that actually happens?

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
```diff
@@ -47,32 +85,47 @@ public TypedTicket typedTicket() {
...
+     */
+    @JsMethod
+    public Promise<JsWidgetExportedObject> reexport() {
+        Ticket reexportedTicket = connection.getConfig().newTicket();
+        return Callbacks.grpcUnaryPromise(c -> {
+            ExportRequest req = new ExportRequest();
+            req.setSourceId(ticket.getTicket());
+            req.setResultId(reexportedTicket);
+            connection.sessionServiceClient().exportFromTicket(req, connection.metadata(), c::apply);
+        }).then(success -> {
```
mofojed, line 102: The downside of this is the extra round trip (not a `TicketAndPromise`), but it is a lot cleaner this way.

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
```diff
@@ -47,32 +85,47 @@ public TypedTicket typedTicket() {
...
     /**
-     * Releases the server-side resources associated with this object, regardless of whether or not other client-side
-     * objects exist that also use that object.
+     * Releases the server-side resources associated with this object, regardless of whether other client-side objects
+     * exist that also use that object. Should not be called after fetch() has been invoked.
      */
     @JsMethod
     public void close() {
-        connection.releaseTicket(ticket.getTicket());
+        if (!fetched.isAvailable()) {
```
mofojed, line 127: Is it worth warning in the `else` case? Or a debug log?

mofojed (changes requested):
Minor doc correction, couple of questions/comments.

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
niloc132: it sure can - make some python object that doesnt have a plugin, and send it from the server using `stream.on_data(..., [my_obj])`. The object will be exported, the client will have a reference to it (but can't `fetch()` it), can pass that back to the server via `dh.Widget.sendMessage()`.

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
niloc132: Oh right - I can make it less clean and less round trippy....

niloc132: Hmm. I can, but then there will be zero round trips, and we would be ignoring failure (if any) from the exportFromTIcket call, and would only find out after trying to use the resulting object.

In that case, we could stop returning promise entirely.

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
mofojed: Ah yea true. Client will want to close those, but not fetch those... If you do try and fetch it, it'll reject, but then you won't be able to close it. Release the ticket here instead, since they already screwed up trying to fetch something they shouldn't?

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
mofojed: I'm indifferent. Could file a ticket to optimize this in the future?

web/client-api/src/main/java/io/deephaven/web/client/api/widget/JsWidgetExportedObject.java
niloc132: I've locally added a guard to prevent fetching (and throw an error in that case), but close() can still be done later.

mofojed (approved):

PR #4941, opened 2023-12-13
Docs CI V2 Rework For Old Rsync

stanbrub:
The original Docs CI V2 did not take into account that the deephaven.io production server uses an older version of Ubuntu (18.04) with an older version of rsync (3.1.2).  So there were errors on a new-ish option (--mkpath) for creating the release/v* destination directory.

- Added a step to make _deephaven-core-v2/*_ directories with _rsync_ without "--delete"
- Removed "--mkpath" option on all jobs
- Added "needs" to doc deploy jobs to ensure _makedirs_ happens first
- Removed unnecessary _actions/checkout_
- Removed unnecessary _concurrency_

### DIFF
--- a/.github/workflows/docs-ci-v2.yml
+++ b/.github/workflows/docs-ci-v2.yml
@@ -10,14 +10,7 @@ jobs:
   makedirs:
     if: ${{ github.event_name == 'push' }}
     runs-on: ubuntu-22.04
-    concurrency:
-      group: makdirs-${{ github.workflow }}-${{ github.ref }}
-      # We don't want to cancel in-progress jobs against main because that might leave the upload in a bad state.
-      cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-
       - name: Make Directories
         run: |
           mkdir -p tmp-deephaven-core-v2/${{ github.ref_name }}/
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-ci-v2.yml
```diff
@@ -7,21 +7,36 @@ on:
     branches: [ 'main', 'release/v*' ]
 
 jobs:
+  makedirs:
+    if: ${{ github.event_name == 'push' }}
```
devinrsmith, line 11: This causes all the downstream jobs to get skipped it seems? I'm not sure if there is another form of dependency that says also runs if the parent job was not run?

stanbrub: I think what I just checked in could work.  The "needs [makedirs]" really is for ordering.  So I moved the conditional to the steps.  So _makedirs_ should run regardless, but if it's not a push, the steps won't run.  I'm assuming the job is still seen as run even if no steps run?

.github/workflows/docs-ci-v2.yml
```diff
@@ -7,21 +7,37 @@ on:
     branches: [ 'main', 'release/v*' ]
 
 jobs:
+  makedirs:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Make Directories
+        if: ${{ github.event_name == 'push' }}
```
devinrsmith, line 14: I don't think it makes sense to conditionally run this - might as well run it every time? Otherwise, looks good.

devinrsmith (approved):

PR #4944, opened 2023-12-13
Barrage Subscription/Snapshot Future's Manage Additional Liveness Until First #get

### LINKED ISSUE #4928: Futures used in BarrageSubscription support need to ensure liveness until their result is published
**Description**

When we return a `Future` from these APIs, we must ensure that result/intermediate liveness is guaranteed until we have provided the result to the user, after which liveness is their problem.

The solution probably looks something like:
1. Acquiring an extra reference count to the result when the future is constructed.
2. `get()` calling `LivenessArtifact.manageWithEnclosingScope()` on the result
3. Dropping the extra reference count on cancel, error arrival, or first successful `get()`

nbauernfeind:
Fixes #4928.

### DIFF
--- a/java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
+++ b/java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
@@ -171,6 +171,11 @@ public class BarrageSubscriptionImpl extends ReferenceCountedLivenessNode implem
     }
 
     @Override
+    public TableDefinition getTableDefinition() {
+        return resultTable.getDefinition();
+    }
+
+    @Override
     public Future<Table> entireTable() {
         return partialTable(null, null, false);
     }
--- a/java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
+++ b/java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
@@ -18,6 +18,7 @@ import io.deephaven.engine.rowset.WritableRowSet;
 import io.deephaven.engine.table.Table;
 import io.deephaven.engine.table.TableDefinition;
 import io.deephaven.engine.table.impl.util.BarrageMessage;
+import io.deephaven.engine.updategraph.DynamicNode;
 import io.deephaven.engine.updategraph.UpdateGraph;
 import io.deephaven.engine.updategraph.UpdateGraphAwareCompletableFuture;
 import io.deephaven.extensions.barrage.BarrageSubscriptionOptions;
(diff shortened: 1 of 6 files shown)


java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
```diff
@@ -456,30 +456,153 @@ private interface FutureAdapter extends Future<Table> {
...
+        volatile int wasReleased;
+
         public UpdateGraphAwareFutureAdapter(@NotNull final UpdateGraph updateGraph) {
             super(updateGraph);
+            resultTable.incrementReferenceCount();
         }
 
         @Override
         public boolean cancel(boolean mayInterruptIfRunning) {
+            maybeRelease();
```
rcaudy, line 557: I am nervous about releasing before we send our cancel.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
```diff
@@ -456,30 +456,153 @@ private interface FutureAdapter extends Future<Table> {
...
+     * <p>
+     * Subsequent calls to {@link Future#get get()} will only succeed if the result is still alive and will increase the
+     * the reference count of the result table.
+     */
     private class CompletableFutureAdapter extends CompletableFuture<Table> implements FutureAdapter {
+
+        volatile int wasReleased;
+
+        public CompletableFutureAdapter() {
+            resultTable.incrementReferenceCount();
```
rcaudy, line 476: Sent patch changing this to work around a "construction liveness scope".

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
```diff
@@ -456,30 +456,153 @@ private interface FutureAdapter extends Future<Table> {
...
+            return super.completeExceptionally(ex);
+        }
+
+        @Override
+        public Table get(final long timeout, @NotNull final TimeUnit unit)
+                throws InterruptedException, ExecutionException, TimeoutException {
+            try {
+                final Table result = super.get(timeout, unit);
+
+                if (result instanceof LivenessArtifact) {
```
rcaudy, line 577: Don't manage the result if it's not refreshing.

java-client/barrage/src/main/java/io/deephaven/client/impl/BarrageSubscriptionImpl.java
```diff
@@ -454,32 +466,153 @@ private interface FutureAdapter extends Future<Table> {
...
+                final Table result = supplier.get();
+
+                if (result instanceof LivenessArtifact) {
+                    ((LivenessArtifact) result).manageWithCurrentScope();
+                }
+
+                return result;
+            } finally {
+                maybeRelease();
+            }
```
rcaudy, line 492: I think we should not be releasing in the case of a timeout. In that scenario, it remains the future's owner's responsibility to cancel or get the future.

rcaudy (approved):
Merge if tests look good

rcaudy (approved):

Issue #4942, opened 2023-12-13
Introduce python api for server scheduler


niloc132:
Bidirectional plugin authors (including deephaven.ui component authors) often have the need to submit work off-thread that will communicate with the actively connected client. We generally don't want to encourage making a new thread for each use case, so instead we should offer access to existing applicable threadpools.

In this case, the server already starts a set of threads, for "serial" and "concurrent" work. In the future we might introduce a "fast" option as well, for https://github.com/deephaven/deephaven-core/issues/3041. Typically, calls that will take an exclusive UGP lock should use the serial executor, and calls that will take the shared lock should use the concurrent executor. Calls that will not touch the update graph in any way (including no table data or operatoins, etc) at all might want to use the future "fast" executor.

Somewhat blocked on #4040, where ongoing refactoring will make it simpler to cleanly provide access to server threads after the python script session has been created.

PR #4949, opened 2023-12-14
Python API to access Java server threads

niloc132:
Provides access to server thread pools from python, enabling plugin or
dhui component authors to move work off-thread as appropriate, without
introducing their own threads.

Partial https://github.com/deephaven/deephaven-core/issues/4942


py/server/deephaven/server/executors.py
```diff
@@ -0,0 +1,56 @@
+#
+# Copyright (c) 2016-2022 Deephaven Data Labs and Patent Pending
```
jmao-denver, line 2:
```suggestion
# Copyright (c) 2016-2023Deephaven Data Labs and Patent Pending
```

py/server/deephaven/server/executors.py
```diff
@@ -0,0 +1,56 @@
...
+
+from typing import Callable, Dict, List
+import jpy
+from deephaven.jcompat import j_runnable
+
+
+_executors: Dict[str, Callable[[Callable[[], None]], None]] = {}
+
+
+def has_named_executor(executor_name: str) -> bool:
```
jmao-denver, line 19: Is it even possible to have 'unnamed' executors? If not, maybe just rename to `has_executor`.

py/server/deephaven/server/executors.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Raises: KeyError if the executor name
+    """
+    _executors[executor_name](task)
+
+
+def _register_named_java_executor(executor_name: str, java_executor: jpy.JType) -> None:
+    """
+    Provides a Java executor for user code to submit tasks to.
+
```
jmao-denver, line 51: Should we highlight that it will be called by the server during the Python session initialization?

py/server/deephaven/server/executors.py
```diff
@@ -0,0 +1,56 @@
...
+
+def executor_names() -> List[str]:
+    """
+    Returns: the List of known executor names
+    """
+    return list(_executors.keys())
+
+
+def submit_task(executor_name: str, task: Callable[[], None]) -> None:
```
jmao-denver, line 33: Is a task truly a callable with zero parameters?

py/server/deephaven/server/executors.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Submits a task to run on a named executor. If no such executor exists, raises KeyError.
+
+    Typically, tasks should not block on other threads. Ensure tasks never block on other tasks submitted to the same executor.
+
+    Args:
+        executor_name: the name of the executor to submit the task to
+        task: the function to run on the named executor
```
jmao-denver, line 41:
```suggestion
        executor_name (str): the name of the executor to submit the task to
        task (Callable[[], None]): the function to run on the named executor
```

py/server/deephaven/server/executors.py
niloc132: Yes, no params - should it be None instead?

py/server/deephaven/server/executors.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Raises: KeyError if the executor name
+    """
+    _executors[executor_name](task)
+
```
jmao-denver, line 46: Should we worry about possible exceptions?

py/server/deephaven/server/executors.py
```diff
@@ -0,0 +1,56 @@
...
+def _register_named_java_executor(executor_name: str, java_executor: jpy.JType) -> None:
+    """
+    Provides a Java executor for user code to submit tasks to.
+
+    Args:
+        executor_name: the name of the executor to register
+        java_executor: a Java Consumer<Runnable> instance
+    """
+    _executors[executor_name] = lambda task: java_executor.accept(j_runnable(task))
```
jmao-denver, line 56: Should we check if the same 'named' executor already exists?

py/server/deephaven/server/executors.py
niloc132: I don't think the executors can throw, except maybe if the server has already shut down (in which case there are lots of things that won't work), aside from the KeyError there are no reasonable errors that can be handled?

jmao-denver (commented):
Minor things (style, docstrings)

py/server/deephaven/server/executors.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def _register_named_java_executor(executor_name: str, java_executor: jpy.JType) -> None:
+    """
+    Provides a Java executor for user code to submit tasks to. Called during server startup.
+
+    Args:
+        executor_name (str): the name of the executor to register
+        java_executor (jpy.JType): a Java Consumer<Runnable> instance
+    """
```
jmao-denver, line 56:
```suggestion
        java_executor (jpy.JType): a Java Consumer<Runnable> instance
        
    Raises:
        DHError
    """
```

py/server/deephaven/server/executors.py
```diff
@@ -0,0 +1,59 @@
...
+    Submits a task to run on a named executor. If no such executor exists, raises KeyError.
+
+    Typically, tasks should not block on other threads. Ensure tasks never block on other tasks submitted to the same executor.
+
+    Args:
+        executor_name (str): the name of the executor to submit the task to
+        task (Callable[[], None]): the function to run on the named executor
+
+    Raises: KeyError if the executor name
```
jmao-denver, line 44:
```suggestion
    Raises: 
        KeyError if the executor name
```

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -120,9 +124,20 @@ public PythonDeephavenSession(
...
         publishInitial();
     }
 
+    private void registerJavaExecutor() {
+        // TODO (deephaven-core#4040) Temporary exec service until we have cleaner startup wiring
+        try (PyModule pyModule = PyModule.importModule("deephaven.server.executors");
+                final PythonDeephavenThreadsModule module = pyModule.createProxy(PythonDeephavenThreadsModule.class)) {
+            ExecutorService executorService = Executors.newFixedThreadPool(1);
+            module._register_named_java_executor("serial", executorService::submit);
+            module._register_named_java_executor("concurrent", executorService::submit);
```
jmao-denver, line 137: Handle possible duplication exception from Python?

niloc132: Given that these are the only two calls here, back to back, I dont think this is possible? - or if it is, it should just throw and be handled above here as "everything is broken, shut down", etc?

jmao-denver (dismissed):
lgtm

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -120,9 +124,20 @@ public PythonDeephavenSession(
...
 
+        registerJavaExecutor();
         publishInitial();
     }
 
+    private void registerJavaExecutor() {
+        // TODO (deephaven-core#4040) Temporary exec service until we have cleaner startup wiring
+        try (PyModule pyModule = PyModule.importModule("deephaven.server.executors");
+                final PythonDeephavenThreadsModule module = pyModule.createProxy(PythonDeephavenThreadsModule.class)) {
+            ExecutorService executorService = Executors.newFixedThreadPool(1);
```
rcaudy, line 135: Don't we want all of our fancy python debugging support for this thread?

niloc132: Rebased (to get the changes where that was passed to this type) and added the thread initializer wiring

jmao-denver (approved):
Reapprove after Java only change

nbauernfeind (approved):

PR #4953, opened 2023-12-15
Distinguish between np type hints from Py ones

### LINKED ISSUE #4933: Unexpected change with type hints not returning as specified type
If I type hint as an `int`, I am unexpectedly getting back a `numpy.int64`. If I don't type hint at all I get an `int`.

```
def my_type(x):
   print(type(x))
t = empty_table(1).update(["x = i", "test = my_type(x)"])
>>> <class 'int'>

def my_type(x:int): # note typing
   print(type(x))
t = empty_table(1).update(["x = i", "test = my_type(x)"])
>>> <class 'numpy.int64'>
```

Note, same goes for float.

jmao-denver:
Fixes #4933

### DIFF
--- a/py/server/tests/test_udf_numpy_args.py
+++ b/py/server/tests/test_udf_numpy_args.py
@@ -392,6 +392,41 @@ def test_udf(col: Optional[{np_type}]) -> bool:
             t2 = t.update(["X1 = f31(null, Y)"])
             self.assertEqual(10, t2.to_string("X1").count("true"))
 
+    def test_non_np_typehints(self):
+        py_types = {"int", "float"}
+
+        for p_type in py_types:
+            with self.subTest(p_type):
+                func_str = f"""
+def f(x: {p_type}) -> bool:  # note typing
+    return type(x) == {p_type}
+"""
+                exec(func_str, globals())
+                t = empty_table(1).update(["X = i", f"Y = f(({p_type})X)"])
+                self.assertEqual(1, t.to_string(cols="Y").count("true"))
+
+
+        np_int_types = {"np.int8", "np.int16", "np.int32", "np.int64"}
+        for p_type in np_int_types:
+            with self.subTest(p_type):
+                func_str = f"""
+def f(x: {p_type}) -> bool:  # note typing
+    return type(x) == {p_type}
+"""
+                exec(func_str, globals())
+                t = empty_table(1).update(["X = i", f"Y = f(X)"])
+                self.assertEqual(1, t.to_string(cols="Y").count("true"))
+
+        np_floating_types = {"np.float32", "np.float64"}
+        for p_type in np_floating_types:
+            with self.subTest(p_type):
+                func_str = f"""
+def f(x: {p_type}) -> bool:  # note typing
+    return type(x) == {p_type}
+"""
+                exec(func_str, globals())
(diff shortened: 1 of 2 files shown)


py/server/tests/test_udf_numpy_args.py
```diff
@@ -392,6 +392,41 @@ def f31(p1: Optional[np.bool_], p2=None) -> bool:
...
+
+        np_floating_types = {"np.float32", "np.float64"}
+        for p_type in np_floating_types:
+            with self.subTest(p_type):
+                func_str = f"""
+def f(x: {p_type}) -> bool:  # note typing
+    return type(x) == {p_type}
+"""
+                exec(func_str, globals())
+                t = empty_table(1).update(["X = i", f"Y = f((float)X)"])
```
chipkent, line 428: This looks like it is only testing the `np.float32` case or is casting.

py/server/tests/test_udf_numpy_args.py
```diff
@@ -392,6 +392,41 @@ def f31(p1: Optional[np.bool_], p2=None) -> bool:
...
+        py_types = {"int", "float"}
+
+        for p_type in py_types:
+            with self.subTest(p_type):
+                func_str = f"""
+def f(x: {p_type}) -> bool:  # note typing
+    return type(x) == {p_type}
+"""
+                exec(func_str, globals())
+                t = empty_table(1).update(["X = i", f"Y = f(({p_type})X)"])
```
chipkent, line 405: Seems like this should also be testing with `long` and `double` since the python types are wide.

py/server/tests/test_udf_numpy_args.py
```diff
@@ -392,6 +392,41 @@ def f31(p1: Optional[np.bool_], p2=None) -> bool:
...
+
+        np_int_types = {"np.int8", "np.int16", "np.int32", "np.int64"}
+        for p_type in np_int_types:
+            with self.subTest(p_type):
+                func_str = f"""
+def f(x: {p_type}) -> bool:  # note typing
+    return type(x) == {p_type}
+"""
+                exec(func_str, globals())
+                t = empty_table(1).update(["X = i", f"Y = f(X)"])
```
chipkent, line 417: This looks like it is only testing the `np.int32` case or is doing casting.

py/server/tests/test_udf_numpy_args.py
jmao-denver: These tests are just to make sure that the converted args' types are what we expected, the actual values are not a concern here and they are tested in other test cases in the same test class.

py/server/tests/test_udf_numpy_args.py
jmao-denver: These tests are just to make sure that the converted args' types are what we expected, the exact types of the values are not a concern here and they are tested in other test cases in the same test class.

py/server/tests/test_udf_numpy_args.py
jmao-denver: see the comment above.

jmao-denver: We don't yet statically check if the parameter types and argument types match. That's something for the improvement of the QLP. Runtime exception however will be thrown if the arg value is outside the range of a numpy integer type.

chipkent (approved):

PR #4961, opened 2023-12-19
Remove unused/broken tests

### LINKED ISSUE #734: Fix DB-test python tests

niloc132:
Closes #734


rcaudy (approved):

jmao-denver (approved):
LGTM

PR #4962, opened 2023-12-19
Rework server tests to avoid using test sources from other projects

niloc132:
Also adds missing explicit dependencies that were pulled along implicitly.

Partial #4040


devinrsmith (approved):

server/test/src/main/java/io/deephaven/server/runner/ExecutionContextUnitTestModule.java
rcaudy, line 1: Why is it attractive to move this to a main module?

rcaudy (approved):

PR #4966, opened 2023-12-19
Add dictionary (mapping) as a valid argument for  new_table()

### LINKED ISSUE #4908: Feature request: Table from Python dictionary via either `new_table` or a new method
As a user, it would be nice to create a table directly from a dictionary in Python. As of now, the best way is to go dict -> DataFrame -> table. This is a roundabout way to do it, and is inefficient (even though new_table isn't meant to be super efficient). Pandas can create a DataFrame directly from a dict, so Deephaven should be able to as well.

Alternatively, another method could be used entirely (called `table_from_dict` or whatever) that gives users this ability.

jmao-denver:
Fixes #4908

### DIFF
--- a/py/server/deephaven/table_factory.py
+++ b/py/server/deephaven/table_factory.py
@@ -5,7 +5,7 @@
 """ This module provides various ways to make a Deephaven table. """
 
 import datetime
-from typing import Callable, List, Dict, Any, Union, Sequence, Tuple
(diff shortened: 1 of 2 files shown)


chipkent (approved):

deephaven-internal:

How-to: https://github.com/deephaven/deephaven.io/issues/3571
Reference: https://github.com/deephaven/deephaven.io/issues/3572

PR #4970, opened 2023-12-20
Extract variable api from ScriptSession, let ScriptSession guard reads

niloc132:
This patch moves from assuming that the ScriptSession owns the ExecContext for locking (with UpdateGraph) and variable access (with QueryScope). Liveness is also moved to the QueryScope, as is the responsibility of unwrapping objects (but delegated to ScriptSession as needed).

Previously, the groovy script session used a lock to interact with its query scope, but python did not under the incorrect assumption that the GIL would guard those reads. Instead, a read/write lock is used to ensure that reads block while an operation is running that could affect values, but avoiding reads blocking each other.

Partial #4040


engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -262,47 +276,144 @@ public Class<?> getVariableType(final String var) {
...
+     * @param name the variable name to set
+     * @param value the new value of the variable
+     */
+    protected abstract void setVariable(String name, @Nullable Object value);
+
     @Override
     public VariableProvider getVariableProvider() {
-        return this;
-    }
+        return new VariableProvider() {
```
niloc132, line 341: Arguably this interface and impl should go away, and QueryScope should be used by autocomplete instead. At this point it still exists to allow direct access to script session variables instead of the "more correct" way of going through the QueryScope.

Note that the abstract methods above are to be implemented by each language, but the locking/etc is handled by the VariableProvider/QueryScope, so those are the public interfaces to get access to data.

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -144,13 +143,6 @@ public Thread newThread(@NotNull Runnable r) {
         }
     }
 
-    @Override
-    @VisibleForTesting
-    public QueryScope newQueryScope() {
-        // depend on the GIL instead of local synchronization
```
niloc132, line 150: Note that this comment was never accurate, even in DHE - python can release the GIL whenever it wants to let another thread advance.

server/src/test/java/io/deephaven/server/appmode/ApplicationTest.java
```diff
@@ -29,15 +29,7 @@ public class ApplicationTest {
     @Rule
     public final EngineCleanup base = new EngineCleanup();
 
-    private AbstractScriptSession session = null;
-
-    @After
-    public void tearDown() {
```
nbauernfeind, line 35: We still need to do something to cleanup the session here. I'm thinking we probably want to `forceReferenceCountToZero` which will `destroy()` (and remove any reference counts the session holds).

(Not sure why I didn't find this reference to `release` when we were chatting.)

server/test/src/main/java/io/deephaven/server/test/FlightMessageRoundTripTest.java
```diff
@@ -482,7 +481,7 @@ public void testLoginHeaderAnonymous() {
         final String ANONYMOUS = "Anonymous";
 
         closeClient();
-        scriptSession.setVariable("test", TableTools.emptyTable(10).update("I=i"));
+        scriptSession.getVariableProvider().setVariable("test", TableTools.emptyTable(10).update("I=i"));
```
nbauernfeind, line 484: Should this match the others?

server/test/src/main/java/io/deephaven/server/test/FlightMessageRoundTripTest.java
```diff
@@ -681,10 +680,10 @@ public void testGetSchema() {
         final Table tickingTable = ExecutionContext.getContext().getUpdateGraph().sharedLock().computeLocked(
                 () -> TableTools.timeTable(1_000_000).update("I = i"));
 
-        try (final SafeCloseable ignored = LivenessScopeStack.open(scriptSession, false)) {
+        try (final SafeCloseable ignored = LivenessScopeStack.open(scriptSession.getQueryScope(), false)) {
```
nbauernfeind, line 683: `scriptSession.getQueryScope` seems slightly silly. Better would be to replace the existing `scriptSession` variable with the `QueryScope` instance. I'd use the member variable over `ExecutionContext.getContext().getQueryScope()`, too.

java-client/session-dagger/src/test/java/io/deephaven/client/ObjectServiceTest.java
```diff
@@ -94,7 +94,7 @@ public MyObjects(Table table, Figure customObject) {
     @Test
     public void fetchable() throws ExecutionException, InterruptedException, TimeoutException,
             InvalidProtocolBufferException, TableHandleException {
-        getScriptSession().setVariable("my_objects", myObjects());
+        getScriptSession().getVariableProvider().setVariable("my_objects", myObjects());
```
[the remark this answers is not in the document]
niloc132: That is fixed on my branch locally, this particular change predates our discussion, and I hadn't rewritten this one yet. 

I still have one or two tests relying on getVariableProvider, will validate that they actually need it...

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -282,7 +283,7 @@ public void bindTableToVariable(
                 ScriptSession scriptSession =
                         exportedConsole != null ? exportedConsole.get() : scriptSessionProvider.get();
                 Table table = exportedTable.get();
-                scriptSession.setVariable(request.getVariableName(), table);
+                ExecutionContext.getContext().getQueryScope().putParam(request.getVariableName(), table);
```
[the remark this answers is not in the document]
niloc132: Note that ScriptSession.getExecContext() is going away in the next-ish patch - but either ScriptSession.getVariableProvider() or ScriptSession.getQueryScope() could suite this.

But really, this method should probably go away in favor of the new publish methods in sessionservice? I'm not confident enough to change the code now, but worth considering.

server/test/src/main/java/io/deephaven/server/test/FlightMessageRoundTripTest.java
niloc132: Oddly, unlike the others, this passes as written. Will investigate.

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -262,47 +276,144 @@ public Class<?> getVariableType(final String var) {
...
+    /**
+     * Retrieve a variable from the script session's bindings.
+     * <p/>
+     * Please use {@link ScriptSession#getVariable(String, Object)} if you expect the variable may not exist.
+     *
+     * @param name the variable to retrieve
+     * @return the variable
+     * @throws QueryScope.MissingVariableException if the variable does not exist
+     */
+    @NotNull
```
niloc132, line 294: This annotation was wrong, variables can be bound as null.

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
niloc132: I think it makes sense to do it now, but I wanted to raise the possibility of keeping separate APIs, with "this is for query scope, which comes from the exec context, and might not be related to a script" vs "this is a specific way to interact with variables in the script session".

Another option (that we're both leaning towards) is to let ScriptSession have a public `getQueryScope()`, so that code that has a script session can be more generic...

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -322,7 +433,7 @@ protected <T> QueryScopeParam<T> createParam(final String name)
                 throw new QueryScope.MissingVariableException("Name " + name + " is invalid");
             }
             // noinspection unchecked
-            return new QueryScopeParam<>(name, (T) scriptSession.getVariable(name));
+            return new QueryScopeParam<>(name, (T) getVariableProvider().getVariable(name, null));
```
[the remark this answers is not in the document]
niloc132: Delegating to readParamValue, I think that is equivalent, lets us only have one place to change if we keep tweaking that behavior.

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -331,57 +442,28 @@ public <T> T readParamValue(final String name) throws QueryScope.MissingVariable
                 throw new QueryScope.MissingVariableException("Name " + name + " is invalid");
             }
             // noinspection unchecked
-            return (T) scriptSession.getVariable(name);
+            return (T) getVariableProvider().getVariable(name, null);
```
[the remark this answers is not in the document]
niloc132: ...it is now, with a throws above it

server/src/main/java/io/deephaven/server/console/ConsoleServiceGrpcImpl.java
```diff
@@ -305,7 +306,7 @@ public StreamObserver<AutoCompleteRequest> autoCompleteStream(
                 final ScriptSession scriptSession = scriptSessionProvider.get();
                 scriptSession.evaluateScript(
                         "from deephaven_internal.auto_completer import jedi_settings ; jedi_settings.set_scope(globals())");
-                settings[0] = (PyObject) scriptSession.getVariable("jedi_settings");
```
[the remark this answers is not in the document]
niloc132: I think it is safe either way, but VariableProvider.getVariable does not have an overload that only takes a name, a default value is required.

For QueryScope, I'll rewrite to use the throwing impl instead

server/src/main/java/io/deephaven/server/console/completer/PythonAutoCompleteObserver.java
```diff
@@ -110,7 +113,7 @@ private void handleAutocompleteRequest(AutoCompleteRequest request,
                 request.getRequestId() > 0 ? request.getRequestId() : request.getGetCompletionItems().getRequestId();
         try {
             final ScriptSession scriptSession = exportedConsole.get();
-            PyObject completer = (PyObject) scriptSession.getVariable("jedi_settings");
+            PyObject completer = (PyObject) scriptSession.getVariableProvider().getVariable("jedi_settings", null);
```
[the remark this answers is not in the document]
niloc132: I have a major problem with scope creep in these refactors, and would rather leave this to someone else who wants to overhaul autocomplete (again).

server/src/test/java/io/deephaven/server/appmode/ApplicationTest.java
niloc132: Shouldn't the EngineCleanup manage this? Its own setup call created a liveness scope, and will close it during teardown.

server/test/src/main/java/io/deephaven/server/test/FlightMessageRoundTripTest.java
niloc132: TBH I'm not sure why this has a liveness scope like this anyway - shouldn't we have a liveness scope that is opened in before, closed in after?

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -141,8 +151,9 @@ public synchronized final Changes evaluateScript(final String script, @Nullable
         final Changes diff;
         // retain any objects which are created in the executed code, we'll release them when the script session
         // closes
+        variableAccessLock.writeLock().lock();
```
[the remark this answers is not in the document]
niloc132: Let's take a few minutes today and consider the three locks in play?
 * UG exclusive/shared lock - exclusive lock is held for the duration of any executed script code
 * ScriptSession state read/write lock, preventing any (well, most) writes while anyone is actively reading
 * ScriptSession instance lock, appears only used to guard snapshots for scope changes. Probably can be folded into the new read/write lock.

External callers could already hold UG shared/exclusive locks when reading from QueryScope, so with the current code we could hit a lock inversion issue - take the write lock, wait to acquire the UG exclusive lock, then in another thread while holding a either UG lock try and acquire the read lock.

I think we can simply move the writelock inside `apply(() -> {...})`, plus as above remove the `synchronized` keyword usage?

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -245,12 +258,13 @@ protected void destroy() {
     protected abstract void evaluate(String command, @Nullable String scriptName);
 
     /**
-     * @return a query scope for this session; only invoked during construction
+     * @return a query scope that wraps
```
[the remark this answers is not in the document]
niloc132: Fixed, pulled up to interface

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -262,47 +276,144 @@ public Class<?> getVariableType(final String var) {
     }
 
 
-    @Override
     public TableDefinition getTableDefinition(final String var) {
```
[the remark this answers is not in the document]
niloc132: Removed entirely, it only ever had one caller.

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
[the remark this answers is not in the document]
niloc132: One more reply from me, then i'll defer to when we do a call: it isnt quite so simple to remove synchronized, since you can't upgrade a read lock to a write lock. If the whole write lock needs to be wider (encompass calls to takeSnapshot, observeScopeChanges, possibly createDiff), then the UG exclusive lock needs to also be correspondingly bigger as well.

server/src/test/java/io/deephaven/server/appmode/ApplicationTest.java
nbauernfeind: Good point; you are right.

server/test/src/main/java/io/deephaven/server/test/FlightMessageRoundTripTest.java
nbauernfeind: Agreed given our offline chat. I suspect these are OK without the liveness scopes.

engine/table/src/main/java/io/deephaven/engine/util/NoLanguageDeephavenSession.java
```diff
@@ -35,32 +33,16 @@ public NoLanguageDeephavenSession(final UpdateGraph updateGraph,
...
-
-    @Override
-    public <T> T getVariable(String name, T defaultValue) {
-        try {
-            // noinspection unchecked
-            return (T) getVariable(name);
-        } catch (QueryScope.MissingVariableException e) {
-            return defaultValue;
+    protected <T> T getVariable(String name) {
+        if (!variables.containsKey(name)) {
```
nbauernfeind, line 41: This needs to be synchronized across the map like you did for the Groovy impl.

engine/table/src/main/java/io/deephaven/engine/util/NoLanguageDeephavenSession.java
```diff
@@ -87,23 +69,18 @@ protected void evaluate(String command, @Nullable String scriptName) {
...
     @Override
-    public Map<String, Object> getVariables() {
-        return Collections.unmodifiableMap(variables);
-    }
-
-    @Override
-    public Set<String> getVariableNames() {
-        return Collections.unmodifiableSet(variables.keySet());
+    protected Set<String> getVariableNames() {
+        return Set.copyOf(variables.keySet());
```
nbauernfeind, line 73: Also synchronized on the map?

engine/table/src/test/java/io/deephaven/engine/table/impl/FuzzerTest.java
```diff
@@ -117,7 +117,7 @@ private void testFuzzerScriptFile(final long timeSeed, String s, boolean realtim
             clock.now += DateTimeUtils.SECOND / 10 * timeRandom.nextInt(20);
         }
 
-        final TimeTable timeTable = (TimeTable) session.getVariable("tt");
+        final TimeTable timeTable = session.getQueryScope().readParamValue("tt", null);
```
nbauernfeind, line 120: Won't this NPE if it doesn't exist? Probably you meant to leave this as throwing a MissingVarException?

open-api/lang-tools/src/main/java/io/deephaven/lang/completion/ChunkerCompleter.java
```diff
@@ -1018,9 +1018,12 @@ private Optional<Class<?>> resolveScopeType(
 
         IsScope o = scope.get(0);
 
-        final Class<?> type = variables.getVariableType(o.getName());
-        if (type != null) {
-            return Optional.of(type);
+        final Object instance = variables.readParamValue(o.getName(), null);
+        if (instance != null) {
+            if (instance instanceof Table) {
+                return Optional.of(Table.class);
```
nbauernfeind, line 1024: I don't get this. Why are we restricting sub-classes of Table to the top-level interface? Does it solve a real problem?

niloc132: This is inlining the old implementation that VariableProvider did, replicated without additional thought.

engine/table/src/test/java/io/deephaven/engine/table/impl/FuzzerTest.java
niloc132: You're not wrong, but the test will fail either way.

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -139,15 +141,17 @@ public synchronized final Changes evaluateScript(final String script, @Nullable
...
         // closes
         try (final S initialSnapshot = takeSnapshot();
-                final SafeCloseable ignored = LivenessScopeStack.open(this, false)) {
+                final SafeCloseable ignored = LivenessScopeStack.open(queryScope, false)) {
 
             try {
                 // Actually evaluate the script; use the enclosing auth context, since AbstractScriptSession's
                 // ExecutionContext never has a non-null AuthContext
                 executionContext.withAuthContext(ExecutionContext.getContext().getAuthContext())
+                        .withQueryScope(queryScope)
```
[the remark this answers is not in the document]
niloc132: It might not be necessary today, that's correct, but if not needed, it will be a no-op. Where possible (and straightforward, and obviously not-incorrect), Ryan and I were trying to make this ready for multiple script sessions, so if it does nothing, I'd be inclined to keep it.

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -2710,10 +2710,10 @@ private void prepareVectorizationArgs(MethodCallExpr n, QueryScope queryScope, E
                 addLiteralArg(expressions[i], argTypes[i], pyCallableWrapper);
             } else if (expressions[i] instanceof NameExpr) {
                 String name = expressions[i].asNameExpr().getNameAsString();
-                try {
+                if (queryScope.hasParamName(name)) {
                     Object param = queryScope.readParamValue(name);
                     pyCallableWrapper.addChunkArgument(new ConstantChunkArgument(param, argTypes[i]));
-                } catch (QueryScope.MissingVariableException ex) {
+                } else {
```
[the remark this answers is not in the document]
niloc132: Correct, doesn't actually work, will revert.

Note that `Optional.ofNullable(null)` returns `empty()`, not "an optional instance that holds null".

server/src/test/java/io/deephaven/server/runner/DeephavenApiServerTestBase.java
```diff
@@ -86,7 +86,8 @@ interface Builder {
     @Rule
     public final GrpcCleanupRule grpcCleanup = new GrpcCleanupRule();
 
-    private ExecutionContext executionContext;
+    @Inject
+    ExecutionContext executionContext;
```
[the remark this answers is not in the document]
niloc132: It needs to be non-private to be injectable as a field (as is apparently the convention in this class), for the `setUp()` code:
```java
        DaggerDeephavenApiServerTestBase_TestComponent.builder()
                .withServerConfig(config)
                .withAuthorizationProvider(new CommunityAuthorizationProvider())
                .withOut(System.out)
                .withErr(System.err)
                .build()
                .injectFields(this);
```

The resulting generated DaggerDeephavenApiServerTestBase_TestComponent &co will assign to these fields from outside the class, so injected fields cannot be private.

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -289,34 +266,43 @@ private Object maybeUnwrap(PyObject o) {
     }
 
     @Override
-    public Set<String> getVariableNames() {
-        return Collections.unmodifiableSet(scope.getKeys().collect(Collectors.toSet()));
+    protected Set<String> getVariableNames() {
+        return scope.getKeys().collect(Collectors.toUnmodifiableSet());
```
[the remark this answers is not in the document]
niloc132: Good points, will pass a predicate, stream results into the unmodifiable set.

Racing with setVariable is somewhat fraught - after all, setVariable isn't the only way to assign values into the collection, python itself could be assigning/removing something. I think technically I can use PyLib.ensureGil() here, to guarantee that reads of python structures can't change while reading, in place of the `synchronized` blocks that Groovy's impl uses). Any AbstractScriptSession override that makes more than one call into python should use this (and note that it is _not_ safe to use this to call into arbitrary python, as that python code could end up dropping the GIL despite our efforts - ensureGil only lets us be sure that we have the GIL in Java, JNI, and our own C).

engine/context/src/main/java/io/deephaven/engine/context/QueryScope.java
```diff
@@ -21,7 +24,7 @@
 /**
  * Variable scope used to resolve parameter values during query execution.
  */
-public abstract class QueryScope implements LogOutputAppendable {
+public abstract class QueryScope extends LivenessScope implements LogOutputAppendable {
```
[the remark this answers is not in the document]
niloc132: `REFERENCE_TRACKING_DISABLED` is package-protected - I think the rest makes sense, I'll push a patch shortly with the rest of this for another pass.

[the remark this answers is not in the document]
niloc132: This breaks "opening" the queryscope as a liveness scope when executing code:
```
        // retain any objects which are created in the executed code, we'll release them when the script session
        // closes
        try (final S initialSnapshot = takeSnapshot();
                final SafeCloseable ignored = LivenessScopeStack.open(queryScope, false)) {
```
We could go back to passing `this`, but that seems to defeat the purpose of that change...

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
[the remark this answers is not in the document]
nbauernfeind: Ah yes, what a bogus API `ofNullable`. My apologies; I misread the javadoc _and the implementation_..

engine/context/src/main/java/io/deephaven/engine/context/QueryScope.java
```diff
@@ -119,6 +114,14 @@ private static Object applyValueConversions(final Object value) {
         return value;
     }
 
+    protected QueryScope() {
+        super(false);
+
+        // if (!Liveness.REFERENCE_TRACKING_DISABLED) {
```
nbauernfeind, line 120: You're right (from slack chat) -- we should likely just check this inside of `incrementReferenceCount` which is a "public" API and used directly elsewhere (like directly by the `JobScheduler`). Let's get Ryan's opinion on this.

nbauernfeind (approved):
LGTM; I believe Ryan wants to pass through this, too (but if not, it's good to me).

engine/context/src/main/java/io/deephaven/engine/context/QueryScope.java
rcaudy: To clarify a little bit:
`incrementReferenceCount()` is a public member of `ReferenceCounted`, but it's an implementation detail for `ReferenceCountedLivenessNode`. It's OK for an implementation to use internally if there's a good reason, but really `retainReference()` is the public API (and the implementation does check that flag).

niloc132:
Passing nightlies: https://github.com/niloc132/deephaven-core/actions/runs/7455151234

niloc132:
Another nightly run after merging and dealing with conflicts: https://github.com/niloc132/deephaven-core/actions/runs/7467840686

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -41,14 +41,13 @@
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
```
rcaudy, line 45: Unused

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -295,34 +274,49 @@ private Object maybeUnwrap(PyObject o) {
...
+            } else {
+                Object wrapped;
+                if (newValue instanceof PyObject) {
+                    wrapped = newValue;
+                } else {
+                    wrapped = PythonObjectWrapper.wrap(newValue);
+                }
+                // This isn't thread safe, we're relying on the GIL being kind to us (as we have historically done).
+                // There
+                // is no built-in for "replace a variable and return the old one".
```
rcaudy, line 306:
```suggestion
                // There is no built-in for "replace a variable and return the old one".
```

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -295,34 +274,49 @@ private Object maybeUnwrap(PyObject o) {
...
+                    return null;
+                }
+            } else {
+                Object wrapped;
+                if (newValue instanceof PyObject) {
+                    wrapped = newValue;
+                } else {
+                    wrapped = PythonObjectWrapper.wrap(newValue);
+                }
+                // This isn't thread safe, we're relying on the GIL being kind to us (as we have historically done).
```
rcaudy, line 304: Is this true, given that we've enclosing with `ensureGil`?

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -295,34 +274,49 @@ private Object maybeUnwrap(PyObject o) {
...
             }
-            globals.setItem(name, newValue);
-        }
+        });
 
         // Observe changes from this "setVariable" (potentially capturing previous or concurrent external changes from
         // other threads)
         observeScopeChanges();
+
+        // This doesn't return the same Java instance of PyObject, so we won't decref it properly, but
```
rcaudy, line 317: I want to make sure I understand this, let's discuss.

rcaudy (commented):
Partial review, I think I'm covering the most important classes, want to discuss.

engine/context/src/main/java/io/deephaven/engine/context/QueryScope.java
```diff
@@ -21,7 +25,7 @@
 /**
  * Variable scope used to resolve parameter values during query execution.
  */
-public abstract class QueryScope implements LogOutputAppendable {
+public abstract class QueryScope extends ReferenceCountedLivenessNode implements LogOutputAppendable {
```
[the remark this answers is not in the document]
niloc132: Based on our call, QueryScope is now a LivenessNode, the StandaloneQueryScope impl is a LivenessArtifact, and the ScriptSessionQueryScope is a LivenessScope. The empty/poisoned implementations throw when asked to manage/unmanage.

engine/context/src/main/java/io/deephaven/engine/context/QueryScope.java
```diff
@@ -34,15 +38,6 @@ public static <T> void addParam(final String name, final T value) {
         ExecutionContext.getContext().getQueryScope().putParam(name, value);
     }
 
-    /**
-     * Adds an object's declared fields to the scope.
-     *
-     * @param object object whose fields will be added.
-     */
-    public static void addObjectFields(final Object object) {
```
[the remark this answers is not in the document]
niloc132: Yes - do we think we need it? DHE didn't use it either, except for a single test (and core+ didn't use it at all).

engine/context/src/main/java/io/deephaven/engine/context/QueryScope.java
```diff
@@ -119,6 +114,12 @@ private static Object applyValueConversions(final Object value) {
         return value;
     }
 
+    protected QueryScope() {
+        super(false);
```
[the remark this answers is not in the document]
niloc132: This question could now apply to ScriptSessionQueryScope, do you still want to consider it?

engine/context/src/main/java/io/deephaven/engine/context/QueryScope.java
```diff
@@ -276,14 +284,19 @@ public <T> T readParamValue(final String name, final T defaultValue) {
 
         @Override
         public <T> void putParam(final String name, final T value) {
+            if (value instanceof LivenessReferent && DynamicNode.notDynamicOrIsRefreshing(value)) {
```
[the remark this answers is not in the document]
niloc132: For StandaloneQueryScope: I'm not aware of another way that things can enter the scope.

For ScriptSessionQueryScope:
The expectation is that if not directly set via putParam, then it was a user from within the script session assigning a global, via executing code. Since executing code requires first opening the query scope as a liveness node, that will then be captured too.

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -243,75 +250,65 @@ protected void destroy() {
...
+     *
+     * @param name the variable name to set
+     * @param value the new value of the variable
+     * @return the old previous value for this name, if any. As with {@link #getVariable(String)}, may need to be
+     *         unwrapped.
+     */
+    protected abstract Object setVariable(String name, @Nullable Object value);
 
     // -----------------------------------------------------------------------------------------------------------------
     // ScriptSession-based QueryScope implementation, with no remote scope or object reflection support
```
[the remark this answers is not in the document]
niloc132: Yes, it looks like that can all go, unless we want the reflective stuff back.

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -243,75 +250,65 @@ protected void destroy() {
...
-    }
+    /**
+     * Inserts a value into the script's scope.
+     *
+     * @param name the variable name to set
+     * @param value the new value of the variable
+     * @return the old previous value for this name, if any. As with {@link #getVariable(String)}, may need to be
+     *         unwrapped.
+     */
+    protected abstract Object setVariable(String name, @Nullable Object value);
```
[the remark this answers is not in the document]
niloc132: This protected method lets each ScriptSession impl define how the variable would be added to that environment, and lets the ScriptSessionQueryScope delegate to that impl.

engine/sql/src/main/java/io/deephaven/engine/sql/Sql.java
```diff
@@ -83,18 +82,16 @@ private static Map<String, Table> currentScriptSessionNamedTables() {
         final Map<String, Table> scope = new HashMap<>();
         // getVariables() is inefficient
         // See SQLTODO(catalog-reader-implementation)
-        for (Entry<String, Object> e : currentScriptSession().getVariables().entrySet()) {
-            if (e.getValue() instanceof Table) {
-                scope.put(e.getKey(), (Table) e.getValue());
+        QueryScope queryScope = ExecutionContext.getContext().getQueryScope();
+        for (String name : queryScope.getParamNames()) {
+            Object paramValue = queryScope.unwrapObject(queryScope.readParamValue(name));
```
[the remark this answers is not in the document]
niloc132: To make the unwrapping safe? At least not for the implementation (only python) we have today, no.

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -243,75 +250,65 @@ protected void destroy() {
...
-        }
-
-        @Override
-        public void putObjectFields(Object object) {
-            throw new UnsupportedOperationException();
-        }
-
+    public class ScriptSessionQueryScope extends QueryScope {
+        /**
+         * Internal workaround to support python calling pushScope.
```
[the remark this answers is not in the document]
niloc132: Yeah, I didn't understand why it existed until I removed it, figured I'd mark it for the next hapless developer.

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
niloc132: Technically no, the GIL doesn't guarantee us practically anything even if this was one call instead of two - `__setitem__` is just a function that could be redefined in python in a way that lets it drop the gil while invoking it...

Practically, we hope that it continued to be kind to us.

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
niloc132: Call `ScriptSessionQueryScope.putParam("foo", somePyObjectInstance)`, and then call `same = readParamValue("foo")` to get it back - the PyObject instances are pointing at the "same" python object, but `same == somePyObjectInstance` is false - they are separate java object instances, and contain separate borrowed pointers into py/c.

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -243,85 +251,74 @@ protected void destroy() {
...
         @Override
         public Set<String> getParamNames() {
-            final Set<String> result = new LinkedHashSet<>();
-            for (final String name : scriptSession.getVariableNames()) {
-                if (NameValidator.isValidQueryParameterName(name)) {
-                    result.add(name);
-                }
-            }
-            return result;
+            return AbstractScriptSession.this.getVariableNames(NameValidator::isValidQueryParameterName);
```
rcaudy, line 307:
```suggestion
            return getVariableNames(NameValidator::isValidQueryParameterName);
```

engine/context/src/main/java/io/deephaven/engine/context/StandaloneQueryScope.java
```diff
@@ -0,0 +1,153 @@
...
+        // noinspection unchecked
+        return (T) valueRetriever.getValue();
+    }
+
+    @Override
+    public <T> void putParam(final String name, final T value) {
+        if (value instanceof LivenessReferent && DynamicNode.notDynamicOrIsRefreshing(value)) {
+            manage((LivenessReferent) value);
+        }
+        NameValidator.validateQueryParameterName(name);
```
rcaudy, line 111: Should validate before managing...
```suggestion
        NameValidator.validateQueryParameterName(name);
        if (value instanceof LivenessReferent && DynamicNode.notDynamicOrIsRefreshing(value)) {
            manage((LivenessReferent) value);
        }
```

engine/table/src/main/java/io/deephaven/engine/util/AbstractScriptSession.java
```diff
@@ -330,57 +327,43 @@ public <T> T readParamValue(final String name) throws QueryScope.MissingVariable
...
+            if (value instanceof LivenessReferent && DynamicNode.notDynamicOrIsRefreshing(value)) {
+                manage((LivenessReferent) value);
+            }
 
-        @Override
-        protected synchronized <T> QueryScopeParam<T> createParam(final String name)
-                throws QueryScope.MissingVariableException {
-            return super.createParam(name);
-        }
+            Object oldValue = AbstractScriptSession.this.setVariable(name, value);
```
rcaudy, line 354:
```suggestion
            Object oldValue = setVariable(name, value);
```

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -295,34 +273,48 @@ private Object maybeUnwrap(PyObject o) {
     }
 
     @Override
-    public Set<String> getVariableNames() {
-        return Collections.unmodifiableSet(scope.getKeys().collect(Collectors.toSet()));
+    protected Set<String> getVariableNames(Predicate<String> allowName) {
```
rcaudy, line 276: `maybeUnwrap(Object)` a few lines above this is apparently unused and functionally the same as `unwrapObject`... should it be removed?
Likewise, I'm convinced that `maybeUnwrap(PyObject)` could just be replaced with `unwrapObject` calls.

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -295,34 +273,48 @@ private Object maybeUnwrap(PyObject o) {
...
+                if (newValue instanceof PyObject) {
+                    wrapped = newValue;
+                } else {
+                    wrapped = PythonObjectWrapper.wrap(newValue);
+                }
+                // This isn't thread safe, we're relying on the GIL being kind to us (as we have historically done).
+                // There is no built-in for "replace a variable and return the old one".
+                Object prev = globals.get(name);
+                globals.setItem(name, wrapped);
+                return prev;
```
rcaudy, line 307: Should we be unwrapping `prev`?

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
rcaudy: What's the solution? Is @jmao-denver aware?

engine/sql/src/main/java/io/deephaven/engine/sql/Sql.java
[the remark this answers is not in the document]
rcaudy: No, to make the `readParamValue` results be consistent with `getParamNames`. Otherwise, you could have a `MissingVariableException` due to concurrent mutation of the `QueryScope`.
Feels like a missing method to get values, to be honest.

open-api/lang-tools/src/main/java/io/deephaven/lang/completion/ChunkerCompleter.java
rcaudy: I would want to think long and hard before pushing `BaseTable` or `QueryTable` methods into auto-complete, to be honest. Let's not change the status quo in this PR.

engine/context/src/main/java/io/deephaven/engine/context/QueryScope.java
```diff
@@ -3,25 +3,19 @@
...
+import java.util.Set;
 
 /**
  * Variable scope used to resolve parameter values during query execution.
  */
-public abstract class QueryScope implements LogOutputAppendable {
+public interface QueryScope extends LivenessNode, LogOutputAppendable {
 
     /**
      * Adds a parameter to the default instance {@link QueryScope}, or updates the value of an existing parameter.
```
rcaudy, line 21: This should document the liveness impacts of `putParam`, too.

server/src/main/java/io/deephaven/server/console/ScopeTicketResolver.java
```diff
@@ -58,29 +51,29 @@ public SessionState.ExportObject<Flight.FlightInfo> flightInfoFor(
...
-                    }
+        QueryScope queryScope = ExecutionContext.getContext().getQueryScope();
+        Object scopeVar = queryScope.readParamValue(scopeName, null);
+        if (scopeVar == null) {
+            throw Exceptions.statusRuntimeException(Code.NOT_FOUND,
+                    "Could not resolve '" + logId + ": no variable exists with name '" + scopeName + "'");
+        }
+        if (!(scopeVar instanceof Table)) {
+            throw Exceptions.statusRuntimeException(Code.NOT_FOUND,
+                    "Could not resolve '" + logId + "': no variable exists with name '" + scopeName + "'");
```
rcaudy, line 62: This is a lie.

server/src/main/java/io/deephaven/server/console/ScopeTicketResolver.java
```diff
@@ -58,29 +51,29 @@ public SessionState.ExportObject<Flight.FlightInfo> flightInfoFor(
...
 
         return SessionState.wrapAsExport(flightInfo);
     }
 
     @Override
     public void forAllFlightInfo(@Nullable final SessionState session, final Consumer<Flight.FlightInfo> visitor) {
-        scriptSessionProvider.get().getVariables().forEach((varName, varObj) -> {
+        QueryScope queryScope = ExecutionContext.getContext().getQueryScope();
+        queryScope.getParamNames().forEach((varName) -> {
+            Object varObj = queryScope.readParamValue(varName);
```
rcaudy, line 76: Same atomicity issue as in `Sql`.

rcaudy (commented):
Mostly happy. A bit concerned about atomicity between listing names and iterating values. A few other nits or questions.

server/src/main/java/io/deephaven/server/console/ScopeTicketResolver.java
niloc132: True, but not a new issue. Do we want a method that returns a map<string,value>?

engine/context/src/main/java/io/deephaven/engine/context/QueryScope.java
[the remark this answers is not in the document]
niloc132: Raffi says kill it.

engine/context/src/main/java/io/deephaven/engine/context/StandaloneQueryScope.java
```diff
@@ -0,0 +1,153 @@
...
+    private final KeyedObjectHashMap<String, ValueRetriever<?>> valueRetrievers =
+            new KeyedObjectHashMap<>(new ValueRetrieverNameKey());
+
+    /**
+     * Apply conversions to certain scope variable values.
+     *
+     * @param value value
+     * @return value, or an appropriately converted substitute.
+     */
+    private static Object applyValueConversions(final Object value) {
```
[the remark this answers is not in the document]
niloc132: Raffi says kill it.

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
niloc132: No, we require callers to invoke unwrap if they want it unwrapped, we don't do it for them - this is consistent with other calls.

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
niloc132: There isn't a great solution in general, except to cleanly close PyObject basically everywhere - java gc is required to mop up all the unclosed PyObject. This is wrong in enough places in DHC that it isn't worth dedicating effort to it here.

server/src/main/java/io/deephaven/server/console/ScopeTicketResolver.java
rcaudy: What do you mean by "doesn't seem constructive"? It seems like pretty valuable debugging information to me.

engine/context/src/main/java/io/deephaven/engine/context/QueryScope.java
```diff
@@ -138,8 +131,16 @@ default QueryScopeParam[] getParams(final Collection<String> names) throws Missi
...
 
     /**
-     * Asks the session to remove any wrapping that exists on scoped objects so that clients can fetch them. Defaults to
-     * returning the object itself.
+     * Returns an immutable map with all objects in the scope. Callers may want to unwrap language-specific values using
+     * {@link #unwrapObject(Object)} before using them.
+     *
+     * @return an immutable map with all known variables and their values.
+     */
+    Map<String, Object> readAllValues();
```
rcaudy, line 139: `toMap()`? `readAllEntries()`?

Integrations/src/main/java/io/deephaven/integrations/python/PythonDeephavenSession.java
```diff
@@ -317,6 +300,12 @@ protected synchronized Object setVariable(String name, @Nullable Object newValue
         return old;
     }
 
+    @Override
+    protected Map<String, Object> getAllValues() {
+        return PyLib
+                .ensureGil(() -> scope.getEntries().collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)));
```
rcaudy, line 306: Not immutable

engine/context/src/main/java/io/deephaven/engine/context/StandaloneQueryScope.java
```diff
@@ -121,6 +76,12 @@ public <T> void putParam(final String name, final T value) {
         }
     }
 
+    @Override
+    public Map<String, Object> readAllValues() {
+        return valueRetrievers.entrySet().stream()
```
rcaudy, line 81: Not immutable

server/src/main/java/io/deephaven/server/console/ScopeTicketResolver.java
rcaudy: It looks like you implemented that method, but didn't apply it to these comments.

rcaudy (commented):
I'm happy enough, but for the missing unwraps in `Sql` and `ScopeTickerResolver`.

rcaudy (approved):
Please merge if nightlies pass.

PR #4985, opened 2023-12-22
Flight: Disable Batching Static Tables

nbauernfeind:
The jsapi cannot handle multiple batches on static tables; note that refreshing tables are already disabled in this regard. 

We'll have to disable batching static tables until after #188 is completed.


niloc132 (approved):
Server changes (and test) approved, tested locally with and without the client additions.

Issue #4975, opened 2023-12-21
UpdateGraph throws `java.lang.IllegalStateException: No outstanding notifications, yet the notification queue is not empty!`


nbauernfeind:
A community member is getting an unexpected shutdown due to the following thrown exception:
```
Server shutdown: Uncaught exception in thread PeriodicUpdateGraph.DEFAULT.refreshThread

java.lang.IllegalStateException: No outstanding notifications, yet the notification queue is not empty!
io.deephaven.engine.updategraph.impl.BaseUpdateGraph.flushNormalNotificationsAndCompleteCycle(BaseUpdateGraph.java:577)
io.deephaven.engine.updategraph.impl.BaseUpdateGraph.flushNotificationsAndCompleteCycle(BaseUpdateGraph.java:518)
io.deephaven.engine.updategraph.impl.BaseUpdateGraph.lambda$doRefresh$5(BaseUpdateGraph.java:915)
io.deephaven.util.locks.FunctionalLock.doLocked(FunctionalLock.java:33)
io.deephaven.engine.updategraph.impl.BaseUpdateGraph.doRefresh(BaseUpdateGraph.java:900)
io.deephaven.engine.updategraph.impl.BaseUpdateGraph.refreshAllTables(BaseUpdateGraph.java:887)
io.deephaven.engine.updategraph.impl.PeriodicUpdateGraph.refreshAllTables(PeriodicUpdateGraph.java:1090)
io.deephaven.engine.updategraph.impl.BaseUpdateGraph.refreshTablesAndFlushNotifications(BaseUpdateGraph.java:779)
io.deephaven.engine.updategraph.impl.PeriodicUpdateGraph.refreshTablesAndFlushNotifications(PeriodicUpdateGraph.java:1014)
io.deephaven.engine.updategraph.impl.PeriodicUpdateGraph.lambda$new$0(PeriodicUpdateGraph.java:139)
java.base/java.lang.Thread.run(Thread.java:1583)
```

Working with them to acquire more information of the query causing the issue.

Found in Deephaven `0.31.0`

nbauernfeind:
Closing as this was a combination of several issues. Several already are already closed/merged (#4980, #4976, and #4986) with the only outstanding piece being #4984 (but was unrelated and found during investigation and is otherwise being worked on).

PR #4986, opened 2023-12-23
Move OperationInitializer Construction to Dagger Object Graph

nbauernfeind:
This sets the proper operation initializer for the refresh thread should the UpdateGraph be started in single threaded mode. This was detected in #4975 and is a part of that fix.

### DIFF
--- a/py/server/test_helper/__init__.py
+++ b/py/server/test_helper/__init__.py
@@ -69,11 +69,18 @@ def start_jvm_for_tests(jvm_props: Dict[str, str] = None):
         # Set up a Deephaven Python session
         py_scope_jpy = jpy.get_type("io.deephaven.engine.util.PythonScopeJpyImpl").ofMainGlobals()
         global py_dh_session
+
+        no_op_thread_factory = jpy.get_type("io.deephaven.util.thread.ThreadInitializationFactory").NO_OP
+        _JOperationInitializationThreadPool = jpy.get_type("io.deephaven.engine.table.impl.OperationInitializationThreadPool")
+        _j_operation_initializer = _JOperationInitializationThreadPool(no_op_thread_factory)
+
         _JPeriodicUpdateGraph = jpy.get_type("io.deephaven.engine.updategraph.impl.PeriodicUpdateGraph")
-        _j_test_update_graph = _JPeriodicUpdateGraph.newBuilder(_JPeriodicUpdateGraph.DEFAULT_UPDATE_GRAPH_NAME).existingOrBuild()
-        no_op_operation_initializer = jpy.get_type("io.deephaven.util.thread.ThreadInitializationFactory").NO_OP
+        _j_test_update_graph = _JPeriodicUpdateGraph.newBuilder(_JPeriodicUpdateGraph.DEFAULT_UPDATE_GRAPH_NAME) \
+                .operationInitializer(_j_operation_initializer) \
+                .existingOrBuild()
+
         _JPythonScriptSession = jpy.get_type("io.deephaven.integrations.python.PythonDeephavenSession")
-        py_dh_session = _JPythonScriptSession(_j_test_update_graph, no_op_operation_initializer, py_scope_jpy)
+        py_dh_session = _JPythonScriptSession(_j_test_update_graph, _j_operation_initializer, no_op_thread_factory, py_scope_jpy)
 
 
 def _expand_wildcards_in_list(elements):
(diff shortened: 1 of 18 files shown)


niloc132 (approved):
I love this, thank you for this massive improvement on what I wrote. The parts I don't like... are my problem anyway, for a followup patch to make ScriptSession creation nicer anyway.

py/server/test_helper/__init__.py
```diff
@@ -69,11 +69,18 @@ def start_jvm_for_tests(jvm_props: Dict[str, str] = None):
         # Set up a Deephaven Python session
         py_scope_jpy = jpy.get_type("io.deephaven.engine.util.PythonScopeJpyImpl").ofMainGlobals()
         global py_dh_session
+
+        no_op_thread_factory = jpy.get_type("io.deephaven.util.thread.ThreadInitializationFactory").NO_OP
```
rcaudy, line 73: I think this should follow the _ convention.

server/src/main/java/io/deephaven/server/runner/updategraph/UpdateGraphModule.java
```diff
@@ -16,9 +19,20 @@ public class UpdateGraphModule {
     @Provides
     @Singleton
     @Named(PeriodicUpdateGraph.DEFAULT_UPDATE_GRAPH_NAME)
-    public static UpdateGraph provideUpdateGraph() {
+    public static UpdateGraph provideUpdateGraph(
+            final ThreadInitializationFactory threadInitializationFactory,
+            final OperationInitializer operationInitializer) {
         return PeriodicUpdateGraph.newBuilder(PeriodicUpdateGraph.DEFAULT_UPDATE_GRAPH_NAME)
                 .numUpdateThreads(PeriodicUpdateGraph.NUM_THREADS_DEFAULT_UPDATE_GRAPH)
+                .threadInitializationFactory(threadInitializationFactory)
```
rcaudy, line 27: @niloc132 Was Python debugging broken without this?

sphinx/source/conf.py
```diff
@@ -105,13 +105,21 @@
...
 import jpy
 py_scope_jpy = jpy.get_type("io.deephaven.engine.util.PythonScopeJpyImpl").ofMainGlobals()
+
+no_op_thread_factory = jpy.get_type("io.deephaven.util.thread.ThreadInitializationFactory").NO_OP
+_JOperationInitializationThreadPool = jpy.get_type("io.deephaven.engine.table.impl.OperationInitializationThreadPool")
+_j_operation_initializer = _JOperationInitializationThreadPool(no_op_thread_factory)
+
 _JUpdateGraph = jpy.get_type("io.deephaven.engine.updategraph.impl.PeriodicUpdateGraph")
-docs_update_graph = _JUpdateGraph.newBuilder("PYTHON_DOCS").build()
+docs_update_graph = _JUpdateGraph.newBuilder("PYTHON_DOCS") \
```
rcaudy, line 114: Same comment as on the test helper. Maybe consult with @jmao-denver to make sure we have the right conventions.

rcaudy (commented):
Minor comments. Please ensure that nightlies are passing.

server/src/main/java/io/deephaven/server/runner/updategraph/UpdateGraphModule.java
niloc132: Likely, yes, but we don't have tests for this.

py/server/test_helper/__init__.py
niloc132: underscore is for top-level stuff that shouldn't be imported from other modules, not necessary here

sphinx/source/conf.py
niloc132: these arguably could be `_` prefixed... but this is configuration for running sphinx, not something that users can use (or that we would import anywhere else), so shouldn't matter.

py/server/test_helper/__init__.py
jmao-denver: underscore doesn't apply to auto-variables in the function scope, so the naming is fine here, though I prefer to use the `_j_ ` prefix if an object/variable comes from Java..

sphinx/source/conf.py
jmao-denver: +1, and the same comment above applies here too.

jmao-denver (approved):
LGTM

rcaudy (approved):
Approving based on Jianfeng's feedback. Ready to merge if nightlies pass.

nbauernfeind:
Nightlies passed: https://github.com/nbauernfeind/deephaven-core/actions/runs/7455524151

PR #4988, opened 2023-12-27
Run JS API unit tests and integration tests in the Gradle build

niloc132:
Restores some unit and integration tests to run as part of the local/CI builds, as part of the `check` task.

Partial #188


web/client-api/src/main/java/io/deephaven/web/client/api/JsColumnStatistics.java
```diff
@@ -12,6 +12,7 @@
 import jsinterop.annotations.JsIgnore;
 import jsinterop.annotations.JsMethod;
 import jsinterop.annotations.JsProperty;
+import jsinterop.annotations.JsType;
```
niloc132, line 15: revert

web/client-api/src/main/java/io/deephaven/web/client/api/WorkerConnection.java
```diff
@@ -1437,7 +1437,7 @@ private void flush() {
                 // TODO #188 support minUpdateIntervalMs
                 double serializationOptionsOffset = BarrageSubscriptionOptions
                         .createBarrageSubscriptionOptions(subscriptionReq, ColumnConversionMode.Stringify, true, 1000,
-                                0, 0);
+                                0, 1_000_000_000);
```
niloc132, line 1440: revert - this was for confirming a different bug

web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
```diff
@@ -1762,8 +1770,8 @@ public void handleSnapshot(TableTicket handle, TableSnapshot snapshot) {
                 viewportRows.size());
     }
 
-
-    protected void processSnapshot() {
+    @JsIgnore
+    public void processSnapshot() {
```
niloc132, line 1774: verify this is necessary, probably revert

web/client-api/src/main/java/io/deephaven/web/client/api/subscription/SubscriptionTableData.java
```diff
@@ -43,7 +44,7 @@ private interface ArrayCopy {
     private RangeSet index;
 
     // mappings from the index to the position of a row in the data array
-    private TreeMap<Long, Long> redirectedIndexes;
+    private TreeMap<Long, Integer> redirectedIndexes;
```
niloc132, line 47: revert this whole file, this is a hack to test a perf issue with un-downsampled charts

web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
niloc132: This is necessary, holdover from viewport debouncing. Post-188, we can probably remove this once and for all.

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
```diff
@@ -0,0 +1,391 @@
+package io.deephaven.web.client.api;
+
+import com.google.gwt.core.client.JavaScriptException;
+import com.google.gwt.junit.client.GWTTestCase;
+import elemental2.core.JsArray;
+import elemental2.core.JsError;
+import elemental2.core.JsString;
+import elemental2.dom.CustomEvent;
+import elemental2.dom.CustomEventInit;
```
mofojed, line 9: Unused

```suggestion
```

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
```diff
@@ -0,0 +1,391 @@
...
+import elemental2.dom.DomGlobal;
+import elemental2.promise.IThenable;
+import elemental2.promise.Promise;
+import io.deephaven.web.client.api.subscription.ViewportData;
+import io.deephaven.web.client.fu.CancellablePromise;
+import io.deephaven.web.client.ide.IdeSession;
+import io.deephaven.web.shared.data.Viewport;
```
mofojed, line 16: Unused
```suggestion
```

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
```diff
@@ -0,0 +1,391 @@
...
+import io.deephaven.web.shared.fu.JsRunnable;
+import io.deephaven.web.shared.fu.RemoverFn;
+import jsinterop.annotations.JsMethod;
+import jsinterop.annotations.JsPackage;
+import jsinterop.annotations.JsProperty;
+import jsinterop.base.Js;
+import jsinterop.base.JsPropertyMap;
+import org.apache.tapestry.INamespace;
```
mofojed, line 24: Unused
```suggestion
```

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
```diff
@@ -0,0 +1,391 @@
...
+            pythonScripts.add(tableName + "=" + python);
+
+            return this;
+        }
+    }
+
+    /**
+     * Set this to a value higher than 1 to get more time to run debugger without timeouts failing.
+     */
+    protected static final int TIMEOUT_SCALE = 2;
```
mofojed, line 72: Feel like this should default to `1`, particularly with the comment before it. Should it default to a different value if we're in a dev mode vs. CI (unsure which flag that is)? Or some other env var?

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
```diff
@@ -0,0 +1,391 @@
...
+        return Promise.reject(error);
+    }
+
+    protected Promise<?> finish(Object input) {
+        finishTest();
+        return Promise.resolve(input);
+    }
+
+    /**
+     * Helper method to add a listener to the promise of a table, and ensure that an update is recieved with the
```
mofojed, line 213:
```suggestion
     * Helper method to add a listener to the promise of a table, and ensure that an update is received with the
```

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
```diff
@@ -0,0 +1,391 @@
...
+     * expected number of items, within the specified timeout.
+     *
+     * Prereq: have already requested a viewport on that table
+     */
+    protected Promise<JsTable> assertUpdateReceived(Promise<JsTable> tablePromise, int count, int timeoutInMillis) {
+        return tablePromise.then(table -> assertUpdateReceived(table, count, timeoutInMillis));
+    }
+
+    /**
+     * Helper method to add a listener to a table, and ensure that an update is recieved with the expected number of
```
mofojed, line 223:
```suggestion
     * Helper method to add a listener to a table, and ensure that an update is received with the expected number of
```

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
```diff
@@ -0,0 +1,391 @@
...
+        }, checkInterval);
+    }
+
+    protected Promise<JsTable> assertNextViewportIs(JsTable table, Function<JsTable, Column> column,
+            String[] expected) {
+        return assertUpdateReceived(table, viewportData -> {
+            String[] actual = Js.uncheckedCast(getColumnData(viewportData, column.apply(table)));
+            assertTrue("Expected " + Arrays.toString(expected) + ", found " + Arrays.toString(actual) + " in table "
+                    + table + " at state " + table.state(), Arrays.equals(expected, actual));
+        }, 2000);
```
mofojed, line 366: Define as constant - same timeout is used in the other `assertNextViewportIs` method.

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
```diff
@@ -0,0 +1,391 @@
...
+        });
+    }
+
+    private Promise<IdeSession> runAllScriptsInOrder(CancellablePromise<IdeSession> ideSession, IdeSession session,
+            List<String> code) {
+        Promise<IdeSession> result = ideSession;
+        for (int i = 0; i < code.size(); i++) {
+            final int index = i;
+            result = result.then(ignore -> {
+                delayTestFinish(4000 + index);
```
mofojed, line 172: Constant? Unsure why 4000 was selected.

mofojed (changes requested):
We should include instructions on how to run these tests as well. I ran `./gradlew gwtUnitTest` and that worked correctly. Then I tried running `./gradle gwtIntegrationTest` and it's just giving me a bunch of `UnableToCompleteException`s for each test:

```
> Task :web-client-api:gwtIntegrationTest FAILED
1 compiler directives added

io.deephaven.web.ClientIntegrationTestSuite > io.deephaven.web.client.api.subscription.ViewportTestGwt.testViewportOnUpdatingTable FAILED
    com.google.gwt.junit.JUnitFatalLaunchException at JUnitShell.java:708

io.deephaven.web.ClientIntegrationTestSuite > io.deephaven.web.client.api.subscription.ViewportTestGwt.testViewportSubsetOfColumns FAILED
    com.google.gwt.core.ext.UnableToCompleteException at JUnitShell.java:1334

io.deephaven.web.ClientIntegrationTestSuite > io.deephaven.web.client.api.subscription.ViewportTestGwt.testViewportWithNoInitialItems FAILED
    com.google.gwt.core.ext.UnableToCompleteException at JUnitShell.java:1334
```

web/client-api/client-api.gradle
```diff
@@ -57,6 +68,89 @@ artifacts {
...
+            ].join(' '),
+            'gwt.persistentunitcachedir': layout.buildDirectory.dir('unitTest-unitCache').get().asFile.absolutePath,
+    ]
+    t.include '**/ClientUnitTestSuite.class'
+    t.useJUnit()
+    t.scanForTestClasses = false
+}
+
+// start a grpc-api server
+String randomSuffix = UUID.randomUUID().toString();
```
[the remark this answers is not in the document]
niloc132: It would yes, but we are doing this all over for naming docker containers, so that you don't accidentally collide in parallel runs, or separate checkouts, etc. Since this is just for tests, we're not concerned with "oh these tests passed last time, so I'm not even going to run them this time" for integration tests.

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
niloc132: Yep, I did intend to revert this... and tweak all the tests to have massively larger timeouts. Will clean that up.

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
niloc132: Actually I want them to be different-ish constants, so you can nail down which timeout failed, since it can be a pain to find them.

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
niloc132: It seems that running python on a brand new server in docker can sometimes be a little slow out of the gate, so I have a "large" timeout here. 

It could make sense to change the strategy entirely and just have a per-test timeout instead, and set it to a few minutes - but it is nice to fail within a few seconds of the failing operation (as we don't have event handlers for all failures, we just let timeouts expire).

web/client-api/client-api.gradle
```diff
@@ -57,6 +68,89 @@ artifacts {
...
+def createSelenium = tasks.register('createSelenium', DockerCreateContainer) { t ->
+    t.dependsOn(Docker.registryTask(project, 'selenium'))
+    t.targetImageId('deephaven/selenium:local-build')
+    t.containerName.set(seleniumContainerId)
+    t.hostConfig.shmSize.set(2L * 1024 * 1024 * 1024)
+
+    // Use "host" networking, so that gradle's test runner can be accessed by the browser. This means
+    // that port 4444 will always be bound, and if already used, this container won't be able to start.
+    t.hostConfig.network.set('host')
+    t.hostConfig.portBindings.set(['4444'])
```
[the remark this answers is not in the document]
niloc132: How would you feel about a gradle property instead, such as `./gradlew -PseleniumPort=4567 :web-client-api:check`?

web/client-api/client-api.gradle
```diff
@@ -57,6 +68,89 @@ artifacts {
...
+def startSelenium = tasks.register('startSelenium', DockerStartContainer) {t ->
+    t.dependsOn(createSelenium)
+    t.containerId.set(seleniumContainerId)
+}
+def seleniumHealthy = project.tasks.register('seleniumHealthy', WaitForHealthyContainer) { task ->
+    task.dependsOn startSelenium
 
-apply from: "$rootDir/gradle/web-gwt-test.gradle"
+    task.awaitStatusTimeout.set 120
+    task.checkInterval.set 100
```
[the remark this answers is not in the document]
niloc132: Agreed... can go back and rewrite those, but I vote not today. This is code taken from the upstream docker-gradle plugin, and we didn't customize it further.

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
mofojed: Yea that's all fine and good, just as a constant you can have those details as a comment so it's clear why you want a timeout of a 4s instead of say 40s or 400ms

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
mofojed: Yup that's fair too, though would be nice to define a base constant and just add an offset for uniqueness

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
niloc132: All of these timeouts are "worst case, sometimes the browser or dh api lags a bit, here's a reasonably sized timeout that likely can't be exceeded in reasonable circumstances". 40ms definitely isn't good enough (browser GC could kill that alone), 400ms probably isn't enough either. 

We _could_ just say "every test should take less than 5 minutes" and not worry about each operation, but historically this has made it easier to figure out which specific operation is stuck/failing.

web/client-api/client-api.gradle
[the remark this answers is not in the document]
mofojed: Yea that would be fine

web/client-api/src/test/java/io/deephaven/web/client/api/AbstractAsyncGwtTestCase.java
mofojed: Right - that all makes sense, now add that "why" information as a comment on that timeout.

mofojed (dismissed):
Works on Mac

web/client-api/src/test/java/io/deephaven/web/client/api/NullValueTestGwt.java
```diff
@@ -0,0 +1,81 @@
...
+            .script("nulltable", "empty_table(2).update([\n" +
+                    "   \"MyInt=i==0?null:i\",\n" +
+                    "   \"MyLong=i==0?null:(long)i\",\n" +
+                    "   \"MyDouble=i==0?null:(double)i\",\n" +
+                    "   \"MyShort=i==0?null:(short)i\",\n" +
+                    "   \"MyFloat=i==0?null:(float)i\",\n" +
+                    "   \"MyChar=i==0?null:(char)i\",\n" +
+                    "   \"MyByte=i==0?null:(byte)i\",\n" +
+                    "   \"MyBoolean=i==0?null:true\",\n" +
+                    "   \"MyDate=i==0?null:epochNanosToInstant(i)\"\n" +
```
mofojed, line 22: Missing String, BigDecimal, BigInteger

devinrsmith (commented):
Talking about selenium port and direct docker IP offline.

mofojed (dismissed):

docker/registry/selenium/gradle.properties
```diff
@@ -0,0 +1,3 @@
+io.deephaven.project.ProjectType=DOCKER_REGISTRY
+deephaven.registry.imageName=selenium/standalone-firefox:4.16.1-20231219
```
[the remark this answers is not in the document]
niloc132: Selenium client has historically been firmly tied to a single release of the browser+driver version, so it rarely will work to update this to "latest" without finding out what version that happens to be, and then update the test dependencies to match that specific version. This way, the version number is clear in both places.

devinrsmith (approved):

mofojed (approved):

Issue #4991, opened 2023-12-28
Reorganize top-level project


devinrsmith:
Our top-level root has grown to about 100 files and folders, which makes it a task to scroll down past all of that on https://github.com/deephaven/deephaven-core to get to the README contents. Ideally, the README header should be visible without having to scroll.

rcaudy:
What's the proposed solution?

devinrsmith:
Create one or more top-level folders and start moving existing subproject folders into it. I don't know what the best hierarchy or naming strategy is.

devinrsmith:
Also, the migration could be piecemeal; doesn't need to be all-at-once.

devinrsmith:
We could consider moving the top-level docs (except README.md) into a `docs/` folder.

PR #4992, opened 2023-12-28
Refactor TypedFunction, ObjectProcessor, and ObjectProcessorFunctions

devinrsmith:
* `io.deephaven.functions` renamed to `io.deephaven.function`, moved to `function/`, `function` project
* `io.deephaven.processor` moved to `engine/processor/`, `engine-processor` project
* `io.deephaven.processor.functions` moved to `engine/processor-functions/`, `engine-processor-functions` project
    
Would like eventual follow-up for `function/` to more appropriate place, ala #4991


java/processor-functions/src/main/java/io/deephaven/processor/functions/ChunkUtils.java
```diff
@@ -0,0 +1,138 @@
...
+
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.ToDoubleFunction;
+import java.util.function.ToIntFunction;
+import java.util.function.ToLongFunction;
+
+final class ChunkUtils {
+
+    // Copy from io.deephaven.kafka.ingest.ChunkUtils
```
rcaudy, line 29: Why are we copying this? Seems like we could just move it somewhere better.

settings.gradle
```diff
@@ -377,6 +377,15 @@ include ':clock-impl'
 
 include ':sql'
 
+include ':functions'
+project(':functions').projectDir = file('java/functions')
```
rcaudy, line 381: I'm not a fan of this naming, specifically the "java/" prefix.

devinrsmith: Hoping to have a generic name where we can dump a lot of projects into. Open to suggestions.

java/processor-functions/src/main/java/io/deephaven/processor/functions/ChunkUtils.java
devinrsmith: It's possible that these should move into code generation of Chunks themselves, but trying to grok that code generation logic and deal with generic typing is a pita and a lot more work.

I'm hoping the kafka ChunkUtils eventually gets deleted, and I don't know how much value there is to making this publicly accessible otherwise until necessary.

devinrsmith:
This is now a breaking change w/ the rename from `io.deephaven.functions` -> `io.deephaven.function`.

devinrsmith:
Planned follow-up wrt changes to SafeCloseable / moving of  `io.deephaven.engine.primitive`

java/processor-functions/src/main/java/io/deephaven/processor/functions/ChunkUtils.java
devinrsmith: I was able to remove the duplication and remove a lot of code by creating `FieldCopierProcessorImpl`

settings.gradle
```diff
@@ -377,6 +383,9 @@ include ':clock-impl'
 
 include ':sql'
 
+include ':function'
+project(':function').projectDir = file('function')
```
rcaudy, line 387: I thought we said we would put this in `Util` and be `:util-function`?

settings.gradle
```diff
@@ -212,6 +212,12 @@ project(':engine-query-constants').projectDir = file('engine/query-constants')
 include(':engine-sql')
 project(':engine-sql').projectDir = file('engine/sql')
 
+include ':engine-processor'
+project(':engine-processor').projectDir = file('engine/processor')
+
+include ':engine-processor-functions'
+project(':engine-processor-functions').projectDir = file('engine/processor-functions')
```
rcaudy, line 219: Why do we mix `/` and `-`? Should it be in `engine/process/functions`?
Plural or singular function?

function/build.gradle
```diff
@@ -0,0 +1,19 @@
+plugins {
+    id 'java-library'
+    id 'io.deephaven.project.register'
+}
+
+dependencies {
+    api project(':qst')
```
rcaudy, line 7: Do we really need to depend on QST here? We can't be "pure Java"?

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/FieldCopierAdapter.java
```diff
@@ -3,117 +3,25 @@
...
-import io.deephaven.functions.ToLongFunction;
-import io.deephaven.functions.ToObjectFunction;
-import io.deephaven.functions.ToPrimitiveFunction;
-import io.deephaven.functions.ToShortFunction;
-import io.deephaven.functions.TypedFunction;
-
-public enum FieldCopierAdapter
-        implements TypedFunction.Visitor<Object, FieldCopier>, ToPrimitiveFunction.Visitor<Object, FieldCopier> {
-    INSTANCE;
+import io.deephaven.function.ToBooleanFunction;
```
rcaudy, line 8: I bet a lot of these are unused now. Do we even need the class?

function/build.gradle
devinrsmith: I'll merge and fix after https://github.com/deephaven/deephaven-core/pull/5054 lands

engine/processor/src/main/java/io/deephaven/processor/ObjectProcessor.java
```diff
@@ -129,7 +129,7 @@ static <T> ObjectProcessor<T> noop(List<Type<?>> outputTypes, boolean fillWithNu
      * </tr>
      * <tr>
      * <td>{@link InstantType}</td>
-     * <td>{@link ChunkType#Long} ({@link io.deephaven.time.DateTimeUtils#epochNanos(Instant)})</td>
+     * <td>{@link ChunkType#Long} (io.deephaven.time.DateTimeUtils#epochNanos(Instant))</td>
```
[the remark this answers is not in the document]
devinrsmith: Actually, this is a good call-out. `engine-processor` shouldn't even need impl dependency on engine-time anymore. I'm apt to not re-add it as api just for the javadoc.

extensions/kafka/src/main/java/io/deephaven/kafka/ingest/FieldCopierAdapter.java
devinrsmith: Removed these unused imports. FieldCopierAdapter is a "bridge" between the world of TypedFunction and the existing kafka implementation; I have hopes we can deprecate existing kafka impl soon and move to one that is ObjectProcessor-first.

settings.gradle
devinrsmith: It's not without precedent; `extensions/source-support`, `engine/test-utils`, `engine/query-constants`.

In my mind, `processor-functions` isn't really hierarchical, it's `engine-processor` + `util-function` at the same level:

```gradle
    api project(':engine-processor')
    api project(':util-function')
```

rcaudy (dismissed):

rcaudy (approved):

PR #4999, opened 2023-12-29
JS API should expose if a table is refreshing

### LINKED ISSUE #4485: Expose `isRefreshing` property through JS API
As a consumer of the JS API, I want to know if a table is refreshing so that I can make assumptions about the table when I display it, such as taking a snapshot of the data.

Should be exposed in `JsTable.isRefreshing` and `JsTreeTable.isRefreshing`

niloc132:
Fixes #4485


niloc132:
If we land this after https://github.com/deephaven/deephaven-core/pull/4988, I'll add integration tests...

web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java
```diff
@@ -1809,6 +1809,17 @@ public boolean isClosed() {
...
 
+    /**
+     * True if this table may receive updates from the server, including size changed events, updated events after
+     * initial snapshot.
+     *
+     * @return boolean
+     */
+    @JsProperty(name = "isRefreshing")
+    public boolean isRefreshing() {
+        return state().isStatic();
```
rcaudy, line 1820: Seems backwards

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/contextual-auth.proto
devinrsmith, line 1: Is this related to this PR?

niloc132: No, should be removed, I'll put it back in proposal for #3133.

niloc132:
> Looks good, though now I'm wondering why I specified JsTable.isRefreshing as opposed to something like JsTable.isStatic. Would isStatic be a better name? Could see isRefreshing being misconstrued as something else. Thoughts?

It looks like the server api exposes this as isRefreshing(), so it makes sense to me for the js client api to do the same - that's probably where you got the name?

jmao-denver (approved):

devinrsmith (approved):

PR #5004, opened 2024-01-03
fix: Fix LazyPromise behaviour

mofojed:
- Ported over from Enterprise change, and then added some more fixes on top:
  1. Errors were thrown after a promise was already fulfilled, instead of doing a no-op
  2. Success value or failure reason value were getting overwritten if the success/fail method was called more than once (again, after a promise is fulfilled it should be a no-op)
  3. The order that chained promises were called in was reversed and non-deterministic
- Remove unnecessary `cancelled` variable - it's just a different kind of failure
- Added unit tests


mofojed:
Needs tests.

web/client-api/client-api.gradle
```diff
@@ -135,6 +121,28 @@ def stopSelenium = project.tasks.register('stopSelenium', DockerRemoveContainer)
     task.removeVolumes.set true
 }
 
+def gwtUnitTest = tasks.register('gwtUnitTest', Test) { t ->
+    t.dependsOn(deephavenDocker.portTask, seleniumHealthy)
+    t.finalizedBy(deephavenDocker.endTask, stopSelenium)
```
niloc132, line 126: (note that you also have `t.finalizedBy(dhDocker.endTask)` below)

if you make this change, there is no longer a reason to have unit vs integration tests. instead, i would suggest removing the deephavenDocker dependencies, so that this only depends on selenium - it will make the unit tests somewhat faster (probably approx 2x) to run after any changes to the js api.

web/client-api/client-api.gradle
```diff
@@ -135,6 +121,28 @@ def stopSelenium = project.tasks.register('stopSelenium', DockerRemoveContainer)
...
+    doFirst {
+        def webdriverUrl = "http://localhost:${seleniumPort}/"
+        t.systemProperty('gwt.args', ["-runStyle io.deephaven.web.junit.RunStyleRemoteWebDriver:${webdriverUrl}?firefox",
+                                      '-ea',
+                                      '-style PRETTY',
+                                      "-war ${layout.buildDirectory.dir('unitTest-war').get().asFile.absolutePath}"
+        ].join(' '))
+        t.classpath += tasks.getByName('gwtCompile').src
+    }
+    t.finalizedBy(deephavenDocker.endTask)
```
niloc132, line 136: remove, regardless of changes above, this is a dup

web/client-api/src/test/java/io/deephaven/web/ClientIntegrationTestSuite.java
```diff
@@ -28,6 +29,7 @@ public static Test suite() {
         suite.addTestSuite(NullValueTestGwt.class);
         suite.addTestSuite(HierarchicalTableTestGwt.class);
         suite.addTestSuite(PartitionedTableTestGwt.class);
+        suite.addTestSuite(LazyPromiseTestGwt.class);
```
niloc132, line 32: Add a comment on this that it isn't really an integration test?

mofojed: I had put a comment in `LazyPromiseTestGwt#getModuleName` itself, but makes sense to have here as well.

niloc132 (approved):

PR #5008, opened 2024-01-04
Revise auto-generated doc issues

### LINKED ISSUE #4978: Revise Auto-generated doc issues
We'd like to downsize to one issue per merged feature, with no "how to" or "reference" labels. The prefix can simply be Community.

mattrunyon:
Fixes #4978 w/ Margaret's requests from the docs team.

It's unnecessary to keep an array of promises since they only want 1 ticket opened, but I didn't want to accidentally introduce a bug by changing the logic around the array since this is kind of a pain to test.


.github/workflows/create-docs-issues.js
```diff
@@ -13,7 +13,7 @@ module.exports = async ({ github, context }) => {
     github.rest.issues.create({
       owner: context.repo.owner,
       repo: ISSUE_REPO_NAME,
-      title: `${type.toUpperCase()}: ${context.payload.pull_request.title}`,
+      title: `${context.payload.pull_request.title}`,
```
mofojed, line 16: We don't want any sort of prefix in the title?

mofojed (approved):

.github/workflows/create-docs-issues.js
mattrunyon: Nope. Margaret requested just the 1 ticket w/ no prefix and adding the `Community` label which will be added since it is the `type`.

PR #5009, opened 2024-01-05
fix: Don't memoize retrieval of tables from JsPartitionedTable#getTable

mofojed:
- Memoizing it is inconsistent with other methods like `WorkerConnection.getTable`, which always returns a new table
- Was unclear who the "owner" of the returned table was
- Memoization can happen at the client level if they need it
- Tested with https://github.com/deephaven/web-client-ui/pull/1663

### DIFF
--- a/web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
+++ b/web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
@@ -22,7 +22,6 @@ import io.deephaven.web.client.api.widget.JsWidget;
 import io.deephaven.web.client.fu.LazyPromise;
 import io.deephaven.web.client.state.ClientTableState;
 import io.deephaven.web.shared.data.RangeSet;
-import io.deephaven.web.shared.fu.JsConsumer;
 import jsinterop.annotations.JsIgnore;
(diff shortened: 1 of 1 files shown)


web/client-api/src/main/java/io/deephaven/web/client/api/JsPartitionedTable.java
```diff
@@ -212,15 +167,36 @@ public Promise<JsTable> getTable(Object key) {
...
             // key doesn't even exist, just hand back a null table
             return Promise.resolve((JsTable) null);
         }
-        return entry.get().then(cts -> Promise.resolve(new JsTable(cts.getConnection(), cts)));
+        final ClientTableState entry = connection.newState((c, cts, metadata) -> {
+            // TODO deephaven-core#2529 parallelize this
+            connection.newTable(
+                    descriptor.getKeyColumnNamesList().asArray(new String[0]),
+                    keyColumnTypes.toArray(new String[0]),
+                    keyList.stream().map(item -> new Object[] {item}).toArray(Object[][]::new),
```
niloc132, line 180: Can you move this up to before the lambda, so we're certain the passed keyList won't have mutated by then? (I realize the original code was "wrong" in this same way.)

niloc132 (approved):

PR #5011, opened 2024-01-05
Auto convert md np array to java array for the output of Python UDFs

### LINKED ISSUE #4989: Feature request: Support multidimensional arrays in query strings returned from Python functions
As a user, I'd like to have the ability to more easily use a Python function that returns a multidimensional array. For instance, something like:

```python
from numpy import typing as npt
import numpy as np

def testA() -> npt.NDArray[np.double]:
  return np.array([[1,2],[2,3]], dtype=np.double)

table = empty_table(1).update([
  "testA = testA()"
])
```

Thanks to `Miek` in community slack, we now know that you can do this in Deephaven:

```python
import jpy 

def testD():
  arrA = jpy.array("double", [1,2])
  arrB = jpy.array("double", [3,4])
  arrC = jpy.array(jpy.get_type('[D'), [arrA, arrB])
  return arrC

table = empty_table(1).update([
  "double_matrix = (double[][]) testD()",
  "top_left = double_matrix[0][0]",
  "top_right = double_matrix[0][1]",
  "bottom_left = double_matrix[1][0]",
  "bottom_right = double_matrix[1][1]",
])
meta = table.meta_table
```

However, it would be nice if it were easier, as in the first example, to get multi-dimensional arrays into a table.

jmao-denver:
Fixes #4989

### DIFF
--- a/py/server/deephaven/jcompat.py
+++ b/py/server/deephaven/jcompat.py
@@ -305,10 +305,11 @@ def _j_array_to_series(dtype: DType, j_array: jpy.JType, conv_null: bool) -> pd.
 
     return s
 
-def _np_array_to_j_array(np_array: np.ndarray) -> Union[jpy.JType | np.ndarray]:
-    """Convert a multidimensional numpy array to a Java array of the same shape and type.
+def _np_array_to_j_array(dtype: dtypes.DType, np_array: np.ndarray) -> Union[jpy.JType | np.ndarray]:
+    """Convert a multidimensional numpy array to a Java array of the same shape based on the given dtype.
 
     Args:
+        dtype (dtypes.DType): The dtype of the Java array, if None, it will be inferred from the numpy array
         np_array (np.ndarray): The numpy array to convert
 
     Returns:
@@ -320,7 +321,9 @@ def _np_array_to_j_array(np_array: np.ndarray) -> Union[jpy.JType | np.ndarray]:
     if np_array is None:
         return None
 
-    dtype = dtypes.from_np_dtype(np_array.dtype)
+    if not dtype:
+        dtype = dtypes.from_np_dtype(np_array.dtype)
+
     array_dtype = _BUILDABLE_ARRAY_DTYPE_MAP.get(dtype, None)
     if array_dtype:
         j_type_char = array_dtype.j_name[-1]
--- a/py/server/deephaven/jcompat.py
+++ b/py/server/deephaven/jcompat.py
@@ -5,7 +5,7 @@
 """ This module provides Java compatibility support including convenience functions to create some widely used Java
 data structures from corresponding Python ones in order to be able to call Java methods. """
 
-from typing import Any, Callable, Dict, Iterable, List, Sequence, Set, TypeVar, Union, Tuple, Literal
+from typing import Any, Callable, Dict, Iterable, List, Sequence, Set, TypeVar, Union, Tuple, Literal, Optional
 
 import jpy
 import numpy as np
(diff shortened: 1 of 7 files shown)


py/server/deephaven/jcompat.py
```diff
@@ -304,3 +304,44 @@ def _j_array_to_series(dtype: DType, j_array: jpy.JType, conv_null: bool) -> pd.
...
+        np_array (np.ndarray): The numpy array to convert
+
+    Returns:
+        jpy.JType | np.ndarray: The Java array or the original numpy array if it could not be converted
+
+    Raises:
+        DHError
+    """
+    if np_array is None:
+        return None
```
chipkent, line 322: Should the type hint and inputs be `Optional`?

py/server/tests/test_udf_return_java_values.py
```diff
@@ -210,7 +212,7 @@ def f() -> np.ndarray[np.int64]:
 
         with self.assertRaises(DHError) as cm:
             t = empty_table(10).update(["X1 = f()"])
-        self.assertIn("not support multi-dimensional arrays", str(cm.exception))
+        self.assertIn("java.lang.ClassCastException: class [[J cannot be cast to class [J", str(cm.exception))
```
chipkent, line 215: Can we get a more python-user-friendly error message, or is the whole message python friendly, and this is just a snippet?

py/server/tests/test_udf_return_java_values.py
```diff
@@ -290,9 +297,30 @@ def test_np_ufunc(self):
...
+            if ndim == 1:
+                return np_1d_array
+            elif ndim == 2:
+                return np_2d_array
+            elif ndim == 3:
+                return np_3d_array
+            else:
+                raise ValueError("ndim must be 1, 2, or 3")
+
+        for ndim in [1, 2, 3]:
```
chipkent, line 319: There should also be an equivalent set of 3 tests with type hints to make sure that our return hints are working properly.

py/server/tests/test_udf_return_java_values.py
```diff
@@ -290,9 +297,30 @@ def test_np_ufunc(self):
...
+            with self.subTest(ndim):
+                cast_str = f"(long{'[]' * ndim})"
+                t = empty_table(10).update([f"X1 = {cast_str}f(ndim)"])
+                self.assertEqual(t.columns[0].data_type, dtypes.from_jtype(jpy.get_type("[" * ndim + "J").jclass))
```
chipkent, line 323: What happens if `f()` returns different dimensioned results?  Do we get an error?  Does it "work" but do weird stuff?  I'd check this case.

py/server/tests/test_udf_return_java_values.py
chipkent, line 1: For the PR, I would adjust the title to indicate that the feature is just addressing return values.

py/server/tests/test_numba_guvectorize.py
```diff
@@ -90,5 +92,15 @@ def g(x, res):
...
 
 
+    def test_2d_array_output(self):
+        @guvectorize([(int64[:], int64[:, :])], "(m)->(m, m)", nopython=True)
+        def g(x, res):
+            for i in range(x.shape[0]):
+                res[i] = x[i] + 5
+
+        t = (empty_table(100).update(["X=i%10", "Y=ii"]).group_by("X")
+             .update("Z=(long[][])g(Y)"))
```
chipkent, line 102: If this feature is working properly, it seems like the cast should not be needed here.

py/server/tests/test_udf_return_java_values.py
chipkent: For the type hinted version, the cast in the query string should not be needed.

py/server/tests/test_udf_return_java_values.py
jmao-denver: It is the most important part of the message (including the full trackback etc.) and it is emitted out of the engine. I don't know there is a safe and easy way to make it more Python-user-friendly.

PR #5013, opened 2024-01-08
Fix pydeephaven-ticking package name and link

rbasralian:
Fix a broken `<a href`, and update the documentation to use the package name (`pydeephaven-ticking`) instead of the module name (`pydeephaven_ticking`).


jmao-denver (approved):
LGTM

rcaudy (approved):

PR #5014, opened 2024-01-08
Allow ExportObject to be RELEASED When Already Managed

### LINKED ISSUE #5002: Occassional TableServiceAsyncTest.longChainAsyncExportAllCancelAllButLast failures
This longChainAsyncExportAllCancelAllButLast test is relying on a potentially incorrect assumption that RELEASED don't cascade (#4781):

> if the export is already EXPORTED (ie, finished initial computation), then the state transitions to RELEASED which does not cancel downstream dependencies

 We've seen this once in nightly tests:

```
  io.deephaven.client.impl.TableServiceAsyncTest > longChainAsyncExportAllCancelAllButLast FAILED
      java.util.concurrent.ExecutionException at TableServiceAsyncTest.java:85
          Caused by: io.deephaven.client.impl.TableHandle$TableHandleException at TableHandle.java:153
```

Natively, this is very hard to reproduce. After introspecting the code though, a small patch makes it happen almost every time:

```diff
diff --git a/java-client/session-dagger/src/test/java/io/deephaven/client/impl/TableServiceAsyncTest.java b/java-client/session-dagger/src/test/java/io/deephaven/client/impl/TableServiceAsyncTest.java
index 1e8b7f188d..47d23c0a89 100644
--- a/java-client/session-dagger/src/test/java/io/deephaven/client/impl/TableServiceAsyncTest.java
+++ b/java-client/session-dagger/src/test/java/io/deephaven/client/impl/TableServiceAsyncTest.java
@@ -24,7 +24,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 public class TableServiceAsyncTest extends DeephavenSessionTestBase {
 
     private static final Duration GETTIME = Duration.ofSeconds(15);
-    private static final int CHAIN_OPS = 250;
+    private static final int CHAIN_OPS = 3;
     private static final int CHAIN_ROWS = 1000;
 
     @Test(timeout = 20000)
diff --git a/server/src/main/java/io/deephaven/server/session/SessionState.java b/server/src/main/java/io/deephaven/server/session/SessionState.java
index 862cb13185..8c53333269 100644
--- a/server/src/main/java/io/deephaven/server/session/SessionState.java
+++ b/server/src/main/java/io/deephaven/server/session/SessionState.java
@@ -1479,6 +1479,11 @@ public class SessionState {
          * @return the submitted export object
          */
         public ExportObject<T> submit(final Callable<T> exportMain) {
+            try {
+                Thread.sleep(1000);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+            }
             export.setWork(exportMain, errorHandler, successHandler, requiresSerialQueue);
             return export;
         }
```

coupled with
```bash
./gradlew java-client-session-dagger:test -PforceTest=true --tests io.deephaven.client.impl.TableServiceAsyncTest.longChainAsyncExportAllCancelAllButLast
```

It seems like:

> doing an isolated FetchTableRequest for each export the user requests

would fix this and #4781.

nbauernfeind:
This fixes #5002.

We intend to allow exports to be `RELEASED` while an `ExportObject` is in limbo waiting for other parent dependencies.  The original intent was to increment the reference count during `setDependencies` to guarantee that if it was `EXPORTED` at that time (or shortly after) that we would be guaranteed to have a valid `result` until the `ExportObject` was `destroy()`'d -- allowing for the external client to `RELEASE` when convenient and without needing to keep track of these dependencies itself.

### DIFF
--- a/server/src/main/java/io/deephaven/server/session/SessionState.java
+++ b/server/src/main/java/io/deephaven/server/session/SessionState.java
@@ -565,6 +565,9 @@ public class SessionState {
 
         /** used to detect when this object is ready for export (is visible for atomic int field updater) */
         private volatile int dependentCount = -1;
+        /** used to detect a parent that was already released prior to having dependencies set */
+        private ExportObject<?> alreadyDeadParent;
+
         @SuppressWarnings("unchecked")
         private static final AtomicIntegerFieldUpdater<ExportObject<?>> DEPENDENT_COUNT_UPDATER =
                 AtomicIntegerFieldUpdater.newUpdater((Class<ExportObject<?>>) (Class<?>) ExportObject.class,
@@ -648,7 +651,9 @@ public class SessionState {
 
             this.parents = parents;
             dependentCount = parents.size();
-            parents.stream().filter(Objects::nonNull).forEach(this::tryManage);
+            alreadyDeadParent = parents.stream()
+                    .filter(p -> p != null && !tryManage(p))
+                    .findFirst().orElse(null);
 
             if (log.isDebugEnabled()) {
                 final Exception e = new RuntimeException();
@@ -678,18 +683,23 @@ public class SessionState {
                 throw new IllegalStateException("export object can only be defined once");
             }
             hasHadWorkSet = true;
-
             if (queryPerformanceRecorder != null && queryPerformanceRecorder.getState() == QueryState.RUNNING) {
                 // transfer ownership of the qpr to the export before it can be resumed by the scheduler
                 queryPerformanceRecorder.suspendQuery();
             }
             this.requiresSerialQueue = requiresSerialQueue;
 
+            // we defer this type of failure until setWork for consistency in error handling
+            if (alreadyDeadParent != null) {
+                onDependencyFailure(alreadyDeadParent);
+            }
+
             if (isExportStateTerminal(state)) {
                 // The following scenarios cause us to get into this state:
                 // - this export object was released/cancelled
                 // - the session expiration propagated to this export object
-                // Note that already failed dependencies will be handled in the onResolveOne method below.
+                // - a parent export was released/dead prior to `setDependencies`
+                // Note that failed dependencies will be handled in the onResolveOne method below.
 
                 // since this is the first we know of the errorHandler, it could not have been invoked yet
                 if (errorHandler != null) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/deephaven/server/session/SessionState.java
```diff
@@ -648,7 +651,9 @@ private synchronized void setDependencies(final List<ExportObject<?>> parents) {
 
             this.parents = parents;
             dependentCount = parents.size();
-            parents.stream().filter(Objects::nonNull).forEach(this::tryManage);
+            alreadyDeadParent = parents.stream()
+                    .filter(p -> p != null && !tryManage(p))
```
devinrsmith, line 655: tryManage is moving to an intermediate operation here as opposed to a terminal operation; this has the potential to not execute tryManage for certain parents.

devinrsmith (changes requested):

server/src/main/java/io/deephaven/server/session/SessionState.java
```diff
@@ -565,6 +565,9 @@ public final static class ExportObject<T> extends LivenessArtifact {
 
         /** used to detect when this object is ready for export (is visible for atomic int field updater) */
         private volatile int dependentCount = -1;
+        /** used to detect a parent that was already released prior to having dependencies set */
```
rcaudy, line 568: Whose dependencies? The parent's, or ours?

server/src/main/java/io/deephaven/server/session/SessionState.java
```diff
@@ -891,44 +901,12 @@ private void onResolveOne(@Nullable final ExportObject<?> parent) {
...
-                                    .endl();
-                        }
-                    }
-
-                    setState(terminalState);
-                    return;
-                }
+            // Is this a cascading failure? Note that we increment the parent reference count in `setDependencies` which
+            // keeps the result live until all children have been exported. This means that the parent is allowed to
+            // be in a RELEASED state, but is not allowed to be in a failure state.
```
rcaudy, line 906:
```suggestion
            // Is this a cascading failure? Note that we manage the parents in `setDependencies` which
            // keeps the parent results live until this child been exported. This means that the parent is allowed to
            // be in a RELEASED state, but is not allowed to be in a failure state.
```

server/src/main/java/io/deephaven/server/session/SessionState.java
```diff
@@ -648,7 +651,9 @@ private synchronized void setDependencies(final List<ExportObject<?>> parents) {
 
             this.parents = parents;
             dependentCount = parents.size();
-            parents.stream().filter(Objects::nonNull).forEach(this::tryManage);
+            alreadyDeadParent = parents.stream()
```
rcaudy, line 654: Should we unmanage all parents we successfully managed, if `alreadyDeadParent` is not null? Are we doing ourselves any favors by preserving extra parents to support this node that cannot succeed?

server/src/main/java/io/deephaven/server/session/SessionState.java
rcaudy: I think this is acceptable/desirable under the circumstances. See https://github.com/deephaven/deephaven-core/pull/5014#discussion_r1445515975

devinrsmith: I disagree. Trying to (ab)use stream with stateful intermediate ops is poor practice even if we think behavior seems ok in this case. Much safer and cleaner to just use for loop and break out for explicitly defined flow control.

nbauernfeind: No biggie. I've pulled this into a traditional for loop.

devinrsmith (approved):
I've tested, and this does seem to fix #5002 so I'm happy. I'll let others review changes to management semantics.

PR #5026, opened 2024-01-10
Python test workflow improvements

rbasralian:
- Support for running only select Python tests (with `-PpyTestFiles=test_file1.py,test_file2.py`).
- Support for running Python tests using local jpy sources (with `-PcustomJpy=/path/to/jpy_sources`). The `customJpy` dir is copied into the container and the wheel is built there.

### DIFF
--- a/Integrations/build.gradle
+++ b/Integrations/build.gradle
@@ -48,7 +48,15 @@ SourceSet test = java.sourceSets.maybeCreate('test')
 
 def runInDocker = { String name, String sourcePath, def command, Closure addConfig = {} ->
     Docker.registerDockerTask(project, name) {
+        final String JPY_SOURCE_PATH = project.hasProperty("customJpy") ? project.property("customJpy") : null
+
         copyIn {
+            if (JPY_SOURCE_PATH != null) {
+                from (JPY_SOURCE_PATH) {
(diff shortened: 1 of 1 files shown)


Integrations/build.gradle
```diff
@@ -44,9 +46,17 @@ sourceSets {
 JavaPluginConvention java = project.convention.plugins.get('java') as JavaPluginConvention
 SourceSet test = java.sourceSets.maybeCreate('test')
 
-def runInDocker = { String name, String sourcePath, List<String> command, Closure addConfig = {} ->
+def runInDocker = { String name, String sourcePath, def command, Closure addConfig = {} ->
     Docker.registerDockerTask(project, name) {
+        final String JPY_SOURCE_PATH = project.hasProperty("customJpy") ? project.property("customJpy") : null
```
niloc132, line 51: Consider moving this to docker/server-jetty/build.gradle, which is upstream of not just this task, but also the client integration tests, and `prepareCompose` itself (allowing devs to test the server directly, not just run through this specific set of tests). Putting it at the end of that set of wiring would make it late in the build (but perhaps not as late as you want, since Integrations will add more stuff after it, though other client integration tests use that image as-is).

PR #5029, opened 2024-01-10
Calendar API changes so that methods accepting strings return strings

chipkent:
Suggested changes to the Calendar API so that methods that accept string inputs return strings.

### DIFF
--- a/engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
+++ b/engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
@@ -1214,14 +1214,15 @@ public class BusinessCalendar extends Calendar {
      * @throws InvalidDateException if the dates are not in the valid range
      * @throws DateTimeUtils.DateTimeParseException if the string cannot be parsed
      */
-    public LocalDate[] businessDates(final String start, final String end, final boolean startInclusive,
+    public String[] businessDates(final String start, final String end, final boolean startInclusive,
             final boolean endInclusive) {
         if (start == null || end == null) {
             return null;
         }
 
-        return businessDates(DateTimeUtils.parseLocalDate(start), DateTimeUtils.parseLocalDate(end), startInclusive,
+        final LocalDate[] dates = businessDates(DateTimeUtils.parseLocalDate(start), DateTimeUtils.parseLocalDate(end), startInclusive,
                 endInclusive);
+        return dates == null ? null : Arrays.stream(dates).map(DateTimeUtils::formatDate).toArray(String[]::new);
     }
 
     /**
@@ -1287,7 +1288,7 @@ public class BusinessCalendar extends Calendar {
      * @throws InvalidDateException if the dates are not in the valid range
      * @throws DateTimeUtils.DateTimeParseException if the string cannot be parsed
      */
-    public LocalDate[] businessDates(final String start, final String end) {
+    public String[] businessDates(final String start, final String end) {
         return businessDates(start, end, true, true);
     }
 
@@ -1357,14 +1358,15 @@ public class BusinessCalendar extends Calendar {
      * @throws InvalidDateException if the dates are not in the valid range
      * @throws DateTimeUtils.DateTimeParseException if the string cannot be parsed
      */
-    public LocalDate[] nonBusinessDates(final String start, final String end, final boolean startInclusive,
+    public String[] nonBusinessDates(final String start, final String end, final boolean startInclusive,
             final boolean endInclusive) {
         if (start == null || end == null) {
             return null;
         }
 
-        return nonBusinessDates(DateTimeUtils.parseLocalDate(start), DateTimeUtils.parseLocalDate(end), startInclusive,
+        final LocalDate[] dates = nonBusinessDates(DateTimeUtils.parseLocalDate(start), DateTimeUtils.parseLocalDate(end), startInclusive,
                 endInclusive);
+        return dates == null ? null : Arrays.stream(dates).map(DateTimeUtils::formatDate).toArray(String[]::new);
     }
 
     /**
@@ -1430,7 +1432,7 @@ public class BusinessCalendar extends Calendar {
      * @throws InvalidDateException if the dates are not in the valid range
      * @throws DateTimeUtils.DateTimeParseException if the string cannot be parsed
      */
-    public LocalDate[] nonBusinessDates(final String start, final String end) {
+    public String[] nonBusinessDates(final String start, final String end) {
         return nonBusinessDates(start, end, true, true);
     }
 
@@ -1748,12 +1750,13 @@ public class BusinessCalendar extends Calendar {
      * @throws InvalidDateException if the date is not in the valid range
      * @throws DateTimeUtils.DateTimeParseException if the string cannot be parsed
      */
-    public LocalDate plusBusinessDays(final String date, final int days) {
+    public String plusBusinessDays(final String date, final int days) {
         if (date == null || days == NULL_INT) {
             return null;
         }
 
-        return plusBusinessDays(DateTimeUtils.parseLocalDate(date), days);
+        final LocalDate d = plusBusinessDays(DateTimeUtils.parseLocalDate(date), days);
+        return d == null ? null : d.toString();
     }
 
     /**
@@ -1840,7 +1843,7 @@ public class BusinessCalendar extends Calendar {
      * @throws InvalidDateException if the date is not in the valid range
      * @throws DateTimeUtils.DateTimeParseException if the string cannot be parsed
      */
-    public LocalDate minusBusinessDays(final String date, final int days) {
+    public String minusBusinessDays(final String date, final int days) {
         if (date == null || days == NULL_INT) {
             return null;
         }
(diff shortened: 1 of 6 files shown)


engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
```diff
@@ -1287,7 +1291,7 @@ public LocalDate[] businessDates(final LocalDate start, final LocalDate end) {
      * @throws InvalidDateException if the dates are not in the valid range
      * @throws DateTimeUtils.DateTimeParseException if the string cannot be parsed
      */
-    public LocalDate[] businessDates(final String start, final String end) {
+    public String[] businessDates(final String start, final String end) {
```
jjbrosnan, line 1294: Every other `businessDate` overload returns a `LocalDate[]`. I don't think the one that takes string inputs needs to be special.

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
```diff
@@ -1357,14 +1361,16 @@ public LocalDate[] nonBusinessDates(final LocalDate start, final LocalDate end,
      * @throws InvalidDateException if the dates are not in the valid range
      * @throws DateTimeUtils.DateTimeParseException if the string cannot be parsed
      */
-    public LocalDate[] nonBusinessDates(final String start, final String end, final boolean startInclusive,
+    public String[] nonBusinessDates(final String start, final String end, final boolean startInclusive,
```
jjbrosnan, line 1364: Same comment as the one on `businessDates`

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
```diff
@@ -1430,7 +1436,7 @@ public LocalDate[] nonBusinessDates(final LocalDate start, final LocalDate end)
      * @throws InvalidDateException if the dates are not in the valid range
      * @throws DateTimeUtils.DateTimeParseException if the string cannot be parsed
      */
-    public LocalDate[] nonBusinessDates(final String start, final String end) {
+    public String[] nonBusinessDates(final String start, final String end) {
```
jjbrosnan, line 1439: Same comment as the one I made on `businessDates`

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
```diff
@@ -165,6 +165,7 @@ private YearData getYearData(final int year) {
 
     // region Getters
 
+    // TODO: should there be a string equivalent?
```
jjbrosnan, line 168: I don't think there needs to be one

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
```diff
@@ -1214,14 +1216,16 @@ public LocalDate[] businessDates(final LocalDate start, final LocalDate end, fin
      * @throws InvalidDateException if the dates are not in the valid range
      * @throws DateTimeUtils.DateTimeParseException if the string cannot be parsed
      */
-    public LocalDate[] businessDates(final String start, final String end, final boolean startInclusive,
+    public String[] businessDates(final String start, final String end, final boolean startInclusive,
```
jjbrosnan, line 1219: Same comment as on the other overload.

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
```diff
@@ -2091,6 +2099,7 @@ public ZonedDateTime minusNonBusinessDays(final ZonedDateTime time, final int da
         return plusNonBusinessDays(time, -days);
     }
 
+    // TODO: should there be a string equivalent?
```
jjbrosnan, line 2102: Isn't there already one?

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
```diff
@@ -2108,6 +2117,7 @@ public LocalDate futureBusinessDate(final int days) {
         return plusBusinessDays(calendarDate(), days);
     }
 
+    // TODO: should there be a string equivalent?
```
jjbrosnan, line 2120: Isn't there already one?

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
```diff
@@ -2125,6 +2135,7 @@ public LocalDate pastBusinessDate(final int days) {
         return minusBusinessDays(calendarDate(), days);
     }
 
+    // TODO: should there be a string equivalent?
```
jjbrosnan, line 2138: Isn't there already one?

engine/time/src/main/java/io/deephaven/time/calendar/Calendar.java
```diff
@@ -89,6 +90,7 @@ public String toString() {
 
     // region Now
 
+    // TODO: should there be a string equivalent?
```
jjbrosnan, line 93: I don't think so

engine/time/src/main/java/io/deephaven/time/calendar/Calendar.java
```diff
@@ -117,6 +119,7 @@ public DayOfWeek dayOfWeek(final LocalDate date) {
         return DateTimeUtils.dayOfWeek(date);
     }
 
+    // TODO: should this return a String?
```
jjbrosnan, line 122: All of the overloads for this method return a `DayOfWeek`. The one that takes string as input doesn't have to be special. It should conform to the behavior the overloads have.

engine/time/src/main/java/io/deephaven/time/calendar/Calendar.java
```diff
@@ -392,6 +397,7 @@ public ZonedDateTime minusDays(final ZonedDateTime time, final int days) {
         return plusDays(time, -days);
     }
 
+    // TODO: should there be a string equivalent?
```
jjbrosnan, line 400: I don't think so.

engine/time/src/main/java/io/deephaven/time/calendar/Calendar.java
```diff
@@ -403,6 +409,7 @@ public LocalDate futureDate(final int days) {
         return plusDays(calendarDate(), days);
     }
 
+    // TODO: should there be a string equivalent?
```
jjbrosnan, line 412: I don't think so.

jjbrosnan (commented):
There were four methods in `BusinessCalendar` that I noticed behavior differed between overloads:

`minusBusinessDays`
`minusNonBusinessDays`
`plusBusinessDays`
`plusNonBusinessDays`

Those four have 4 overloads each - in all but 1 overload, the method returns the same type that's passed in. The only one where that's not the case is when a String is passed in. In that case, a `LocalDate` is returned. Those are the only 4 methods I think should change so that the behavior is consistent across all overloads.

After glancing at `Calendar` as well, there are two methods where I think this should change in the same way:

`minusDays`
`plusDays`

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
chipkent: I assumed that String in implies String out.  If someone is working in strings, do you expect that they would not want this result in strings as well?

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
chipkent: I assumed that String in implies String out.  If someone is working in strings, do you expect that they would not want this result in strings as well?

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
chipkent: I assumed that String in implies String out.  If someone is working in strings, do you expect that they would not want this result in strings as well?

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
chipkent: I assumed that String in implies String out.  If someone is working in strings, do you expect that they would not want this result in strings as well?

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
rbasralian: I agree; people can convert these to strings on their own, or stick with java.time types (as we'd generally recommend).

I don't strongly oppose a `firstValidDateStr()`; I just wouldn't add it proactively. We can always add it later if people ask for it.

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendar.java
rbasralian: I think `String in -> String out`, `LocalDate in -> LocalDate out` is the most user-friendly

rbasralian (commented):
Looks good to me — `String in -> String out`, `LocalDate in -> LocalDate out`, and always `LocalDate` if there are no date arguments.

It does make me wonder if we should make `parseLocalDate()` accept nulls — it would be nice to be able to do `parseLocalDate(someFunctionResultOrJoinedOnColumnThatMayBeNull)` whenever you have a string date, without having to handle nulls manually. We have `parseLocalDateQuiet()` (and similar for other parse methods), but those also swallow format exceptions.

jjbrosnan (approved):

deephaven-internal:

Community: https://github.com/deephaven/deephaven.io/issues/3616

Issue #53, opened 2021-02-01
Consider task cancelation implementation


niloc132:
This could potentially be a check in somewhere like ChunkPoolReleaseTracking.onGive/onTake for the thread's interrupt status (or some other threadlocal flag) and throw a specific exception to indicate that the operation didn't fail, but was canceled. This means in places like ConstructSnapshot.callDataSnapshotFunction where exceptions are expected (for "try again a few more times"), we would need to handle this exception specially.

Probably want to disable this feature during LTM refresh to avoid accidentally killing the ltm thread.

rcaudy:
Ideally, we want to re-integrate interruption for this purpose, because it allows us to break out of blocking I/O and blocking lock waits.

* We can get `Future`s associated with `ExportObject`s that have been submitted with a minor refactoring.
* We still have interrupt checks in `QueryPerformanceRecorder.getNugget().
* We need to audit our I/O, and make sure we're resilient if we encounter closed-by-interrupt errors.
* We're chunky these days. Conceivably, we could check for interrupt in any/every `getChunk` or `fillChunk` call, or `getNextRowSequenceXYZ`.
* We need to examine a lot of our locking and switch to `lockInterruptibly`.

nbauernfeind:
Note that `QueryPerformanceImpl#abortQuery` and `SessionState#doWork` do not yet consider premature abortion patterns.

PR #5032, opened 2024-01-12
Re-implement `ParallelWhere` using `JobScheduler` semantics.

### LINKED ISSUE #4984: ParallelWhere Does Not Ensure Exactly Once Error Delivery
While looking into #4975, @rcaudy and I discovered that parallel where does not propagate errors properly.

The error notifications cannot be executed as the WhereListener's `satisfied` impl checks against `finalNotificationStep` which is never set when an error occurs.
It also does not guarantee exactly one error should multiple errors occur.

Ideally we should be using the IterationManager which was designed for this type of use-case.

lbooker42:
Closes #4984


engine/table/src/main/java/io/deephaven/engine/table/impl/WhereListener.java
```diff
@@ -214,25 +211,26 @@ void setFinalExecutionStep() {
...
                 final boolean runModifiedFilters,
-                final ModifiedColumnSet sourceModColumns,
-                final int filterIndex) {
-            super(WhereListener.this.sourceTable, WhereListener.this.filters, addedInput, addStart, addEnd, modifyInput,
-                    modifyStart, modifyEnd, parent, false, runModifiedFilters, sourceModColumns, filterIndex);
+                final ModifiedColumnSet sourceModColumns) {
+            super(WhereListener.this.sourceTable, WhereListener.this.filters, addedInput, modifyInput,
+                    false, runModifiedFilters, sourceModColumns);
+            // Create the proper JobScheduler for the following parallel tasks
+            if (getUpdateGraph().parallelismFactor() > 1) {
```
nbauernfeind, line 229: Need to check both `permitParallelization` and `QueryTable.FORCE_PARALLEL_WHERE`.

engine/table/src/main/java/io/deephaven/engine/table/impl/WhereListener.java
```diff
@@ -214,25 +211,26 @@ void setFinalExecutionStep() {
...
+                jobScheduler = new UpdateGraphJobScheduler(getUpdateGraph());
+            } else {
+                jobScheduler = ImmediateJobScheduler.INSTANCE;
+            }
         }
 
         @Override
```
nbauernfeind, line 236: The method `doParallelization` can be removed and replaced with `doParallelizationBase` which should be renamed to something like `shouldParallelizeFilter`.

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -1068,7 +1068,11 @@ void doRefilter(
                 final WhereListener.ListenerFilterExecution filterExecution =
                         listener.makeFilterExecution(source.getRowSet().copy());
                 filterExecution.scheduleCompletion(
-                        fe -> completeRefilterUpdate(listener, upstream, update, fe.addedResult));
+                        () -> completeRefilterUpdate(listener, upstream, update, filterExecution.addedResult),
+                        exception -> {
+                            // ignore? Shouldn't we notify the listeners of the failure?
```
nbauernfeind, line 1073: As discussed in your overview, yes we need to propagate the error to table listeners. I would create a method analogous to the `completeRefilterUpdate` which notifies listeners of the error. The implementation must also `upstream.release()` (see `completeRefilterUpdate`).

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -1089,7 +1093,10 @@ void doRefilter(
                         }
                         newMapping.insert(previouslyMatched);
                     }
-                    completeRefilterUpdate(listener, upstream, update, fe.addedResult);
+                    completeRefilterUpdate(listener, upstream, update, filterExecution.addedResult);
+                }, exception -> {
+                    // ignore? Shouldn't we notify the listeners of the failure?
+                    System.out.println("Exception in refilter: " + exception.getMessage());
```
nbauernfeind, line 1099: Should be able to re-use your helper for the instance on line 1074.

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -1105,7 +1112,11 @@ void doRefilter(
                 final WhereListener.ListenerFilterExecution filterExecution =
                         listener.makeFilterExecution(matchedClone);
                 filterExecution.scheduleCompletion(
-                        fe -> completeRefilterUpdate(listener, upstream, update, fe.addedResult));
+                        () -> completeRefilterUpdate(listener, upstream, update, filterExecution.addedResult),
+                        exception -> {
+                            // ignore? Shouldn't we notify the listeners of the failure?
+                            System.out.println("Exception in refilter: " + exception.getMessage());
```
nbauernfeind, line 1118: Should be able to re-use your helper for the instance on line 1074.

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -1226,41 +1237,20 @@ private QueryTable whereInternal(final WhereFilter... filters) {
...
-                                                currentMappingFuture.completeExceptionally(x.exceptionResult);
-                                            } else {
-                                                currentMappingFuture.complete(x.addedResult.toTracking());
-                                            }
-                                        }
+                                            this, filters, rowSetToUse.copy(), usePrev);
+                                    final TrackingWritableRowSet currentMapping;
+                                    initialFilterExecution.scheduleCompletion(() -> {
+                                        currentMappingFuture.complete(initialFilterExecution.addedResult.toTracking());
+                                    }, exception -> {
```
nbauernfeind, line 1246: Cutely, you can pass in the method reference instead: `currentMappingFuture::completeExceptionally`

engine/table/src/main/java/io/deephaven/engine/table/impl/AbstractFilterExecution.java
```diff
@@ -48,273 +43,287 @@ abstract class AbstractFilterExecution extends AbstractNotification {
...
-            this.modifyResult = other.modifyResult;
-        } else if (other.modifyResult != null) {
-            this.modifyResult.insert(other.modifyResult);
+    private void doFilter(
+            final WhereFilter filter,
+            final RowSet addsToUse,
+            final RowSet modsToUse,
+            final FilterComplete onComplete,
+            final Consumer<Exception> onError) {
+        if (Thread.interrupted()) {
```
nbauernfeind, line 156: I'm a little torn on making this suggestion. It feels like this should defer to the other method. The `subSetByPositionRange` will detect that the appropriate thing to do is to do a `copy()`. On the other hand, both methods are super straightforward and this implementation is ever-so-slightly faster.

Leaving the snippet just for consideration. Maybe Ryan will have a strong preference.
```java
        doFilter(filter, addsToUse, 0, addsToUse == null ? 0 : addsToUse.size(),
                modsToUse, 0, modsToUse == null ? 0 : modsToUse.size(),
                onComplete, onError);
```

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -982,6 +980,37 @@ public static class FilteredTable extends QueryTable implements WhereFilter.Reco
         private boolean refilterUnmatchedRequested = false;
         private MergedListener whereListener;
 
+        @ReferentialIntegrity
+        private Runnable delayedErrorReference;
+
+        private static final class DelayedErrorNotifier implements Runnable {
```
nbauernfeind, line 986: You should rebase and use the version of this that dropped in `io.deephaven.engine.table.impl.util`.

See https://github.com/deephaven/deephaven-core/blob/main/engine/table/src/main/java/io/deephaven/engine/table/impl/util/DelayedErrorNotifier.java.

engine/table/src/main/java/io/deephaven/engine/table/impl/QueryTable.java
```diff
@@ -1148,6 +1179,19 @@ private void completeRefilterUpdate(
...
 
+        private void errorRefilterUpdate(final WhereListener listener, final Exception e) {
+            // Notify listeners that we had an issue refreshing the table.
+            if (getLastNotificationStep() == updateGraph.clock().currentStep()) {
+                if (listener != null) {
+                    listener.forceReferenceCountToZero();
+                }
+                delayedErrorReference = new DelayedErrorNotifier(e, this);
+            } else {
+                notifyListenersOnError(e, null);
```
nbauernfeind, line 1190: This `null` parameter should be `listener == null ? null : listener.entry`.

engine/table/src/main/java/io/deephaven/engine/table/impl/WhereListener.java
```diff
@@ -39,13 +44,43 @@ class WhereListener extends MergedListener {
...
     private final boolean permitParallelization;
     private final int segmentCount;
 
     private volatile long initialNotificationStep = NotificationStepReceiver.NULL_NOTIFICATION_STEP;
     private volatile long finalNotificationStep = NotificationStepReceiver.NULL_NOTIFICATION_STEP;
 
+    @ReferentialIntegrity
+    private Runnable delayedErrorReference;
+
+    private static final class DelayedErrorNotifier implements Runnable {
```
nbauernfeind, line 56: Same comment on this `DelayedErrorNotifier`; rebase + use the version in `io.deephaven.engine.impl.util`.

engine/table/src/main/java/io/deephaven/engine/table/impl/WhereListener.java
```diff
@@ -184,6 +216,17 @@ private void completeUpdate(
         setFinalExecutionStep();
     }
 
+    private void errorUpdate(final Exception e) {
+        // Notify listeners that we had an issue refreshing the table.
+        if (result.getLastNotificationStep() == result.updateGraph.clock().currentStep()) {
+            forceReferenceCountToZero();
+            delayedErrorReference = new DelayedErrorNotifier(e, result);
+        } else {
+            result.notifyListenersOnError(e, null);
```
nbauernfeind, line 225: replace `null` with `entry`.

engine/table/src/test/java/io/deephaven/engine/table/impl/QueryTableWhereTest.java
```diff
@@ -227,6 +228,7 @@ public void testWhereOneOfTwo() {
                 i(4, 6, 8).toTracking(), col("x", 21, 3, 4), col("y", 'x', 'c', 'f')));
     }
 
+    @Ignore
```
nbauernfeind, line 231: Let's add a comment and refer to the existing issue https://github.com/deephaven/deephaven-core/issues/53

engine/table/src/test/java/io/deephaven/engine/table/impl/QueryTableWhereTest.java
```diff
@@ -1175,4 +1060,44 @@ public void testBigTableInitial() {
...
+                col("x", 1, 2, 3, 4),
+                col("y", "a", "b", "c", null));
+
+        try {
+            final QueryTable whereResult = (QueryTable) table.where("y.length() > 0");
+        } catch (Exception e) {
+            Assert.eqTrue(e instanceof FormulaEvaluationException
+                    && e.getCause() != null && e.getCause() instanceof NullPointerException,
+                    "NPE causing FormulaEvaluationException expected.");
+        }
```
nbauernfeind, line 1077: We should fail if the exception is not thrown.

nbauernfeind (dismissed):

rcaudy (commented):
See additional patch we worked on together.

engine/table/src/main/java/io/deephaven/engine/table/impl/WhereListener.java
```diff
@@ -220,15 +234,11 @@ public boolean satisfied(long step) {
...
+    ListenerFilterExecution makeRefilterExecution(final RowSet refilter) {
         return new ListenerFilterExecution(refilter, null, false, ModifiedColumnSet.ALL);
     }
 
-    void setFinalExecutionStep() {
-        finalNotificationStep = getUpdateGraph().clock().currentStep();
-    }
-
-    ListenerFilterExecution makeFilterExecution() {
+    ListenerFilterExecution makeRefilterExecution() {
```
rcaudy, line 241: Should be `makeFilterExecution()` or `makeUpdateExecution()`.

rcaudy (dismissed):

rcaudy (approved):

PR #5036, opened 2024-01-12
Parse legacy calendar XML files.

chipkent:
Parse legacy calendar XML files.


engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendarXMLParser.java
```diff
@@ -103,12 +122,16 @@ private static BusinessCalendarInputs parseBusinessCalendarInputs(@NotNull final
...
             calendarElements.timeZone = TimeZoneAliases.zoneId(getText(getRequiredChild(root, "timeZone")));
-            calendarElements.description = getText(getRequiredChild(root, "description"));
-            calendarElements.firstValidDate =
-                    DateTimeUtils.parseLocalDate(getText(getRequiredChild(root, "firstValidDate")));
-            calendarElements.lastValidDate =
-                    DateTimeUtils.parseLocalDate(getText(getRequiredChild(root, "lastValidDate")));
+            calendarElements.description = getText(root.getChild("description"));
             calendarElements.holidays = parseHolidays(root, calendarElements.timeZone);
+            final String firstValidDateStr = getText(root.getChild("firstValidDate"));
+            calendarElements.firstValidDate = DateTimeUtils.parseLocalDate(
```
devinrsmith, line 128: Shouldn't need to parse in the null case; can just use the min value, yeah? Same w/ max.

engine/time/src/main/java/io/deephaven/time/calendar/Calendar.java
```diff
@@ -42,7 +42,7 @@ public class Calendar {
      */
     Calendar(final String name, final String description, final ZoneId timeZone) {
         this.name = Require.neqNull(name, "name");
-        this.description = Require.neqNull(description, "description");
+        this.description = description;
```
devinrsmith, line 45: javadoc incorrect now.

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendarXMLParser.java
chipkent: I think you are asking if `firstValidDate` even needs to be present.  It does.  The calendar could be valid before the first holiday.

devinrsmith: I'm saying when the element is null, we shouldn't need to take go from LocalDate -> String -> LocalDate, ie:

`DateTimeUtils.parseLocalDate(Collections.min(calendarElements.holidays.keySet()).toString())`

devinrsmith: `Collections.min(calendarElements.holidays.keySet())` should suffice.

devinrsmith (approved):

deephaven-internal:

Community: https://github.com/deephaven/deephaven.io/issues/3619

PR #5037, opened 2024-01-12
Move calendar initialization to server startup

devinrsmith:


cpwright (dismissed):

engine/time/src/main/java/io/deephaven/time/calendar/BusinessCalendarXMLParser.java
```diff
@@ -89,38 +92,78 @@ public static BusinessCalendar loadBusinessCalendar(@NotNull final String file)
...
+     * @param resource XML input stream
+     * @return business calendar.
+     */
+    public static BusinessCalendar loadBusinessCalendarFromResource(String resource) throws IOException {
+        final InputStream in = Calendars.class.getResourceAsStream(resource);
+        if (in == null) {
+            throw new RuntimeException("Could not open resource " + resource + " from classpath");
+        }
+        try (final InputStream bin = new BufferedInputStream(in)) {
+            return BusinessCalendarXMLParser.loadBusinessCalendar(bin);
```
chipkent, line 127:
```suggestion
            return loadBusinessCalendar(bin);
```

chipkent (approved):
LGTM.  One optional comment.

This will conflict with #5036, so we should try to merge #5036 first, if possible.

engine/time/src/test/java/io/deephaven/time/calendar/CalendarInit.java
```diff
@@ -0,0 +1,17 @@
...
+
+final class CalendarInit {
+
+    static {
+        for (BusinessCalendar calendar : Calendars.calendarsFromConfiguration()) {
+            Calendars.addCalendar(calendar);
+        }
+    }
+
+    static void noop() {
```
chipkent, line 14: I would document why this method exists.  Also add a class doc about the purpose of the class.

chipkent (approved):

PR #5038, opened 2024-01-13
BarrageTable: Use WritableRowRedirectionLockFree in Most Cases

### LINKED ISSUE #5035: Barrage Client Leaks Memory with Some Index Patterns
**Description**

Produce the following table:
```
sourceTable = timeTable("PT0.001s").tail(1000)
st = new Table[64]
for (int ii = 0; ii < st.length; ++ii) {
    st[ii] = sourceTable
}
m = merge(st)
```

Subscribe to it.  The redirection index in the barrage client table grows because we cycle through the indices; and the SparseArraySource backing it does not know that the block is now null and meaningless.

**Versions**
 - Deephaven: ...
 - OS: ...
 - Browser: ...
 - Docker: ...

nbauernfeind:
Fixes #5035.

### DIFF
--- a/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/table/BarrageTable.java
+++ b/extensions/barrage/src/main/java/io/deephaven/extensions/barrage/table/BarrageTable.java
@@ -461,21 +461,17 @@ public abstract class BarrageTable extends QueryTable implements BarrageMessage.
         } else {
             final WritableRowRedirection rowRedirection;
             final boolean isFlat = getAttribute.test(BarrageUtil.TABLE_ATTRIBUTE_IS_FLAT);
-            if (getAttribute.test(Table.APPEND_ONLY_TABLE_ATTRIBUTE)
-                    || getAttribute.test(Table.ADD_ONLY_TABLE_ATTRIBUTE)
-                    || isFlat) {
+            if (getAttribute.test(Table.APPEND_ONLY_TABLE_ATTRIBUTE) || isFlat) {
                 rowRedirection = new LongColumnSourceWritableRowRedirection(new LongSparseArraySource());
             } else {
-                rowRedirection = WritableRowRedirectionLockFree.FACTORY.createRowRedirection(1024);
+                rowRedirection = WritableRowRedirection.FACTORY.createRowRedirection(1024);
             }
 
             final LinkedHashMap<String, ColumnSource<?>> finalColumns =
                     makeColumns(columns, writableSources, rowRedirection);
             table = new BarrageRedirectedTable(
-                    registrar, queue, executor, finalColumns, writableSources, rowRedirection, attributes, vpCallback);
-            if (isFlat) {
-                table.setFlat();
-            }
+                    registrar, queue, executor, finalColumns, writableSources, rowRedirection, attributes, isFlat,
+                    vpCallback);
         }
 
         return table;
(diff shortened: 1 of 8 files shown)


extensions/barrage/src/main/java/io/deephaven/extensions/barrage/table/BarrageTable.java
```diff
@@ -449,18 +449,33 @@ public static BarrageTable make(
...
+                rowRedirection = new LongColumnSourceWritableRowRedirection(new LongSparseArraySource());
+            } else {
+                rowRedirection = WritableRowRedirectionLockFree.FACTORY.createRowRedirection(1024);
+            }
+
             final LinkedHashMap<String, ColumnSource<?>> finalColumns =
                     makeColumns(columns, writableSources, rowRedirection);
             table = new BarrageRedirectedTable(
                     registrar, queue, executor, finalColumns, writableSources, rowRedirection, attributes, vpCallback);
+            if (isFlat) {
```
rcaudy, line 476: Maybe we should pass this into the constructor, for consistency with `BarrageBlinkTable` in terms of where we `setFlat`?

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/table/BarrageTable.java
```diff
@@ -449,18 +449,33 @@ public static BarrageTable make(
...
             final LinkedHashMap<String, ColumnSource<?>> finalColumns = makeColumns(columns, writableSources);
             table = new BarrageBlinkTable(
                     registrar, queue, executor, finalColumns, writableSources, attributes, vpCallback);
         } else {
-            final WritableRowRedirection rowRedirection =
-                    new LongColumnSourceWritableRowRedirection(new LongSparseArraySource());
+            final WritableRowRedirection rowRedirection;
+            final boolean isFlat = getAttribute.test(BarrageUtil.TABLE_ATTRIBUTE_IS_FLAT);
+            if (getAttribute.test(Table.APPEND_ONLY_TABLE_ATTRIBUTE)
+                    || getAttribute.test(Table.ADD_ONLY_TABLE_ATTRIBUTE)
```
rcaudy, line 465: This (add-only) is a little bit dangerous, as discussed in the DM. Are we sure?

extensions/barrage/src/main/java/io/deephaven/extensions/barrage/table/BarrageTable.java
```diff
@@ -449,18 +449,33 @@ public static BarrageTable make(
...
+                    || isFlat) {
+                rowRedirection = new LongColumnSourceWritableRowRedirection(new LongSparseArraySource());
+            } else {
+                rowRedirection = WritableRowRedirectionLockFree.FACTORY.createRowRedirection(1024);
```
rcaudy, line 469: `FACTORY` is on `io.deephaven.engine.table.impl.util.WritableRowRedirection`:
```suggestion
                rowRedirection = WritableRowRedirection.FACTORY.createRowRedirection(1024);
```

It also has wrong JavaDoc:
```
    /**
     * Factory for producing WritableRowSets and their components.
     */
```

rcaudy (approved):
Approved, with some nits. Unsure about add-only.

rcaudy (approved):

PR #5041, opened 2024-01-16
Fix mismatched list and count in docs

rbasralian:
Fixes a place where the Python docs say "all **three** packages" and then proceed to list **four** packages.

### DIFF
--- a/py/client-ticking/README.md
+++ b/py/client-ticking/README.md
@@ -8,8 +8,8 @@ create a thin wrapper around the Deephaven native C++ library.
 ## Prerequisites
 
 Before installing this library, you will need to install the Deephaven Core C++ library and the
-pydeephaven python package. All three packages (Deephaven Core C++, pydeephaven, Deephaven Core,
-and pydeephaven-ticking) are present in the Deephaven Core github repository. We assume you
+pydeephaven python package. All four packages (Deephaven Core C++, pydeephaven, Deephaven Core,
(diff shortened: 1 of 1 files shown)


chipkent (approved):

py/client-ticking/README.md
```diff
@@ -8,8 +8,8 @@ create a thin wrapper around the Deephaven native C++ library.
 ## Prerequisites
 
 Before installing this library, you will need to install the Deephaven Core C++ library and the
-pydeephaven python package. All three packages (Deephaven Core C++, pydeephaven, Deephaven Core,
-and pydeephaven-ticking) are present in the Deephaven Core github repository. We assume you
+pydeephaven python package. All four packages (Deephaven Core C++, pydeephaven, Deephaven Core,
+and pydeephaven-ticking) are present in the Deephaven Core GitHub repository. We assume you
```
jmao-denver, line 12: I don't see 'Deephaven Core' used in the README. Maybe we should keep the number of packages and remove it instead?

chipkent:
```suggestion
pydeephaven python package. All three packages (Deephaven Core C++, pydeephaven, 
and pydeephaven-ticking) are present in the Deephaven Core GitHub repository. We assume you
```
I think you are suggesting this, which makes sense.

cpwright (dismissed):

margaretkennedy (approved):

PR #5050, opened 2024-01-18
fix: Mark JsPartitionedTable.getTable as nullable

mofojed:
- JsPartitionedTable.getTable returns `null` if the key is not found, as the table may not exist yet.


mofojed:
@niloc132 is there a recommended way of throwing a specific error/error type so that I can react to this type of exception? In this case the client can easily just select a different key and recover.

PR #5053, opened 2024-01-18
Chunk pools should be easier to emulate in GWT

niloc132:
Several changes to simplify how Chunk pooling will work in the JS API:
 - close() calls of borrowed instances should return the instance to the pool consistently
 - Separate pool implementations from interface
 - Flattened method calls required to borrow/create a chunk

### DIFF
--- a/engine/chunk/src/main/java/io/deephaven/chunk/util/pools/MultiChunkPool.java
+++ b/engine/chunk/src/main/java/io/deephaven/chunk/util/pools/MultiChunkPool.java
@@ -4,6 +4,34 @@
 package io.deephaven.chunk.util.pools;
 
 import io.deephaven.chunk.ChunkType;
+import io.deephaven.chunk.ResettableBooleanChunk;
+import io.deephaven.chunk.ResettableByteChunk;
+import io.deephaven.chunk.ResettableCharChunk;
+import io.deephaven.chunk.ResettableDoubleChunk;
+import io.deephaven.chunk.ResettableFloatChunk;
+import io.deephaven.chunk.ResettableIntChunk;
+import io.deephaven.chunk.ResettableLongChunk;
+import io.deephaven.chunk.ResettableObjectChunk;
+import io.deephaven.chunk.ResettableShortChunk;
+import io.deephaven.chunk.ResettableWritableBooleanChunk;
+import io.deephaven.chunk.ResettableWritableByteChunk;
+import io.deephaven.chunk.ResettableWritableCharChunk;
+import io.deephaven.chunk.ResettableWritableDoubleChunk;
+import io.deephaven.chunk.ResettableWritableFloatChunk;
+import io.deephaven.chunk.ResettableWritableIntChunk;
+import io.deephaven.chunk.ResettableWritableLongChunk;
+import io.deephaven.chunk.ResettableWritableObjectChunk;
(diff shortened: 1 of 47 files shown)


engine/chunk/src/main/java/io/deephaven/chunk/util/pools/MultiChunkPool.java
```diff
@@ -29,37 +62,37 @@ public static MultiChunkPool forThisThread() {
...
+    private final ByteChunkPool byteChunkPool = new ByteChunkSoftPool();
+    private final ShortChunkPool shortChunkPool = new ShortChunkSoftPool();
+    private final IntChunkPool intChunkPool = new IntChunkSoftPool();
+    private final LongChunkPool longChunkPool = new LongChunkSoftPool();
+    private final FloatChunkPool floatChunkPool = new FloatChunkSoftPool();
+    private final DoubleChunkPool doubleChunkPool = new DoubleChunkSoftPool();
+    private final ObjectChunkPool objectChunkPool = new ObjectChunkSoftPool();
 
-    private final Map<ChunkType, ChunkPool> pools;
+    private final Map<ChunkType, Supplier<ChunkPool>> pools;
```
rcaudy, line 75: I know it feels silly since the generic `ChunkPool` interface is unused, but it seems to me like we could use the `asChunkPool` methods once each to populate the map, instead of invoking the supplier on demand. It would probably use no more space, etc.

engine/chunk/src/main/java/io/deephaven/chunk/util/pools/MultiChunkPool.java
```diff
@@ -4,17 +4,50 @@
...
 import java.util.EnumMap;
 import java.util.Map;
+import java.util.function.Supplier;
 
 /**
  * Provides a set of per-type {@link ChunkPool}s. Normally accessed via a {@link ThreadLocal}, to allow some threads to
  * share a common pool and others to allocate their own.
+ * <p>
+ * This type isn't compatible with GWT, and needs to be replaced with a different class that doesn't use the soft pool
+ * instances.
```
rcaudy, line 47: For some reason, I'm not in love with having comments about GWT around the rest of the code base. This one, in particular, seems more like a note to you describing future work than anything for readers of `MultiChunkPool` to care about.

niloc132: Fair - we're going to need to address the topic somewhere though so that other developers can:
 * see that their changes affect the js api, either by breaking things accidentally, or by automatically adding features there
 * see that in some cases that isn't true, where the js api has to maintain its own separate copy of classes

It could all live in web/client-api's developer docs (i.e. readme and such), but I'm not sure that will be enough. Unless we get very aggressive about splitting various utility types, we're inevitably going to need to add actual `@GwtIncompatible` annotations here and there in source too.

engine/chunk/src/main/java/io/deephaven/chunk/util/pools/MultiChunkPool.java
niloc132: Good call - I had it that way to begin with, but rewrote it this way, easy to revert.

rcaudy (approved):

PR #5058, opened 2024-01-22
Replace usage of trove in Chunk APIs

niloc132:
Partial #188


extensions/barrage/src/main/java/io/deephaven/extensions/barrage/chunk/BooleanChunkInputStreamGenerator.java
```diff
@@ -175,7 +175,7 @@ static WritableChunk<Values> extractChunkFromInputStreamWithConversion(
             final StreamReaderOptions options,
             final ByteConversion conversion,
             final Iterator<FieldNodeInfo> fieldNodeIter,
-            final TLongIterator bufferInfoIter,
+            final PrimitiveIterator.OfLong bufferInfoIter,
```
rcaudy, line 178: Replace usages of `next()` with `nextLong()`.

extensions/barrage/src/test/java/io/deephaven/extensions/barrage/chunk/BarrageColumnRoundTripTest.java
```diff
@@ -570,14 +571,15 @@ private static <T> void testRoundTripSerialization(
                 final ArrayList<ChunkInputStreamGenerator.FieldNodeInfo> fieldNodes = new ArrayList<>();
                 column.visitFieldNodes((numElements, nullCount) -> fieldNodes
                         .add(new ChunkInputStreamGenerator.FieldNodeInfo(numElements, nullCount)));
-                final TLongArrayList bufferNodes = new TLongArrayList();
+                final LongStream.Builder bufferNodes = LongStream.builder();
```
rcaudy, line 574: This might be a good use case for a RowSetBuilderSequential or RowSetBuilderRandom.

rcaudy (changes requested):

rcaudy (commented):
We still have 2 usages of Trove in the module, once we clean up the unused imports.
It looks to me like we can eliminate both trivially, if needed.

extensions/barrage/src/test/java/io/deephaven/extensions/barrage/chunk/BarrageColumnRoundTripTest.java
nbauernfeind: They are definitely ordered and sorted. These are offsets into the payload of where to find each buffer.

rcaudy: Agreed. It's a test anyway, not worth the effort.

rcaudy (approved):

PR #5069, opened 2024-01-23
Add support for ZonedDateTime literals to the query language

chipkent:
Add support for ZonedDateTime literals to the query language

### DIFF
--- a/engine/time/src/main/java/io/deephaven/time/TimeLiteralReplacedExpression.java
+++ b/engine/time/src/main/java/io/deephaven/time/TimeLiteralReplacedExpression.java
@@ -73,7 +73,8 @@ public class TimeLiteralReplacedExpression {
         final StringBuilder convertedFormula = new StringBuilder();
 
         int localDateIndex = 0;
-        int dateTimeIndex = 0;
+        int instantIndex = 0;
+        int zdtIndex = 0;
         int nanosIndex = 0;
         int periodIndex = 0;
         int durationIndex = 0;
@@ -90,13 +91,12 @@ public class TimeLiteralReplacedExpression {
             }
 
             if (DateTimeUtils.parseInstantQuiet(s) != null) {
-                matcher.appendReplacement(convertedFormula, "_instant" + dateTimeIndex);
-                instanceVariablesString.append("        private Instant _instant").append(dateTimeIndex)
+                matcher.appendReplacement(convertedFormula, "_instant" + instantIndex);
+                instanceVariablesString.append("        private Instant _instant").append(instantIndex)
                         .append("=DateTimeUtils.parseInstant(\"")
                         .append(expression, matcher.start() + 1, matcher.end() - 1).append("\");\n");
-                newVariables.put("_instant" + dateTimeIndex, Instant.class);
-
-                dateTimeIndex++;
+                newVariables.put("_instant" + instantIndex, Instant.class);
+                instantIndex++;
             } else if (DateTimeUtils.parsePeriodQuiet(s) != null) {
                 matcher.appendReplacement(convertedFormula, "_period" + periodIndex);
                 instanceVariablesString.append("        private java.time.Period _period").append(periodIndex)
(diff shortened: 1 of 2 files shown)


alexpeters1208 (dismissed):
Tested, LGTM

devinrsmith (commented):
I'm not sure I understand the exactly where this fits into the full scope of things... that said, all of the other examples are direct calls into DateTimeUtils. If we want to support `Z` prefix, should that instead be implemented and documented on `DateTimeUtils.parseZonedDateTimeQuiet` / `io.deephaven.time.DateTimeUtils#parseZonedDateTime` (and everything that calls it)?

chipkent:
Good suggestion.  Changes were made.

The root of the feature is that we have query string literals for all of the key Java datetime types except for ZonedDateTime.  This adds a literal for completeness.  e.g.
```
t = t.update(["Timestamp = 'Z2023-11-23T12:34 ET' + HOUR"])
```

alexpeters1208 (approved):
Tested locally, LGTM

engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -3886,6 +3886,11 @@ public static Instant parseInstant(@NotNull final String s) {
                 return epochNanosToInstant(nanos);
             }
 
+            if (s.startsWith("Z") || s.startsWith("z")) {
+                throw new RuntimeException(
+                        "A 'Z' or 'z' prefix is not allowed for Instant parsing.  Use parseZonedDateTime instead.");
+            }
```
devinrsmith, line 3892: I'm not a fan of the way this is structured implementation-wise; also, `parseEpochNanos` delegates to parseZonedDateTime - I'm not sure if we _want_ parseEpochNanos to accept `Z` prefix, but if we do, it should have javadoc updated too.

I think it would be better to have a `parseZonedDateTimeImpl` that is private and _doesn't_ accept `Z`, and existing callers (here and maybe parseEpochNanos) can instead call the Impl version.

devinrsmith (commented):
Okay. Your comment makes sense. It does give me a bit of pause about my original suggestion to move it into the utils layer, but I think this does deserve a wider audience about query expression literal support. I'm assuming the idea is that the literals get parsed at query compile time as opposed to query execution time, and we want some way to express that.

While a prefix is an easy solution now for this specific situation, it's worthwhile asking if it's ergonomic enough and something we actually want to do now.

For example, I can theoretically imagine support for "any" literal constant at query compile time via some theoretical syntax:

```java
t = t.update(["ZonedTs = <parseZonedDateTime(`2023-11-23T12:34 ET`)>"])
```

where the idea is that `<...>` is computed once at query compile time. Of course, I'm also interested in more declarative / QST ways of provided a constant, but I'm not against improvements that wash through compile layer.

cc @rbasralian @rcaudy

cpwright:
> Okay. Your comment makes sense. It does give me a bit of pause about my original suggestion to move it into the utils layer, but I think this does deserve a wider audience about query expression literal support. I'm assuming the idea is that the literals get parsed at query compile time as opposed to query execution time, and we want some way to express that.
> 
> While a prefix is an easy solution now for this specific situation, it's worthwhile asking if it's ergonomic enough and something we actually want to do now.
> 
> For example, I can theoretically imagine support for "any" literal constant at query compile time via some theoretical syntax:
> 
> ```java
> t = t.update(["ZonedTs = <parseZonedDateTime(`2023-11-23T12:34 ET`)>"])
> ```
> 
> where the idea is that `<...>` is computed once at query compile time. Of course, I'm also interested in more declarative / QST ways of provided a constant, but I'm not against improvements that wash through compile layer.
> 
> cc @rbasralian @rcaudy

I agree that if we are to expand the kinds of literals we support, then we should do it generally.  I believe the most correct and standard way to do that would be via a syntax compatible with String Templates.
https://openjdk.org/jeps/430

The more custom things we have in our query language parser, the harder it gets to maintain and extend.  It is one of the things I am always the most nervous about changing.

rcaudy:
Do we really need more automatic-conversions? Users can make a ZoneDateTime and put it in the QueryScope. It's hard for me to want to encourage ZDT usage, since it's got more ambiguity than Instant. What's the use-case?

rcaudy:
I think @cpwright's recommendation boils down to:
If we're going to do this, make the syntax more like `ZDT.`

PR #5077, opened 2024-01-26
Use a Deque to avoid StackOverflow in ImmediateJobScheduler

nbauernfeind:
This helps avoid StackOverflowErrors during select/update operations with many layers.


engine/table/src/main/java/io/deephaven/engine/table/impl/util/ImmediateJobScheduler.java
```diff
@@ -49,7 +51,7 @@ public void submit(
                 job.run();
             }
         } finally {
-            processingThread.set(null);
+            PROCESSING_THREAD_UPDATER.set(this, null);
```
rcaudy, line 54: This is OK, but you can also just do the assignment:
```
processingThread = null;
```

rcaudy (approved):

PR #5079, opened 2024-01-26
Support the comparison of String and LocalDate

chipkent:
For backwards compatibility reasons, it is desirable to be able to directly compare String and LocalDate objects.  There is potential overhead from doing this, but the overhead was determined to be acceptable for not breaking compatibility.

For example:
```python

t = empty_table(1).update([
    "X = '2021-01-13'",
    "Y = new String[]{`2021-01-12`,`2021-01-13`,`2021-01-14`}"
]).ungroup()

t2 = t.where("X > Y")
```

image


rcaudy (changes requested):
This is exactly the kind of change I had in mind. It needs to go through `io.deephaven.engine.table.impl.lang.QueryLanguageFunctionGenerator` however.

chipkent:
I believe the final decision was to not make conversions.

PR #5085, opened 2024-01-30
Add `RollingFormula` feature to `UpdateBy`

### LINKED ISSUE #4052: Add `RollingFormula` operator to `UpdateBy`
Similar to `AggFormula`, add a `RollingFormula` operator to `UpdateBy` allowing a user to supply a user-defined-function to be applied to a window of values for a column (or list of columns). 

See #4195
See #4194

lbooker42:
Closes #4052

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateByOperatorFactory.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateByOperatorFactory.java
@@ -16,6 +16,7 @@ import io.deephaven.engine.table.impl.updateby.fill.*;
 import io.deephaven.engine.table.impl.updateby.minmax.*;
 import io.deephaven.engine.table.impl.updateby.prod.*;
 import io.deephaven.engine.table.impl.updateby.rollingcount.*;
+import io.deephaven.engine.table.impl.updateby.rollingformula.*;
 import io.deephaven.engine.table.impl.updateby.rollinggroup.*;
 import io.deephaven.engine.table.impl.updateby.rollingavg.*;
 import io.deephaven.engine.table.impl.updateby.rollingminmax.*;
@@ -513,6 +514,20 @@ public class UpdateByOperatorFactory {
             Arrays.stream(pairs)
                     .filter(p -> !isTimeBased || !p.rightColumn().equals(timestampCol))
                     .map(fc -> makeRollingCountOperator(fc,
+                            source,
+                            spec))
+                    .forEach(ops::add);
+            return null;
+        }
+
+        @Override
+        public Void visit(@NotNull final RollingFormulaSpec spec) {
+            final boolean isTimeBased = spec.revWindowScale().isTimeBased();
+            final String timestampCol = spec.revWindowScale().timestampCol();
+
+            Arrays.stream(pairs)
+                    .filter(p -> !isTimeBased || !p.rightColumn().equals(timestampCol))
+                    .map(fc -> makeRollingFormulaOperator(fc,
                             source,
                             spec))
                     .forEach(ops::add);
@@ -1341,5 +1356,69 @@ public class UpdateByOperatorFactory {
 
             throw new IllegalArgumentException("Can not perform RollingWAvg on type " + csType);
         }
+
+        private UpdateByOperator makeRollingFormulaOperator(@NotNull final MatchPair pair,
+                @NotNull final Table source,
+                @NotNull final RollingFormulaSpec rs) {
+            // noinspection rawtypes
+            final ColumnSource columnSource = source.getColumnSource(pair.rightColumn);
+            final Class<?> csType = columnSource.getType();
+
+            final String[] affectingColumns;
+            if (rs.revWindowScale().timestampCol() == null) {
+                affectingColumns = new String[] {pair.rightColumn};
+            } else {
+                affectingColumns = new String[] {rs.revWindowScale().timestampCol(), pair.rightColumn};
+            }
+
+            final long prevWindowScaleUnits = rs.revWindowScale().getTimeScaleUnits();
+            final long fwdWindowScaleUnits = rs.fwdWindowScale().getTimeScaleUnits();
+
+            if (csType == char.class || csType == Character.class) {
+                return new CharRollingFormulaOperator(pair, affectingColumns, rowRedirection,
+                        rs.revWindowScale().timestampCol(),
+                        prevWindowScaleUnits, fwdWindowScaleUnits, rs.formula(), rs.paramToken(), columnSource);
+            }
+
+            if (csType == Boolean.class || csType == boolean.class) {
+                return new ByteRollingFormulaOperator(pair, affectingColumns, rowRedirection,
+                        rs.revWindowScale().timestampCol(),
+                        prevWindowScaleUnits, fwdWindowScaleUnits, rs.formula(), rs.paramToken(), columnSource,
+                        NULL_BOOLEAN_AS_BYTE);
+            } else if (csType == byte.class || csType == Byte.class) {
+                return new ByteRollingFormulaOperator(pair, affectingColumns, rowRedirection,
+                        rs.revWindowScale().timestampCol(),
+                        prevWindowScaleUnits, fwdWindowScaleUnits, rs.formula(), rs.paramToken(), columnSource,
+                        NULL_BYTE);
+            } else if (csType == char.class || csType == Character.class) {
+                return new CharRollingFormulaOperator(pair, affectingColumns, rowRedirection,
+                        rs.revWindowScale().timestampCol(),
+                        prevWindowScaleUnits, fwdWindowScaleUnits, rs.formula(), rs.paramToken(), columnSource);
+            } else if (csType == short.class || csType == Short.class) {
+                return new ShortRollingFormulaOperator(pair, affectingColumns, rowRedirection,
+                        rs.revWindowScale().timestampCol(),
+                        prevWindowScaleUnits, fwdWindowScaleUnits, rs.formula(), rs.paramToken(), columnSource);
+            } else if (csType == int.class || csType == Integer.class) {
+                return new IntRollingFormulaOperator(pair, affectingColumns, rowRedirection,
+                        rs.revWindowScale().timestampCol(),
+                        prevWindowScaleUnits, fwdWindowScaleUnits, rs.formula(), rs.paramToken(), columnSource);
+            } else if (csType == long.class || csType == Long.class) {
+                return new LongRollingFormulaOperator(pair, affectingColumns, rowRedirection,
+                        rs.revWindowScale().timestampCol(),
+                        prevWindowScaleUnits, fwdWindowScaleUnits, rs.formula(), rs.paramToken(), columnSource);
+            } else if (csType == float.class || csType == Float.class) {
+                return new FloatRollingFormulaOperator(pair, affectingColumns, rowRedirection,
+                        rs.revWindowScale().timestampCol(),
+                        prevWindowScaleUnits, fwdWindowScaleUnits, rs.formula(), rs.paramToken(), columnSource);
+            } else if (csType == double.class || csType == Double.class) {
+                return new DoubleRollingFormulaOperator(pair, affectingColumns, rowRedirection,
+                        rs.revWindowScale().timestampCol(),
+                        prevWindowScaleUnits, fwdWindowScaleUnits, rs.formula(), rs.paramToken(), columnSource);
+            }
+            return new ObjectRollingFormulaOperator(pair, affectingColumns, rowRedirection,
+                    rs.revWindowScale().timestampCol(),
+                    prevWindowScaleUnits, fwdWindowScaleUnits, rs.formula(), rs.paramToken(), columnSource);
+        }
+
     }
 }
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateByOperatorFactory.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateByOperatorFactory.java
@@ -9,6 +9,7 @@ import io.deephaven.api.updateby.spec.*;
 import io.deephaven.engine.table.ColumnSource;
 import io.deephaven.engine.table.impl.MatchPair;
 import io.deephaven.engine.table.Table;
+import io.deephaven.engine.table.impl.select.FormulaColumn;
 import io.deephaven.engine.table.impl.updateby.em.*;
 import io.deephaven.engine.table.impl.updateby.emstd.*;
 import io.deephaven.engine.table.impl.updateby.delta.*;
@@ -525,11 +526,12 @@ public class UpdateByOperatorFactory {
             final boolean isTimeBased = spec.revWindowScale().isTimeBased();
             final String timestampCol = spec.revWindowScale().timestampCol();
 
+            // These operators can re-use formula columns when the types match.
+            final Map<Class<?>, FormulaColumn> formulaColumnMap = new HashMap<>();
+
             Arrays.stream(pairs)
                     .filter(p -> !isTimeBased || !p.rightColumn().equals(timestampCol))
-                    .map(fc -> makeRollingFormulaOperator(fc,
-                            source,
-                            spec))
+                    .map(fc -> makeRollingFormulaOperator(fc, source, formulaColumnMap, spec))
                     .forEach(ops::add);
             return null;
         }
@@ -1359,6 +1361,7 @@ public class UpdateByOperatorFactory {
 
         private UpdateByOperator makeRollingFormulaOperator(@NotNull final MatchPair pair,
                 @NotNull final Table source,
+                @NotNull final Map<Class<?>, FormulaColumn> formulaColumnMap,
                 @NotNull final RollingFormulaSpec rs) {
             // noinspection rawtypes
             final ColumnSource columnSource = source.getColumnSource(pair.rightColumn);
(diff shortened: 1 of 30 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateByOperatorFactory.java
```diff
@@ -519,6 +521,21 @@ public Void visit(@NotNull final RollingCountSpec spec) {
             return null;
         }
 
+        @Override
+        public Void visit(@NotNull final RollingFormulaSpec spec) {
+            final boolean isTimeBased = spec.revWindowScale().isTimeBased();
+            final String timestampCol = spec.revWindowScale().timestampCol();
```
devinrsmith, line 527: Probably something that should have been brought up in earlier review... but, do we expect `WindowScale#isTimeBased` and `WindowScale#timestampCol` to be the same for rev and fwd? If so, I'm assuming we should add that check to `RollingOpSpec` and potentially elevate isTimeBased / timestampCol helpers to RollingOpSpec? (This might effect construction and use of @Default)

table-api/src/main/java/io/deephaven/api/updateby/spec/RollingFormulaSpec.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+@Immutable
+@BuildableStyle
+public abstract class RollingFormulaSpec extends RollingOpSpec {
+
+    @Value.Parameter
+    public abstract String formula();
+
+    @Value.Parameter
+    public abstract String paramToken();
```
devinrsmith, line 20: `@Parameter` is only applicable when using `@SimpleStyle`.

devinrsmith (commented):
Didn't dig into the operation specifics.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateByOperatorFactory.java
lbooker42: We don't do this currently because we don't require the user to supply a fwd window scale.  A default is provided of `WindowScale.ofTicks(0)` and we use only the `timestampCol()` from rev window scale (which is always required).

If we enforce this check, we need also to rewrite all `Rolling` op calls that skip the `fwd` parameter to construct a new `WindowScale` object with the matching timestamp column name.  This might be a good thing, but it isn't a trivial change.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingformula/BaseRollingFormulaOperator.java
```diff
@@ -0,0 +1,291 @@
...
+
+        final String outputColumnName = pair.leftColumn;
+
+        final Class<?> columnType = tableDef.getColumn(pair.rightColumn).getDataType();
+        vectorType = getVectorType(columnType);
+
+        // Handle the rare (and probably not useful) case where the formula is an identity formula. We need to make
+        // a copy of the RingBuffer wrapper and store that as a DirectVector. If not, we will point to the live data
+        // in the ring.
+        final String formulaToUse = formula.equals(paramToken) ? formula + ".getDirect()" : formula;
```
nbauernfeind, line 110: This isn't the only way to give back the input value; you can imagine doing so only conditionally (like returning the input value if it doesn't contain any null values).

If you really want to handle this, it would be better to do this in the chunk setter, like so:
```java
        return i -> {
            Object result = formulaOutputSource.get(0);
            if (result instanceof Vector) {
                // Handle the rare (and probably not useful) case where the formula is an identity formula. We need to
                // make a copy of the RingBuffer wrapper and store that as a DirectVector. If not, we will point to the
                // live data in the ring.
                result = ((Vector<?>) result).getDirect();
            }
            valueChunk.asWritableObjectChunk().set(i, result);
        };
```

It might be ideal to create a marker interface for `RingBufferVectorWrapper` and check `instanceof` on that (instead of `Vector` as in my example) to guarantee the call to `directVector` is happening only on the operation's input vector.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingformula/BaseRollingFormulaOperator.java
```diff
@@ -0,0 +1,291 @@
...
+
+        // Re-use the formula column if it's already been created for this type. No need to synchronize; these
+        // operators are created serially.
+        // TODO: does generic Object need to be handled uniquely?
```
nbauernfeind, line 114: I can't think of a reason why an `Object` based formula column couldn't be memoized. The user will need to handle type checking on their own, and each `Object` column will need the same formula to be invoked anyway.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingformula/BaseRollingFormulaOperator.java
```diff
@@ -0,0 +1,291 @@
...
+            @NotNull final TableDefinition tableDef) {
+        super(pair, affectingColumns, timestampColumnName, reverseWindowScaleUnits, forwardWindowScaleUnits, true);
+        this.formulaColumnMap = formulaColumnMap;
+        this.tableDef = tableDef;
+
+        final Class<?> columnType = tableDef.getColumn(pair.rightColumn).getDataType();
+        this.vectorType = vectorType;
+
+        // Re-use the formula column that already been created for this type.
+        // TODO: does generic Object need to be handled uniquely?
```
nbauernfeind, line 143: Don't forget to remove this TODO (I believe it only applies to the above constructor anyway).

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingformula/BaseRollingFormulaOperator.java
```diff
@@ -0,0 +1,291 @@
...
+            maybeInnerSource = null;
+            // region create-sparse
+            outputSource = SparseArrayColumnSource.getSparseMemoryColumnSource(0, formulaColumn.getReturnedType());
+            // endregion create-sparse
+        }
+        outputChunkType = outputSource.getChunkType();
+        nullValue = getNullValue(outputChunkType);
+    }
+
+    private static Class<?> getVectorType(final Class<?> type) {
```
nbauernfeind, line 168: Note that `DhFormulaColumn#getVectorType` checks for boxed primitives; you may want to ensure that `Long` source columns still use `LongRollingFormulaOperator` and not the `Object` impl.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingformula/BaseRollingFormulaOperator.java
```diff
@@ -0,0 +1,291 @@
...
+        if (chunkType == ChunkType.Long) {
+            return (o, i) -> valueChunk.asWritableLongChunk().set(i, o == null ? NULL_LONG : (long) o);
+        }
+        if (chunkType == ChunkType.Short) {
+            return (o, i) -> valueChunk.asWritableShortChunk().set(i, o == null ? NULL_SHORT : (short) o);
+        }
+        return (o, i) -> valueChunk.asWritableObjectChunk().set(i, o);
+    }
+
+    private static Object getNullValue(final ChunkType chunkType) {
```
nbauernfeind, line 225: rm in favor of `outputValues.fillWithNullValue`

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingformula/CharRollingFormulaOperator.java
```diff
@@ -0,0 +1,205 @@
...
+
+            int pushIndex = 0;
+
+            // chunk processing
+            for (int ii = 0; ii < len; ii++) {
+                final int pushCount = pushChunk.get(ii);
+                final int popCount = popChunk.get(ii);
+
+                if (pushCount == NULL_INT) {
+                    writeNullToOutputChunk(ii);
```
nbauernfeind, line 97: prefer `outputValues.fillWithNullValue(ii, 1);`

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingformula/CharRollingFormulaOperator.java
```diff
@@ -0,0 +1,205 @@
...
+
+            for (int ii = 0; ii < count; ii++) {
+                charWindowValues.removeUnsafe();
+            }
+        }
+
+        @Override
+        public void writeToOutputChunk(int outIdx) {
+            // If not empty (even if completely full of null), run the formula over the window values.
+            outputSetter.accept(formulaOutputSource.get(0), outIdx);
```
nbauernfeind, line 142: I suggested saving the `outputSetter` in this `Context` sub-class as an `IntConsumer` fetched via `getChunkSetter` passing in the output source so it can directly invoke `getChar` or `getInt` etc without needing to box.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingformula/ObjectRollingFormulaOperator.java
```diff
@@ -0,0 +1,195 @@
...
+
+            for (int ii = 0; ii < count; ii++) {
+                windowValues.removeUnsafe();
+            }
+        }
+
+        @Override
+        public void writeToOutputChunk(int outIdx) {
+            // If not empty (even if completely full of null), run the formula over the window values.
+            formulaInputSource.set(new ObjectVectorDirect<>(windowValues.getAll()));
```
nbauernfeind, line 139: Any reason to not introduce an `ObjectRingBufferVectorWrapper`? It would save on some garbage generation.

lbooker42: I thought I had a reason, but pretty sure I'm wrong. Good catch.

py/server/deephaven/updateby.py
```diff
@@ -1360,4 +1358,90 @@ def rolling_wavg_time(ts_col: str, wcol: str, cols: Union[str, List[str]], rev_t
...
+        |  `rev_ticks = 11, fwd_ticks = -1` - contains 10 rows, beginning at 10 rows before, ending at 1 row before the
+            current row (inclusive)
+        |  `rev_ticks = -5, fwd_ticks = 10` - contains 5 rows, beginning 5 rows following, ending at 10 rows  following the
+            current row (inclusive)
+
+    Args:
+        formula (str): the user defined formula to apply to each window of data
+        formula_param (str): the parameter name within the formula
+        cols (Union[str, List[str]]): the column(s) to be operated on, can include expressions to rename the output,
+            i.e. "new_col = col"; when empty, update_by perform the rolling formula operation on all columns.
```
chipkent, line 1386:
```suggestion
            i.e. "new_col = col"; when empty, update_by performs the rolling formula operation on all columns.
```

chipkent, line 1386: This signature looks very locked into a single column in / out implementation.  Is the plan to have a multiple in implementation have a different function?  If not, this function signature should be evaluated.

py/server/deephaven/updateby.py
```diff
@@ -1360,4 +1358,90 @@ def rolling_wavg_time(ts_col: str, wcol: str, cols: Union[str, List[str]], rev_t
...
+            current row timestamp (inclusive), this is a purely backwards looking window
+        |  `rev_time = "-PT00:05:00", fwd_time = "PT00:10:00"` - contains rows from 5m following through 10m
+            following the current row timestamp (inclusive), this is a purely forwards looking window
+
+    Args:
+        ts_col (str): the timestamp column for determining the window
+        formula (str): the user defined formula to apply to each window of data
+        formula_param (str): the parameter name within the formula
+        cols (Union[str, List[str]]): the column(s) to be operated on, can include expressions to rename the output,
+            i.e. "new_col = col"; when empty, update_by perform the rolling formula operation on all columns.
```
chipkent, line 1429:
```suggestion
            i.e. "new_col = col"; when empty, update_by performs the rolling formula operation on all columns.
```

py/server/deephaven/updateby.py
```diff
@@ -1360,4 +1358,90 @@ def rolling_wavg_time(ts_col: str, wcol: str, cols: Union[str, List[str]], rev_t
...
+        |  `rev_time = "PT00:10:00", fwd_time = "-PT00:05:00"` - contains rows from 10m before through 5m before the
+            current row timestamp (inclusive), this is a purely backwards looking window
+        |  `rev_time = "-PT00:05:00", fwd_time = "PT00:10:00"` - contains rows from 5m following through 10m
+            following the current row timestamp (inclusive), this is a purely forwards looking window
+
+    Args:
+        ts_col (str): the timestamp column for determining the window
+        formula (str): the user defined formula to apply to each window of data
+        formula_param (str): the parameter name within the formula
+        cols (Union[str, List[str]]): the column(s) to be operated on, can include expressions to rename the output,
```
chipkent, line 1428: See the comment above about multiple input formulas.

py/server/tests/test_updateby.py
```diff
@@ -154,6 +153,14 @@ def setUpClass(cls) -> None:
...
             rolling_wavg_time(ts_col="Timestamp", wcol="b", cols=["rwavg_b = b", "rwavg_e = e"], rev_time=10_000_000_000, fwd_time=-10_000_000_00),
             rolling_wavg_time(ts_col="Timestamp", wcol="b", cols=["rwavg_b = b", "rwavg_e = e"], rev_time="PT30S", fwd_time="-PT00:00:20"),
+            # rolling formula
+            rolling_formula_tick(formula="sum(x)", formula_param="x", cols=["formula_a = a", "formula_d = d"], rev_ticks=10),
+            rolling_formula_tick(formula="avg(x)", formula_param="x", cols=["formula_a = a", "formula_d = d"], rev_ticks=10, fwd_ticks=10),
+            rolling_formula_time(formula="sum(x)", formula_param="x", ts_col="Timestamp", cols=["formula_b = b", "formula_e = e"], rev_time="PT00:00:10"),
+            rolling_formula_time(formula="avg(x)", formula_param="x", ts_col="Timestamp", cols=["formula_b = b", "formula_e = e"], rev_time=10_000_000_000,
+                             fwd_time=-10_000_000_00),
+            rolling_formula_time(formula="sum(x)", formula_param="x", ts_col="Timestamp", cols=["formula_b = b", "formula_e = e"], rev_time="PT30S",
```
chipkent, line 162: The `formula_param` variable makes me wonder what is going on under the covers.  Might there be a regex to fool?  How would something like this behave?

```python
 rolling_formula_tick(formula="sum(u)", formula_param="u", cols=["formula_a = a", "formula_d = d"], rev_ticks=10),
```

py/server/deephaven/updateby.py
lbooker42: This matches the `AggBy` `Formula` syntax which was the primary goal of this PR.  The multiple column syntax is TBD for both `Agg` and `UpdateBy` formula operators.

py/server/tests/test_updateby.py
lbooker42: The replacing call is `io.deephaven.engine.table.impl.select.FormulaUtil#replaceFormulaTokens` which doesn't leverage regex but does work hard to avoid over-replacements.  I tested `formula="sum(x)", formula_param="x"` and it handles it correctly.

lbooker42: Also, providing aliases for the input columns  (`formula_a = a`) is handled properly.  The output column (after the formula is applied) is properly called `formula_a`

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingformula/BooleanRollingFormulaOperator.java
```diff
@@ -0,0 +1,190 @@
...
+
+            for (int ii = 0; ii < count; ii++) {
+                final byte val = influencerValuesChunk.get(pos + ii);
+                windowValues.addUnsafe(val == NULL_BYTE ? null : val != 0);
+            }
+        }
+
+        @Override
+        public void pop(int count) {
+            Assert.geq(windowValues.size(), "byteWindowValues.size()", count);
```
nbauernfeind, line 136: variable name in string is mismatched

nbauernfeind (dismissed):

rcaudy (approved):

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingformula/BaseRollingFormulaOperator.java
```diff
@@ -0,0 +1,271 @@
...
+            outputSource = SparseArrayColumnSource.getSparseMemoryColumnSource(0, formulaColumn.getReturnedType());
+            // endregion create-sparse
+        }
+
+        primitiveOutputSource = ReinterpretUtils.maybeConvertToWritablePrimitive(outputSource);
+
+        outputChunkType = primitiveOutputSource.getChunkType();
+    }
+
+    private static Class<?> getVectorType(final Class<?> type) {
```
rcaudy, line 164: I think you want to use `io.deephaven.vector.VectorFactory.forElementType()`, which gives you a `vectorType()`.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingformula/BaseRollingFormulaOperator.java
```diff
@@ -0,0 +1,271 @@
...
+    protected static IntConsumer getChunkSetter(
+            final WritableChunk<? extends Values> valueChunk,
+            final ColumnSource<?> formulaOutputSource) {
+        final ChunkType chunkType = valueChunk.getChunkType();
+        if (chunkType == ChunkType.Boolean) {
+            throw new IllegalStateException(
+                    "Output chunk type should not be Boolean but should have been reinterpreted to byte");
+        }
+        if (chunkType == ChunkType.Byte) {
+            return i -> valueChunk.asWritableByteChunk().set(i, formulaOutputSource.getByte(0));
```
rcaudy, line 198: I would do the cast just once, and define the lambda over the result.

table-api/src/main/java/io/deephaven/api/updateby/UpdateByOperation.java
```diff
@@ -1906,6 +1906,146 @@ static UpdateByOperation RollingWAvg(String timestampCol, long revTime, long fwd
...
+     * <li>{@code revTicks = -5, fwdTicks = 10} - contains 5 rows, beginning 5 rows following, ending at 10 rows
+     * following the current row (inclusive)</li>
+     * </ul>
+     *
+     * @param revTicks the look-behind window size (in rows/ticks)
+     * @param fwdTicks the look-ahead window size (in rows/ticks)
+     * @param pairs The input/output column name pairs
+     * @return The aggregation
+     */
+    static UpdateByOperation RollingFormula(long revTicks, long fwdTicks, String formula, String paramToken,
```
rcaudy, line 1952: JavaDoc for formula and paramToken. Maybe improve pyDoc for paramToken.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingformula/ringbuffervectorwrapper/CharRingBufferVectorWrapper.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    public CharVector subVector(long fromIndexInclusive, long toIndexExclusive) {
+        throw new UnsupportedOperationException("subVector not supported on CharRingBufferVectorWrapper");
+    }
+
+    @Override
+    public CharVector subVectorByPositions(long[] positions) {
+        throw new UnsupportedOperationException("subVectorByPositions not supported on CharRingBufferVectorWrapper");
```
rcaudy, line 31: Can use the existing slice and subvector classes to implement this. Not insanely efficient, but works...

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/UpdateByOperator.java
```diff
@@ -126,20 +129,32 @@ public abstract void accumulateRolling(RowSequence inputKeys,
...
+     *
+     * @return a copy of this operator
+     */
+    public abstract UpdateByOperator copy();
+
+    /**
+     * Initialize this operator with a specific source table (and row redirection if needed). This will be called
+     * exactly once per operator.
+     */
+    public abstract void initializeSources(@NotNull final Table source, @Nullable final RowRedirection rowRedirection);
```
rcaudy, line 156:
```suggestion
    public abstract void initializeSources(@NotNull Table source, @Nullable RowRedirection rowRedirection);
```

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/rollingformula/BooleanRollingFormulaOperator.java
```diff
@@ -0,0 +1,190 @@
...
+            writeToOutputColumn(inputKeys);
+        }
+
+        @Override
+        public void push(int pos, int count) {
+            windowValues.ensureRemaining(count);
+
+            for (int ii = 0; ii < count; ii++) {
+                final byte val = influencerValuesChunk.get(pos + ii);
+                windowValues.addUnsafe(val == NULL_BYTE ? null : val != 0);
```
rcaudy, line 130: `io.deephaven.util.BooleanUtils#byteAsBoolean`

py/client/pydeephaven/agg.py
```diff
@@ -172,11 +172,15 @@ def first(cols: Union[str, List[str]] = None) -> Aggregation:
...
+        |  Built-in functions such as `min`, `max`, etc.
+        |  Mathematical arithmetic such as `*`, `+`, `/`, etc.
+        |  User-defined functions
 
     Args:
-        formula (str): the user defined formula to apply to each group
-        formula_param (str): the parameter name within the formula
+        formula (str): the user defined formula to apply to each group.
+        formula_param (str): the parameter name within the formula. If formula is `max(each)`, then `each` is the
+            formula_param.
```
rcaudy, line 183:
```suggestion
        formula_param (str): the parameter name for the input column's vector within the formula. If formula is `max(each)`, then `each` is the
            formula_param. 
```

table-api/src/main/java/io/deephaven/api/updateby/UpdateByOperation.java
```diff
@@ -1946,7 +1946,16 @@ static UpdateByOperation RollingFormula(long revTicks, String formula, String pa
...
      * @param fwdTicks the look-ahead window size (in rows/ticks)
-     * @param pairs The input/output column name pairs
+     * @param formula the user-defined formula to apply to each group. This formula can contain any combination of the
+     *        following:
+     *        <ul>
+     *        <li>Built-in functions such as min, max, etc.</li>
+     *        <li>Mathematical arithmetic such as *, +, /, etc.</li>
+     *        <li>User-defined functions</li>
+     *        </ul>
+     * @param paramToken the parameter name within the formula. If formula is <em>max(each)</em>, then <em>each</em> is
```
rcaudy, line 1956:
```suggestion
     * @param paramToken the parameter name for the input column's vector within the formula. If formula is <em>max(each)</em>, then <em>each</em> is
```

table-api/src/main/java/io/deephaven/api/updateby/UpdateByOperation.java
```diff
@@ -1968,6 +1977,15 @@ static UpdateByOperation RollingFormula(long revTicks, long fwdTicks, String for
...
      * @param timestampCol the name of the timestamp column
      * @param revDuration the look-behind window size (in Duration)
+     * @param formula the user-defined formula to apply to each group. This formula can contain any combination of the
+     *        following:
+     *        <ul>
+     *        <li>Built-in functions such as min, max, etc.</li>
+     *        <li>Mathematical arithmetic such as *, +, /, etc.</li>
+     *        <li>User-defined functions</li>
+     *        </ul>
+     * @param paramToken the parameter name within the formula. If formula is <em>max(each)</em>, then <em>each</em> is
```
rcaudy, line 1987:
```suggestion
     * @param paramToken the parameter name for the input column's vector within the formula. If formula is <em>max(each)</em>, then <em>each</em> is
```

rcaudy (dismissed):

py/client/pydeephaven/agg.py
```diff
@@ -172,11 +172,15 @@ def first(cols: Union[str, List[str]] = None) -> Aggregation:
...
+        |  Built-in functions such as `min`, `max`, etc.
+        |  Mathematical arithmetic such as `*`, `+`, `/`, etc.
+        |  User-defined functions
 
     Args:
-        formula (str): the user defined formula to apply to each group
-        formula_param (str): the parameter name within the formula
+        formula (str): the user defined formula to apply to each group.
+        formula_param (str): the parameter name for the input column's vector within the formula. If formula is
+            `max(each)`, then `each` is the formula_param.
```
chipkent, line 183: I agree that these changes are an improvement.  They should be incorporated into the functions in `updateby.py`.

rcaudy (dismissed):

py/server/tests/test_updateby.py
chipkent: great.

py/server/deephaven/updateby.py
chipkent: Sounds good.

py/server/deephaven/updateby.py
```diff
@@ -417,7 +417,7 @@ def emstd_tick(decay_ticks: float, cols: Union[str, List[str]],
     Args:
         decay_ticks (float): the decay rate in ticks
         cols (Union[str, List[str]]): the column(s) to be operated on, can include expressions to rename the output,
-            i.e. "new_col = col"; when empty, update_by perform the ems operation on all columns.
+            i.e. "new_col = col"; when empty, update_by performs the ems operation on all columns.
```
chipkent, line 420: these changes also need to be in the client.

deephaven-internal:

Community: https://github.com/deephaven/deephaven.io/issues/3699

PR #5089, opened 2024-01-31
Fix ReadableByteChannel.read usages

### LINKED ISSUE #5088: Helpers.readBytes assumes ReadableByteChannel#read will always read fully
This was caused during exploration of the new s3 feature, https://github.com/deephaven/deephaven-core/pull/4972.

```
caused by java.io.IOException: Expected for bytes, only read 7244 while it expected 38744
        at io.deephaven.parquet.base.util.Helpers.readBytes(Helpers.java:20)
        at io.deephaven.parquet.base.ParquetFileReader.<init>(ParquetFileReader.java:89)
        at io.deephaven.parquet.table.ParquetTools.getParquetFileReaderChecked(ParquetTools.java:1163)
        at io.deephaven.parquet.table.ParquetTools.getParquetFileReader(ParquetTools.java:1132)
        ... 27 more
```

This was easy to reproduce using a small fragment size `io.deephaven.extensions.s3.S3Instructions#fragmentSize`. While not caused by this feature, we should aim to fix this to make sure the new feature works as expected.

devinrsmith:
Fixes #5088

### DIFF
--- a/extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
+++ b/extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
@@ -158,7 +158,11 @@ public class ArrowWrapperTools {
 
                 final ByteBuffer metadataBuf = ByteBuffer.wrap(rawMetadataBuf, 0, block.getMetadataLength());
                 while (metadataBuf.hasRemaining()) {
-                    if (channel.read(metadataBuf) == -1) {
+                    final int read = channel.read(metadataBuf);
+                    if (read == 0) {
+                        throw new IllegalStateException("ReadableByteChannel.read returned 0");
+                    }
+                    if (read == -1) {
                         throw new IOException(
                                 "Unexpected end of input trying to read block " + ii + " of '" + path + "'");
                     }
--- a/extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
+++ b/extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
@@ -157,11 +157,12 @@ public class ArrowWrapperTools {
                 channel.position(block.getOffset());
 
                 final ByteBuffer metadataBuf = ByteBuffer.wrap(rawMetadataBuf, 0, block.getMetadataLength());
-                int numRead = channel.read(metadataBuf);
-                if (numRead != block.getMetadataLength()) {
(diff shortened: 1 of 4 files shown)


malhotrashivam (approved):

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
```diff
@@ -157,11 +157,12 @@ public static QueryTable readFeather(@NotNull final String path) {
                 channel.position(block.getOffset());
 
                 final ByteBuffer metadataBuf = ByteBuffer.wrap(rawMetadataBuf, 0, block.getMetadataLength());
-                int numRead = channel.read(metadataBuf);
-                if (numRead != block.getMetadataLength()) {
-                    throw new IOException("Unexpected end of input trying to read block " + ii + " of '" + path + "'");
+                while (metadataBuf.hasRemaining()) {
+                    if (channel.read(metadataBuf) == -1) {
```
nbauernfeind, line 161: We should probably throw if `<= 0` to ensure that each call is making progress.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/util/Helpers.java
```diff
@@ -15,22 +15,26 @@
...
-        }
+        readExact(f, buffer);
         buffer.flip();
         return BytesInput.from(buffer);
     }
 
+    public static void readExact(ReadableByteChannel f, ByteBuffer buffer) throws IOException {
+        final int expected = buffer.remaining();
+        while (buffer.hasRemaining()) {
+            if (f.read(buffer) == -1) {
```
nbauernfeind, line 31: Same; probably should complain if we aren't making progress by checking `<= 0`.

nbauernfeind (approved):
I won't throw my hands in the air if you don't like the suggestion or feel that it would not be correct.

extensions/arrow/src/main/java/io/deephaven/extensions/arrow/ArrowWrapperTools.java
devinrsmith: It's a valid point. Assuming the channel is in blocking mode, we _can_ throw an error on 0 (which _shouldn't_ occur)

> It is guaranteed, however, that if a channel is in blocking mode and there is at least one byte remaining in the buffer then this method will block until at least one byte is read.

I'll make the change.

nbauernfeind (approved):

PR #5100, opened 2024-02-01
Add support of 1D arrays in UDF vectorization

### LINKED ISSUE #4593: Python vectorization should support parameters of array types.
See https://numba.pydata.org/numba-doc/dev/user/vectorize.html#the-guvectorize-decorator.

The [guvectorize()](https://numba.pydata.org/numba-doc/dev/reference/jit-compilation.html#numba.guvectorize) decorator takes the concept one step further and allows you to write ufuncs that will work on an arbitrary number of elements of input arrays, and take and return arrays of differing dimensions.

But the vectorization implementation currently doesn't support non scalar types and would throw an exception to cause the query to fail. 
Two things might lead to lifting this restriction:
1. the ability to pass Java arrays (2D or nD) as usable Python data - i.e. numpy arrays/Pandas serires, and the reverse
2. the ability to fill a chunk of array column (nD),  not as an array of object,  but as n + 1 dimension arrays

### LINKED ISSUE #4649: Support of vectorizing functions that return arrays
Currently if a Python function returns 1D arrays of primitive types and String, Instant, it is vectorizable due to:
1. they are treated as generic object chunks when vectorized 
2. they are not converted to usable Java arrays

jmao-denver:
Fixes #4593 
Fixes #4649

### DIFF
--- a/py/server/tests/test_vectorization.py
+++ b/py/server/tests/test_vectorization.py
@@ -4,7 +4,7 @@
 import random
 import unittest
 
-from typing import Optional
+from typing import Optional, Union
 import numpy as np
 
 from deephaven import DHError, empty_table, dtypes
@@ -14,6 +14,8 @@ from deephaven.filters import Filter, and_
 import deephaven._udf as _udf
 from deephaven._udf import _dh_vectorize as dh_vectorize
 from tests.testbase import BaseTestCase
+
+from tests.test_udf_numpy_args import _J_TYPE_NULL_MAP, _J_TYPE_NP_DTYPE_MAP, _J_TYPE_J_ARRAY_TYPE_MAP
 
 
 class VectorizationTestCase(BaseTestCase):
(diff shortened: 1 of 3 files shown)


py/server/tests/test_vectorization.py
```diff
@@ -278,6 +280,86 @@ def pyfunc(p1: np.int32, p2: np.int32, p3: Optional[np.int32]) -> Optional[int]:
...
+                    with self.assertRaises(DHError) as cm:
+                        tbl.update("Col3 = test_udf(Col1, Col2)")
+
+                self.assertEqual(_udf.vectorized_count, 1)
+                _udf.vectorized_count = 0
+
+    def test_1d_str_bool_datetime_array(self):
+        with self.subTest("str"):
+            def f1(p1: np.ndarray[str]) -> bool:
+                return (p1 == 'None').any()
```
chipkent, line 335: Is this right?  It looks like it is encoding the null `None` as the string `"None"`.

py/server/tests/test_vectorization.py
```diff
@@ -278,6 +280,86 @@ def pyfunc(p1: np.int32, p2: np.int32, p3: Optional[np.int32]) -> Optional[int]:
...
+            _udf.vectorized_count = 0
+
+        with self.subTest("boolean"):
+            def f3(p1: np.ndarray[np.bool_]) -> np.ndarray[np.bool_]:
+                return np.invert(p1)
+
+            t = empty_table(10).update(["X = i % 3", "Y = i % 2 == 0? true : false"]).group_by("X").update("Y = Y.toArray()")
+            t1 = t.update(["X1 = f3(Y)"])
+            self.assertEqual(_udf.vectorized_count, 1)
+            _udf.vectorized_count = 0
```
chipkent, line 362: Seems like there should also be a null boolean test.

py/server/tests/test_vectorization.py
jmao-denver: Yes. NumPy automatically converts Python None to 'None'. Also NumPy string arrays are padded to the same length,

chipkent: I had a hard time beliving that Numpy did something so bad, but they did.  Here is my test to verify:
```python

import numpy as np

sa = np.array(["A", "BC", "DEF", None], dtype="str")

print(sa)

for i in range(len(sa)):
    print(f"{i} {sa[i]} {type(sa[i])}")
```
image

I guess we can't do anything better here.

chipkent (approved):

deephaven-internal:

Community: https://github.com/deephaven/deephaven.io/issues/3680

PR #5105, opened 2024-02-01
[WIP] Added support to write metadata files in parquet

malhotrashivam:


extensions/parquet/base/build.gradle
```diff
@@ -9,6 +9,12 @@ dependencies {
     api project(':util-channel')
 
     Classpaths.inheritParquetHadoop(project)
+    Classpaths.inheritParquetHadoopConfiguration(project)
+
+// TODO These are the additional dependencies that are required for the "fs.file.impl.disable.cache", check
+//  MetadataFileWriter code for more details.
+//    implementation group: 'org.apache.commons', name: 'commons-configuration2', version: '2.0'
+//    implementation group: 'org.apache.hadoop', name: 'hadoop-auth', version: '3.3.3'
```
malhotrashivam, line 18: Remove TODO when done

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/MetadataFileWriter.java
```diff
@@ -0,0 +1,80 @@
...
+ * provided Parquet files.
+ */
+public final class MetadataFileWriter implements MetadataFileWriterBase {
+    // TODO Please suggest a better name for this class and the interface
+
+    private final String metadataRootDir;
+    private final List<Footer> footers;
+
+    public MetadataFileWriter(final String metadataRootDir, final File[] destinations) {
+        // TODO Discuss with Ryan about this check
```
malhotrashivam, line 27: Remove TODO when done

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/MetadataFileWriter.java
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public void addFooter(final File parquetFile, final ParquetMetadata parquetMetadata) {
+        footers.add(new Footer(new Path(parquetFile.getAbsolutePath()), parquetMetadata));
+    }
+
+    public void writeMetadataFiles() {
+        final Configuration conf = new Configuration();
+        conf.setBoolean("fs.file.impl.disable.cache", true);
+
+        // TODO Discuss with Ryan about this setting
```
malhotrashivam, line 55: Remove TODO when done

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -101,30 +106,45 @@ public static void write(
...
                     final File groupingDestFile = entry.getValue().destFile;
                     cleanupFiles.add(groupingDestFile);
                     tableInfoBuilder.addGroupingColumns(GroupingColumnInfo.of(parquetColumnName,
-                            destDirPath.relativize(metadataFilePath.toPath()).toString()));
-                    write(auxiliaryTable, auxiliaryTable.getDefinition(), writeInstructions,
-                            groupingDestFile.getAbsolutePath(), Collections.emptyMap(), TableInfo.builder());
+                            destDirPath.relativize(metadataGroupingFilePath.toPath()).toString()));
+                    // We don't include schema and row group information from grouping files in the metadata files
+
+                    // TODO Remove this comment
```
malhotrashivam, line 133: Remove TODO when done

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -522,12 +536,27 @@ public static void writeParquetTables(@NotNull final Table[] sources,
                     groupingColumnWritingInfoMap.values().forEach(gcwi -> shadowFiles.add(gcwi.destFile));
 
                     final Table sourceTable = sources[tableIdx];
-                    ParquetTableWriter.write(sourceTable, definition, writeInstructions,
-                            shadowDestFiles[tableIdx].getPath(),
-                            Collections.emptyMap(), groupingColumnWritingInfoMap);
+                    ParquetTableWriter.write(sourceTable, definition, writeInstructions, shadowDestFiles[tableIdx],
+                            tableDestination, Collections.emptyMap(), groupingColumnWritingInfoMap, metadataFileWriter);
                 }
             }
```
malhotrashivam, line 543: Remove TODO when done

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -467,6 +468,86 @@ private static void writeReadTableTest(final Table table, final File dest,
...
+                new File[] {firstDataFile, secondDataFile}, null);
+
+        final Table fromDisk = readTable(parentDir);
+        assertTableEquals(source, fromDisk);
+
+        final File metadataFile = new File(rootFile, "_metadata");
+        final Table fromDiskWithMetadata = readTable(metadataFile);
+        assertTableEquals(source, fromDiskWithMetadata);
+    }
+
```
malhotrashivam, line 550: TODO Add test for the failure JJ noticed.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/MetadataFileWriter.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.List;
+
+import static org.apache.parquet.hadoop.ParquetFileWriter.writeMetadataFile;
+
+/**
+ * Uses the hadoop implementation of {@link org.apache.parquet.hadoop.ParquetFileWriter} to generate metadata files for
+ * provided Parquet files.
+ */
+public final class MetadataFileWriter implements MetadataFileWriterBase {
+    // TODO Please suggest a better name for this class and the interface
```
malhotrashivam, line 21: Remove TODO when done

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
rcaudy: We need to merge this "manually". Use `io.deephaven.parquet.table.metadata.TableInfo#deserializeFromJSON` to turn values into `TableInfo`, and `io.deephaven.parquet.table.metadata.TableInfo#serializeToJSON` to make new values.

`groupingColumns`, `dataIndexes` should probably always be dropped.
`sortingColumns` and `version` are optional.
`columnTypes` must be the same for all partitions, I think.

rcaudy (commented):
Answered some questions

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
malhotrashivam: I added this logic but after this, adding metadata from grouping files to the overall metadata files has a weird behavior. Consider the following screenshot:

Screenshot 2024-02-16 at 11 11 13 PM

We can see that the combined schema now has columns dh_key, dh_begin_pos and dh_end_pos and the blocks has references to the grouping file (block[0] has path ".dh_metadata/indexes/vvv/index_vvv_parquetWithGroupingDataAndMetadataTest.parquet"). So now if I do a read on the directory, it would go to the metadata file, locate all these blocks (including the ones from the grouping file) and add all these columns to the main table. I can walk you through this on a screen share session as well.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetMetadataFileWriterImpl.java
```diff
@@ -0,0 +1,177 @@
...
+        ParquetFileMetadata(final File file, final ParquetMetadata metadata) {
+            this.file = file;
+            this.metadata = metadata;
+        }
+    }
+
+    /**
+     * A class to accumulate metadata from multiple parquet files.
+     */
+    private static final class AccumulatedMetadata {
```
malhotrashivam, line 49: There are a lot of variables with names related to metadata here, let me know if you want me to rename any of these.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -269,10 +288,12 @@ private static ParquetFileWriter getParquetFileWriter(
...
             @NotNull final RowSet tableRowSet,
             @NotNull final Map<String, ? extends ColumnSource<?>> columnSourceMap,
-            @NotNull final String path,
+            @NotNull final File destFile,
+            @NotNull final File metadataFilePath,
             @NotNull final ParquetInstructions writeInstructions,
             @NotNull final Map<String, String> tableMeta,
-            @NotNull final TableInfo.Builder tableInfoBuilder) throws IOException {
+            @NotNull final TableInfo.Builder tableInfoBuilder,
+            @NotNull final ParquetMetadataFileWriter metadataFileWriter) throws IOException {
```
lbooker42, line 296: Missing Javadoc for `metadataFileWriter`

lbooker42 (commented):
This looks good, a few comments about Javadoc parameters but this is easy to follow.

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -467,6 +468,85 @@ private static void writeReadTableTest(final Table table, final File dest,
...
+        writeTable(table, destFile, writeInstructions);
+
+        final File metadataFile = new File(rootFile, "_metadata");
+        assertTrue(metadataFile.exists());
+        assertTrue(new File(rootFile, "_common_metadata").exists());
+
+        final Table fromDisk = readTable(destFile);
+        assertTableEquals(table, fromDisk);
+
+        final Table fromDiskWithMetadata = readTable(metadataFile);
```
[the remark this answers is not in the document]
malhotrashivam: Calling `readTable()` on metadata files reads the table definition from the metadata file and sets up all the linkages internally (check `ParquetMetadataFileLayout` constructor for more details). When we call `select()` or access any particular row of the table, that is when we load the actual parquet file and data.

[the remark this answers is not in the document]
lbooker42: Great explanation, thanks!

lbooker42 (dismissed):
LGTM

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -456,15 +464,124 @@ private static Map<String, ParquetTableWriter.GroupingColumnWritingInfo> groupin
...
+     *        created.
+     * @param baseName The base name for the individual partitioned tables. For example, a base name of "table" will
+     *        result in files named "partition1/table.parquet", "partition2/table.parquet", etc.
+     * @param writeInstructions Write instructions for customizations while writing
+     */
+    public static void writeKeyValuePartitionedTable(@NotNull final Table sourceTable,
+            @NotNull final File destinationDir,
+            @NotNull final String baseName,
+            @NotNull final ParquetInstructions writeInstructions) {
+        // TODO Should I take an additional write instruction for partitioning column names in case they are different
```
malhotrashivam, line 483: TODO Confirm with @rcaudy

lbooker42: The indexing PR that @rcaudy is working on (#3851) has made a lot of changes to table writing, including the generation of companion index files.  I think this partition-level dividing and writing to disk needs to be harmonized with that PR. 

This looks like good tooling, maybe the indexing PR can/should leverage it to add partition-writing to the other changes.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -456,15 +464,124 @@ private static Map<String, ParquetTableWriter.GroupingColumnWritingInfo> groupin
...
+        }
+        final String[] partitioningColNames = partitioningColumns.stream()
+                .map(ColumnDefinition::getName)
+                .toArray(String[]::new);
+        final PartitionedTable partitionedTable = sourceTable.partitionBy(partitioningColNames);
+        writeKeyValuePartitionedTable(partitionedTable, destinationDir, baseName, writeInstructions);
+    }
+
+    /**
+     * Write a partitioned tables to disk in parquet format with all the {@link PartitionedTable#keyColumnNames() key
```
rcaudy, line 505:
```suggestion
     * Write a partitioned table to disk in parquet format with all the {@link PartitionedTable#keyColumnNames() key
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -456,15 +464,124 @@ private static Map<String, ParquetTableWriter.GroupingColumnWritingInfo> groupin
...
+    public static void writeKeyValuePartitionedTable(@NotNull final PartitionedTable partitionedTable,
+            @NotNull final File destinationDir,
+            @NotNull final String baseName,
+            @NotNull final ParquetInstructions writeInstructions) {
+        final String[] partitioningColumnNames = partitionedTable.keyColumnNames().toArray(String[]::new);
+        final Table keyTable;
+        if (partitionedTable.uniqueKeys()) {
+            keyTable = partitionedTable.table().view(partitioningColumnNames);
+        } else {
+            keyTable = partitionedTable.table().selectDistinct(partitioningColumnNames);
```
rcaudy, line 525: This suggests that you *do* need indexes in a base name template.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -456,15 +464,124 @@ private static Map<String, ParquetTableWriter.GroupingColumnWritingInfo> groupin
...
+        final Collection<File> destinations = new ArrayList<>();
+        keyTable.getRowSet().forAllRowKeys(key -> {
+            final Object[] keyValues = Arrays.stream(partitioningColumnNames)
+                    .map(keyTable::getColumnSource)
+                    .map(colSource -> colSource.get(key))
+                    .toArray();
+            final StringBuilder partitionTableRelativePath = new StringBuilder();
+            for (int i = 0; i < partitioningColumnNames.length; i++) {
+                partitionTableRelativePath.append(File.separator)
+                        .append(partitioningColumnNames[i]).append("=").append(keyValues[i]);
```
rcaudy, line 537: This is using `toString()` to get the string representation of the partitioning column value.
I'm wondering if we want something more explicit, symmetrical with `PartitionParser`, e.g. `PartitionFormatter`.

This gives us two properties:
1. We can guarantee that we'll be able to parse the types we support.
2. We can refuse to write data that we won't be able to read back. This is controversial, maybe.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -456,15 +464,124 @@ private static Map<String, ParquetTableWriter.GroupingColumnWritingInfo> groupin
...
+            }
+            partitionTableRelativePath.append(File.separator).append(baseName).append(".parquet");
+            destinations.add(new File(destinationDir, partitionTableRelativePath.toString()));
+            constituentTables.add(partitionedTable.constituentFor(keyValues));
```
rcaudy, line 541: This may break if we hit the `!uniqueKeys()` case.
Maybe we want to just "assume the worst" and use:
```
        Table withGroupConstituents = partitionedTable.table().groupBy(partitionedTable.keyColumnNames().toArray(String[]::new));
```
So we can operate on `ObjectVector<? extends Table>`.

We should insist on a UUID or an index in the name template if `!uniqueKeys()`.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -456,15 +464,124 @@ private static Map<String, ParquetTableWriter.GroupingColumnWritingInfo> groupin
...
+        // pyspark has names of the form "part.{i}.parquet" and allows passing a naming function which takes an integer
+        // and generates a file name.
+        final List<ColumnDefinition<?>> partitioningColumns = sourceTable.getDefinition().getPartitioningColumns();
+        if (partitioningColumns.isEmpty()) {
+            throw new IllegalArgumentException("Table must have partitioning columns to write partitioned data");
+        }
+        final String[] partitioningColNames = partitioningColumns.stream()
+                .map(ColumnDefinition::getName)
+                .toArray(String[]::new);
+        final PartitionedTable partitionedTable = sourceTable.partitionBy(partitioningColNames);
```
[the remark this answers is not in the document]
malhotrashivam: Yes, keeping keys and not adding them to metadata.
So there's not much advantage of doing "drop keys" because in the other api which takes a partitioned table, we cannot assume that the user has dropped the keys.

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -468,6 +474,361 @@ private static void writeReadTableTest(final Table table, final File dest,
...
+        final File parentDir = new File(rootFile, "testAllPartitioningColumnTypes");
+        final ParquetInstructions writeInstructions = ParquetInstructions.builder()
+                .setMetadataRootDir(parentDir.getAbsolutePath())
+                .build();
+        writeKeyValuePartitionedTable(inputData, parentDir, "data", writeInstructions);
+
+        // Verify that we can read the partition values, but types like LocalDate or LocalTime will be read as strings
+        // Therefore, we cannot compare the tables directly
+        final Table fromDisk = readKeyValuePartitionedTable(parentDir, EMPTY).select();
+
```
malhotrashivam, line 823: Comparing fromDisk with input data leads to this error
```
Expected type of PC4 is byte actual type is int
Expected type of PC5 is short actual type is int
Expected type of PC7 is long actual type is int
Expected type of PC8 is float actual type is double
Expected type of PC10 is class java.math.BigInteger actual type is int
Expected type of PC11 is class java.math.BigDecimal actual type is double
Expected type of PC13 is class java.time.LocalDate actual type is class java.lang.String
Expected type of PC14 is class java.time.LocalTime actual type is class java.lang.String
```

Basically types of columns are not deduced properly, unless we read from the metadata file. Then it works as expected, as shown in the next few statements.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -456,15 +465,244 @@ private static Map<String, ParquetTableWriter.GroupingColumnWritingInfo> groupin
...
+                    partitionedData.add(constituent);
+                    count++;
+                }
+                row++;
+            }
+        }
+        // If needed, generate schema for _common_metadata file from key table
+        final ParquetInstructions updatedWriteInstructions;
+        if (!ParquetInstructions.DEFAULT_METADATA_ROOT_DIR.equals(writeInstructions.getMetadataRootDir())) {
+            final MessageType commonSchema = getSchemaForTable(partitionedTable.table(), definition, writeInstructions);
```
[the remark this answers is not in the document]
malhotrashivam: Found another way of doing this, will discuss during review.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -533,7 +535,11 @@ public static void writeKeyValuePartitionedTable(@NotNull final Table sourceTabl
                 .map(ColumnDefinition::getName)
                 .toArray(String[]::new);
         final PartitionedTable partitionedTable = sourceTable.partitionBy(partitioningColNames);
-        writeKeyValuePartitionedTable(partitionedTable, definition, destinationDir, baseName, writeInstructions);
+        final TableDefinition keyTableDefinition = TableDefinition.of(partitioningColumns);
+        final TableDefinition leafDefinition =
+                getNonKeyTableDefinition(Arrays.asList(partitioningColNames), definition);
```
rcaudy, line 540: Set instead of List?

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ParquetMetadataFileWriter.java
```diff
@@ -0,0 +1,34 @@
+package io.deephaven.parquet.base;
+
+import org.apache.parquet.hadoop.metadata.ParquetMetadata;
+
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Used to write _metadata and _common_metadata files for Parquet.
```
rcaudy, line 9: Maybe use value constants in the JavaDoc.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/PositionedBufferedOutputStream.java
```diff
@@ -5,17 +5,22 @@
...
+public final class PositionedBufferedOutputStream extends BufferedOutputStream {
 
     private final SeekableByteChannel writeChannel;
 
+    public PositionedBufferedOutputStream(final SeekableByteChannel writeChannel) {
+        super(Channels.newOutputStream(writeChannel));
+        this.writeChannel = writeChannel;
+    }
+
     PositionedBufferedOutputStream(final SeekableByteChannel writeChannel, final int size) {
```
rcaudy, line 17: I would make this constructor public, too. Why not have them all on equal footing?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetInstructions.java
```diff
@@ -164,6 +166,11 @@ public final String getColumnNameFromParquetColumnNameOrDefault(final String par
      */
     public abstract boolean isRefreshing();
 
+    /**
+     * @return should we generate _meteadata and _common_metadata files while writing parquet files?
```
rcaudy, line 170: Constants, maybe?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetInstructions.java
```diff
@@ -647,6 +667,23 @@ public Builder setSpecialInstructions(final Object specialInstructions) {
...
         }
 
+        /**
+         * Set whether to generate _metadata and _common_metadata files while writing parquet files.
+         * <ul>
+         * <li>When writing a single parquet file, metadata files are generated in the same parent directory as the
+         * parquet file.
+         * <li>When writing multiple parquet files in a single write call, metadata files are only generated if all
+         * parquet files are written to the same parent directory. The metadata files will also be written to the same
+         * parent directory.
```
rcaudy, line 677: This is inaccurate. If the instructions call for writing metadata files, the code insists that all files be written to the same parent directory.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -90,40 +92,56 @@ static class GroupingColumnWritingInfo {
...
+     *
      * @throws IOException For file writing related errors
      */
     static void write(
             @NotNull final Table t,
             @NotNull final TableDefinition definition,
             @NotNull final ParquetInstructions writeInstructions,
-            @NotNull final String destPathName,
+            @NotNull final File destFile,
+            @NotNull final File metadataFilePath,
```
rcaudy, line 115: This feels like a confusing variable name. Since it feels like "the path to the metadata file" rather than "the real path to the destination". `destFilePathForMetadata`. Whatever you decide, please propagate the convention.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTableWriter.java
```diff
@@ -144,30 +162,33 @@ static void write(
...
+     * @param destFile The destination path
+     * @param metadataFilePath The destination path to store in the metadata files. This can be different from
+     *        {@code destFile} if we are writing the parquet file to a shadow location first since the metadata should
+     *        always hold the accurate path.
      * @param tableMeta A map of metadata values to be stores in the file footer
-     * @param tableInfoBuilder A partially-constructed builder for the metadata object
-     * @throws SchemaMappingException Error creating a parquet table schema for the given table (likely due to
-     *         unsupported types)
+     * @param tableInfoBuilder A partially constructed builder for the metadata object
+     * @param metadataFileWriter The writer for the _metadata and _common_metadata files
```
rcaudy, line 171: Constants.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/PartitionFormatter.java
```diff
@@ -0,0 +1,132 @@
...
+        typeMap.put(Float.class, ForFloat);
+        typeMap.put(Double.class, ForDouble);
+        typeMap.put(BigInteger.class, ForBigInteger);
+        typeMap.put(BigDecimal.class, ForBigDecimal);
+        typeMap.put(Instant.class, ForInstant);
+        typeMap.put(LocalDate.class, ForLocalDate);
+        typeMap.put(LocalTime.class, ForLocalTime);
+    }
+
+    abstract String format(@NotNull Object obj);
```
rcaudy, line 122: I'm not sure we shouldn't have some option for `null` partitioning column values *if* we can make them work symmetrically. Empty? `"null"`?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/PartitionFormatter.java
```diff
@@ -0,0 +1,132 @@
...
+
+    static String formatToString(@NotNull final Object obj) {
```
rcaudy, line 124: We should do a single lookup based on schema, rather than doing the lookup for each value.

`io.deephaven.util.type.TypeUtils#getBoxedType`

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/PartitionFormatter.java
```diff
@@ -0,0 +1,132 @@
...
+import java.time.LocalDate;
+import java.time.LocalTime;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * This class takes an object read from a parquet file and formats it to a string used for generate partitioning column
+ * values.
+ */
+enum PartitionFormatter {
```
rcaudy, line 18: This is *not* inherently Parquet-specific. Since it's in parquet-table, it *could* be in engine-impl to be adjacent to `PartitionParser`.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -56,12 +57,15 @@ private static final class PageState<ATTR extends Any> {
             final long mask,
             @NotNull final ToPage<ATTR, ?> toPage) throws IOException {
         super(pageCache, columnChunkReader, mask, toPage);
-        offsetIndex = columnChunkReader.getOffsetIndex();
+        try (final SeekableChannelContext channelContext =
```
rcaudy, line 60: We should consider passing in a context. Maybe infeasible. We've deferred the loading, at the cost of not doing it "in bulk" with a single context.

Maybe we can defer loading page stores when we create regions?

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -139,6 +163,24 @@ private ChunkPage<ATTR> getPageImpl(@Nullable FillContext fillContext, int pageN
...
+            initialize(fillContext);
+            return getPageContainingImpl(fillContext, rowKey);
+        }
+        // Create a fill context and use it to initialize this class as well as read the page
+        // TODO What should be the chunk capacity here?
+        try (final FillContext fillContext1 = makeFillContext(1, null)) {
+            initialize(fillContext1);
+            return getPageContainingImpl(fillContext1, rowKey);
+        }
+    }
```
malhotrashivam, line 180: Please suggest if there is a better way to do this, also what should I keep the chunk capacity here.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetMetadataFileLayout.java
```diff
@@ -196,15 +200,43 @@ public ParquetMetadataFileLayout(
                     partitions.put(partitionKey, partitionValue);
                 }
             }
-            final ParquetTableLocationKey tlk = new ParquetTableLocationKey(new File(directory, filePathString),
+            final File partitionFile = new File(directory, filePathString);
+            final ParquetTableLocationKey tlk = new ParquetTableLocationKey(partitionFile,
                     partitionOrder.getAndIncrement(), partitions, inputInstructions);
             tlk.setFileReader(metadataFileReader);
-            tlk.setMetadata(metadataFileMetadata);
+            tlk.setMetadata(getParquetMetadataForFile(partitionFile, metadataFileMetadata));
```
malhotrashivam, line 207: I verified that that right metadata is being propagated here, but I don't know any tests which I can write for this, since column type is not partitioning, so not really reading any grouping data.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetMetadataFileWriterImpl.java
```diff
@@ -0,0 +1,286 @@
...
+            // TODO How should I change the basename in the API for writeKeyValuePartitioned data for this check?
+            if (i > 0) {
+                // We use filename to generate the key for each file's metadata in the common metadata file, therefore
+                // all files must have unique names.
+                if (destination.getName().equals(firstDestination.getName())) {
+                    throw new UncheckedDeephavenException("When generating common metadata for multiple parquet files, "
+                            + "all files must have unique names, but " + destination.getName() + " is repeated.");
+                }
+            }
+        }
```
malhotrashivam, line 100: I have added this check here for unique names, but now I think its better to just enforce a {uuid} in the file name for partitioned writing. Do you have any suggestions for this?

py/server/deephaven/parquet.py
```diff
@@ -280,6 +280,90 @@ def write(
         raise DHError(e, "failed to write to parquet data.") from e
 
 
+def write_key_value_partitioned_table(
```
malhotrashivam, line 283: Is this method name okay?

jmao-denver: It is OK because it matches the Java one? Or maybe just `write_partitioned_table` because that's the only/default way to write out a PartitionedTabled?

py/server/deephaven/parquet.py
```diff
@@ -43,6 +43,7 @@ def _build_parquet_instructions(
     is_refreshing: bool = False,
     for_read: bool = True,
     force_build: bool = False,
+    generate_metadata_files: Optional[bool] = False,
```
chipkent, line 46: Are you sure this shouldn't be:
```suggestion
    generate_metadata_files: Optional[bool] = None,
```
As is, you will rarely get in the `if not any` code block, which I suspect is your target block.

py/server/deephaven/parquet.py
```diff
@@ -234,6 +239,7 @@ def write(
     max_dictionary_keys: Optional[int] = None,
     max_dictionary_size: Optional[int] = None,
     target_page_size: Optional[int] = None,
+    generate_metadata_files: Optional[bool] = False,
```
chipkent, line 242: Same here
```suggestion
    generate_metadata_files: Optional[bool] = None,
```

py/server/deephaven/parquet.py
```diff
@@ -248,6 +254,7 @@ def write(
         max_dictionary_keys (Optional[int]): the maximum dictionary keys allowed, if not specified, defaults to 2^20 (1,048,576)
         max_dictionary_size (Optional[int]): the maximum dictionary size (in bytes) allowed, defaults to 2^20 (1,048,576)
         target_page_size (Optional[int]): the target page size in bytes, if not specified, defaults to 2^20 bytes (1 MiB)
+        generate_metadata_files (Optional[bool]): whether to generate _metadata and _common_metadata files, default is False
```
chipkent, line 257: 1. Default value.
2. This docstring is not useful to a user.  It needs to explain what these files are.

py/server/deephaven/parquet.py
chipkent: Matching the java name is not the key objective.  It needs to be clear and concise first.

py/server/deephaven/parquet.py
```diff
@@ -280,6 +288,93 @@ def write(
...
+        table: Union[Table, PartitionedTable],
+        destination_dir: str,
+        col_definitions: Optional[List[Column]] = None,
+        col_instructions: Optional[List[ColumnInstruction]] = None,
+        compression_codec_name: Optional[str] = None,
+        max_dictionary_keys: Optional[int] = None,
+        max_dictionary_size: Optional[int] = None,
+        target_page_size: Optional[int] = None,
+        base_name: Optional[str] = None,
+        generate_metadata_files: Optional[bool] = False,
```
chipkent, line 301: Same here.
```suggestion
        generate_metadata_files: Optional[bool] = None,
```

py/server/deephaven/parquet.py
chipkent: The prior name is `write`.  It seems like this should be `write_partitioned`.

Are there multiple ways partitioning can be done?  Is key_value the only way?  If there are mutliple ways, it is likely more pythonic to have function arguments to control the type instead of using the function name.

py/server/deephaven/parquet.py
```diff
@@ -280,6 +288,93 @@ def write(
...
+) -> None:
+    """ Write a table to a Parquet file in a key-value partitoned format.
```
chipkent, line 303: A user is not going to know what a "key-value partitioned format" is.  Provide some details.

py/server/deephaven/parquet.py
```diff
@@ -280,6 +288,93 @@ def write(
...
+
+    Args:
+        table (Table): the source table or partitioned table
+        destination_dir (str): the destination root directory to store partitioned data in nested format. Non-existing
+            directories are created.
```
chipkent, line 308: Not clear for a user.

py/server/deephaven/parquet.py
```diff
@@ -280,6 +288,93 @@ def write(
...
+        col_definitions (Optional[List[Column]]): the column definitions to use, default is None
+        col_instructions (Optional[List[ColumnInstruction]]): instructions for customizations while writing, default is None
```
chipkent, line 310: Neither of these explain what the defaults do so they are not useful for users.

py/server/deephaven/parquet.py
```diff
@@ -280,6 +288,93 @@ def write(
...
+        compression_codec_name (Optional[str]): the default compression codec to use, if not specified, defaults to SNAPPY
```
chipkent, line 311: Are these values from a DH enum or from parquet?  If there is a specific set of values, we should consider listing them here.

py/server/deephaven/parquet.py
```diff
@@ -280,6 +288,93 @@ def write(
...
+        max_dictionary_keys (Optional[int]): the maximum dictionary keys allowed, if not specified, defaults to 2^20 (1,048,576)
+        max_dictionary_size (Optional[int]): the maximum dictionary size (in bytes) allowed, defaults to 2^20 (1,048,576)
```
chipkent, line 313: User will have no clue what the significance of these is.

py/server/deephaven/parquet.py
```diff
@@ -280,6 +288,93 @@ def write(
...
+        target_page_size (Optional[int]): the target page size in bytes, if not specified, defaults to 2^20 bytes (1 MiB)
+        base_name (Optional[str]): The base name for the individual partitioned tables, if not specified, defaults to
+            `{uuid}`, so files will have names of the format `<uuid>.parquet` where `uuid` is a randomly generated UUID.
+            Users can provide the following tokens in the base_name:
+            - The token `{uuid}` will be replaced with a random UUID. For example, a base name of
+            "table-{uuid}" will result in files named like "table-8e8ab6b2-62f2-40d1-8191-1c5b70c5f330.parquet.parquet".
+            - The token `{partitions}` will be replaced with an underscore-delimited, concatenated string of
+            partition values. For example, a base name of "{partitions}-table" will result in files like
+            "PC1=partition1/PC2=partitionA/PC1=partition1_PC2=partitionA-table.parquet", where "PC1" and "PC2" are
+            partitioning columns.
```
chipkent, line 323: 1. Does this case create subdirectories?
2. I'm confused by this.  Why are the partition values listed 2x?

py/server/deephaven/parquet.py
```diff
@@ -280,6 +288,93 @@ def write(
...
+            - The token `{i}` will be replaced with an automatically incremented integer for files in a directory. For
+            example, a base name of "table-{i}" will result in files named like "PC=partition1/table-0.parquet",
+            "PC=partition1/table-1.parquet", etc.
+        generate_metadata_files (Optional[bool]): whether to generate _metadata and _common_metadata files, default is False
```
chipkent, line 327: See comments about users needing more description and the default.

py/server/deephaven/parquet.py
chipkent: Looking more, it might make sense, but it will need clarification for users.

py/server/deephaven/parquet.py
```diff
@@ -290,6 +385,7 @@ def batch_write(
     max_dictionary_size: Optional[int] = None,
     target_page_size: Optional[int] = None,
     grouping_cols: Optional[List[str]] = None,
+    generate_metadata_files: Optional[bool] = False,
```
chipkent, line 388: Same
```suggestion
    generate_metadata_files: Optional[bool] = None,
```

py/server/deephaven/parquet.py
```diff
@@ -310,6 +406,7 @@ def batch_write(
         max_dictionary_size (Optional[int]): the maximum dictionary size (in bytes) allowed, defaults to 2^20 (1,048,576)
         target_page_size (Optional[int]): the target page size in bytes, if not specified, defaults to 2^20 bytes (1 MiB)
         grouping_cols (Optional[List[str]]): the group column names
+        generate_metadata_files (Optional[bool]): whether to generate _metadata and _common_metadata files, default is False
```
chipkent, line 409: See other comments on this docstring.

py/server/deephaven/parquet.py
chipkent: Should this method be separate from `write` or part of `write`?  I may lean to separate, but our python code should be checked for precedent.

py/server/tests/test_parquet.py
```diff
@@ -576,6 +581,85 @@ def test_read_parquet_from_s3(self):
             read("s3://dh-s3-parquet-test1/multiColFile.parquet", special_instructions=s3_instructions).select()
         # TODO(deephaven-core#5064): Add support for local S3 testing
 
+    def verify_metadata_files(self, root_dir):
```
chipkent, line 584: I would put this function before the tests to make it easier on the next dev.

py/server/tests/test_parquet.py
```diff
@@ -576,6 +581,85 @@ def test_read_parquet_from_s3(self):
...
+        metadata_file_path = os.path.join(root_dir, '_metadata')
+        self.assertTrue(os.path.exists(metadata_file_path))
+        common_metadata_file_path = os.path.join(root_dir, '_common_metadata')
+        self.assertTrue(os.path.exists(common_metadata_file_path))
+
+    def test_writing_partitioned_data(self):
```
chipkent, line 590:
```suggestion
    def test_write_partitioned_data(self):
```
Just made the naming consistent with the other tests -- I think.

py/server/tests/test_parquet.py
```diff
@@ -576,6 +581,85 @@ def test_read_parquet_from_s3(self):
...
+        # Test all different APIs
+        write_key_value_partitioned_table(partitioned_table, destination_dir=root_dir, base_name=base_name,
+                                          max_dictionary_keys=max_dictionary_keys)
+        from_disk = read(root_dir)
+        definition = from_disk.columns
+        verify_table_from_disk(from_disk)
+        verify_file_names()
+        from_disk = None
+
+        shutil.rmtree(root_dir)
```
chipkent, line 626: If you need to make certain these root_dirs get cleaned up in case there is a problem and the test is rerun, consider try-finally or a shutdown for the tests or putting a rmtree before the first test.
https://www.geeksforgeeks.org/try-except-else-and-finally-in-python/

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -143,6 +163,24 @@ private ChunkPage<ATTR> getPageImpl(@Nullable FillContext fillContext, int pageN
...
+        if (fillContext != null) {
+            initialize(fillContext);
+            return getPageContainingImpl(fillContext, rowKey);
+        }
+        // Create a fill context and use it to initialize this class as well as read the page
+        // TODO What should be the chunk capacity here?
+        try (final FillContext fillContext1 = makeFillContext(1, null)) {
+            initialize(fillContext1);
+            return getPageContainingImpl(fillContext1, rowKey);
+        }
```
rcaudy, line 179:
```suggestion
        // We don't really use chunk capacity in our FillContext. In practice, however, this method is only invoked with
        // a null FillContext for single-element "get" methods.
        try (final FillContext allocatedFillContext = fillContext != null ? null : makeFillContext(1, null)) {
            final FillContext fillContextToUse = fillContext != null ? fillContext : allocatedFillContext;
            initialize(fillContextToUse);
            return getPageContainingImpl(fillContextToUse, rowKey);
        }
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -47,39 +49,57 @@ private static final class PageState<ATTR extends Any> {
...
-            fixedPageSize = numRows();
+        isInitialized = false;
+        numPages = NUM_PAGES_NOT_INITIALIZED;
+        fixedPageSize = PAGE_SIZE_NOT_FIXED;
+        offsetIndex = null;
+        pageStates = null;
+        columnPageDirectAccessor = null;
+    }
+
+    private void initialize(@Nullable final FillContext fillContext) {
```
rcaudy, line 70:
```suggestion
    private void ensureInitialized(@Nullable final FillContext fillContext) {
```

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/pagestore/OffsetIndexBasedColumnChunkPageStore.java
```diff
@@ -143,6 +163,24 @@ private ChunkPage<ATTR> getPageImpl(@Nullable FillContext fillContext, int pageN
     @Override
     @NotNull
     public ChunkPage<ATTR> getPageContaining(@Nullable final FillContext fillContext, long rowKey) {
```
rcaudy, line 165: `rowKey` can be `final` now

rcaudy (commented):
.

py/server/tests/test_parquet.py
malhotrashivam: No, I don't want to rerun any tests. The test can fail the first time it fails.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -697,9 +691,11 @@ private static void writeKeyValuePartitionedTableImpl(@NotNull final Partitioned
         }
         final Map<String, Map<ParquetCacheTags, Object>> computedCache =
                 buildComputedCache(() -> sourceTable.orElseGet(partitionedTable::merge), leafDefinition);
-        writeParquetTablesImpl(partitionedData.toArray(Table[]::new), leafDefinition, writeInstructions,
-                destinations.toArray(File[]::new), leafDefinition.getGroupingColumnNamesArray(),
-                partitioningColumnsSchema, new File(destinationRoot), computedCache);
+        final Table[] partitionedDataArray = partitionedData.toArray(Table[]::new);
+        // TODO Verify correctness with Ryan/Larry
```
malhotrashivam, line 695: Work with Larry/Ryan

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -696,12 +702,14 @@ public void writeKeyValuePartitionedDataWithIntegerPartitionsTest() {
...
                 ColumnDefinition.ofInt("PC2").withPartitioning(),
-                ColumnDefinition.ofLong("I").withGrouping());
+                ColumnDefinition.ofLong("I"));
         final Table inputData = ((QueryTable) TableTools.emptyTable(1_000_000)
                 .updateView("PC1 = (int)(ii%3)",
                         "PC2 = (int)(ii%2)",
                         "I = ii"))
                 .withDefinitionUnsafe(definition);
+        DataIndexer.getOrCreateDataIndex(inputData, "I");
+        // TODO verify correctness of the indexing data
```
malhotrashivam, line 712: Work with Larry/Ryan

extensions/parquet/table/src/test/java/io/deephaven/parquet/table/ParquetTableReadWriteTest.java
```diff
@@ -587,28 +593,28 @@ public void basicParquetWithMetadataTest() {
     }
 
     @Test
-    public void parquetWithGroupingDataAndMetadataTest() {
+    public void parquetWithIndexingDataAndMetadataTest() {
+        // TODO verify correctness of the indexing data
```
malhotrashivam, line 597: Work with Ryan/Larry

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -483,15 +501,265 @@ private static List<ParquetTableWriter.IndexWritingInfo> indexInfoBuilderHelper(
...
+     * method will call {@link Table#partitionBy(String...) partitionBy} on all the partitioning columns of provided
+     * table. The generated parquet files will have names of the format provided by
+     * {@link ParquetInstructions#baseNameForPartitionedParquetData()}.
+     *
+     * @param sourceTable The table to partition and write
+     * @param destinationDir The path to destination root directory to store partitioned data in nested format.
+     *        Non-existing directories are created.
+     * @param writeInstructions Write instructions for customizations while writing
+     */
+    public static void writeKeyValuePartitionedTable(@NotNull final Table sourceTable,
```
malhotrashivam, line 516: Can we name these methods as `writePartitioned` instead?

py/server/deephaven/parquet.py
```diff
@@ -239,22 +244,30 @@ def write(
...
-        col_definitions (Optional[List[Column]]): the column definitions to use, default is None
-        col_instructions (Optional[List[ColumnInstruction]]): instructions for customizations while writing, default is None
-        compression_codec_name (Optional[str]): the default compression codec to use, if not specified, defaults to SNAPPY
-        max_dictionary_keys (Optional[int]): the maximum dictionary keys allowed, if not specified, defaults to 2^20 (1,048,576)
-        max_dictionary_size (Optional[int]): the maximum dictionary size (in bytes) allowed, defaults to 2^20 (1,048,576)
+        col_definitions (Optional[List[Column]]): the column definitions to use for writing, instead of the definitions
+            implied by the table. Default is None, which means use the column definitions implied by the table
+        col_instructions (Optional[List[ColumnInstruction]]): instructions for customizations while writing particular
+            columns, default is None, which means no specialization for any column
+        compression_codec_name (Optional[str]): the default compression codec to use, if not specified, defaults to SNAPPY.
```
chipkent, line 260: Given that the compression codec must come from a discrete set, we should probably type hint as a `Literal`.
https://docs.python.org/3/library/typing.html#typing.Literal
Confirm the view with @jmao-denver

py/server/deephaven/parquet.py
```diff
@@ -298,33 +305,45 @@ def write_key_value_partitioned_table(
...
-        max_dictionary_keys (Optional[int]): the maximum dictionary keys allowed, if not specified, defaults to 2^20 (1,048,576)
-        max_dictionary_size (Optional[int]): the maximum dictionary size (in bytes) allowed, defaults to 2^20 (1,048,576)
+        destination_dir (str): The path to destination root directory in which the partitioned parquet data will be stored
+            in a nested directory structure format. Non-existing directories in the provided path will be created.
+        col_definitions (Optional[List[Column]]): the column definitions to use for writing, instead of the definitions
+            implied by the table. Default is None, which means use the column definitions implied by the table
+        col_instructions (Optional[List[ColumnInstruction]]): instructions for customizations while writing particular
+            columns, default is None, which means no specialization for any column
+        compression_codec_name (Optional[str]): the default compression codec to use, if not specified, defaults to SNAPPY.
+            Allowed values are "UNCOMPRESSED", "SNAPPY", "GZIP", "LZO", "BROTLI", "LZ4", "LZ4_RAW", "ZSTD".
```
chipkent, line 325: See note above about `Literal` type hint.

py/server/deephaven/parquet.py
```diff
@@ -399,14 +397,21 @@ def batch_write(
...
             this method unsafe for concurrent use
-        col_definitions (List[Column]): the column definitions to use
+        col_definitions (Optional[List[Column]]): the column definitions to use for writing, instead of the definitions
+            implied by the table. Default is None, which means use the column definitions implied by the table
         col_instructions (Optional[List[ColumnInstruction]]): instructions for customizations while writing
-        compression_codec_name (Optional[str]): the compression codec to use, if not specified, defaults to SNAPPY
-        max_dictionary_keys (Optional[int]): the maximum dictionary keys allowed, if not specified, defaults to 2^20 (1,048,576)
-        max_dictionary_size (Optional[int]): the maximum dictionary size (in bytes) allowed, defaults to 2^20 (1,048,576)
+        compression_codec_name (Optional[str]): the compression codec to use, if not specified, defaults to SNAPPY.
+            Allowed values are "UNCOMPRESSED", "SNAPPY", "GZIP", "LZO", "BROTLI", "LZ4", "LZ4_RAW", "ZSTD".
```
chipkent, line 404: Same type hint

py/server/deephaven/parquet.py
malhotrashivam: I have made the change but I am not too confident if I have made it correctly. Basically, the same codec list is to be passed to three methods. So I declared it as a global variable, but that wouldn't give the users any hint, right? Because the list is actually not visible to users.
So should I make copies at three places (in write, batch_write and write_partitioned)?

jmao-denver: I am not sure using Literal is a good idea here as the supported compression types are more than 3 and could be expanded in the future. That said, it could be helpful to the users to list all the supported types in the docstring.

malhotrashivam: Okay, I will change it back.

py/server/deephaven/parquet.py
```diff
@@ -6,14 +6,14 @@
 Parquet files. """
 from dataclasses import dataclass
 from enum import Enum
-from typing import List, Optional, Union, Dict
+from typing import List, Optional, Union, Dict, Literal
```
chipkent, line 9:
```suggestion
from typing import List, Optional, Union, Dict
```
If we aren't using `Literal`, this doesn't belong.

py/server/deephaven/parquet.py
```diff
@@ -234,20 +244,30 @@ def write(
...
-        col_instructions (Optional[List[ColumnInstruction]]): instructions for customizations while writing, default is None
-        compression_codec_name (Optional[str]): the default compression codec to use, if not specified, defaults to SNAPPY
-        max_dictionary_keys (Optional[int]): the maximum dictionary keys allowed, if not specified, defaults to 2^20 (1,048,576)
-        max_dictionary_size (Optional[int]): the maximum dictionary size (in bytes) allowed, defaults to 2^20 (1,048,576)
+        col_definitions (Optional[List[Column]]): the column definitions to use for writing, instead of the definitions
+            implied by the table. Default is None, which means use the column definitions implied by the table
+        col_instructions (Optional[List[ColumnInstruction]]): instructions for customizations while writing particular
+            columns, default is None, which means no specialization for any column
+        compression_codec_name (Optional[str]): the default compression codec to use, if not specified, defaults to SNAPPY.
+            Allowed values are "UNCOMPRESSED", "SNAPPY", "GZIP", "LZO", "LZ4", "LZ4_RAW", "ZSTD".
```
chipkent, line 261:
```suggestion
            Allowed values include "UNCOMPRESSED", "SNAPPY", "GZIP", "LZO", "LZ4", "LZ4_RAW", "ZSTD".
```

py/server/deephaven/parquet.py
```diff
@@ -260,22 +280,95 @@ def write(
...
+    Args:
+        table (Table): the source table or partitioned table
+        destination_dir (str): The path to destination root directory in which the partitioned parquet data will be stored
+            in a nested directory structure format. Non-existing directories in the provided path will be created.
+        col_definitions (Optional[List[Column]]): the column definitions to use for writing, instead of the definitions
+            implied by the table. Default is None, which means use the column definitions implied by the table
+        col_instructions (Optional[List[ColumnInstruction]]): instructions for customizations while writing particular
+            columns, default is None, which means no specialization for any column
+        compression_codec_name (Optional[str]): the default compression codec to use, if not specified, defaults to SNAPPY.
+            Allowed values are "UNCOMPRESSED", "SNAPPY", "GZIP", "LZO", "LZ4", "LZ4_RAW", "ZSTD".
```
chipkent, line 325:
```suggestion
            Allowed values include "UNCOMPRESSED", "SNAPPY", "GZIP", "LZO", "LZ4", "LZ4_RAW", "ZSTD".
```

py/server/deephaven/parquet.py
```diff
@@ -303,13 +397,21 @@ def batch_write(
...
             this method unsafe for concurrent use
-        col_definitions (List[Column]): the column definitions to use
+        col_definitions (Optional[List[Column]]): the column definitions to use for writing, instead of the definitions
+            implied by the table. Default is None, which means use the column definitions implied by the table
         col_instructions (Optional[List[ColumnInstruction]]): instructions for customizations while writing
-        compression_codec_name (Optional[str]): the compression codec to use, if not specified, defaults to SNAPPY
-        max_dictionary_keys (Optional[int]): the maximum dictionary keys allowed, if not specified, defaults to 2^20 (1,048,576)
-        max_dictionary_size (Optional[int]): the maximum dictionary size (in bytes) allowed, defaults to 2^20 (1,048,576)
+        compression_codec_name (Optional[str]): the compression codec to use, if not specified, defaults to SNAPPY.
+            Allowed values are "UNCOMPRESSED", "SNAPPY", "GZIP", "LZO", "LZ4", "LZ4_RAW", "ZSTD".
```
chipkent, line 404:
```suggestion
            Allowed values include "UNCOMPRESSED", "SNAPPY", "GZIP", "LZO", "LZ4", "LZ4_RAW", "ZSTD".
```

chipkent (dismissed):
Python LGTM

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
rcaudy: In the branch Larry and I shipped, which only looks at single-table indexes, we've taken the second way.

You could do that, by assuming that:
1. A `PartitionedTable` will have indexes for you to write (or not) on its constituents. Conceivably, those constituents should be considered individually for writing indexes.
2. A Table that you're going to partition before writing will have indexes you can intersect. I wouldn't want the generic `partitionBy` to intersect parent table indexes, but we could easily do a `transform` to allow us to intersect the index tables' row sets with the constituent row sets, then filter the empty row sets.

Pseudocode for intersecting, probably with bad parentheses, etc:
```
partitionedSource = sourceTable.partitionBy(partitioningColumns)
for dataIndex in sourceTable indexes:
    indexTable = dataIndex.table()
    partitionedIndexTable = partitionedSource.transform(c ->
        // Do this with `FunctionalColumns` to avoid compiles
        indexTable
            .update(List.of(new FunctionalColumn(dataIndex.rowSetColumnName(), RowSet.class, dataIndex.rowSetColumnName(), RowSet.class, dataIndexRowSet -> dataIndexRowSet.intersect(c.getRowSet()))))
            .updateView(List.of(new FunctionalColumn(dataIndex.rowSetColumnName(), RowSet.class, "__NON_EMPTY__", Boolean.class, RowSet::isNonempty)))
            .where(Filter.isTrue(ColumnName.of("__NON_EMPTY__"))))
    // We *could* add `StandaloneDataIndex`es to the constituents, or just write from the partitioned index table
```

All of that said, I'm not opposed to using instructions, but so far we've just said "the user tells us what to index using the indexing APIs". If we're going to use instructions, we need to add that support to non-partitioned table writing code.

malhotrashivam: Created issue #5292 for this

py/server/deephaven/parquet.py
```diff
@@ -43,6 +43,8 @@ def _build_parquet_instructions(
     is_refreshing: bool = False,
     for_read: bool = True,
     force_build: bool = False,
+    generate_metadata_files: Optional[bool] = None,
+    base_name: Optional[str] = None,
     special_instructions: Optional[s3.S3Instructions] = None,
```
rcaudy, line 48: Add support for passing lists of column lists to index. Consult Jianfeng for signature help. Here, and on old write method as well.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -797,18 +768,18 @@ private static void writeKeyValuePartitionedTableImpl(@NotNull final Partitioned
             while (constituentIterator.hasNext()) {
                 final ObjectVector<? extends Table> constituentVector = constituentIterator.next();
                 final List<String> partitionStrings = partitionStringsList.get(row);
-                final File relativePath = new File(destinationRoot, String.join("/", partitionStrings));
+                final File relativePath = new File(destinationRoot, String.join(File.separator, partitionStrings));
```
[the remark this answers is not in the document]
malhotrashivam: This can be an issue because the relative path written down to metadata file would have delimiter of windows and Linux won't understand it.
I would need to test it though.
If its an issue, we would need to normalize the relative paths first at the time of reading to use the correct file separator.

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/ParquetTools.java
```diff
@@ -798,27 +802,38 @@ private static void writeKeyValuePartitionedTableImpl(@NotNull final Partitioned
         } else {
             partitioningColumnsSchema = null;
         }
-        final Map<String, Map<ParquetCacheTags, Object>> computedCache =
-                buildComputedCache(() -> sourceTable.orElseGet(partitionedTable::merge), leafDefinition);
         final Table[] partitionedDataArray = partitionedData.toArray(Table[]::new);
         try (final SafeCloseable ignored = LivenessScopeStack.open()) {
             // TODO(deephaven-core#5292): Optimize creating index on constituent tables
```
rcaudy, line 807: I think this matches the wrong line now.

py/server/deephaven/parquet.py
```diff
@@ -382,23 +410,19 @@ def batch_write(
...
 
     Args:
         tables (List[Table]): the source tables
-        paths (List[str]): the destinations paths. Any non existing directories in the paths provided are
+        paths (List[str]): the destination paths. Any non-existing directories in the paths provided are
             created. If there is an error, any intermediate directories previously created are removed; note this makes
             this method unsafe for concurrent use
-        col_definitions (Optional[List[Column]]): the column definitions to use for writing, instead of the definitions
-            implied by the table. Default is None, which means use the column definitions implied by the table
+        col_definitions (List[Column]): the column definitions to use for writing.
```
rcaudy, line 425: Was this a good change?

malhotrashivam: So I made a mistake by changing it from List[Column] to Optional[List[Column]] earlier. But have reverted it back now.

rcaudy (dismissed):

py/server/deephaven/parquet.py
```diff
@@ -234,20 +248,121 @@ def write(
...
+        target_page_size (Optional[int]): the target page size in bytes, if not specified, defaults to 2^20 bytes (1 MiB)
+        generate_metadata_files (Optional[bool]): whether to generate parquet _metadata and _common_metadata files,
+            defaults to False. Generating these files can help speed up reading of partitioned parquet data because these
+            files contain metadata (including schema) about the entire dataset, which can be used to skip reading some
+            files.
+        index_columns (Optional[List[List[str]]]): Lists containing the column names for indexes to persist. The write
+            operation will store the index info as sidecar tables. By default, data indexes to write are determined by
+            those present on the source table. This argument is used to narrow the set of indexes to write, or to be
+            explicit about the expected set of indexes present on all sources. Indexes that are specified but missing
+            will be computed on demand.
```
chipkent, line 280: I'm confused by the type hint and documentation.  The docs sounds like it is a list of column names (e.g. `List[str]`, but the type hint is `List[List[str]]`.
1. If this should be `List[List[str]]`, the doc should clarify what the argument contains, since I don't think I can guess with certainty.
2. I think @jmao-denver has been using `Sequence` in place of `List` in many places because it is more general.  You should discuss the `List` type hints with him.

py/server/deephaven/parquet.py
```diff
@@ -234,20 +248,121 @@ def write(
...
+            _JParquetTools.writeTable(table.j_table, path, table_definition, write_instructions)
+    except Exception as e:
+        raise DHError(e, "failed to write to parquet data.") from e
+
+
+def write_partitioned(
+        table: Union[Table, PartitionedTable],
+        destination_dir: str,
+        col_definitions: Optional[List[Column]] = None,
+        col_instructions: Optional[List[ColumnInstruction]] = None,
```
chipkent, line 315: See prior comment on `List` vs `Sequence`.

py/server/deephaven/parquet.py
```diff
@@ -234,20 +248,121 @@ def write(
...
+            "PC=partition1/table-1.parquet", etc.
+        generate_metadata_files (Optional[bool]): whether to generate parquet _metadata and _common_metadata files,
+            defaults to False. Generating these files can help speed up reading of partitioned parquet data because these
+            files contain metadata (including schema) about the entire dataset, which can be used to skip reading some
+            files.
+        index_columns (Optional[List[List[str]]]): Lists containing the column names for indexes to persist. The write
+            operation will store the index info as sidecar tables. By default, data indexes to write are determined by
+            those present on the source table. This argument is used to narrow the set of indexes to write, or to be
+            explicit about the expected set of indexes present on all sources. Indexes that are specified but missing
+            will be computed on demand.
```
chipkent, line 365: See prior comment on type hint vs text.

py/server/deephaven/parquet.py
```diff
@@ -289,27 +410,36 @@ def batch_write(
...
-        grouping_cols (Optional[List[str]]): the group column names
+        generate_metadata_files (Optional[bool]): whether to generate parquet _metadata and _common_metadata files,
+            defaults to False. Generating these files can help speed up reading of partitioned parquet data because these
+            files contain metadata (including schema) about the entire dataset, which can be used to skip reading some
+            files.
+        index_columns (Optional[List[List[str]]]): Lists containing the column names for indexes to persist. The write
+            operation will store the index info as sidecar tables. By default, data indexes to write are determined by
+            those present on the source table. This argument is used to narrow the set of indexes to write, or to be
+            explicit about the expected set of indexes present on all sources. Indexes that are specified but missing
+            will be computed on demand.
```
chipkent, line 442: See comment above.

py/server/deephaven/parquet.py
malhotrashivam: I have added more details to the comments explaining the argument for now. Will wait for Jianfeng to have a look regarding Sequence vs List.

jmao-denver: Sequence is more general and flexible. I'd prefer Sequence.

chipkent (dismissed):

extensions/parquet/table/src/main/java/io/deephaven/parquet/table/layout/ParquetMetadataFileLayout.java
malhotrashivam, line 1: Metadata files written on windows by pyarrow cannot be read by deephaven becuase of the code that does
```
                final Path filePath = Paths.get(filePathString);
                final int numPartitions = filePath.getNameCount() - 1;
                if (numPartitions != partitioningColumns.size()) {
                    throw new TableDataException(String.format(
                            "Unexpected number of path elements in %s for partitions %s",
                            filePathString, partitions.keySet()));
                }
```

At this point, filePathString has file separator "\" which Paths.get doesn't understand and thus cannot identify the partitions.
Whereas pyarrow is able to such metadata files on macos.

malhotrashivam: Added a fix for this.

rcaudy (approved):

deephaven-internal:

Community: https://github.com/deephaven/deephaven-docs-community/issues/180

PR #5110, opened 2024-02-02
DEEPHAVEN_VERSION must be set when installing py-client wheel

niloc132:
Without this, the pip install apparently fails (silently?) and tests do not run.

Followup #4865


rcaudy (approved):

devinrsmith (approved):

jmao-denver (approved):
LGTM

chipkent (approved):

PR #5123, opened 2024-02-08
PythonScopeJpyImpl: Disable Conversion Cache

### LINKED ISSUE #5112: The auto conversion feature of the Python UDF decorator has a performance problem.
This was found by the latest bench marking effort to measure the performance impact from the usability improvement on Python UDF.

nbauernfeind:
This fixes a memory leak discovered in #5112 via repeated runs of:

```python
import time, jpy
from deephaven import empty_table, garbage_collect

row_count = 10_000_000

begin_time = time.perf_counter_ns()
source = empty_table(row_count).update(["X = repeat(ii % 250, 100)"]) 
empty_table(1).select('Y = 1')
print('Rows / Sec:', row_count / ((time.perf_counter_ns() - begin_time) / 1_000_000_000))

Runtime = jpy.get_type('java.lang.Runtime')
print('Gigs Used Before GC:',
        (Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory()) / 1024 / 1024 / 1024)

del source;
del row_count;
del begin_time
for i in range(10):
    garbage_collect()

print('Gigs Used After GC:',
        (Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory()) / 1024 / 1024 / 1024)
del Runtime
```

Nightlies running here: https://github.com/nbauernfeind/deephaven-core/actions/runs/7833860264/

### DIFF
--- a/engine/context/src/main/java/io/deephaven/engine/context/StandaloneQueryScope.java
+++ b/engine/context/src/main/java/io/deephaven/engine/context/StandaloneQueryScope.java
@@ -6,9 +6,11 @@ import io.deephaven.engine.liveness.LivenessReferent;
 import io.deephaven.engine.updategraph.DynamicNode;
 import io.deephaven.hash.KeyedObjectHashMap;
 import io.deephaven.hash.KeyedObjectKey;
+import org.jetbrains.annotations.NotNull;
 
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 /**
@@ -77,9 +79,11 @@ public class StandaloneQueryScope extends LivenessArtifact implements QueryScope
     }
 
     @Override
-    public Map<String, Object> toMap() {
+    public Map<String, Object> toMap(@NotNull final Predicate<Map.Entry<String, Object>> predicate) {
         return valueRetrievers.entrySet().stream()
-                .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().value));
+                .map(e -> Map.entry(e.getKey(), (Object) e.getValue().value))
+                .filter(predicate)
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
     }
 
     private static class ValueRetriever<T> {
--- a/engine/context/src/main/java/io/deephaven/engine/context/StandaloneQueryScope.java
+++ b/engine/context/src/main/java/io/deephaven/engine/context/StandaloneQueryScope.java
@@ -6,6 +6,7 @@ import io.deephaven.engine.liveness.LivenessReferent;
 import io.deephaven.engine.updategraph.DynamicNode;
 import io.deephaven.hash.KeyedObjectHashMap;
 import io.deephaven.hash.KeyedObjectKey;
+import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.HashMap;
@@ -82,7 +83,7 @@ public class StandaloneQueryScope extends LivenessArtifact implements QueryScope
     public Map<String, Object> toMap(@NotNull final Predicate<Map.Entry<String, Object>> predicate) {
         final HashMap<String, Object> result = new HashMap<>();
         valueRetrievers.entrySet().stream()
-                .map(e -> Map.entry(e.getKey(), (Object) e.getValue().value))
+                .map(e -> ImmutablePair.of(e.getKey(), (Object) e.getValue().value))
                 .filter(predicate)
                 .forEach(e -> result.put(e.getKey(), e.getValue()));
         return result;
--- a/engine/context/src/main/java/io/deephaven/engine/context/StandaloneQueryScope.java
+++ b/engine/context/src/main/java/io/deephaven/engine/context/StandaloneQueryScope.java
@@ -8,10 +8,10 @@ import io.deephaven.hash.KeyedObjectHashMap;
 import io.deephaven.hash.KeyedObjectKey;
 import org.jetbrains.annotations.NotNull;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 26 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
```diff
@@ -189,14 +190,43 @@ public final class QueryLanguageParser extends GenericVisitorAdapter<Class<?>, Q
...
+     * {@link #getResult()}} method.
+     *
+     * @param expression The query language expression to parse
+     * @param packageImports Wildcard package imports
+     * @param classImports Individual class imports
+     * @param staticImports Wildcard static imports. All static variables and methods for the given classes are
+     *        imported.
+     * @param variables A map of the names of scope variables to their types
+     * @param variableTypeArguments A map of the names of scope variables to their type arguments
+     * @param possibleParams A mutable map of the names of query scope variables to their values
```
jmao-denver, line 217: Don't quite like the name `possibleParams`, I think `queryScopeVariables` might be easier for readers.

jmao-denver (dismissed):
1. weak-values cache makes sense for the main global scope
2. like the decoupling of QueryScope from QLP

engine/table/src/main/java/io/deephaven/engine/table/impl/lang/QueryLanguageParser.java
rcaudy: This is a good point, and I endorse it whole-heartedly.

rcaudy (commented):
.

engine/context/src/main/java/io/deephaven/engine/context/StandaloneQueryScope.java
```diff
@@ -77,9 +79,13 @@ public <T> void putParam(final String name, final T value) {
     }
 
     @Override
-    public Map<String, Object> toMap() {
-        return valueRetrievers.entrySet().stream()
-                .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().value));
+    public Map<String, Object> toMap(@NotNull final Predicate<Map.Entry<String, Object>> predicate) {
+        final HashMap<String, Object> result = new HashMap<>();
+        valueRetrievers.entrySet().stream()
+                .map(e -> Map.entry(e.getKey(), (Object) e.getValue().value))
```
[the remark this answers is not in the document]
nbauernfeind: as discussed offline, this is to unwrap the value retriever. I assume the retriever is allowed to return null, but that the retrievers themselves are non-null.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/AbstractConditionFilter.java
```diff
@@ -98,8 +94,9 @@ public synchronized void init(TableDefinition tableDefinition) {
 
         try {
             final QueryScope queryScope = ExecutionContext.getContext().getQueryScope();
-            final Map<String, Object> possibleParams = queryScope.toMap();
-            for (Map.Entry<String, Object> param : possibleParams.entrySet()) {
+            final Map<String, Object> queryScopeVariables = queryScope.toMap(
```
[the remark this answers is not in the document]
nbauernfeind: I'm willing to do this if this is your preference. I was considering the other usages. They both filter to Table instances -- unfortunately they also need to unwrap first, and probably we would prefer to unwrap once.

engine/table/src/main/java/io/deephaven/engine/util/PythonEvaluatorJpy.java
```diff
@@ -47,6 +47,8 @@ public static PythonEvaluatorJpy withGlobals() throws IOException, InterruptedEx
 
     private PythonEvaluatorJpy(PyDictWrapper globals) {
         this.globals = globals;
+        // forcefully load this helper class which imports python modules
+        PyCallableWrapperJpyImpl.init();
```
rcaudy, line 51: I'm worried this is some kind of band-aid allowing us to mask a deeper issue.

server/src/main/java/io/deephaven/server/console/python/PythonImportInitializer.java
```diff
@@ -0,0 +1,17 @@
+package io.deephaven.server.console.python;
+
+import io.deephaven.engine.util.PyCallableWrapperJpyImpl;
+import io.deephaven.integrations.python.PythonObjectWrapper;
+
+public abstract class PythonImportInitializer {
```
rcaudy, line 6: Feels like this pattern could be replaced with service loader.

engine/table/src/main/java/io/deephaven/engine/util/PythonScope.java
```diff
@@ -160,14 +160,8 @@ default Collection<String> getKeysCollection() {
      */
     default Map<String, Object> getEntriesMap() {
```
rcaudy, line 161: This method feels inconsistent. Do we still need it?

engine/table/src/main/java/io/deephaven/engine/util/PythonScope.java
```diff
@@ -134,13 +135,12 @@ default Stream<String> getKeys() {
...
-     * and the values via {@link #convertValue(Object)}.
+     * but the values still need conversion via {@link #convertValue(Object)}.
      *
-     * @return the string keys and converted values
+     * @return the string keys and original values
      */
-    default Stream<Entry<String, Object>> getEntries() {
-        return getEntriesRaw()
-                .map(e -> new SimpleImmutableEntry<>(convertStringKey(e.getKey()), convertValue(e.getValue())));
+    default Stream<Entry<String, PyObj>> getEntries() {
```
rcaudy, line 142: To your point, this adds very little value now.

server/src/main/java/io/deephaven/server/console/python/PythonImportInitializer.java
devinrsmith: I think the problem w/ service loader is that the classes are on the classpath regardless of whether we are running python or groovy. In this current pattern, it's only invoked when we are coming from known-python script session.

rcaudy (approved):

Issue #5124, opened 2024-02-08
Propagate sortable/filterable/key_columns/sorted attributes in select/update/view/updateView


nbauernfeind:
When investigating #2474, @rcaudy and I audited all existing attributes to consider whether or not there were additional attributes that should propagate.

We discussed that we might want to propagate attributes for these:
- `SORTABLE_COLUMNS_ATTRIBUTE`
- `FILTERABLE_COLUMNS_ATTRIBUTE`
- `KEY_COLUMNS_ATTRIBUTE`
- `SORTED_COLUMNS_ATTRIBUTE`

We agree that we must not propagate them across anything but a source-column (i.e. a retain / rename).

PR #5125, opened 2024-02-08
Propagate Blink Attribute Downstream of View/UpdateView/Select/UpdateView

### LINKED ISSUE #2474: Select/Update do not propagate Blink Table attribute

nbauernfeind:
Fixes #2474.

Potential follow up work if it ever gets priority: #5124 
Nightlies: https://github.com/nbauernfeind/deephaven-core/actions/runs/7836660158/

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/table/impl/BaseTable.java
+++ b/engine/table/src/main/java/io/deephaven/engine/table/impl/BaseTable.java
@@ -299,13 +299,14 @@ public abstract class BaseTable<IMPL_TYPE extends BaseTable<IMPL_TYPE>> extends 
 
         tempMap.put(SNAPSHOT_VIEWPORT_TYPE, EnumSet.allOf(CopyAttributeOperation.class));
 
+        // Note: The logic applying this attribute for select/update/view/updateView is in SelectAndViewAnalyzerWrapper.
         tempMap.put(ADD_ONLY_TABLE_ATTRIBUTE, EnumSet.of(
                 CopyAttributeOperation.DropColumns,
                 CopyAttributeOperation.RenameColumns,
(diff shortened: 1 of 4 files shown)


engine/table/src/main/java/io/deephaven/engine/table/impl/BaseTable.java
```diff
@@ -347,8 +347,6 @@ public enum CopyAttributeOperation {
                 CopyAttributeOperation.Flatten,
                 CopyAttributeOperation.PartitionBy,
                 CopyAttributeOperation.Preview,
-                CopyAttributeOperation.View, // and Select, if added
```
rcaudy, line 350: Let's add "breadcrumb" comments to the blocks for attributes that we *do* (conditionally) pass through select/update/view/updateView/lazyUpdate.

engine/table/src/main/java/io/deephaven/engine/table/impl/select/analyzers/SelectAndViewAnalyzerWrapper.java
```diff
@@ -80,6 +80,10 @@ public QueryTable applyShiftsAndRemainingColumns(
                 // be convenient for test authors by propagating the test source table attribute
                 queryTable.setAttribute(Table.TEST_SOURCE_TABLE_ATTRIBUTE, true);
             }
+            if (sourceTable.isBlink()) {
```
rcaudy, line 83: Maybe a trivial unit test for this?

rcaudy (approved):

PR #5144, opened 2024-02-12
Add exception for exponential operators providing non-descending timestamps.

### LINKED ISSUE #4848: `UpdateBy` `em` time operators not handling out-of-order timestamps properly
**Description**
The operators for `update_by` `em*_time` expect non-descending timestamps but will treat negative time deltas as valid deltas.  This results in data corruption and invalid results.  

**Steps to reproduce**
Run the following code:
```
from deephaven.updateby import ema_time
import random

def random_char() -> str:
    return random.choice(['A', 'B'])

result = time_table("PT1S").update(formulas=["Sym = random_char()", "Numbers = randomInt(0, 100)"]).reverse()

result_w_ema = result.update_by([ema_time("Timestamp", "PT1S", "NumbersEMA = Numbers")], by=["Sym"])

ema_A = result_w_ema.where(["Sym == `A`"])

ema_fig = Figure()
ema_plot = ema_fig\
    .plot_xy(series_name="Numbers", t=ema_A, x="Timestamp", y="Numbers")\
    .plot_xy(series_name="EMA", t=ema_A, x="Timestamp", y="NumbersEMA")\
    .show()
```

**Expected results**

A clear exception explaining the need for ordered timestamps.

**Actual results**

Obviously incorrect data (negative EMA from all positive values).

**Versions**
 - Deephaven: <= 0.31.1

lbooker42:
Previous versions accepted non-descending time data and attempted to produce values using negative time deltas.  This is undefined behavior and an exception should be thrown explaining this to the user.

Closes #4848


engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/em/BigIntegerEMOperator.java
```diff
@@ -58,22 +60,25 @@ public void accumulateCumulative(@NotNull RowSequence inputKeys,
                         handleBadData(this, true);
                     } else if (isNullTime) {
                         // no change to curVal and lastStamp
+                    } else if (curVal == null) {
+                        // If the data looks good, and we have a null computed value, accept the current value
+                        curVal = new BigDecimal(input, control.bigValueContextOrDefault());
+                        lastStamp = timestamp;
                     } else {
                         final BigDecimal decimalInput = new BigDecimal(input, control.bigValueContextOrDefault());
```
malhotrashivam, line 68: Minor nitpick, you can move definition of decimalInput closer to line 80 where it is used after the if checks.

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/emstd/BigIntegerEmStdOperator.java
```diff
@@ -85,31 +86,34 @@ public void accumulateCumulative(@NotNull final RowSequence inputKeys,
...
-                                curVariance = alpha.multiply(
-                                        curVariance.add(
-                                                oneMinusAlpha.multiply(decInput.subtract(curEma).pow(2, mathContext)), mathContext),
-                                        mathContext);
-
-                                final BigDecimal decayedEmaVal = curEma.multiply(alpha, mathContext);
-                                curEma = decayedEmaVal.add(oneMinusAlpha.multiply(decInput, mathContext));
-                                curVal = curVariance.sqrt(mathContext);
-                                lastStamp = timestamp;
+                        final BigDecimal decInput = new BigDecimal(input, mathContext);
```
malhotrashivam, line 94: Same as earlier

malhotrashivam (commented):
I have a tangential comment which might not be your responsibility.
Basically, the ema [documentation page](https://deephaven.io/core/groovy/docs/reference/table-operations/update-by-operations/ema/) doesn't explain how ema calculation would work if we are using a timestamp column. I had to look through the code to understand it. Do you think there is any value in adding it, or is this more common knowledge that I was not familiar with?

engine/table/src/main/java/io/deephaven/engine/table/impl/updateby/em/IntEMOperator.java
```diff
@@ -74,10 +75,15 @@ public void accumulateCumulative(@NotNull RowSequence inputKeys,
                     } else if (isNullTime) {
                         // no change to curVal and lastStamp
                     } else if (curVal == NULL_DOUBLE) {
+                        // If the data looks good, and we have a null computed value, accept the current value
```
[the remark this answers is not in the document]
lbooker42: I rewrote this to (hopefully) better express what is happening.

malhotrashivam (approved):

PR #5148, opened 2024-02-13
Provide upperBin and lowerBin signatures that accept Durations

### LINKED ISSUE #5147: upperBin and lowerBin should have methods that accept Durations
The `upperBin` and `lowerBin` methods should have signatures that accept `Duration`s.

chipkent:
Provide upperBin and lowerBin signatures that accept Durations.

Resolves #5147

### DIFF
--- a/engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
+++ b/engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
@@ -3259,8 +3259,7 @@ public class DateTimeUtils {
      *
      * @param instant instant for which to evaluate the start of the containing window
      * @param interval size of the window
-     * @param offset The window start offset. For example, a value of 'PT1m' would offset all windows by
-     *        one minute.
+     * @param offset The window start offset. For example, a value of 'PT1m' would offset all windows by one minute.
      * @return {@code null} if any input is {@code null}; otherwise, an {@link Instant} representing the start of the
      *         window
      */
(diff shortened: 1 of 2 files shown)


engine/time/src/main/java/io/deephaven/time/DateTimeUtils.java
```diff
@@ -3170,6 +3170,26 @@ public static Instant lowerBin(@Nullable final Instant instant, long intervalNan
         return epochNanosToInstant(Numeric.lowerBin(epochNanos(instant), intervalNanos));
     }
 
+    /**
+     * Returns an {@link Instant} value, which is at the starting (lower) end of a time range defined by the interval
+     * nanoseconds. For example, a 5*MINUTE intervalNanos value would return the instant value for the start of the
```
devinrsmith, line 3175: I don't think `5*MINUTE intervalNanos` is correct in this context; maybe just `5 minute {@code interval}`?

devinrsmith, line 3175: This is also a criticism of the existing javadoc - we're considering *epoch* nanoseconds as the start of the interval. I do worry a bit generally that there may be some sloppiness around what the user wants and what we calculate, potentially as it relates to leap seconds and such.

For example, are we guaranteed that `lowerBin(x, Duration.ofSeconds(1)).equals(x.truncatedTo(ChronoUnit.SECONDS))`? Ditto wrt MICROS, MILLIS, MINUTES, HOURS, HALF_DAYS, DAYS.

rcaudy: Or use {@code 5 * MINUTE}

chipkent: I guess there is still stuff about time that I didn't know.

1. Java time does not use UTC, they have their own "Java Time-Scale".
2. "Java Time Scale" divides the day into exactly 86400 subdivisions known as seconds.  These seconds differ from SI seconds but closely match the "international civil time scale".
3. From 1972-11-03 to now, Java Time Scale exactly matches UTC-SLS on days that do not have leap seconds.  
4. On days that do have leap seconds, the leap second is spread equally over the last 1000 seconds of the day.

Details can be found at: https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html

Given this weirdness, I think the upperBin/lowerBin return what is expected.

devinrsmith (approved):

deephaven-internal:

Community: https://github.com/deephaven/deephaven.io/issues/3740

PR #5151, opened 2024-02-13
Provide query library functions for computing differences and sort order

### LINKED ISSUE #3413: rank() function to return indices of sorted values
There should be a `rank(Vector)` function that returns a vector of indices that would sort that input (so, each element's rank within the input vector).

For example:

```groovy
rank(5, 3, 4, 1, 2)
```

should return:

```
[ 4, 2, 3, 0, 1 ]
```

(this also lets you calculate percentile rankings by doing `rank(myVector)/len(myVector)`)

### LINKED ISSUE #5101: diff() function for calculating differences between values within an array
There should be a `diff(Vector)` function that returns a vector of differences between the values.

For example:

`diff(5, 3, 4, 1, 2)`

should return:

`[ NULL_INT, -2, 1, -3, 1 ]`

or possibly skip the obvious null and just return `[ -2, 1, -3, 1 ]`.

It should take a `io.deephaven.api.updateby.BadDataBehavior` as an argument (or have an overload that does, and do whatever we decide is reasonable/safe if you don't specify the arugment).

(This is kind of related to #3413.)

chipkent:
Query library functions for:
1) Computing array differences
2) Returning sort order

Resolves #5101 
Resolves #3413

### DIFF
--- a/engine/function/src/templates/Numeric.ftl
+++ b/engine/function/src/templates/Numeric.ftl
@@ -1596,7 +1596,6 @@ public class Numeric {
      * Differences off the end of the input vector are null.
      *
      * @param stride number of elements to skip between consecutive elements.
-     *     Positive strides are forward differences, negative strides are backward differences.
      * @param values input vector.
      * @return difference between elements.
      */
@@ -1612,7 +1611,6 @@ public class Numeric {
      * Differences off the end of the input vector are null.
      *
      * @param stride number of elements to skip between consecutive elements.
-     *     Positive strides are forward differences, negative strides are backward differences.
      * @param values input vector.
      * @return difference between elements.
      */
@@ -1628,7 +1626,6 @@ public class Numeric {
      * Differences off the end of the input vector are null.
      *
      * @param stride number of elements to skip between consecutive elements.
-     *     Positive strides are forward differences, negative strides are backward differences.
      * @param values input vector.
      * @return difference between elements.
      */
--- a/engine/function/src/templates/Numeric.ftl
+++ b/engine/function/src/templates/Numeric.ftl
@@ -1589,6 +1589,76 @@ public class Numeric {
     }
 
     /**
+     * Returns the difference between elements in the input vector.
+     * A stride of 2 returns v(i)=e(i+2)-e(i), where v(i) is the ith computed value e(i) is the ith input value.
+     * A stride of -2 returns v(i)=e(i-2)-e(i), where v(i) is the ith computed value e(i) is the ith input value.
+     * The result is the same length as the input vector.
+     * Differences off the end of the input vector are null.
+     *
+     * @param stride number of elements to skip between consecutive elements.
+     *     Positive strides are forward differences, negative strides are backward differences.
+     * @param values input vector.
+     * @return difference between elements.
+     */
+    public static ${pt.primitive}[] diff(int stride, ${pt.boxed}[] values) {
+        return diff(stride, unbox(values));
+    }
+
+    /**
+     * Returns the difference between elements in the input vector.
+     * A stride of 2 returns v(i)=e(i+2)-e(i), where v(i) is the ith computed value e(i) is the ith input value.
+     * A stride of -2 returns v(i)=e(i-2)-e(i), where v(i) is the ith computed value e(i) is the ith input value.
+     * The result is the same length as the input vector.
+     * Differences off the end of the input vector are null.
+     *
+     * @param stride number of elements to skip between consecutive elements.
+     *     Positive strides are forward differences, negative strides are backward differences.
+     * @param values input vector.
+     * @return difference between elements.
+     */
+    public static ${pt.primitive}[] diff(int stride, ${pt.primitive}... values) {
+        return diff(stride, new ${pt.vectorDirect}(values));
+    }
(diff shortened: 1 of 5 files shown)


engine/function/src/templates/Sort.ftl
```diff
@@ -202,6 +318,63 @@ public class Sort {
         return vs;
     }
 
+    /**
+     * Returns the indices of values sorted from smallest to largest.
+     *
+     * @param values values.
+     * @return sorted indices.
+     */
+    public static int[] sortIndex(final ${pt.vector} values) {
```
rbasralian, line 327: I'd prefer calling this `rank()`, since it's shorter and it's the name both pandas and kx use

engine/function/src/templates/Sort.ftl
```diff
@@ -202,6 +318,63 @@ public class Sort {
...
+        if (values == null) {
+            return null;
+        }
+
+        if (values.isEmpty()) {
+            return new int[]{};
+        }
+
+        return IntStream.range(0, values.intSize("sortIndex"))
+            .boxed().sorted((i, j) -> ${pt.boxed}.compare(values.get(i),values.get(j)) )
```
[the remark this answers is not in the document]
chipkent: Resolved, but it came with boxing and copies.

rbasralian (approved):

engine/function/src/templates/Numeric.ftl
```diff
@@ -1588,6 +1588,73 @@ public class Numeric {
         return product(new ${pt.vectorDirect}(values));
     }
 
+    /**
+     * Returns the differences between elements in the input vector separated by a stride.
+     * A stride of k returns v(i)=e(i+k)-e(i), where v(i) is the ith computed value e(i) is the ith input value.
```
kosak, line 1593: Is there a connecting word missing here before e(i)? like "and"?

where v(i) is the ith computed value, and e(i) is the ith input value.

engine/function/src/templates/Numeric.ftl
```diff
@@ -2788,6 +2855,55 @@ public class Numeric {
...
     </#if>
 
+
+    /**
+     * Compares two specified values.  Deephaven null values are less than normal numbers which are less than NaN values.
+     *
+     * @param v1 the first value to compare.
+     * @param v2 the second value to compare.
+     * @returns the value 0 if v1 is numerically equal to v2; a value of less than 0 if v1 is numerically less than v2; and a value greater than 0 if v1 is numerically greater than v2.
+     *      Deephaven null values are less than normal numbers which are less than NaN values.
```
kosak, line 2865: We *might* want to add a line to assuage any IEEE 754 hardliners among our customers. Something like:

Unlike standard Java, we treat NaN values as ordered. In particular two NaN values will compare equal to each other, and a NaN value will compare greater than any other value.

engine/function/src/templates/Sort.ftl
```diff
@@ -138,6 +198,63 @@ public class Sort {
...
+     * @param comparator value comparator.
+     * @return sorted indices.
+     */
+    static public <T extends Comparable<? super T>> int[] rankDescendingObj(final ObjectVector<T> values, final Comparator<T> comparator) {
+        if (values == null) {
+            return null;
+        }
+
+        if (values.isEmpty()) {
+            return new int[]{};
```
kosak, line 214: micro-nit: maybe better as `new int[0]`?

engine/function/src/templates/Sort.ftl
```diff
@@ -193,13 +310,66 @@ public class Sort {
...
+     * @param values values.
+     * @return sorted indices.
+     */
+    public static int[] rank(final ${pt.vector} values) {
+        if (values == null) {
+            return null;
+        }
+
+        if (values.isEmpty()) {
+            return new int[]{};
```
kosak, line 330: micro-nit: maybe better as `new int[0]`?

engine/function/src/templates/Sort.ftl
```diff
@@ -193,13 +310,66 @@ public class Sort {
...
+     * @param values values.
+     * @return sorted indices.
+     */
+    public static int[] rank(final ${pt.boxed}[] values) {
+        if (values == null) {
+            return null;
+        }
+
+        if (values.length == 0) {
+            return new int[]{};
```
kosak, line 364: micro-nit: maybe better as `new int[0]`?

engine/function/src/templates/Sort.ftl
```diff
@@ -254,6 +422,58 @@ public class Sort {
...
+     * @param values values.
+     * @return sorted indices.
+     */
+    public static int[] rankDescending(final ${pt.vector} values) {
+        if (values == null) {
+            return null;
+        }
+
+        if (values.isEmpty()) {
+            return new int[]{};
```
kosak, line 437: micro-nit: maybe better as `new int[0]`?

kosak (commented):
I have a couple of small comments and micro-nits.

kosak (approved):

rbasralian (approved):

deephaven-internal:

Community: https://github.com/deephaven/deephaven-docs-community/issues/173

PR #5153, opened 2024-02-14
GRPC support for MultiJoin, improved support for RangeJoin

### LINKED ISSUE #4281: GRPC support for `MultiJoin` feature
As a user, I want to create and use `MultiJoin` tables from GRPC clients.

lbooker42:
This PR adds GRPC support for `MultiJoin`, allowing clients to efficiently join multiple tables.  It also improves support for `RangeJoin` by exposing a `string` input for specifying the match criteria.

Note that this implementation directly returns a `Table` as opposed to the server-side calls which return `MultiJoinTable` objects.  The more complex interface can be exposed to GRPC along-side this implementation when we add some additional desired functionality to `MultiJoinTable`.

Closes #4281


nbauernfeind (dismissed):

server/src/main/java/io/deephaven/server/table/ops/RangeJoinGrpcImpl.java
```diff
@@ -82,7 +108,12 @@ public Table create(RangeJoinTablesRequest request, List<ExportObject<Table>> so
         final Table leftTable = sourceTables.get(0).get();
         final Table rightTable = sourceTables.get(1).get();
         final Collection<JoinMatch> exactMatches = JoinMatch.from(request.getExactMatchColumnsList());
-        final RangeJoinMatch rangeMatch = parseRangeMatch(request);
+        final RangeJoinMatch rangeMatch;
+        if (request.getRangeMatch().isEmpty()) {
```
devinrsmith, line 112: Regardless of whether we check empty or presence bit, I think we should have a helper function for the few callsites that are doing this.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -765,12 +784,17 @@ message RangeJoinTablesRequest {
...
   string left_start_column = 5;
   RangeStartRule range_start_rule = 6;
   string right_range_column = 7;
   RangeEndRule range_end_rule = 8;
   string left_end_column = 9;
   repeated Aggregation aggregations = 10;
+  // Specifies the range match parameters as a parseable string. Providing `range_match` in the GRPC call is the
+  // alternative to detailed range match parameters provided in the `range_start_column`, `range_start_rule`,
+  // `right_range_column`, `range_end_rule`, and `left_end_column` fields.
+  string range_match = 11;
```
devinrsmith, line 797: Do we want to expose this to `io.deephaven.qst.table.RangeJoinTable`? (The answer may be no.)

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -748,6 +753,20 @@ message AjRajTablesRequest {
   repeated string columns_to_add = 6;
 }
 
+message MultiJoinInput {
+  TableReference source_id = 1;
+  repeated string columns_to_match = 3;
+  repeated string columns_to_add = 4;
+}
+
+message MultiJoinTablesRequest {
```
devinrsmith, line 762: Is now the time to plumb this through `io.deephaven.qst` / `io.deephaven.api.TableOperations`? If not, please create a ticket.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -748,6 +753,20 @@ message AjRajTablesRequest {
   repeated string columns_to_add = 6;
 }
 
+message MultiJoinInput {
+  TableReference source_id = 1;
+  repeated string columns_to_match = 3;
+  repeated string columns_to_add = 4;
```
niloc132, line 759: where did 2 go?
```suggestion
  repeated string columns_to_match = 2;
  repeated string columns_to_add = 3;
```

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
```diff
@@ -748,6 +753,20 @@ message AjRajTablesRequest {
...
+}
+
+message MultiJoinTablesRequest {
+  Ticket result_id = 1;
+  // It is considered an error if both `source_ids` and `multi_join_inputs` are provided.
```
niloc132, line 764: As a comment, this will appear on the `source_ids` field in each language, rather than be general for the body of the message. Consider moving to just before the message itself, or duplicate it to the `multi_join_inputs` as well?

niloc132: ...that said, this might be better expressed as a `oneof`? Reading the logic in the java server impl, if `multi_join_inputs` is provided, neither `source_ids` nor `columns_to_match` can be specified (i.e. "will be empty")?

I'm not above just making the runtime logic encode this instead of the message format, but there is something to be said for making the serialization api simply _not support_ invalid states.

lbooker42: as @devinrsmith pointed out, it's really `(source_ids + columns_to_match) ^ multi_join_inputs` that needs to be enforced.  `oneof` doesn't seem to work here without adding another nesting layer which I think would break backward compatibility.

devinrsmith: What do you mean "break backward compatibility" - this is ostensibly a brand new message, yeah?

server/src/main/java/io/deephaven/server/table/ops/MultiJoinGrpcImpl.java
```diff
@@ -47,6 +51,10 @@ public void validateRequest(final MultiJoinTablesRequest request) throws StatusR
             throw Exceptions.statusRuntimeException(Code.INVALID_ARGUMENT,
                     "If `multi_join_inputs` are provided, `columns_to_match` must remain empty.");
         }
+
+        // Validate well-formed source table ids.
+        request.getSourceIdsList().forEach(Common::validate);
+        request.getMultiJoinInputsList().forEach(input -> Common.validate(input.getSourceId()));
```
devinrsmith, line 57: We should ideally also check each MultiJoinInput has no unknown fields.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
lbooker42: Good point.  RangeJoin is published, MultiJoin is new.

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
lbooker42: Created https://github.com/deephaven/deephaven-core/issues/5158

proto/proto-backplane-grpc/src/main/proto/deephaven/proto/table.proto
lbooker42: I don't think so, this adds a shortcut / additional way to populate the `RangeJoinMatch` but doesn't replace it.

devinrsmith (approved):

niloc132 (approved):
Approved of proto structure, documentation.

rcaudy (approved):

PR #5166, opened 2024-02-18
Fix test failures from wrapping job scheduler errors in `TableInitializationExceptions`

nbauernfeind:
Nightlies: https://github.com/nbauernfeind/deephaven-core/actions/runs/7947180962/

### DIFF
--- a/engine/table/src/test/java/io/deephaven/engine/table/impl/updateby/TestEmMinMax.java
+++ b/engine/table/src/test/java/io/deephaven/engine/table/impl/updateby/TestEmMinMax.java
@@ -4,10 +4,12 @@ import io.deephaven.api.updateby.BadDataBehavior;
 import io.deephaven.api.updateby.OperationControl;
 import io.deephaven.api.updateby.UpdateByControl;
 import io.deephaven.api.updateby.UpdateByOperation;
+import io.deephaven.base.verify.Assert;
 import io.deephaven.chunk.Chunk;
 import io.deephaven.chunk.ObjectChunk;
 import io.deephaven.chunk.attributes.Values;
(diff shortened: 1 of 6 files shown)


engine/table/src/test/java/io/deephaven/engine/table/impl/updateby/TestEmMinMax.java
```diff
@@ -356,73 +358,103 @@ public void testThrowBehaviors() {
...
                         UpdateByOperation.EmMin(OperationControl.builder().onNullValue(BadDataBehavior.THROW).build(),
-                                10)),
-                "Encountered null value during EMS processing");
-
-        assertThrows(TableDataException.class,
+                                10)));
+        err = err.getCause();
+        Assert.eqTrue(err.getClass() == TableDataException.class,
+                "err.getClass() == TableDataException.class");
+        Assert.eqTrue(err.getMessage().contains("Encountered null value during Exponential Moving output processing"),
```
rcaudy, line 406: This doesn't match the previous message. I don't recall a change to the message in your original PR.

rcaudy (commented):
Ready to approve, just want to understand the error message validation changes. Are we maybe validating the cause message instead of the wrapper message now?

engine/table/src/test/java/io/deephaven/engine/table/impl/updateby/TestEmMinMax.java
rcaudy: Ah. It appears that the old code was not validating anything at all. The message was for the assertion error, not for validation. I don't love message validations, but at least I see why this is correct now.

rcaudy (approved):

PR #5167, opened 2024-02-19
Use reader's schema for kafka avro deserialization

### LINKED ISSUE #5163: avro schema changes
The current kafka avro implementation does not explicitly account for avro schema changes. In the best case, the parsing logic _may_ continue to work, implicitly correct. In the worst case, the parsing logic _may_ continue running, implicitly putting the wrong data into a column. In some cases, the parsing logic will cause the ingestion to crash. Ultimately, this is due to the implementation using parsing logic based on a single `org.apache.avro.Schema`. The kafka serdes protocol allows for a schema id on a per-message basis: https://docs.confluent.io/platform/current/schema-registry/fundamentals/serdes-develop/index.html#messages-wire-format

A typed function prototype was developed (based on some code-generation that @cpwright wrote) that bypasses the GenericRecord altogether, and may relevant to a solution here. https://github.com/devinrsmith/deephaven-core/tree/avro-raw-impl

devinrsmith:
Fixes #5163

### DIFF
--- a/extensions/kafka/src/main/java/io/deephaven/kafka/AvroImpl.java
+++ b/extensions/kafka/src/main/java/io/deephaven/kafka/AvroImpl.java
@@ -37,6 +37,7 @@ import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.util.Utf8;
 import org.apache.commons.lang3.mutable.MutableInt;
 import org.apache.commons.lang3.mutable.MutableObject;
+import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.serialization.Deserializer;
 import org.apache.kafka.common.serialization.Serializer;
 import org.jetbrains.annotations.NotNull;
@@ -66,7 +67,7 @@ class AvroImpl {
         private static final Pattern NESTED_FIELD_NAME_SEPARATOR_PATTERN =
                 Pattern.compile(Pattern.quote(NESTED_FIELD_NAME_SEPARATOR));
 
-        private final Schema schema;
+        private Schema schema;
         private final String schemaName;
         private final String schemaVersion;
         /** fields mapped to null are skipped. */
@@ -75,7 +76,7 @@ class AvroImpl {
         private final boolean useUTF8Strings;
 
         AvroConsume(final Schema schema, final Function<String, String> fieldPathToColumnName) {
-            this.schema = schema;
+            this.schema = Objects.requireNonNull(schema);
             this.schemaName = null;
             this.schemaVersion = null;
             this.fieldPathToColumnName = fieldPathToColumnName;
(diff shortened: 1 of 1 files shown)


devinrsmith:
For reference, the schema transitioned from

```json
{
    "type": "record",
    "name": "yellow3",
    "fields": [
        {
            "name": "passenger_count",
            "type": [
                "int",
                "null"
            ]
        },
        {
            "name": "VendorID",
            "type": [
                "long",
                "null"
            ]
        }
    ]
}
```

to

```json
{
    "type": "record",
    "name": "yellow3",
    "fields": [
        {
            "name": "VendorID",
            "type": [
                "long",
                "null"
            ]
        },
        {
            "name": "passenger_count",
            "type": [
                "int",
                "null"
            ]
        }
    ]
}
```

in the testing I did. Other sorts of schema updates are very likely to cause similar issues without this fix.

jcferretti:
Ref your schema change (in the 'for reference...' comment above).

I mentioned over slack, a schema update where you don't add a default value for a new field is not supposed to work:

https://docs.confluent.io/platform/current/schema-registry/fundamentals/schema-evolution.html#backward-compatibility

```
Note that the new field favorite_color has the default value "green". This allows data encoded with the old schema to be read with the new one. The default value specified in the new schema will be used for the missing field when deserializing the data encoded with the old schema. Had the default value been omitted in the new field, the new schema would not be backward compatible with the old one since it's not clear what value should be assigned to the new field, which is missing in the old data.
```

rcaudy (commented):
Looks OK. `KafkaAvroDeserializer` seems to be doing an awful lot of work, however. Is there a way to get it to cache migrations, etc?

jcferretti:
How did you make the change in schema registry?  Did you create a new version of the schema, or did you remove/overwrite?

jcferretti:
https://github.com/confluentinc/schema-registry/blob/802dc1a4dfbca7919375e28dfe2cbe793ae57e0c/core/src/test/java/io/confluent/kafka/schemaregistry/rest/json/RestApiTest.java#L240

https://github.com/confluentinc/schema-registry/pull/2064

devinrsmith:
> Looks OK. `KafkaAvroDeserializer` seems to be doing an awful lot of work, however. Is there a way to get it to cache migrations, etc?

I believe "migrations" are a confluent hooks that allow more control over the transformation of data. Afaik, we don't use them, so they should always be empty.

devinrsmith:
I've verified in main today and in this fix, the majority of the migration / rule code is bypassed in the deserializer.

jcferretti:
> > How did you make the change in schema registry? Did you create a new version of the schema, or did you remove/overwrite? Did you do the test on redpanda's schema registry?
> 
> I did a couple of manual _updates_ (not remove / overwrite) through the Redpanda Console UI, but also went through DHC avro producer (on that front, filed #5168).

The specific question here for me is, was this change done as a new version of the same schema id in schema service.  The reason I am asking is because in my reading of the docs, that should just work without the need for us to change any code.  Also I am curious about the validation that schema service is supposed to do in updates (new version of same schema) and whether it would swallow a change that is just about ordering of records (which should be not really a new version as per my reading of the docs).

A second question is whether you have tested adding a new field with a default value (an actual legal backwards compatible schema update), as opposed to just a reordering of existing fields.

jcferretti:
I was reading the sources around KafkaAvroDeserializer in hoping we could do this in a simpler way, but I am close to convincing myself there isn't one if we still use GenericRecord.

The constructor does seem to use the reader schema without the need to do anything else when using SpecificRecord:
(https://github.com/confluentinc/schema-registry/blob/0288317a8823e57f42d0b1c4c06f46a9f5ec9969/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroDeserializer.java#L120 
But alas, we don't use that.

jcferretti (approved):

devinrsmith:
> I was reading the sources around KafkaAvroDeserializer in hoping we could do this in a simpler way, but I am close to convincing myself there isn't one if we still use GenericRecord.
> 
> The constructor does seem to use the reader schema without the need to do anything else when using SpecificRecord: (https://github.com/confluentinc/schema-registry/blob/0288317a8823e57f42d0b1c4c06f46a9f5ec9969/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroDeserializer.java#L120 But alas, we don't use that.

In general, we _can't_ use that unless the generated java class is on the classpath. In protobuf, there is a very similar bifurcation (`io.deephaven.kafka.protobuf.DescriptorSchemaRegistry` vs `io.deephaven.kafka.protobuf.DescriptorMessageClass`). We could theoretically take advantage of generated java classes / SpecificRecord if we 1) updated our internal code to work against `IndexedRecord`, and 2) had a way to allow end user to configure as such.

PR #5169, opened 2024-02-20
Make 'numba' a soft dependency for the Python server API

### LINKED ISSUE #3082: Clarify dependency requirements wrt numba
`numba` used to be a "soft" dependency - we installed it for compatible python versions by default, but we didn't have any "hard" dependency on it. This changed as of https://github.com/deephaven/deephaven-core/pull/4502. We should either:

1. Declare `numba` a "hard" dependency (ie, remove the version qualifier around our dependency `numba; python_version < "..."`
2. Fix our code to work gracefully when `numba` is not installed

It's worth noting that declaring a hard dependency on `numba` tighly couples our supportable matrix and delivery timeline to numba, which has typically lagged 6 months or more for new python releases.

### LINKED ISSUE #5133: Recent Python UDF change added a hard dependency of numba
Which makes experimenting DH with newer, not-yet-supported-by-numba Python versions almost impossible.

Related to #3082, which is more about test code's dependency on numba

jmao-denver:
Fixes #5133
Fixes #3082

### DIFF
--- a/py/server/deephaven/_dep.py
+++ b/py/server/deephaven/_dep.py
@@ -1,0 +1,21 @@
+#
+#     Copyright (c) 2016-2023 Deephaven Data Labs and Patent Pending
+#
+"""This module is an internal module to support dependencies management"""
+
+import importlib
+from types import ModuleType
+from typing import Optional
+
+
+def soft_dependency(module_name) -> Optional[ModuleType]:
+    """Attempt to import a module and return it if it exists, otherwise return None.
+    Args:
+        module_name (str): the name of the module to import
+    Returns:
+        the module if it exists, otherwise None
+    """
+    try:
+        return importlib.import_module(module_name)
+    except ImportError:
+        return None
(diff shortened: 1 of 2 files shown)


py/server/deephaven/_dep.py
```diff
@@ -0,0 +1,21 @@
+#
+#     Copyright (c) 2016-2023 Deephaven Data Labs and Patent Pending
+#
+"""This module is an internal module to support dependencies management"""
```
chipkent, line 4:
```suggestion
"""This module is an internal module to support dependency management"""
```

py/server/deephaven/_dep.py
```diff
@@ -0,0 +1,21 @@
...
+
+import importlib
+from types import ModuleType
+from typing import Optional
+
+
+def soft_dependency(module_name) -> Optional[ModuleType]:
```
chipkent, line 11:
```suggestion
def soft_dependency(module_name: str) -> Optional[ModuleType]:
```

chipkent (approved):

devinrsmith (approved):

PR #5176, opened 2024-02-21
Added support to read mixed encoded parquet files

### LINKED ISSUE #5162: Failure to read Parquet file with mixed dictionary and plain encoding
**Description**

A Core+ customer has a Parquet file that can be read with PyArrow and Pandas,  but throws a CCE because of mixed dictionary encoding when we try to read it with Core Parquet.read().  We have done a bit of digging in the parquet metadata (attached)

The issue exists only for the single file, and has something to do with the change of encoding from dictionary to plain.  There doesn’t seem to be anything really interesting in any of the meta outputs, other than the fact that after 3 dictionary pages it falls back to plain encoding.

We generated several Parquet files using both PyArrow and Pandas that had the same structure (same dictionary size, same number of initial dictionary pages etc). None of which Core had any issues with.  There is something very specific about the file that Core does not like.  I’ve requested the file itself, but I don’t think we will get it, given that it contains price data.

We used variations of the following script code to generate files
```python
from deephaven import numpy as dhnp
import numpy as np
import pyarrow as pa
import pyarrow.parquet as pq

q = empty_table(2000000).update("Broken=String.format(`%015d`,  ii < 1200000 ? (ii % 30000) : ii)")
np_source = dhnp.to_numpy(table=q, cols=["Broken"])
ndarray_table = pa.table({"Broken": np_source[:, 0],})
pa.parquet.write_table(ndarray_table, "/tmp/ndarray.parquet")
```
if you vary the conditional you can change the dictionary size and how many pages of dictionary you end up with.

**Expected results**

We read the data fine

**Actual results**

Class cast exception
image-20240213-205242

[pages2.txt](https://github.com/deephaven/deephaven-core/files/14312174/pages2.txt)
[meta.txt](https://github.com/deephaven/deephaven-core/files/14312175/meta.txt)
[footer.txt](https://github.com/deephaven/deephaven-core/files/14312176/footer.txt)
[check-stats.txt](https://github.com/deephaven/deephaven-core/files/14312177/check-stats.txt)
[pages.txt](https://github.com/deephaven/deephaven-core/files/14312178/pages.txt)

**Versions**
 - Deephaven: 0.30.4
 - OS: Windows
 - Browser: Chrome
 - Docker: N/A

malhotrashivam:
Closes #5162


extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReaderImpl.java
```diff
@@ -294,58 +287,51 @@ private static org.apache.parquet.format.Encoding getEncoding(PageHeader pageHea
...
+        }
+    }
+
     private static int getNumValues(PageHeader pageHeader) {
         return pageHeader.isSetData_page_header()
                 ? pageHeader.getData_page_header().getNum_values()
                 : pageHeader.getData_page_header_v2().getNum_values();
     }
 
-    private final class ColumnPageReaderIteratorIndexImpl implements ColumnPageReaderIterator {
```
malhotrashivam, line 303: To give you some context behind this class, we need to understand the two page reader implementations which are used for accessing any page in the file:
- `VariablePageSizeColumnChunkPageStore`: To access page i, this class iterates through all the pages leading upto i and builds a reader objects.
Internally, this class can use two implementations of ColumnPageReaderIterator:
    - `ColumnPageReaderIteratorImpl`: This class will read the page header for each page, decides whether each page needs a dictionary and then finally builds and returns reader objects.
    - `ColumnPageReaderIteratorIndexImpl`: This class uses offset index to locate the pages and thus does not need read the page header. It can directly find offset of each page in the file and build a reader object.
- `OffsetIndexBasedColumnChunkPageStore`: To access page i, this class directly uses offset index to locate the page in the file and builds a reader objects. Thus no iterating needed.

So now your question might be, when do we use each of them:
- If the parquet file has offset index and doesn't have the offset-index bug from deephaven version below 0.31.0, then we use `OffsetIndexBasedColumnChunkPageStore` (check `ColumnChunkPageStore::canUseOffsetIndexBasedPageStore` for more details).
- Else we use `VariablePageSizeColumnChunkPageStore`
      - If the parquet file has offset index, we use `ColumnPageReaderIteratorIndexImpl`
      - If the parquet file doesn't have offset index, we use `ColumnPageReaderIteratorImpl`

So overall `ColumnPageReaderIteratorIndexImpl` is used to handle parquet files written by deephaven version below 0.31.0 with buggy offset index information.
Note that these files can also be read by `ColumnPageReaderIteratorImpl` because it doesn't even use the offset index information. So is not affected by the bug. But we used `ColumnPageReaderIteratorIndexImpl` because it was more efficient in that it didn't need to read the page header.

But now, after this new bug, we know that we always need to read the page header to decide dictionary encoding of each page. Therefore, I decided to remove `ColumnPageReaderIteratorIndexImpl` because now there is no performance advantage over `ColumnPageReaderIteratorImpl`.

And this helped me make more improvements in how we compute data offset and numValues. Basically, now since we are always reading page header up front, we can directly compute the offset of data in the page and the number of values in the page.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java
```diff
@@ -95,8 +90,10 @@ final class ColumnPageReaderImpl implements ColumnPageReader {
         this.path = path;
         this.uri = uri;
         this.fieldTypes = fieldTypes;
-        this.offset = offset;
+        this.dataOffset = dataOffset;
```
malhotrashivam, line 93: Now we always get the offset of the data in the page, page header and number of values at the time of construction.

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java
```diff
@@ -126,51 +123,11 @@ public IntBuffer readKeyValues(IntBuffer keyDest, int nullPlaceholder,
...
-    private static int readNumValuesFromPageHeader(@NotNull final PageHeader header) throws IOException {
-        switch (header.type) {
-            case DATA_PAGE:
-                return header.getData_page_header().getNum_values();
-            case DATA_PAGE_V2:
-                return header.getData_page_header_v2().getNum_values();
-            default:
-                throw new IOException(String.format("Unexpected page of type {%s}", header.getType()));
-        }
-    }
```
malhotrashivam, line 172: No need for all this because this is already done at the time of constructing this reader object.

extensions/s3/src/main/java/io/deephaven/extensions/s3/BufferPool.java
```diff
@@ -1,9 +1,7 @@
 package io.deephaven.extensions.s3;
 
 import io.deephaven.base.reference.PooledObjectReference;
-import io.deephaven.base.reference.SimpleReference;
```
malhotrashivam, line 4: Should I remove this change if we plan to back port this fix, to avoid merge conflicts?

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnPageReaderImpl.java
```diff
@@ -95,8 +90,10 @@ final class ColumnPageReaderImpl implements ColumnPageReader {
         this.path = path;
         this.uri = uri;
         this.fieldTypes = fieldTypes;
-        this.offset = offset;
+        this.dataOffset = dataOffset;
+        Assert.neqNull(pageHeader, "pageHeader");
         this.pageHeader = pageHeader;
+        Assert.geqZero(numValues, "numValues");
         this.numValues = numValues;
```
rcaudy, line 97:
```suggestion
        this.pageHeader = Require.neqNull(pageHeader, "pageHeader");
        this.numValues = Require.geqZero(numValues, "numValues");
```

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReaderImpl.java
```diff
@@ -230,9 +226,9 @@ private final class ColumnPageReaderIteratorImpl implements ColumnPageReaderIter
         private long nextHeaderOffset;
         private long remainingValues;
 
-        ColumnPageReaderIteratorImpl(final long startOffset, final long numValues) {
+        ColumnPageReaderIteratorImpl(final long dataPageStartOffset, final long numValues) {
```
rcaudy, line 229:
```suggestion
        ColumnPageReaderIteratorImpl(final long firstHeaderOffset, final long numValues) {
```

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReaderImpl.java
```diff
@@ -294,58 +287,51 @@ private static org.apache.parquet.format.Encoding getEncoding(PageHeader pageHea
...
+                    throw new IllegalStateException("Expected data page, but got " + pageType + " for page number "
+                            + pageNum + " for column " + getURI());
+                }
+                final Function<SeekableChannelContext, Dictionary> pageDictionarySupplier =
+                        getPageDictionarySupplier(pageHeader);
+                return new ColumnPageReaderImpl(channelsProvider, decompressor, pageDictionarySupplier,
+                        nullMaterializerFactory, path, getURI(), fieldTypes, dataOffset, pageHeader,
+                        getNumValues(pageHeader));
+            } catch (final IOException e) {
+                throw new UncheckedDeephavenException("Error reading page header for page number " + pageNum + " for " +
```
rcaudy, line 332: I would use this better error message in the iterator case as well, after correcting "column" to "file".

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReaderImpl.java
```diff
@@ -251,37 +247,34 @@ public ColumnPageReader next(@NotNull final SeekableChannelContext channelContex
...
-                    // Dictionary page; skip it
-                    return next(holder.get());
-                }
-                if (!pageHeader.isSetData_page_header() && !pageHeader.isSetData_page_header_v2()) {
-                    throw new IllegalStateException(
-                            "Expected data page, but neither v1 nor v2 data page header is set in file "
-                                    + ch + " at offset " + headerOffset);
+                final PageType pageType = pageHeader.type;
+                if (pageType != PageType.DATA_PAGE && pageType != PageType.DATA_PAGE_V2) {
+                    throw new IllegalStateException("Expected data page, but got " + pageType + " in file " + ch +
```
rcaudy, line 256:
```suggestion
                    throw new IllegalStateException("Expected data page, but got " + pageType + " in file " + getURI() +
```

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReaderImpl.java
```diff
@@ -294,58 +287,51 @@ private static org.apache.parquet.format.Encoding getEncoding(PageHeader pageHea
...
+            final long headerOffset = offsetIndex.getOffset(pageNum);
+            try (
+                    final ContextHolder holder = SeekableChannelContext.ensureContext(channelsProvider, channelContext);
+                    final SeekableByteChannel ch = channelsProvider.getReadChannel(holder.get(), getURI())) {
+                ch.position(headerOffset);
+                final PageHeader pageHeader = readPageHeader(ch);
+                final long dataOffset = ch.position();
+                final PageType pageType = pageHeader.type;
+                if (pageType != PageType.DATA_PAGE && pageType != PageType.DATA_PAGE_V2) {
+                    throw new IllegalStateException("Expected data page, but got " + pageType + " for page number "
```
rcaudy, line 323: Make this match the other one. URI is a file, not a column.

rcaudy (approved):

extensions/parquet/base/src/main/java/io/deephaven/parquet/base/ColumnChunkReaderImpl.java
```diff
@@ -252,9 +251,15 @@ public ColumnPageReader next(@NotNull final SeekableChannelContext channelContex
                 final long dataOffset = ch.position();
                 nextHeaderOffset = dataOffset + pageHeader.getCompressed_page_size();
                 final PageType pageType = pageHeader.type;
+                if (pageType == PageType.DICTIONARY_PAGE && headerOffset == columnChunk.meta_data.getData_page_offset()
+                        && columnChunk.meta_data.getDictionary_page_offset() == 0) {
+                    // https://stackoverflow.com/questions/55225108/why-is-dictionary-page-offset-0-for-plain-dictionary-encoding
+                    // Skip the dictionary page and jump to the data page
+                    return next(holder.get());
```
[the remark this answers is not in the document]
malhotrashivam: Yea, I can add an explicit check for this here.

rcaudy (approved):

PR #5187, opened 2024-02-22
Remove old benchmark script

devinrsmith:


stanbrub (commented):
The README.md indicates that this was part of a nightly workflow.  Were references to this code already removed elsewhere?

devinrsmith:
Yep, looks like we probably should have deleted these files as part of https://github.com/deephaven/deephaven-core/pull/3981

stanbrub (approved):

PR #5190, opened 2024-02-23
Include all Right Table Values in Match Column

### LINKED ISSUE #5184: Full Outer Join Does not Properly Display Right-Values on Joined Column Without Match
```python
from deephaven.experimental.outer_joins import full_outer_join, left_outer_join

left = empty_table(5).update(["X1 = ii", "Y = Math.sin(X1)"])
right = empty_table(5).update(["X2 = ii * 2", "Y = Math.cos(X2)"])

result = full_outer_join(l_table=left, r_table=right, on=["X1 = X2"], joins=["Z = Y"])
```
Screenshot 2024-02-22 at 8 37 47 AM

It does work, however, if you first rename the right column:
```python
from deephaven.experimental.outer_joins import full_outer_join, left_outer_join

left = empty_table(5).update(["X1 = ii", "Y = Math.sin(X1)"])
right = empty_table(5).update(["X2 = ii * 2", "Y = Math.cos(X2)"])

result = full_outer_join(l_table=left, r_table=right.rename_columns(["X1 = X2"]), on=["X1"], joins=["Z = Y"])
```
Screenshot 2024-02-22 at 8 38 01 AM

Thanks to @robbcamera for reporting.

nbauernfeind:
Fixes #5184.

I had kept the behavior from DHE, but I suspect we do actually want for match columns to show the right values when there are no rows.

### DIFF
--- a/engine/table/src/main/java/io/deephaven/engine/util/OuterJoinTools.java
+++ b/engine/table/src/main/java/io/deephaven/engine/util/OuterJoinTools.java
@@ -7,7 +7,6 @@ import com.google.common.collect.Streams;
 import io.deephaven.api.Selectable;
 import io.deephaven.api.TableOperationsDefaults;
 import io.deephaven.engine.table.ColumnDefinition;
-import io.deephaven.engine.table.ColumnSource;
 import io.deephaven.engine.table.impl.MatchPair;
 import io.deephaven.engine.table.Table;
 import io.deephaven.engine.table.impl.CrossJoinHelper;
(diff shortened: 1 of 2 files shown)


rcaudy (commented):
I find this "weird". I think the commingling of values is "fine", but it feels weird to me to only use the LHS name. Should we create a new name that represents a pair of values? Should we never merge values? The code change is fine, but I want confirmation from someone more user-facing.

nbauernfeind:
This approach is being abandoned in favor of this:

- Do not include RHS columns by default (do not merge by default as is today)
- Add a union parameter that uses LHS column names and merges RHS into them (new feature)
- Allow RHS renames via the join-list (if the join-list is empty it brings all RHS columns, not jut key-columns; existing behaviors)
- Fail if joined columns have any name collisions (existing behavior)

PR #5191, opened 2024-02-23
Update publish-ci to continue on certain failures

devinrsmith:
There was an error observed during the v0.33.0 release where the deephaven-server artifact was too large for PyPi. There are other failure scenarios that we could potentially run into here as well - network connectivity, expired tokens, temporary service interruptions, etc. Given that all of the release artifacts have already been successfully uploaded as a GitHub artifact, we should prefer to continue with publishing to all the sources even if one of the sources fails.

### DIFF
--- a/.github/workflows/publish-ci.yml
+++ b/.github/workflows/publish-ci.yml
@@ -89,6 +89,7 @@ jobs:
           user: __token__
           password: ${{ secrets.DEEPHAVENCORE_PYPI_TOKEN }}
           packages_dir: py/server/build/wheel/
+        continue-on-error: true
 
       - name: Publish deephaven-server to PyPi
         if: ${{ startsWith(github.ref, 'refs/heads/release/v') }}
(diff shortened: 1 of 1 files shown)


rcaudy (approved):
Approving as code-owner. Please defer to @stanbrub for a review before merge.

stanbrub:
I assume this would still show up as a failed run so that it doesn't get overlooked?  There is a section in the README.md to verify that things like wheels have been published, so that is covered somewhat.

devinrsmith:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

> Prevents a job from failing when a step fails. Set to true to allow a job to pass when this step fails.

So, I read this as "the step will still fail, but the job will not". I don't know if the top level job will have some sort of warning state, but we should still be able to see the individual steps as failed. Our release doc does mention:

> Once the workflow job is done, ensure all publication sources have the new artifacts.

We can be mindful during release to make sure we double-check the step states and physically check the publication sources.

stanbrub (approved):

